PR #14, opened 2018-09-07
Added line number errors in the tokenizer

crw5996:
I wasn't sure about what constituted a "column", so I just incremented it every tab.

Let me know if there's anything I need to change!

### DIFF
--- a/src/sqltokenizer.rs
+++ b/src/sqltokenizer.rs
@@ -33,7 +33,7 @@ pub enum Token {
     /// Comma
     Comma,
     /// Whitespace (space, tab, etc)
-    Whitespace,
+    Whitespace(char),
     /// Equality operator `=`
     Eq,
     /// Not Equals operator `!=` or `<>`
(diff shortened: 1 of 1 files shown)


src/sqltokenizer.rs
```diff
@@ -158,13 +162,20 @@ impl Tokenizer {
         let mut tokens: Vec<Token> = vec![];
 
         while let Some(token) = self.next_token(&mut peekable)? {
+
+            if token == Token::Whitespace('\n') {
+                self.line += 1;
+                self.col = 0;
```
andygrove: I think col should reset to 1 here since it starts at 1 on the first line

src/sqltokenizer.rs
```diff
@@ -158,13 +162,20 @@ impl Tokenizer {
...
+            } else if token == Token::Whitespace('\t') {
+                self.col += 1;
```
andygrove: so col should increase for each character in the SQL statement, and for `\t` I think it should increase by 4 (I know that's a bit arbitrary but I just checked in a text editor and that is the default there)

Issue #15, opened 2018-09-08
CASE expressions


andygrove:
https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_6_12_case_expression

nickolay:
With the changes from #25 and #59 we support `CASE WHEN <condition> .. THEN <expr> [WHEN..THEN ...] ELSE <expr> END` and `CASE <expr> WHEN <expr> THEN ...`. We don't support "half-expressions" in WHEN like `CASE foo WHEN >1`, nor multi-value `WHEN 1,2,3`.

LucaCappelletti94:
Closing as resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/25 and https://github.com/apache/datafusion-sqlparser-rs/pull/59

PR #25, opened 2018-10-08
Support "searched" CASE expressions

nickolay:
To get acquainted with the code I tried to add something simple and settled on part of #15 - https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#searched-case

I'm sharing it in the hope it's helpful, although:

* The PR doesn't have enough tests (mainly because I didn't find a common pattern to make exhaustive tests without too much boilerplate and didn't want to invent something).
* I didn't feel too good about adding keywords to the dialects without testing the other dialects (should some keywords be shared?). (Also I'm not sure if you're aware: the keyword story in SQL is a mess - every DB has its own list of keywords and whether each can be used as an identifier, so only the absolute minimum should be included in the generic dialect, IMO - not certainly not ROW or PARQUET.)
* I'm not very familiar with Rust and tried to re-use patterns from the surrounding code.
* Calling `parse_expr` when expecting different types of expressions (e.g. boolean vs value) seemed wrong, especially since it also accepts the `SELECT` expression. I also wasn't sure if I needed to do something with binding priority, I assume not.


andygrove:
@nickolay This looks great. I will be happy to merge when the merge conflict is fixed (some other PRs were merged which caused this).

The current design of this repo is wrong in the way that we have one parser supporting multiple dialects. We should one parser per dialect instead, delegating as much as possible to the "generic" dialect or other shared code.

I was hoping to have more time to work on this and I have let this get out of control.

nickolay:
@andygrove The conflict was because git couldn't handle two new `#[test]` functions being added, so I fixed it by rebasing, making no other changes. Hope it's OK.

PR #27, opened 2018-10-16
consume semi colon at the end of select and delete queries

virattara:


andygrove:
@virattara Could you add a unit test for this please

PR #33, opened 2018-12-16
implement NOT

andygrove:
@cswinter Here you go


andygrove:
Actually there's a little more to this ... leave it with me and I'll get `NOT` and `NOT LIKE` working later today

andygrove:
So this implemented the general case of `NOT` as a unary operator and the special case of `NOT LIKE` as a binary operator (I prefer modeling it as `left NOT LIKE right` rather than `NOT (left LIKE right)` but either way would be valid.

Sadly this breaks DDL where we have things like `DEFAULT expr NOT NULL` because `NOT` is now recognized by the infix parser.

I now have to handle both cases but need to think of the least hacky way of doing this.

nickolay:
> Sadly this breaks DDL where we have things like DEFAULT expr NOT NULL because NOT is now recognized by the infix parser.

The spec [says](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#default-clause) you can only use literals, "&lt;datetime value function>", and a few keywords after DEFAULT, not any kind of expression. Using a specialized function instead of parse_expr would solve the problem.

By the way I found it confusing that parse_expr also handles things (like INSERT) that make no sense in [expression context](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_6_26_value_expression) and that it's invoked to handle things that are not expressions (like after FROM) - is that a intentional decision?

andygrove:
@nickolay yes this is intentional because queries can also be expressions e.g. `SELECT a, (SELECT 1) AS b` or `WHERE b IN (SELECT ....)`

nickolay:
@andygrove thanks. I know SELECT can be a subquery (only parenthesized though), and I was asking about non-query keywords (INSERT can't appear inside a SELECT query, can it). If that's the reason, I'll make a PR with the improvements I was thinking about sometime later.

PR #34, opened 2019-01-10
Miscellaneous fixes

nickolay:
These are a few unrelated fixes I came up with while reading through the code. See the commit messages for more details about each one.

I added tests where there wasn't sufficient coverage, except for timestamps where the code being fixed is effectively unused at this moment, since we don't have the code to handle `TIMESTAMP '...'` literals.


nickolay:
The CI failure is due to https://github.com/rust-lang/rust/issues/57488...

PR #37, opened 2019-01-13
Rework keyword/identifier parsing

nickolay:
This depends on PRs #36 and #38, I'll rebase after those are merged.

The commits view on GitHub is a mess (it seems to order the commits by the date of their first revision), but the main commits are 6295c92cca6b09bf2f5ef51ec6cdae29176545a0 and 06fe053e59685c0983d92c34b6468602c0f6230d.

The goals are:
- to pave the way towards supporting quoted/"delimited" identifiers
- to make it possible to use `parse`/`expect_keyword` machinery for non-reserved keywords (that can be used as identifiers)
- to make the generic parser (which could be used to parse SQL from existing systems) as liberal as possible by avoiding a global list of "reserved" keywords (which can't be used as identifiers) that are not necessary to prevent ambiguities in parsing.

The main changes are:
- folding Token::{Keyword, Identifier, DoubleQuotedString} into one Token::SQLWord, which has the necessary information (was it a known keyword and/or was it quoted).
- adding support for table aliases to showcase how the context-sensitive "reserved keywords" would work
- replacing dialect-specific lists of reserved keywords by the list of `ALL_KEYWORDS` and context-specific lists (only `RESERVED_FOR_TABLE_ALIAS` ATM).


andygrove:
This is great. I look forward to reviewing it once it is rebased.

nickolay:
@andygrove actually what you did (merging 'master' back into this branch) seems to have had better effect on GitHub than rebasing, I think the PR is in good state for review in its current state.

andygrove:
@nickolay merged and released in 0.2.3 ... thanks again

PR #41, opened 2019-02-16
Support parsing table & column aliases.

zhzy0077:
This PR add the support of table & column aliases.
The following SQL now can be parsed w/o error.
```SQL
SELECT c.id as ID FROM customer AS c
```

The `SQLNameAliasExpr` will change the usage. So it should be a new version of this modification.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -40,6 +40,11 @@ pub enum ASTNode {
     SQLIsNull(Box<ASTNode>),
     /// `IS NOT NULL` expression
     SQLIsNotNull(Box<ASTNode>),
+    /// Column projection with alias.
+    SQLProjectionExpr {
+        column_name: Box<ASTNode>,
+        alias: Option<String>
+    },
     /// Binary expression e.g. `1 + 1` or `foo > bar`
     SQLBinaryExpr {
         left: Box<ASTNode>,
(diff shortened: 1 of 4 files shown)


nickolay:
I don't think this is able to handle `(a+b) AS c`?

[My personal fork](https://github.com/nickolay/sqlparser-rs) (note: I'm force-pushing to it) implements both [table](https://github.com/nickolay/sqlparser-rs/commit/536fa6e428d733071c352fbd105388834197385a) and [SELECT](https://github.com/nickolay/sqlparser-rs/commit/bed03abe44035c4b3d8e9f5d6acafcddd1a90463) aliases among other things. I took a different implementation approach: instead of adding the alias parsing to `parse_expr` (and using an extra state variable to remember if we're parsing the `SELECT`, `FROM` or another clause), I [changed parse_expr to only handle real expressions](https://github.com/nickolay/sqlparser-rs/commit/b57c60a78c894c4f5d6d3621682aaaef25ef23de), and parse the aliases in context-specific methods like `parse_select_list` and `parse_table_factor`.

zhzy0077:
@nickolay You are right. I am using this crate in my project which has some urgent requirements of alias, so I did this pr with minimal modifications.
Since you are doing the same thing with a complete solution, I will close this pr and look forward to your updates!

rw:
@nickolay Thanks for your fork -- aliases work great for me. Cheers!

PR #43, opened 2019-03-08
What's the best way to merge my fork?

nickolay:
In my fork I have a number of new features and refactorings:
* Separate `SQLStatement` enum from `ASTNode` (up to 2dec65fdb45f91d45b27f5a42b9e59daa698d43e) and support parsing multiple statements separated with a semicolon (707c58ad57baac0c4323886c92f0964fca00bc7d)
* Only use parse_expr() for expressions (b57c60a78c894c4f5d6d3621682aaaef25ef23de) and related fixes (up to 07790fe4c4ecb2c4429e970a7d0231a41b5181d7)
* Assorted refactorings and changes to the AST (up to 346d1ff2e454d4df7b52798515c8609424e006f5), notably:
  * Store the way an identifier was quoted, as the easiest way to make `ast.to_string()` not produce invalid SQL (e0ceacd1adeb67c802e81feb5f6770cc2e254ae6)
  * Move the remaining ASTNode variant which doesn't make sense in the context of an expression (9967031cbaed216460b8533d951321184c79697f)
* Support CREATE VIEW (0c0cbcaff4bfaca1d3e3512e5c44659ae084e20f)
* A CLI program (2e9da53ed3fce705df82defec5944981aae1e74c)
* Tokenizer fixes (up to f958e9d3cf9cd2a8b8264dd9e93cf3bbe5c5cccd): dialect-specific identifier quoting (e.g. MS SQL's `[foo bar]`), national string literals (`N'...'`), comments.
* Support for more SQL constructs (up to 0621f8d43c19c626eaa413661186ea2a721c4392): WITH, aliases and qualified wildcards in SELECT, IN, BETWEEN, unary +/-, UNION/EXCEPT/INTERSECT, data types tweaks: the `CHAR` synonym and support `NUMERIC` not followed by `(p,s)`.
* CREATE MATERIALIZED VIEW from PR https://github.com/nickolay/sqlparser-rs/pull/1

Initially I intended to submit this as one PR per item, and deal with the conflicts arising from reviews by rebasing. Now that the pile grew quite large and others indicated they pulled from my fork, I'm not sure rebasing is a good idea. I can still submit multiple PRs to make reviewing easier.

What do you think we should do?


zimond:
Wow, I just ran into this crate and found this PR. No intention to spam but really thanks for your hard work. This fork works like a charm

andygrove:
@zimond Thanks for commenting on this .. I had not seen this PR until now.

@nickolay Apologies for missing this, and generally being unresponsive. I've been putting a lot of energy into Apache Arrow and neglecting the work here.

Maybe I should just make you a collaborator on this repo since you are driving it more than me now.

The only requirements I have in the short term are to not break compatibility without warning for datafusion and locustdb, the two dependent crates.

The description of your changes sounds good. I'll go ahead and test against datafusion from your branch and see if there are any issues.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "sqlparser"
 description = "Extensible SQL Lexer and Parser with support for ANSI SQL:2011"
-version = "0.2.2-alpha.0"
+version = "0.2.3-alpha.0"
```
andygrove: Please bump the version to `0.3.0` since these are breaking changes

andygrove:
@nickolay 

> This change looks good. Could you create a PR for this one?

I was trying to improve individual commits within the PR ... that doesn't work I guess

andygrove:
I think the best bet is going to be to rebase this, bump the version to 0.3.0 and I will merge and then create a PR to upgrade Arrow to the new version. We should get @cswinter to take a look at this too.

nickolay:
@andygrove, thanks for getting back to me; don't worry about the delays!

I merged andygrove/master (after reverting the revert from #42...) back into my fork for now. As indicated, I avoid rebasing as it will make it harder for those who pulled from my fork.

Bumped the version to 0.3.0, as requested.

You previously noted in #42 that you found a "breaking change" in this, did you mean the (intentional) AST changes or is there a problem I missed?

cswinter:
🎉

PR #46, opened 2019-04-10
Support parsing create external table.

zhzy0077:
Add the possibility to parse  the following SQL:

```sql
CREATE EXTERNAL TABLE uk_cities (
         name VARCHAR(100) NOT NULL,
         lat DOUBLE NULL,
         lng DOUBLE NULL
)
STORED AS TEXTFILE
LOCATION '/tmp/example.csv'
```


andygrove:
@nickolay Any opinion on this PR?

nickolay:
Thanks!

LGTM as well, though I'd remove the second (`parse_create_external_table_newline`) test, since it appears to differ only in whitespace (newline), which is stripped by the tokenizer anyway.

I don't think creating a separate SQLStatement variant _for `EXTERNAL` tables specifically_ is right, since most of the syntax is shared by `CREATE TABLE` and `CREATE EXTERNAL TABLE` _(I believe the only difference is that the `LOCATION` parameter is required for `EXTERNAL` tables and can't be specified for managed ones)._ I don't know how the dialect-specific DDL would be best represented though, so a separate variant, as in this PR, looks as good a starting point as any.

Syntax:
* https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-CreateTable
* https://www.cloudera.com/documentation/enterprise/6/6.2/topics/impala_create_table.html

zhzy0077:
@nickolay I think it's more intuitive if we have inheritance here. `CreateExternalTable` will be a subclass of `CreateTable` and share most fields. However I am a newbie to Rust and don't know how to represent it in rustlang.

nickolay:
Rust is “composition over inheritance”, so what you’re thinking about is not an option. The only alternative way to add EXTERNAL table parsing without way more work is to add `Option`s for the location and the file format in the existing `SQLCreateTable` variant. Either way is fine with me.

andygrove:
Thanks @nickolay for the review.

For dialect, I am keen on using Hive syntax. I would like to see official support for HiveQL in this crate eventually.

andygrove:
@zhzy0077 I agree that it makes sense to combine `SQLCreateTable` and `SQLCreateExternalTable` and use `Option` and add an attribute to indicate whether this is `EXTERNAL` or not. Just a `bool` would be fine for now.

zhzy0077:
@andygrove I have updated this pr according to your and Nickolay's advice. Thanks for the comprehensive explanation! @nickolay

PR #47, opened 2019-04-14
Support parsing unsigned data_type

zhzy0077:


nickolay:
@zhzy0077 I'm curious, which dialect did you need this for? A quick Google search indicates only MySQL supports it, did I miss something?

@andygrove Since supported data types and DDL syntax vary greatly across dialects, I wasn't sure listing them all in one location was useful and didn't submit a PR with my changes to support MSSQL-specific data types. Also in my private fork I added some code to "skip" DDL constructs I wasn't interested in, so that parsing would complete without an error and let me extract the interesting data (like the list of columns or constraints). Having to extend the AST types for the less interesting features is quite a drag.

If we want a generic cross-dialect DDL parser, I think that just storing the custom type name (e.g. `UNSIGNED INT`) as a string would be simpler. The parser would still need to be taught that in this case two keywords make up the data type name, but with this approach no AST changes would be required for most new data types.

zhzy0077:
@nickolay At work I mainly use MySQL and MariaDB. Currently I want to parse unsigned datatypes in this pr [https://github.com/apache/arrow/pull/4147](https://github.com/apache/arrow/pull/4147) and enables unsigned datatypes.
I have walk through many open-sourced SQL parser libs. Some of them treat signed as a field in `DataType` struct, so I do it in this way. 

Anyway, I think you're right. What's your opinion if we have a trait like `DataTypeResolver` for dialects to implement and have their own datatype resolving logic? (However in this way we must separate `SQLType` enum to a trait w/ many structs since dialects need to add more)

nickolay:
Thanks for the info.

Regarding the trait idea, I imagined that any user of this library would create their own enum of supported data types and have a function like `fn convert_type(sqlparser::asttypes::SQLType) -> MyType` that converts known types and bails out on unsupported ones. A quick look through the PR you referenced [seems to confirm this suspicion](https://github.com/apache/arrow/pull/4147/files#diff-8273e76b6910baa123f3a25a967af3b5R122). I'm not sure how a trait would help with this.

You could just make the parser represent unsigned types as, say, `SQLType::SQLCustom("UNSIGNED INT")` and match on that in your SQLType-to-my-own-enum function.

To be clear: I'm perfectly fine with merging this PR as is, just wanted to get some clarifications if a similar PR adding some MSSQLisms like `nvarchar` would be welcome.

andygrove:
I really would like to allocate some time to work on the design so we can support multiple dialects in an elegant way. This will be my next priority once I have parallel query execution working in DataFusion (I'm actively working on that now).

One idea I had was that we can have an enum for all the base types and then have a variant for vendor specific types e.g. something like

```rust
enum Types<T> where T: Dialect {
  Int,
  Varchar(usize),
  CustomType<T>
}
```

nickolay:
@andygrove that does look nice!

Note that data types is just one of the differences between the dialects' DDL - for example look at the number of options available as part of [<column_definition> in T-SQL](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017), and a `<column_definition>` is not even the only thing that can appear inside the CREATE TABLE statement: MS also supports constraints, indices, computed columns (`value AS qty * price`), something called a "column set", and `PERIOD FOR`.

[Here's](https://github.com/nickolay/sqlparser-rs/commit/4b636049f2d0105744e12a1d09e330fd221894d5) the code that's able to accept a small subset of that just for one dialect (with some comments indicating dialect-specific portions). I'm not sure how we'd split that into "core" and "dialect" parsers... Perhaps delegating the parsing of the whole `CREATE TABLE` statement to a dialect would be simpler. The "generic" dialect could then recognize these constructs but not attempt to represent them all in a fully-typed AST (perhaps storing the recognized tokens themselves instead?)

alamb:
I am closing what look like stale PRs in this repo; I apologize in advance if this is a mistake -- please feel free to reopen if you want to keep working on this issue.

PR #49, opened 2019-04-20
Support SELECT DISTINCT, and a few minor tweaks

nickolay:
The grammar [describes](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#query-specification) this as:

        <query specification> ::=
            SELECT [ <set quantifier> ] <select list> <table expression>

        <set quantifier> ::=
            DISTINCT
          | ALL

I've chosen not to add `ALL`, as it's equivalent to omitting the quantifier and I have not ever seen it used in practice (I'm also afraid that some dialects might allow ALL as a column name).

(I've also included some minor unrelated changes as separate commits in this PR.)


nickolay:
Thanks for the quick review!

Issue #52, opened 2019-04-22
Qualified wildcard to_string() is invalid


thomas-jeepe:
Line in question: https://github.com/andygrove/sqlparser-rs/blob/master/src/sqlast/mod.rs#L114

Calling to_string() on a qualified wildcard returns `Qualification*` instead of `Qualification.*`. Should be as simple as changing the line to: `q.join(".") + ".*"`

nickolay:
Thanks! While this fix should probably be applied anyway, I wonder how you found this.

Do you know which dialects allow qualified wildcards "inside" expressions, like `count(foo.*)` in your PR's testcase? ANSI SQL [defines](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_10_9_aggregate_function) `<aggregate function> ::= COUNT <left paren> <asterisk> <right paren> ...`, MSSQL agrees by reporting a syntax error in `count(foo.*)`, and the construct doesn't make sense to me. I guess `COUNT(distinct foo.*)` could make sense, but I don't know of a RDBMs that supports that either.

(The widely supported `SELECT foo.*` is handled correctly now, because we convert `ASTNode::SQLQualifiedWildcard` to `SQLSelectItem::QualifiedWildcard` [here](https://github.com/andygrove/sqlparser-rs/blob/bbf1805729d4fb0238c889d258b092e64ad4bcad/src/sqlparser.rs#L1560) -- I was initially confused how this was not caught by an existing test.)

thomas-jeepe:
I am not too sure exactly which databases support it, but it is a valid syntax.

[Stack overflow link recommending its use.](https://stackoverflow.com/questions/2710621/count-vs-count1-vs-countpk-which-is-better)

nickolay:
Thanks! no need to test, I was wondering if you knew one off-hand. Turns out Postgres [does support it](https://www.db-fiddle.com/f/j1ykWCPKzPpko966NUmweR/0), though I didn't find it mentioned in their docs.

nickolay:
The PR was merged. Thanks again!

PR #53, opened 2019-04-22
Fix qualified wildcard stringifying

thomas-jeepe:
Issue #52

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -111,7 +111,7 @@ impl ToString for ASTNode {
         match self {
             ASTNode::SQLIdentifier(s) => s.to_string(),
             ASTNode::SQLWildcard => "*".to_string(),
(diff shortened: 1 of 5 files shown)


nickolay:
I'd move the `roundtrip_qualified_wildcard` test from sqlparser_ansi to sqlparser_generic, as the latter has helper functions that let you test the round-tripping easily:

    let sql = "SELECT COUNT(Employee.*) FROM northwind.Order JOIN northwind.Employee ON Order.employee = Employee.id";
    let _select = verified_only_select(&sql);

(This would also avoid the conflict with the change to `SQLStatement::SQLSelect` in #51)

See also my question in #52 about validity of this SQL; I think a comment about whether this SQL is valid in some dialects would be helpful.

For the second patch, you might want to [run `cargo fmt`](https://github.com/rust-lang/rustfmt#quick-start), which we use for auto-formatting the code -- sorry it's not mentioned anywhere!

nickolay:
Excellent, thank you.

Issue #48, opened 2019-04-19
Add support for prepared queries.


Jekshmek:
Add support for prepared queries.
```sql
let sql = "SELECT * FROM users WHERE name = $1";
```
```bash
AST: Err(
    ParserError(
        "Did not expect Char(\'$\') at the beginning of an expression"
    )
)

```

nickolay:
I'm guessing you want this for a library that will let the user execute SQL in two steps (prepare + run)?

A quick search indicates that the dollar sign:

* starts a `money` literal in [MSSQL](https://stackoverflow.com/a/30407243/1026)
* starts an identifier in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html)
* can start a double-dollar quoted string in [H2](http://www.h2database.com/html/grammar.html#dollar_quoted_string)

Another common character to indicate parameters in prepared queries is colon (`WHERE name = :1`).

Perhaps this could be implemented by adding functions that recognize the query parameters (similar to `is_identifier_start`/`is_identifier_part`) to `Dialect`, and additional variants to `Token` and `Value`? @andygrove what do you think?

Jekshmek:
Yes, I use sql in two steps (prepare + run)
Similar to the library (implemented as a compiler plugin) [rust-postgres-macros](https://github.com/sfackler/rust-postgres-macros)
it is necessary to parse the sql at the compilation stage in order to exclude a non-valid sql at runtime.
For database queries I use [rust-postgres](https://github.com/sfackler/rust-postgres) which uses `$1` as a parameter of a prepared request.

It may be possible to add the same: `:1`.
Perhaps you should separate the implementation from `parse_sql` and add the function` parse_sql_prepare`?

nickolay:
Thanks for the pull request (#55). Reading through it gave me another idea: you could implement a dialect (in your own crate) that accepts `'$'` in its `is_identifier_start`: this would lead to `$1` getting parsed as `ASTNode::SQLIdentifier`. There's some difference in parser's treatment of identifiers vs literals, but I can't think of a case where it would matter. Would this work for you?

Jekshmek:
@andygrove  Thanks, this solution is right for me.

```rust
use sqlparser::dialect::Dialect;
use sqlparser::sqlast::SQLStatement;
use sqlparser::sqlparser::ParserError;
use sqlparser::sqlparser::Parser;

pub struct MyPostgreSqlDialect {}

[...]

PR #55, opened 2019-04-26
Add prepare literal

Jekshmek:
Added ability to use prepared request.


nickolay:
Closed per discussion in #48

PR #61, opened 2019-05-03
Reorganize tests, remove datetime code, and add basic MSSQL dialect

nickolay:
Currently:
* Most of the tests for sqlparser are written as "integration tests" located in  `tests/sqlparser_<dialect>.rs`.
* Most tests are in `sqlparser_generic.rs` (which actually tested the generic and the Postgres dialects), but some tests - not all of them postgres-specific - are in `sqlparser_postgres.rs`. (There's also an `sqlparser_ansi.rs` with a single test because we don't test the ANSI dialect elsewhere.)
* The helper functions like `one_statement_parses_to()` are partially duplicated among the "generic" and the "postgres" files, so moving tests to a different file in `tests/` or to a unit test inside the sqlparser crate, as well as adding a new dialect, requires copying the helpers or doing some refactoring.

This PR adds (478dbe940d071b84b9a373456c197dd6b4d1dfa8) an `sqlparser::test_utils` module with the helpers, that were previously implemented in `tests/sqlparser_generic.rs`. I had to make the module [public and enabled in non-test configurations](https://stackoverflow.com/a/44541071/1026), so that it could be used from both the integration tests and the unit tests within the sqlparser crate (like `test_prev_index`, which confusingly part of `sqlparser_postgres.rs`)

This PR also lets us pick which dialects to run a given test with (all dialects, a specific dialect, or a specific dialect and the generic one) by moving the helpers to methods on a new `TestedDialects` struct, which stores the dialects to test with.

I've also renamed `sqlparser_generic.rs` to `sqlparser_common.rs` and made the tests there run for all dialects (removing the need for `sqlparser_ansi.rs`). Rearranged the tests between the files, so that `sqlparser_postgres.rs` only contained tests with PG-specific syntax (some of those tests are parsed with the generic parser too, though).

While doing this reorganization I saw the effectively disabled date-time tests, and decided to remove them along with the unused code for date/time support (de177f107cd0f879aba60a5893778a71e6602452), as I believe it will not be used in its current form even when we implement support for dates.

I've also noticed one test in `sqlparser_generic.rs` (for `SELECT @@version`), which was not compatible with the ANSI dialect, so I added the MSSQL dialect I've been meaning to add for a while, and updated the rules for parsing identifiers for Postgres dialect while I was at it (304710d59ad7b89378148eafec02c2045daaf826).


nickolay:
Thanks!

PR #72, opened 2019-05-22
Implement std::error::Error for ParserError

benesch:


nickolay:
Hey @benesch, thanks for this and the other PRs! 

As I understand, [the point is](https://stevedonovan.github.io/rust-gentle-intro/6-error-handling.html) to let the caller use `?` to pass an `sqlparser::ParserError`, as well as any other kind of error, up the call stack as a `Result<..., Box<Error>>`.

The change looks good to me; a minor point (that also applies to 74 and 75) is that sqlparser is auto-formatted [using rustfmt](https://github.com/rust-lang/rustfmt#on-the-stable-toolchain) (even though sometimes the resulting formatting is sub-optimal). I think Github will let me fix it up when merging though.

benesch:
> As I understand, the point is to let the caller use `?` to pass an `sqlparser::ParserError`, as well as any other kind of error, up the call stack as a `Result<..., Box<Error>>`.

Yes, that's exactly right. I've run this through `cargo fmt`, so should be good to merge once 

> Hey @benesch, thanks for this and the other PRs!

You are very welcome! Thanks for your maintainership of this library. :) 

I meant to drop you a line in one of these PRs and it looked like it slipped my mind: the context for all this work is that I'm upstreaming patches from our internal fork at @MaterializeInc. I'm erring on the side of upstreaming everything possible, and I don't think any of it will be objectionable, but definitely let me know if you think any of it would be better left to us downstream!

PR #74, opened 2019-05-22
Support arbitrary WITH options for CREATE [TABLE|VIEW]

benesch:
Both Postgres and MSSQL accept this syntax, though the particular
options they accept differ.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -379,6 +379,7 @@ pub enum SQLStatement {
         name: SQLObjectName,
         query: Box<SQLQuery>,
         materialized: bool,
+        with_options: Vec<SQLOption>,
     },
     /// CREATE TABLE
     SQLCreateTable {
@@ -387,6 +388,7 @@ pub enum SQLStatement {
         /// Optional schema
         columns: Vec<SQLColumnDef>,
         constraints: Vec<TableConstraint>,
+        with_options: Vec<SQLOption>,
         external: bool,
         file_format: Option<FileFormat>,
         location: Option<String>,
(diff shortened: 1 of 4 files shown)


nickolay:
I'd like to share a few thoughts on this, even though I don't have specific suggestions at this time:
* Both Postgres and MS support options without values (e.g. Postgres `CREATE TABLE .. WITH (OIDS)`). (MS additionally does not wrap the CREATE VIEW options in parentheses.)
* There's a standard `WITH .. CHECK OPTION` clause, distinct from the "WITH options" this PR implements, which makes for two sets of "with options"...
* If we want to support parsing statements that are not `;`-delimited, care should be taken to distinguish `WITH (options)` at the end of a CREATE statement and a `WITH cte AS` at the beginning of the next query.

Some links to documentation (apparently these options can appear in other contexts, this is just a sample):
* Postgres:
  * [`CREATE TABLE` (storage_parameter)](https://www.postgresql.org/docs/11/sql-createtable.html#SQL-CREATETABLE-STORAGE-PARAMETERS),
  * [`CREATE VIEW .. [ WITH ( view_option, ... ) ] AS ..  [ WITH [ CASCADED | LOCAL ] CHECK OPTION ]`](https://www.postgresql.org/docs/11/sql-createview.html)
* MS:
  * [`CREATE TABLE .. WITH (<table_option>, ...)`](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017)
  * [`CREATE VIEW .. [ WITH <view_attribute>, ... ] AS ... [ WITH CHECK OPTION ]`](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-view-transact-sql?view=sql-server-2017)

nickolay:
One thing I'd like to request is that you put a realistic SQL in the testcase (possibly moving it to a dialect-specific suite); the extensions can be implemented in a follow-up.

My idea with `sqlparser_common.rs` was that it runs (and must succeed) for all dialects. While currently only the lexer does dialect-specific processing, I imagine we'll come to a point where different dialects must parse a given SQL differently. I hoped that being clear about which dialect is supposed to handle certain non-common testcases would make that easier.

benesch:
Ok, added a more realistic test case to the Postgres dialect. Thanks for the review!

nickolay:
Thanks! Given that you kept `WITH (foo = 'bar', a = 123)` in the "general" tests I'd like to clarify - do you rely on a variation of this in your product?

benesch:
> do you rely on a variation of this in your product?

We do, yeah—but I think we'd be happy to declare that we are most closely aligned with the Postgres dialect, and ask for that dialect when we parse. I.e., you should definitely feel free to delete the general test case, or move it to the Postgres-specific test case.

PR #75, opened 2019-05-22
Support DROP [TABLE|VIEW]

benesch:


src/sqlast/mod.rs
```diff
@@ -608,3 +614,28 @@ impl FromStr for FileFormat {
...
+    pub restrict: bool,
+}
+
+impl SQLDrop {
+    fn to_string_internal(&self, object_type: &str) -> String {
+        format!(
+            "DROP {}{} {}{}{}",
+            object_type,
+            if self.if_exists { " IF EXISTS" } else { "" },
+            self.names
```
nickolay: `comma_separated_string(&self.names),` seems to work here.

src/sqlast/mod.rs
```diff
@@ -608,3 +614,28 @@ impl FromStr for FileFormat {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SQLDrop {
+    pub if_exists: bool,
+    pub names: Vec<SQLObjectName>,
+    pub cascade: bool,
+    pub restrict: bool,
```
nickolay: Like with `SELECT ALL` (#76) I wonder if it makes sense to store the `restrict` in the AST, given it's the default.

benesch: Agreed, fixed.

src/sqlast/mod.rs
benesch: Good point. Done.

benesch:
> Oh, interesting. We didn't notice that before writing this implementation. Definitely not opposed to having SQLDrop store the object type (table/view), rather than vice-versa, though I'd prefer to have a proper enum rather than a string type, so that it's possible to exhaustively handle all the cases. Thoughts?

I decided I like this a lot, and adjusted the PR accordingly.

nickolay:
Great, thanks! I've asked @andygrove if he wants to take a look at this and 72, 73, 76 before merging. I hope to get to the other PRs "soon":)

PR #76, opened 2019-05-22
Support SELECT ALL

benesch:


nickolay:
This is about `<set quantifier> ::= DISTINCT | ALL` in [`<query specification> ::= SELECT [ <set quantifier> ] <select list> ...`](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_7_12_query_specification), where `ALL` is the same as lack of an explicit set quantifier.

Given that parsing text->AST currently loses information unimportant to the query semantics, perhaps we shouldn't add the `all` flag to the AST? Did you need it for some reason?

benesch:
> Given that parsing text->AST currently loses information unimportant to the query semantics, perhaps we shouldn't add the all flag to the AST? Did you need it for some reason?

No particular reason! I think we just hadn't fully adopted the "it's ok if queries don't roundtrip exactly, as long as they roundtrip to something equivalent" rule. I've updated the PR to drop the explicit `all` flag and simplified the test accordingly.

Thanks for the review!

PR #77, opened 2019-05-22
Support COUNT(DISTINCT x) and similar

benesch:


nickolay:
[The standard says](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_10_9_aggregate_function) DISTINCT/ALL is allowed in a few ("set") functions, accepting a single argument (`<general set function> ::= <set function type> <left paren> [ <set quantifier> ] <value expression> <right paren>`), and they may or may not be a [window function](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#_6_10_window_function) (i.e. followed by `OVER (...)`).

This PR extends `ASTNode::SQLFunction` to hold `{ name, all, distinct, args, over }`, which I think is OK for now, but going forward I think that we might want to switch to using a enum to denote different style of the function arguments (`COUNT(*)` vs `EXTRACT(YEAR FROM val)` vs `COUNT(DISTINCT val)`).

 `all` could be removed for the same reasons as in #76, and I think that putting `distinct` before `args` would make the `SQLFunction` declaration slightly easier to read. LGTM either way, thanks!

benesch:
> This PR extends ASTNode::SQLFunction to hold { name, all, distinct, args, over }, which I think is OK for now, but going forward I think that we might want to switch to using a enum to denote different style of the function arguments (COUNT(*) vs EXTRACT(YEAR FROM val) vs COUNT(DISTINCT val)).

Yeah, agreed that the type system could do more for us here. SQL really blurs the line between function calls and dedicated syntax. (Consider `CAST(datum AS type)`, for example, which is handled with its own ASTNode enum variant already.)

> all could be removed for the same reasons as in #76, and I think that putting distinct before args would make the SQLFunction declaration slightly easier to read. LGTM either way, thanks!

Good point, done.

PR #78, opened 2019-05-22
Add an AST visitor

benesch:
AST visitors can make traversal of an AST much simpler when only part of the AST is interesting. E.g., it's quite easy to write a visitor that answers a question like "are there any aggregate functions in this query?". 

This commit is pretty dependent on the exact structure of the AST, so I think it depends on basically all the other PRs I just submitted (#72, #73, #74, #75, #76, and #77). The relevant diff is in the last commit: https://github.com/andygrove/sqlparser-rs/commit/c7c89bd2c9b37892a329d71ba10b518e4e400225

### DIFF
--- a/src/sqlast/visit.rs
+++ b/src/sqlast/visit.rs
@@ -1,0 +1,1284 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! SQL AST traversal.
+
+// Visiting an enum struct variant with many arguments forces our hand here.
+// If, over time, we convert these large variants to dedicated structs, we
+// can remove this escape hatch.
+#![allow(clippy::too_many_arguments)]
+
+// Disable lints that want us to rewrite `&SQLIdent` as `&str`, as `&str` is not
+// as self-documenting as &SQLIdent.
+#![allow(clippy::ptr_arg)]
+
+use super::*;
+
+/// A trait that represents a visitor that walks through a SQL AST.
+///
+/// Each function corresponds to a node in the SQL AST, and has a default
+/// implementation that visits all of its child nodes. Implementors of this
+/// trait can override functions as desired to hook into AST traversal without
+/// writing code to traverse the entire AST.
+pub trait Visit<'ast> {
+    fn visit_statement(&mut self, statement: &'ast SQLStatement) {
+        visit_statement(self, statement)
+    }
+
+    fn visit_query(&mut self, query: &'ast SQLQuery) {
+        visit_query(self, query)
+    }
+
+    fn visit_cte(&mut self, cte: &'ast Cte) {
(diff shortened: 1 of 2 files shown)


nickolay:
I don't have a firm opinion about this:

1. on one hand this increases the burden on the authors of the future PRs to sqlparser
2. on the other hand, it probably makes it easier to walk the AST for some consumers

For my own use-case I convert the AST to JSON via serde and analyze it outside of rust, so I can't feel the benefits of (2) on my own, and I didn't find time to get familiar with the work @cswinter and @andygrove do based on sqlparser to see if this would be beneficial to them.

To evaluate how much of a burden is (1) I'd have to implement a new feature on top of this. I expect the type system will make it trivial, and the hardest part would be to update the testcase, so I'm leaning towards giving it a try.

benesch:
Yep, those are exactly the tradeoffs! It is kind of a pain to update the visitor as well as the AST, so I'd be happy to wait on this until the API is more stable. (You may have noticed I was particularly slow to rebase this PR.) I think at some point the scales will tip firmly in favor of bundling the visitor with the library, but I'm not totally convinced we've hit that point yet.

src/sqlast/visit.rs
```diff
@@ -0,0 +1,1284 @@
...
+
+/// A trait that represents a visitor that walks through a SQL AST.
+///
+/// Each function corresponds to a node in the SQL AST, and has a default
+/// implementation that visits all of its child nodes. Implementors of this
+/// trait can override functions as desired to hook into AST traversal without
+/// writing code to traverse the entire AST.
+pub trait Visit<'ast> {
+    fn visit_statement(&mut self, statement: &'ast SQLStatement) {
+        visit_statement(self, statement)
```
andygrove: I'm confused by the default implementation ... it looks like it recurses straight back into itself. I'm sure I'm missing something obvious here but could you explain this?

benesch: Sure thing! I stole this pattern from the syn crate's visitor. The goal is to minimize the amount of code that the consumer needs to write.

Basically, there's a trait method with a default implementation that calls a free function (with the same name, hence the confusion), and that free function that does the actual walking of the AST. Here's the first few lines of `visit_statement`, for example:

```rust
pub fn visit_statement<'ast, V: Visit<'ast> + ?Sized>(
     visitor: &mut V,
     statement: &'ast SQLStatement,
 ) {
     match statement {
         SQLStatement::SQLQuery(query) => visitor.visit_query(query),
         SQLStatement::SQLInsert {
             table_name,
             columns,
             source,
         } => visitor.visit_insert(table_name, columns, source),
         // ...
     }
 }
```

Here's a more concrete example of a visitor that makes use of the free function avoid reimplementing the AST-walking code:

```rust
struct MyVisitor {
    // arbitrary state
}

impl<'ast> Visit<'ast> for MyVisitor<'ast> {
    fn visit_function(&mut self, func: &'ast SQLFunction) {
        // pre-visit
        sqlparser::visit::visit_function(func);
        // post-visit
    }
}
```

Basically, overriding trait methods is really convenient, because you only have to override the methods whose behavior you want to change. But sometimes you don't want to override the method entirely; you just want to tack on some behavior to the beginning or end. Having a public free function makes this quite ergonomic. You override the trait method, do whatever you need to, and then call the code to keep recursion when you're ready—or perhaps not at all, if you've already found what you need.

andygrove: Got it. Thanks for explaining that to me.

andygrove:
Sorry @benesch this somehow broke the master build and I had to revert

benesch:
No problem, @andygrove—I fixed the merge skew and resubmitted as #114.

PR #79, opened 2019-05-22
Standardize license headers

### LINKED ISSUE #58: Remove copyright from headers
The file headers are very inconsistent right now. We should probably have CI fail if files don't have ASL headers so that this code is kept in compliance with ASL and could potentially be contributed to an Apache project in the future. We should certainly remove `Copyright 2018 Grove Enterprises LLC` from all headers.

benesch:
Standardize the license header, removing the Grove Enterprise copyright
notice where it exists per #58. Also add a CI check to ensure that files
without license headers don't get merged.

Fix #58.

@andygrove it'd be good to get your eyes on this, because I'm removing your copyright notice (based on what you said in #58) and I'm not sure if I should add something else in its place.


andygrove:
Thanks @benesch I approve of removing the Grove Enterprises mentions

PR #80, opened 2019-05-24
Support nested expressions in BETWEEN

benesch:
`BETWEEN <thing> AND <thing>` allows <thing> to be any expr that doesn't
contain boolean operators. (Allowing boolean operators would wreak
havoc, because of the repurposing of AND as both a boolean operation
and part of the syntax of BETWEEN.)

### DIFF
--- a/src/sqlparser.rs
+++ b/src/sqlparser.rs
@@ -478,9 +478,12 @@ impl Parser {
 
     /// Parses `BETWEEN <low> AND <high>`, assuming the `BETWEEN` keyword was already consumed
     pub fn parse_between(&mut self, expr: ASTNode, negated: bool) -> Result<ASTNode, ParserError> {
-        let low = self.parse_prefix()?;
+        // Stop parsing subexpressions for <low> and <high> on tokens with
+        // precedence lower than that of `BETWEEN`, such as `AND`, `IS`, etc.
+        let prec = self.get_precedence(&Token::make_keyword("BETWEEN"))?;
+        let low = self.parse_subexpr(prec)?;
         self.expect_keyword("AND")?;
-        let high = self.parse_prefix()?;
+        let high = self.parse_subexpr(prec)?;
         Ok(ASTNode::SQLBetween {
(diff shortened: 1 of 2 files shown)


nickolay:
Huh, I completely missed the fact that BETWEEN supports expressions when I wrote this (https://github.com/andygrove/sqlparser-rs/pull/43/commits/786b1cf18abad4bfbee5966bc31e0225eeb0fca0). Thanks for figuring this out!

I think this works by re-using `BETWEEN`'s precedence in two distinct scenarios:

1. as the left-binding precedence (this worked before this change), which:
    * being lower than the precedence of binary arithmetic operators makes `1+1 BETWEEN ..` parse as `(1+1) BETWEEN ..`
    * being higher than the precedence of AND/OR makes `expr AND bar BETWEEN ..` parse as `expr AND (bar BETWEEN ..)`
2. as the right-binding precedence, which stops parsing the expressions for the range bounds whenever a token with a lower precedence (such as `AND`, but also `IS` and others) is encountered.

I feel an urge to add a comment to that effect, maybe:

    // Stop parsing subexpressions for <low> and <high> on tokens with
    // precedence lower than that of `BETWEEN` (such as `AND`, `IS`, etc.)
    let prec = self.get_precedence(&Token::make_keyword("BETWEEN"))?; 
    let low = self.parse_subexpr(prec)?;

I'd also extend the test to check that `where 1=1 and 1+x between ...` still parses correctly as we don't appear to have an existing test for this.

benesch:
> Huh, I completely missed the fact that BETWEEN supports expressions when I wrote this (786b1cf). Thanks for figuring this out!

You bet! I'm working towards getting the parser to parse all the queries in https://www.sqlite.org/sqllogictest. There's only about 3k queries left that cause parse failures.

Your explanation of how this works is spot on. I've added the comment as you suggested.

> I'd also extend the test to check that where 1=1 and 1+x between ... still parses correctly as we don't appear to have an existing test for this.

Good idea. Done.

PR #82, opened 2019-05-27
Fix the precedence of NOT LIKE

### LINKED ISSUE #81: Precedence of NOT isn't quite right
It took me a while to come up with this example, and it's not totally realistic, but it's probably still worth fixing. Basically, the precedence of `NOT LIKE` is mishandled, because the precedence of `NOT` is fixed to its unary operator precedence, when in it needs to vary based on whether the NOT keyword is followed by LIKE/BETWEEN/IN/<something else>.

Concretely, `SELECT a LIKE b IS NULL` parses correctly, with the LIKE binding more tightly than the IS NULL:

```rust
SQLIsNull(
    SQLBinaryExpr {
        left: SQLIdentifier(
            "a"
        ),
        op: Like,
        right: SQLIdentifier(
            "b"
        )
    }
)
```

But when NOT LIKE is substituted for LIKE, the IS NULL incorrectly binds more tightly than the NOT LIKE:

```rust
SQLBinaryExpr {
    left: SQLIdentifier(
        "a"
    ),
    op: NotLike,
    right: SQLIsNull(
        SQLIdentifier(
            "b"
        )
    )
}
```

benesch:
~NOT LIKE has the same precedence as the LIKE operator. The parser was
previously assigning it the precedence of the unary NOT operator.~

~The fix is a bit involved, as the precedence of NOT depends on the keyword that follows it.~

~FYI this will conflict with #80, but I'll rebase whichever gets merged second.~

NOT LIKE has the same precedence as the LIKE operator. The parser was
previously assigning it the precedence of the unary NOT operator. NOT
BETWEEN and NOT IN are treated similarly, as they are equivalent, from a
precedence perspective, to NOT LIKE.

The fix for this requires associating precedences with sequences of
tokens, rather than single tokens, so that "NOT LIKE" and "NOT <expr>"
can have different preferences. Perhaps surprisingly, this change is not
very invasive.

An alternative I considered involved adjusting the tokenizer to lex
NOT, NOT LIKE, NOT BETWEEN, and NOT IN as separate tokens. This broke
symmetry in strange ways, though, as NotLike, NotBetween, and NotIn
gained dedicated tokens, while LIKE, BETWEEN, and IN remained as
stringly identifiers.

Fixes #81.

### DIFF
--- a/src/sqlparser.rs
+++ b/src/sqlparser.rs
@@ -190,13 +190,10 @@ impl Parser {
                 }
                 "CASE" => self.parse_case_expression(),
                 "CAST" => self.parse_cast_expression(),
-                "NOT" => {
-                    let p = self.get_precedence(&Token::make_keyword("NOT"))?;
-                    Ok(ASTNode::SQLUnary {
-                        operator: SQLOperator::Not,
-                        expr: Box::new(self.parse_subexpr(p)?),
-                    })
-                }
+                "NOT" => Ok(ASTNode::SQLUnary {
+                    operator: SQLOperator::Not,
+                    expr: Box::new(self.parse_subexpr(Self::UNARY_NOT_PREC)?),
+                }),
                 // Here `w` is a word, check if it's a part of a multi-part
                 // identifier, a function call, or a simple identifier:
                 _ => match self.peek_token() {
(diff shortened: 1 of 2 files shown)


benesch:
This is rebased and ready to go! I split the original patch into two commits to hopefully make the details of the change more clear.

nickolay:
LGTM, thanks!
(I didn't double-check the testcases, as it looks like you've given it much thought already.)

PR #85, opened 2019-05-31
Enable Clippy and rustfmt in CI

benesch:
@nickolay what do you think about this? I've found that if Clippy isn't enabled in CI, lint violations creep in quite quickly.


benesch:
I'm inclined to ignore Coveralls here. The decrease is a quirk of how assert!(false) is parsed vs. unreachable!().

nickolay:
Looks great, thanks for doing this!

(Regarding coveralls, I find it very noisy and am not really paying attention to it..)

benesch:
> (Regarding coveralls, I find it very noisy and am not really paying attention to it..)

Makes sense. I find it useful to have around when you're explicitly looking for stuff that's undertested, but it's definitely pretty spammy on PRs.

Issue #83, opened 2019-05-27
Support join associativity


benesch:
Alternate join associativity can be forced with parentheses, as in:

```sql
SELECT * FROM a NATURAL JOIN (b NATURAL JOIN c);
```

The current parser doesn't support it, nor is the design of the SQLSelect struct particularly conducive to it.

PR #87, opened 2019-06-02
Refactor parse_joins

nickolay:
While preparing to add support for the non-standard `CROSS`/`OUTER APPLY` ([an alternative syntax for LATERAL invented by MS and adopted by Oracle](https://blog.jooq.org/2013/12/18/add-lateral-joins-or-cross-apply-to-your-sql-tool-chain/)) I've tweaked the internals of `parse_joins` to reduce code duplication, added tests for `NATURAL` and fixed a couple minor issues (extra whitespace in the serializer and NATURAL being silently eaten when not followed by a JOIN).

(This does _not_ change the AST, and doesn't deal with #83.)

### DIFF
--- a/src/sqlast/query.rs
+++ b/src/sqlast/query.rs
@@ -271,14 +271,14 @@ impl ToString for Join {
         }
         fn suffix(constraint: &JoinConstraint) -> String {
             match constraint {
-                JoinConstraint::On(expr) => format!("ON {}", expr.to_string()),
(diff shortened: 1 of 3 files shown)


benesch:
Heh, I've got a patch somewhere in my stack that addresses that same whitespace bug. Everything LGTM!

benesch:
I'm going to merge this now, actually, because I've got a patch for #83 coming up that's going to need a rebase.

PR #88, opened 2019-06-03
Implement Hash on all AST nodes

benesch:
It is convenient for downstream libraries to be able to stash bits of
ASTs into hash maps, e.g., for performing simple common subexpression
elimination.

The only downside to this change is that it requires that the f64 in the
Value enum be wrapped in an OrderedFloat, which provides the necessary
equality semantics to allow Hash to be drived. The reason f64 doesn't
implement Hash by default is because NaN is typically not equal to
itself, so it's not clear what it should hash to. That's less of a
concern in a SQL context, because every SQL database I've looked at
treats NaN as equal to itself, in violation of the IEEE standard, in
order to permit indexing and sorting of float columns.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ path = "src/lib.rs"
 
 [dependencies]
 log = "0.4.5"
+ordered-float = "1.0.2"
 
 [dev-dependencies]
 simple_logger = "1.0.1"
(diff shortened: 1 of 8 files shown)


benesch:
> ... although I don't think Value::Double(f64) was correct anyway - we parse it as a decimal, and I'm not sure the SQL standard mandates converting to 64-bit floating point, losing precision for decimals longer than 15 digits...

Oof, good point.

PR #89, opened 2019-06-03
Extract a SQLFunction struct

benesch:
This variant is getting rather large, and it's useful downstream to be
able to pass around SQLFunction nodes without reinventing the wheel.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -108,13 +108,7 @@ pub enum ASTNode {
     /// SQLValue
     SQLValue(Value),
     /// Scalar function call e.g. `LEFT(foo, 5)`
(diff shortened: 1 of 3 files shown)


src/sqlast/mod.rs
```diff
@@ -346,6 +324,7 @@ impl ToString for SQLWindowFrameBound {
 }
 
 /// A top-level statement (SELECT, INSERT, CREATE, etc.)
+#[allow(clippy::large_enum_variant)]
```
nickolay: I see that clippy points at `SQLStatement::SQLUpdate` as being a "large variant" after this change. I don't understand why exactly, but I think going forward we might extract individual statement types making this a moot point.

PR #90, opened 2019-06-03
Support EXISTS subqueries

benesch:


benesch:
> I'm assuming NOT EXISTS() will parse correctly with unary NOT, unsure if it deserves a test of its own.

Yep! I added a test just to be sure.

> This fixes the last remaining part of #5, thanks!

Oh, nice. I've updated the commit message to close out #5 when this merges.

PR #91, opened 2019-06-03
Improve VALUES-related syntax parsing

benesch:
The first commit allows VALUES clauses in more places; the second commit teaches INSERT to accept things besides VALUES clauses. See individual commit messages for details.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -20,10 +20,12 @@ mod sql_operator;
 mod sqltype;
 mod value;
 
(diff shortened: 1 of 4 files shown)


benesch:
> Any reason you chose to serialize the VALUES prefix as part of SQLSetExpr::to_string() and not as part of SQLValues::to_string()?

No particular reason! Just an artifact of how this changeset organically evolved. I've adjusted SQLValues to serialize with the `VALUES` prefix.

PR #71, opened 2019-05-21
Don't panic on weird infix garbage

benesch:
Submitting on behalf of @jamii.

PR #93, opened 2019-06-03
Parse column constraints in any order

benesch:
`CREATE TABLE t (a INT NOT NULL DEFAULT 1 PRIMARY KEY)` is as valid as `CREATE TABLE t (a INT DEFAULT 1 PRIMARY KEY NOT NULL)`.

@nickolay I got annoyed by the tests in sqlparser_postgres and adjusted them so that the first test asserts every column specification, and the remainder simply verify that the SQL roundtrips. Let me know if you'd prefer to be more precise with the later tests.

#

### DIFF
--- a/src/sqlast/ddl.rs
+++ b/src/sqlast/ddl.rs
@@ -1,6 +1,6 @@
 //! AST types specific to CREATE/ALTER variants of `SQLStatement`
 //! (commonly referred to as Data Definition Language, or DDL)
-use super::{ASTNode, SQLIdent, SQLObjectName};
+use super::{ASTNode, SQLIdent, SQLObjectName, SQLType};
 
 /// An `ALTER TABLE` (`SQLStatement::SQLAlterTable`) operation
 #[derive(Debug, Clone, PartialEq, Hash)]
@@ -48,11 +48,6 @@ pub enum TableConstraint {
 
 impl ToString for TableConstraint {
     fn to_string(&self) -> String {
-        fn format_constraint_name(name: &Option<SQLIdent>) -> String {
-            name.as_ref()
-                .map(|name| format!("CONSTRAINT {} ", name))
-                .unwrap_or_default()
-        }
         match self {
             TableConstraint::Unique {
                 name,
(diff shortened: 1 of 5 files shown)


nickolay:
This got rather long, sorry!

1) The `CHECK` constraint is only partially implemented: it's in the AST, but the parser doesn't handle it. Please make a note if you want to leave it unimplemented for now.

2) A thought on parsing of the `DEFAULT` expression:
> We want to allow as many DEFAULT expressions as possible, but calling self.parse_expr() will choke on expressions like "DEFAULT 1 NOT NULL". Typically this would be a syntax error, as "1 NOT NULL" is not a valid SQL expression, but in this case we know that NOT NULL is part of the next column constraint.

A syntax error happens whenever a "token" (or - with the change from #82 - a token pair, like "NOT IN") following the `DEFAULT` expression has a [get_next_precedence](https://github.com/andygrove/sqlparser-rs/blob/1f8708390675254794da11b38618daa01316e3f6/src/sqlparser.rs#L576) > 0. The [Pratt parser then attempts to](https://github.com/andygrove/sqlparser-rs/blob/1f8708390675254794da11b38618daa01316e3f6/src/sqlparser.rs#L147) parse it as an "infix", and in the case of `NOT NULL` fails with "Expected IN or BETWEEN after NOT, found: null".

> As it turns out, the only expressions that cause trouble all have precedence less than or equal to BETWEEN, so we pass BETWEEN_PREC to stop parsing on tokens with precedence less than or equal to BETWEEN.
> The same trick is [employed by PostgreSQL](https://github.com/postgres/postgres/blob/56b78626c/src/backend/parser/gram.y#L13366-L13370)

I've looked through the docs for a few dialects (below) and I can't think of anything other than `NOT` (and `COLLATE` -  see below) that's under a risk of being incorrectly treated as a DEFAULT expression's infix, while being a valid keyword starting a new constraint or a column attribute definition.

After #82 `get_next_precedence` is only used to get the [left-binding precedence](https://github.com/andygrove/sqlparser-rs/pull/80#issuecomment-496328019) of the peeked token (unlike before when `get_precedence(make_keyword("NOT"))` was used to get the right-binding precedence to pass to `parse_subexpr` - we now use constants for that).

I don't see why `get_next_precedence` needs to treat `NOT` not followed by `IN`/`BETWEEN` any differently than other keywords that are not "infix" operators. It currently returns lbp=`UNARY_NOT_PREC` for `NOT` in infix position instead of `0`, so you have to call `parse_subexpr` with rbp=`BETWEEN_PREC` to make it stop (per rbp>=lbp) when encountering `NOT`. If we defined lbp(`NOT`) = 0, the parser invoked with the default rbp=0 would stop on `NOT` followed by `NULL` just like it stops on any unknown keyword. (And the `unreachable!` removed in #71 could be restored then.)

What do you think?

3) On `COLLATE`:
* ANSI puts `COLLATE` as the last clause in a column definition, but it doesn't allow arbitrary expressions in DEFAULT, so it doesn't conflict.
* PG/MS both seem to expect `COLLATE` before `DEFAULT` in column definition, presumably to avoid the ambiguity of `DEFAULT 'foo' COLLATE bar`.

As we parse COLLATE as an optional suffix at the end of `parse_prefix`, the precedence trick doesn't help with this anyway, and to support ANSI syntax we'd need to use a different radically simpler parser for the `DEFAULT` clause. This is not in the scope of this PR, of course.

4) On the AST design:

ANSI allows `CONSTRAINT <name>` to be used with `NOT NULL`, `UNIQUE`/`PRIMARY KEY`, and `REFERENCES`. MSSQL doesn't seem to support it for `NOT NULL`, but supports it for `DEFAULT` instead. Postgres allows both `NOT NULL` and `DEFAULT`, plus `NULL` and `GENERATED` as constraints.

All three dialects agree that `COLLATE` isn't a constraint, but expect it in different positions. MS has a number of custom options interleaved with the standard ones.

Maybe this could be modeled as a `Vec<ColumnOption>` where a ColumnOption is either a `Constraint { name: Option<String>, constraint: Box<ColumnOption>, deferred, deferrable, etc }`, one of standard options (including constraints, COLLATE, DEFAULT, and GENERATED), or a dialect-specific String?

5) I didn't get to the tests, but given that you're being very thorough with testing, I'll probably agree with whatever approach you decided on.

6) Relevant parts of the docs:

* [ANSI](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#column-definition)
  ```
  <column definition> ::= <column name> [ <data type or domain name> ]
      [ <default clause> | <identity column specification> | <generation clause>
        | <system time period start column specification>
        | <system time period end column specification>
      ]
      [ <column constraint definition>... ]
      [ <collate clause> ]

  <column constraint definition> ::= [CONSTRAINT ..]
    { NOT NULL
      | <unique specification>
      | <references specification>
      | <check constraint definition>
    }
    { in any order:
      [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
      [ [ NOT ] DEFERRABLE ]
    }
    [ [ NOT ] ENFORCED ]
  ```
* [Postgres](https://www.postgresql.org/docs/11/sql-createtable.html)
  ```
  column_name data_type [ COLLATE collation ] [ column_constraint [ ... ] ]
  and column_constraint is:
  [ CONSTRAINT constraint_name ]
  { NOT NULL | NULL | CHECK | DEFAULT .. | GENERATED .. |
  UNIQUE .. | PRIMARY KEY .. | REFERENCES .. }
  [ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
  ```
* [MSSQL](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017)
  ```
  <column_definition> ::= column_name <data_type>
    [ FILESTREAM ]
    [ COLLATE collation_name ]
    [ custom stuff ]
    [ CONSTRAINT constraint_name [ DEFAULT constant_expression ] ]
    [ GENERATED .. ]
    [ NULL | NOT NULL ]
    [ more custom stuff ]
    [ <column_constraint> [, ...n ] ]
    [ <column_index> ]

  <column_constraint> ::= [ CONSTRAINT constraint_name ]
    { CHECK .. 
    | FOREIGN KEY .. 
    | { PRIMARY KEY | UNIQUE } [ custom ] }
  ```

benesch:
> After #82 get_next_precedence is only used to get the left-binding precedence of the peeked token (unlike before when get_precedence(make_keyword("NOT")) was used to get the right-binding precedence to pass to parse_subexpr - we now use constants for that).

> I don't see why get_next_precedence needs to treat NOT not followed by IN/BETWEEN any differently than other keywords that are not "infix" operators. It currently returns lbp=UNARY_NOT_PREC for NOT in infix position instead of 0, so you have to call parse_subexpr with rbp=BETWEEN_PREC to make it stop (per rbp>=lbp) when encountering NOT. If we defined lbp(NOT) = 0, the parser invoked with the default rbp=0 would stop on NOT followed by NULL just like it stops on any unknown keyword. (And the unreachable! removed in #71 could be restored then.)

This is brilliant! You're entirely right that there's a bug here. We should absolutely be returning 0 if we see a unary NOT in get_next_precedence, because unary NOT is not an infix operator and therefore has no left binding power. 

I'm going to split that change out into its own PR.

benesch:
Ok, I've created a separate PR for the unary NOT issue (#107), and rebased this PR on top of it.

> 1. The CHECK constraint is only partially implemented: it's in the AST, but the parser doesn't handle it. Please make a note if you want to leave it unimplemented for now.

Oops, thanks! This is fixed now, and I've added a test.

> 2. A thought on parsing of the DEFAULT expression...

This was spot on, and is addressed in #107.

> 3. On COLLATE...
> 4. On the AST design...

I took a shortcut here and added `COLLATE` as an `Option<SQLObjectName>` on the `SQLColumnDef` struct. I'll try to find time soon to rework the structs as you've suggested.

nickolay:
> I'll try to find time soon to rework the structs as you've suggested.

Did you mean you wanted to merge this PR before reworking? I’ll try look at this over the weekend, but please let me know if it blocks you.

benesch:
> Did you mean you wanted to merge this PR before reworking? I’ll try look at this over the weekend, but please let me know if it blocks you.

Not blocked! Just wanted to write down which chunks of your feedback I've addressed and which chunks I haven't, so that I remember when I get back to it.

src/sqlast/ddl.rs
```diff
@@ -84,3 +79,106 @@ impl ToString for TableConstraint {
...
+                .join("")
+        )
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Hash)]
+pub enum ColumnConstraint {
+    /// `NULL`
+    ///
+    /// The ANSI specification technically allows NULL constraints to have a
```
nickolay: Sorry for nit-picking: I [don't believe ANSI SQL allows CONSTRAINT foo NULL](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#column-constraint-definition) - it's Postgres that seems to allow it judging from its docs...

src/sqlast/ddl.rs
```diff
@@ -84,3 +79,106 @@ impl ToString for TableConstraint {
...
+                .iter()
+                .map(|c| format!(" {}", c.to_string()))
+                .collect::<Vec<_>>()
+                .join("")
+        )
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Hash)]
+pub enum ColumnConstraint {
```
nickolay: The current AST seems to be partially modeled after Postgres in which any column option (except COLLATE) is called a "column constraint", even though it doesn't constrain anything. Shall we add "GENERATED" to _this_ struct when it's implemented? Other options, like MS-specific ones? Would we keep the `ColumnConstraint` name then? If we plan to rename, it might be better to rename it before landing.

src/sqlast/ddl.rs
benesch: Oh man, good point. What a nightmare. Fixed.

src/sqlast/ddl.rs
benesch: Good points. I've pretty substantially reworked the design of the AST. I've renamed the main enum to `ColumnOption`, to better reflect the fact that it might not contain only constraints. I've also moved the optional name to a wrapping `ColumnOptionDef` struct, which better allows *any* of the options to take on a name, to allow for Postgres syntax, where practically everything can have a name. 

I think this should provide the right amount of future flexibility. GENERATED naturally fits as a new enum variant in `ColumnOption`, and hopefully it's less grating now that the struct doesn't refer to "constraints." MSSQL-specific options can also go into `ColumnOption`, if that makes sense, or they can live as fields on `SQLColumnDef`, if they're the kind of options (like SPARSE) that can't be named and must appear in a fixed order.

nickolay:
Looks great, thanks!

PR #94, opened 2019-06-03
Support CREATE TABLE with no columns

benesch:
This is weird, but supported by PostgreSQL.


benesch:
> To be clear: I don't ask you to move the test, rather wondering what you think about organizing the tests in such a way and if you'd object to moving it there at a later point.

I'm very much in favor in that! I almost wonder if we should aim for dialects to explicitly reject syntax they don't support. I.e., perhaps we should be explicitly testing that `CREATE TABLE t ()` returns an error on all the non-Postgres dialects.

PR #95, opened 2019-06-03
Support hexadecimal string literals

benesch:


src/sqlparser.rs
```diff
@@ -1037,6 +1040,16 @@ impl Parser {
                 Token::NationalStringLiteral(ref s) => {
                     Ok(Value::NationalStringLiteral(s.to_string()))
                 }
+                Token::HexStringLiteral(ref s) => {
+                    if let Some(bad) = s.chars().find(|c| !c.is_ascii_hexdigit()) {
```
nickolay: [The spec](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#binary-string-literal) is more lenient, allowing whitespace as well, as far as I can see. Perhaps leave validation to the consumers?

src/sqltokenizer.rs
```diff
@@ -286,6 +289,21 @@ impl<'a> Tokenizer<'a> {
                         }
                     }
                 }
+                x @ 'x' | x @ 'X' => {
```
nickolay: Which dialect supports the lower-case 'x'?

src/sqltokenizer.rs
```diff
@@ -286,6 +289,21 @@ impl<'a> Tokenizer<'a> {
...
+                    chars.next(); // consume, to check the next char
+                    match chars.peek() {
+                        Some('\'') => {
+                            // X'...' - a <national character string literal>
```
nickolay: `<binary string literal>`

src/sqltokenizer.rs
benesch: PostgreSQL, at least. I've added a comment to this effect.

src/sqltokenizer.rs
benesch: Good catch, thanks.

src/sqlparser.rs
benesch: Sure thing. Done.

PR #96, opened 2019-06-03
Support EXTRACT function-like operator

benesch:
The EXTRACT function, for extracting components of a date from a
timestamp, has special syntax: `EXTRACT(<field> FROM <timestamp>)`.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -100,6 +100,10 @@ pub enum ASTNode {
         expr: Box<ASTNode>,
         data_type: SQLType,
     },
+    SQLExtract {
+        field: SQLDateTimeField,
+        expr: Box<ASTNode>,
+    },
     /// `expr COLLATE collation`
     SQLCollate {
         expr: Box<ASTNode>,
@@ -186,6 +190,9 @@ impl ToString for ASTNode {
                 expr.as_ref().to_string(),
                 data_type.to_string()
             ),
+            ASTNode::SQLExtract { field, expr } => {
+                format!("EXTRACT({} FROM {})", field.to_string(), expr.to_string())
+            }
             ASTNode::SQLCollate { expr, collation } => format!(
                 "{} COLLATE {}",
                 expr.as_ref().to_string(),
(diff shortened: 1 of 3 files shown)


nickolay:
I'd prefer `field: SQLDateTimeField` to be a `String` instead of a custom enum, as everyone has their own idea of what values are allowed here:
* [ANSI](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#time-zone-field) defines `TIMEZONE_HOUR | TIMEZONE_MINUTE` in addition to the ones you have implemented,
* [Oracle](https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions050.htm) additionally has TIMEZONE_ABBR and TIMEZONE_REGION,
* and [MySQL](https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_extract) wins the contest with [a dozen of additional supported fields](https://dev.mysql.com/doc/refman/5.5/en/expressions.html#temporal-intervals)... 

Alternatively, I'd be happy to hear any suggestions on how to deal with future PRs wanting those extensions to be supported - just add everything to the enum?

nickolay:
Given that #99 depends on this, I'm fine with merging this "as is", as long as we agree on the future direction.

benesch:
> Given that #99 depends on this, I'm fine with merging this "as is", as long as we agree on the future direction.

Ok, great, I'm going to go ahead and merge, as I'm definitely OK with SQLDateTimeField evolving to a string in the future.

> Alternatively, I'd be happy to hear any suggestions on how to deal with future PRs wanting those extensions to be supported - just add everything to the enum?

Yeah, I don't see anything particularly appealing here. All of the options I can think of have some serious cons:

    * Use conditional compilation to add enum variants based on feature flags. This is appealing because users would get all the usual compiler support around exhaustive matching of enums, and wouldn't see anything irrelevant to their dialect. But we'd seriously break the existing API, because users would have to choose, at compile time, what dialect to support. And perhaps there are users who depend on being able to parse multiple dialects?

    * Write dialect-specific ASTs (that are largely copies of the general AST, but omitting irrelevant fields and enum variants), and then write transformations that lower the general AST to the dialect-specific ASTs. This would be a colossal amount of work for any change that affected multiple dialects (i.e., most dialects).

    * Make the enum generic over `Dialect`, and add a `DialectSpecific(_)` variant. I think @andygrove suggested this somewhere. This is a bit ugly, but maybe a nice balance between maintainability and user convenience.

PR #98, opened 2019-06-03
Support column names in more places

benesch:
See individual commits for details.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -25,7 +25,7 @@ use std::ops::Deref;
 pub use self::ddl::{AlterTableOperation, TableConstraint};
 pub use self::query::{
     Cte, Fetch, Join, JoinConstraint, JoinOperator, SQLOrderByExpr, SQLQuery, SQLSelect,
-    SQLSelectItem, SQLSetExpr, SQLSetOperator, SQLValues, TableFactor,
+    SQLSelectItem, SQLSetExpr, SQLSetOperator, SQLValues, TableAlias, TableFactor,
 };
 pub use self::sqltype::SQLType;
 pub use self::value::Value;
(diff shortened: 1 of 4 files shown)


benesch:
> I'm always unsure if I want to introduce a new struct vs tolerate some duplication in existing structs, so if you have a couple of words to say about why you chose to extract TableAlias, I'd appreciate it. Note that I don't have anything against it, so just merging is fine as well.

Sure thing! I thought having a separate struct that could be wrapped in an Option better communicated the set of possibilities. That is, sticking both a `table_alias` and a `column_aliases` field in the outer struct, like

```rust
struct Table {
    table_alias: Option<SQLObjectName>,
    column_aliases: Vec<SQLIdentifier>
}
```

indicates that it might be valid to have column aliases without a table alias, when it is not in fact valid. (You can specify a table alias without specifying column aliases, but you cannot specify a column alias without specifying a table alias.)

> (Ouch, this probably conflicts with the recently landed "WITH options" PR -- the expected order in CREATE VIEW is view_name (columns, list) WITH (options) AS I think? Having dealt with merge conflicts a lot, I feel your pain - thanks for doing this again! and if I can make this easier, please let me know.)

No problem! The rebase is in, and the conflicts weren't too bad. I really appreciate the speed with which you're moving through these PRs, and I'm happy to do a bit of work to split and reintegrate these changes to make the review easier. In other words, it's the overall rate of divergence that's scary and time-consuming, not the latency of any individual review, and the rate of divergence is very solidly negative at the moment! :)

nickolay:
> You can specify a table alias without specifying column aliases, but you cannot specify a column alias without specifying a table alias.

Ah, I have not thought of this. Thanks for clarifying - this makes perfect sense to me.

PR #100, opened 2019-06-03
Support nested joins

### LINKED ISSUE #83: Support join associativity

benesch:
Fix #83.

@nickolay this is everything, for the moment! Feel free to review/merge in any order. I'll take care of rebasing whatever winds up conflicting (though feel free to rebase yourself, if you're so inclined).

### DIFF
--- a/src/sqlast/query.rs
+++ b/src/sqlast/query.rs
@@ -215,6 +215,10 @@ pub enum TableFactor {
         subquery: Box<SQLQuery>,
         alias: Option<TableAlias>,
     },
+    NestedJoin {
+        base: Box<TableFactor>,
+        joins: Vec<Join>,
+    },
 }
 
 impl ToString for TableFactor {
@@ -252,6 +256,13 @@ impl ToString for TableFactor {
                     s += &format!(" AS {}", alias.to_string());
                 }
                 s
+            }
+            TableFactor::NestedJoin { base, joins } => {
+                let mut s = base.to_string();
+                for join in joins {
+                    s += &join.to_string();
+                }
+                format!("({})", s)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


benesch:
> I thought fixing this would have to deal with the weird way we store joins, with FROM foo NATURAL JOIN bar, baz NATURAL JOIN qux stored as vec![Natural "bar", Implicit "baz", Natural "qux"] instead of From [Natural ("foo", "bar"), Natural("baz", "qux")].

Oh wow, that is some headscratching syntax. I've never seen anyone combine the implicit and explicit syntaxes before! I wasn't even aware that was possible. Sigh.

> This doesn't handle ((a NATURAL JOIN b)) just as we didn't handle ((SELECT ..)) derived tables, which I think is OK.

Oh, good catch. I'm going to merge for now, but I'll investigate making a follow-on PR to support double parens.

benesch:
> > This doesn't handle ((a NATURAL JOIN b)) just as we didn't handle ((SELECT ..)) derived tables, which I think is OK.
> 
> Oh, good catch. I'm going to merge for now, but I'll investigate making a follow-on PR to support double parens.

Wait... I think we do support this! Unless I'm misunderstanding. See #108.

benesch:
> > I thought fixing this would have to deal with the weird way we store joins, with FROM foo NATURAL JOIN bar, baz NATURAL JOIN qux stored as vec![Natural "bar", Implicit "baz", Natural "qux"] instead of From [Natural ("foo", "bar"), Natural("baz", "qux")].
> 
> Oh wow, that is some headscratching syntax. I've never seen anyone combine the implicit and explicit syntaxes before! I wasn't even aware that was possible. Sigh.

I took a stab at fixing this in #109. Discussion to follow there.

PR #102, opened 2019-06-03
Speed up CI

benesch:
Details in individual commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: rust
 addons:
   apt:
     packages:
+      - kcov
       - libcurl4-openssl-dev
       - libelf-dev
       - libdw-dev
@@ -12,6 +13,11 @@ addons:
       - cmake
     sources:
       - kalakris-cmake
+      # The version of kcov shipped with Xenial (v25) doesn't support the
+      # --verify option that `cargo coveralls` passes. This PPA has a more
+      # up-to-date version. It can be removed if Ubuntu ever ships a newer
+      # version, or replaced with another PPA if this one falls out of date.
+      - sourceline: ppa:sivakov512/kcov
 
 rust:
   - stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,18 @@ rust:
 
 before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
-  - cargo install --force cargo-travis && export PATH=$HOME/.cargo/bin:$PATH
+  - export PATH=$HOME/.cargo/bin:$PATH
+  # `cargo install` fails if the specified binary is already installed, and
+  # doesn't yet support a `--if-not-installed` option [0], so for now assume
+  # failures mean the package is already installed. If installation truly
+  # failed, the build will fail later anyway, when we try to use the installed
+  # binary. Note that `cargo install --force` is not a solution, as it always
+  # rebuilds from scratch, ignoring the cache entirely.
+  #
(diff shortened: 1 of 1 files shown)


benesch:
This brings builds down to about 4m with a warm cache, which is a huge upgrade over the ~14m build times before.

.travis.yml
```diff
@@ -1,24 +1,34 @@
-dist: trusty
+dist: xenial
 sudo: required
-cache: cargo
```
nickolay: Given that [the docs say](https://docs.travis-ci.com/user/caching/#rust-cargo-cache) "[cache: cargo] caches $HOME/.cargo and $TRAVIS_BUILD_DIR/target" and the new value of `cache` below, you effectively removed `target` and `.cargo/registry` from the cache.

Comparing [a build log before these changes](https://travis-ci.org/andygrove/sqlparser-rs/builds/540583671?utm_source=github_status&utm_medium=notification) and [a build with the new config](https://travis-ci.org/andygrove/sqlparser-rs/builds/540865573?utm_source=github_status&utm_medium=notification), the downloading and re-building takes more time with the newer build, as expected. I suppose there's a benefit to not accumulating the dead weight in the cache? Care to clarify?

.travis.yml
```diff
@@ -1,24 +1,34 @@
...
     packages:
+      - kcov
       - libcurl4-openssl-dev
       - libelf-dev
       - libdw-dev
       - binutils-dev
       - cmake
     sources:
       - kalakris-cmake
+      - sourceline: ppa:sivakov512/kcov
```
nickolay: I guess this is needed only for the new `kcov` package reference above to work, despite being added in a separate commit? In case the PPA gets broken, we can just remove these two lines, right?

I'm curious how you found the PPA, there's not a lot of Google hits for its name.

.travis.yml
```diff
@@ -1,24 +1,34 @@
-dist: trusty
```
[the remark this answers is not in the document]
benesch: That's right. I added a note to the commit message explaining this. Xenial does tend to boot up a bit faster than Trusty, due to some changes Travis made in the Xenial build environment, so it falls on the general umbrella of CI speedups. :)

.travis.yml
benesch: Yeah, at some point the scale will tip in favor of a clean `cargo build`, as downloading the cache will take longer than just compiling from scratch. I'm happy to kick that can down the road, though, since I think you're right that, at the moment, the scales are firmly in favor of using the default `cache: cargo` settings. I've updated the PR accordingly.

.travis.yml
benesch: > I guess this is needed only for the new kcov package reference above to work, despite being added in a separate commit? In case the PPA gets broken, we can just remove these two lines, right?

Yep, that's right. I screwed up a rebase, which is why this ended up in the wrong commit—but I've fixed that now—thanks! I've added some details (like "feel free to remove this PPA line if you can get a good version of kcov some other way") in a comment in the file.

> I'm curious how you found the PPA, there's not a lot of Google hits for its name.

Heh, I googled "kcov ppa" in the hopes that someone had already done the hard work of creating a PPA.

nickolay:
Wow, I'm not sure if I'm more impressed with the 3x+ speedup you've achieved or with the way you document your changes! Please merge. (As for me, I can't even find the "approve" button :)

benesch:
> Wow, I'm not sure if I'm more impressed with the 3x+ speedup you've achieved or with the way you document your changes! Please merge.

And I've been impressed by the speed at which you're reviewing all these PRs, considering how thoughtful your reviews have been! Thanks again for all the work!

PR #106, opened 2019-06-06
Transaction support

SamuelMarks:
Basic transaction support


benesch:
Thanks for the PR, @SamuelMarks. Transaction statements are a great addition!

I have one rather large comment about the design of the new AST node, though. From a syntactic perspective, `BEGIN` and `COMMIT` should be modeled as their own standalone statements. We don't want to be in the business of deciding exactly when and where a transaction starts and ends, as that's the job of a semantic analyzer or planner downstream. (For example, a `ROLLBACK` statement should also end a transaction.)

Would you be open to introducing separate `SQLStartTransaction` and `SQLCommit` variants, rather than the one `SQLTransaction` variant you've introduced? (Note that `BEGIN` is actually nonstandard syntax, albeit widely-supported syntax, and we typically try to use the standard SQL names where possible, which is why I've suggested naming the variant after `START TRANSACTION` instead.)

Let me know if you'd like any help with the refactor!

benesch:
Great, thanks for the refactor! I've pushed some additional changes to support the various options that `START TRANSACTION` and `COMMIT` accept, and added support for `BEGIN`, `ROLLBACK`, and `SET TRANSACTION` as well. Let me know if I've somehow made this harder for you to use downstream, but I think the patch is strictly more general than it was before.

@nickolay, can I get your eyes on this?

nickolay:
@benesch I trust your review, especially since I'm not familiar with transaction control syntax.

I'd only like to remind that it [might be helpful for the future work on dialects to explicitly mark the dialect-specific parts of the parser and the relevant tests](https://github.com/andygrove/sqlparser-rs/pull/94#issuecomment-498509045).

benesch:
> @benesch I trust your review, especially since I'm not familiar with transaction control syntax.

Ok, great! Merging.

> I'd only like to remind that it [might be helpful for the future work on dialects to explicitly mark the dialect-specific parts of the parser and the relevant tests](https://github.com/andygrove/sqlparser-rs/pull/94#issuecomment-498509045).

Sure thing. Perhaps surprisingly, the only dialect-specific bit here is that PostgreSQL allows omitting the commas between transaction modes. I'd previously only noted that in the test, but I've copied that comment to the parsing code too.

PR #107, opened 2019-06-06
Fix precedence of unary NOT

benesch:
get_next_precedence deals with left-binding power, not right binding
power. Therefore, when it encounters a standalone NOT operator (i.e., a
"NOT" token that is not followed by "BETWEEN", "LIKE", or "IN"), it
should return 0, because unary NOT is not an infix operator, it's a
prefix operator, and therefore it has no left-binding power.

@nickolay this is as discussed in #93. Making this change does indeed mean that parsing DEFAULT doesn't need to do any special precedence hacks.


nickolay:
I would keep the `const UNARY_NOT_PREC`, otherwise this looks great.

benesch:
> I would keep the const UNARY_NOT_PREC, otherwise this looks great.

You bet. Done.

PR #108, opened 2019-06-08
Test that redundant nesting is supported

benesch:
SELECT * FROM (((SELECT 1))) is just as valid as
SELECT * FROM (SELECT 1). Add a test to ensure that we can parse the
first form.

Addresses a comment from #100.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -1466,7 +1466,7 @@ fn parse_join_nesting() {
     }
 
     macro_rules! nest {
-        ($base:expr, $($join:expr),*) => {
+        ($base:expr $(, $join:expr)*) => {
             TableFactor::NestedJoin {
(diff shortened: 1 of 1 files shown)


benesch:
> Ah, I see: unlike [in the spec](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#parenthesized-joined-table), TableFactor::NestedJoin can represent a non-join TableFactor in parens, so `((SELECT 1))` gets represented as `NestedJoin{ base: Derived(Select...), joins: vec![] }`. The spec has separate productions for `((foo NATURAL JOIN bar))` and for `FROM ((SELECT 1 a))x`, the latter is allowed in [`<query primary>`](https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#query-primary).

Oh boy. That'll be fun to deal with. 🙃

PR #109, opened 2019-06-08
Properly handle mixed implicit and explicit joins

benesch:
Parse a query like

    SELECT * FROM a NATURAL JOIN b, c NATURAL JOIN d

as the SQL specification requires, i.e.:

    from: [
        TableReference {
            relation: TableFactor::Table("a"),
            joins: [Join {
                relation: TableFactor::Table("b"),
                join_operator: JoinOperator::Natural,
            }]
        },
        TableReference {
            relation: TableFactor::Table("c"),
            joins: [Join {
                relation: TableFactor::Table("d"),
                join_operator: JoinOperator::Natural,
            }]
        }
    ]

Previously we were parsing such queries as

    relation: TableFactor::Table("a"),
    joins: [
        Join {
            relation: TableFactor::Table("b"),
            join_operator: JoinOperator::Natural,
        },
        Join {
            relation: TableFactor::Table("c"),
            join_operator: JoinOperator::Implicit,
        },
        Join {
            relation: TableFactor::Table("d"),
            join_operator: JoinOperator::Natural,
        },
    ]

which did not make the join hierarchy clear.

### DIFF
--- a/src/sqlast/query.rs
+++ b/src/sqlast/query.rs
@@ -113,9 +113,7 @@ pub struct SQLSelect {
     /// projection expressions
     pub projection: Vec<SQLSelectItem>,
     /// FROM
-    pub relation: Option<TableFactor>,
-    /// JOIN
-    pub joins: Vec<Join>,
+    pub from: Vec<TableWithJoins>,
     /// WHERE
     pub selection: Option<ASTNode>,
     /// GROUP BY
@@ -131,11 +129,8 @@ impl ToString for SQLSelect {
             if self.distinct { " DISTINCT" } else { "" },
             comma_separated_string(&self.projection)
         );
-        if let Some(ref relation) = self.relation {
-            s += &format!(" FROM {}", relation.to_string());
-        }
-        for join in &self.joins {
-            s += &join.to_string();
+        if !self.from.is_empty() {
+            s += &format!(" FROM {}", comma_separated_string(&self.from));
         }
         if let Some(ref selection) = self.selection {
             s += &format!(" WHERE {}", selection.to_string());
@@ -197,6 +192,22 @@ impl ToString for SQLSelectItem {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Hash)]
+pub struct TableWithJoins {
+    pub relation: TableFactor,
+    pub joins: Vec<Join>,
+}
+
+impl ToString for TableWithJoins {
+    fn to_string(&self) -> String {
+        let mut s = self.relation.to_string();
+        for join in &self.joins {
+            s += &join.to_string();
+        }
+        s
+    }
+}
+
 /// A table name or a parenthesized subquery with an optional alias
 #[derive(Debug, Clone, PartialEq, Hash)]
 pub enum TableFactor {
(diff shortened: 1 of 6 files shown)


benesch:
@nickolay I realize this is likely to be a bit prickly, as it is a substantial refactoring of a core AST. I wanted to flesh out what it looked like, though, and it didn't turn out to be too bad.

Note that because we now support nested joins, we could alternatively parse `FROM a NATURAL JOIN b, c NATURAL JOIN d` as:

```
relation: TableFactor::Table("a"),
joins: [
    Join { relation: TableFactor::Table("b"), operator: Natural },
    Join {
        relation: TableFactor::NestedJoin {
            base: TableFactor::Table("c"),
            joins: [Join { relation: TableFactor::("d"), operator: Natural }]
        },
        operator: Cross
    }
]
```

But perhaps downstream consumers will want to distinguish between implicit and explicit cross joins.

src/sqlast/query.rs
```diff
@@ -197,6 +192,23 @@ impl ToString for SQLSelectItem {
     }
 }
 
+/// A table reference is one table factor followed by any number of joins.
```
nickolay: I get that this follows the names of the productions in ANSI SQL syntax, but maybe `TableWithJoins`?

src/sqlast/query.rs
```diff
@@ -215,10 +227,7 @@ pub enum TableFactor {
         subquery: Box<SQLQuery>,
         alias: Option<TableAlias>,
     },
-    NestedJoin {
-        base: Box<TableFactor>,
-        joins: Vec<Join>,
-    },
+    NestedJoin(Box<TableReference>),
```
nickolay: Given https://github.com/andygrove/sqlparser-rs/pull/108#pullrequestreview-247376543 maybe call it `TableFactor::Nested`?

src/sqlparser.rs
```diff
@@ -1526,14 +1528,19 @@ impl Parser {
...
-            relation,
-            joins,
             group_by,
             having,
         })
     }
 
+    pub fn parse_table_reference(&mut self) -> Result<TableReference, ParserError> {
+        let relation = self.parse_table_factor()?;
+        let joins = self.parse_joins()?;
```
nickolay: Would `parse_joins` be useful elsewhere? Perhaps inline it here?

src/sqlast/query.rs
benesch: Something about this whole design wasn't sitting right with me, so I decided to address the underlying issue you brought up in #108, rather than renaming this variant. I've pushed two commits that make it so that NestedJoin always contains a table reference with at least one join. A nice side effect of this work was supporting derived tables with set operations, like `SELECT * FROM (((SELECT 1) UNION (SELECT 2)) t1 NATURAL JOIN t2)`.

nickolay:
Could you please extract the new `parse_top_level()` test (with the required changes in `parse_statement`) into its own commit and land it along with the already reviewed changes in this PR? Or land all of it and let me think about it later?

Making two parse attempts bothers me, plus the names from the spec (`parse_derived_table_factor`/`parse_table_reference`) are hard to remember; I'll need some time to let it sink in.

src/sqlast/query.rs
benesch: Yeah, I find them hard to remember too. I've taken your suggestion for `TableWithJoins`.

src/sqlparser.rs
benesch: Nope, not that I can see. Inlined!

benesch:
> Could you please extract the new parse_top_level() test (with the required changes in parse_statement) into its own commit and land it along with the already reviewed changes in this PR? 

Yep, working on that now. I'll open a new PR for the remaining changes as soon as this lands.

PR #110, opened 2019-06-09
Minor code clean-ups

nickolay:


nickolay:
Thanks for the review! I've also tried factoring the common parts even further, like:
```
        let options = self.parse_comma_separated(|parser| {
            let name = parser.parse_identifier()?;
            parser.expect_token(&Token::Eq)?;
            let value = parser.parse_value()?;
            Ok(SQLOption { name, value })
        })?;
```
..but decided against it for now, as I didn't like that sometimes the parser is referred to as `self`, and sometimes as `parser`.

PR #111, opened 2019-06-10
Refine join parsing

benesch:
Additional join tweaks for better compliance with the standard. Extracted from #109.

@nickolay no rush at all on this! I'm not currently in need of the changes here, but I wanted to cross the i's and dot the t's while the code was fresh in my mind.


benesch:
Yeah, the new commits look great! I'll squash and merge in just a moment.

PR #112, opened 2019-06-10
Improve consistency of binary/unary op nodes

benesch:
This was extracted from #105. See individual commits for details.


benesch:
Thanks for the review, @andygrove!

PR #68, opened 2019-05-18
Add LIMIT as RESERVED_FOR_TABLE_ALIAS

ivanceras:
Closes [Issue67](https://github.com/andygrove/sqlparser-rs/issues/67)

---
This change is [Reviewable](https://reviewable.io/reviews/andygrove/sqlparser-rs/68)

PR #116, opened 2019-06-16
Support HAVING/LIMIT/OFFSET/FETCH without FROM and other follow-ups

nickolay:
#68 made me review the keywords in `RESERVED_FOR_TABLE_ALIAS` and `RESERVED_FOR_COLUMN_ALIAS`, and I realized more of them need to be "Reserved as both a table and a column alias".

Also after `TableAlias` was introduced in #98 to represent `alias [ (col1, col2, ...) ]` as part of a TableFactor, it makes sense to re-use it in `Cte`.

### DIFF
--- a/src/dialect/keywords.rs
+++ b/src/dialect/keywords.rs
@@ -420,7 +420,7 @@ define_keywords!(
 /// can be parsed unambiguously without looking ahead.
 pub const RESERVED_FOR_TABLE_ALIAS: &[&str] = &[
     // Reserved as both a table and a column alias:
-    WITH, SELECT, WHERE, GROUP, ORDER, LIMIT, OFFSET, FETCH, UNION, EXCEPT, INTERSECT,
+    WITH, SELECT, WHERE, GROUP, HAVING, ORDER, LIMIT, OFFSET, FETCH, UNION, EXCEPT, INTERSECT,
     // Reserved only as a table alias in the `FROM`/`JOIN` clauses:
     ON, JOIN, INNER, CROSS, FULL, LEFT, RIGHT, NATURAL, USING,
 ];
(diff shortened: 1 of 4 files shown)


src/dialect/keywords.rs
```diff
@@ -420,16 +420,16 @@ define_keywords!(
...
-    ON, JOIN, INNER, CROSS, FULL, LEFT, RIGHT, NATURAL, USING, LIMIT, OFFSET, FETCH,
+    ON, JOIN, INNER, CROSS, FULL, LEFT, RIGHT, NATURAL, USING,
 ];
 
 /// Can't be used as a column alias, so that `SELECT <expr> alias`
 /// can be parsed unambiguously without looking ahead.
 pub const RESERVED_FOR_COLUMN_ALIAS: &[&str] = &[
     // Reserved as both a table and a column alias:
-    WITH, SELECT, WHERE, GROUP, ORDER, UNION, EXCEPT, INTERSECT,
+    WITH, SELECT, WHERE, GROUP, ORDER, LIMIT, OFFSET, FETCH, UNION, EXCEPT, INTERSECT,
```
benesch: I know this isn't the fault of this commit, but it's a shame that these keywords have to be duplicated between the two constants. I don't have any obvious solutions, and definitely doesn't need to be considered in this PR; just thought I'd flag it in case you happened to have any ideas.

nickolay: I don't like it either, but [the solutions](https://users.rust-lang.org/t/how-to-concatenate-array-literals-in-compile-time/21141) seemed too complex for a trivial problem...

These lists will become dialect-specific at some point, and we would be able to build the reserved lists in the dialect constructor then, I guess.

PR #119, opened 2019-06-18
Rename ASTNode to Expr

benesch:
The ASTNode enum was confusingly named. In the past, the name made
sense, as the enum contained nearly all of the nodes in the AST, but
over time, pieces have been split into different structs, like
SQLStatement and SQLQuery. The ASTNode enum now contains only contains
expression nodes, so Expr is a better name.

Extracted from #105.

### DIFF
--- a/src/sqlast/mod.rs
+++ b/src/sqlast/mod.rs
@@ -53,7 +53,7 @@ pub type SQLIdent = String;
 /// (e.g. boolean vs string), so the caller must handle expressions of
 /// inappropriate type, like `WHERE 1` or `SELECT 1=1`, as necessary.
 #[derive(Debug, Clone, PartialEq, Hash)]
-pub enum ASTNode {
+pub enum Expr {
     /// Identifier e.g. table name or column name
     SQLIdentifier(SQLIdent),
     /// Unqualified wildcard (`*`). SQL allows this in limited contexts, such as:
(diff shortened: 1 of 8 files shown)


src/sqlast/query.rs
```diff
@@ -177,9 +177,9 @@ impl ToString for Cte {
 #[derive(Debug, Clone, PartialEq, Hash)]
 pub enum SQLSelectItem {
     /// Any expression, not followed by `[ AS ] alias`
-    UnnamedExpression(ASTNode),
+    UnnamedExpression(Expr),
```
nickolay: Should we keep `UnnamedExpression` and `ExpressionWithAlias` or rename them to use `Expr`? Same for parse_.._expression() methods.

src/sqlast/mod.rs
```diff
@@ -69,55 +69,55 @@ pub enum ASTNode {
...
+        right: Box<Expr>,
     },
     /// Unary operation e.g. `NOT foo`
     SQLUnaryOp {
         op: SQLUnaryOperator,
-        expr: Box<ASTNode>,
+        expr: Box<Expr>,
     },
     /// CAST an expression to a different data type e.g. `CAST(foo AS VARCHAR(123))`
     SQLCast {
```
nickolay: My version of `cargo fmt` reformats this on one line. I wonder why CI doesn't catch it..

benesch: What version of the Rust toolchain are you using? Perhaps the formatting here is different in different versions.

benesch: Never mind. I think something's broken in CI.

src/sqlast/query.rs
benesch: Oh, good idea! Done.

nickolay:
This looks ready to merge to me!

PR #121, opened 2019-06-19
[do-not-merge] check that ci enforces rustfmt

benesch:
Sanity check that CI actually enforces rustfmt. I suspect something is broken.

### DIFF
--- a/src/sqlast/query.rs
+++ b/src/sqlast/query.rs
@@ -17,7 +17,9 @@ use super::*;
 #[derive(Debug, Clone, PartialEq, Hash)]
 pub struct SQLQuery {
     /// WITH (common table expressions, or CTEs)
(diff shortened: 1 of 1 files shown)


benesch:
Well, that proves it.

benesch:
Ok, turns out the crash is a rustfmt bug that I've filed upstream: https://github.com/rust-lang/rustfmt/pull/3655.

benesch:
Ok, once rustfmt is fixed we can squash this PR down and merge, and we'll actually have rustfmt working in CI. Whoops!

benesch:
@nickolay, does this LGTY? The panic in rustfmt (https://github.com/rust-lang/rustfmt/pull/3655) is finally fixed in nightly, so should be good to go.

PR #122, opened 2019-06-25
Remove "SQL" prefix from "SQLDateTimeField" struct

benesch:
I realized a moment too late that I'd missed a type name in
when removing the "SQL" prefix from types in ac555d7e8. As far as I can
tell, this was the only oversight.


benesch:
> (Not sure there’s a point in reviewing changes like this before merging.)

Sure, happy to just merge tiny stuff like this without waiting for approval!

PR #123, opened 2019-06-25
Implement Eq alongside Hash

benesch:
It turns out implementing Hash alone is not very useful, as
std::collection::HashMap keys are required to implement both Hash and
Eq.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -13,7 +13,7 @@
 use super::ObjectName;
 
 /// SQL data types
-#[derive(Debug, Clone, PartialEq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum DataType {
     /// Fixed-length character type e.g. CHAR(10)
     Char(Option<u64>),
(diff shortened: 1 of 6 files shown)


nickolay:
Would you mind commenting on why Eq makes sense for AST types? Otherwise I’ll get to it later - i’d have to reread the docs before I can say I understand this change.

Not for this PR: have you looked into whether it’s possible to share a common set of `#[...]` attributes across all of our AST types?

benesch:
> Would you mind commenting on why Eq makes sense for AST types? Otherwise I’ll get to it later - i’d have to reread the docs before I can say I understand this change.

You bet! The difference between `PartialEq` and `Eq` is that `PartialEq` does not require reflexivity (a == a for all a), while `Eq` does. The canonical example is that floats cannot implement `Eq` because `NaN` is not equal to itself. (Aside: this isn't an issue for us since we're already using OrderedFloat to provide Hash/Eq/PartialEq/Cmp/PartialCmd on our Value::Float variant.) There's essentially no situation where you don't want `Eq` if it makes sense, since otherwise you can't use the type as a key in a hash map.

benesch:
Awesome, thanks for the review!

PR #124, opened 2019-06-30
implement fmt::Display instead of ToString

vemoo:
As mentioned in the documentation for [`ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html) implementing [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) should be prefered.

This should be faster and use less memory because no intermediate `String`s are created.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -11,6 +11,7 @@
 // limitations under the License.
 
 use super::ObjectName;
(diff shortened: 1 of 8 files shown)


nickolay:
>if @nickolay has any thoughts

Nope - if you think this is right, let's land it!

PR #126, opened 2019-07-01
Update comments after the renaming done in PR #105

nickolay:


nickolay:
@benesch that was fast! Thanks.

benesch:
Lucky coincidence! I just happened to be in the process of reviewing #124 so I was primed to see your PR come through.

PR #130, opened 2019-07-08
Don't lose precision when parsing decimal fractions

benesch:
The SQL standard requires that numeric literals with a decimal point,
like 1.23, are represented exactly, up to some precision. That means
that parsing these literals into f64s is invalid, as it is impossible
to represent many decimal numbers exactly in binary floating point (for
example, 0.3).

This commit instead parses these literals into a big decimal type that
is capable of representing every decimal fraction exactly. Downstream
consumers can lower this representation into something more efficient
but less precise (like the old f64) if desired.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1183,7 +1183,13 @@ impl Parser {
                         return parser_err!(format!("No value parser for keyword {}", k.keyword));
                     }
                 },
-                Token::Number(ref n) => Ok(Value::Number(n.to_string())),
+                // The call to n.parse() returns a bigdecimal when the
+                // bigdecimal feature is enabled, and is otherwise a no-op
+                // (i.e., it returns the input string).
+                Token::Number(ref n) => match n.parse() {
+                    Ok(n) => Ok(Value::Number(n)),
+                    Err(e) => parser_err!(format!("Could not parse '{}' as number: {}", n, e)),
+                },
                 Token::SingleQuotedString(ref s) => Ok(Value::SingleQuotedString(s.to_string())),
                 Token::NationalStringLiteral(ref s) => {
                     Ok(Value::NationalStringLiteral(s.to_string()))
@@ -1192,6 +1198,16 @@ impl Parser {
                 _ => parser_err!(format!("Unsupported value: {:?}", t)),
             },
             None => parser_err!("Expecting a value, but found EOF"),
+        }
+    }
+
+    pub fn parse_number_value(&mut self) -> Result<Value, ParserError> {
+        match self.parse_value()? {
+            v @ Value::Number(_) => Ok(v),
+            _ => {
+                self.prev_token();
+                self.expected("literal number", self.peek_token())
+            }
         }
     }
 
@@ -1856,16 +1872,13 @@ impl Parser {
         if self.parse_keyword("ALL") {
             Ok(None)
         } else {
-            self.parse_literal_uint()
-                .map(|n| Some(Expr::Value(Value::Number(n.to_string()))))
+            Ok(Some(Expr::Value(self.parse_number_value()?)))
         }
     }
 
     /// Parse an OFFSET clause
     pub fn parse_offset(&mut self) -> Result<Expr, ParserError> {
-        let value = self
-            .parse_literal_uint()
-            .map(|n| Expr::Value(Value::Number(n.to_string())))?;
+        let value = Expr::Value(self.parse_number_value()?);
         self.expect_one_of_keywords(&["ROW", "ROWS"])?;
         Ok(value)
     }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1183,14 +1183,7 @@ impl Parser {
                         return parser_err!(format!("No value parser for keyword {}", k.keyword));
                     }
                 },
-                Token::Number(ref n) if n.contains('.') => match n.parse::<f64>() {
-                    Ok(n) => Ok(Value::Double(n.into())),
-                    Err(e) => parser_err!(format!("Could not parse '{}' as f64: {}", n, e)),
-                },
-                Token::Number(ref n) => match n.parse::<u64>() {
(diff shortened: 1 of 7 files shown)


benesch:
> One thing that bothers me is that we use Value::Date(String) in the AST for DATE literals, instead of bundling chrono; while for decimals you've elected to rely on bigdecimal. Is there a reason for this discrepancy or do you see us adding chrono for date/time values later?

Yes, this bothers me too! The reason I went with this is because it seemed really disruptive to remove all integer/float parsing, but I'm starting to lean towards that approach. There's a problem right now in that someone can write `SELECT 1234123412341234123412341242314123431241324321412341234`, which is perfectly parseable as an arbitrary-precision decimal in Postgres, and our SQL parser will fail to parse it because the literal isn't representable in a u64. Really the best thing to do is to return the literal as a string, and let downstream consumers handle it.

> Is having values parsed in the AST useful in context of an SQL engine? I haven't studied any SQL implementations, but I imagined one would have to make at least one pass over the AST (to resolve the object names), producing a different structure before inspecting the values. I'd like to know if that assumption is wrong.

Nope, it's not that useful. It was useful when we (Materialize) were just prototyping our SQL engine, but as its evolved we've found that the values parsing was too opinionated. It's fine when the parser produces *exactly* the literal representation we need (e.g., using this bigdecimal crate, the ordered float crate, etc.), but that approach certainly doesn't scale as we have additional users with different and conflicting requirements.

> I have a slight preference for adding a way to parse the literals to this crate as an optional cargo feature, but I realize that my obsession with build times can sometimes be unhealthy:)

Very on board with that! How do you feel about this as the standard `Value` enum?

```rust
pub enum Value {
    /// Numeric literal
    Number(String),
    /// 'string value'
    SingleQuotedString(String),
    /// N'string value'
    NationalStringLiteral(String),
    /// X'hex value'
    HexStringLiteral(String),
    /// Boolean value true or false
    Boolean(bool),
    /// `DATE '...'` literals
    Date(String),
    /// `TIME '...'` literals
    Time(String),
    /// `TIMESTAMP '...'` literals
    Timestamp(String),
    /// INTERVAL literals, roughly in the following format:
    /// `INTERVAL '<value>' <leading_field> [ (<leading_precision>) ]
    /// [ TO <last_field> [ (<fractional_seconds_precision>) ] ]`,
    /// e.g. `INTERVAL '123:45.67' MINUTE(3) TO SECOND(2)`.
    ///
    /// The parser does not validate the `<value>`, nor does it ensure
    /// that the `<leading_field>` units >= the units in `<last_field>`,
    /// so the user will have to reject intervals like `HOUR TO YEAR`.
    Interval {
        value: String,
        leading_field: DateTimeField,
        /// or as `__ TO SECOND(x)`.
    },
    /// `NULL` value
    Null,
}
```

It feels a bit strange at first, but seems quite sensible once you sit with it. We could retain the existing `Value` variants behind a feature flag, plus perhaps this decimal variant too. Let me rework this PR accordingly.

nickolay:
> How do you feel about this as the standard Value enum?

So a single `Value::Number(String)` instead of `Value::Long(u64)` / `Value::Double(OrderedFloat<f64>)`. That makes sense!

I'm not sure I understand: why provide an option to keep existing variants, instead of just providing an optional `bigdecimal`?

benesch:
> So a single Value::Number(String) instead of Value::Long(u64) / Value::Double(OrderedFloat<f64>). That makes sense!

Yep, this is done now! Please take a look when you get a chance.

> I'm not sure I understand: why provide an option to keep existing variants, instead of just providing an optional bigdecimal?

I think we're just saying the same thing in different ways. I coded up what I meant, at least, in the second commit in this PR—let me know if that's also what you had in mind!

src/parser.rs
```diff
@@ -1854,20 +1847,17 @@ impl Parser {
     }
 
     /// Parse a LIMIT clause
-    pub fn parse_limit(&mut self) -> Result<Option<Expr>, ParserError> {
+    pub fn parse_limit(&mut self) -> Result<Option<u64>, ParserError> {
```
nickolay: I'm not against this change, but I didn't make it sooner because at least [Postgres allows expressions here](https://www.postgresql.org/docs/11/sql-select.html#SQL-LIMIT).

src/parser.rs
```diff
@@ -1854,20 +1847,17 @@ impl Parser {
...
+            Ok(Some(self.parse_literal_uint()?))
         }
     }
 
     /// Parse an OFFSET clause
-    pub fn parse_offset(&mut self) -> Result<Expr, ParserError> {
-        let value = self
-            .parse_literal_uint()
-            .map(|n| Expr::Value(Value::Long(n)))?;
+    pub fn parse_offset(&mut self) -> Result<u64, ParserError> {
```
nickolay: Same here. This was added in this PR https://github.com/andygrove/sqlparser-rs/pull/69#issuecomment-497160698

src/parser.rs
```diff
@@ -1176,7 +1176,10 @@ impl Parser {
                         return parser_err!(format!("No value parser for keyword {}", k.keyword));
                     }
                 },
-                Token::Number(ref n) => Ok(Value::Number(n.to_string())),
+                Token::Number(ref n) => match n.parse() {
```
nickolay: I was confused by this at first, as the IDE suggested that `parse()` returns a `String`. Perhaps add a comment that this is for feature=bigdecimal?

src/parser.rs
benesch: Oh, good point. I've backed this bit out.

src/parser.rs
benesch: Backed this bit out too.

src/parser.rs
benesch: Sure thing, done!

nickolay:
> I believe I've addressed all your feedback.

Yep, thank you!

PR #135, opened 2019-08-13
Support MySQL `SHOW COLUMNS` statement

benesch:
Submitting this on behalf of @quodlibetor. This was extracted from our fork at @MaterializeInc and evolved to support a more complete version of the `SHOW COLUMNS` statement that MySQL supports, so that it is no longer Materialize-specific. @nickolay, mind taking a look?

Upstream grammar: https://dev.mysql.com/doc/refman/8.0/en/show-columns.html


benesch:
Added another commit to support `SHOW <var>` and `SET <var>`.

src/parser.rs
```diff
@@ -1593,6 +1593,30 @@ impl Parser {
         })
     }
 
+    pub fn parse_set(&mut self) -> Result<Statement, ParserError> {
```
nickolay: Oh, this will be fun. I've been holding off adding support for [MSSQL interpretation of `SET`](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql?view=sql-server-2017).

src/parser.rs
```diff
@@ -1593,6 +1593,30 @@ impl Parser {
...
+                local: modifier == Some("LOCAL"),
+                variable,
+                value,
+            })
+        } else if variable == "TRANSACTION" && modifier.is_none() {
+            Ok(Statement::SetTransaction {
+                modes: self.parse_transaction_modes()?,
+            })
+        } else {
+            self.expected("variable name", self.peek_token())
```
nickolay: I think an equals sign is expected here?

src/parser.rs
benesch: Oh boy. We may need to start thinking more about dialect-specific code.

src/parser.rs
benesch: Oh, good catch, thanks! Fixed, and added some error test cases as well.

PR #139, opened 2019-09-13
Allow semicolon after start transaction

gaffneyk:
Currently, the parser returns an `Err` if there is a semicolon after a start transaction statement, despite this being valid SQL. For example,
```sql
START TRANSACTION
```
parses correctly, but
```sql
START TRANSACTION; SELECT * FROM T
```
does not.

This change would allow a semicolon delimiter/terminator after a start transaction statement.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2026,7 +2026,7 @@ impl Parser {
                 TransactionMode::AccessMode(TransactionAccessMode::ReadOnly)
             } else if self.parse_keywords(vec!["READ", "WRITE"]) {
                 TransactionMode::AccessMode(TransactionAccessMode::ReadWrite)
-            } else if required || self.peek_token().is_some() {
+            } else if required {
                 self.expected("transaction mode", self.peek_token())?
             } else {
                 break;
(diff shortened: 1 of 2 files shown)


nickolay:
@gaffneyk thanks for catching this, and for the patch! Would you like to add a test as well, so that this doesn't regress in the future?

@benesch, could you take a look?

Why don't we let the [common code](https://github.com/andygrove/sqlparser-rs/blob/abf68c6af6a90794d6cb06d19db09afd15df1a5d/src/parser.rs#L101) handle extraneous tokens after `START TRANSACTION`? (i.e. only raise an error in the `if required` case.)

benesch:
Good catch, @gaffneyk, and thanks for the patch! @nickolay you're completely right that we can drop the peek entirely and let the common code handle the error. I've pushed an update to this branch with that approach. I'll land when CI goes green.

gaffneyk:
No problem, @nickolay and @benesch, I'm happy to contribute. Thanks for taking care of that!

benesch:
Very much appreciate you taking the time to put together a PR, @gaffneyk! Even though the approach that’s landing is slightly different, it’s quite motivating when someone takes the time to put together a working fix.

Please don’t hesitate to contribute in the future! :)

PR #143, opened 2019-10-18
Support quoted identifiers in AST as well

### LINKED ISSUE #140: Parser mishandles quoted column names
The tokenizer nicely handles quoted column names, independent of dialect, e.g., `[column name]`, `"column name"`, and `` `column name` `` are all correctly recognized and the unquoted identifier `column name` is stored in the `value` slot of the corresponding `Word`. But then the parser's [`as_ident()`](https://github.com/andygrove/sqlparser-rs/blob/c7021ed145c59f9684bac6a14d63ddb66ec4aa17/src/parser.rs#L2069) restores the quotes by calling `to_string()` instead of just returning the value of `self.value`. I can't speak to all 7 lines that invoke this method but this seems the wrong implementation at least when parsing [column names](https://github.com/andygrove/sqlparser-rs/blob/c7021ed145c59f9684bac6a14d63ddb66ec4aa17/src/parser.rs#L979) as well as [parsing identifiers](https://github.com/andygrove/sqlparser-rs/blob/c7021ed145c59f9684bac6a14d63ddb66ec4aa17/src/parser.rs#L1369). If there are no objections, I'm gonna submit a pull request to fix this.

apparebit:
As described in #140, the tokenizer already handles quoted identifiers but the current AST drops that information onto the floor again. This pull request adds the same kind of quoted identifier support to the AST, replacing the `Ident` type alias for `String` with a struct that closely resembles `Word`. The changes are relatively small, with most lines uplifting the tests.

This pull request closes #140. I am out of office next week (i.e., the one starting Monday 10/21). So please feel free to merge (or to make minor changes).

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -68,8 +68,61 @@ where
     DisplaySeparated { slice, sep: ", " }
 }
 
-/// Identifier name, in the originally quoted form (e.g. `"id"`)
-pub type Ident = String;
+/// An identifier, decomposed into its value or character data and the quote style.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Ident {
+    /// The value of the identifier without quotes.
+    pub value: String,
+    /// The starting quote if any. Valid quote characters are the single quote,
+    /// double quote, backtick, and opening square bracket.
+    pub quote_style: Option<char>,
+}
+
+impl Ident {
+    /// Create a new identifier with the given value and no quotes.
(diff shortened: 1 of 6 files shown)


benesch:
Thanks, Robert, this looks great! No review feedback whatsoever. I'm `cargo fmt`ing, squashing, and merging.

apparebit:
Good to hear that it was acceptable. And thank you for doing the cleanup. I have CLion configured to rustfmt on Save but it appears to be unreliable. Sorry about that.

Issue #149, opened 2020-01-02
SELECT TOP N syntax (i/o LIMIT) for MSSQL


alexkyllo:
This is minor but seemed like potentially a good first issue, and the README mentioned you were interested in PRs to add support for RDBMS like MS SQL Server:

SQL Server syntax for limiting the result set is `SELECT TOP N ...` or `SELECT TOP(N) ...` optionally followed by `PERCENT` and/or `WITH TIES` -- see [Microsoft Docs link](https://docs.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-ver15). I don't see `TOP` in the keyword list and was able to create a failing test for parsing this. Thinking of submitting a PR to add support for this syntax.

benesch:
Please do! That sounds like a great first issue to tackle.

PR #150, opened 2020-01-04
Add support for MSSQL's SELECT TOP N syntax

alexkyllo:
Microsoft SQL Server does not use LIMIT syntax to limit the query result set. Instead, the syntax is `SELECT TOP (numeric expression) [PERCENT] [WITH TIES]`
See: https://docs.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql

This is for Issue #149

This adds a `Top` child struct to the `Select` struct. It may or may not be problematic that `Select` would now have `Top` nested in it, while the outer `Query` struct has a `limit` field, when `TOP` and `LIMIT` have basically the same meaning in different SQL dialects. But since `TOP` is part of the `SELECT` clause in MSSQL's grammar, this seemed to be the most straightforward way to achieve support for it. Open to suggestions!

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -27,7 +27,7 @@ pub use self::ddl::{
 pub use self::operator::{BinaryOperator, UnaryOperator};
 pub use self::query::{
     Cte, Fetch, Join, JoinConstraint, JoinOperator, OrderByExpr, Query, Select, SelectItem,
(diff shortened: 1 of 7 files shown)


alexkyllo:
I didn't get the `clippy::double-must-use` error when I ran `cargo clippy` locally. Trying to reproduce...

benesch:
Thanks very much for this! Very appreciated.

Issue #154, opened 2020-03-31
Fail to parse - "select * from (table_name)"


eyalleshem:
I think it's fail here - https://github.com/andygrove/sqlparser-rs/blob/master/src/parser.rs#L1797 
because the code assume that inside the Parenthesis should be TableFactor::Derived  or TableFactor::NestedJoin,  but it's not take in account it's could be TableFactor::Table

nickolay:
`from (table_name)` is not actually valid SQL. Note that `<table primary>` in the [ANSI SQL grammar](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#_7_5_from_clause) does not allow parentheses around `<table or query name>`. A quick test shows that at least MS SQL Server follows the spec here. What database are you working with?

eyalleshem:
Thanks for the quick response :)  

I testest against snowflake ..
Is that something that could be solve by building a dialect ?

nickolay:
Theoretically, yes. Practically in the current implementation the selected dialect only affects the tokenizer, and the parser unconditionally tries to parse some non-standard constructs.

eyalleshem:
So a code that support such syntax could be accepted?

nickolay:
I guess so, if it's going to be a simple change, that doesn't introduce problems parsing the standard syntax (this specific bit of the parser is quite complex as it is). @benesch what do you think?

eyalleshem:
well - i tried :)

nickolay:
PR #155 is now merged, closing this. Thanks!

PR #155, opened 2020-03-31
Support "Select * from (a)" queries (when the table name is inside Parenthesis)

eyalleshem:
Following the discussion in issue #154  , 
I tried to add support to parenthesis around table name, 
waiting for reviews :)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1771,6 +1771,7 @@ impl Parser {
             //                   ^ ^ ^ ^
             //                   | | | |
             //                   | | | |
+            //                   | | | |
             //                   | | | (4) belongs to a SetExpr::Query inside the subquery
             //                   | | (3) starts a derived table (subquery)
             //                   | (2) starts a nested join
@@ -1793,18 +1794,12 @@ impl Parser {
                     // Ignore the error and back up to where we were before.
                     // Either we'll be able to parse a valid nested join, or
                     // we won't, and we'll return that error instead.
+                    //
+                    // Even the SQL spec prohibits derived tables and bare
+                    // tables from appearing alone in parentheses, we allowed it
+                    // as some Db's allowed that (snowflake as example)
                     self.index = index;
                     let table_and_joins = self.parse_table_and_joins()?;
-                    match table_and_joins.relation {
-                        TableFactor::NestedJoin { .. } => (),
-                        _ => {
-                            if table_and_joins.joins.is_empty() {
-                                // The SQL spec prohibits derived tables and bare
-                                // tables from appearing alone in parentheses.
-                                self.expected("joined table", self.peek_token())?
-                            }
-                        }
-                    }
                     self.expect_token(&Token::RParen)?;
                     Ok(TableFactor::NestedJoin(Box::new(table_and_joins)))
                 }
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -1806,11 +1806,19 @@ fn parse_join_nesting() {
         vec![join(nest!(nest!(nest!(table("b"), table("c")))))]
     );
 
-    let res = parse_sql_statements("SELECT * FROM (a NATURAL JOIN (b))");
-    assert_eq!(
-        ParserError::ParserError("Expected joined table, found: )".to_string()),
-        res.unwrap_err()
-    );
+    // Parenthesized table names are non-standard, but supported in Snowflake SQL
+    let sql = "SELECT * FROM (a NATURAL JOIN (b))";
+    let select = verified_only_select(sql);
+    let from = only(select.from);
+
+    assert_eq!(from.relation, nest!(table("a"), nest!(table("b"))));
+
+    // Double parentheses around table names are non-standard, but supported in Snowflake SQL
+    let sql = "SELECT * FROM (a NATURAL JOIN ((b)))";
+    let select = verified_only_select(sql);
+    let from = only(select.from);
+
+    assert_eq!(from.relation, nest!(table("a"), nest!(nest!(table("b")))));
 }
 
 #[test]
@@ -1953,10 +1961,24 @@ fn parse_derived_tables() {
         }))
     );
 
-    let res = parse_sql_statements("SELECT * FROM ((SELECT 1) AS t)");
+    // Nesting a subquery in parentheses is non-standard, but supported in Snowflake SQL
+    let sql = "SELECT * FROM ((SELECT 1) AS t)";
+    let select = verified_only_select(sql);
+    let from = only(select.from);
+
     assert_eq!(
-        ParserError::ParserError("Expected joined table, found: )".to_string()),
-        res.unwrap_err()
+        from.relation,
+        TableFactor::NestedJoin(Box::new(TableWithJoins {
+            relation: TableFactor::Derived {
+                lateral: false,
+                subquery: Box::new(verified_query("SELECT 1")),
+                alias: Some(TableAlias {
+                    name: "t".into(),
+                    columns: vec![],
+                })
+            },
+            joins: Vec::new(),
+        }))
     );
 }
 


src/parser.rs
```diff
@@ -1794,47 +1829,29 @@ impl Parser {
...
+                    match self.parse_table_inside_parenthesis() {
+                        Ok(res) => Ok(res), 
+                        Err(_) => {
+                            self.index = index;
+                            let table_and_joins = self.parse_table_and_joins()?; 
+                            match table_and_joins.relation {
+                                TableFactor::NestedJoin { .. } => (),
+                                _ => {
+                                    if table_and_joins.joins.is_empty() {
+                                        // The SQL spec prohibits derived tables and bare
```
nickolay, line 1841: I expected the change to be here, as _"bare tables .. appearing alone in parentheses"_ is exactly the case you reported as supported in Snowflake SQL. Not sure about the "derived tables .. alone in parentheses" case.

eyalleshem:
Hi , 
From (foo bar) should fail ,  i also added test for that . 

I think added the bare table before the join - as I don't want the object to be "table with joins" but without join's   - i think it's will be more correct that bare table  will be just a table_factor (I also updated the comment's in that section) .

eyalleshem:
image
It's look like the CI don't failed because code issue - but because some infrastructure issue..
Any option to re-run it ?

nickolay:
> It's look like the CI don't failed because code issue - but because some infrastructure issue..
> Any option to re-run it ?

I don't know how travis permissions work, but it allows me to restart the build once I log in with github. Judging from https://rust-lang-nursery.github.io/rust-toolstate/ the nightly rustfmt we use is currently unavailable, so restarting the build until that's fixed will not help.

nickolay:
> I think added the bare table before the join - as I don't want the object to be "table with joins" but without join's - i think it's will be more correct that bare table will be just a table_factor 

The joins in `TableWithJoins` are optional by design. A more appropriate name could be `TableFactorOptionallyFollowedByJoins`, but that's too long.

The code you have now strips the extraneous parens, while it would arguably be more correct to keep them in the AST via `TableFactor::NestedJoin` -- probably renamed it to `Nested` and its comments updated https://github.com/andygrove/sqlparser-rs/blob/172ba42001003bce79708141923ba6100cc7d4f2/src/ast/query.rs#L228-L232

I believe this would be less code, less nesting, and less intrusive patch overall.

eyalleshem:
But - if i understand right you suggestion : 

"From a" - will be parsed as  : 
```
TableWithJoins  {
    relation :  TableFactor::Table {
        name :a 
      [...] 
   }
}
```

"From (a)" - will be parsed as  : 
```
TableWithJoins  {
    relation :  TableFactor::NestedJoin (
        TableWithJoins  {
              [Table factor for a]
         }
   ) 
}
```

but we don't have any nested join in the query...

nickolay:
Right. `TableFactor::NestedJoin` was called that because standard SQL only allows joins to be nested. If we want other things to be nested, it should _"probably renamed it to Nested and its comments updated," as I noted._

You may argue that an object representing the extra parentheses does NOT belong to an _abstract_ syntax tree, but in other cases we prefer preserving detail in the AST, and in this particular case IMO this should make the particularly complex piece of the parser easier to manage.

eyalleshem:
Changed according to your suggestion. 
It's also will be better - as it's will support multi level of parenthesis. 
(select *from ((a)))

src/parser.rs
```diff
@@ -1793,18 +1795,12 @@ impl Parser {
                     // Ignore the error and back up to where we were before.
                     // Either we'll be able to parse a valid nested join, or
```
nickolay, line 1796: This whole comment should be reworded (as well as the bit in ast.rs I linked to earlier), as otherwise it becomes self-contradictory. I guess I'll do it myself later.

tests/sqlparser_common.rs
```diff
@@ -1806,11 +1806,34 @@ fn parse_join_nesting() {
         vec![join(nest!(nest!(nest!(table("b"), table("c")))))]
     );
 
-    let res = parse_sql_statements("SELECT * FROM (a NATURAL JOIN (b))");
+
+    let sql = "SELECT * FROM a NATURAL JOIN (((b NATURAL JOIN c)))";
```
nickolay, line 1810: This appears to duplicate the test above, no?

tests/sqlparser_common.rs
```diff
@@ -1806,11 +1806,34 @@ fn parse_join_nesting() {
...
+    let from = only(select.from);
+    assert_eq!(from.relation, table("a"));
     assert_eq!(
-        ParserError::ParserError("Expected joined table, found: )".to_string()),
-        res.unwrap_err()
+        from.joins,
+        vec![join(nest!(nest!(nest!(table("b"), table("c")))))]
+    );
+
+    let sql = "SELECT * FROM (a NATURAL JOIN (b))";
```
nickolay, line 1819: It would be cleaner to have only the generic dialect accept this, and have a TODO test that makes the ANSI dialect fail it. But I guess that introducing the necessary infrastructure is too much for this PR.

So at least we must note that this test is supposed to succeed with the Snowflake dialect only:
```
// Parenthesized table names are non-standard, but supported in Snowflake SQL
```

tests/sqlparser_common.rs
```diff
@@ -1806,11 +1806,34 @@ fn parse_join_nesting() {
...
+    let select = verified_only_select(sql);
+    let from = only(select.from); 
+
+    assert_eq!(
+        from.relation,
+        nest!(table("a"), nest!(table("b")))
+    );
+
+    let sql = "SELECT * FROM (a NATURAL JOIN ((b)))";
```
nickolay, line 1828:
```
// Double parentheses around table names are non-standard, but supported in Snowflake SQL
```

Please check that it is.

tests/sqlparser_common.rs
```diff
@@ -1953,11 +1976,26 @@ fn parse_derived_tables() {
         }))
     );
 
-    let res = parse_sql_statements("SELECT * FROM ((SELECT 1) AS t)");
```
nickolay, line 1956:
```
// Nesting a subquery in parentheses is non-standard, but supported in Snowflake SQL
```

Please check that it is.

src/parser.rs
```diff
@@ -1767,10 +1767,12 @@ impl Parser {
...
-            //     SELECT * FROM ( ( ( (SELECT 1) UNION (SELECT 2) ) AS t1 NATURAL JOIN t2 ) )
-            //                   ^ ^ ^ ^
-            //                   | | | |
-            //                   | | | |
+            //     SELECT * FROM ( ( ( (SELECT 1) UNION (SELECT 2) ) AS t1 NATURAL JOIN (t2) ) )
+            //                   ^ ^ ^ ^                                                ^
+            //                   | | | |                                                |
+            //                   | | | | /----------------------------------------------|
+            //                   | | | | |  
+            //                   | | | | (5) some db's (snowflake) allow bare table inside parenthesis
```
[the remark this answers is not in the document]
eyalleshem: removed

src/parser.rs
eyalleshem: Ok , i didn't touched  this comment - and according to my english skills i don't sure that I could improve it :) .

tests/sqlparser_common.rs
eyalleshem: yep , my mistake - removed

tests/sqlparser_common.rs
eyalleshem: Agree that dialect  will be much cleaner , Sound like small project  :) . 
Added the comment.

tests/sqlparser_common.rs
eyalleshem: Added.

eyalleshem: and checked :)

tests/sqlparser_common.rs
eyalleshem: Added , thank for the help with the comment wording. 
(checked in snowflake..)

eyalleshem:
Hi , 
The Ci failed because some clippy warning in era that i didn't touched (examples/cli.rs:43:26), 
and i resolved all the comments  - but for some reason the PR still in state of "Change requested" - don't sure how i need to solve that..

nickolay:
>  The Ci failed because some clippy warning in era that i didn't touched (examples/cli.rs:43:26),

Thanks, I opened #159 to fix that.

> some reason the PR still in state of "Change requested" - don't sure how i need to solve that..

I think you should be able to expand that section and "Re-request review" to change its state.

That's not the reason I didn't get to it, though, sorry!

PR #157, opened 2020-04-04
Adding BigQuery dialect

eddyxu:
[WIP] This PR adds BigQuery dialect support  in `sqlparser-rs`


eddyxu:
My immediate interest is to support BigQuery's SQL UDFs. 

However, UDF definitions are very different from vendor to vendors. So i'd like to take some feedbacks regarding to how to support UDFs in this project.

src/parser.rs
```diff
@@ -1916,6 +1970,46 @@ impl Parser {
         }
     }
 
+    pub fn parse_parameter(&mut self) -> Result<Expr, ParserError> {
```
nickolay, line 1973: I'm not sure what's the point of unnamed parameters in practice and why the grammar in the SQL spec looks like they are possible. Is there an implementation that supports it? BigQuery doesn't seem to, judging from their docs..

Even if so, this doesn't seem to be the right way to handle those, as data types don't have to be exactly 1 token long. I think the logic ought to use backtracking - if we can't parse the name followed by the data type, try parsing the data type alone.

tests/sqlparser_common.rs
```diff
@@ -854,8 +854,8 @@ fn parse_cast() {
     );
 
     one_statement_parses_to(
-        "SELECT CAST(id AS decimal) FROM customer",
-        "SELECT CAST(id AS numeric) FROM customer",
+        "SELECT CAST(CELL(a) AS decimal) FROM customer",
```
nickolay, line 857: This was supposed to test the parsing of `decimal` as `numeric`. The fact that more complex expressions can be nested in CAST can be tested separately.

src/ast/mod.rs
```diff
@@ -223,6 +223,13 @@ pub enum Expr {
     /// A parenthesized subquery `(SELECT ...)`, used in expression like
     /// `SELECT (subquery) AS x` or `WHERE (subquery) = x`
     Subquery(Box<Query>),
+    /// A parameter declaration.
```
nickolay, line 226: I think calling it "parameter declaration" elsewhere (i.e. the variant name and the parser methods) too would be preferred. Otherwise we end up with parse_optional_{args/parameters}, though arguments and parameters are synonymous in my understanding.

src/ast/mod.rs
```diff
@@ -298,6 +305,20 @@ impl fmt::Display for Expr {
             }
             Expr::Exists(s) => write!(f, "EXISTS ({})", s),
             Expr::Subquery(s) => write!(f, "({})", s),
+            Expr::Parameter {
```
nickolay, line 308: Why this is a variant of `Expr`?

src/parser.rs
```diff
@@ -906,6 +926,40 @@ impl Parser {
...
+    /// Parse Create User Defined Function statement
+    ///
+    /// 1. https://jakewheat.github.io/sql-overview/sql-2011-foundation-grammar.html#schema-function
+    /// 2. https://cloud.google.com/bigquery/docs/reference/standard-sql/user-defined-functions
+    pub fn parse_create_function(&mut self, or_replace: bool) -> Result<Statement, ParserError> {
+        let if_not_exists = self.parse_keywords(vec!["IF", "NOT", "EXISTS"]);
+
+        let name = self.parse_object_name()?;
+        self.expect_token(&Token::LParen)?;
+        let args = self.parse_optional_parameters()?; // TODO: support arguments
```
nickolay, line 938: What does this TODO refer to?

src/parser.rs
```diff
@@ -906,6 +926,40 @@ impl Parser {
...
+
+        let temporary = self.parse_keyword("TEMPORARY") || self.parse_keyword("TEMP");
```
nickolay, line 940: Why is this here? The [doc you linked to](https://cloud.google.com/bigquery/docs/reference/standard-sql/user-defined-functions) has this: `CREATE [OR REPLACE] {TEMPORARY | TEMP} FUNCTION`. Would you add a test for this as well?

src/parser.rs
```diff
@@ -906,6 +926,40 @@ impl Parser {
         })
     }
 
+    /// Parse Create User Defined Function statement
```
nickolay, line 929: "Parse `CREATE FUNCTION` statement"?

src/parser.rs
eddyxu: Sure will do.

src/parser.rs
```diff
@@ -393,7 +401,15 @@ impl Parser {
     /// Parse a SQL CAST function e.g. `CAST(expr AS FLOAT)`
     pub fn parse_cast_expr(&mut self) -> Result<Expr, ParserError> {
         self.expect_token(&Token::LParen)?;
-        let expr = self.parse_expr()?;
+        // Can be an expr or a function
```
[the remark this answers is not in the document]
eddyxu: Please bear with me. This is my project to learn rust :p

src/ast/mod.rs
eddyxu: Makes sense. I will move this to a separate struct

src/ast/mod.rs
eddyxu: I will move this out, and add some tests

src/parser.rs
eddyxu: Hmm, you are right. This syntax is probably from postgres: https://www.postgresql.org/docs/9.5/sql-createfunction.html

I will just re-implement it according to BigQuery' spec then.

PR #158, opened 2020-04-14
Add support for OFFSET without the ROWS keyword

mjibson:
MySQL doesn't support the ROWS part of OFFSET. Teach the parser to
remember which variant it saw, including just ROW.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -26,8 +26,8 @@ pub use self::ddl::{
 };
 pub use self::operator::{BinaryOperator, UnaryOperator};
 pub use self::query::{
-    Cte, Fetch, Join, JoinConstraint, JoinOperator, OrderByExpr, Query, Select, SelectItem,
-    SetExpr, SetOperator, TableAlias, TableFactor, TableWithJoins, Top, Values,
+    Cte, Fetch, Join, JoinConstraint, JoinOperator, Offset, OffsetRows, OrderByExpr, Query, Select,
+    SelectItem, SetExpr, SetOperator, TableAlias, TableFactor, TableWithJoins, Top, Values,
 };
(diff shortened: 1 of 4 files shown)


mjibson:
Test failed due to a rustfmt nightly issue. Bumping to re-run.

mjibson:
Ok well this keeps failing with `error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly` which I don't think is related to this PR.

mjibson:
Ahh https://github.com/andygrove/sqlparser-rs/pull/159 seems to maybe fix it.

nickolay:
Yep, the rustfmt issue should be fixed in #159. I didn't manage to fix coverage yesterday, before I got distracted.

Sorry you didn't get a reply. I took a quick glance at the patch yesterday - and it looked great (thanks!), but I wanted to fix CI first before taking a closer look and merging.

Note to self: this quirk is documented in https://dev.mysql.com/doc/refman/8.0/en/select.html

>  For compatibility with PostgreSQL, MySQL also supports the `LIMIT row_count OFFSET offset` syntax. 

(note the lack of `ROWS` after _offset_)

mjibson:
Rebased on master.

nickolay:
Excellent (and thanks for rebasing promptly)!

Do I understand it correctly that you needed MySQL-specific syntax to round-trip properly for your use-case?

mjibson:
Luckily I am currently only producing SQL, so things are fine as is. I do have an upcoming project that would require me to parse SQL, and this is one of the first obvious things that would force me to do silly stuff like use regexes to remove or insert the keywords. I took a look at your custom parser doc because I was curious how this could be possible with the current code.

PR #162, opened 2020-04-19
derive default for GenericDialect

alex-dukhno:
The motivation behind this changes is that whatever user structure has a field of `GenericDialect` it is trivial to `impl Default` by deriving it like:
```rust
#[derive(Default)]
struct MySqlProcessor {
  dialect: GenericDialect
}
```


nickolay:
I don't see why you'd want to do that, but I suppose it doesn't hurt so I merged it, thanks!

PR #163, opened 2020-04-19
create if not exists for all dialect

alex-dukhno:
This PR attempts to implement `CREATE TABLE IF NOT EXISTS table_name` for all dialects.
However, I am not sure that it is a standard SQL statement - please correct me if I am wrong.


src/parser.rs
```diff
@@ -932,6 +933,37 @@ impl Parser {
     }
 
     pub fn parse_create_table(&mut self) -> Result<Statement, ParserError> {
+        let if_keyword = self.parse_keyword("IF");
```
nickolay, line 936: I think you can simply use `let if_not_exists = self.parse_keywords(vec!["IF", "NOT", "EXISTS"]);`, why not?

tests/sqlparser_common.rs
```diff
@@ -995,6 +996,102 @@ fn parse_create_table() {
...
+            columns,
+            constraints,
+            with_options,
+            if_not_exists: true,
+            external: false,
+            file_format: None,
+            location: None,
+        } => {
+            assert_eq!("uk_cities", name.to_string());
+            assert_eq!(
```
nickolay, line 1024: Please remove the columns copy-paste, I feel the cost of duplication outweighs the benefit of making this test more realistic.

src/parser.rs
alex-dukhno: If I use `let if_not_exists = self.parse_keywords(vec!["IF", "NOT", "EXISTS"]);` I wouldn't be able to say if user typed in just `CREATE TABLE` or missed one of keyword because `self.parse_keywords` returns `bool`

alex-dukhno:
@nickolay thank you for review, I've rebased over master and moved tests to only PostgreSQL dialect module, but I have general question how do I make sure that `CREATE TABLE IF NOT EXISTS` applies only to Postgres dialect?

src/parser.rs
nickolay: Not sure I fully understand what you're saying.. When `parse_keywords` returns `false` the current token is left unchanged. So `CREATE TABLE if (...)` would be accepted (which seems more appropriate for the generic parser at least; I don't know if Postgres rejects that), and missing some of the keywords like `CREATE TABLE IF NOT ...` would fail with an error like "Expected end of statement, found: NOT" instead of more specific error message, which seems acceptable too.

Listing all possible errors for multiple dialects doesn't seem maintainable, and we already use `parse_keywords` in other places.

nickolay:
>And please run cargo fmt

Note that CI still thinks you haven't run rustfmt before comitting: https://travis-ci.org/github/andygrove/sqlparser-rs/builds/677142460#L641 - did you?

nickolay:
Thank you!

PR #167, opened 2020-05-15
Add create index and drop index support

mashuai:
This PR adds support for:

    CREATE [ UNIQUE ] INDEX [ IF NOT EXISTS ]
        <index_name>
        ON <table_name>
        ( col_name [, ...] )

    DROP INDEX <index_name>

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -476,6 +476,15 @@ pub enum Statement {
         file_format: Option<FileFormat>,
         location: Option<String>,
     },
+    /// CREATE INDEX
+    CreateIndex {
+        /// index name
+        name: ObjectName,
+        table_name: ObjectName,
+        columns: Vec<Ident>,
+        unique: bool,
+        if_not_exists: bool,
+    },
     /// ALTER TABLE
     AlterTable {
         /// Table name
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -476,6 +476,15 @@ pub enum Statement {
         file_format: Option<FileFormat>,
         location: Option<String>,
     },
+    /// CREATE INDEX
+    CreateIndex {
+        /// index name
+        name: ObjectName,
+        table_name: ObjectName,
+        columns: Vec<ObjectName>,
+        if_unique: bool,
```
nickolay, line 485: Simply `unique` seems more appropriate.

src/dialect/keywords.rs
```diff
@@ -50,6 +50,7 @@ macro_rules! define_keywords {
 }
 
 define_keywords!(
+    INDEX,
```
nickolay, line 53: This ought to be further below.

src/parser.rs
```diff
@@ -912,8 +916,10 @@ impl Parser {
             ObjectType::Table
         } else if self.parse_keyword("VIEW") {
             ObjectType::View
+        } else if self.parse_keyword("INDEX") {
+            ObjectType::Index
         } else {
-            return self.expected("TABLE or VIEW after DROP", self.peek_token());
+            return self.expected("TABLE ,VIEW or INDEX after DROP", self.peek_token());
```
nickolay, line 922: Typo.

mashuai:
@nickolay Thanks for the comment. And yes I should use `self.parse_comma_separated(Parser::parse_identifier)?` to parse column names.
 Actually I don't intend to support trailing comma, just because I find `parse_columns` support it, so I do the same thing.

nickolay:
Alright, supporting `CREATE INDEX idx_name ON test (col_name[, ...])` sounds great to me!

In this case, there's no reason to have `parse_index_column` as a separate method (it's rather short now!), this would lead to changing `let object_names` into `let columns`, as expected by `Statement::CreateIndex`. If you can make this last change and squash the changes into a single commit rebased on top of the current master, I'll merge it now. Or I can take it from here, if you prefer.

mashuai:
Hi @nickolay, thanks for the advice, I have moved `parse_index_column` into `parse_create_index` and squash the changes into a single commit rebased on top of the current master.

nickolay:
Excellent, thanks so much for bearing with me and for taking it to the finish line!

I just realized this can now use `parse_parenthesized_column_list`, but as I wanted to credit you in the CHANGELOG anyway, I've gone ahead and made the last minute change myself.

PR #173, opened 2020-05-28
added create and drop schema

alex-dukhno:
This PR adds functionality to parse `CREATE SCHEMA schema_name` and `DROP SCHEMA schema_name`

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -864,9 +864,19 @@ impl Parser {
             self.parse_create_view()
         } else if self.parse_keyword("EXTERNAL") {
             self.parse_create_external_table()
+        } else if self.parse_keyword("SCHEMA") {
+            self.parse_create_schema()
         } else {
-            self.expected("TABLE, VIEW or INDEX after CREATE", self.peek_token())
+            self.expected(
+                "TABLE, VIEW, INDEX or SCHEMA after CREATE",
+                self.peek_token(),
+            )
         }
+    }
+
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -864,11 +864,21 @@ impl Parser {
             self.parse_create_view()
         } else if self.parse_keyword("EXTERNAL") {
             self.parse_create_external_table()
+        } else if self.parse_keyword("SCHEMA") {
+            self.parse_create_schema()
         } else {
-            self.expected("TABLE, VIEW or INDEX after CREATE", self.peek_token())
+            self.expected(
+                "TABLE, VIEW, INDEX or SCHEMA after CREATE",
```
nickolay, line 871: I think with the next addition it makes sense to change the message, so that it doesn't duplicate the list above (as it's incorrect already, missing "EXTERNAL").

alex-dukhno:
> Both SQL:2016 and Posgres have a number of options for the CREATE SCHEMA statement. They also support including other CREATE statements as part of the schema definition, rather than separate statements.
I think it will be gradually added. From my side, I am contributing to parser those parts that I need to right now to develop Postgres compatible database 😄 So whenever I need support of including other statements I will come back to this repo 😃

nickolay:
Yep, I agree that implementing bits and pieces as you need them makes the most sense. Thanks for sharing them!

I'm just making a note of the the relevant docs and the bits that I think are helpful for maintaining and maybe re-using the parser.

PR #174, opened 2020-05-29
provide LISTAGG implementation

### LINKED ISSUE #169: `listagg` support?
It seems the parser doesn't currently support [listagg](https://modern-sql.com/feature/listagg). For instance, the following Redshift example fails to parse:

```sql
select listagg(sellerid) 
within group (order by dateid) as sellers,
listagg(dateid) as dates
from winsales;
```

maxcountryman:
This patch provides an initial implemenation of LISTAGG[1]. Notably this
implemenation deviates from ANSI SQL by allowing both WITHIN GROUP and
the delimiter to be optional. We do so because Redshift SQL works this
way and this approach is ultimately more flexible.

Fixes #169.

[1] https://modern-sql.com/feature/listagg


src/parser.rs
```diff
@@ -423,6 +424,74 @@ impl Parser {
...
+            Some(Box::new(separtor.clone()))
+        } else {
+            None
+        };
+        let on_overflow = if self.parse_keywords(vec!["ON", "OVERFLOW"]) {
+            let error = self.parse_keyword("ERROR");
+            let filler = if !error {
+                self.expect_keyword("TRUNCATE")?;
+                Some(Box::new(self.parse_expr()?))
+            } else {
```
nickolay, line 450: This should become more natural when `ListAggOnOverflow` is a enum, as you won't have to initialize irrelevant properties to `None` in the `!error` case.

src/parser.rs
```diff
@@ -423,6 +424,74 @@ impl Parser {
...
+        // Once again ANSI SQL requires WITHIN GROUP, but Redshift does not. Again we choose the
+        // more general implementation.
+        let within_group = if self.parse_keywords(vec!["WITHIN", "GROUP"]) {
+            self.expect_token(&Token::LParen)?;
+            self.expect_keywords(&["ORDER", "BY"])?;
+            Some(self.parse_comma_separated(Parser::parse_order_by_expr)?)
+        } else {
+            None
+        };
+        if within_group.is_some() {
```
nickolay, line 483: Please combine with the above `if`, like this: https://github.com/andygrove/sqlparser-rs/blob/c918ff042df521fdde0e780947a5d33745708a51/src/parser.rs#L1984-L1986

src/ast/mod.rs
```diff
@@ -812,6 +815,71 @@ impl FromStr for FileFormat {
...
+        let distinct = if self.distinct { "DISTINCT " } else { "ALL " };
+        let args = if let Some(separator) = &self.separator {
+            format!(
+                "{}",
+                display_comma_separated(&[self.expr.clone(), separator.clone()])
+            )
+        } else {
+            format!("{}", self.expr)
+        };
+        if let Some(on_overflow) = &self.on_overflow {
```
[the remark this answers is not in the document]
maxcountryman: It's certainly possible, but it seemed a little cleaner to define `args` one time and then format `LISTAGG(...)`. But I can rework that if you prefer.

src/parser.rs
```diff
@@ -423,6 +424,74 @@ impl Parser {
         })
     }
 
+    /// Parse a SQL LISTAGG expression, e.g. `LISTAGG(...) WITHIN GROUP (ORDER BY ...)`.
+    pub fn parse_listagg_expr(&mut self) -> Result<Expr, ParserError> {
+        self.expect_token(&Token::LParen)?;
+        let all = self.parse_keyword("ALL");
```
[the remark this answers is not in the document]
maxcountryman: I've gone ahead and done this but my first pass of this generalizes the error message in a way that removes preexisting context. I completely understand if this isn't desirable and would be happy to adapt it so that information isn't lost. On the other hand, I wasn't necessarily sure it was needed. So, I'll defer to you and then update accordingly.

maxcountryman:
@nickolay thank you for the thorough and detailed feedback. It's most appreciated! I've made a number of updates and left a couple of things open just to clarify or get your feedback.

src/parser.rs
```diff
@@ -851,6 +900,18 @@ impl Parser {
         Ok(values)
     }
 
+    /// Parse either `ALL` or `DISTINCT`. Returns `true` if `DISTINCT` is parsed and results in a
+    /// `ParserError` if both `ALL` and `DISTINCT` are fround.
+    pub fn parse_all_or_distinct(&mut self) -> Result<bool, ParserError> {
```
maxcountryman, line 905: I wasn't sure exactly where to place this so I kind of guessed. If there's a better spot, please let me know.

src/parser.rs
```diff
@@ -851,6 +900,18 @@ impl Parser {
...
+        let all = self.parse_keyword("ALL");
+        let distinct = self.parse_keyword("DISTINCT");
+        if all && distinct {
+            return parser_err!("Cannot specify both ALL and DISTINCT".to_string());
```
maxcountryman, line 909: I mentioned this elsewhere, but this does reduce the info contained in the error messages by removing some context we previously had--I really don't know enough to say how important this is so please let me know what the preference is.

src/ast/mod.rs
```diff
@@ -812,6 +815,71 @@ impl FromStr for FileFormat {
...
+    pub within_group: Option<Vec<OrderByExpr>>,
+}
+
+impl fmt::Display for ListAgg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let distinct = if self.distinct { "DISTINCT " } else { "ALL " };
+        let args = if let Some(separator) = &self.separator {
+            format!(
+                "{}",
+                display_comma_separated(&[self.expr.clone(), separator.clone()])
```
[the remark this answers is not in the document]
nickolay: It's less complex: both contenders already use `format!()`, but what's expressed as an extra 4 characters `", {}"` in one case, requires an additional `display_comma_separated()`call, two `.clone()` calls, `&[]` slice construction, which rustfmt spreads across 4 lines...

`display_comma_separated()` was introduced to avoid boilerplate working with sequences of unrestricted length, and we don't use for a fixed-length sequences elsewhere.

src/ast/mod.rs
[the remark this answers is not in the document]
nickolay: I don't think it's cleaner, as there's more duplication this way. And there's an argument for  consistency: I think most if not all of formatting code follows the approach I've linked to.

src/parser.rs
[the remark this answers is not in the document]
nickolay: I wouldn't worry too much about the error message. Good error messages should provide location information, not explain it with words.

src/parser.rs
nickolay: There's no well-defined order of the members; this spot makes sense to me!

src/parser.rs
nickolay: As noted above, I think this is fine.

src/ast/mod.rs
```diff
@@ -857,28 +857,32 @@ impl fmt::Display for ListAgg {
 
 /// The `ON OVERFLOW` clause of a LISTAGG invocation
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct ListAggOnOverflow {
-    pub error: bool,
-    pub filler: Option<Box<Expr>>,
-    pub with_count: bool,
+pub enum ListAggOnOverflow {
+    Error,
```
nickolay, line 861: Please add doc-comments for these variants as well: `ON OVERFLOW ERROR` / `TRUNCATE [ <filler> ] WITH[OUT] COUNT`

src/ast/mod.rs
```diff
@@ -857,28 +857,32 @@ impl fmt::Display for ListAgg {
...
-                    format!("{} ", filler)
+        write!(f, " ON OVERFLOW")?;
+        match self {
+            ListAggOnOverflow::Error => write!(f, " ERROR"),
+            ListAggOnOverflow::Truncate { filler, with_count } => {
+                write!(f, " TRUNCATE")?;
+                if let Some(filler) = filler {
+                    write!(f, " {}", filler)?;
+                }
+                if *with_count {
```
nickolay, line 878: Perhaps I value vertical space too much, but I'd write this as: `write!(f, " {} COUNT", if *with_count { "WITH" } else { "WITHOUT" })`. Leaving it as is is fine too.

src/parser.rs
```diff
@@ -444,28 +444,24 @@ impl Parser {
         };
         let on_overflow = if self.parse_keywords(vec!["ON", "OVERFLOW"]) {
             let error = self.parse_keyword("ERROR");
```
nickolay, line 431: You don't need the variable anymore. Simply: `if self.parse_keyword("ERROR") {`.

src/parser.rs
```diff
@@ -444,28 +444,24 @@ impl Parser {
...
-            let filler = if !error {
-                self.expect_keyword("TRUNCATE")?;
-                Some(Box::new(self.parse_expr()?))
+            Some(if error {
```
nickolay, line 432: We avoid wrapping long expressions in Some. Please use `Some(ListAggOnOverflow::Error)` instead.

src/parser.rs
```diff
@@ -444,28 +444,24 @@ impl Parser {
...
-            };
-            let with_count = if !error {
-                let with_count = self.parse_keywords(vec!["WITH", "COUNT"]);
-                let without_count = self.parse_keywords(vec!["WITHOUT", "COUNT"]);
-                if !with_count && !without_count {
-                    return parser_err!(
-                        "Expected either WITH COUNT or WITHOUT COUNT in LISTAGG".to_string()
-                    );
+                self.expect_keyword("TRUNCATE")?;
+                let filler = Some(Box::new(self.parse_expr()?));
```
nickolay, line 436: As implemented, the `filler` isn't optional. I'm fine with this if we change the spec and the AST, or you could peek_token() to see if the next token is either a `WITH` or a `WITHOUT` keyword before attempting to `parse_expr`

src/parser.rs
```diff
@@ -444,28 +444,24 @@ impl Parser {
...
+                let with_count = if !error {
```
nickolay, line 437: `!error` always holds here.

src/parser.rs
```diff
@@ -444,28 +444,24 @@ impl Parser {
...
+                    let with_count = self.parse_keywords(vec!["WITH", "COUNT"]);
+                    let without_count = self.parse_keywords(vec!["WITHOUT", "COUNT"]);
+                    if !with_count && !without_count {
+                        return parser_err!(
+                            "Expected either WITH COUNT or WITHOUT COUNT in LISTAGG".to_string()
+                        );
+                    };
+                    with_count
```
nickolay, line 445: I think this is slightly shorter:
```
let with_count = self.parse_keyword("WITH");
if !with_count && !self.parse_keyword("WITHOUT") {
    self.expected("either WITH or WITHOUT in LISTAGG", self.peek_token())?;
}
self.expect_keyword("COUNT")?;
```

maxcountryman:
@nickolay I think I've addressed your additional comments. Thank you for your patience, I appreciate it: I'm still very much a Rust novice. 🙂

src/ast/mod.rs
maxcountryman: I left it for now, but happy to clean it up later.

src/ast/mod.rs
maxcountryman: Oops, I misread this. I'll add that now.

src/parser.rs
```diff
@@ -423,6 +417,61 @@ impl Parser {
...
+                    Some(Token::Word(kw)) if kw.keyword == "WITH" || kw.keyword == "WITHOUT" => {
+                        None
+                    }
+                    Some(Token::SingleQuotedString(_)) => Some(Box::new(self.parse_expr()?)),
+                    _ => self.expected("either filler, WITH, or WITHOUT", self.peek_token())?,
+                };
+                let with_count = self.parse_keyword("WITH");
+                if !with_count {
+                    self.expect_keyword("WITHOUT")?;
+                }
```
maxcountryman, line 447: I reworked this a bit since now the check for `WITH` or `WITHOUT` should be happening on line 442.

src/parser.rs
```diff
@@ -435,21 +435,15 @@ impl Parser {
...
-                        {
-                            None
-                        } else {
-                            Some(Box::new(self.parse_expr()?))
-                        }
+                    Some(Token::Word(kw)) if kw.keyword == "WITH" || kw.keyword == "WITHOUT" => {
+                        None
                     }
-                    None => None,
+                    Some(Token::SingleQuotedString(_)) => Some(Box::new(self.parse_expr()?)),
```
nickolay, line 441: If you want to only support literal string filler, use `parse_literal_string()` and make it a `filler: Option<String>`.

maxcountryman: Ah thanks. That seems [to be to spec](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#listagg-truncation-filler), right?

nickolay: I doubt this is a good idea though, as I bet someone will go and ask us to at least extend support to `N'...'` literals and then Postgres will accept expressions here...

maxcountryman: Hm, it does seem like we should be a little more concrete here: not every expression is valid in this position.

Also for context, I don't think Postgres supports LISTAGG altogether (or maybe I'm just failing to find it).

src/parser.rs
```diff
@@ -423,6 +417,66 @@ impl Parser {
...
+                Some(ListAggOnOverflow::Error)
+            } else {
+                self.expect_keyword("TRUNCATE")?;
+                let filler = match self.peek_token() {
+                    Some(Token::Word(kw)) if kw.keyword == "WITH" || kw.keyword == "WITHOUT" => {
+                        None
+                    }
+                    Some(Token::SingleQuotedString(_))
+                    | Some(Token::NationalStringLiteral(_))
+                    | Some(Token::HexStringLiteral(_)) => Some(Box::new(self.parse_expr()?)),
```
maxcountryman, line 443: This allows for any supported string literal to be parsed as a filler and uses `parse_expr` to handle this ergonomically. (There may be a more precise way of handling this, but I wasn't immediately able to find it.)

Also it might be a good idea to be stricter with the `separator` parsing, but I'll defer to your feedback.

src/parser.rs
nickolay: > not every expression is valid in this position

The same can be said about 'delimiter'. Omitting which is also not "valid", yet you yourself wrote "we choose to make the separator optional as this provides the more general implementation". The same logic can be applied here: "we choose to allow any expression not starting with `WITH` or `WITHOUT` as a `<filler>` as this provides the more general implementation".

I [know](https://github.com/andygrove/sqlparser-rs/issues/169#issuecomment-634108887) Postgres doesn't support LISTAGG at the moment. I was making a prediction that someone will deviate from the spec by supporting something other than `'string'` literals here. I brought up Postgres, as I have the impression they like to make things more general whenever possible...

I see you decided to keep it an `Expr` in the AST, but restrict the allowed tokens to certain kinds of literals (`''`, `N''`, `X''`). OK, let's do it this way.

src/parser.rs
nickolay: The WITH/WITHOUT lookahead you're referring to should not affect this code, as this can execute after the lookahead failed and we've parsed the 'filler'.

This code is equivalent to the previous one in the happy case, but will report a slightly confusing error of "Expected keyword WITHOUT, found FOO"  when parsing `'filler' FOO`.

nickolay:
The Travis failure is not a problem with your code. Please let me know what you think about https://github.com/andygrove/sqlparser-rs/pull/174#discussion_r432849585 while I'm trying to figure out what's wrong with it.

src/parser.rs
maxcountryman: You're right that the same logic could be applied here, re: the more general implementation. What occurred to me however is this question: What possible values actually make sense here? In other words, what non-string filler would work in practice? I couldn't really think of any, but I very well may be overlooking the obvious. 🙂

I'm more than happy to do what you think is best, so if you decide you'd prefer another implementation then I will swap that in for what we have here.

src/parser.rs
maxcountryman: Ah shoot, good catch! I'm not sure why I keep running into this.

nickolay:
Ah, travis runs on the results of a merge between the PR and the upstream master. Would you like to do the merge and add the two `nulls_first` in the `parse_listagg` testcase? Or I can do it when merging.

maxcountryman:
> Ah, travis runs on the results of a merge between the PR and the upstream master. Would you like to do the merge and add the two `nulls_first` in the `parse_listagg` testcase? Or I can do it when merging.

I'll try to rebase once you feel confident the implementation is acceptable.

maxcountryman:
All right, it seems like the rebase was clean. I'll force push the rebased branch when you give the go ahead. 🙂

src/parser.rs
nickolay: Your argument makes perfect sense, but in practice "LISTAGG(ALL" shouldn't exist, yet in SQL it does ;p

`'.'+'.'+'.'` is a _string_ filler just not a literal string.

But as I said, the way you did it is fine by me. If/when someone comes around wanting to lift this restriction, it would be a trivial change, rather than if you followed my original idea of changing `Expr` to `String` in the AST.

nickolay:
Feel free to either rebase and force-push or commit a merge with upstream master - I recently learned that Github's "Squash and merge" is pretty useful.

maxcountryman:
> Feel free to either rebase and force-push or commit a merge with upstream master - I recently learned that Github's "Squash and merge" is pretty useful.

I love Squash and Merge. 🙂

All right, rebased off master, force pushed to my branch. Hopefully should be passing on Travis shortly.

nickolay:
Sweet.

Thank you!

PR #176, opened 2020-05-30
add nulls first/last support to order by expression

houqp:
this is needed for adding sort support to datafusion's sql planner


nickolay:
Thanks!

PR #181, opened 2020-06-03
Support bitwise and, or, xor

Dandandan:
Supporting bitwise and, or, xor.

Operator precedence is coming from:

https://cloud.google.com/bigquery/docs/reference/standard-sql/operators


nickolay:
Thanks!

PR #182, opened 2020-06-03
Accept &str in `Parse::parse_sql`

Dandandan:
It is more generic to accept a `&str` than a `String` in an API, and avoids having to convert a the string to a `String` when not needed, avoiding a copy.


nickolay:
Good catch! Thanks.

PR #190, opened 2020-06-07
Add benchmarks using cargo bench / criterion

Dandandan:
Adds two simple benchmarks, using `cargo bench` and `criterion`.

On my machine:

```
sqlparser-rs parsing benchmark/sqlparser::select                        
                        time:   [13.409 us 13.797 us 14.171 us]

sqlparser-rs parsing benchmark/sqlparser::with_select                        
                        time:   [69.353 us 71.665 us 74.025 us]
```


nickolay:
Thanks! As I feared, this slows down the clean build (e.g. `cargo test`) considerably - 6 wall-clock seconds / 23 user CPU vs 23/120 seconds - a 4x slowdown, this is too large a price for something rarely used.

https://github.com/rust-lang/cargo/issues/1596#issuecomment-623009259 describes a possible solution, would you be willing to do that? Alternatively we could keep this on a branch of its own and include instructions on cherry-picking it in the README on `master`.

Dandandan:
That would be a nice solution. I'll be looking into that!

Dandandan:
I moved them to `sqlparser_bench`

nickolay:
Awesome, thank you!

sqlparser_bench/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(test)]
```
maxcountryman, line 1: Was this intended to be committed?

Dandandan: Good catch, is the result of running `cargo new --lib`, didn't notice that.

nickolay: Thanks @maxcountryman! I removed this file from master.

PR #191, opened 2020-06-07
Use binary search to speed up matching keywords

Dandandan:
Speeds up parsing benchmarks by ~10%


nickolay:
I like this, but because it forces people to keep the list sorted, and not because of the performance. Most of the keywords in that list are unused, and could simply be removed, if performance was an issue.

PR #195, opened 2020-06-10
Use Token::EOF instead of Option<Token>

nickolay:
This simplifies codes slightly, removing the need deal with the EOF case explicitly.

@Dandandan what do you think?

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -98,7 +98,7 @@ impl Parser {
                 expecting_statement_delimiter = false;
             }
 
-            if parser.peek_token().is_none() {
+            if parser.peek_token() == Token::EOF {
                 break;
             } else if expecting_statement_delimiter {
                 return parser.expected("end of statement", parser.peek_token());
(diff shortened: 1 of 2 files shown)


Dandandan:
I like two thing about this change:

* Makes the encoding of EOF more explicit, rather than relying on Option
* Reduces the amount of nesting in matches / special casing needed EOF vs normal token

src/parser.rs
```diff
@@ -98,7 +98,7 @@ impl Parser {
                 expecting_statement_delimiter = false;
             }
 
-            if parser.peek_token().is_none() {
+            if parser.peek_token() == Token::EOF {
                 break;
             } else if expecting_statement_delimiter {
```
Dandandan, line 103: no else needed here?

src/parser.rs
```diff
@@ -699,59 +680,53 @@ impl Parser {
...
+    pub fn peek_token(&self) -> Token {
         self.peek_nth_token(0)
     }
 
     /// Return nth non-whitespace token that has not yet been processed
-    pub fn peek_nth_token(&self, mut n: usize) -> Option<Token> {
+    pub fn peek_nth_token(&self, mut n: usize) -> Token {
         let mut index = self.index;
         loop {
             index += 1;
```
Dandandan, line 724: This index += 1 I think would be better at the and of the `loop` so it can be just `self.tokens.get(index)`, or convert to a normal for loop?
(just looking at the code surrounding the changes)

nickolay: There's a `continue` later on, so moving it below won't work. If you have suggestions for improvements, feel free to raise separate PRs.

src/parser.rs
nickolay: I don't have an opinion on this, fixed.

PR #196, opened 2020-06-10
Add serde support to AST structs and enums

panarch:
Apply serde to AST structs and enums to be serializable/deserializable.
serde support is optional, can be activated by feature named "serde".

It resolves the issue https://github.com/andygrove/sqlparser-rs/issues/183

I applied to all enums and structs which has `#[derive(Debug, Clone, PartialEq, Eq, Hash)]` in ast directory.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ name = "sqlparser"
 path = "src/lib.rs"
 
 [dependencies]
-bigdecimal = { version = "0.1.0", optional = true }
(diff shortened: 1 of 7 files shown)


nickolay:
Excellent, thanks! I think it might be a good idea to add a note about this in README and/or update the cli example to output this instead of `Debug` representation when available, but this can be done separately.

panarch:
@nickolay wow, really fast, thanks!
It also sounds good to add code example, I'll make a new PR for that.

PR #197, opened 2020-06-10
Add code example to see how to use serde

panarch:
Related to PR: https://github.com/andygrove/sqlparser-rs/pull/196

Because `serde` is an optional dependency, if I modify existing example codes, then it may make other new comers feel bothersome.
So I added new code example named `examples/serde.rs`.
@nickolay How do you think of this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,12 +18,19 @@ edition = "2018"
 name = "sqlparser"
 path = "src/lib.rs"
 
+[features]
+# Enable JSON output in the `cli` example:
+json_example = ["serde_json", "serde"]
+
 [dependencies]
 bigdecimal = { version = "0.1.0", features = ["serde"], optional = true }
 log = "0.4.5"
 serde = { version = "1.0", features = ["derive"], optional = true }
+# serde_json is only used in examples/cli, but we have to put it outside
+# of dev-dependencies because of
+# https://github.com/rust-lang/cargo/issues/1596
+serde_json = { version = "1.0", optional = true }
 
 [dev-dependencies]
-serde_json = "1.0"
 simple_logger = "1.0.1"
 matches = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,5 +24,6 @@ log = "0.4.5"
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


nickolay:
I forgot that doing that would build serde unconditionally in dev mode (as cargo doesn't allow optional dev-dependencies). I try to minimize the amount of stuff we build by default to keep the build times down. Which means we'd have to move the `cli` example to a separate crate like we did for the benchmarks in https://github.com/andygrove/sqlparser-rs/pull/190

Or move serde_json to `[dependencies]` just to make it optional. I did the latter in the follow-up push: https://github.com/andygrove/sqlparser-rs/pull/197/commits/3a05b32e4d80e6a483fcb797fcebae0575b1933c - what do you think of this?

panarch:
Wow, that both two options sound very reasonable, it's quite tough to pick only one.
I didn't think about whether adding `optional` to dev-dependencies are possible or not.
If that is possible, then that could be simple and clean solution for now. But it does not work so...

Ok, than how about taking your adding `serde_example` feature option?

If `cli` has more roles and `examples/` has more files to see, than it would be worth enough to split into different crates.
But current `cli` usage example is simple enough to embed in main `README.md` and current main `Cargo.toml` is also simple for now. So although it requires some comments like this https://github.com/andygrove/sqlparser-rs/commit/3a05b32e4d80e6a483fcb797fcebae0575b1933c#diff-80398c5faae3c069e4e6aa2ed11b28c0R29-R31, I think it's clear enough to see what's going on.

By the way, https://github.com/andygrove/sqlparser-rs/pull/190 was also so good idea to take, thanks for sharing, it was so fun to see what was going on there.

nickolay:
> If cli has more roles and examples/ has more files to see, than it would be worth enough to split into different crates.

This makes sense. So I merged the single-crate variant for now. Thanks for your help!

panarch:
https://github.com/andygrove/sqlparser-rs/commit/a0f076acda6bf0a849bf9070700ae7728c999f77#diff-6d3ae06537fccb04d134faab9e782c45R67-R70
```rust
if cfg!(feature = "json_example") {
    #[cfg(feature = "json_example")]
    {
        ...
    }
...
```
by the way, cool, this looks better than put `#[cfg(..)]` in else statement. thanks :)

PR #199, opened 2020-06-11
Take slice as input for parse_keywords

Dandandan:
Just a purely mechanical change to change `parse_keywords` take a slice instead of a `Vec`.


nickolay:
Nice! I think the point is that a slice is a static constant whereas `vec![..]` builds the value dynamically?

Dandandan:
Yes, that's the point. This is a bit more "rusty".

You can also pass a vec to this function, but not a slice to a function expecting a get.

PR #200, opened 2020-06-12
Make FileFormat case insensitive

Dandandan:
Fixes

https://github.com/andygrove/sqlparser-rs/issues/198


nickolay:
I think we shouldn't use `parse_identifier` for keywords (like in https://github.com/andygrove/sqlparser-rs/pull/193#discussion_r438399675 ). If the idea is to  support `STORED AS "TEXTFILE"`, we should store it as an identifier in the AST. Otherwise, why not add the keywords to the global list?

Dandandan:
I agree it makes sense to also parse it as an keyword and fix it in a more "general' way

nickolay:
OK, I've merged this PR to fix the bug, leaving the general way as a possible future improvement.

PR #201, opened 2020-06-12
Replace FromStr with normal parser function for FileFormat

Dandandan:
As discussed in https://github.com/andygrove/sqlparser-rs/pull/200


nickolay:
Thanks!

PR #202, opened 2020-06-13
Output DataType capitalized

Dandandan:
This makes it consistent with other output which also prints keywords capitalized.


nickolay:
I noticed that too, thanks for taking the time to fix it!

PR #203, opened 2020-06-16
implement alter table for sqlite grammar

mashuai:
implement alter table for SQLite grammar:https://www.sqlite.org/lang_altertable.html

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -23,56 +23,34 @@ use std::fmt;
 pub enum AlterTableOperation {
     /// `ADD <table_constraint>`
     AddConstraint(TableConstraint),
-    AddColumn {
-        has_column_identifier: bool,
-        column_def: ColumnDef,
-    },
+    /// `ADD [ COLUMN ] <column-def>`
+    AddColumn { column_def: ColumnDef },
     /// TODO: implement `DROP CONSTRAINT <name>`
-    DropConstraint {
-        name: Ident,
-    },
+    DropConstraint { name: Ident },
+    /// `RENAME [ COLUMN ] <column-name> TO <column-name>`
     RenameColumn {
-        has_column_identifier: bool,
         old_column_name: Ident,
         new_column_name: Ident,
     },
-    RenameTable {
-        table_name: Ident,
-    },
+    /// `RENAME TO <table-name>`
+    RenameTable { table_name: Ident },
 }
 
 impl fmt::Display for AlterTableOperation {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             AlterTableOperation::AddConstraint(c) => write!(f, "ADD {}", c),
-            AlterTableOperation::AddColumn {
-                has_column_identifier,
-                column_def,
-            } => write!(
-                f,
-                "ADD {}{}",
-                if *has_column_identifier {
-                    "COLUMN "
-                } else {
-                    ""
-                },
-                column_def.to_string()
-            ),
+            AlterTableOperation::AddColumn { column_def } => {
+                write!(f, "ADD COLUMN {}", column_def.to_string())
+            }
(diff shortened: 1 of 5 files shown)


nickolay:
Hey @mz, thanks for the PR! Please let me know when this is ready for review, as I see you're still working on it.

mashuai:
Hi @nickolay I'm working on the test coverage, I have already covered alter table test, but it still says that coverage decreased, can you help me to see what's wrong.

nickolay:
The coverage is somewhat flaky unfortunately. Given that in this case it thinks that an existing line lost coverage, whereas it wasn't covered in the first place, I think it's not worth worrying about:
```
AlterTableOperation::DropConstraint { name } => write!(f, "DROP CONSTRAINT {}", name),
```

mashuai:
> The coverage is somewhat flaky unfortunately

Thanks, @nickolay , It's ready for review.

src/ast/ddl.rs
```diff
@@ -23,15 +23,60 @@ use std::fmt;
 pub enum AlterTableOperation {
     /// `ADD <table_constraint>`
     AddConstraint(TableConstraint),
+    AddColumn {
+        has_column_identifier: bool,
```
nickolay, line 27: I think this represents whether the `COLUMN` keyword is there, and `ADD` vs `ADD COLUMN` is purely a stylistic difference? In that case I think we could drop this from the AST, parse `COLUMN` [like this](https://github.com/andygrove/sqlparser-rs/blob/fab6e2827195b6067cac5e5b3a1a9bb6bc161ae7/src/parser.rs#L1315), and prefer  `ADD COLUMN` when serializing back to string. This is mainly to avoid coming up with a less confusing name than `has_column_identifier`.

src/ast/ddl.rs
```diff
@@ -23,15 +23,60 @@ use std::fmt;
...
+        name: Ident,
+    },
+    RenameColumn {
+        has_column_identifier: bool,
+        old_column_name: Ident,
+        new_column_name: Ident,
+    },
+    RenameTable {
+        table_name: Ident,
+    },
```
nickolay, line 41: Please add comments / remove `has_column_identifier` from these too.

mz:
> Hey @mz, thanks for the PR! Please let me know when this is ready for review, as I see you're still working on it.

Hey, I'm @mz and he is @mashuai :-) but nice to meet you both anyway !

src/parser.rs
```diff
@@ -1122,6 +1122,29 @@ impl Parser {
         })
     }
 
+    fn parse_column(&mut self) -> Result<ColumnDef, ParserError> {
```
nickolay, line 1125: This is taken from `parse_columns`, right? Why not re-use it there now?

src/parser.rs
```diff
@@ -1318,10 +1341,33 @@ impl Parser {
...
+                AlterTableOperation::RenameColumn {
+                    has_column_identifier,
+                    old_column_name,
+                    new_column_name,
+                }
             }
         } else {
-            return self.expected("ADD after ALTER TABLE", self.peek_token());
+            return self.expected(
+                "ADD,RENAME TO or RENAME after ALTER TABLE",
```
nickolay, line 1368: Why not simply "ADD or RENAME"?

src/parser.rs
mashuai: Yes, this is taken from `parse_columns`, I can't re-use it , because in `alter table ` grammar, it doesn't start with `LParen`, if re-use `parse_columns`, it will always return an empty vector.
https://github.com/andygrove/sqlparser-rs/blob/fab6e2827195b6067cac5e5b3a1a9bb6bc161ae7/src/parser.rs#L1128

nickolay: I mean call the new `parse_column` method _from_ `parse_columns`, not the other way around.

PR #208, opened 2020-06-24
Add without id support for SQLite

mashuai:
SQLite supports `WITHOUT ROWID` at the end of `CREATE TABLE` [SQLite Create Table](https://sqlite.org/lang_createtable.html)

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -482,6 +482,7 @@ pub enum Statement {
         file_format: Option<FileFormat>,
         location: Option<String>,
         query: Option<Box<Query>>,
+        without_rowid: Option<bool>,
     },
     /// CREATE INDEX
     CreateIndex {
@@ -647,6 +648,7 @@ impl fmt::Display for Statement {
                 file_format,
                 location,
                 query,
+                without_rowid,
             } => {
                 // We want to allow the following options
                 // Empty column list, allowed by PostgreSQL:
@@ -686,6 +688,11 @@ impl fmt::Display for Statement {
                 }
                 if let Some(query) = query {
                     write!(f, " AS {}", query)?;
+                }
+                if let Some(without_rowid) = *without_rowid {
+                    if without_rowid {
+                        write!(f, "WITHOUT ROWID")?;
+                    }
                 }
                 Ok(())
             }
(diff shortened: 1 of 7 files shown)


src/ast/mod.rs
```diff
@@ -687,6 +689,11 @@ impl fmt::Display for Statement {
                 if let Some(query) = query {
                     write!(f, " AS {}", query)?;
                 }
+                if let Some(without_rowid) = *without_rowid {
+                    if without_rowid {
+                        write!(f, "WITHOUT ROWID")?;
```
Dandandan, line 694: I think we should add a space before `WITHOUT` to be consistent.

tests/sqlparser_common.rs
```diff
@@ -1023,7 +1023,7 @@ fn parse_create_table() {
                constrained INT NULL CONSTRAINT pkey PRIMARY KEY NOT NULL UNIQUE CHECK (constrained > 0),
                ref INT REFERENCES othertable (a, b),\
                ref2 INT references othertable2 on delete cascade on update no action\
-               )";
+               ) WITHOUT ROWID";
```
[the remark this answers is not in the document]
mashuai: That's good, I add a new `sqlparser_sqlite.rs` to test that.

tests/sqlparser_sqlite.rs
```diff
@@ -0,0 +1,74 @@
...
+        Statement::CreateTable {
+            name,
+            columns,
+            constraints,
+            with_options: _with_options,
+            if_not_exists: false,
+            external: false,
+            file_format: None,
+            location: None,
+            query: _query,
```
Dandandan, line 37: maybe also use `query: None` or keep `with_options` and `query` out of the test (`,..` as well)?

src/ast/mod.rs
```diff
@@ -689,10 +689,8 @@ impl fmt::Display for Statement {
                 if let Some(query) = query {
                     write!(f, " AS {}", query)?;
                 }
-                if let Some(without_rowid) = *without_rowid {
-                    if without_rowid {
-                        write!(f, " WITHOUT ROWID")?;
-                    }
+                if *without_rowid {
+                    write!(f, " WITHOUT ROWID")?;
                 }
```
nickolay, line 694: This ought to be moved as well, to mirror the changes in the parsing code.

nickolay:
I also unresolved two comments that I don't think you have addressed.

tests/sqlparser_common.rs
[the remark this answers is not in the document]
nickolay: .. and don't add it to a test that runs for all dialects, I meant.

nickolay:
Awesome, thanks!

PR #209, opened 2020-06-26
Support SQLite `CREATE VIRTUAL TABLE`

mashuai:
Support SQLite [`CREATE VIRTUAL TABLE` ](https://www.sqlite.org/lang_createvtab.html)


nickolay:
I think that we'll eventually fold the CreateVirtualTable AST variant into the generic Statement::CreateTable (and re-use the table schema parser for `module_args`), but I think this is good for now. Thanks!

Please note that it's preferable to use one of the `verified_...` methods in the test, instead of duplicating the SQL text in the tests (60f60d9)

PR #211, opened 2020-06-27
Add fuzzer based on cargo-fuzz

Dandandan:
Currently it returns (after fuzzing):

`Error: Fuzz target exited with signal: 11` 

That looks like a problem in the nightly compiler (LLVM?)

### DIFF
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -1,0 +1,9 @@
+# Fuzzing
+
+sqlparser uses the tool `cargo-fuzz`.  This tool can be installed using: `cargo install cargo-fuzz`.
+Cargo-fuzz also requires nightly to be installed: `rustup install nightly`
+
+Running the fuzzer is as easy as running in the `fuzz` directory:
+
+`cargo +nightly fuzz run parse_sql`
+
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -5,5 +5,5 @@ Cargo-fuzz also requires nightly to be installed: `rustup install nightly`
 
 Running the fuzzer is as easy as running in the `fuzz` directory:
 
-`cargo +nightly fuzz run parse_sql -- -max_len=64`
+`cargo +nightly fuzz run parse_sql -- -max_len=64 -jobs=16 -timeout=1`
 
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -5,5 +5,5 @@ Cargo-fuzz also requires nightly to be installed: `rustup install nightly`
 
 Running the fuzzer is as easy as running in the `fuzz` directory:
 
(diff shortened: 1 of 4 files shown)


nickolay:
I'm not sure I understand the value of fuzzing sqlparser in general, and of fuzzing without first devising a way to generate SQL-looking input, in particular. Or did you mean to mark this PR as a WIP? Could you elaborate?

Dandandan:
The fuzzer (quickly) generates bytes.

Because of a few things at starts generating:

* It quickly generates bytes, starting from smaller inputs
* It tries to discover new code paths, and tries to find smaller examples for existing paths
* It uses a genetic algorithm to generate new inputs
* It keeps a corpus of which paths have been tries

It can also run in parallel (-- -njobs=16) to make it even faster.
Some more details here https://llvm.org/docs/LibFuzzer.html and here https://rust-fuzz.github.io/book/cargo-fuzz.html

The example comes from here: https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html

Could also use this to make it a bit more smarter: https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html . For example, we could generate arbitrary strings instead of bytes. 

some random things in the corpus (after running for a few secods), looks like it covers many things in the tokenizer:

`((*`
`/,,$,,,,,,,,,,,,3,,M* `
`'''`
`>>!=!=!X`
`(INDEX<=A(<=A(NX<=((((((<=}666F\`

The thing is that currently, because of the error above, it doesn't generate a cool report.

Although it may also be an actual problem like https://github.com/MaterializeInc/materialize/pull/3429

Dandandan:
Also to elaborate on the general goal to parse sqlparser in is mainly to find inputs which make the parser crash. This could be because of any use of partial functions like `unwrap`, out of bounds access, stack overflows, etc.

Dandandan:
It looks like the size of the problems indeed is the problem (I think related to a recursion problem).
By adding a maximum length, it just continues and covers already >1600 lines of the program (also many parser functions). I'll just keep it running for a while to see whether it can get it to crash!

alamb:
@Dandandan  shall we close this based on the contribution from @PsiACE  in https://github.com/sqlparser-rs/sqlparser-rs/pull/312 ?

Dandandan:
> @Dandandan shall we close this based on the contribution from @PsiACE in #312 ?

Yeah sounds good. According to some articles, different fuzzers can detect different bugs, but for now it seems we should pick just one option.
https://theultramarine19.github.io/data/736.pdf

alamb:
I also filed https://github.com/apache/arrow-datafusion/issues/913 in DataFusion for using a "domain specific" fuzzer (aka that generates valid sql) which may be applicable to sqlparser-rs as well

PR #217, opened 2020-06-28
Fix performance of very nested SQL files, add example to benchmark

Dandandan:
This adds one of the inputs found during fuzzing as benchmarking input.

https://github.com/ballista-compute/sqlparser-rs/issues/216

According to benchmark,the query takes 75.814 ms on average. That is much more than the other benchmarks, which are more than 1000x as fast (3 / 17 microseconds).

I included a fix which memoizes the unsuccesful usage of parse_derived_table_factor.

This brings down the benchmark to 109.24 microseconds (687 times speedup for this extreme example).

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -37,6 +37,19 @@ fn basic_queries(c: &mut Criterion) {
     group.bench_function("sqlparser::with_select", |b| {
         b.iter(|| Parser::parse_sql(&dialect, with_query));
     });
+
+    let nested_query = "
+    SELECT
+@ FROM((((((((((SELECT
+@ FROM((((((SELECT
+@ FROM((((((((((SELECT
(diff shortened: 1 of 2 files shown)


Dandandan:
According to benchmark, this takes 75.814 ms . That is much more than the other benchmarks, which are more than 1000x as fast.

Dandandan:
I included a fix which memoizes the unsuccesful usage of `parse_derived_table_factor`.

This brings down the benchmark to 114.16 microseconds (656 times speedup for this extreme example).

alamb:
@Dandandan   Is this PR still something we should  work on to help contribute?

Dandandan:
> @Dandandan   Is this PR still something we should  work on to help contribute?

I think the benchmark is a useful example of some problematic input for the parser.
Not sure whether this PR is a good way to solve it (better would be to rewrite the parser to avoid this)

alamb:
Closing old PRs. Please feel free to re-open if this was a mistake

PR #219, opened 2020-07-03
update travis badge to point to actions status

maxcountryman:
It seems Travis is currently not updating as expected, likely as a side effect of the repo moving. If we're comfortable leaning on Actions, then we can switch out the badge here and plan to remove Travis entirely. Alternatively we could reconfigure Travis to work with the new repo name.


Dandandan:
What about changing the link to

https://github.com/ballista-compute/sqlparser-rs/actions?query=workflow%3ARust+branch%3Amain

so it also shows the results only from the main branch.

I see no problems with going using GitHub Actions entirely.

PR #229, opened 2020-07-17
Remove PostgreSQL version of assert

Dandandan:
I linked earlier to this page but didn't notice this was part of PL/pgSQL (so no valid SQL statement in PostgreSQL).

In that case we can better remove the postgres part and simplify the code a bit.
https://www.postgresql.org/docs/10/plpgsql-errors-and-messages.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -556,7 +556,6 @@ pub enum Statement {
     Assert {
         condition: Expr,
         // AS or ,
-        separator: String,
         message: Option<Expr>,
     },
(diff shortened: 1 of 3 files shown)


nickolay:
In #226 I assumed both variants (BigQuery and Postgres) of ASSERT were procedural statements. But please feel free to simplify by dropping support for PG's syntax!

Dandandan:
Yeah I think that's better, as they are not valid non procedural statements that we should mix with the other ones.

PR #234, opened 2020-07-22
Support AUTO_INCREMENT for MySQL 

mashuai:
Support AUTO_INCREMENT for MySQL [MySQL Create Table](https://dev.mysql.com/doc/refman/8.0/en/create-table.html)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1266,6 +1266,8 @@ impl Parser {
             let expr = self.parse_expr()?;
             self.expect_token(&Token::RParen)?;
             ColumnOption::Check(expr)
+        } else if self.parse_keyword(Keyword::AUTO_INCREMENT) {
+            ColumnOption::AutoIncrement
         } else {
             return self.expected("column option", self.peek_token());
         };
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1268,10 +1268,26 @@ impl Parser {
             ColumnOption::Check(expr)
         } else if self.parse_keyword(Keyword::AUTO_INCREMENT) {
             // Support AUTO_INCREMENT for MySQL
-            ColumnOption::MySQLAutoIncrement
+            let word = Word {
+                value: "AUTO_INCREMENT".to_string(),
+                quote_style: None,
+                keyword: Keyword::AUTO_INCREMENT,
+            };
+            let token = Token::Word(word);
+            let mut vec = vec![];
+            vec.push(token);
+            ColumnOption::DialectSpecific(vec)
         } else if self.parse_keyword(Keyword::AUTOINCREMENT) {
             // Support AUTOINCREMENT for SQLite
-            ColumnOption::SQLiteAutoIncrement
+            let word = Word {
+                value: "AUTOINCREMENT".to_string(),
+                quote_style: None,
+                keyword: Keyword::AUTOINCREMENT,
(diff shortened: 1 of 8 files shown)


nickolay:
Yes, looks good to me too, but _please_ add a note about this being MySQL-specific as a comment in the code (better in all of: AST, the parser and the tests), as:
* PostgreSQL has [`SERIAL`](https://stackoverflow.com/a/787774/1026)
* MSSQL has [`IDENTITY [ (seed, increment) ]`](https://stackoverflow.com/questions/10991894/auto-increment-primary-key-in-sql-server-management-studio-2012)
* ANSI uses [`GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( ... )  ]`](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#identity-column-specification)

Documenting the differences between dialects like this should be useful both for our users and for designing a better solution for the dialects problem.

`AUTO_INCREMENT` also seems to be a good case for trying `ColumnOption::DialectSpecific(Vec<Token>)`, as [suggested earlier](https://github.com/ballista-compute/sqlparser-rs/issues/207#issuecomment-648208390), what do you think?

mashuai:
> Yes, looks good to me too, but _please_ add a note about this being MySQL-specific as a comment in the code (better in all of: AST, the parser and the tests), as:
> 
> * PostgreSQL has [`SERIAL`](https://stackoverflow.com/a/787774/1026)
> * MSSQL has [`IDENTITY [ (seed, increment) ]`](https://stackoverflow.com/questions/10991894/auto-increment-primary-key-in-sql-server-management-studio-2012)
> * ANSI uses [`GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( ... )  ]`](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#identity-column-specification)
> 
> Documenting the differences between dialects like this should be useful both for our users and for designing a better solution for the dialects problem.
> 
> `AUTO_INCREMENT` also seems to be a good case for trying `ColumnOption::DialectSpecific(Vec<Token>)`, as [suggested earlier](https://github.com/ballista-compute/sqlparser-rs/issues/207#issuecomment-648208390), what do you think?

mashuai:
Thank you for your suggestions. I find SQLite support `AUTOINCREMENT`, so add `MySQLAutoIncremnt` and `SQLiteAutoIncrement` to solve the problem.

src/parser.rs
```diff
@@ -1268,10 +1268,26 @@ impl Parser {
...
-            ColumnOption::MySQLAutoIncrement
+            let word = Word {
+                value: "AUTO_INCREMENT".to_string(),
+                quote_style: None,
+                keyword: Keyword::AUTO_INCREMENT,
+            };
+            let token = Token::Word(word);
+            let mut vec = vec![];
+            vec.push(token);
+            ColumnOption::DialectSpecific(vec)
```
nickolay, line 1279: This would be much simpler with `Token::make_keyword()`, I think.

I hoped there was a way to avoid keyword duplication by remembering the position in the token stream before trying to parse. When parsing is successful, we could simply copy the tokens starting from that position into the DialectSpecfic variant. I'm not up to trying to implement that now, so `make_keyword` is fine with me too.

PR #235, opened 2020-07-23
Implement Hive QL Parsing

hntd187:
This is a highly experimental start to implementing Hive QL in the parser. Luckily, the parser's design lends itself well to adding in some of these things, but there are still a few oddities and things to implement to make something for Hive complete. I'd like to get people reviewing this early because in my opinion I do some janky things here I should clean up and I'm hoping early review will help me nip that in the bud.

Also, my tests are a bit on the light side right now, but I have verified a lot of this functionality with some internal HQL that I cannot share publicly. I am confident that this covers the majority of edge cases on how not to write HQL.


src/ast/mod.rs
```diff
@@ -627,6 +753,25 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
+            Statement::CreateDatabase {
+                db_name,
+                ine,
```
Dandandan, line 758: `if_not_exists` , better to write it out without abbreviations

src/ast/data_type.rs
```diff
@@ -61,6 +61,8 @@ pub enum DataType {
     Regclass,
     /// Text
     Text,
+    /// String (Hive)
```
Dandandan, line 64: Also some other dialects besides hive use `string`

src/ast/mod.rs
```diff
@@ -566,15 +607,100 @@ impl fmt::Display for Statement {
...
+                    }
+                )
+            }
+            Statement::Truncate {
+                table_name,
+                partitions,
+            } => {
+                write!(f, "TRUNCATE TABLE {}", table_name)?;
+                if let Some(ref parts) = partitions {
+                    if !parts.is_empty() {
```
Dandandan, line 649: Is `PARTITIONS ()` also possible? If so, then this check is not needed. Otherwise, we could check the partitions type to a vec instead of optional vec.

src/ast/mod.rs
```diff
@@ -425,6 +425,28 @@ impl fmt::Display for WindowFrameBound {
...
+    /// Truncate (Hive)
+    Truncate {
+        table_name: ObjectName,
+        partitions: Option<Vec<Expr>>,
+    },
+    /// Msck (Hive)
+    Msck {
+        table_name: ObjectName,
+        repair: bool,
+        add_partitions: bool,
```
Dandandan, line 446: Add/drop/sync could be an enum I think? Also simplifies the parser code a bit and allows for easier patter matching.

Dandandan:
Looks great, would love to have this! I added a few comments

src/ast/mod.rs
hntd187: I don't believe `PARTITIONS ()` is valid, you have to specify at the very least a column.

src/ast/mod.rs
```diff
@@ -420,11 +420,49 @@ impl fmt::Display for WindowFrameBound {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum PartitionAction {
+    ADD,
```
Dandandan, line 426: Can we change naming to `Add`, `Drop`, `Sync` ?

hntd187:
So @Dandandan I ran into a discussion point here. Hive allows you to have column aliases beginning with numbers such as `SELECT a as 3b from table` with no quotes or anything of the sort around it. This basically breaks most ident parsing, which led me to disable the hive dialect in the ansi tests for now. While I don't believe hive has ever conformed to anything close to ansi sql there might be a better way to do this or support numbers as beginning identifiers.

Dandandan:
@hntd187 did you try adding numbers to

`is_identifier_start` for this dialect? I am not sure how that combines with other parts of the tokenizer, but if this causes any issue, I think the tokenizer could be adapted for this case.

hntd187:
Yes I did @Dandandan it basically breaks 30 tests in the ansi dialect comparison leading me to remove that change for now. I think Hive _should_ parse the standard ANSI statements the exact same way and do it's own thing for the weird off standard syntax, but it seems like no dialect supports numbers first.

Dandandan:
I think the tokenizer / dialect trait needs to change a bit for numbers in the start position.
If the start character is a number, there needs to be a alphabetical character as well (otherwise it would be a number).

E.g. changing this in the tokenizer (yes,code is a bit messy, but this can be cleaned up / made easier / redesign a bit) succeeds all the test after allowing to start on `0-9`.

```
                ch if self.dialect.is_identifier_start(ch) => {
                    chars.next(); // consume the first char
                    let s = self.tokenize_word(ch, chars);

                    if s.chars().all(|x| x >= '0' && x <= '9' || x == '.') {
                        let mut s = peeking_take_while(&mut s.chars().peekable(), |ch| match ch {
                            '0'..='9' | '.' => true,
                            _ => false,
                        });
                        let s2 = peeking_take_while(chars, |ch| match ch {
                            '0'..='9' | '.' => true,
                            _ => false,
                        });
                        s += s2.as_str();
                        return Ok(Some(Token::Number(s)));
                    }
                    Ok(Some(Token::make_word(&s, None)))
                }```

**Comment by nickolay** [comment]
> no dialect supports numbers first.

[MySQL does](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html), which leads them to make recommendations like _"avoid using 1e as an identifier, because an expression such as 1e+3 is ambiguous. Depending on context, it might be interpreted as the expression 1e + 3 or as the number 1e+3."_, which makes me want to start with documenting the expected behavior before discussing the implementation. I would suggest separating this from the rest of Hive support.

**Comment by hntd187** [comment]
I think that's reasonable, but idents starting with numerics is part of hive support, but since it's not hive specific should happen at the tokenizer level like Dan suggested, I suppose the MySQL Dialect would need something similar in it to what hive has.

**Comment by andygrove** [comment]
@hntd187 @Dandandan Seems like we should go ahead and merge this now? I will skim through and merge if there are no objections.

File: src/ast/value.rs [hunk 6]
```diff
@@ -22,15 +22,17 @@ use std::fmt;
 pub enum Value {
     /// Numeric literal
     #[cfg(not(feature = "bigdecimal"))]
-    Number(String),
+    Number(String, bool),
```
andygrove, line 25: nit: I wonder if a new type should have been added rather than change this type?

andygrove (approved):
LGTM

src/tokenizer.rs
```diff
@@ -35,7 +35,7 @@ pub enum Token {
     /// A keyword (like SELECT) or an optionally quoted SQL identifier
     Word(Word),
     /// An unsigned numeric literal
-    Number(String),
+    Number(String, bool),
```
Dandandan, line 38: I think this could be better extracted to a named field, e.g. `long: bool`?.
This helps readability

src/ast/value.rs
Dandandan, line 25: For those values as well

Dandandan:
@hntd187 @andygrove 
To me it looks great as well, added some final comments, could also be done in a follow-up.
Thanks @hntd187, for the great addition and for your patience 😎

PR #238, opened 2020-07-25
automate crate publishing

maxcountryman:
This introduces a new section of the GitHub Actions workflow which will
publish the crate upon a new tag being pushed. Note that this requires a
new project secret, `CRATES_TOKEN`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,3 +34,6 @@ serde_json = { version = "1.0", optional = true }
 [dev-dependencies]
 simple_logger = "1.0.1"
 matches = "0.1"
+
+[package.metadata.release]
+disable-publish = true
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,0 +1,32 @@
+# Releasing
+
+Releasing, i.e. crate publishing, has been automated via GitHub Actions.
+
+In order to author a new release, you simply tag the desired revision and push
+the resulting tag.
+
+**Before releasing** ensure `CHANGELOG.md` is updated appropriately as well as
+`Cargo.toml`.
+
+## Process
+
+Please ensure you follow the correct format when creating new tags. For
+instance:
+
+```
+git tag -a '0.6.0' -m '(cargo-release) sqlparser version 0.6.0'
+```
+
+This will create a new tag, `0.6.0` which the message,
+`(cargo-release) sqlparser version 0.6.0`.
+
+Once the tag is created, pushing the tag upstream will trigger a publishing
+process to crates.io. Now to push our example tag:
+
+```
+git push origin 0.6.0
+```
+
+(Note that this process is fully automated; credentials
+for authoring in this way are securely stored in the repo secrets as
+`CRATE_TOKEN`.)
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -2,20 +2,25 @@
 
 Releasing, i.e. crate publishing, has been automated via GitHub Actions.
 
-In order to author a new release, you simply tag the desired revision and push
-the resulting tag.
+We use the [`cargo release`](https://github.com/sunng87/cargo-release)
+subcommand to ensure correct versioning. Install via:
 
-**Before releasing** ensure `CHANGELOG.md` is updated appropriately as well as
-`Cargo.toml`.
+```
+$ cargo install cargo-release
+```
+
+**Before releasing** ensure `CHANGELOG.md` is updated appropriately.
 
 ## Process
 
-Please ensure you follow the correct format when creating new tags. For
-instance:
+Using `cargo-release` we can author a new minor release like so:
 
 ```
-git tag -a '0.6.0' -m '(cargo-release) sqlparser version 0.6.0'
+$ cargo release minor --skip-publish
 ```
+
+**Ensure publishing is skipped** since pushing the resulting tag upstream will
+handle crate publishing automatically.
 
 This will create a new tag, `0.6.0` with the message,
 `(cargo-release) sqlparser version 0.6.0`.
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -17,7 +17,7 @@ instance:
 git tag -a '0.6.0' -m '(cargo-release) sqlparser version 0.6.0'
 ```
 
-This will create a new tag, `0.6.0` which the message,
+This will create a new tag, `0.6.0` with the message,
 `(cargo-release) sqlparser version 0.6.0`.
 
 Once the tag is created, pushing the tag upstream will trigger a publishing
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,3 +63,16 @@ jobs:
       uses: coverallsapp/github-action@master
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  publish-crate:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    needs: [test]
+    steps:
+      - name: Set up Rust
+        uses: hecrj/setup-rust-action@v1
+      - uses: actions/checkout@v2
+      - name: Publish
+        shell: bash
+        run: |
+          cargo publish --token ${{ secrets.CRATES_TOKEN }}


maxcountryman:
@nickolay or @andygrove, if one of you could add the `CRATES_TOKEN` to the project secrets then this should be good to go. 🙂

Dandandan:
@maxcountryman maybe could use some docs/instructions to describe how to publish (pushing with the right tags)

nickolay:
I don't think I have the rights to do that.

maxcountryman:
> @maxcountryman maybe could use some docs/instructions to describe how to publish (pushing with the right tags)

Yep--I wonder where best to keep that?

andygrove:
@maxcountryman I've temporarily made you an admin on the repo

maxcountryman:
Thanks @andygrove, I created a new token and added it to the repo secrets. I shouldn't need admin access any longer.

Dandandan:
> > @maxcountryman maybe could use some docs/instructions to describe how to publish (pushing with the right tags)
> 
> Yep--I wonder where best to keep that?

Maybe a new file like `RELEASING.md`?

maxcountryman:
> > > @maxcountryman maybe could use some docs/instructions to describe how to publish (pushing with the right tags)
> 
> > 
> 
> > Yep--I wonder where best to keep that?
> 
> 
> 
> Maybe a new file like `RELEASING.md`?

Sure. I can add something to this PR.

maxcountryman:
I noticed [our tags](https://github.com/ballista-compute/sqlparser-rs/tags) are named slightly inconsistently. Does anyone have a strong preference or could be move towards omitting the `v` prefix?

Dandandan:
Without v sounds good to me

maxcountryman:
@nickolay can I just check you're all right with this methodology? I don't want to commit us to this process without everyone's buy-in. 🙂

nickolay:
Push-to-publish is great!
I have no opinion about vX.Y vs X.Y for the tag names.
As for the documentation, the process to update the version number on master is not documented (I think Andy uses some tool for that?), including a reminder to follow semver, as described in CHANGELOG.md.

Thanks!

maxcountryman:
> As for the documentation, the process to update the version number on master is not documented (I think Andy uses some tool for that?), including a reminder to follow semver, as described in CHANGELOG.md.

Thanks for pointing that out. @andygrove would you mind sharing the process/tool we should use for that? I can update the release doc accordingly.

andygrove:
I've been using cargo release.

https://github.com/sunng87/cargo-release

Cargo.toml
```diff
@@ -34,3 +34,6 @@ serde_json = { version = "1.0", optional = true }
 [dev-dependencies]
 simple_logger = "1.0.1"
 matches = "0.1"
+
+[package.metadata.release]
+disable-publish = true
```
nickolay, line 39: Please add a comment as to why this is needed. I initially thought this was for cargo-release, but the docs below recommend using the `--skip-publish` parameter.

docs/releasing.md
```diff
@@ -2,21 +2,26 @@
...
+$ cargo install cargo-release
+```
+
+**Before releasing** ensure `CHANGELOG.md` is updated appropriately.
 
 ## Process
 
-Please ensure you follow the correct format when creating new tags. For
-instance:
+Using `cargo-release` we can author a new minor release like so:
```
nickolay, line 16: I'm not thinking of the new releases as "minor", so perhaps it would be useful to remind that this means `0.(N+1)` in our context?

Cargo.toml
maxcountryman: It is for cargo release. This is a fallback in case we forget to use the flag.

docs/releasing.md
maxcountryman: Sure that makes sense to me. Maybe we should simply link to the CHANGELOG where that's described?

nickolay: Sure, but CHANGELOG doesn't mention "minor" releases either. Technically, "minor" is the correct term with regards to semver, but in context of sqlparser using it without mentioning semver is slightly weird. Do what you think is best, I don't have a strong opinion on if or how this should be addressed.

Dandandan:
Anything that remains on this one? Would like to see this in action!

nickolay:
I was wondering about that as well!

A couple of ideas for the future:
* I noticed that the `publish-crate` "check" is displayed as "skipped" on this PR. Is it at all possible to only run it on pushes to certain tags ([the docs seem to say so](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags))?
* When/if https://github.com/actions-rs/meta/issues/6 is resolved we should consider using the standardized solution.

Dandandan:
> I was wondering about that as well!
> 
> A couple of ideas for the future:
> 
> * I noticed that the `publish-crate` "check" is displayed as "skipped" on this PR. Is it at all possible to only run it on pushes to certain tags ([the docs seem to say so](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags))?
> * When/if [actions-rs/meta#6](https://github.com/actions-rs/meta/issues/6) is resolved we should consider using the standardized solution.

* I like the idea of making this only on a tag! I think it has to move to a new workflow/file in that case.

* Also, indeed might make a note to use that in the future.

maxcountryman:
Sorry, I haven't had a chance to get back to this. I'll take another look this evening--those all sound like good ideas but perhaps we can circle back and prioritize getting this live first?

nickolay:
> prioritize getting this live first?

Sure, this is exactly what I meant by calling my notes "ideas for the future"!

docs/releasing.md
maxcountryman: I'll leave this as-is for now.

maxcountryman:
All right, this is merged. Hopefully I configured it correctly. 🤞

nickolay:
Great! I pushed a PR updating the changelog (#261) so that we can test out the new release procedure.

I also unresolved my comments above, so that they would be visible by default, if you don't mind.

Cargo.toml
nickolay: Thanks for dealing with this in https://github.com/ballista-compute/sqlparser-rs/pull/263 !

nickolay:
Continuing from https://github.com/ballista-compute/sqlparser-rs/pull/238#issuecomment-667204734

> I noticed [our tags](https://github.com/ballista-compute/sqlparser-rs/tags) are named slightly inconsistently. Does anyone have a strong preference or could be move towards omitting the `v` prefix? (It shouldn't make a difference for automation, but I figured as long as we're documenting the process we could establish a consistent pattern.)

it seems you didn't implement this? I'll note that the newer tags start with `v` due to a [change in cargo-release](https://github.com/sunng87/cargo-release/issues/124). They claim that the "v" prefix is more common.

maxcountryman:
> Continuing from https://github.com/ballista-compute/sqlparser-rs/pull/238#issuecomment-667204734
> 
> 
> 
> > I noticed [our tags](https://github.com/ballista-compute/sqlparser-rs/tags) are named slightly inconsistently. Does anyone have a strong preference or could be move towards omitting the `v` prefix? (It shouldn't make a difference for automation, but I figured as long as we're documenting the process we could establish a consistent pattern.)
> 
> 
> 
> it seems you didn't implement this? I'll note that the newer tags start with `v` due to a [change in cargo-release](https://github.com/sunng87/cargo-release/issues/124). They claim that the "v" prefix is more common.
> 
> 
> 
> [edit] also note that future changes are being discussed in https://github.com/ballista-compute/sqlparser-rs/pull/263

I'm not sure I follow? I implemented what Andy had suggested, which I believe was to use cargo release. So the naming is handled by cargo release now.

nickolay:
Andy didn't say anything about the naming of tags. Older cargo-release used to create X.Y tags and the current version creates vX.Y tags, which explains the inconsistency you found.

You initially intended to standardize on using tags named "X.Y" and even put `git push origin 0.6.0` in [docs/releasing.md](https://github.com/ballista-compute/sqlparser-rs/blob/9351efb437f27d2c36fa4f1582ff9a1e01d50e15/docs/releasing.md#process), while the current version of cargo-release creates tags named vX.Y as we didn't configure it to do otherwise. We should either accept the cargo-release's new default and update the docs, or add the configuration to Cargo.toml, so that cargo-release creates X.Y (sans the `v` prefix) tags.

PR #239, opened 2020-07-26
Support create or replace view/table

Dandandan:
Adds support for create or replace (external) table

https://github.com/ballista-compute/sqlparser-rs/issues/228

Examples:
https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language
https://docs.snowflake.com/en/sql-reference/sql/create-external-table.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -462,25 +462,25 @@ pub enum Statement {
     },
     /// CREATE VIEW
     CreateView {
+        or_replace: bool,
+        materialized: bool,
         /// View name
         name: ObjectName,
         columns: Vec<Ident>,
         query: Box<Query>,
-        materialized: bool,
-        or_replace: bool,
         with_options: Vec<SqlOption>,
     },
     /// CREATE TABLE
     CreateTable {
+        or_replace: bool,
+        external: bool,
+        if_not_exists: bool,
         /// Table name
         name: ObjectName,
         /// Optional schema
         columns: Vec<ColumnDef>,
         constraints: Vec<TableConstraint>,
         with_options: Vec<SqlOption>,
-        or_replace: bool,
-        if_not_exists: bool,
-        external: bool,
         file_format: Option<FileFormat>,
         location: Option<String>,
         query: Option<Box<Query>>,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -464,6 +464,7 @@ pub enum Statement {
     CreateView {
         /// View name
         name: ObjectName,
+        or_replace: bool,
         columns: Vec<Ident>,
         query: Box<Query>,
         materialized: bool,
@@ -630,12 +631,18 @@ impl fmt::Display for Statement {
             }
             Statement::CreateView {
                 name,
+                or_replace,
                 columns,
                 query,
                 materialized,
                 with_options,
             } => {
                 write!(f, "CREATE")?;
+
+                if *or_replace {
+                    write!(f, " OR REPLACE")?;
+                }
+
                 if *materialized {
                     write!(f, " MATERIALIZED")?;
                 }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -464,10 +464,10 @@ pub enum Statement {
     CreateView {
         /// View name
         name: ObjectName,
-        or_replace: bool,
         columns: Vec<Ident>,
         query: Box<Query>,
         materialized: bool,
+        or_replace: bool,
         with_options: Vec<SqlOption>,
     },
     /// CREATE TABLE
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -477,6 +477,7 @@ pub enum Statement {
         columns: Vec<ColumnDef>,
         constraints: Vec<TableConstraint>,
         with_options: Vec<SqlOption>,
+        or_replace: bool,
         if_not_exists: bool,
         external: bool,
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -988,16 +988,25 @@ impl Parser {
     /// Parse a SQL CREATE statement
     pub fn parse_create(&mut self) -> Result<Statement, ParserError> {
         if self.parse_keyword(Keyword::TABLE) {
-            self.parse_create_table()
+            self.parse_create_table(false)
+        } else if self.parse_keywords(&[Keyword::OR, Keyword::REPLACE, Keyword::TABLE]) {
+            self.parse_create_table(true)
```
nickolay, line 993: Would this be clearer as:

```
let or_replace = self.parse_keywords(&[Keyword::OR, Keyword::REPLACE]);
if self.parse_keyword(Keyword::TABLE) {
   self.parse_create_table(or_replace)
} ...
```
?

src/parser.rs
```diff
@@ -988,16 +988,25 @@ impl Parser {
...
         } else if self.parse_keyword(Keyword::INDEX) {
             self.parse_create_index(false)
         } else if self.parse_keywords(&[Keyword::UNIQUE, Keyword::INDEX]) {
             self.parse_create_index(true)
         } else if self.parse_keyword(Keyword::MATERIALIZED) || self.parse_keyword(Keyword::VIEW) {
             self.prev_token();
-            self.parse_create_view()
+            self.parse_create_view(false)
+        } else if self.parse_keywords(&[Keyword::OR, Keyword::REPLACE])
+            && (self.parse_keyword(Keyword::MATERIALIZED) || self.parse_keyword(Keyword::VIEW))
```
nickolay, line 1002: For `CREATE OR REPLACE MATERIALIZED VIEW` and for `CREATE OR REPLACE EXTERNAL TABLE` below please either add a test, mentioning which dialects support that or don't add support parsing that.

src/ast/mod.rs
```diff
@@ -464,6 +464,7 @@ pub enum Statement {
     CreateView {
         /// View name
         name: ObjectName,
+        or_replace: bool,
```
nickolay, line 467: Why did you put `or_replace` between `name` and `columns`?

src/parser.rs
Dandandan: Thought about it too, however tries to parse "or replace" for every other alternative too.
Code is a bit cleaner though.

src/parser.rs
Dandandan: :+1:

src/parser.rs
nickolay: If by "tries to parse" you mean "accepts", that doesn't have to be the case. I meant 
```
let or_replace = self.parse_keywords(&[Keyword::OR, Keyword::REPLACE]);
if self.parse_keyword(Keyword::TABLE) {
   self.parse_create_table(or_replace)
} else if self.parse_keyword(Keyword::VIEW) {
   self.parse_create_table(or_replace)
} else if or_replace {
  // error
} else if self.parse_keyword(Keyword::INDEX) {
  ...
```

If you literally mean *tries*, then this https://github.com/ballista-compute/sqlparser-rs/pull/239/files#diff-2c09afcdc3c420ab0678ba9b5e83959cR992 already tries to parse "OR REPLACE" for every alternative.

src/parser.rs
Dandandan: Added tests for those options as well

src/ast/mod.rs
nickolay: I would be fine with any order, as long as it can be explained:)

What do you think about using the source code order, when possible? I.e. `or_replace`, `materialized`, `name`, ...?

src/parser.rs
```diff
@@ -1072,7 +1077,7 @@ impl Parser {
         }
     }
 
-    pub fn parse_create_view(&mut self) -> Result<Statement, ParserError> {
+    pub fn parse_create_view(&mut self, or_replace: bool) -> Result<Statement, ParserError> {
         let materialized = self.parse_keyword(Keyword::MATERIALIZED);
         self.expect_keyword(Keyword::VIEW)?;
         // Many dialects support `OR REPLACE` | `OR ALTER` right after `CREATE`, but we don't (yet).
```
nickolay, line 1083: Now we do!

src/parser.rs
nickolay: Your current version silently eats "OR REPLACE" with some object types, like CREATE INDEX - is that intentional?

src/ast/mod.rs
Dandandan: Makes sense, also applied that for CreateTable:
```
        or_replace
        external
        if_not_exists
        name
```

src/parser.rs
Dandandan: Hm that's right, after changing it with the 

`let or_replace = self.parse_keywords(&[Keyword::OR, Keyword::REPLACE]);`

it will consume those. Maybe better to revert to the first implementation, or do you have some other suggestion?

Dandandan: Ah I see the error part, will do that.

Dandandan: Resolved now

nickolay:
I've already "approved" this in GitHub UI earlier, but to confirm: this looks great to me!

Dandandan:
Awesome @nickolay ! Thanks for the review

PR #243, opened 2020-07-27
Add Postgres-specific statements PREPARE, EXECUTE and DEALLOCATE

### LINKED ISSUE #242: PREPARE, EXECUTE and DEALLOCATE are not top-level statements
Hi, I am using this crate to parse SQLs, and working on [Postgres Extended Query](https://www.postgresql.org/docs/current/sql-prepare.html#SQL-PREPARE-EXAMPLES). e.g.

```
prepare stmt (smallint) AS select * from schema_name.table_name where si_column_2 = $1;
execute stmt(1, 2);
deallocate stmt;
```

I tried to find solutions by `Dialect`, but it seems that `PREPARE`, `EXECUTE` and `DEALLOCATE` are not top-level statements.

Should they be added as top-level statements? Or there are other better solutions. Thanks.

silathdiir:
Closes #242 

Adds top-statements `PREPARE`, `EXECUTE` and `DEALLOCATE` for Postgres-specific feature `prepared statement`.

https://www.postgresql.org/docs/current/sql-prepare.html
https://www.postgresql.org/docs/current/sql-execute.html
https://www.postgresql.org/docs/current/sql-deallocate.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -551,31 +551,24 @@ pub enum Statement {
     Rollback { chain: bool },
     /// CREATE SCHEMA
     CreateSchema { schema_name: ObjectName },
-    /// ASSERT <condition> [AS <message>]
+    /// `ASSERT <condition> [AS <message>]`
     Assert {
         condition: Expr,
         message: Option<Expr>,
     },
-    /// DEALLOCATE [ PREPARE ] { name | ALL }
+    /// `DEALLOCATE [ PREPARE ] { name | ALL }`
     ///
     /// Note: this is a PostgreSQL-specific statement.
-    Deallocate {
-        name: Option<ObjectName>,
-        all: bool,
-        prepare: bool,
-    },
-    /// EXECUTE name [ ( parameter [, ...] ) ]
+    Deallocate { name: Ident, prepare: bool },
+    /// `EXECUTE name [ ( parameter [, ...] ) ]`
     ///
     /// Note: this is a PostgreSQL-specific statement.
-    Execute {
-        name: ObjectName,
-        parameters: Vec<Expr>,
-    },
-    /// PREPARE name [ ( data_type [, ...] ) ] AS statement
+    Execute { name: Ident, parameters: Vec<Expr> },
+    /// `PREPARE name [ ( data_type [, ...] ) ] AS statement`
     ///
     /// Note: this is a PostgreSQL-specific statement.
     Prepare {
-        name: ObjectName,
+        name: Ident,
         data_types: Vec<DataType>,
         statement: Box<Statement>,
     },
@@ -846,17 +839,12 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
-            Statement::Deallocate { name, all, prepare } => {
-                f.write_str("DEALLOCATE ")?;
-                if *prepare {
-                    f.write_str("PREPARE ")?;
-                }
-                if *all {
-                    f.write_str("ALL")
-                } else {
-                    write!(f, "{}", name.as_ref().unwrap())
-                }
-            }
+            Statement::Deallocate { name, prepare } => write!(
+                f,
+                "DEALLOCATE {prepare}{name}",
+                prepare = if *prepare { "PREPARE " } else { "" },
+                name = name,
+            ),
             Statement::Execute { name, parameters } => {
                 write!(f, "EXECUTE {}", name)?;
                 if !parameters.is_empty() {
@@ -873,7 +861,6 @@ impl fmt::Display for Statement {
                 if !data_types.is_empty() {
                     write!(f, "({}) ", display_comma_separated(data_types))?;
                 }
-
                 write!(f, "AS {}", statement)
             }
         }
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -551,12 +551,34 @@ pub enum Statement {
     Rollback { chain: bool },
     /// CREATE SCHEMA
     CreateSchema { schema_name: ObjectName },
-
     /// ASSERT <condition> [AS <message>]
     Assert {
         condition: Expr,
         message: Option<Expr>,
     },
+    /// DEALLOCATE [ PREPARE ] { name | ALL }
```
nickolay, line 559: This needs to be enclosed in backticks (as well as the `ASSERT` comment above -- sorry about missing that!), so that `cargo doc` output is properly formatted. Same below.

src/ast/mod.rs
```diff
@@ -551,12 +551,34 @@ pub enum Statement {
...
+    ///
+    /// Note: this is a PostgreSQL-specific statement.
+    Deallocate {
+        name: Option<ObjectName>,
+        all: bool,
```
nickolay, line 564: You shouldn't need to handle `ALL` separately, from the parser viewpoint, it's as good an object name, as any. Can `name` be schema-qualified? The docs are not clear on that.

src/ast/mod.rs
```diff
@@ -551,12 +551,34 @@ pub enum Statement {
...
+        prepare: bool,
```
nickolay, line 565: This seems to be meaningless syntax. It's fine to keep it, but maybe note that PostgreSQL docs say it's ignored?

src/ast/mod.rs
```diff
@@ -824,6 +846,36 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
+            Statement::Deallocate { name, all, prepare } => {
+                f.write_str("DEALLOCATE ")?;
+                if *prepare {
+                    f.write_str("PREPARE ")?;
+                }
```
nickolay, line 853: I see we're inconsistent about that, but I believe using the same style as in the Statement::CreateTable branch will yield smaller code (especially we figure out the ALL vs object name question above).

src/ast/mod.rs
```diff
@@ -824,6 +846,36 @@ impl fmt::Display for Statement {
...
+            Statement::Prepare {
+                name,
+                data_types,
+                statement,
+            } => {
+                write!(f, "PREPARE {} ", name)?;
+                if !data_types.is_empty() {
+                    write!(f, "({}) ", display_comma_separated(data_types))?;
+                }
+
```
nickolay, line 876: Again, I know we're inconsistent about this, but it seems that we usually don't add blank lines in situations like this.

src/dialect/postgresql.rs
```diff
@@ -20,7 +20,7 @@ impl Dialect for PostgreSqlDialect {
         // See https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
         // We don't yet support identifiers beginning with "letters with
         // diacritical marks and non-Latin letters"
-        (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || ch == '_'
+        (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || ch == '_' || ch == '$'
```
nickolay, line 23: This doesn't match the documentation linked just above this line (Postgres seems to have `$1` and `$$comments$$`, but I don't see how `$foo` is supposed to be parsed), so I'd like to split this separately. You [should be able](https://github.com/ballista-compute/sqlparser-rs/issues/48) to keep it in a dialect of your own for now.

src/parser.rs
```diff
@@ -2386,6 +2391,49 @@ impl Parser {
...
+        };
+
+        Ok(Statement::Deallocate { name, all, prepare })
+    }
+
+    fn parse_execute(&mut self) -> Result<Statement, ParserError> {
+        let name = self.parse_object_name()?;
+
+        let mut parameters = vec![];
+        if self.expect_token(&Token::LParen).is_ok() {
```
nickolay, line 2412: We have confusingly named `consume_token` for that. Same below.

tests/sqlparser_postgres.rs
```diff
@@ -423,6 +423,139 @@ fn parse_show() {
     )
 }
 
+#[test]
+fn parse_deallocate() {
+    let stmt = pg().verified_stmt("DEALLOCATE a");
```
nickolay, line 428: Use `pg_and_generic()` please.

src/ast/mod.rs
silathdiir: Sorry. I tried to check and test again. It is not schema-qualified. I will remove `all` and fix type of `name` to an `Ident`.

src/ast/mod.rs
silathdiir: Yes. The syntax is meaningless. But I think if we should output (format) the useless keyword `PREPARE` as the SQL passed into. What do you think?

Issue #241, opened 2020-07-27
Dialect-specific parsing and Snowflake JSON support


eyalleshem:
Hi , 

Currently the lib support different databases (postgresql-sql, mysql , etc .. ) dialect only in the tokenizer.

My question is - if there was any discussions/plans about supporting per-db dialects also in the parser (or do you you think it's good idea to start such discussions )? 

Where most of the DB's support ANSI sql , when it's come to more advance features like json-parsing   - There is great variability between the different databases.

nickolay:
Hey, I wrote about this recently in https://github.com/ballista-compute/sqlparser-rs/issues/7#issuecomment-659857344

nickolay:
If the question was specifically about discussing a change like this https://github.com/ballista-compute/sqlparser-rs/commit/c67d6670d4f5fb7fee5450b96a799f89b8660b2f then I was too quick to close this as a duplicate, as this is something we'll need eventually.

eyalleshem:
thanks @nikolay for the fast response :)  ,
 
Follow your answer i follow issue #7 , and the issues that mentation there . 

What i currently hope  to do is to parse snowflake-json queries with this this lib, and I wonder what  is the correct way for doing that. 

As json-Sql query in snowflake could look not native in ANSI-sql perspective -so the first idea that came into my mind is dialect.
(something like you mention in: [https://github.com/ballista-compute/sqlparser-rs/commit/c67d6670d4f5fb7fee5450b96a799f89b8660b2f](url) ).

json sql in snowflake could look like  :  
`SELECT b.this, a.V['is_dog'] FROM a1 a, table(FLATTEN(input => a.v)) as b  where a.v:is_dog::String like 'true';`

I could think also about some other approaches to deal with that - but maybe it's should be in another  issue , with more suitable title  :)

nickolay:
Right. This is a good example where different dialects conflict, as `a.V['is_dog']` means `a.V AS "'is_dog'"` in MSSQL. It is because `['is_dog']` is a valid identifier ([already configured in the dialect](https://github.com/ballista-compute/sqlparser-rs/blob/f053383c718b3d4942751224783c594ff7e9f0b1/src/dialect/mssql.rs#L19)) and because MSSQL does not support `[]` for subscripts.

[...]

PR #244, opened 2020-07-28
Use dialects in the parser for support snowflake aliasing syntax

eyalleshem:
Snowflake DB allow single table to be within parenthesis.
This behaviour is diffrent than other DB ,
and it has some impact on the parsing table factor.

For supporting we do the following :
1. Add refrence to the dialect in the parser
2. Add Snowflake dialect
3. add function to the dialect trait the identify if single table inside parenthesis allowed
4. When parsing table factor in the allow/deny single table inside parenthesis according to dialect

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -16,7 +16,7 @@ pub mod keywords;
 mod mssql;
 mod mysql;
 mod postgresql;
-
+mod snowflake;
 use std::fmt::Debug;
 
 pub use self::ansi::AnsiDialect;
@@ -24,6 +24,7 @@ pub use self::generic::GenericDialect;
 pub use self::mssql::MsSqlDialect;
 pub use self::mysql::MySqlDialect;
 pub use self::postgresql::PostgreSqlDialect;
+pub use self::snowflake::SnowflakeDialect;
 
 pub trait Dialect: Debug {
     /// Determine if a character starts a quoted identifier. The default
@@ -38,4 +39,8 @@ pub trait Dialect: Debug {
     fn is_identifier_start(&self, ch: char) -> bool;
     /// Determine if a character is a valid unquoted identifier character
     fn is_identifier_part(&self, ch: char) -> bool;
+
+    fn alllow_single_table_in_parenthesis(&self) -> bool {
+        false
+    }
 }
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -1,0 +1,26 @@
+use crate::dialect::Dialect;
+
+#[derive(Debug, Default)]
+pub struct SnowflakeDialect;
+
+impl Dialect for SnowflakeDialect {
+    //Revisit: currently copied from Genric dialect
+    fn is_identifier_start(&self, ch: char) -> bool {
+        (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || ch == '_' || ch == '#' || ch == '@'
+    }
+
+    //Revisit: currently copied from Genric dialect
+    fn is_identifier_part(&self, ch: char) -> bool {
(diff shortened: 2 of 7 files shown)


src/dialect/mod.rs
```diff
@@ -38,4 +39,8 @@ pub trait Dialect: Debug {
     fn is_identifier_start(&self, ch: char) -> bool;
     /// Determine if a character is a valid unquoted identifier character
     fn is_identifier_part(&self, ch: char) -> bool;
+
+    fn alllow_single_table_in_parenthesis(&self) -> bool {
+        false
+    }
```
nickolay, line 45: Please note https://github.com/ballista-compute/sqlparser-rs/issues/241#issuecomment-664377731

src/dialect/snowflake.rs
```diff
@@ -0,0 +1,26 @@
+use crate::dialect::Dialect;
+
+#[derive(Debug, Default)]
+pub struct SnowflakeDialect;
```
nickolay, line 4: I'd like the new dialect (including a separate test file), the dialect-specific parsing infrastructure, and the table factor parsing fix to be landed separately.

tests/sqlparser_snowflake.rs
```diff
@@ -0,0 +1,152 @@
...
+
+macro_rules! nest {
+    ($base:expr $(, $join:expr)*) => {
+        TableFactor::NestedJoin(Box::new(TableWithJoins {
+            relation: $base,
+            joins: vec![$(join($join)),*]
+        }))
+    };
+}
+
```
nickolay, line 40: Please don't duplicate code.

tests/sqlparser_snowflake.rs
```diff
@@ -0,0 +1,152 @@
...
+fn sf() -> TestedDialects {
+    TestedDialects {
+        dialects: vec![Box::new(SnowflakeDialect {})],
+    }
+}
```
nickolay, line 45: Please place this at the end of the file, and include a `sf_and_generic()` helper, as in other files.

tests/sqlparser_snowflake.rs
```diff
@@ -0,0 +1,152 @@
...
+
+fn get_from_section_from_select_query(query: &str) -> Vec<TableWithJoins> {
```
nickolay, line 47: I haven't been able to carefully review the rest, because the move of existing tests and the addition of new ones are combined in a single commit (see the request above).

src/dialect/mod.rs
eyalleshem: Sorry :| -  I don't understand what you mean by that ..  
(to note #241 in code comment ? or something in the code is not aligned with what you commented in [#241 (comment) ](https://github.com/ballista-compute/sqlparser-rs/issues/241#issuecomment-664377731)  )
Can you please elborate ?

src/dialect/snowflake.rs
eyalleshem: You mean by different PR's or commits ?

eyalleshem:
Thanks for your comment and for the commit with the suggestions (i also tech me some features or rust that i was not aware of  ) , agree thats more elegant - and i will merge it into the PR .  

I also asked inline for some clarifications for some off the comments , as i don't sure what the action item for them ..

src/dialect/mod.rs
nickolay: In that comment I tried to argue for using `if self.dialect <is snowflake>` checks rather than `self.dialect.alllow_single_table_in_parenthesis()`, at least until we better understand when to use which approach.

src/dialect/snowflake.rs
nickolay: I meant PRs.

eyalleshem:
Ok, so i am closing this PR - and will open 3 smaller..

PR #245, opened 2020-07-28
Minor code cleanups

nickolay:
See individual commits. @Dandandan would you take a look?

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -750,7 +750,7 @@ impl fmt::Display for Statement {
                 if_not_exists,
             } => write!(
                 f,
-                "CREATE {unique}INDEX {if_not_exists}{name} ON {table_name}({columns});",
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -793,28 +777,27 @@ impl fmt::Display for Statement {
...
-            } => {
-                f.write_str("SET ")?;
-                if *local {
-                    f.write_str("LOCAL ")?;
-                }
-                write!(f, "{} = {}", variable, value)
-            }
+            } => write!(
+                f,
+                "SET{} {} = {}",
```
Dandandan, line 782: I think it makes sense to use named arguments here `{local}` etc as well

Dandandan:
One small comment which could make it a bit more consistent, looking great!

src/ast/mod.rs
nickolay: Right. I was too focussed on the other inconsistencies that I missed how I was introducing another one myself. Thanks for catching that!

PR #246, opened 2020-07-29
Don't fail parsing ALTER TABLE ADD COLUMN ending with a semicolon

mashuai:
Fix `parse_column_def` https://github.com/ballista-compute/sqlparser-rs/issues/233

> ALTER TABLE tab ADD COLUMN foo TEXT;

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1197,7 +1197,7 @@ impl Parser {
         let mut options = vec![];
         loop {
             match self.peek_token() {
-                Token::EOF | Token::Comma | Token::RParen => break,
+                Token::EOF | Token::Comma | Token::RParen | Token::SemiColon => break,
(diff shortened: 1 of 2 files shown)


mashuai:
I have compared `CREATE TABLE t (a int);` and  `ALTER TABLE tab ADD COLUMN foo TEXT;`.  We invoke `parse_column_def` to parse CREATE statement it will break the loop when it meets `Token::RParen`, but for ALTER statement, it will not match either of them in the end. That's why I add `Token::SemiColon` to fix that.

nickolay:
> It's not obvious that this is the correct fix. How is {the set of tokens we break on in parse_column_def} determined? Why is it OK to break on irrelevant tokens? This should be noted in a comment I think.

I've opened https://github.com/ballista-compute/sqlparser-rs/pull/251 to fix that.

PR #249, opened 2020-07-29
add IndexedColumn to support SQLite create index grammar

mashuai:
ref: https://www.sqlite.org/syntax/indexed-column.html

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -139,31 +139,6 @@ impl fmt::Display for TableConstraint {
     }
 }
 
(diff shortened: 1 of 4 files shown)


nickolay:
Thanks! Any reason you prefer `indexed_columns` to simply `columns` or is it a left-over from the previous version of the PR?

mashuai:
Because I think `column-def` and `indexed-column` are different, so I use `indexed-columns` to distinguish the difference, but it can be implied to `columns` because I define it inside of `CreateIndex`.

PR #250, opened 2020-07-29
Add support for named argument functions

eyalleshem:
Add support for named argument function

This commit support functions with argument names.
the format is :
"Select some_function( a => exp, b => exp2 .. ) FROM table1
OR
"select * from table(function(a => exp)) f;"

see:
https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#named-argument-assignment-token
or example is snwflake db :
https://docs.snowflake.com/en/sql-reference/functions/flatten.html
 This commit does not belong to any branch on this repository.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -873,12 +873,28 @@ impl fmt::Display for Assignment {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum FunctionArg {
+    Named { name: Ident, arg: Expr },
+    Unnamed(Expr),
+}
+
+impl fmt::Display for FunctionArg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            FunctionArg::Named { name, arg } => write!(f, "{} => {}", name, arg),
+            FunctionArg::Unnamed(unnamed_arg) => write!(f, "{}", unnamed_arg),
+        }
+    }
+}
+
 /// A function call
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Function {
     pub name: ObjectName,
-    pub args: Vec<Expr>,
+    pub args: Vec<FunctionArg>,
     pub over: Option<WindowSpec>,
     // aggregate functions may specify eg `COUNT(DISTINCT x)`
     pub distinct: bool,
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2203,11 +2203,24 @@ impl Parser {
         Ok(Assignment { id, value })
     }
 
-    pub fn parse_optional_args(&mut self) -> Result<Vec<Expr>, ParserError> {
+    fn parse_function_args(&mut self) -> Result<FunctionArg, ParserError> {
+        if self.peek_nth_token(1) == Token::RArrow {
+            let name = self.parse_identifier()?;
+
+            self.expect_token(&Token::RArrow)?;
+            let arg = self.parse_expr()?;
+
+            Ok(FunctionArg::Named { name, arg })
+        } else {
+            Ok(FunctionArg::Unnamed(self.parse_expr()?))
+        }
+    }
+
+    pub fn parse_optional_args(&mut self) -> Result<Vec<FunctionArg>, ParserError> {
         if self.consume_token(&Token::RParen) {
             Ok(vec![])
         } else {
-            let args = self.parse_comma_separated(Parser::parse_expr)?;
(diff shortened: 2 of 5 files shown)


eyalleshem:
Sorry about some style changes that made by `cargo fmt`

nickolay:
I find it weird that you added NamedArgument as a new variant of Expr. This makes the parser accept many weird constructs, which shouldn't be accepted. I expected this to be a new SelectItem-like enum. You can use `peek_nth_token(1) == RArrow` to determine if it's a named argument.

The argument name is a simple identifier, not an ObjectName.

Please check the PR for typos (FUCNTION, rigth, agument, // vs /// (which led to rustfmt reformatting the whole enum))

> the format is : "Select * from ( a => exp, b=> exp2 .. )

Surely you meant `SELECT myfunc(args)` rather than `FROM (args)`

eyalleshem:
Sorry for all the typos , I fixed them (Hope that all of them..) 
I also changed the the object name to ident - and fixed the commit message with the correct query . 

But i don't sure what you means when you say that the named-argument should be `SelectItem` 
It's a comma-separated argument list that passed to function, In the `struct Function` the args is comma-separated `Vec<Expr>` , And because that my thought was that creating new `Expr` type will be the best suit .

nickolay:
I said a _new_ enum, a little like SelectItem (currently with two variants: unnamed argument, named argument).

eyalleshem:
I done the changes according to comments .

src/ast/mod.rs
```diff
@@ -873,12 +873,42 @@ impl fmt::Display for Assignment {
...
+    pub name: Ident,
+    pub arg: Expr,
+}
+
+impl fmt::Display for FunctionNamedArg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{} => {}", self.name, self.arg)?;
+        Ok(())
+    }
+}
```
nickolay, line 888: Is there a reason to make this a separate struct, rather than `FunctionArg::Named { name, arg }`?

_(Another option is to make the argument type a simple struct with an optional `name`, and the reason I suggested `enum` initially because of the [other possible kinds of SQL arguments](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#SQL-argument)...)_

src/ast/mod.rs
```diff
@@ -873,12 +873,42 @@ impl fmt::Display for Assignment {
...
+pub enum FunctionArg {
+    Named(FunctionNamedArg),
+    Unnamed(Expr),
+}
+
+impl fmt::Display for FunctionArg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            FunctionArg::Named(named_arg) => named_arg.fmt(f),
+            FunctionArg::Unnamed(unnamed_arg) => unnamed_arg.fmt(f),
```
nickolay, line 901: We use `write!(f, "{}", ...)` in other places for this.

src/parser.rs
```diff
@@ -2201,11 +2201,24 @@ impl Parser {
         Ok(Assignment { id, value })
     }
 
-    pub fn parse_optional_args(&mut self) -> Result<Vec<Expr>, ParserError> {
+    fn parse_function_args(&mut self) -> Result<FunctionArg, ParserError> {
+        if self.peek_nth_token(1) == Token::RArrow {
+            let name = self.parse_identifier()?;
+            // the `=>`
+            let _ = self.next_token();
```
nickolay, line 2208: Please use `expect_token` instead.

src/tokenizer.rs
```diff
@@ -323,7 +326,7 @@ impl<'a> Tokenizer<'a> {
                         }
                     }
                 }
-                // identifier or keyword
+                // identifier or Keyword
```
nickolay, line 329: Why this change?

src/ast/mod.rs
```diff
@@ -141,7 +141,7 @@ impl fmt::Display for ObjectName {
 
 /// An SQL expression of any type.
 ///
-/// The parser does not distinguish between expressions of different types
+/// The parser does not distinguish between qessions of different types
```
nickolay, line 144: Typo?

src/ast/mod.rs
eyalleshem: Fixed , 
I agree that enum is better.

src/ast/mod.rs
eyalleshem: fixed

src/parser.rs
eyalleshem: fixed

src/tokenizer.rs
eyalleshem: sorry ,  by mistake ..

src/ast/mod.rs
eyalleshem: sorry ,  by mistake ..

nickolay:
Thanks!

PR #253, opened 2020-08-02
Support table function derived table

eyalleshem:
Support `TABLE(...)` syntax in `FROM`, for example:

    select * from TABLE(SOME_FUNCTION(some_arg))

The ANSI spec allows routine invocations (and some other kinds of expressions we don't currently support) inside TABLE:
https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#PTF-derived-table
https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#table-function-derived-table

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -235,6 +235,11 @@ pub enum TableFactor {
         subquery: Box<Query>,
         alias: Option<TableAlias>,
     },
+    /// `TABLE(<expr>)[ AS <alias> ]`
+    TableFunction {
+        expr: Expr,
+        alias: Option<TableAlias>,
+    },
     /// Represents a parenthesized table factor. The SQL spec only allows a
     /// join expression (`(foo <JOIN> bar [ <JOIN> baz ... ])`) to be nested,
     /// possibly several times, but the parser also accepts the non-standard
@@ -273,6 +278,13 @@ impl fmt::Display for TableFactor {
                     write!(f, "LATERAL ")?;
                 }
                 write!(f, "({})", subquery)?;
+                if let Some(alias) = alias {
+                    write!(f, " AS {}", alias)?;
+                }
+                Ok(())
+            }
+            TableFactor::TableFunction { expr, alias } => {
+                write!(f, "TABLE({})", expr)?;
                 if let Some(alias) = alias {
                     write!(f, " AS {}", alias)?;
                 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2066,10 +2066,15 @@ impl Parser {
             if !self.consume_token(&Token::LParen) {
                 self.expected("subquery after LATERAL", self.peek_token())?;
             }
-            return self.parse_derived_table_factor(Lateral);
-        }
-
-        if self.consume_token(&Token::LParen) {
+            self.parse_derived_table_factor(Lateral)
+        } else if self.parse_keyword(Keyword::TABLE) {
+            // parse table function (SELECT * FROM TABLE (<expr>) [ AS <alias> ])
+            self.expect_token(&Token::LParen)?;
+            let expr = self.parse_expr()?;
+            self.expect_token(&Token::RParen)?;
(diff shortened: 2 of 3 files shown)


tests/sqlparser_common.rs
```diff
@@ -1864,6 +1864,39 @@ fn parse_simple_math_expr_minus() {
     verified_only_select(sql);
 }
 
+#[test]
+fn parse_table_function() {
+    let select = verified_only_select("SELECT * FROM TABLE(FUN('1')) AS a");
+
+    match only(select.from).relation {
+        TableFactor::TableFunction { expr, alias } => {
+            let excepted_exp = Expr::Function(Function {
```
nickolay, line 1873: excepted_exp -> expected_expr

src/parser.rs
```diff
@@ -2069,6 +2081,10 @@ impl Parser {
             return self.parse_derived_table_factor(Lateral);
         }
 
+        if self.parse_keyword(Keyword::TABLE) {
+            return self.parse_table_fucntion();
```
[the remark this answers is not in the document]
eyalleshem: done

tests/sqlparser_common.rs
eyalleshem: done (sorry )

src/parser.rs
```diff
@@ -2069,6 +2081,10 @@ impl Parser {
             return self.parse_derived_table_factor(Lateral);
         }
 
+        if self.parse_keyword(Keyword::TABLE) {
```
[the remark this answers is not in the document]
eyalleshem: done

src/parser.rs
```diff
@@ -2059,6 +2059,18 @@ impl Parser {
...
+    fn parse_table_fucntion(&mut self) -> Result<TableFactor, ParserError> {
+        self.expect_token(&Token::LParen)?;
+
+        let expr = self.parse_expr()?;
+
+        self.expect_token(&Token::RParen)?;
+
+        let alias = self.parse_optional_table_alias(keywords::RESERVED_FOR_TABLE_ALIAS)?;
+
+        Ok(TableFactor::TableFunction { expr, alias })
```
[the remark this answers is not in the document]
eyalleshem: done

src/ast/query.rs
```diff
@@ -278,6 +283,13 @@ impl fmt::Display for TableFactor {
                 }
                 Ok(())
             }
+            TableFactor::TableFunction { expr, alias } => {
+                write!(f, "TABLE({})", expr)?;
```
Dandandan, line 287: Wouldn't it be more consistent to add a space after `TABLE`?

nickolay: This is currently formatted like a function call and follows the formatting used in snowflake's documentation. What do you think it should it be consistent with?

src/parser.rs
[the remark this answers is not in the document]
nickolay: Please ditch the `return` above the `else`? And make the same change in the next return, so that the whole function is one large if-elseif-else expression.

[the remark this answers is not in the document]
eyalleshem: Done

src/ast/query.rs
eyalleshem: I don't care about removing this whitespace or not .. 
whats looks better for you - it's fine for me ..

Dandandan: I think the function call inside the `TABLE` should have no space, but after it it is just a parenthesized expression? So I would say, to be consistent with the docs (which also has a space after table) and other constructs a space makes sense here.

Dandandan: Ah I see the snowflake documentation now. That too makes sense if you see it like a function call... Maybe in docs for the enum, it could have the space removed too?

eyalleshem: I removed the space from the enum definition.

eyalleshem:
sorry for all the commit's - i got into some mess between my branch's , 
But fixed now  ..

PR #255, opened 2020-08-04
Dependency updates

Dandandan:
Some manual dependency updates.

We might also enable dependabot to look for some updates in the future. I'll create a PR for this.


nickolay:
I don't believe the changes to Cargo.toml have any effect on the actual versions used when testing or building the library, but I agree it's good to be more clear!

PR #256, opened 2020-08-04
Enable dependabot for this repository

Dandandan:
Adds a basic configuration for dependabot


dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: "daily"
+  directory: "/"
+  schedule:
+    interval: daily
```
nickolay, line 9: I don't think this duplication is intentional?

dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+  open-pull-requests-limit: 10
+  allow:
+  - dependency-type: direct
```
nickolay, line 12: Why is this necessary? I couldn't figure out what's the default for this from the docs, but we don't commit Cargo.lock, so there shouldn't be any difference, no?

dependabot.yml
nickolay: Though we might want to get update notifications on sqlparser_bench as well.

dependabot.yml
Dandandan: Should be no difference indeed, based on only having a `Cargo.toml` file

PR #260, opened 2020-08-12
[snowflake] Support `FROM (table_name) alias`

### LINKED ISSUE #223: Fail to parse - "select * from (table_name) alias" (Snowflake)
Related to https://github.com/ballista-compute/sqlparser-rs/issues/154

sqlparser now parses `select * from (table_name)`. However, it does not parse `select * from (table_name) alias`. 

Would it make sense to add this functionality? If so, would that mean adding an `alias` field to `TableWithJoins`?

eyalleshem:
Initially discussed in https://github.com/ballista-compute/sqlparser-rs/issues/223 and  https://github.com/ballista-compute/sqlparser-rs/pull/244. Closes #223.

The description copied from the commit:

Snowflake diverges from the standard and from most of the other implementations by allowing extra parentheses not only around a join, but around lone table names (e.g. `FROM (mytable [AS alias])`) and around derived tables (e.g. `FROM ((SELECT ...)  [AS alias])`) as well.

Initially this was implemented in https://github.com/ballista-compute/sqlparser-rs/issues/154 by (ab)using `TableFactor::NestedJoin` to represent anything nested in extra set of parens.

Afterwards we learned in https://github.com/ballista-compute/sqlparser-rs/issues/223 that in cases of such extraneous nesting Snowflake allows specifying the alias both inside and outside parens, but not both - consider:

    FROM (table_factor AS inner_alias) AS outer_alias

We've considered implementing this by changing `TableFactor::NestedJoin` to a `TableFactor::Nested { inner: TableWithJoins, alias: Option<TableAlias> }`, but that seemed too generic, as no known dialect supports duplicate aliases, as shown above, nor naming nested joins `(foo NATURAL JOIN bar) alias`. So we decided on making a smaller change (with no modifications to the AST), that is also more appropriate to the contributors to the Snowflake dialect:

1) Revert #154 by rejecting `FROM (table or derived table)` in most dialects.

2) For `dialect_of!(self is SnowflakeDialect | GenericDialect)` parse
and strip the extraneous parentheses:

   `(mytable) AS alias` -> `(mytable AS alias)`

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -18,12 +18,15 @@
 //! sqlparser regardless of the chosen dialect (i.e. it doesn't conflict with
 //! dialect-specific parsing rules).
 
-use matches::assert_matches;
+#[macro_use]
+#[path = "macros/mod.rs"]
+mod macros;
 
+use matches::assert_matches;
 use sqlparser::ast::*;
 use sqlparser::dialect::keywords::ALL_KEYWORDS;
 use sqlparser::parser::ParserError;
-use sqlparser::test_utils::{all_dialects, expr_from_projection, number, only};
+use sqlparser::test_utils::{all_dialects, expr_from_projection, join, number, only, table};
 
 #[test]
 fn parse_insert_values() {
@@ -2282,31 +2285,6 @@ fn parse_complex_join() {
 
 #[test]
 fn parse_join_nesting() {
-    fn table(name: impl Into<String>) -> TableFactor {
-        TableFactor::Table {
-            name: ObjectName(vec![Ident::new(name.into())]),
-            alias: None,
-            args: vec![],
-            with_hints: vec![],
-        }
-    }
-
-    fn join(relation: TableFactor) -> Join {
-        Join {
-            relation,
-            join_operator: JoinOperator::Inner(JoinConstraint::Natural),
-        }
-    }
-
-    macro_rules! nest {
-        ($base:expr $(, $join:expr)*) => {
-            TableFactor::NestedJoin(Box::new(TableWithJoins {
-                relation: $base,
-                joins: vec![$(join($join)),*]
-            }))
-        };
-    }
-
     let sql = "SELECT * FROM a NATURAL JOIN (b NATURAL JOIN (c NATURAL JOIN d NATURAL JOIN e)) \
                NATURAL JOIN (f NATURAL JOIN (g NATURAL JOIN h))";
     assert_eq!(
@@ -2337,20 +2315,6 @@ fn parse_join_nesting() {
         from.joins,
         vec![join(nest!(nest!(nest!(table("b"), table("c")))))]
     );
-
-    // Parenthesized table names are non-standard, but supported in Snowflake SQL
-    let sql = "SELECT * FROM (a NATURAL JOIN (b))";
-    let select = verified_only_select(sql);
-    let from = only(select.from);
-
-    assert_eq!(from.relation, nest!(table("a"), nest!(table("b"))));
-
-    // Double parentheses around table names are non-standard, but supported in Snowflake SQL
-    let sql = "SELECT * FROM (a NATURAL JOIN ((b)))";
-    let select = verified_only_select(sql);
-    let from = only(select.from);
-
-    assert_eq!(from.relation, nest!(table("a"), nest!(nest!(table("b")))));
 }
 
 #[test]
@@ -2488,26 +2452,6 @@ fn parse_derived_tables() {
                 },
                 join_operator: JoinOperator::Inner(JoinConstraint::Natural),
             }],
-        }))
-    );
-
-    // Nesting a subquery in parentheses is non-standard, but supported in Snowflake SQL
-    let sql = "SELECT * FROM ((SELECT 1) AS t)";
-    let select = verified_only_select(sql);
-    let from = only(select.from);
-
-    assert_eq!(
-        from.relation,
-        TableFactor::NestedJoin(Box::new(TableWithJoins {
-            relation: TableFactor::Derived {
-                lateral: false,
-                subquery: Box::new(verified_query("SELECT 1")),
-                alias: Some(TableAlias {
-                    name: "t".into(),
-                    columns: vec![],
-                })
-            },
-            joins: Vec::new(),
         }))
     );
 }
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -19,14 +19,13 @@
 //! dialect-specific parsing rules).
 
 #[macro_use]
-#[path = "macros/mod.rs"]
-mod macros;
+mod test_utils;
+use test_utils::{all_dialects, expr_from_projection, join, number, only, table};
 
 use matches::assert_matches;
 use sqlparser::ast::*;
 use sqlparser::dialect::keywords::ALL_KEYWORDS;
 use sqlparser::parser::ParserError;
-use sqlparser::test_utils::{all_dialects, expr_from_projection, join, number, only, table};
 
 #[test]
 fn parse_insert_values() {
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -20,7 +20,7 @@
 
 #[macro_use]
 mod test_utils;
-use test_utils::{all_dialects, expr_from_projection, join, number, only, table};
+use test_utils::{all_dialects, expr_from_projection, join, number, only, table, table_alias};
 
 use matches::assert_matches;
 use sqlparser::ast::*;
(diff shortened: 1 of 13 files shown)


tests/sqlparser_common.rs
```diff
@@ -17,6 +17,14 @@
 //! that 1) it's either standard or widely supported and 2) it can be parsed by
 //! sqlparser regardless of the chosen dialect (i.e. it doesn't conflict with
 //! dialect-specific parsing rules).
+//!
+//!
```
nickolay, line 21: Any reason for this addition?

tests/sqlparser_common.rs
```diff
@@ -17,6 +17,14 @@
...
+
+#[macro_use]
+#[path = "utils/mod.rs"]
+mod utils;
+
+use utils::*;
```
nickolay, line 27: There's `test_utils.rs` for that already.

tests/sqlparser_snowflake.rs
```diff
@@ -9,9 +9,26 @@
...
+#[path = "utils/mod.rs"]
+mod utils;
+
 use sqlparser::ast::*;
 use sqlparser::dialect::{GenericDialect, SnowflakeDialect};
+use sqlparser::parser::ParserError;
 use sqlparser::test_utils::*;
+use utils::*;
+
+fn table_alias(alias: &str) -> TableAlias {
```
nickolay, line 23: Maybe keep `table(name)` as is to avoid changes to existing tests, and add a `table_with_alias(table, alias)` instead?

tests/sqlparser_snowflake.rs
```diff
@@ -24,6 +41,120 @@ fn test_snowflake_create_table() {
     }
 }
 
+fn get_from_section_from_select_query(query: &str) -> Vec<TableWithJoins> {
```
nickolay, line 44: Can't you use `verified_only_select(sql).from` instead of creating a new helper?

tests/sqlparser_common.rs
```diff
@@ -2492,26 +2473,6 @@ fn parse_derived_tables() {
             }],
         }))
     );
-
-    // Nesting a subquery in parentheses is non-standard, but supported in Snowflake SQL
-    let sql = "SELECT * FROM ((SELECT 1) AS t)";
```
nickolay, line 2497: To reiterate, it's hard to review when you change code at the same time you move it around. I forgot about this when reviewing the PR for the new dialect, but the idea was to do the moves there.

I found the new location of the other two moved tests, but not this one.. Where was this test case moved?

tests/sqlparser_snowflake.rs
```diff
@@ -24,6 +41,120 @@ fn test_snowflake_create_table() {
...
+        }
+    );
+}
+
+#[test]
+fn test_single_table_in_parenthesis() {
+    //Parenthesized table names are non-standard, but supported in Snowflake SQL
+    let from = get_from_section_from_select_query("SELECT * FROM (a NATURAL JOIN (b))");
+
+    assert_eq!(from[0].relation, nest!(table("a", None), table("b", None)));
```
nickolay, line 82: Better use `only(select.from)`, as `from[0]` doesn't check there are no other items in the FROM.

tests/sqlparser_snowflake.rs
```diff
@@ -24,6 +41,120 @@ fn test_snowflake_create_table() {
...
+
+    let from = get_from_section_from_select_query("SELECT * FROM (a NATURAL JOIN ((b)))");
```
nickolay, line 84: This used to have a comment:
```
// Double parentheses around table names are non-standard, but supported in Snowflake SQL
```

tests/sqlparser_snowflake.rs
```diff
@@ -24,6 +41,120 @@ fn test_snowflake_create_table() {
...
+    assert_eq!(from[0].relation, nest!(table("a", None), table("b", None)));
+}
+
+#[test]
+fn test_single_table_in_parenthesis_with_alias() {
+    let sql = "SELECT * FROM (a NATURAL JOIN (b) c )";
+    let table_with_joins = get_from_section_from_select_query(sql)[0].clone();
```
nickolay, line 91: Shouldn't need the `clone`s.

tests/sqlparser_snowflake.rs
```diff
@@ -24,6 +41,120 @@ fn test_snowflake_create_table() {
...
+    let res = snowflake().parse_sql_statements("SELECT * FROM (a b) c");
+    assert_eq!(
+        ParserError::ParserError("duplicate alias b".to_string()),
+        res.unwrap_err()
+    );
+}
+
+fn snowflake() -> TestedDialects {
+    TestedDialects {
+        // we don't have a separate SQLite dialect, so test only the generic dialect for now
```
nickolay, line 153: Please remove.

tests/sqlparser_common.rs
eyalleshem: sorry , removed

tests/sqlparser_common.rs
eyalleshem: The actual reason that i don't use the test util is that i want to share the macro between the test , 
And if I understand it correct , the `src` and the the `test` are not considered in the same crate , so i can export macro only via `#[macro_export]` directive, but that mean the macro will be available not only  for the test but to all users of the lib , and that something that i want to avoid (please correct me if i wrong..). 
Anyway - I moved the join and table macros to `test_utils.rs` and change the `utils` dir that i crated to `macros` dir, hope you ok with that. 
btw - why `test_util.rs` located inside the source ?

tests/sqlparser_snowflake.rs
eyalleshem: Done

tests/sqlparser_common.rs
eyalleshem: Sorry about that , i also comment in the [previous pr](https://github.com/ballista-compute/sqlparser-rs/pull/259#issuecomment-671435447) that i don't sure what needed to be in the test (now i am understand :) ) . 

Anyway added also this test ..

tests/sqlparser_snowflake.rs
eyalleshem: Done

tests/sqlparser_snowflake.rs
eyalleshem: It's a problem to use it because we remove the information about [ we ignore the information about  the redundant parenthesis in the AST ](https://github.com/ballista-compute/sqlparser-rs/issues/223#issuecomment-666334528)  , so the input sql will   be different than the output .

tests/sqlparser_snowflake.rs
eyalleshem: This comment relevant to both test in the function , so i put here as the first line in the function , and then blank line , so it's relate to the function and not specific line ..

tests/sqlparser_snowflake.rs
eyalleshem: done

tests/sqlparser_snowflake.rs
eyalleshem: done , sorry - a  rebase issue

nickolay:
I first rebased this onto `main` without changes, then split into two commits (moving things around and implementing the new functionality), as I wanted to do all along...

nickolay:
Regarding the need to share helper macros, as well as helper functions, discussed in https://github.com/ballista-compute/sqlparser-rs/pull/260#discussion_r470006301 -- I added a commit https://github.com/ballista-compute/sqlparser-rs/pull/260/commits/a3a0d7b1de791091786a52ebfe2dec9654f9cc66 that demonstrates how I'd like to solve this. I added some comments that I hope will explain my reasoning, please let me know if they're not clear.

The second part of the PR I want to look over again before merging.

eyalsatori:
Thanks a lot for the rebasing and the fixes!  
The solution of the test_utils indeed much more elegant . 

Is there any change needed from me ,  or you just want to continue review it?

nickolay:
No, I was just running out of time that night, and wanted to share the part I believed to be ready. I'll try to get to the other one sometime "soon"...

nickolay:
> > Can't you use verified_only_select(sql).from instead of creating a new helper?
>
> It's a problem to use it because we remove the information about we ignore the information about the redundant parenthesis in the AST , so the input sql will be different than the output .

I don't know why it took me so long to realize it, but we use `one_statement_parses_to` in cases like this. I added commits that do that to this PR and if you're OK with this, I'm ready to merge.

eyalsatori:
I am ok with that , 
Thanks a lot :)

nickolay (approved):

waitingkuo:
@eyalleshem  @nickolay i made a pull request #551 for nested table with alias

PR #263, opened 2020-08-14
add a note about cargo release config

maxcountryman:


Cargo.toml
```diff
@@ -36,4 +36,6 @@ simple_logger = "1.6"
 matches = "0.1"
 
 [package.metadata.release]
+# We want to ensure we don't publish via `cargo release` since Actions
+# handles this for us.
 disable-publish = true
```
nickolay, line 41: Thanks. May I suggest:
```
# Instruct `cargo release` to not run `cargo publish` locally:
# https://github.com/sunng87/cargo-release/blob/master/docs/reference.md#config-fields
# See docs/releasing.md for details.
```

nickolay:
BTW it appears that `cargo-release` by default pushes the tags as part of its process, and defaults to `origin` as the remote. Should we suggest `--skip-push` (or `--push-remote` to specify the actual remote name, which is `upstream` in my case)?

maxcountryman:
I didn't explicitly call that out (I should have) but I assumed that that would be a desirable side effect. I wouldn't suggest having it not push tags because then we might accidentally push the wrong tags if we do so manually. So there is a benefit to encapsulating that responsibility. However it could be surprising that running the command will in fact create a Crates release.

nickolay:
The problem is that the [docs](https://github.com/ballista-compute/sqlparser-rs/blob/9351efb437f27d2c36fa4f1582ff9a1e01d50e15/docs/releasing.md) make it seem that pushing is a separate step.

Frankly I don't see the problem of "accidentally push[ing] the wrong tags", **but** as cargo-release defaults to doing everything in one step, I think it's fine to do so --- as long as we don't have contradicting documentation.

Also, in my opinion, if we decide we want to release in one command, it shouldn't be documented as `cargo release minor --skip-publish`, as it does **not** skip publishing. The reader shouldn't be expected to learn all the intricacies of the implementation in order not to be surprised.

maxcountryman:
It's pretty easy to push local tags that you may not mean to push. That might not be desirable and the sub command should help avoid that.

Feel free to make any edits you see fit. 🙂

nickolay:
> Feel free to make any edits you see fit.

I had a go at it, what do you think?

PR #267, opened 2020-08-30
added parsing for PostgreSQL operations

alex-dukhno:
Added support to parse mathematical operations for PostgreSQL
[See table 9.4](https://www.postgresql.org/docs/12/functions-math.html)

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -191,7 +191,11 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// Unary operation e.g. `NOT foo`
-    UnaryOp { op: UnaryOperator, expr: Box<Expr> },
+    UnaryOp {
+        op: UnaryOperator,
+        expr: Box<Expr>,
+        infix: bool,
+    },
     /// CAST an expression to a different data type e.g. `CAST(foo AS VARCHAR(123))`
     Cast {
         expr: Box<Expr>,
@@ -282,7 +286,13 @@ impl fmt::Display for Expr {
                 high
             ),
             Expr::BinaryOp { left, op, right } => write!(f, "{} {} {}", left, op, right),
-            Expr::UnaryOp { op, expr } => write!(f, "{} {}", op, expr),
+            Expr::UnaryOp { op, expr, infix } => {
+                if *infix {
+                    write!(f, "{}{}", expr, op)
+                } else {
+                    write!(f, "{} {}", op, expr)
+                }
+            }
             Expr::Cast { expr, data_type } => write!(f, "CAST({} AS {})", expr, data_type),
             Expr::Extract { field, expr } => write!(f, "EXTRACT({} FROM {})", field, expr),
             Expr::Collate { expr, collation } => write!(f, "{} COLLATE {}", expr, collation),
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -191,11 +191,7 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// Unary operation e.g. `NOT foo`
-    UnaryOp {
-        op: UnaryOperator,
-        expr: Box<Expr>,
-        infix: bool,
-    },
+    UnaryOp { op: UnaryOperator, expr: Box<Expr> },
     /// CAST an expression to a different data type e.g. `CAST(foo AS VARCHAR(123))`
     Cast {
         expr: Box<Expr>,
@@ -286,8 +282,8 @@ impl fmt::Display for Expr {
                 high
             ),
             Expr::BinaryOp { left, op, right } => write!(f, "{} {} {}", left, op, right),
-            Expr::UnaryOp { op, expr, infix } => {
-                if *infix {
+            Expr::UnaryOp { op, expr } => {
+                if op == &UnaryOperator::PGPostfixFactorial {
                     write!(f, "{}{}", expr, op)
                 } else {
                     write!(f, "{} {}", op, expr)
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -21,11 +21,17 @@ pub enum UnaryOperator {
     Plus,
     Minus,
     Not,
+    /// Bitwise Not, e.g. `~9` (PostgreSQL-specific)
     PGBitwiseNot,
-    PGSqrt,
-    PGCbrt,
-    PGFactorial,
-    PGInfixFactorial,
+    /// Square root, e.g. `|/9` (PostgreSQL-specific)
+    PGSquareRoot,
+    /// Cube root, e.g. `||/27` (PostgreSQL-specific)
+    PGCubeRoot,
+    /// Factorial, e.g. `9!` (PostgreSQL-specific)
+    PGPostfixFactorial,
+    /// Factorial, e.g. `!!9` (PostgreSQL-specific)
+    PGPrefixFactorial,
+    /// Absolute value, e.g. `@ -9` (PostgreSQL-specific)
     PGAbs,
 }
 
@@ -36,10 +42,10 @@ impl fmt::Display for UnaryOperator {
             UnaryOperator::Minus => "-",
             UnaryOperator::Not => "NOT",
             UnaryOperator::PGBitwiseNot => "~",
-            UnaryOperator::PGSqrt => "|/",
-            UnaryOperator::PGCbrt => "||/",
-            UnaryOperator::PGFactorial => "!",
-            UnaryOperator::PGInfixFactorial => "!!",
+            UnaryOperator::PGSquareRoot => "|/",
+            UnaryOperator::PGCubeRoot => "||/",
+            UnaryOperator::PGPostfixFactorial => "!",
+            UnaryOperator::PGPrefixFactorial => "!!",
             UnaryOperator::PGAbs => "@",
         })
     }
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -21,6 +21,12 @@ pub enum UnaryOperator {
     Plus,
     Minus,
     Not,
+    PGBitwiseNot,
+    PGSqrt,
+    PGCbrt,
+    PGFactorial,
+    PGInfixFactorial,
+    PGAbs,
 }
 
 impl fmt::Display for UnaryOperator {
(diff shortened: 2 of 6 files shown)


alex-dukhno:
@nickolay can you please take a look?

nickolay:
This seems good. Could you fix the conflicts with the recent merge, and use the `dialect_of!` macro to limit the new operations to the PostgreSQL dialect?

alex-dukhno:
@nickolay can you have a look? 
thank you

src/tokenizer.rs
```diff
@@ -406,7 +434,18 @@ impl<'a> Tokenizer<'a> {
                 '|' => {
                     chars.next(); // consume the '|'
                     match chars.peek() {
-                        Some('|') => self.consume_and_return(chars, Token::StringConcat),
+                        Some('/') if dialect_of!(self is PostgreSqlDialect) => {
```
alex-dukhno, line 437: @nickolay, am I right that this is not going to work with custom dialect as you suggest here https://github.com/ballista-compute/sqlparser-rs/pull/243#discussion_r461094779?

src/ast/mod.rs
```diff
@@ -191,7 +191,11 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// Unary operation e.g. `NOT foo`
-    UnaryOp { op: UnaryOperator, expr: Box<Expr> },
+    UnaryOp {
+        op: UnaryOperator,
+        expr: Box<Expr>,
+        infix: bool,
```
nickolay, line 197: It seems that `infix` can be deduced from `op`. Do we need to store it?

src/ast/operator.rs
```diff
@@ -21,6 +21,12 @@ pub enum UnaryOperator {
     Plus,
     Minus,
     Not,
+    PGBitwiseNot,
+    PGSqrt,
+    PGCbrt,
```
nickolay, line 26: I think adding doc comments for the non-trivial operators is useful, as they are easily available on hover in many IDEs. E.g.
```
Cube root, e.g. `||/ 9` (PostgreSQL-specific)
```

src/tokenizer.rs
```diff
@@ -102,6 +103,24 @@ pub enum Token {
     RBrace,
     /// Right Arrow `=>`
     RArrow,
+    /// Sharp `#` use for PostgreSQL Bitwise XOR operator
```
nickolay, line 106: "used"?

src/tokenizer.rs
```diff
@@ -102,6 +103,24 @@ pub enum Token {
...
+    Sharp,
+    /// Tilde `~` use for PostgreSQL Bitwise NOT operator
+    Tilde,
+    /// Bitwise left operator `<<` use for PostgreSQL
```
nickolay, line 110:
```
`<<`, a bitwise shift left operator in PostgreSQL
```

src/tokenizer.rs
```diff
@@ -102,6 +103,24 @@ pub enum Token {
...
+    ShiftLeft,
+    /// Bitwise right operator `>>` use for PostgreSQL
+    ShiftRight,
+    /// Exclamation Mark `!` use for PostgreSQL factorial operator
+    ExclamationMark,
+    /// Exclamation Mark `!!` use for PostgreSQL prefix factorial operator
+    DoubleExclamationMark,
+    /// Ampersat `@` use for PostgreSQL abs operator
+    Ampersat,
```
nickolay, line 119: It's commonly known as the "At sign". I had to Google it to make sure it wasn't a typo, and wikipedia says this term has not achieved wide use. Use `AtSign`?

src/tokenizer.rs
```diff
@@ -102,6 +103,24 @@ pub enum Token {
...
+    /// PostgreSQL square root math operator
+    SquareRoot,
+    /// PostgreSQL cube root math operator
+    CubeRoot,
```
nickolay, line 123: Please mention that these are `|/` and `||/` respectively. Consider using the same names as for the UnaryOperator. I'm in favor of PGSquareRoot and PGCubeRoot, as these are probably very specific to postgres, and CubeRoot is more obvious than "cbrt".

src/parser.rs
```diff
@@ -283,6 +284,31 @@ impl<'a> Parser<'a> {
...
+                infix: false,
+            }),
+            Token::SquareRoot => Ok(Expr::UnaryOp {
+                op: UnaryOperator::PGSqrt,
+                expr: Box::new(self.parse_subexpr(0)?),
+                infix: false,
+            }),
+            Token::CubeRoot => Ok(Expr::UnaryOp {
+                op: UnaryOperator::PGCbrt,
+                expr: Box::new(self.parse_subexpr(0)?),
```
nickolay, line 304: This is probably not right, as this parses `WHERE ||/27 = 3` as `||/ ( 27=3 )`, which is not a valid expression. I don't see precedence mentioned in the documentation you linked to, and I think a better guess would be to default all the unary operators to the `PLUS_MINUS_PREC` (moving the new ops to the `tok @` case below).

src/parser.rs
```diff
@@ -658,6 +685,9 @@ impl<'a> Parser<'a> {
             Token::Caret => Some(BinaryOperator::BitwiseXor),
             Token::Ampersand => Some(BinaryOperator::BitwiseAnd),
             Token::Div => Some(BinaryOperator::Divide),
+            Token::ShiftLeft => Some(BinaryOperator::PGBitwiseShiftLeft),
+            Token::ShiftRight => Some(BinaryOperator::PGBitwiseShiftRight),
+            Token::Sharp => Some(BinaryOperator::PGBitwiseXor),
```
nickolay, line 690: I think the natural place for the `dialect_of!` checks is here, not in the tokenizer.

src/tokenizer.rs
```diff
@@ -958,6 +1014,65 @@ mod tests {
         compare(expected, tokens);
     }
 
+    #[test]
+    fn tokenize_postgresql_bitwise_operations() {
```
nickolay, line 1018: I think it's usually enough to test the parser. Is there any point to the tokenizer-specific tests?

tests/sqlparser_postgres.rs
```diff
@@ -551,6 +551,83 @@ fn parse_prepare() {
...
+                right: Box::new(Expr::Identifier(Ident::new("b"))),
+            }),
+            select.projection[0]
+        );
+    }
+}
+
+#[test]
+fn parse_pg_bitwise_unary_ops() {
+    let bitwise_ops = &[("~", UnaryOperator::PGBitwiseNot)];
```
nickolay, line 577: I believe this can be merged with the following test?

tests/sqlparser_postgres.rs
```diff
@@ -551,6 +551,83 @@ fn parse_prepare() {
...
+                infix: false,
+            }),
+            select.projection[0]
+        );
+    }
+}
+
+#[test]
+fn parse_pg_math_ops() {
+    let bitwise_ops = &[("!", UnaryOperator::PGFactorial)];
```
nickolay, line 616: `parse_pg_postfix_factorial` seems more appropriate here.

nickolay (changes requested):
I'm sorry for taking so long!

src/tokenizer.rs
alex-dukhno: I have custom `Dialect` https://github.com/alex-dukhno/database/blob/master/src/parser/src/lib.rs#L221-L231, is it true that it wouldn't parse `|/` into `SquareRoot` because of `if dialect_of!(self is PostgreSqlDialect)`?

nickolay: Ah, I see now. Yes, I assumed `|/` is something very Postgres-specific, so limiting it to PostgreSqlDialect _(and GenericDialect, which I didn't mention explicitly)_ seemed to make sense.

In the older comment you're referring to I described a stop-gap measure that would be used until someone figured out the proper way to deal with `$...` in the parser.

We could make the new operators parse unconditionally for now if that makes your life easier, but without a clear use-case (and accompanying testcase) this will likely break in the future.

src/parser.rs
```diff
@@ -283,11 +283,30 @@ impl<'a> Parser<'a> {
...
+            | tok @ Token::PGSquareRoot
+            | tok @ Token::PGCubeRoot
+            | tok @ Token::AtSign
+            | tok @ Token::Tilde => {
+                let op = match tok {
+                    Token::Plus => UnaryOperator::Plus,
+                    Token::Minus => UnaryOperator::Minus,
+                    Token::DoubleExclamationMark if dialect_of!(self is PostgreSqlDialect) => {
+                        UnaryOperator::PGPrefixFactorial
+                    }
```
nickolay, line 298: If we keep the dialect-specific parsing, it probably would be clearer to handle +/- in a separate case from the PG-specific stuff. That way you would be able to specify the `if dialect_of` only once instead of repeating it for every operator.

src/parser.rs
```diff
@@ -707,6 +735,12 @@ impl<'a> Parser<'a> {
             }
         } else if Token::DoubleColon == tok {
             self.parse_pg_cast(expr)
+        } else if Token::ExclamationMark == tok {
+            // PostgreSQL factorial operation
+            Ok(Expr::UnaryOp {
```
nickolay, line 740: Ideally we'd consider how postfix factorial interacted with other operators (the postfix `::type`, another factorial, and prefix operators), but I don't think this has to block merging this PR.

nickolay (approved):
I'm ready to merge this, waiting for your decision on removing the dialect-specific parsing.

src/tokenizer.rs
alex-dukhno: I think it's better to have `dialect_of` checks and have a solution for dealing with `$...`. Having my knowledge of crate I'd suggest to add `Variable(Ident)` or something similar to `ast::Expr` enum. It has to be done in a separate PR of course. WDYT?

nickolay: I have @-mentioned you in #265, where this discussion can continue.

The hard part is describing what specifically we're trying to implement and why. Your suggestion, for example, solves the problem of representing `$` (but not another leader) followed by an identifier _(quoted or not; consisting of characters from an unspecified set)_ anywhere an arbitrary expression is allowed _(but not in more restricted contexts, like `OFFSET __ ROWS`)._ I'm not sure if this matches Postgres or the "common denominator" of the usual dialects; answering that requires some research. From the examples I've seen so far, a new `Value` variant with `{leader, ident}` fields might be more appropriate.

src/parser.rs
nickolay: Were you planning to address this and maybe adding GenericDialect to the list of dialects that support the new operations? Or can I merge this as is?

alex-dukhno: addressed

nickolay:
Excellent, thank you!

PR #276, opened 2020-10-01
create drop schema conditionally

alex-dukhno:
added `CREATE SCHEMA IF NOT EXISTS ...` for PostgreSQL
added test for `DROP SCHEMA IF EXISTS ...` as it was supported but no test for the case

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -558,7 +558,10 @@ pub enum Statement {
     /// `ROLLBACK [ TRANSACTION | WORK ] [ AND [ NO ] CHAIN ]`
     Rollback { chain: bool },
     /// CREATE SCHEMA
-    CreateSchema { schema_name: ObjectName },
+    CreateSchema {
+        schema_name: ObjectName,
+        if_not_exists: bool,
+    },
     /// `ASSERT <condition> [AS <message>]`
     Assert {
         condition: Expr,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -829,7 +832,19 @@ impl fmt::Display for Statement {
             Statement::Rollback { chain } => {
                 write!(f, "ROLLBACK{}", if *chain { " AND CHAIN" } else { "" },)
             }
-            Statement::CreateSchema { schema_name } => write!(f, "CREATE SCHEMA {}", schema_name),
+            Statement::CreateSchema {
+                schema_name,
+                if_not_exists,
+            } => write!(
+                f,
+                "CREATE SCHEMA{if_not_exists}{name}",
```
nickolay, line 840: Using `"CREATE SCHEMA {if_not_exists}{name}"` and an empty string for `if_not_exists` when it needs to be omitted both is consistent with other implementations, and results in more compact code.

tests/sqlparser_postgres.rs
```diff
@@ -300,6 +300,33 @@ fn parse_bad_if_not_exists() {
     );
 }
 
+#[test]
+fn parse_create_schema_if_not_exists() {
+    let sql = "CREATE SCHEMA IF NOT EXISTS schema_name";
+    let ast = pg().verified_stmt(sql);
```
nickolay, line 306: Please default to `pg_and_generic()` unless testing a construct that can't be parsed by the generic parser. The idea of the generic parser is to be able to parse as much SQL code as possible.

nickolay (approved):
Thanks! See my two comments below.

This continues from https://github.com/ballista-compute/sqlparser-rs/pull/173 and I'll copy the spec links from that PR: [SQL:2016](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#schema-name-clause) and [Posgres](https://www.postgresql.org/docs/12/sql-createschema.html). The `IF NOT EXISTS` is a Postgres-specific clause.

PR #278, opened 2020-10-11
Add Support for Recursive CTEs

rhanqtl:
See https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#with-clause

Fixes https://github.com/ballista-compute/sqlparser-rs/issues/277

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -20,7 +20,7 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Query {
     /// WITH (common table expressions, or CTEs)
-    pub ctes: Vec<Cte>,
+    pub with: Option<With>,
     /// SELECT or UNION / EXCEPT / INTECEPT
     pub body: SetExpr,
     /// ORDER BY
@@ -35,8 +35,8 @@ pub struct Query {
 
 impl fmt::Display for Query {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if !self.ctes.is_empty() {
-            write!(f, "WITH {} ", display_comma_separated(&self.ctes))?;
+        if self.with.is_some() {
+            write!(f, "{}", self.with.as_ref().unwrap())?;
         }
         write!(f, "{}", self.body)?;
         if !self.order_by.is_empty() {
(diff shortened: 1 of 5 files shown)


rhanqtl:
```
error: match expression looks like `matches!` macro
   --> src/tokenizer.rs:385:60
    |
385 |                       let s = peeking_take_while(chars, |ch| match ch {
    |  ____________________________________________________________^
386 | |                         '0'..='9' | '.' => true,
387 | |                         _ => false,
388 | |                     });
    | |_____________________^ help: try this: `matches!(ch, '0'..='9' | '.')`
    |
    = note: `-D clippy::match-like-matches-macro` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro
```

src/ast/query.rs
```diff
@@ -35,8 +35,8 @@ pub struct Query {
 
 impl fmt::Display for Query {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if !self.ctes.is_empty() {
-            write!(f, "WITH {} ", display_comma_separated(&self.ctes))?;
+        if self.with.is_some() {
+            write!(f, "{}", self.with.as_ref().unwrap())?;
```
nickolay, line 39: `if let Some(ref with) = self.with`? Also: I think it makes more sense to print the space between the CTEs and the main query here, as part of `Query` formatting, rather than trailing make `With` end with a trailing space.

src/ast/query.rs
```diff
@@ -157,6 +157,29 @@ impl fmt::Display for Select {
...
+            write!(f, "RECURSIVE ")?;
+        }
+        let s = self
+            .cte_tables
+            .iter()
+            .map(|cte| format!("{}", cte))
+            .collect::<Vec<String>>()
+            .join(", ");
+        write!(f, "{} ", s)
+    }
```
nickolay, line 180:
```
write!(f, "WITH {}{} ",
    if self.recursive { "RECURSIVE " } else { "" },
    display_comma_separated(&self.cte_tables)
)
```

src/parser.rs
```diff
@@ -1795,11 +1795,14 @@ impl<'a> Parser<'a> {
     /// by `ORDER BY`. Unlike some other parse_... methods, this one doesn't
     /// expect the initial keyword to be already consumed
     pub fn parse_query(&mut self) -> Result<Query, ParserError> {
-        let ctes = if self.parse_keyword(Keyword::WITH) {
+        let with = if self.parse_keyword(Keyword::WITH) {
             // TODO: optional RECURSIVE
```
nickolay, line 1799: This TODO should be removed now.

nickolay (changes requested):
Thanks, this PR looks great, apart from a couple of nits below.

Regarding the clippy error, it's probably a lint enabled in a new clippy version? Would you like to fix it while you're at it?

nickolay (approved):

nickolay:
Thanks again!

PR #281, opened 2020-11-06
feat: support sqlite `insert or` statement

zhangli-pear:
https://www.sqlite.org/lang_insert.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -486,6 +486,8 @@ pub enum Statement {
     Query(Box<Query>),
     /// INSERT
     Insert {
+        /// Only for Sqlite
+        or: Option<SqliteOnConflict>,
         /// TABLE
         table_name: ObjectName,
         /// COLUMNS
@@ -782,6 +784,7 @@ impl fmt::Display for Statement {
                 Ok(())
             }
             Statement::Insert {
+                or,
                 table_name,
                 overwrite,
                 partitioned,
@@ -790,13 +793,17 @@ impl fmt::Display for Statement {
                 source,
                 table,
             } => {
-                write!(
-                    f,
-                    "INSERT {act}{tbl} {table_name} ",
-                    table_name = table_name,
-                    act = if *overwrite { "OVERWRITE" } else { "INTO" },
-                    tbl = if *table { " TABLE" } else { "" }
-                )?;
+                if let Some(action) = or {
+                    write!(f, "INSERT OR {} INTO {} ", action, table_name)?;
+                } else {
+                    write!(
+                        f,
+                        "INSERT {act}{tbl} {table_name} ",
+                        table_name = table_name,
+                        act = if *overwrite { "OVERWRITE" } else { "INTO" },
+                        tbl = if *table { " TABLE" } else { "" }
+                    )?;
+                }
                 if !columns.is_empty() {
                     write!(f, "({}) ", display_comma_separated(columns))?;
                 }
@@ -810,6 +817,7 @@ impl fmt::Display for Statement {
                 }
                 write!(f, "{}", source)
             }
+
             Statement::Copy {
                 table_name,
                 columns,
@@ -1529,6 +1537,32 @@ impl fmt::Display for SetVariableValue {
         match self {
             Ident(ident) => write!(f, "{}", ident),
             Literal(literal) => write!(f, "{}", literal),
+        }
+    }
+}
+
+/// Sqlite specific syntax
+///
+/// https://sqlite.org/lang_conflict.html
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum SqliteOnConflict {
+    Rollback,
+    Abort,
+    Fail,
+    Ignore,
+    Replace,
+}
+
+impl fmt::Display for SqliteOnConflict {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use SqliteOnConflict::*;
+        match self {
+            Rollback => write!(f, "ROLLBACK"),
+            Abort => write!(f, "ABORT"),
+            Fail => write!(f, "FAIL"),
+            Ignore => write!(f, "IGNORE"),
+            Replace => write!(f, "REPLACE"),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -435,6 +435,7 @@ pub enum Statement {
     Query(Box<Query>),
     /// INSERT
     Insert {
+        or: Option<crate::dialect::keywords::Keyword>,
```
Dandandan, line 438: I would encode this as a new struct (named OrAction) which explicitly lists the actions allowed.

src/ast/mod.rs
```diff
@@ -526,6 +527,14 @@ pub enum Statement {
         /// `RESTRICT` or no drop behavior at all was specified.
         cascade: bool,
     },
+    Replace {
+        /// TABLE
```
Dandandan, line 531: It would be better to have this encoded in a ReplaceAction which is part of the OrAction.
Replace itself is no statement!

Dandandan:
Thanks for the contribution! I left a few comments :)

src/ast/mod.rs
nickolay: Thanks for bringing this up, Dan. Judging from https://sqlite.org/lang_conflict.html the more appropriate name would be something like `SqliteConflictResolutionAlgorithm`, as these values can appear in many contexts, not just after `INSERT OR`.

src/ast/mod.rs
nickolay: I guess this implements this bit: _For compatibility with MySQL, the parser allows the use of the single keyword REPLACE as an alias for "INSERT OR REPLACE"._ I'm not sure it's appropriate to introduce a duplicate of Statement::Insert just to implement another syntax. Perhaps normalize it to `INSERT OR REPLACE` on parsing? Change `Statement::Insert::or` to a tristate-enum: None, Replace, Or(SqliteConflictResolutionAlgorithm)?

src/ast/mod.rs
nickolay: Oh and this needs a comment indicating this implements an sqlite-specific`INSERT OR action` to specify a constraint conflict resolution method.

src/ast/mod.rs
zhangli-pear: I did try some of these methods.Finally, I think a `replace` statement is more reasonable. Because the implementation is much easier, on the other hand, one statement to support one syntax is straightforward. I don't think we should care whether it is an alias for `insert` during parsing the sql. And saying about normalize `replace` to `insert or replace`, we will lost the original details, I'm not sure is it necessary to hold the detail. If we do so, we can't build a `replace` sql from ast any longer.

src/ast/mod.rs
zhangli-pear: `SqliteConflictResolutionAlgorithm` is really a long name, how about `OnConflictAction`  or `OnconflictAlgorithm`?

nickolay: The "Sqlite" bit is pretty important part of the name, as `ON CONFLICT` exists in other dialects, and has very little in common with sqlite's. SqliteOnConflict is fine with me if you really want to make it shorter.

src/ast/mod.rs
nickolay: Our AST is lossy in many aspects, so unless someone _needs_ to be able to distinguish REPLACE from INSERT OR REPLACE, I'd rather we didn't add a separate variant. But I can live with either solution.

zhangli-pear:
I’v changed code following discussion above, please review again.BTW, how long does it take to publish a new version to crates.io?

zhangli-pear:
Is there anybody still following this thread?

Dandandan:
Thanks @zhangli-pear for building this feature.
Sorry for waiting so long to have a look at it.

src/ast/mod.rs
```diff
@@ -1155,3 +1163,29 @@ impl fmt::Display for SetVariableValue {
         }
     }
 }
+
+/// Sqlite specific syntax
+///
+/// https://sqlite.org/lang_conflict.html
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum SqliteOnConfilict {
```
Dandandan, line 1172: typo `Confilict` `->` `Conflict`

Dandandan (approved):

Dandandan:
LGTM now @nickolay do you maybe have any comments?

src/parser.rs
```diff
@@ -2272,11 +2276,32 @@ impl<'a> Parser<'a> {
...
+                Some(SqliteOnConflict::Replace)
+            } else if self.parse_keywords(&[Keyword::OR, Keyword::ROLLBACK]) {
+                Some(SqliteOnConflict::Rollback)
+            } else if self.parse_keywords(&[Keyword::OR, Keyword::ABORT]) {
+                Some(SqliteOnConflict::Abort)
+            } else if self.parse_keywords(&[Keyword::OR, Keyword::FAIL]) {
+                Some(SqliteOnConflict::Fail)
+            } else if self.parse_keywords(&[Keyword::OR, Keyword::IGNORE]) {
+                Some(SqliteOnConflict::Ignore)
+            } else if self.parse_keyword(Keyword::REPLACE) {
```
Dandandan, line 2290: I think this is not right? looking at https://sqlite.org/lang_insert.html it should be combined always with `OR`?

src/parser.rs
```diff
@@ -2272,11 +2276,32 @@ impl<'a> Parser<'a> {
 
     /// Parse an INSERT statement
     pub fn parse_insert(&mut self) -> Result<Statement, ParserError> {
+        let or = if dialect_of!(self is SQLiteDialect) {
```
Dandandan, line 2279: This can be reverted / simplified with early return so it doesn't need double nesting, something like:
```rust
if ! dialect_of!(self is SQLiteDialect)  {
   None
...

```

src/parser.rs
zhangli-pear: `REPLACE INTO table ...` is the same as `INSERT OR REPLACE INTO table ...`. 

> For compatibility with MySQL, the parser allows the use of the single keyword REPLACE as an alias for "INSERT OR REPLACE".

Dandandan:
@zhangli-pear can you rebase against the main branch?

zhangli-pear:
Done. I didn't notice the conflict before😂.

Dandandan:
Thanks @zhangli-pear for the changes :+1:

PR #284, opened 2020-12-22
feat: Introduce support for EXPLAIN/ANALYZE

ovr:
Hello!

Thanks

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -431,6 +431,12 @@ impl fmt::Display for WindowFrameBound {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Statement {
+    //
+    Explain {
+        analyze: bool,
+        /// A SQL query that specifies what to explain
+        query: Box<Query>,
+    },
     /// SELECT
     Query(Box<Query>),
     /// INSERT
@@ -591,6 +597,15 @@ impl fmt::Display for Statement {
     #[allow(clippy::cognitive_complexity)]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            Statement::Explain { analyze, query } => {
+                write!(f, "EXPLAIN ")?;
+
+                if *analyze {
+                    write!(f, "ANALYZE ")?;
+                }
+
+                write!(f, "{}", query)
+            }
             Statement::Query(s) => write!(f, "{}", s),
             Statement::Insert {
                 table_name,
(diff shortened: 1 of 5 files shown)


ovr:
Hello @Dandandan @nickolay 

Can you please take a look?

Thanks!

src/parser.rs
```diff
@@ -1790,6 +1791,13 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_explain(&mut self) -> Result<Statement, ParserError> {
+        let analyze = self.parse_keyword(Keyword::ANALYZE);
+        let query = Box::new(self.parse_query()?);
```
Dandandan, line 1796: It looks like most databases support a statement in explain (analyze) instead of "only" a query?

src/ast/mod.rs
```diff
@@ -431,6 +431,12 @@ impl fmt::Display for WindowFrameBound {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Statement {
+    //
```
Dandandan, line 434: I think you want to add something like `/// EXPLAIN`  here?

Dandandan:
Thanks for your PR @ovr ! Looking good. I left some review comments

src/parser.rs
ovr: Goot catch! Thanks

src/ast/mod.rs
ovr: Added, thanks

ovr:
@Dandandan, Thank for your comments, I've updated PR. Added `ExplainLevel` enum to support VERBOSE/ANALYZE levels.

src/ast/mod.rs
```diff
@@ -426,11 +426,34 @@ impl fmt::Display for WindowFrameBound {
...
+}
+
 /// A top-level statement (SELECT, INSERT, CREATE, etc.)
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Statement {
+    // EXPLAIN
+    Explain {
+        level: Option<ExplainLevel>,
```
Dandandan, line 453: I think it's not an enum but rather a combination of options.
see: https://www.postgresql.org/docs/9.1/sql-explain.html

I think all the following are correct:

`EXPLAIN`
`EXPLAIN ANALYZE`
`EXPLAIN VERBOSE`
`EXPLAIN ANALYZE VERBOSE`

So maybe this would be a better encoding?
```rust
Explain {
    analyze: bool
    verbose: bool
}
```

ovr: Yeap, you are right.

I checked What combintations work with PostgreSQL:

```
EXPLAIN VERBOSE SELECT * from test;
EXPLAIN ANALYSE SELECT * from test;
EXPLAIN ANALYZE VERBOSE SELECT * from test;
```

What doesnt work:

```
EXPLAIN VERBOSE ANALYZE SELECT * from test;
```

Find very funny thing, that they support `ANALYSE`, but it's not mentioned inside docs.

Dandandan:
Thanks a lot @ovr for your contribution!

PR #285, opened 2020-12-28
Support ANALYZE TABLE syntax

Dandandan:
Also FYI @andygrove I want to add this for DataFusion.

There are quite some different syntax versions for this, so for now I think it is best to stick to `ANALYZE TABLE [table_name]` and support further syntax later on. 

https://www.postgresql.org/docs/9.0/sql-analyze.html
https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-aux-analyze-table.html
https://dev.mysql.com/doc/refman/5.6/en/analyze-table.html


andygrove (approved):
LGTM

PR #286, opened 2021-01-02
Add support for no_std + alloc

joshwd36:
I'm looking to use this in a no_std + alloc environment and so it would be really helpful to allow that. I've implemented this by having a `std` feature that is enabled by default, so it shouldn't affect existing users. The only thing the library uses that isn't available in core or alloc is the error trait, and a few `println!`s in tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ path = "src/lib.rs"
 [features]
 # Enable JSON output in the `cli` example:
 json_example = ["serde_json", "serde"]
+std = []
+default = ["std"]
 
 [dependencies]
 bigdecimal = { version = "0.2", features = ["serde"], optional = true }
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -11,9 +11,11 @@
 // limitations under the License.
 
 use super::ObjectName;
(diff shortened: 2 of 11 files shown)


joshwd36:
The failed clippy lints seem to be elsewhere in the code, so I'll fix them in a separate PR.

Dandandan:
Interesting @joshwd36 ! Will try to look at this soon. I am curious what you are planning to use it for :)

joshwd36:
> 
> 
> I am curious what you are planning to use it for :)

I'm hoping to write a basic database engine for a unikernel OS, that doesn't have std (yet, although I may work on that), but is still functional, and wouldn't be used on the low-power devices that would normally be associated with no_std, but on hypervisors on a server.

yjhmelody:
Is this PR still going on? Supporting no-std is great

alamb:
As noted by @koushiro, this PR is likely superceded by https://github.com/sqlparser-rs/sqlparser-rs/pull/332

joshwd36:
> As noted by @koushiro, this PR is likely superceded by #332

That's fine. Should I close this now or wait for #332 to be merged?

alamb:
Let's leave it open and I think when we merge #332 the Github API will close this one too

PR #293, opened 2021-02-07
Implement SUBSTRING(col [FROM <expr>] [FOR <expr>]) syntax

Dandandan:
This implements syntax for `substring(string [from expr] [for expr])`

This makes TCP-H Query 22 parse (Enabled it in the tests).

FYI @andygrove


andygrove (approved):
LGTM

PR #299, opened 2021-03-17
Implement TRY_CAST

seddonm1:
Hi @andygrove 

This change adds the TRY_CAST alongside CAST so that the choice on how to handle data casting errors can be differentiated.

See: https://docs.microsoft.com/en-us/sql/t-sql/functions/try-cast-transact-sql?view=sql-server-ver15

Also fixed a clippy error that must have been added recently.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -201,6 +201,12 @@ pub enum Expr {
         expr: Box<Expr>,
         data_type: DataType,
     },
+    /// TRY_CAST an expression to a different data type e.g. `TRY_CAST(foo AS VARCHAR(123))`
+    //  this differs from CAST in the choice of how to implement invalid conversions
+    TryCast {
+        expr: Box<Expr>,
+        data_type: DataType,
+    },
     /// EXTRACT(DateTimeField FROM <expr>)
     Extract {
         field: DateTimeField,
@@ -309,6 +315,7 @@ impl fmt::Display for Expr {
                 }
             }
             Expr::Cast { expr, data_type } => write!(f, "CAST({} AS {})", expr, data_type),
+            Expr::TryCast { expr, data_type } => write!(f, "TRY_CAST({} AS {})", expr, data_type),
             Expr::Extract { field, expr } => write!(f, "EXTRACT({} FROM {})", field, expr),
             Expr::Collate { expr, collation } => write!(f, "{} COLLATE {}", expr, collation),
             Expr::Nested(ast) => write!(f, "({})", ast),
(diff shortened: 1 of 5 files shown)


andygrove (approved):
LGTM. Thanks @seddonm1

seddonm1:
How does the release process work for this library?

andygrove:
> How does the release process work for this library?

I'm embarrassed to say that I don't really know anymore. @Dandandan released a version more recently than me so maybe he or @maxcountryman can help with that.

seddonm1:
@andygrove I think we can let you off the hook for not knowing given the work you have done building DataFusion and Ballista  😁

Dandandan:
@seddonm1 @andygrove 

thanks a lot, I will have a look at this PR later.
The release process is documented here: https://github.com/ballista-compute/sqlparser-rs/blob/main/docs/releasing.md
Basically fully automated after doing a version bump.

seddonm1:
Thanks @Dandandan

maxcountryman:
Happy to help with a release. As @Dandandan mentioned, it should be automated once we've pushed a tag.

seddonm1:
@Dandandan are you able to review and hopefully release? This will allow us to add the CastOptions https://github.com/apache/arrow/pull/9682 to DataFusion - hopefully before the next major release.

src/ast/mod.rs
```diff
@@ -201,6 +201,12 @@ pub enum Expr {
         expr: Box<Expr>,
         data_type: DataType,
     },
+    /// TRY_CAST an expression to a different data type e.g. `TRY_CAST(foo AS VARCHAR(123))`
+    //  this differs from CAST in the choice of how to implement invalid conversions
+    TryCast {
```
Dandandan, line 206: I am wondering, why the choice `TryCast` / `try_cast`?

I know bigquery uses `SAFE_CAST` https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#safe_casting

It would make sense to support both maybe?

Dandandan (approved):

Dandandan:
I think this looks good @seddonm1 ! I think we are good to go, do you want to address the `SAFE_CAST` question?

src/ast/mod.rs
seddonm1: I could easily change this to add SAFE_CAST too. There was some rationale discussed here: https://github.com/apache/arrow/pull/9682#issuecomment-799831930

basically it was just that I think SQL Server got there first with TRY_CAST

Dandandan:
@seddonm1 will be released as 0.9.0 , thanks!

PR #300, opened 2021-03-20
provide ILIKE support

maxcountryman:
This introduces support for ILIKE and NOT ILIKE. ILIKE is the
case-insensitive variant of LIKE. Systems such as Postgres, Redshift,
and Snowflake provide this variant.[1][2][3]

[1] https://www.postgresql.org/docs/7.3/functions-matching.html
[2] https://docs.aws.amazon.com/redshift/latest/dg/r_patternmatching_condition_like.html
[3] https://docs.snowflake.com/en/sql-reference/functions/ilike.html


maxcountryman:
This [lint issue](https://github.com/ballista-compute/sqlparser-rs/pull/300/checks?check_run_id=2155988613) seems to also happen on `main`.

Dandandan (approved):

Dandandan:
looks good! could you merge/rebase to get in the clippy lint fixes?

PR #307, opened 2021-04-14
Implement parsing of Postgres regex match operators

returnString:
Hey all,

This PR adds support for parsing Postgres' [POSIX regex matching operators](https://www.postgresql.org/docs/13/functions-matching.html#FUNCTIONS-POSIX-REGEXP). These are similar in nature to `LIKE` et al but with full support for regex patterns rather than being restricted to wildcards, so you can use `select * from my_table where name ~ '^pattern$'`.

I've restricted the parsing to only run when using the Postgres dialect, but I've not given lexing the same treatment as that doesn't _seem_ to be the done thing (e.g. `PGBitwiseXor` is lexed unconditionally). If this is actually desired, please let me know!

I'm not entirely happy with the token names, but given the overlap (e.g. `~` is both a unary and bitwise op now) I didn't want to attach names that were overly specific either.

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -80,6 +80,10 @@ pub enum BinaryOperator {
     PGBitwiseXor,
     PGBitwiseShiftLeft,
     PGBitwiseShiftRight,
+    PGRegexMatch,
+    PGRegexIMatch,
+    PGRegexNotMatch,
+    PGRegexNotIMatch,
 }
 
 impl fmt::Display for BinaryOperator {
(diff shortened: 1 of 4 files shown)


Dandandan:
I think this is looking good @returnString . Can you fix or ignore the linting errors?

alamb:
Possibly made redundant by https://github.com/ballista-compute/sqlparser-rs/pull/328

alamb:
Closing this PR as https://github.com/sqlparser-rs/sqlparser-rs/pull/328 was merged

PR #312, opened 2021-06-03
Add fuzzer based on honggfuzz

PsiACE:
fuzzing

Also see: #211

### DIFF
--- a/fuzz/fuzz_targets/fuzz_parse_sql.rs
+++ b/fuzz/fuzz_targets/fuzz_parse_sql.rs
@@ -1,0 +1,12 @@
+use honggfuzz::fuzz;
+use sqlparser::dialect::GenericDialect;
+use sqlparser::parser::Parser;
+
+fn main() {
+    loop {
+        fuzz!(|data: String| {
+            let dialect = GenericDialect {};
+            let _ = Parser::parse_sql(&dialect, &data);
+        });
+    }
+}
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2202,8 +2202,8 @@ impl<'a> Parser<'a> {
             Ok(Query {
                 with,
                 body,
-                limit,
                 order_by,
+                limit,
                 offset,
                 fetch,
             })
@@ -2358,8 +2358,7 @@ impl<'a> Parser<'a> {
                         ]) // This couldn't possibly be a bad idea
                     })?
                     .into_iter()
-                    .filter(|i| i.is_some())
-                    .map(|i| i.unwrap())
+                    .flatten()
                     .collect();
 
                 lateral_views.push(LateralView {
@@ -2414,8 +2413,8 @@ impl<'a> Parser<'a> {
             top,
             projection,
             from,
-            selection,
             lateral_views,
+            selection,
             group_by,
             cluster_by,
             distribute_by,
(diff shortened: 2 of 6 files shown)


BohuTANG:
Hi, @Dandandan @andygrove 

With this patch, we can easily to run the fuzz tool for sqlparser-rs and find more fuzz issues, great for robustness.
In the future we can make a CI to run it (limited time, i.e 10 minutes).

src/parser.rs
```diff
@@ -2358,8 +2358,7 @@ impl<'a> Parser<'a> {
                         ]) // This couldn't possibly be a bad idea
                     })?
                     .into_iter()
-                    .filter(|i| i.is_some())
-                    .map(|i| i.unwrap())
+                    .flatten()
```
alamb, line 2361: Is it correct that  this is a real bug that was found/fixed with the fuzzer?

fuzz/fuzz_targets/fuzz_parse_sql.rs
```diff
@@ -0,0 +1,12 @@
+use honggfuzz::fuzz;
+use sqlparser::dialect::GenericDialect;
+use sqlparser::parser::Parser;
+
+fn main() {
+    loop {
+        fuzz!(|data: String| {
```
alamb, line 7: This is pretty cool. Where does the data come from? Is it just random strings? Or does this somehow know how to make valid sql statements?

alamb (commented):
Sorry for the delay in review. I am trying to pitch in and help out now to clear the backlog. 

This is very cool @PsiACE  and @BohuTANG  -- thank you. It is cool that it seems you have found a bug with this fuzzer already.

My biggest question is about the input -- does the fuzzer simply check against random inputs or does it somehow know how to randomly make sql statements?

fuzz/fuzz_targets/fuzz_parse_sql.rs
alamb: I see some related comments here: https://github.com/sqlparser-rs/sqlparser-rs/pull/211#issuecomment-650716166

alamb (approved):
I think it is cool and having initial fuzzing, even if just random strings to start, is a good first step.

src/parser.rs
PsiACE: no, just make clippy happy

fuzz/fuzz_targets/fuzz_parse_sql.rs
PsiACE: For now it is some random bytes, but can be used to find bugs. Optimization can be continued later, refer to: https://www.cockroachlabs.com/blog/sqlsmith-randomized-sql-testing/

alamb: In a past project, we had a harness that could generate random SQL queries and it found **many** bugs -- such tests are a wonderful way to help database software mature

alamb: Filed https://github.com/apache/arrow-datafusion/issues/913 for DataFusion. Thanks for the pointer @PsiACE

houqp (approved):
This is pretty cool. Looking forward to the day when we could have sql syntax specific fuzzer :)

Issue #311, opened 2021-06-03
Use identifiable structs for AST variant fields


cmoog:
When consuming the generated AST for further processing, one needs to match against and deconstruct `ast::Statement` instances. Currently, the fields of this enum's variants are defined inline... which makes it impossible for downstream code to represent the nested structures as types. For example, consider this structure:

https://github.com/ballista-compute/sqlparser-rs/blob/35ef0eee3857c59827b2587bc813218a00dce264/src/ast/mod.rs#L490-L497

I propose we migrate to the following pattern.
```rust
pub struct AnalyzeStatement {
  table_name: ObjectName, 
  partitions: Option<Vec<Expr>>, 
  for_columns: bool, 
  ....
}
pub enum Statement { 
  Analyze(AnalyzeStatement),
  ...
}
```

Obviously this is a massive breaking change, but that seems OK given the `0.9.x` crates.io version. 

Thoughts?

alamb:
Is the primary usecase of this proposal so that downstream code can re-use `AnalyzeStatement` directly?

cmoog:
> Is the primary usecase of this proposal so that downstream code can re-use AnalyzeStatement directly?

Yes, exactly.

tvallotton:
I totally agree with this proposal. This would simplify my code considerably. I've got structs that are complete translations of some enum variants.

tvallotton:
I forked the project and made those changes if you want to see how it would look.

alamb:
I don't feel qualified to approve / accept this level of change without some additional buy in from @nickolay  / @andygrove  / @Dandandan . https://github.com/sqlparser-rs/sqlparser-rs/issues/311#issuecomment-902861958 seems like a good usecase to me but it is a pretty major change for all users of sqlparser-rs.

nickolay:
https://github.com/sqlparser-rs/sqlparser-rs/issues/40 could be marked as a duplicate of this issue as well. As mentioned there, I didn't think creating a separate struct for _every single variant_ (even those holding one or two fields) of every enum was worth the breakage, but extracting complex variants like `Statement::Analyze` is something we have done before.

In https://github.com/sqlparser-rs/sqlparser-rs/pull/315#issuecomment-916948007 @Dandandan suggested doing this for all variants of the `Statement` enum, which seems a sensible suggestion too, as most statements exceed the complexity threshold anyway.

alamb:
Your guidance makes a lot of sense to me @nickolay  -- thank you

PR #315, opened 2021-06-03
extract enum fields to named structs

cmoog:
Addresses #311 

Maybe it's appropriate to continue this refactor for `Expr` as well as the remaining `Statement` variants? Those seem much less pressing though since nearly all have <= 3 fields.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -485,168 +485,242 @@ impl fmt::Display for AddDropSync {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct AnalyzeStatement {
+    pub table_name: ObjectName,
+    pub partitions: Option<Vec<Expr>>,
+    pub for_columns: bool,
+    pub columns: Vec<Ident>,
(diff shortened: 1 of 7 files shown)


alamb (commented):
This seems like a useful improvement to me (though as @cmoog  points out a fairly large backwards incompatible change which will cause pain for downstream crates)

Any other thoughts?

cc @Dandandan  and @andygrove

Dandandan:
> Addresses #311
> 
> Maybe it's appropriate to continue this refactor for `Expr` as well as the remaining `Statement` variants? Those seem much less pressing though since nearly all have <= 3 fields.

Thank you for this PR.

I think it make sense if we do this, consistently do it for all statement variants.

I would be OK to not do the same thing for `Expr` now, I am not sure if there is a similar need to do it there too?

alamb:
Closing old PRs. Please feel free to re-open if this was a mistake

AugustoFKL:
@alamb @Dandandan I currently would love to have that approach, since my usage is requiring a lot of "unreachable!()" because of that.

Is that idea still OK? I can work on that this week

alamb:
Hi @AugustoFKL  -- I think given that the sqlparser API is basically stable, I don't really want to contemplate such a large change at this time :(

PR #324, opened 2021-07-15
Support Parsing of hexadecimal literals that start with 0x

### LINKED ISSUE #321: Allow Parsing of Hexadecimal Literals that start with '0x'
I am starting a project in which I am parsing some MSSQL queries and I came across:
`select collationname(0x0904D02400)`
Which resulted in a parse error: `Expected ), found: x0904D02400`

It doesn't seem to be in the official SQL spec but both [MySQL](https://dev.mysql.com/doc/refman/5.7/en/hexadecimal-literals.html) and [Sql Server](https://docs.microsoft.com/en-us/sql/t-sql/data-types/constants-transact-sql?view=sql-server-ver15#binary-constants) supports it.

TheSchemm:
Fixes #321
Please let me know if more needs to be added!

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -397,6 +397,17 @@ impl<'a> Tokenizer<'a> {
                 // numbers and period
                 '0'..='9' | '.' => {
                     let mut s = peeking_take_while(chars, |ch| matches!(ch, '0'..='9'));
+
+                    // match binary literal that starts with 0x
+                    if s == "0" && chars.peek() == Some(&'x') {
+                        chars.next();
+                        let s2 = peeking_take_while(
+                            chars,
+                            |ch| matches!(ch, '0'..='9' | 'A'..='F' | 'a'..='f'),
+                        );
+                        return Ok(Some(Token::HexStringLiteral(s2)));
+                    }
+
                     // match one period
                     if let Some('.') = chars.peek() {
                         s.push('.');
--- a/tests/sqlparser_mssql.rs
+++ b/tests/sqlparser_mssql.rs
@@ -115,7 +115,7 @@ fn parse_mssql_top() {
 
 #[test]
 fn parse_mssql_bin_literal() {
-  let _ = ms_and_generic().one_statement_parses_to("SELECT 0xdeadBEEF", "SELECT X'deadBEEF'");
+    let _ = ms_and_generic().one_statement_parses_to("SELECT 0xdeadBEEF", "SELECT X'deadBEEF'");
 }
 
 fn ms() -> TestedDialects {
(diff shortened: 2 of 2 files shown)


src/tokenizer.rs
```diff
@@ -397,6 +397,17 @@ impl<'a> Tokenizer<'a> {
                 // numbers and period
                 '0'..='9' | '.' => {
                     let mut s = peeking_take_while(chars, |ch| matches!(ch, '0'..='9'));
+
+                    // match binary literal that starts with 0x
+                    if s == "0" && chars.peek() == Some(&'x') {
```
alamb, line 402: Given it appears that postgres seems to treat `0xf00`  as an identifier I wonder if we should check `dialect.is_identifier_start()` prior to matching on this token?

https://dba.stackexchange.com/questions/203358/how-do-i-write-a-hex-literal-in-postgresql

tests/sqlparser_mssql.rs
```diff
@@ -113,6 +113,11 @@ fn parse_mssql_top() {
     let _ = ms_and_generic().one_statement_parses_to(sql, "SELECT TOP (5) bar, baz FROM foo");
 }
 
+#[test]
+fn parse_mssql_bin_literal() {
+    let _ = ms_and_generic().one_statement_parses_to("SELECT 0xdeadBEEF", "SELECT X'deadBEEF'");
+}
```
alamb, line 119: I recommend a test that the postgres dialect treats this as an identifier

alamb (commented):
Thank you for the contribution @TheSchemm ! Sorry for the delay in review.

I think we can fix the CI failures on this PR by rebasing it against the latest master (clippy has become more strict)

src/tokenizer.rs
TheSchemm: In section 4.1.1 (paragraph 2) in the [PostgreSQL docs](https://www.postgresql.org/docs/9.1/sql-syntax-lexical.html) they say that the identifiers have to start with a letter or an underscore.

alamb: 🤔  I tried this with postgres and the example `select 0xCC` still parses strangely, but as you mention, using `0x` as column names without quoting them doesn't work

```shell
(arrow_dev) alamb@MacBook-Pro:~/Software/arrow-datafusion$ psql
psql (13.4)
Type "help" for help.

alamb=# SELECT 0xCC;
 xcc 
-----
   0
(1 row)

alamb=# create table foo(0xCC VARCHAR);
ERROR:  syntax error at or near "0"
LINE 1: create table foo(0xCC VARCHAR);

alamb=# select c1 as 0x44 from test ;
ERROR:  syntax error at or near "0"
LINE 1: select c1 as 0x44 from test ;

alamb=# select c1 as "0x44" from test ;
 0x44 
------
    0
    3
     
    1
    3
(5 rows)


```

alamb (approved):
I think this looks good enough to me. Does anyone else have thoughts?

alamb:
Thanks again @TheSchemm

PR #332, opened 2021-08-20
Support `no_std`

koushiro:
* Add `std` feature and `default` feature
* Support `no_std`
* Update CI for `no_std` check

Close #286 
Close #327

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,12 +13,15 @@ include = [
     "Cargo.toml",
 ]
 edition = "2018"
+resolver = "2"
 
 [lib]
 name = "sqlparser"
 path = "src/lib.rs"
 
 [features]
+default = ["std"]
+std = []
 # Enable JSON output in the `cli` example:
 json_example = ["serde_json", "serde"]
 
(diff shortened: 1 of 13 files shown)


koushiro:
@Dandandan PTAL

Cargo.toml
```diff
@@ -13,12 +13,15 @@ include = [
     "Cargo.toml",
 ]
 edition = "2018"
+resolver = "2"
```
alamb, line 16: Is the change to resolver v2 needed for this change, or is it a nice to have / drive by cleanup?

reference https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2

alamb (approved):
Thank you @koushiro !

I reviewed the code and the output of the new CI check https://github.com/sqlparser-rs/sqlparser-rs/pull/332/checks?check_run_id=3379689095 and it looks good to me

cc @Dandandan  let me know if you have any issues with merging this PR

Cargo.toml
koushiro: Actually, it was added to pass CI 'compile-no-std' check

Because the dev-dependency 'simple_logger' introduces the std feature of log, if resolver=2 is not used, libstd will be implicitly introduced in the CI 'compile-no-std'

Dandandan (approved):

PR #334, opened 2021-08-21
Add TrimWhereField for `parse_trim_expr`

ever0de:


Dandandan:
Changes look good, could you add some tests @ever0de ?

alamb (commented):
Hi @ever0de  thank you for the contribution! 

Can you please add tests, as a way of documentation:
1. The positive cases (e.g. `TRIM(LEADING FROM foo)
2. The negative cases (e.g. `TRIM(FOO from FOO)`

Thanks!

ever0de:
> Changes look good, could you add some tests @ever0de ?

Some tests have been added to #331.
I'll add a case of failure.

alamb:
Thanks for the pointer to #331  👍  looks good

alamb:
I also double checked the postgres and mysql manuals that the syntax in this PR is correct

ever0de:
> I also double checked the postgres and mysql manuals that the syntax in this PR is correct

Is there any place to modify other than the test case?

alamb (approved):
Looks great @ever0de ! Thank you

PR #335, opened 2021-08-22
Fix the term 'Modulus' to 'Modulo'

MRGRAVITY817:
I've found out that the term for indicating `%` is currently `Modulus`, but I believe it should be `Modulo`.

Reference
---
- Definition of `Modulo`: https://en.wikipedia.org/wiki/Modulo_operation


alamb (approved):
I agree the proper term for the mathematical remainder operation is `Modulo` -- thanks @MRGRAVITY817 !

alamb:
cc @Dandandan and @andygrove

Dandandan (approved):

Dandandan:
Thanks @MRGRAVITY817

MRGRAVITY817:
> Thanks @MRGRAVITY817 

You're welcome 😀

Issue #279, opened 2020-10-20
Parse create table statement error


Thearas:
Got an error while parsing a create statement:
```sql
CREATE TABLE T (
    id int(4)
);
```
```sh
Parsing sql 'CREATE TABLE T (
        id int(4)
    );'...
Error: sql parser error: Expected ',' or ')' after column definition, found: (
```
But the sql can be successfully executed on MySQL.

Dandandan:
Thanks for creating the issue. The parser doesn't support the int(4) part yet.

PR #337, opened 2021-08-24
Support minimum display width for integer data types

vasilev-alex:
This PR address the #279 issue as MySQL supports the [minimum display width](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html) for integer data types.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -42,13 +42,13 @@ pub enum DataType {
     /// Floating point with optional precision e.g. FLOAT(8)
     Float(Option<u64>),
     /// Tiny integer
-    TinyInt,
+    TinyInt(Option<u64>),
     /// Small integer
-    SmallInt,
+    SmallInt(Option<u64>),
     /// Integer
-    Int,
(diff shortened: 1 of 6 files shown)


src/ast/data_type.rs
```diff
@@ -42,13 +42,13 @@ pub enum DataType {
     /// Floating point with optional precision e.g. FLOAT(8)
     Float(Option<u64>),
     /// Tiny integer
-    TinyInt,
+    TinyInt(Option<u64>),
```
alamb, line 45: I suggest updating the docstrings to explain what the `Option` means:

like 
```rust
/// Tiny integer, with optional precision (e.g. TINYINT(2))
```

alamb (approved):
I think this is looking good -- thank you @vasilev-alex .

It looks like there is some small CI issue (need to run `cargo fmt`) and I do think updating the docstrings is important. 

Overall, looks great

cc @Dandandan  / @andygrove

alamb (approved):

PR #338, opened 2021-08-24
feat: Support SHOW CREATE TABLE|EVENT|FUNCTION (MySQL specific)

ovr:
Hello!

Thanks

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -104,29 +104,21 @@ fn parse_show_columns() {
 fn parse_show_create() {
     let obj_name = ObjectName(vec![Ident::new("myident")]);
 
-    assert_eq!(
-        mysql_and_generic().verified_stmt("SHOW CREATE TABLE myident"),
-        Statement::ShowCreate {
-            obj_type: ShowCreateObject::Table,
-            obj_name: obj_name.clone(),
-        }
-    );
-
-    assert_eq!(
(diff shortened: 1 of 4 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -100,6 +100,35 @@ fn parse_show_columns() {
     }
 }
 
+#[test]
+fn parse_show_create() {
+    let obj_name = ObjectName(vec![Ident::new("myident")]);
+
+    assert_eq!(
```
alamb, line 107: I wonder if it would be possible to check all the possible variants here (perhaps with a loop / expected values)?

alamb (approved):
Look good to me 👍  Thank you @ovr 

For anyone else following along, here are some links to the Mysql reference manual:

https://dev.mysql.com/doc/refman/8.0/en/show-create-table.html
https://dev.mysql.com/doc/refman/8.0/en/show-create-trigger.html
https://dev.mysql.com/doc/refman/5.7/en/show-create-function.html
https://dev.mysql.com/doc/refman/5.7/en/show-create-event.html
https://dev.mysql.com/doc/refman/5.7/en/show-create-procedure.html

Looks like there may be a few more `SHOW CREATE` statements (we could always add them in a future PR, I just figured I would point them out on this PR):
https://dev.mysql.com/doc/refman/5.7/en/show.html

```sql
SHOW CREATE USER
SHOW CREATE VIEW
SHOW CREATE DATABASE
```

alamb:
Thanks @ovr !

PR #340, opened 2021-08-25
feat: Support DESCRIBE table_name

ovr:
Hello!

I am working on MySQL shim for BI systems and It's a popular way to introspect schemas in MySQL, will be cool to support it in this parser.

Thanks

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -735,7 +735,8 @@ pub enum Statement {
         data_types: Vec<DataType>,
         statement: Box<Statement>,
     },
-    /// Note: this is a MySQL-specific statement.
+    /// EXPLAIN TABLE
+    /// Note: this is a MySQL-specific statement. See <https://dev.mysql.com/doc/refman/8.0/en/explain.html>
     ExplainTable {
         // If true, query used the MySQL `DESCRIBE` alias for explain
         describe_alias: bool,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -735,7 +735,14 @@ pub enum Statement {
         data_types: Vec<DataType>,
         statement: Box<Statement>,
     },
-    /// EXPLAIN
+    /// Note: this is a MySQL-specific statement.
+    ExplainTable {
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -735,8 +735,17 @@ pub enum Statement {
         data_types: Vec<DataType>,
         statement: Box<Statement>,
     },
-    /// EXPLAIN
+    /// Note: this is a MySQL-specific statement.
+    ExplainTable {
```
alamb, line 739:
```suggestion
    /// EXPLAIN TABLE
    /// Note: this is a MySQL-specific statement. See <https://dev.mysql.com/doc/refman/8.0/en/explain.html>
    ExplainTable {
```

src/ast/mod.rs
```diff
@@ -735,8 +735,17 @@ pub enum Statement {
...
+        // MySQL supports DESCRIBE alias for EXPLAIN
```
alamb, line 740:
```suggestion
        // If true, query used the MySQL `DESCRIBE` alias for explain
```

src/ast/mod.rs
```diff
@@ -735,8 +735,17 @@ pub enum Statement {
...
+        describe_alias: bool,
+        // Table name
+        table_name: ObjectName,
+    },
+    /// EXPLAIN / DESCRIBE for select_statement
     Explain {
+        // MySQL supports DESCRIBE alias for EXPLAIN
```
alamb, line 747:
```suggestion
        // If true, query used the MySQL `DESCRIBE` alias for explain
```

alamb (commented):
Thanks @ovr . Looks like a good feature to me 👍 

Looking at https://dev.mysql.com/doc/refman/8.0/en/explain.html

I wonder if you have considered limiting `DESCRIBE` support to only the `MySql` dialect: https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/dialect/mssql.rs

alamb (approved):

PR #344, opened 2021-09-02
Split parser into multiple files

koushiro:

I found that the parser module is only in one large file, which is not very friendly for viewing the code and resolving git conflicts, so I split it by the way when I view the code, but I don’t know if this big PR can be accepted (without changing the parser logic)

visibility changes:

- `pub fn parse_identifiers` ==> `fn parse_identifiers`
- `fn parse_prepare` ==> `pub fn parse_prepare`
- `fn parse_execute` ==> `pub fn parse_execute`
- `fn parse_deallocate` ==> `pub fn parse_deallocate`


koushiro:
The commit `Differentiate methods` caused a lot of git diff content, but in fact it only adjusted the order of the methods.

If this is unacceptable, I will revert the commit.

alamb (commented):
Thank you for the contribution @koushiro 

I am not opposed to this change, but I am not sure if the additional value is worth the potential risk to breaking downstream consumers. I will let others weigh in as well

koushiro:
I dropped the commit `Differentiate methods` to make it easier to be reviewed.

/cc @alamb @Dandandan Is there any progress?

alamb:
> /cc @alamb @Dandandan Is there any progress?

I am of two minds here -- I like that you are trying to improve the codebase, but I am worried about doing a substantial reorganization that offers contributors to the library some benefit but imposes costs on users of the library.

koushiro:
> > /cc @alamb @Dandandan Is there any progress?
> 
> I am of two minds here -- I like that you are trying to improve the codebase, but I am worried about doing a substantial reorganization that offers contributors to the library some benefit but imposes costs on users of the library.

At least this PR has no other breaking changes except for the visibility changes of some functions. I think this will not impose costs on the users of the library.

PR #347, opened 2021-09-09
Improve tokenizer error

koushiro:

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -68,10 +68,7 @@ use IsLateral::*;
 
 impl From<TokenizerError> for ParserError {
     fn from(e: TokenizerError) -> Self {
-        ParserError::TokenizerError(format!(
-            "{} at Line: {}, Column {}",
-            e.message, e.line, e.col
-        ))
+        ParserError::TokenizerError(e.to_string())
     }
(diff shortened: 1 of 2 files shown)


koushiro:
@alamb @Dandandan PTAL

alamb (commented):
Thank you for the contribution @koushiro  -- looks like an improvement to me 👍 

Can you please add at least one test case demonstrating the functionality? The rationale is to ensure that if someone changes / breaks this functionality in the future tests will catch it.

koushiro:
> Can you please add at least one test case demonstrating the functionality? The rationale is to ensure that if someone changes / breaks this functionality in the future tests will catch it.

I am not sure about what your meaning, test case about what?

alamb:
> I am not sure about what your meaning, test case about what?

Sorry for not being clear -- I guess I was thinking something that ensures that `TokenizerError` implements `std::error::Error` -- the rationale being that if someone removed the line

```rust
impl std::error::Error for TokenizerError {}
```

by accident in a future PR we would likely never know the difference as no tests would fail.

alamb (approved):
Thanks @koushiro

PR #349, opened 2021-09-09
Implement some MySQL specific syntax and extend the UPDATE statement

0xA537FD:
Implement the `ON DUPLICATE KEY UPDATE` syntax on an `INSERT` statement which is MySQL specific as well as the ability to use joins in an `UPDATE` statement.

Besides that a breaking change was introduced on the UPDATE statement to support table aliases (which is also defined in the 2016 SQL spec).

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2939,10 +2939,10 @@ impl<'a> Parser<'a> {
             let after_columns = self.parse_parenthesized_column_list(Optional)?;
 
             let source = Box::new(self.parse_query()?);
-            let on = if dialect_of!(self is MySqlDialect) && self.parse_keyword(Keyword::ON) {
-                self.expect_keyword(Keyword::DUPLICATE);
-                self.expect_keyword(Keyword::KEY);
-                self.expect_keyword(Keyword::UPDATE);
+            let on = if self.parse_keyword(Keyword::ON) {
+                self.expect_keyword(Keyword::DUPLICATE)?;
+                self.expect_keyword(Keyword::KEY)?;
+                self.expect_keyword(Keyword::UPDATE)?;
                 let l = self.parse_comma_separated(Parser::parse_expr)?;
 
                 Some(OnInsert::DuplicateKeyUpdate(l))
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1914,7 +1914,7 @@ impl<'a> Parser<'a> {
                 data_type,
                 options,
             }
-        }else {
+        } else {
             return self.expected(
                 "ADD, RENAME, PARTITION or DROP after ALTER TABLE",
                 self.peek_token(),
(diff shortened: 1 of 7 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -216,6 +216,260 @@ fn parse_create_table_with_minimum_display_width() {
...
+                on
+            );
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn parse_update_with_joins() {
+    let sql = "UPDATE orders AS o JOIN customers AS c ON o.customer_id = c.id SET o.completed = true WHERE c.firstname = 'Peter'";
```
alamb, line 378: I didn't find this type of syntax in the MySql manual https://dev.mysql.com/doc/refman/8.0/en/update.html

However, I tried such a thing in `mysql` and it worked 🤯 

```sql
mysql> create table items(id int, price double);
Query OK, 0 rows affected (0.01 sec)

mysql> create table month(id int, price double);

mysql> UPDATE items,month SET items.price=month.price
    -> WHERE items.id=month.id;
Query OK, 0 rows affected (0.01 sec)
Rows matched: 0  Changed: 0  Warnings: 0

mysql> UPDATE items JOIN month ON items.id = month.id SET items.price=month.price WHERE items.id=month.id;
Rows matched: 0  Changed: 0  Warnings: 0
```

tests/sqlparser_mysql.rs
```diff
@@ -216,6 +216,260 @@ fn parse_create_table_with_minimum_display_width() {
...
+    let sql1 = "ALTER TABLE orders CHANGE COLUMN description desc TEXT NOT NULL";
+    match mysql().verified_stmt(sql1) {
+        Statement::AlterTable { name, operation } => {
+            assert_eq!(expected_name, name);
+            assert_eq!(expected_operation, operation);
+        }
+        _ => unreachable!(),
+    }
+
+    let sql2 = "ALTER TABLE orders CHANGE description desc TEXT NOT NULL";
```
alamb, line 463: I tested this locally in mysql -- looks good

src/parser.rs
```diff
@@ -2139,16 +2159,41 @@ impl<'a> Parser<'a> {
         Ok(ObjectName(idents))
     }
 
+    /// Parse identifiers strictly i.e. don't parse keywords
```
alamb, line 2162: maybe calling this `parse_identifiers_non_keywords` would be a more specific name

alamb (commented):
Thank you for this @0xA537FD -- I reviewed this PR fairly carefully and it looks good to me. 

There is a conflict and there are a few lints to resolve, but otherwise I think this is looking good to me 👍

alamb:
Will try and have a look at this over the weekend

alamb (approved):
This one totally fell off my radar -- I am sorry @0xA537FD . The changes look good

This PR now has some conflicts -- I have fixed them in https://github.com/sqlparser-rs/sqlparser-rs/pull/375 if you want to pull that commit into this branch.

I'll also understand if you gave up on us and are working on something else now -- in that case I can merge #375 as well.

alamb:
Included in https://github.com/sqlparser-rs/sqlparser-rs/pull/375 so closing this PR. Thanks again @0xA537FD  and sorry for the delay

PR #351, opened 2021-09-10
Split parser into multiple files

koushiro:
part of #344 

I dropped the commit `Differentiate methods` of #344 to make it easier to be reviewed.

visibility changes (**no breaking changes**):

`fn parse_prepare` ==> `pub fn parse_prepare`
`fn parse_execute` ==> `pub fn parse_execute`
`fn parse_deallocate` ==> `pub fn parse_deallocate`

/cc @alamb


alamb:
Checking this one out

alamb:
FWIW I did do a spot check and try this code out with DataFusion and DataFusion's tests still pass

houqp:
I also did a spot check, mostly looks good to me. Couple good improvements on the doc string as well :+1: 

I have to admit though, personally I think refactoring code into mods makes sense when they are mostly decoupled, which doesn't seem like the case here. I actually find it easier to navigate and search the code base when coupled code are in the same file. That said, it's not a strong no from me.

PR #352, opened 2021-09-10
Move the `keywords` module

koushiro:


alamb:
So something that would help me understand these PRs is the rationale for making the change. For example, this PR perhaps makes the code structure a bit more logical but won't it require a bunch of rework for all users of the library?

koushiro:
> this PR perhaps makes the code structure a bit more logical but won't it require a bunch of rework for all users of the library?

yes

houqp:
perhaps we can reexport the keywords module in dialect module to make it backwards compatible until we do the next major release?

koushiro:
> perhaps we can reexport the keywords module in dialect module to make it backwards compatible until we do the next major release?

@alamb  has been released v0.11

alamb:
I guess I am just thinking of users of this crate -- when they upgrade to a new version of sqlparser I feel the best experience is to minimize the amount of changes they have to make to their code.

For changes required to support new features (e.g. new enum variants or new fields in existing variants) I don't see a way around requiring the user code to change.

However, I don't like the idea of causing work for downstream users for changes that are basically internal improvements to `sqlparser-rs` 

Thus, I would be happy to merge this PR if it was backwards compatible (and @houqp  has suggested a good way to do that)

koushiro:
@alamb Now the change is backwards compatible, I originally planned to add deprecated to re-export, like below:

```
#[deprecated(
    since = "0.11.1",
    note = "The keywords module has been moved into the root of the library, \
    please use sqlparser::keywords instead of sqlparser::dialect::keywords"
)]
pub use crate::keywords;
```
but it doesn't seem to work,  see https://github.com/rust-lang/rust/issues/30827

alamb (approved):
Thanks @koushiro !

PR #353, opened 2021-09-10
Rname `Token::Mult` to `Token::Mul`

koushiro:

I think rust developers will be more inclined to use the abbr `Mul` for multiplication operator.


alamb (approved):
Thank you @koushiro . I think this is a nice improvement and the cost on users during upgrade is minimal.

PR #356, opened 2021-09-16
Enable map access for numbers, multiple nesting levels

Igosuki:
This enables number based map access in the AST. I am unsure if this required a new type rather than just patching MapAccess but here goes.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -281,21 +281,15 @@ impl fmt::Display for Expr {
         match self {
             Expr::Identifier(s) => write!(f, "{}", s),
             Expr::MapAccess { column, keys } => {
-                write!(
-                    f,
-                    "{}{}",
-                    column,
-                    keys.iter()
-                        .map(|k| {
-                            match k {
-                                k @ Value::Number(_, _) => format!("[{}]", k),
-                                Value::SingleQuotedString(s) => format!("[\"{}\"]", s),
-                                _ => format!("[{}]", k),
-                            }
-                        })
-                        .collect::<Vec<String>>()
-                        .join("")
-                )
+                write!(f, "{}", column)?;
+                for k in keys {
+                    match k {
+                        k @ Value::Number(_, _) => write!(f, "[{}]", k)?,
+                        Value::SingleQuotedString(s) => write!(f, "[\"{}\"]", s)?,
+                        _ => write!(f, "[{}]", k)?
+                    }
+                }
+                Ok(())
             }
             Expr::Wildcard => f.write_str("*"),
             Expr::QualifiedWildcard(q) => write!(f, "{}.*", display_separated(q, ".")),
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -286,7 +286,7 @@ impl fmt::Display for Expr {
                     match k {
                         k @ Value::Number(_, _) => write!(f, "[{}]", k)?,
                         Value::SingleQuotedString(s) => write!(f, "[\"{}\"]", s)?,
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -251,7 +251,7 @@ pub enum Expr {
     },
     MapAccess {
```
alamb, line 252: I wonder if we can improve the docstrings here while we are modifying this file?

alamb (approved):
Thanks @Igosuki !

Looks good to me. FYI @hntd187 as `MapAccess` seems to have been introduced in https://github.com/sqlparser-rs/sqlparser-rs/pull/235

Broadly speaking looks compatible with https://www.postgresql.org/docs/9.1/arrays.html too

alamb:
This PR also adds support for multiple levels of map access as well, right? Like not just ["foo"] also ["foo"]["bar"]

houqp (approved):
LGTM, thanks @Igosuki !

alamb:
looks like there are some CI tests to clean up as well @Igosuki FYI

hntd187:
Looks good to me, I remember map access being something I was not entirely confident in for my initial work, but perhaps if one was able to extend it then it wasn't as bad as my own perception was of it.

Igosuki:
Well, there will be occasions to expand upon this. For instance to support ranges, which are in the pg dialect (e.g. ```foo[1:2]```). Also I think the display is wrong for map access regarding strings, because ```foo['bar']``` and ``foo[bar]``` prints ```foo["'bar'"]```
I added a test for string keys in map access as well to wrap it up.

Igosuki:
It breaks the hive test for map_access, since we don't differentiate between double and single quoted strings, I think this needs further fixing

Igosuki:
Since map access belongs into the OLAP world and there is already the hive dialect, I went and enforced double quoted strings for map access in the display.

Igosuki:
So now MapAccess is an infinite depth path access to nested values. I think allowing compound field selection within this should be a separate addition as it requires patching the tokenization.

src/ast/mod.rs
```diff
@@ -280,7 +280,23 @@ impl fmt::Display for Expr {
...
+            Expr::MapAccess { column, keys } => {
+                write!(
+                    f,
+                    "{}{}",
+                    column,
+                    keys.iter()
+                        .map(|k| {
+                            match k {
+                                k @ Value::Number(_, _) => format!("[{}]", k),
+                                Value::SingleQuotedString(s) => format!("[\"{}\"]", s),
```
alamb, line 292: It is odd to me that  a `SingleQuotedString` is emitted with double quotes (`"`) but it is consistent with the existing implementation

alamb (approved):

alamb:
>  I think allowing compound field selection within this should be a separate addition as it requires patching the tokenization so that '.' is allowed after brackets 

makes sense to me

alamb:
FYI @Igosuki  I fixed a [CI failure](https://github.com/sqlparser-rs/sqlparser-rs/pull/356/checks?check_run_id=3703131981) with c9e3cfba47d2df923fdb287e0af7032025d3f4aa

Igosuki:
Thanks, now to enable this in data fusion

Le ven. 24 sept. 2021 à 20:22, Andrew Lamb ***@***.***> a
écrit :

houqp:
Thank you for your work on this @Igosuki !

Igosuki:
Thanks for merging

alamb:
> Thanks, now to enable this in data fusion

See https://github.com/apache/arrow-datafusion/pull/1052

PR #357, opened 2021-09-17
Add logical xor

zmrdltl:
In the existing sqlparser, `^`, which is `BitwiseXor`, is implemented, but the logical operator `XOR` is not implemented.
ex) 
```sql
SELECT true XOR true, false XOR false, true XOR false, false XOR true
```

ref. https://mariadb.com/docs/reference/mdb/operators/XOR/

The two Boolean values of `XOR` were added.

After adding this, it will be used for `GlueSql`.
ref. https://github.com/gluesql/gluesql/issues/354


houqp (approved):

alamb (approved):
Thank you for the contribution @zmrdltl !

XOR is also part of Mysql -- https://dev.mysql.com/doc/refman/8.0/en/logical-operators.html#operator_xor 👍 

Skimming that page, it seems like sqlparser also does not implement the `NOT` operator -- maybe something to add in a follow on PR if you need it.

alamb:
🤔  there appears to be something wrong with coveralls (returning a 405):

https://github.com/sqlparser-rs/sqlparser-rs/pull/357/checks?check_run_id=3639889624

zmrdltl:
thanks :)

alamb:
> I wonder which part of the not is not parsing.

I am not sure either. If someone has time to look at it that would be great

PR #358, opened 2021-09-19
Add support MySQL syntax

wqc200:
There are mainly the following four grammars:
1. create table with options, example: create table table_name (...) engine=innodb
2. show variables
3. show table status
4. value only string


src/ast/value.rs
```diff
@@ -158,6 +161,25 @@ pub fn escape_single_quote_string(s: &str) -> EscapeSingleQuoteString<'_> {
     EscapeSingleQuoteString(s)
 }
 
+pub struct EscapeDoubleQuoteString<'a>(&'a str);
```
alamb, line 164: 👍

src/tokenizer.rs
```diff
@@ -413,6 +417,25 @@ impl<'a> Tokenizer<'a> {
                     let s = self.tokenize_single_quoted_string(chars)?;
                     Ok(Some(Token::SingleQuotedString(s)))
                 }
+                '\"' => {
+                    let s = self.tokenize_double_quoted_string(chars)?;
+                    Ok(Some(Token::DoubleQuotedString(s)))
+                }
+                // quote_start if self.dialect.is_delimited_string_start(quote_start) => {
```
alamb, line 424: Was it meant to leave this code commented out?

alamb (commented):
Thanks @wqc200 ! I think this needs tests before we merge it, mostly so that someone doesn't accidentally break this functionality with a future refactor

alamb:
Closing old seemingly stalled PR. Please re-open if this was a mistake

PR #361, opened 2021-10-14
Implement is [not] distinct from

Dandandan:
FYI @xudong963 @alamb


xudong963:
Dan dan, nice work! 👍🏻

PR #365, opened 2021-11-11
Add basic support for GRANT and REVOKE

blx:
Hello 👋 

Preliminary PR here for basic `GRANT` and `REVOKE` support. I'm not sure if permissions statements like GRANT are in scope here; if not please feel free to close!

My personal experience is mostly postgres-flavored but I've tried to stick to the spec here. Note, this only addresses granting privileges regarding objects (like tables); there is also a `GRANT <role_name> TO <role_specification>` for adding users to groups, which is not considered here.

This also doesn't yet consider ~column permissions (`GRANT SELECT ON a (id, created) ...`), nor~ other object types like `DATABASE` or `DOMAIN`.

Please let me know if there's interest in this, thanks!


src/parser.rs
```diff
@@ -2881,6 +2882,82 @@ impl<'a> Parser<'a> {
...
 
+    /// Parse a GRANT statement.
+    pub fn parse_grant(&mut self) -> Result<Statement, ParserError> {
+        let privileges = if self.parse_keyword(Keyword::ALL) {
+            GrantPrivileges::All { with_privileges_keyword: self.parse_keyword(Keyword::PRIVILEGES) }
+        } else {
+            GrantPrivileges::Privileges(
+                self.parse_comma_separated(Parser::parse_grant_permission)?
+                    .iter().map(|kw| match kw {
+                        Keyword::DELETE     => keywords::DELETE.into(),
```
blx, line 2893: Please let me know if there is an easier way to stringify these 😅 

Actually, I'm guessing that a new explicit `GrantPrivilege` enum would be preferable here, as opposed to these strings?

src/ast/mod.rs
```diff
@@ -1359,6 +1409,137 @@ impl fmt::Display for Statement {
...
 }
 
+/// Privileges granted in a GRANT statement or revoked in a REVOKE statement.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum Privileges {
+    /// All privileges applicable to the object type
+    All {
+        /// Optional keyword from the spec, ignored in practice
+        with_privileges_keyword: bool,
```
alamb, line 1419: Maybe worth removing it from the struct then?

alamb (approved):
Thank you @blx  -- sorry for the lengthy delay -- I didn't see this PR until recently. 

I think this PR is well constructed and well tested.

alamb:
@blx  I merged your code to master and fixed some clippy warnings

src/ast/mod.rs
blx: @alamb I erred on the side of losslessness here, but I wasn't sure to what degree that's a goal.

I guess there are already many cases where an implied default becomes explicit when parsed, and would then produce different output after a roundtrip if the input had omitted the default (eg. `cascade` in `Statement::Drop`).

If you think it would be cleaner I'm happy to PR removing this.

alamb: If this field is needed to survive the round trip I think keeping it makes sense.

PR #366, opened 2021-11-21
parse grouping sets, rollup, and cube for postgresql

Jimexist:


Jimexist:
need to rebase onto #367

houqp (approved):

tests/sqlparser_common.rs
```diff
@@ -1039,6 +1039,24 @@ fn parse_select_group_by() {
...
+            Expr::Identifier(Ident::new("size")),
+            Expr::GroupingSets(vec![
+                vec![Expr::Identifier(Ident::new("brand"))],
+                vec![Expr::Identifier(Ident::new("size"))],
+                vec![],
+            ])
+        ],
+        select.group_by
+    );
+}
```
alamb, line 1058: I think we should also add some test here for ROLLUP and CUBE

src/parser.rs
```diff
@@ -557,6 +557,63 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// parse a group by expr. a group by expr can be one of group sets, roll up, cube, or simple
+    /// expr.
+    fn parse_group_by_expr(&mut self) -> Result<Expr, ParserError> {
+        if self.parse_keywords(&[Keyword::GROUPING, Keyword::SETS]) {
```
alamb, line 563: I double checked at MySql also supports GROUPING, though it seems it supports `WITH ROLLUP` rather than `ROLLUP`,  https://dev.mysql.com/doc/refman/8.0/en/group-by-modifiers.html 

but seemingly not CUBE https://dev.mysql.com/doc/search/?d=201&p=1&q=cube 

I think we should make parsing this part of the syntax conditional on the the postgresql dialect

tests/sqlparser_common.rs
Jimexist: added

src/parser.rs
Jimexist: done now

houqp (approved):

alamb (commented):
Thanks @Jimexist !

PR #370, opened 2021-11-30
support select limit offset syntax

QuantumBear:
Support the syntax "select fields from table limit offset, count", which MySQL support.


src/parser.rs
```diff
@@ -3065,11 +3065,26 @@ impl<'a> Parser<'a> {
...
 
     /// Parse a LIMIT clause
-    pub fn parse_limit(&mut self) -> Result<Option<Expr>, ParserError> {
+    pub fn parse_limit(&mut self) -> Result<Option<Limit>, ParserError> {
         if self.parse_keyword(Keyword::ALL) {
             Ok(None)
         } else {
-            Ok(Some(Expr::Value(self.parse_number_value()?)))
+            let first_number = self.parse_number_value()?;
+            let second_number = if  self.consume_token(&Token::Comma){
```
alamb, line 3073: Since this is specific to the mysql dialect, can you please add a guard here like

```
if dialect_of!(self is MySqlDialect) {
}
```

src/ast/query.rs
```diff
@@ -524,6 +524,24 @@ impl fmt::Display for OrderByExpr {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Limit {
+    pub offset: Option<Expr>,
```
alamb, line 530:
```suggestion
    // MySQL specific
    pub offset: Option<Expr>,
```

alamb (commented):
This is looking good -- thank you @QuantumBear 

Other than conditionalizing on the dialect, I think this is ready to go 👍

alamb:
Sorry in the delay in revieww

QuantumBear:
Sorry for late because I exprienced a crazy work week (996). And I try to add some test for this syntax. Add this code to tests/sqlparser_mysql.rs.

`#[test]
fn parse_limit_with_offset_count() {
    // In dialects supporting LIMIT with offset and count
     match mysql().verified_stmt("SELECT id FROM customer LIMIT 2,4") {
        Statement::Query(query ) => {
            match query.limit {
                Some(Limit { offset, count }) => {
                    assert_eq!(Some(Expr::Value(Value::Number("2".to_string(), false))), offset);
                    assert_eq!(Expr::Value(Value::Number("4".to_string(), false)), count);
                }
                _ => unreachable!(),
            }
        }
        _ => unreachable!(),
    }
}
`
The weird thing is, in unit test, dialect_of!(self is MsSqlDialect) returns false.  Is that correct?

alamb:
> The weird thing is, in unit test, dialect_of!(self is MsSqlDialect) returns false. Is that correct?


It doesn't sound correct to me 🤔

src/parser.rs
```diff
@@ -3380,11 +3380,33 @@ impl<'a> Parser<'a> {
...
 
     /// Parse a LIMIT clause
-    pub fn parse_limit(&mut self) -> Result<Option<Expr>, ParserError> {
+    pub fn parse_limit(&mut self) -> Result<Option<Limit>, ParserError> {
         if self.parse_keyword(Keyword::ALL) {
             Ok(None)
         } else {
-            Ok(Some(Expr::Value(self.parse_number_value()?)))
+            let first_number = self.parse_number_value()?;
+            if dialect_of!(self is MsSqlDialect) {
```
alamb, line 3388: MsSQL is microsoft sql server, rather than mysql I think

Issue #362, opened 2021-10-16
MySql dialect doesn't handle slash-escaped single quotes


WesleyAC:
The following fails to parse:

```rust
fn main() {
    println!("{:?}",
        sqlparser::parser::Parser::parse_sql(
            &sqlparser::dialect::MySqlDialect {},
            r#"INSERT INTO `example` (`col`) VALUES ('foo \'bar\' baz');"#
        )
    );
}
```

Instead returning:

```
Err(ParserError("Expected ), found: bar"))
```

But I believe it shouldn't — `mysqldump` generates SQL that is escaped like this, and the [docs](https://dev.mysql.com/doc/refman/8.0/en/string-literals.html) explicitly mention this as a valid escaping (see Table 9.1).

This is, according to [this doc](https://www.ibm.com/docs/en/informix-servers/12.10?topic=statements-quotation-marks-escape-characters) a difference between ANSI SQL and MySQL, but unfortunately this needs to be fixed in the tokenizer, which AFAICT the dialect system doesn't have a way to hook into?

sanxiyn:
This seems fixed. If so, please close.

LucaCappelletti94:
Closing as completed in https://github.com/apache/datafusion-sqlparser-rs/pull/373

PR #373, opened 2021-12-05
fix: mysql backslash escaping

vasilev-alex:
Addresses this issue #362 

Works only when a raw string provided as input

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -411,6 +411,7 @@ impl<'a> Tokenizer<'a> {
                 // string
                 '\'' => {
                     let s = self.tokenize_single_quoted_string(chars)?;
+
                     Ok(Some(Token::SingleQuotedString(s)))
                 }
                 // delimited (quoted) identifier
@@ -646,6 +647,24 @@ impl<'a> Tokenizer<'a> {
                         chars.next();
                     } else {
                         return Ok(s);
+                    }
+                }
+                '\\' => {
+                    chars.next();
+                    match chars.peek() {
+                        None => {
+                            return self.tokenizer_error("Unterminated escape sequence");
+                        }
+                        Some(&c) => {
+                            chars.next();
+                            match c {
+                                '\'' => {
+                                    s.push('\\');
+                                    s.push('\'');
+                                }
+                                x => s.push(x),
+                            }
+                        }
                     }
                 }
                 _ => {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -31,8 +31,8 @@ use core::str::Chars;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-use crate::dialect::Dialect;
 use crate::dialect::SnowflakeDialect;
+use crate::dialect::{Dialect, MySqlDialect};
 use crate::keywords::{Keyword, ALL_KEYWORDS, ALL_KEYWORDS_INDEX};
 
 /// SQL Token enumeration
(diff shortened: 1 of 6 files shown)


src/tokenizer.rs
```diff
@@ -648,6 +649,24 @@ impl<'a> Tokenizer<'a> {
...
+                    chars.next();
+                    match chars.peek() {
+                        None => {
+                            return self.tokenizer_error("Unterminated escape sequence");
+                        }
+                        Some(&c) => {
+                            chars.next();
+                            match c {
+                                '\'' => {
+                                    s.push('\\');
```
alamb, line 662: I would have expected the parser to convert `\'` to `'` ?

```suggestion
```

src/tokenizer.rs
```diff
@@ -648,6 +649,24 @@ impl<'a> Tokenizer<'a> {
                         return Ok(s);
                     }
                 }
+                '\\' => {
```
alamb, line 652: As postgres's quoting is different (e.g. it uses `''` instead of `\'` I think this case should be wrapped in a `if dialect_of!(self is MySqlDialect)` or something

tests/sqlparser_mysql.rs
```diff
@@ -176,6 +176,33 @@ fn parse_quote_identifiers() {
...
+#[test]
+fn parse_escaped_string() {
+    let sql = r#"SELECT 'I\'m fine'"#;
+
+    let projection = mysql().verified_only_select(sql).projection;
+    let item = projection.get(0).unwrap();
+
+    match &item {
+        SelectItem::UnnamedExpr(Expr::Value(value)) => {
+            assert_eq!("'I\\'m fine'", value.to_string());
```
alamb, line 188: I would expect this to be

```suggestion
            assert_eq!("'I'm fine'", value.to_string());
```

tests/sqlparser_mysql.rs
```diff
@@ -176,6 +176,33 @@ fn parse_quote_identifiers() {
...
+        }
+        _ => unreachable!(),
+    }
+
+    let sql = r#"SELECT 'I''m fine'"#;
+
+    let projection = mysql().verified_only_select(sql).projection;
+    let item = projection.get(0).unwrap();
+
+    match &item {
```
alamb, line 198: Can you please also add a test for a string ending in `\`(unterminated sequence) as well as for a non MySql dialect ?

alamb (commented):
Thank you for the contribution @vasilev-alex ! This PR looks close

tests/sqlparser_mysql.rs
vasilev-alex: @alamb you're right here.

Looking at the current implementation a `SingleQuotedString` value for `'I''m fine'` will be `I'm fine` as consecutive quote is removed (added back in the formatter). 

I can use the same approach (which seems totally correct) and a `SingleQuotedString` value for `'I\'m fine'` will 'be the same as above `I'm fine`. The only problem here is that this test will be failing
https://github.com/sqlparser-rs/sqlparser-rs/blob/9569d1b21596c150643a212d09607aacd8f60677/src/test_utils.rs#L95

as `SELECT 'I\'m fine'` will never be equals to `SELECT 'I''m fine'`. Also, doesn't seem like we can tweak the formatter as it's unaware of the context

https://github.com/sqlparser-rs/sqlparser-rs/blob/9569d1b21596c150643a212d09607aacd8f60677/src/ast/value.rs#L148

alamb: Are you saying that the `Display` impl from`to_string()` function is what is adding `\\` back? 

if so, perhaps we could add a check like 

```rust
if let Value::SingleQuotedString(value) = value {
  assert_eq!("I'm fine", value)
} else { 
  unreachable!()
}
```

to show that the value has been properly un-escaped?

tests/sqlparser_mysql.rs
```diff
@@ -176,19 +178,36 @@ fn parse_quote_identifiers() {
     }
 }
 
+#[test]
+fn parse_unterminated_escape() {
+    let sql = r#"SELECT 'I\'m not fine\'"#;
```
alamb, line 183: 👍

tests/sqlparser_mysql.rs
```diff
@@ -176,19 +178,36 @@ fn parse_quote_identifiers() {
...
-        SelectItem::UnnamedExpr(Expr::Value(value)) => {
-            assert_eq!("'I\\'m fine'", value.to_string());
-        }
+    match stmt {
+        Statement::Query(query) => match query.body {
+            SetExpr::Select(value) => {
+                let expr = expr_from_projection(only(&value.projection));
+                assert_eq!(
+                    *expr,
+                    Expr::Value(Value::SingleQuotedString("I'm fine".to_string()))
```
alamb, line 204: 👍

alamb (approved):
Looks great -- thank you @vasilev-alex

PR #378, opened 2021-12-11
Add `FunctionArgExpr` and remove `Expr::[Qualified]Wildcard`,

panarch:
There is no use case of `Expr::Wildcard` and `Expr::QualifiedWildcard` only except for function arguments.
Add `FunctionArgExpr` to have `Wildcard` and `QualifiedWildcard`.
Remove `Expr::Wildcard` and `Expr::QualifiedExpr`.

There is `WildcardExpr` which is only being used in parser module.
Both `SelectItem` and `FunctionArgExpr` use wildcards but usage is a bit different, so I added `parse_wildcard_expr` and convert `WildcardExpr` to `SelectItem` or `FunctionArgExpr`.

Now, no more `SELECT * + * FROM foo;` like queries are accepted.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -66,6 +66,22 @@ pub enum IsLateral {
 
 use IsLateral::*;
 
+pub enum WildcardExpr {
+    Expr(Expr),
+    QualifiedWildcard(ObjectName),
+    Wildcard,
+}
+
+impl From<WildcardExpr> for FunctionArgExpr {
+    fn from(wildcard_expr: WildcardExpr) -> Self {
+        match wildcard_expr {
+            WildcardExpr::Expr(expr) => Self::Expr(expr),
+            WildcardExpr::QualifiedWildcard(prefix) => Self::QualifiedWildcard(prefix),
+            WildcardExpr::Wildcard => Self::Wildcard,
+        }
+    }
+}
+
 impl From<TokenizerError> for ParserError {
     fn from(e: TokenizerError) -> Self {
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -3262,18 +3301,15 @@ impl<'a> Parser<'a> {
...
-        } else if let Expr::QualifiedWildcard(prefix) = expr {
-            Ok(SelectItem::QualifiedWildcard(ObjectName(prefix)))
-        } else {
-            // `expr` is a regular SQL expression and can be followed by an alias
-            if let Some(alias) = self.parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)? {
-                Ok(SelectItem::ExprWithAlias { expr, alias })
-            } else {
-                Ok(SelectItem::UnnamedExpr(expr))
-            }
+        match self.parse_wildcard_expr()? {
```
alamb, line 3304: I think this is roughly what postgres does too (explicitly allow `*` / `A_Star` nodes in certain places rather than anywhere a generic Expr would be allowed)

https://github.com/postgres/postgres/blob/170aec63cd7139b453c52ad52bbeb83993faa31d/src/backend/parser/gram.y#L11479-L11497

alamb (approved):
> Now, no more SELECT * + * FROM foo; like queries are accepted.

Nice -- thank you @panarch 

I went through this PR carefully and it makes sense to me.

cc @houqp  @Dandandan @andygrove

alamb:
Thanks again @panarch

Issue #380, opened 2021-12-13
Will the key in MapAccess support Express?


monadbobo:
The index of an array in ClickHouse can be an function such as :

```sql
string.values[indexOf(string.names, 'endpoint')]='/health'
```

PS: If sqlparser wants to support this syntax,I can submit a PR to support it.

houqp:
Looks like a reasonable feature to add? ping @alamb @Dandandan @Jimexist

alamb:
I agree extending support for MapAccess sounds like a good idea to me 👍

LucaCappelletti94:
Closing as completed in https://github.com/apache/datafusion-sqlparser-rs/pull/382

PR #382, opened 2021-12-14
Enable map access for function, add ClickHouse dialect

monadbobo:
#380


src/ast/mod.rs
```diff
@@ -294,8 +294,11 @@ impl fmt::Display for Expr {
                 write!(f, "{}", column)?;
                 for k in keys {
                     match k {
-                        k @ Value::Number(_, _) => write!(f, "[{}]", k)?,
-                        Value::SingleQuotedString(s) => write!(f, "[\"{}\"]", s)?,
+                        Expr::Value(v) => match v {
+                            k @ Value::Number(_, _) => write!(f, "[{}]", k)?,
+                            Value::SingleQuotedString(s) => write!(f, "[\"{}\"]", s)?,
+                            _ => write!(f, "[{}]", k)?,
+                        },
```
alamb, line 301: Do we need to match the types of `key`? Could it all be replated with `write!(f, "[{}], k)`?

Also it would be cool to avoid the duplication. If we need a match, how about

```rust
match k { 
  k@ Expr::Value(Value::Number(_, _) => write!(f, "[{}]", k)?,
  Expr::Value(Value::SingleQuotedString(s)) => write!(f, "[\"{}\"]", s)?,
  _ => write!(f, "[{}]", k)?,
}
```

alamb (approved):
Looks good -- thank you @monadbobo 
I left some comments but I don't think they are critical

src/ast/mod.rs
monadbobo: Thank you for your reply, I'll update it.

alamb:
@monadbobo  it seems like the tests no longer pass on this PR

monadbobo:
> @monadbobo it seems like the tests no longer pass on this PR

Yes, my problem, I fixed the compilation error after merging the main branch code.

alamb:
Thanks again @monadbobo

gamife:
Perhaps a corresponding dialect should also be added in "examples/cli.rs".

src/ast/mod.rs
```diff
@@ -284,8 +284,8 @@ impl fmt::Display for Expr {
                 write!(f, "{}", column)?;
                 for k in keys {
                     match k {
-                        k @ Value::Number(_, _) => write!(f, "[{}]", k)?,
-                        Value::SingleQuotedString(s) => write!(f, "[\"{}\"]", s)?,
+                        k @ Expr::Value(Value::Number(_, _)) => write!(f, "[{}]", k)?,
+                        Expr::Value(Value::SingleQuotedString(s)) => write!(f, "[\"{}\"]", s)?,
```
gamife, line 288: I have a question why don't we use Value::DoubleQuotedString here

alamb: I don't have an answer 🤷

PR #383, opened 2021-12-15
Fix #369 - Add Support in Date/Time keywords

### LINKED ISSUE #369: Trino - Add Support in Date/Time keywords
The following functions are not supported:

current_timestamp 
current_date

yuval-illumex:
Fix https://github.com/sqlparser-rs/sqlparser-rs/issues/369

Add Support for:
CURRENT_TIMESTAMP
CURRENT_DATE
CURRENT_TIME

Which are function without args.


tests/sqlparser_common.rs
```diff
@@ -3620,3 +3620,18 @@ fn verified_only_select(query: &str) -> Select {
 fn verified_expr(query: &str) -> Expr {
     all_dialects().verified_expr(query)
 }
+
+#[test]
+fn parse_time_functions() {
```
alamb, line 3625: I think we should also test:
1.  the no `()` syntax as well (e. g `SELECT CURRENT_TIMESTAMP`)
2. test coverage for CURRENT_TIMESTAMP, CURRENT_TIME, CURRENT_DATE

alamb (commented):
I double checked that both mysql and postgres indeed do treat these specially: 
👍 
Thanks @yuval-illumex 

```sql
mysql> select current_timestamp;
+---------------------+
| current_timestamp   |
+---------------------+
| 2021-12-17 12:30:40 |
+---------------------+
1 row in set (0.00 sec)

mysql> select current_time;
+--------------+
| current_time |
+--------------+
| 12:30:45     |
+--------------+
1 row in set (0.00 sec)

mysql> select current_date;
+--------------+
| current_date |
+--------------+
| 2021-12-17   |
+--------------+
1 row in set (0.00 sec)

mysql> ^DBye
(arrow_dev) alamb@MacBook-Pro-2:~/Software/arrow-datafusion$ psql
psql (14.1)
Type "help" for help.

alamb=# select current_timestamp;
       current_timestamp       
-------------------------------
 2021-12-17 12:30:59.547784-05
(1 row)

alamb=# select current_time;
    current_time    
--------------------
 12:31:02.464701-05
(1 row)

alamb=# select current_date;
 current_date 
--------------
 2021-12-17
(1 row)
```

tests/sqlparser_common.rs
yuval-illumex: @alamb I totally agree, the problem is when I try to test the following:

```
let sql = "SELECT CURRENT_TIMESTAMP";
let select = verified_only_select(sql);
```

I got the following error:

```
thread 'parse_time_functions' panicked at 'assertion failed: `(left == right)`
  left: `"SELECT CURRENT_TIMESTAMP"`,
 right: `"SELECT CURRENT_TIMESTAMP()"`', src/test_utils.rs:95:13
```

In the one_statement_parses_to function, The to_string() function of the command returns it with () (Which makes sense, because it's a function without arguments) and I'm not sure how to tackle this one. Would love for some help.

alamb: perhaps you could use something like `one_statement_parses_to()` instead of `verified_only_select()`

yuval-illumex: @alamb Updated, Thanks!

alamb (approved):
Looks good -- thanks @yuval-illumex

houqp (approved):

alamb:
These appear to have been included in https://github.com/sqlparser-rs/sqlparser-rs/pull/391 so closing this one down

PR #385, opened 2021-12-16
Support Order of `OFFSET` and `LIMIT`

### LINKED ISSUE #384: Trino - LIMIT and OFFSET order is diffrent
While in databases like postgres the order is LIMIT and then OFFSET,
In Trino it's the other way around - OFFSET and then LIMIT.

https://trino.io/docs/current/sql/select.html

yuval-illumex:
Fix https://github.com/sqlparser-rs/sqlparser-rs/issues/384

Support OFFSET LIMIT and LIMIT OFFSET

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -357,7 +357,7 @@ impl<'a> Parser<'a> {
                     self.prev_token();
                     Ok(Expr::Value(self.parse_value()?))
                 }
-                Keyword::CURRENT_TIMESTAMP => self.parse_current_timestamp(ObjectName(vec![w.to_ident()])),
+                Keyword::CURRENT_TIMESTAMP | Keyword::CURRENT_TIME | Keyword::CURRENT_DATE => self.parse_time_functions(ObjectName(vec![w.to_ident()])),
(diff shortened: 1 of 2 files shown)


alamb:
Note these is a similar PR from @QuantumBear https://github.com/sqlparser-rs/sqlparser-rs/pull/370

src/parser.rs
```diff
@@ -2294,17 +2294,18 @@ impl<'a> Parser<'a> {
...
-            };
+            let mut limit = None;
+            let mut offset = None;
 
-            let offset = if self.parse_keyword(Keyword::OFFSET) {
-                Some(self.parse_offset()?)
-            } else {
-                None
-            };
+            for _x in 0..2 {
```
alamb, line 2300: as written this will also parse `LIMIT LIMIT` and `OFFSET OFFSET` I think.

alamb (commented):
As I don't have permissions to fix this PR, I made a new one based on this change: https://github.com/sqlparser-rs/sqlparser-rs/pull/413

Thanks for the contribution @yuval-illumex  and I am sorry for the delay in reviewing

PR #387, opened 2021-12-20
add SUPER keyword

flaneur2020:
the `SUPER` keyword exists in the MySQL dialect which indicates a privilege on some system admin commands like kill queries, set global variables etc.

i'm not sure whether the keywords set is accepting the keywords of all the dialects or not. thank you for advices in advance!


houqp (approved):

alamb:
Thanks @flaneur2020

PR #388, opened 2021-12-21
add support for array expression and map index key as expression

poonai:
This PR adds following support
-  Array expression eg: `select (array['Yes', 'No', 'Maybe'])`
-  `MapAccess` expression to have key expression  eg: `select (array['Yes', 'No', 'Maybe'])[floor(random() * 3 + 1)]`

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -352,6 +352,10 @@ impl<'a> Parser<'a> {
 
             expr = self.parse_infix(expr, next_precedence)?;
         }
+
+        if self.consume_token(&Token::LBracket) {
+            return self.parse_map_access(expr);
+        }
         Ok(expr)
     }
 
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -1180,7 +1183,7 @@ impl<'a> Parser<'a> {
             Token::Mul | Token::Div | Token::Mod | Token::StringConcat => Ok(40),
             Token::DoubleColon => Ok(50),
             Token::ExclamationMark => Ok(50),
-            Token::LBracket | Token::RBracket => Ok(10),
+            Token::LBracket => Ok(51),
```
poonai, line 1186: The value is bumped up because array index doesn't have right have operand. So, I bumped up to the highest value

nickolay: I know this was added before you, but it seems inappropriate to assign a precedence to a square bracket at all, instead of parsing it directly, like we do with [parens](https://github.com/sqlparser-rs/sqlparser-rs/blob/9569d1b21596c150643a212d09607aacd8f60677/src/parser.rs#L488).

poonai: I had the same thought. I made it in a way, so that it can fit in the existing code. Now, I've removed the precedence.

monadbobo:
Hi, if ARRAY is used as a keyword, is it possible to handle the array function, for example in clickhouse you can define an array like [this](https://clickhouse.com/docs/en/sql-reference/data-types/array/):    `array(1, 2)`

poonai:
Closing this. will raise one later after clearing the conflicts

PR #390, opened 2021-12-23
Allow parentheses in GROUP BY

yuval-illumex:
In TrinoDB, GROUP BY Can have parentheses with multiple values.

Example Query:

```
SELECT orderstatus,orderpriority,COUNT(*)
FROM tpch.sf1.orders 
GROUP BY (orderstatus,orderpriority)
LIMIT 10;
```


alamb (commented):
thank you for the contribution -- sorry for the delay in response @yuval-illumex 

So I was playing around with postgres and there is a difference between `col1, col2` and `(col1, col2)`. Specifically, I think `(col1, col2)` is treated as some tuple type. 

I think this PR effectively strips off the parenthesis

```sql
alamb=# select (t1_id, t1_name), count(*) from t1 group by (t1_id, t1_name);
  row   | count 
--------+-------
 (44,d) |     1
 (77,e) |     1
 (11,a) |     1
 (22,b) |     1
 (33,c) |     1
(5 rows)

alamb=# select t1_id, t1_name, count(*) from t1 group by t1_id, t1_name;
 t1_id | t1_name | count 
-------+---------+-------
    44 | d       |     1
    77 | e       |     1
    11 | a       |     1
    22 | b       |     1
    33 | c       |     1
(5 rows)
```

Turns out that mysql just rejects this syntax

```sql
mysql> select (t1_id, t1_name), count(*) from t1 group by (t1_id, t1_name);
ERROR 1241 (21000): Operand should contain 1 column(s)
```
I didn't catch it quite in time, but the same thing is true for distinct: https://github.com/sqlparser-rs/sqlparser-rs/pull/391

I don't have a `trino` installation around to test -- what does trino do with the syntax above?

yuval-illumex:
Thanks @alamb 

The following syntax works in Trino:

image


This is the default schemas and tables coming with tirno, so you can tun Trino with docker to play with using this query.

I'm more worried about GroupBy rather then Select - I think the parentheses are more common in GroupBy (and Distinct)

alamb:
Thanks @yuval-illumex  -- it looks to me like in this case too Trino is treating `GROUP BY (a, b)` as `GROUP BY {set_expr}` while the change in this PR effectively ignores the parenthesis and treats `GROUP BY (a, b)` as `GROUP BY a, b`

alamb:
@yuval-illumex  can you please look into parsing `GROUP BY (a, b)` as a tuple / row expr? 
https://trino.io/docs/current/language/types.html?highlight=data%20type#row

I think you may be able to `parse_tuple`: use https://github.com/sqlparser-rs/sqlparser-rs/blob/2614576dbf140824cb659bcc7026b21bd1b82f0a/src/parser.rs#L652

yuval-illumex:
@alamb parse_tuple do works, but I could use some help.
I don't understand how it goes with `parse_group_by_expr`:
https://github.com/sqlparser-rs/sqlparser-rs/blob/2614576dbf140824cb659bcc7026b21bd1b82f0a/src/parser.rs#L623

It parses expressions, and I'm not sure how to identify when it should be an expression and when should it be a tuple. as the function must return an expression.

alamb:
> @alamb parse_tuple do works, but I could use some help.

Looking into it

alamb:
@yuval-illumex  this is what I had in mind (it should cover all paren cases): https://github.com/sqlparser-rs/sqlparser-rs/pull/414#

PR #397, opened 2021-12-30
fix: export all methods of parser

neverchanje:
Almost all methods of parser are exported to the public except some that were fixed in this PR.
Exporting all methods can help databases implement their custom statement which may not align with postgres dialect, but still reuse parser's existing code.


alamb (approved):

alamb:
Thank you @neverchanje

PR #408, opened 2022-01-28
support MySQL double quoted string

daiguadaidai:
MySQL double quoated should be string type, like single quoated
example:
```
SELECT "abc_123"
```

`"abc_123"` should be string type, not ident(column name)

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1093,7 +1093,7 @@ mod tests {
         assert_eq!(
             tokenizer.tokenize(),
             Err(TokenizerError {
-                message: "Unterminated string literal".to_string(),
+                message: "Unterminated string literal with single quoted".to_string(),
                 line: 1,
                 col: 8
             })
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -49,7 +49,7 @@ pub enum Token {
     Char(char),
     /// Single quoted string: i.e: 'string'
     SingleQuotedString(String),
-    /// Double quoted string: i.e: 'string'
+    /// Double quoted string: i.e: "string"
     DoubleQuotedString(String),
     /// "National" string literal: i.e: N'string'
     NationalStringLiteral(String),
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -49,6 +49,8 @@ pub enum Token {
     Char(char),
     /// Single quoted string: i.e: 'string'
     SingleQuotedString(String),
+    /// Double quoted string: i.e: 'string'
+    DoubleQuotedString(String),
     /// "National" string literal: i.e: N'string'
     NationalStringLiteral(String),
     /// Hexadecimal string literal: i.e.: X'deadbeef'
@@ -149,6 +151,7 @@ impl fmt::Display for Token {
             Token::Number(ref n, l) => write!(f, "{}{long}", n, long = if *l { "L" } else { "" }),
             Token::Char(ref c) => write!(f, "{}", c),
             Token::SingleQuotedString(ref s) => write!(f, "'{}'", s),
+            Token::DoubleQuotedString(ref s) => write!(f, "{:#?}", s),
             Token::NationalStringLiteral(ref s) => write!(f, "N'{}'", s),
             Token::HexStringLiteral(ref s) => write!(f, "X'{}'", s),
             Token::Comma => f.write_str(","),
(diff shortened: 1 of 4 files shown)


alamb:
Hi @daiguadaidai  -- this PR is not passing CI checks -- can you please fix the issues?

tests/sqlparser_mysql.rs
```diff
@@ -222,6 +222,39 @@ fn parse_escaped_string() {
     }
 }
 
+#[test]
+fn parse_escaped_double_quoated_string() {
```
alamb, line 226:
```suggestion
fn parse_escaped_double_quoted_string() {
```

daiguadaidai:
already fix

src/tokenizer.rs
```diff
@@ -49,6 +49,8 @@ pub enum Token {
     Char(char),
     /// Single quoted string: i.e: 'string'
     SingleQuotedString(String),
+    /// Double quoted string: i.e: "string"
```
alamb, line 52:
```suggestion
    /// Double quoted string: i.e: "string" (mysql dialect)
```

src/tokenizer.rs
```diff
@@ -149,6 +151,7 @@ impl fmt::Display for Token {
             Token::Number(ref n, l) => write!(f, "{}{long}", n, long = if *l { "L" } else { "" }),
             Token::Char(ref c) => write!(f, "{}", c),
             Token::SingleQuotedString(ref s) => write!(f, "'{}'", s),
+            Token::DoubleQuotedString(ref s) => write!(f, "{:?}", s),
```
alamb, line 154: Shouldn't the display also print out the double quotes? Something like:

```suggestion
            Token::DoubleQuotedString(ref s) => write!(f, r#""{:?}""#, s),
```

tests/sqlparser_mysql.rs
```diff
@@ -222,6 +222,39 @@ fn parse_escaped_string() {
     }
 }
 
+#[test]
+fn parse_escaped_double_quoted_string() {
+    let sql = r#"SELECT "I\"m fine""#;
+
+    let stmt = mysql().one_statement_parses_to(sql, "");
```
alamb, line 229: Shouldn't this be something like `one_statement_parses_to(sql, sql);`?

tests/sqlparser_mysql.rs
```diff
@@ -222,6 +222,39 @@ fn parse_escaped_string() {
...
+                let expr = expr_from_projection(only(&value.projection));
+                assert_eq!(
+                    *expr,
+                    Expr::Value(Value::DoubleQuotedString("I\"m fine".to_string()))
+                );
+            }
+            _ => unreachable!(),
+        },
+        _ => unreachable!(),
+    };
```
alamb, line 243: You could potentially write this in a single match statement like


```suggestion
    match stmt {
        Statement::Query({ 
          body: SetExpr::Select(value),
          ..
        } => {
                let expr = expr_from_projection(only(&value.projection));
                assert_eq!(
                    *expr,
                    Expr::Value(Value::DoubleQuotedString("I\"m fine".to_string()))
                );
        },
        _ => unreachable!(),
    };
```

tests/sqlparser_mysql.rs
```diff
@@ -222,6 +222,39 @@ fn parse_escaped_string() {
...
+    let item = projection.get(0).unwrap();
+
+    match &item {
+        SelectItem::UnnamedExpr(Expr::Value(value)) => {
+            assert_eq!(*value, Value::DoubleQuotedString("I\"m fine".to_string()));
+        }
+        _ => unreachable!(),
+    }
+}
+
```
alamb, line 257: I think we should also add tests for the following cases
1. single quoted strings elsewhere in the query (not just the select list)
2. a double quoted string with no escapes
3. A double quoted string with at least two escaped `"`
4. An incorrectly quoted string (e.g. `SELECT "f"" from foo`



```
mysql> select * from bar where x != "foo";
+------+
| x    |
+------+
|    1 |
+------+
1 row in set, 1 warning (0.00 sec)

```

alamb: Also a test that `SELECT "foo" ` is treated as `SELECT foo` in other dialects (like postgres)

alamb (commented):
Thank you for the contribution @daiguadaidai  -- this is a good start. I left some review comments.

Today I learned something about MySQL!

```sql
mysql> select "f" from bar;
+---+
| f |
+---+
| f |
| f |
+---+
2 rows in set (0.00 sec)
```

src/tokenizer.rs
andygrove: It is also worth noting that in ANSI SQL, double quotes are used for identifiers and not for strings. For example, `monthly_income * 12 AS "Annual Income"`.

alamb:
Thanks again for this PR. I am closing it for now to clean up the PR list. Please reopen if that was a mistake.

PR #409, opened 2022-02-03
Add support for `FROM <filename>`, `DELIMITER`, and `CSV HEADER` options for `COPY` command

poonai:
This PR adds support for delimiter and csv header options in copy command

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1051,7 +1051,7 @@ impl fmt::Display for Statement {
                 values,
                 delimiter,
                 filename,
-                csv_header
+                csv_header,
             } => {
                 write!(f, "COPY {}", table_name)?;
                 if !columns.is_empty() {
@@ -1083,7 +1083,7 @@ impl fmt::Display for Statement {
                         }
                     }
                 }
-                if filename.is_none(){
+                if filename.is_none() {
                     write!(f, "\n\\.")?;
                 }
                 Ok(())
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -670,6 +670,12 @@ pub enum Statement {
         columns: Vec<Ident>,
         /// VALUES a vector of values to be copied
         values: Vec<Option<String>>,
+        /// file name of the data to be copied from
+        filename: Option<Ident>,
+        /// delimiter character
+        delimiter: Option<Ident>,
+        /// CSV HEADER
```
alamb, line 677: I didn't think `CSV HEADER` was part of the syntax. Postgres seems to only support `HEADER`:

https://www.postgresql.org/docs/13/sql-copy.html

I think CSV is used as part of the FORMAT option, like

```sql
COPY ... FORMAT CSV ...
```

alamb (commented):
Thanks for the contribution @poonai . I had a question about `CSV HEADER` 

the lint CI failures should be resolved if you `merge` up to main

src/ast/mod.rs
poonai: compactability

If you scroll down, you'll find the compatibility section. There you can find CSV HEADER spec. 

Sure, I merge the main

alamb (approved):
Thanks again @poonai  for the contribution. If/when people want further support for COPY options, we can always add it as a follow on PR

PR #411, opened 2022-02-04
fix: Handle double quotes inside quoted identifiers correctly

### LINKED ISSUE #410: Quoted identifiers do not handle escaped quotes
Looking at the code https://github.com/sqlparser-rs/sqlparser-rs/blob/ea0eb1be617635b29d49d712be22a9a568d02366/src/tokenizer.rs#L417-L430 and writing a naive test it doesn't seem like quoted identifiers `"quoted "" identifier"` handle escaped quotes (`""`) in anyway. Unless I am missing something I'd be happy to make a PR to implement these.

https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#delimited-identifier

```rust

#[test]
fn parse_quoted_identifier() {
    pg_and_generic().verified_stmt(r#"SELECT "quoted "" ident""#);
}

---- parse_quoted_identifier stdout ----
thread 'parse_quoted_identifier' panicked at 'assertion failed: `(left == right)`
  left: `"SELECT \"quoted \"\" ident\""`,
 right: `"SELECT \"quoted \" AS \" ident\""`', src/test_utils.rs:95:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Marwes:
This fixes #410 for standard SQL, however I don't know enough about other dialects to know if they
handle this differently. May need more extensive testing as well.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1304,11 +1304,11 @@ mod tests {
         let tokens = tokenizer.tokenize().unwrap();
         let expected = vec![
             Token::Whitespace(Whitespace::Space),
-            Token::make_word(r#"a " b"#.into(), Some('"')),
+            Token::make_word(r#"a " b"#, Some('"')),
             Token::Whitespace(Whitespace::Space),
-            Token::make_word(r#"a ""#.into(), Some('"')),
+            Token::make_word(r#"a ""#, Some('"')),
             Token::Whitespace(Whitespace::Space),
-            Token::make_word(r#"c """#.into(), Some('"')),
+            Token::make_word(r#"c """#, Some('"')),
             Token::Whitespace(Whitespace::Space),
         ];
         compare(expected, tokens);
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -418,8 +418,26 @@ impl<'a> Tokenizer<'a> {
                 quote_start if self.dialect.is_delimited_identifier_start(quote_start) => {
                     chars.next(); // consume the opening quote
                     let quote_end = Word::matching_end_quote(quote_start);
-                    let s = peeking_take_while(chars, |ch| ch != quote_end);
-                    if chars.next() == Some(quote_end) {
+                    let mut last_char = None;
+                    let s = {
+                        let mut s = String::new();
+                        while let Some(ch) = chars.next() {
+                            if ch == quote_end {
+                                if chars.peek() == Some(&quote_end) {
+                                    chars.next();
+                                    s.push(ch);
+                                } else {
+                                    last_char = Some(quote_end);
+                                    break;
+                                }
+                            } else {
+                                s.push(ch);
+                            }
+                        }
+                        s
+                    };
+
+                    if last_char == Some(quote_end) {
                         Ok(Some(Token::make_word(&s, Some(quote_start))))
                     } else {
                         self.tokenizer_error(format!(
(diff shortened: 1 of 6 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -834,6 +834,11 @@ fn parse_comments() {
     }
 }
 
+#[test]
+fn parse_quoted_identifier() {
+    pg_and_generic().verified_stmt(r#"SELECT "quoted "" ident""#);
+}
```
alamb, line 840: This is definitely how postgres handles things

```sql
alamb=# SELECT "quoted "" ident" from t1;
ERROR:  column "quoted " ident" does not exist
LINE 1: SELECT "quoted "" ident" from t1;
             
```

alamb: However, msql treats this differently

```sql
Database changed
mysql> SELECT "quoted "" ident" from t1;
+----------------+
| quoted " ident |
+----------------+
| quoted " ident |
| quoted " ident |
| quoted " ident |
| quoted " ident |
| quoted " ident |
+----------------+
5 rows in set (0.00 sec)

mysql> SELECT "quoted  ident" from t1;
+---------------+
| quoted  ident |
+---------------+
| quoted  ident |
| quoted  ident |
| quoted  ident |
| quoted  ident |
| quoted  ident |
+---------------+
5 rows in set (0.00 sec)

```

tests/sqlparser_postgres.rs
```diff
@@ -834,6 +834,11 @@ fn parse_comments() {
     }
 }
 
+#[test]
+fn parse_quoted_identifier() {
+    pg_and_generic().verified_stmt(r#"SELECT "quoted "" ident""#);
```
alamb, line 839: perhaps we can add a test for dialects (e.g. mysql) that this isn't parsed as an identifier?

src/tokenizer.rs
```diff
@@ -418,8 +418,26 @@ impl<'a> Tokenizer<'a> {
                 quote_start if self.dialect.is_delimited_identifier_start(quote_start) => {
                     chars.next(); // consume the opening quote
                     let quote_end = Word::matching_end_quote(quote_start);
-                    let s = peeking_take_while(chars, |ch| ch != quote_end);
-                    if chars.next() == Some(quote_end) {
+                    let mut last_char = None;
```
alamb, line 421: I think this code would be easier to understand if it were pulled out into its own function (`peeking_take_while_ident`?) which could be unit tested, to ensure it handles cases such as
```
foo "" bar "" 
foo bar""
```

alamb (commented):
Thanks @Marwes  -- this is looking good. I think with a few more tests this would be good to go

tests/sqlparser_postgres.rs
Marwes: `"` is used for strings in mysql unless `ANSI_QUOTES` is enabled https://dev.mysql.com/doc/refman/8.0/en/identifiers.html . Neither double quoted strings or  `ANSI_QUOTES` seems implemented in this crate and I don't think it is up to this PR to fix that.

tests/sqlparser_postgres.rs
Marwes: I added a test for ` quoted strings in mysql

alamb (approved):
Thank you @Marwes

alamb:
I hope to work through some various other PRs that are open this week and make a release later in the week

PR #412, opened 2022-02-05
Fix new clippy errors

alamb:
Fixes new clippy lints introduced in newer `rustc` releases

Also check for missing paren in code added in https://github.com/sqlparser-rs/sqlparser-rs/pull/391 (found by clippy lint)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2649,7 +2649,7 @@ impl<'a> Parser<'a> {
         };
 
         // Not Sure if Top should be cheked here as well. Trino doesn't support TOP.
-        let is_l_parent = if distinct {
+        let is_l_paren = if distinct {
             self.consume_token(&Token::LParen)
         } else {
             false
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -2649,16 +2649,16 @@ impl<'a> Parser<'a> {
...
             self.consume_token(&Token::LParen)
         } else {
             false
         };
 
         let projection = self.parse_comma_separated(Parser::parse_select_item)?;
 
-        if is_l_parent {
-            self.consume_token(&Token::RParen);
+        if is_l_paren && !self.consume_token(&Token::RParen) {
```
alamb, line 2660: clippy was complaining (rightly) that the return value from `consume_token` was not used

tests/sqlparser_common.rs
```diff
@@ -343,6 +343,15 @@ fn parse_select_distinct_two_fields() {
     );
 }
 
+#[test]
+fn parse_select_distinct_missing_paren() {
```
alamb, line 347: test for new code

PR #391, opened 2021-12-26
Support parentheses in DISTINCT clause, CURRENT_TIMESTAMP, CURRENT_TIME, and CURRENT_DATE

yuval-illumex:
The following query works in TrinoDB:

```
SELECT DISTINCT (orderstatus,orderpriority)
FROM tpch.sf1.orders 
LIMIT 10

```

PR #414, opened 2022-02-08
Add `Expr::Tuple` and parsing support, remove special case in `DISTINCT` clause

alamb:
# Rationale:
@yuval-illumex  is trying to support parenthesized expressions like `(a, b)` in various places (see https://github.com/sqlparser-rs/sqlparser-rs/pull/390, and https://github.com/sqlparser-rs/sqlparser-rs/pull/391)

I believe these are more commonly called `tuple` or `row` types in postgres or trino. Rather than adding special case handling for parens in various places, it would be better to simply support parsing tuples in general 

# Changes
1. Support parsing tuple expressions in select statement
2. Tests for same
2. Remove special case added in #390 for `distinct` clause
2. Add tests in #391  demonstrating this code can parse paren'd expressions in the group by

# Reference:

This syntax is supported by postgres:

```sql
alamb=# select (1, 'foo');
   row   
---------
 (1,foo)
(1 row)
```

It is not supported by `mysql`:

```sql
mysql> select (1, 2);
ERROR 1241 (21000): Operand should contain 1 column(s)
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -494,7 +494,12 @@ impl<'a> Parser<'a> {
                         self.prev_token();
                         Expr::Subquery(Box::new(self.parse_query()?))
                     } else {
-                        Expr::Nested(Box::new(self.parse_expr()?))
+                        let exprs = self.parse_comma_separated(Parser::parse_expr)?;
+                        match exprs.len() {
+                            0 => unreachable!(), // parse_comma_separated ensures 1 or more
+                            1 => Expr::Nested(Box::new(exprs.into_iter().next().unwrap())),
+                            _ => Expr::Tuple(exprs),
+                        }
                     };
                 self.expect_token(&Token::RParen)?;
                 Ok(expr)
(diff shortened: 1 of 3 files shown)


tests/sqlparser_common.rs
```diff
@@ -1121,6 +1171,12 @@ fn parse_select_group_by() {
         ],
         select.group_by
     );
+
+    // Tuples can also be in the set
+    one_statement_parses_to(
```
alamb, line 1176: FYI @yuval-illumex

src/parser.rs
```diff
@@ -2680,19 +2685,8 @@ impl<'a> Parser<'a> {
             None
         };
 
-        // Not Sure if Top should be cheked here as well. Trino doesn't support TOP.
```
alamb, line 2683: this was added in #390  but is now covered by the general purpose `Tuple` parsing

e-dard (approved):
I took a whizz through this and it made sense to me 👍

yuval-illumex:
@alamb Much appreciated!!! Looks great :)

PR #418, opened 2022-02-10
feat: add FOR UPDATE/FOR SHARE clause

gamife:


houqp (approved):

src/ast/query.rs
```diff
@@ -577,6 +582,23 @@ impl fmt::Display for Fetch {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum LockType {
+    Shared,
+    Exclusive,
```
alamb, line 589: Since this is always parsed to/from `SHARE` and `FOR UPDATE` what do you think about using `Share` and `Update` as the enum variant names so that the Rust structures more closely mirror the parsed SQL structure and is more consistent with the rest of this crate?

alamb (commented):
Thank you @gamife -- other than the struct name this looks good to go.

src/ast/query.rs
gamife: I'm not insistent, It's just that `Update` doesn't look like a type of lock.  
If you still think update is better for this crate, I will make changes. Looking forward to reply :->

alamb: I think for this crate it is better to match the original SQL as closely as possible. 

Downstream crates that actually implement a database can then then map that SQL structure on to actual locking implementations (like shared / exclusive, etc)

gamife: I got it, thanks for the reply !!

baoyachi:
What progress has been made so far? It seems that it can be merged

houqp:
I don't have merge access and @alamb is on vacation right now, let's wait a little bit more :)

houqp:
actually, @Dandandan and @andygrove should be able to help review as well.

Dandandan (approved):

Dandandan:
Thanks a lot!

alamb:
🎉

alamb:
I'll try and make a sqlparser release sometime early next week too to pick up this change

PR #419, opened 2022-02-10
Support array expressions such as `ARRAY[1,2]` , `foo[1]` and `INT[][]`

gamife:
1. Array expression: ARRAY[1,2] 
2. Array index expression: foo[1]
3. Extend the array data type: INT[][] 
```sql
SELECT (CAST(ARRAY[ARRAY[2, 3]] AS INT[][]))[1][2]
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -420,6 +420,7 @@ impl<'a> Parser<'a> {
                 Keyword::TRIM => self.parse_trim_expr(),
                 Keyword::INTERVAL => self.parse_literal_interval(),
                 Keyword::LISTAGG => self.parse_listagg_expr(),
+                Keyword::ARRAY => self.parse_array_expr(),
                 Keyword::NOT => Ok(Expr::UnaryOp {
                     op: UnaryOperator::Not,
                     expr: Box::new(self.parse_subexpr(Self::UNARY_NOT_PREC)?),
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -1083,13 +1091,32 @@ impl<'a> Parser<'a> {
                 expr: Box::new(expr),
             })
         } else if Token::LBracket == tok {
+            if dialect_of!(self is PostgreSqlDialect) {
```
alamb, line 1094: 👍  https://www.postgresql.org/docs/9.1/arrays.html

tests/sqlparser_postgres.rs
```diff
@@ -774,51 +771,32 @@ fn parse_pg_regex_match_ops() {
 }
 
 #[test]
-fn parse_map_access_expr() {
```
alamb, line 777: Could you please add back the the existing test parsing the simpler expression `SELECT foo[0] FROM foos` to make sure that that syntax is still supported (it would be fine if the resulting parse tree is slightly different)

I think it is important to ensure that this PR does not introduce a regression in terms of the syntax that is supported (I don't think it does, but I would like the test to ensure that this is the case)

alamb (commented):
Thank you for the contribution @gamife

tests/sqlparser_postgres.rs
gamife: Ok. Thanks for the review!

alamb (approved):
Thank you @gamife

Issue #145, opened 2019-11-22
Add support for parsing placeholders


mitar:
Maybe I missed something, but I think placeholders are not yet supported and cannot be parsed. I would like to ask if this could be added.

Ideally, with few extensions than a regular parser:
* In regular SQL queries, only SELECT, INSERT, UPDATE queries can have placeholders. But for my use case it would be great if also other statements can have them, like `CREATE VIEW` which takes a query, and that one could have placeholders, because `CREATE VIEW` is specially processed in my case. (See https://github.com/mit-pdos/noria/issues/142 for more information.)
* Common SQL databases support position-based placeholders, I would love it some way of having keyword-based ones could be added. This then maps very well to how one can provide values with a dict structure instead of position.

benesch:
In the meantime you can apply this patch:
https://github.com/MaterializeInc/sqlparser/pull/17/files

I’ll look into getting it merged here. It works as you suggest with regards
to CREATE VIEW et al supporting expressions with placeholders. It does not
presently accept named placeholders, however.

mitar:
Yes, something like `$foobar` would be awesome to have for placeholders. (Then mapping to/from GraphQL can be made much smoother.)

schreter:
I'm also missing parameter placeholder support.

We considered using sqlparser as the parser for a (for now) "demo" Rust-based frontend to our Rust-based data store backend. Ideally, it should become the default frontend. However, to support prepared statements, placeholder support is needed. In our SQL dialect, those are specified with a question mark, as used in most "large" RDBMSs, like MS SQL, Oracle SQL, SAP HANA DB SQL, etc.

Would it be possible to add support for this as well? It should be fairly trivial extension to the patch above (though, it didn't land in the source code yet, so not sure whether it can be even applied).

Thanks.

alamb:
@gamife has contributed an implementation here: https://github.com/sqlparser-rs/sqlparser-rs/pull/420

I'll try and cut a new release in the next week or two

PR #420, opened 2022-02-11
Support prepared statement placeholder arg `?` and `$`

gamife:
#145 support `?` or `$foo` prepared statement arg placeholder.
```sql
SELECT * FROM student WHERE id = ?;
SELECT * FROM student WHERE id = $Id1;
```

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -59,6 +59,8 @@ pub enum Value {
     },
     /// `NULL` value
     Null,
+    /// `?` or `$` Prepared statement arg placeholder
+    Placeholder(String),
 }
 
 impl fmt::Display for Value {
@@ -111,6 +113,7 @@ impl fmt::Display for Value {
                 Ok(())
             }
             Value::Null => write!(f, "NULL"),
+            Value::Placeholder(v) => write!(f, "{}", v),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/ast/value.rs
```diff
@@ -59,6 +59,8 @@ pub enum Value {
     },
     /// `NULL` value
     Null,
+    /// `?` or `$` Prepared statement arg placeholder
```
alamb, line 62: I am not familiar with the `$` syntax for prepared argument placeholders. What database system do they come from?

I have only ever seen `?` used for prepared statement placeholders

alamb (commented):
Thanks (again) for the contribution @gamife

src/ast/value.rs
gamife: Here is an [example](https://www.postgresql.org/docs/current/sql-prepare.html#:~:text=pg_prepared_statements%20system%20view.-,Examples,-Create%20a%20prepared) from pgsql's document.
```sql
PREPARE fooplan (int, text, bool, numeric) AS
    INSERT INTO foo VALUES($1, $2, $3, $4);
EXECUTE fooplan(1, 'Hunter Valley', 't', 200.00);
```
However, only numbers follow the $.
Then I found that there was a need for $foo in the issue. https://github.com/sqlparser-rs/sqlparser-rs/issues/145#issuecomment-557332621

So the final product is this:
```sql
SELECT * FROM student WHERE id = $Id1;
```
Maybe there can't be a separate $placeholder here.

Thank you even more for your continued contribution. I'm just a spoon of water in the ocean :-)

alamb (approved):
👍  thanks @gamife

PR #422, opened 2022-02-18
Add ROLE keyword

flaneur2020:
this PR adds an ROLE keyword definition, so we can parse the role related statements like: https://www.postgresql.org/docs/9.0/database-roles.html


houqp (approved):

alamb:
Thanks @flaneur2020

PR #426, opened 2022-02-20
Support `IN UNNEST(expression)`

komukomo:
This PR adds support `IN UNNEST(array_expression)` for BigQuery.

```
search_value [NOT] IN value_set

value_set:
  {
    (expression[, ...])
    | (subquery)
    | UNNEST(array_expression)
  }
```
https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#in_operators

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -189,6 +189,12 @@ pub enum Expr {
         subquery: Box<Query>,
         negated: bool,
     },
+    /// `[ NOT ] IN UNNEST(array_expression)`
+    InUnnest {
+        expr: Box<Expr>,
+        array_expr: Box<Expr>,
+        negated: bool,
+    },
     /// `<expr> [ NOT ] BETWEEN <low> AND <high>`
     Between {
         expr: Box<Expr>,
@@ -334,6 +340,17 @@ impl fmt::Display for Expr {
                 expr,
                 if *negated { "NOT " } else { "" },
                 subquery
+            ),
+            Expr::InUnnest {
+                expr,
+                array_expr,
+                negated,
+            } => write!(
+                f,
+                "{} {}IN UNNEST({})",
+                expr,
+                if *negated { "NOT " } else { "" },
+                array_expr
             ),
             Expr::Between {
                 expr,
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -905,16 +905,23 @@ fn parse_in_subquery() {
 
 #[test]
 fn parse_in_unnest() {
-    let sql = "SELECT * FROM customers WHERE segment IN UNNEST(expr)";
-    let select = verified_only_select(sql);
-    assert_eq!(
-        Expr::InUnnest {
-            expr: Box::new(Expr::Identifier(Ident::new("segment"))),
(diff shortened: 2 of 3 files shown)


houqp (approved):

src/ast/mod.rs
```diff
@@ -189,6 +189,12 @@ pub enum Expr {
         subquery: Box<Query>,
         negated: bool,
     },
+    /// `[ NOT ] IN UNNEST(array_expression)`
```
alamb, line 192: I was going to suggest making this somewhat more general with any expr (rather than specific for UNEST function) like:

```rust
/// [NOT] IN <expr>
InExpr {
  expr: Box<Expr>,
  in_expr: Box<Expr>,
  negated: bool
}
```

However, I see that https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#in_operators treats it as a special case so I think it would be ok to do here as well.

Actually, looking more deeply, it seems like UNNEST actually supports `with offset` as well

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unnest_operator

tests/sqlparser_common.rs
```diff
@@ -903,6 +903,20 @@ fn parse_in_subquery() {
     );
 }
 
+#[test]
+fn parse_in_unnest() {
```
alamb, line 907: Can you please also add a test for `NOT IN UNNEST` as well?

alamb (approved):
I apologize for the late review, I have been away for the last week. Thank you for the contribution @komukomo

tests/sqlparser_common.rs
komukomo: Thanks! I will fix it!

src/ast/mod.rs
komukomo: Thank you for your valuable review.

As you said, it seems to be treated as a special case and any other functions can't used, so I did like this.

> Actually, looking more deeply, it seems like UNNEST actually supports with offset as well

I did not know about `with offset`, thank you.
However, `WITH OFFSET` seems to be available when UNNEST returns a table.
In the case of `IN UNNSET(array_expression)`, `WITH OFFSET` cannot be used because this `IN UNNEST(array_expression)` is a syntax sugar of subquery:
`IN (SELECT element FROM UNNEST(array_expression) AS element)`.

✅  `SELECT * FROM  UNNEST(ARRAY[1,2,3]) WITH OFFSET`
✅  `SELECT 1 IN UNNEST(ARRAY[1,2,3])`
❌  `SELECT 1 IN UNNEST(ARRAY[1,2,3]) WITH OFFSET`

So, if we support `with offset`, I think it is better to separate it from this PR, because the implementation and testing will be completely different. (I will try it in the future, if possible).

komukomo:
I fixed it!

src/ast/mod.rs
alamb: > So, if we support with offset, I think it is better to separate it from this PR,

I agree -- that makes sense. If anyone needs `WITH OFFSET` support it can be added as a follow on 👍

alamb:
Thanks again @komukomo

PR #429, opened 2022-03-06
Support for ClickHouse array types (e.g. [1,2,3])

monadbobo:


tests/sqpparser_clickhouse.rs
```diff
@@ -40,16 +41,32 @@ fn parse_map_access_expr() {
...
-            })]
+                distinct: false,
+            })],
         },
         expr_from_projection(only(&select.projection)),
     );
 }
 
+#[test]
+fn parse_array_expr() {
```
alamb, line 55: Can you please also add a test for the `named` variant.

I think you added support for `SELECT ARRAY['1', '2'])` as well but it isn't covered in the tests

alamb (commented):
Thank you for the contribution @monadbobo

alamb:
FYI I hope to make the 0.15.0 release in the next day or two, so hopefully we can get this PR in for that release;

tests/sqpparser_clickhouse.rs
monadbobo: named array has been tested in sqlparser_postgres.rs, so I didn't test named array separately here。

alamb: Found it https://github.com/sqlparser-rs/sqlparser-rs/blob/1da49c1/tests%2Fsqlparser_postgres.rs#L826-L843 👍 

Thanks

alamb:
🤔  this change seems to have broken a downstream test in DataFusion : https://github.com/apache/arrow-datafusion/pull/1956

In particular, sqlparser now rejects

```sql
SELECT array(c1, cast(c2 as varchar)) FROM test
```

As invalid (in this case `array` is a function name). I'll see if I can get a fix

PR #441, opened 2022-03-17
Add support for `COPY TO`

matthewmturner:
Syntax reference: https://www.postgresql.org/docs/current/sql-copy.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -736,7 +736,7 @@ pub enum Statement {
         file_format: Option<FileFormat>,
         source: Box<Query>,
     },
-    Copy {
+    CopyFrom {
         /// TABLE
         table_name: ObjectName,
         /// COLUMNS
@@ -744,6 +744,20 @@ pub enum Statement {
         /// VALUES a vector of values to be copied
         values: Vec<Option<String>>,
         /// file name of the data to be copied from
+        filename: Option<Ident>,
+        /// delimiter character
+        delimiter: Option<Ident>,
+        /// CSV HEADER
+        csv_header: bool,
+    },
+    CopyTo {
+        /// TABLE
+        table_name: ObjectName,
+        /// COLUMNS
+        columns: Vec<Ident>,
+        /// VALUES a vector of values to be copied
+        values: Vec<Option<String>>,
+        /// file name of the data to be copied to
         filename: Option<Ident>,
         /// delimiter character
         delimiter: Option<Ident>,
@@ -1136,7 +1150,7 @@ impl fmt::Display for Statement {
                 }
             }
 
-            Statement::Copy {
+            Statement::CopyFrom {
                 table_name,
                 columns,
                 values,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1159,12 +1159,10 @@ impl fmt::Display for Statement {
                     } else {
                         write!(f, " FROM {}", name)?;
                     }
+                } else if *to {
+                    write!(f, " TO stdin ")?
                 } else {
-                    if *to {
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -750,6 +750,20 @@ pub enum Statement {
         /// CSV HEADER
         csv_header: bool,
     },
+    CopyTo {
```
alamb, line 753: It looks to me like both `CopyFrom` and `CopyTo` share all the same fields.

What would you think about just adding a new field to `Copy` rather than making `CopyTo` and `CopyFrom`?

Something like:


```rust
  Copy {
        /// TABLE
        table_name: ObjectName,

...
        /// CSV HEADER
        csv_header: bool,
        /// If true, represents COPY TO, else COPY FROM
        to: bool,        
    },
```

It looks to me like the only difference between `COPY TO` and `COPY FROM` is that COPY FROM also supports a `WHERE` clause in postgres.

matthewmturner: Indeed, they are effectively the same asides from COPY TO taking a query as well. and im definitely fine with updating to your proposal.  I was naively under the impression that each statement should have its own variant.

btw below is from postgres.
```
COPY table_name [ ( column_name [, ...] ) ]
    FROM { 'filename' | STDIN }
    [ [ WITH ] ( option [, ...] ) ]

COPY { table_name [ ( column_name [, ...] ) ] | ( query ) }
    TO { 'filename' | STDOUT }
    [ [ WITH ] ( option [, ...] ) ]

where option can be one of:

    FORMAT format_name
    OIDS [ boolean ]
    DELIMITER 'delimiter_character'
    NULL 'null_string'
    HEADER [ boolean ]
    QUOTE 'quote_character'
    ESCAPE 'escape_character'
    FORCE_QUOTE { ( column_name [, ...] ) | * }
    FORCE_NOT_NULL ( column_name [, ...] )
    ENCODING 'encoding_name'
```

alamb: > Indeed, they are effectively the same asides from COPY TO taking a query as well. and im definitely fine with updating to your proposal. I was naively under the impression that each statement should have its own variant.

I guess I was thinking the `COPY` was the statement rather than `COPY FROM` and `COPY TO` being separate statements

matthewmturner:
I think this is good now

src/ast/mod.rs
```diff
@@ -1143,14 +1145,22 @@ impl fmt::Display for Statement {
...
+                to,
             } => {
                 write!(f, "COPY {}", table_name)?;
                 if !columns.is_empty() {
                     write!(f, " ({})", display_comma_separated(columns))?;
                 }
 
                 if let Some(name) = filename {
-                    write!(f, " FROM {}", name)?;
+                    // println!("To: {}", *to);
```
alamb, line 1156:
```suggestion
```

src/parser.rs
```diff
@@ -2233,7 +2233,12 @@ impl<'a> Parser<'a> {
     pub fn parse_copy(&mut self) -> Result<Statement, ParserError> {
         let table_name = self.parse_object_name()?;
         let columns = self.parse_parenthesized_column_list(Optional)?;
-        self.expect_keywords(&[Keyword::FROM])?;
+        let to_or_from = self.expect_one_of_keywords(&[Keyword::FROM, Keyword::TO])?;
+        let to: bool = match to_or_from {
```
alamb, line 2237: 👍

matthewmturner:
> LGTM -- I had some small suggestions but overall looks great. Thank you @matthewmturner

ok will fix shortly.

matthewmturner:
Updated

alamb:
Thanks again @matthewmturner

PR #443, opened 2022-03-21
feat: add support for parsing mysql intervals

kyle-mccarthy:
[MySQL handles intervals differently from the other dialects](https://dev.mysql.com/doc/refman/8.0/en/expressions.html#temporal-intervals). This PR adds support for parsing those intervals when using the MySQL dialect. 

One thing to note, in the tests, I have the following case:

```rust
mysql().one_statement_parses_to(
    "SELECT NOW() + INTERVAL CAST(6/4 AS DECIMAL(3,1)) HOUR",
    "SELECT NOW() + INTERVAL CAST(6 / 4 AS NUMERIC(3,1)) HOUR",
);
```

This may be wrong... The destination type probably should remain as `DECIMAL(3,1)`. I say "may" because a [bug was reported](https://bugs.mysql.com/bug.php?id=84558) to the MySQL team in 2017; however, it isn't clear if this bug or documentation issue, and there hasn't been any progress since then.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -14,7 +14,7 @@
 //! (commonly referred to as Data Definition Language, or DDL)
 
 #[cfg(not(feature = "std"))]
-use alloc::{boxed::Box, string::String, string::ToString, vec::Vec};
+use alloc::{boxed::Box, string::String, vec::Vec};
 use core::fmt;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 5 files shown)


alamb:
Hi @kyle-mccarthy  -- I am sorry it looks like this PR never got reviewed and it has now bitrotted. If you would like to pursue this can you please update the PR? 

alamb:
Closing as stale -- please reopen if you still plan to work on it

PR #446, opened 2022-03-25
Add support for more postgres `COPY` options

wangrunji0408:
This PR implements almost all `COPY` syntax in [current version](https://www.postgresql.org/docs/current/sql-copy.html) and legacy syntax [before v9.0](https://www.postgresql.org/docs/8.4/sql-copy.html).

Three enums (`CopyOption`, `CopyLegacyOption` and `CopyLegacyCsvOption`) were introduced and some fields in `Statment::Copy {..}` were changed.


src/parser.rs
```diff
@@ -2233,53 +2233,166 @@ impl<'a> Parser<'a> {
...
+            CopyTarget::Program {
+                command: self.parse_literal_string()?,
+            }
+        } else {
+            CopyTarget::File {
+                filename: self.parse_literal_string()?,
             }
-            break;
+        };
+        let _ = self.parse_keyword(Keyword::WITH);
```
alamb, line 2254: Shouldn't this be

```suggestion
        self.parse_keyword(Keyword::WITH)?;
```

So that the error is properly returned?

tests/sqlparser_postgres.rs
```diff
@@ -402,121 +402,201 @@ PHP	₱ USD $
...
             values: vec![],
-            delimiter: None,
-            csv_header: false,
-            to: false
         }
     );
+}
 
-    let stmt = pg().verified_stmt("COPY users FROM 'data.csv' DELIMITER ','");
+#[test]
```
alamb, line 472: Would you mind keeping the existing test cases (to show how this PR changes their behavior) and the tests for the new behavior as new cases?

alamb (commented):
This looks great @wangrunji0408  -- thank you so much for the contribution.

I had a few comments but overall this is looking wonderful 🏅

src/parser.rs
wangrunji0408: The `WITH` keyword is optional, so the error should be ignored here.

tests/sqlparser_postgres.rs
wangrunji0408: restored in 13d0f3f

alamb (approved):
Looks great -- thank you @wangrunji0408

alamb:
I believe the CI check is failing as the latest version of rustfmt (in nightly) changed a bit

@wangrunji0408  can you please do `rustup update` and then `cargo +nightly fmt` and check in the result?

I tried to do so but I do not have permission:

```
(arrow_dev) alamb@MacBook-Pro-2:~/Software/sqlparser-rs$ git commit -a -m 'fix: nightly fmt'
[copy f25492f] fix: nightly fmt
 1 file changed, 4 insertions(+), 4 deletions(-)
(arrow_dev) alamb@MacBook-Pro-2:~/Software/sqlparser-rs$ git push
ERROR: Permission to risinglightdb/sqlparser-rs.git denied to alamb.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

alamb:
BTW I hope to get this PR merged in over the next day or two prior to releasing version 0.16

wangrunji0408:
I merged the main branch and it seems to fix the fmt problem.

alamb:
Thanks again @wangrunji0408

PR #450, opened 2022-03-30
Support `UPDATE SET FROM` statement

### LINKED ISSUE #449: Support `UPDATE SET FROM` statement
In some database like Postgres, it is allow to run `Update` statement like the following one:
``` SQL
UPDATE t1
SET name = t2.name FROM ( 
    SELECT name, id
    FROM t1
    GROUP BY id
) t2
WHERE t1.id = t2.id
```

slhmy:
Close #449

This PR adjust the `Update` statement with an additional `from` member, which can hold the contents behind `FROM`.
Since I can only approve this statement is supported in Postgres, currently I only allow `PostgreSqlDialect` to parse `FROM` part.

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -402,6 +402,84 @@ PHP	₱ USD $
 }
 
 #[test]
+fn parse_update_set_from() {
+    let sql = "UPDATE t1 SET name = t2.name FROM (SELECT name, id FROM t1 GROUP BY id) AS t2 WHERE t1.id = t2.id";
+    let stmt = pg().verified_stmt(sql);
+    assert_eq!(
+        stmt,
+        Statement::Update {
+            table: TableWithJoins {
+                relation: TableFactor::Table {
+                    name: ObjectName(vec![Ident::new("t1")]),
+                    alias: None,
+                    args: vec![],
+                    with_hints: vec![],
+                },
+                joins: vec![],
+            },
+            assignments: vec![Assignment {
+                id: vec![Ident::new("name")],
+                value: Expr::CompoundIdentifier(vec![Ident::new("t2"), Ident::new("name")])
+            }],
+            from: Some(TableWithJoins {
+                relation: TableFactor::Derived {
+                    lateral: false,
+                    subquery: Box::new(Query {
+                        with: None,
+                        body: SetExpr::Select(Box::new(Select {
+                            distinct: false,
+                            top: None,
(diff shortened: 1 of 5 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -401,6 +401,21 @@ PHP	₱ USD $
...
+    SET name = t2.name FROM ( \
+        SELECT name, id \
+        FROM t1 \
+        GROUP BY id \
+    ) t2 \
+    WHERE t1.id = t2.id";
+    let ast = pg().one_statement_parses_to(sql, "");
+    println!("{:#?}", ast);
+    println!("{}", ast);
+    //assert_eq!(sql, ast.to_string());
```
alamb, line 416: It seems like this test is missing the expected ast. Can you please update that?

alamb (commented):
Thanks @slhmy  -- the code looks great -- I think a little work on the tests and this PR will be good to merge.

tests/sqlparser_postgres.rs
slhmy: Added, not sure whether I'm doing right 😄.

tests/sqlparser_postgres.rs
```diff
@@ -401,6 +401,109 @@ PHP	₱ USD $
...
+#[test]
+fn parse_update_set_from() {
+    let sql = "UPDATE t1
+    SET name = t2.name FROM ( \
+        SELECT name, id \
+        FROM t1 \
+        GROUP BY id \
+    ) t2 \
+    WHERE t1.id = t2.id";
+    let ast = pg().one_statement_parses_to(sql, "");
```
alamb, line 413: Would it be possible to follow the model of the test `test_copy_from` (which is immediately below this one)?

So something like

```rust
    let stmt = pg().verified_stmt(sql);
    assert_eq!(
        stmt,
        Statement::Update {
...
    });
```

slhmy: Good suggestion, I didn't notice that before.

Changed.

alamb:
Thanks @slhmy !

alamb:
Looks like there is now a small compiler error 🤔

slhmy:
> Looks like there is now a small compiler error 🤔

Since I hadn't fetch up with recent commits, we met these compile error.
Fixed.

alamb:
Thanks for sticking with this @slhmy  -- I hope to release the next version in the next few days which will include this change

PR #456, opened 2022-04-11
Support  `global`,`local`, `on commit` for `create temporary table`

gandronchik:
CREATE [GLOBAL | LOCAL] TEMPORARY TABLE table (col INT) ON COMMIT [DELETE ROWS | DROP | PRESERVE ROWS]

add support GLOBAL | LOCAL
add support ON COMMIT [DELETE ROWS | DROP | PRESERVE ROWS]

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1285,7 +1285,7 @@ impl fmt::Display for Statement {
                     external = if *external { "EXTERNAL " } else { "" },
                     global = global
                         .map(|global| {
-                            if global == true {
+                            if global {
                                 "GLOBAL "
                             } else {
                                 "LOCAL "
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -1276,9 +1280,18 @@ impl fmt::Display for Statement {
...
                     external = if *external { "EXTERNAL " } else { "" },
+                    global = if global.is_some() {
+                        if global.unwrap() == true {
+                            "GLOBAL "
+                        } else {
+                            "LOCAL "
+                        }
+                    } else {
+                        ""
+                    },
```
alamb, line 1294: I think more idomatic rust would be the following (it avoids `unwrap` when you know it is already `Some`)

```suggestion
                    global = global.map(|global| {
                        if global.unwrap() == true {
                            "GLOBAL "
                        } else {
                            "LOCAL "
                        }
                    }).unwrap_or(""),
```

alamb (approved):
I had one tiny style suggestion, but otherwise this looks great 👍

alamb (approved):
Thanks @gandronchik

alamb:
Seems like there is a clippy failure @gandronchik

gandronchik:
> 



> Seems like there is a clippy failure @gandronchik

Oh, god, sorry) fixed

alamb:
Thanks again @gandronchik  👍

PR #458, opened 2022-04-13
Add support for postgres json operators `->`, `->>`, `#>`, and `#>>`

poonai:
`arrow` are used to represent nested field of json.

```
SELECT info->'items'->>'product' FROM orders
```

This PR adds support to parse identifier which are separated using `arrow` 
 
### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -197,11 +197,11 @@ pub enum JsonOperator {
 impl fmt::Display for JsonOperator {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            JsonOperator::LongArrow => {
-                write!(f, "->>")
-            }
             JsonOperator::Arrow => {
                 write!(f, "->")
+            }
+            JsonOperator::LongArrow => {
+                write!(f, "->>")
             }
             JsonOperator::HashArrow => {
                 write!(f, "#>")
@@ -225,7 +225,7 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
-    /// JsonIdentifier eg: data->'tags'
+    /// JSON access (postgres)  eg: data->'tags'
     JsonAccess {
         left: Box<Expr>,
         operator: JsonOperator,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -180,6 +180,29 @@ impl fmt::Display for Array {
     }
 }
 
+/// JsonOperator
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum JsonOperator {
+    /// -> keeps the value as json
+    Arrow,
+    /// ->> keeps the value as text or int.
+    LongArrow,
+}
+
+impl fmt::Display for JsonOperator {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            JsonOperator::LongArrow => {
+                write!(f, "->>")
+            }
+            JsonOperator::Arrow => {
+                write!(f, "->")
+            }
+        }
+    }
+}
+
 /// An SQL expression of any type.
 ///
 /// The parser does not distinguish between expressions of different types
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -192,6 +215,12 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// JsonIdentifier eg: data->'tags'
+    JsonIdentifier {
```
alamb, line 219: Since this is more like a map access, what do you think about calling this `JsonAccess` to be consistent with `MapAccess` (for `foo[4]` type expressions)?

src/ast/mod.rs
```diff
@@ -192,6 +215,12 @@ pub enum Expr {
...
+        ident: Ident,
```
alamb, line 220: I is there a reason `ident` isn't also a `Box<Expr>`? The examples from the postgres docs seem to show expressions being used as the left hand side (e.g. `'foo'::json`)

src/tokenizer.rs
```diff
@@ -141,6 +141,10 @@ pub enum Token {
     PGCubeRoot,
     /// `?` or `$` , a prepared statement arg placeholder
     Placeholder(String),
+    /// ->, used as a operator to extract json field in PostgreSQL
+    Arrow,
+    /// -->, used as a operator to extract json field in PostgreSQL
```
alamb, line 146:
```suggestion
    /// -->, used as a operator to extract json field as text in PostgreSQL
```

alamb (commented):
According to 

https://www.postgresql.org/docs/current/functions-json.html

It appears there are also `#>` and `#>>` operators. I wonder if you have thought about supporting them?

src/tokenizer.rs
poonai: done

src/ast/mod.rs
poonai: I missed it. I've made necessary changes and test is added for that.

src/ast/mod.rs
poonai: done

poonai:
> It appears there are also #> and #>> operators. I wonder if you have thought about supporting them?

done 😃

src/ast/mod.rs
```diff
@@ -180,6 +180,39 @@ impl fmt::Display for Array {
...
+
+impl fmt::Display for JsonOperator {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            JsonOperator::LongArrow => {
+                write!(f, "->>")
+            }
+            JsonOperator::Arrow => {
+                write!(f, "->")
+            }
```
alamb, line 205: I don't think it matters, but it would be nice to keep the order of the variants consistent

```suggestion
            JsonOperator::Arrow => {
                write!(f, "->")
            }
            JsonOperator::LongArrow => {
                write!(f, "->>")
            }
```

src/ast/mod.rs
```diff
@@ -192,6 +225,12 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// JsonIdentifier eg: data->'tags'
```
alamb, line 228:
```suggestion
    /// JSON access (postgres)  eg: data->'tags'
```

src/tokenizer.rs
```diff
@@ -141,6 +141,14 @@ pub enum Token {
     PGCubeRoot,
     /// `?` or `$` , a prepared statement arg placeholder
     Placeholder(String),
+    /// ->, used as a operator to extract json field in PostgreSQL
+    Arrow,
+    /// -->, used as a operator to extract json field as text in PostgreSQL
```
alamb, line 146:
```suggestion
    /// ->>, used as a operator to extract json field as text in PostgreSQL
```

src/tokenizer.rs
```diff
@@ -197,6 +205,10 @@ impl fmt::Display for Token {
             Token::PGSquareRoot => f.write_str("|/"),
             Token::PGCubeRoot => f.write_str("||/"),
             Token::Placeholder(ref s) => write!(f, "{}", s),
+            Token::Arrow => write!(f, "->"),
+            Token::LongArrow => write!(f, "-->"),
```
alamb, line 209:
```suggestion
            Token::LongArrow => write!(f, "->>"),
```

alamb: I wonder why the tests didn't cover this? I think there is a function that verifies that a parsed sql statement is then displayed in the same way

alamb (commented):
Thanks @poonai  -- I think this is quite close (the only thing I think that is required is the `Display` for `Token::LongArrow`)

src/ast/mod.rs
poonai: done

src/ast/mod.rs
poonai: done

src/tokenizer.rs
poonai: done

src/tokenizer.rs
poonai: We are printing `JsonOperator` not the `Token`. so it escaped :sweat_smile:

alamb (approved):
Thanks @poonai !

PR #462, opened 2022-04-17
Add support of NVARCHAR data type

yuval-illumex:
Redshift support the NVARCHAR datatype ([Redshift Data Types](https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html))

The following query can't be parsed using sqlparser:

```
SELECT cast(field as nvarchar(50)) as "field" 
FROM a

```

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -101,7 +101,7 @@ impl fmt::Display for DataType {
                 format_type_with_optional_length(f, "CHARACTER VARYING", size, false)
             }
             DataType::Nvarchar(size) => {
-                format_type_with_optional_length(f, "NUMERIC CHARACTER VARYING", size, false)
+                format_type_with_optional_length(f, "NVARCHAR", size, false)
             }
             DataType::Uuid => write!(f, "UUID"),
             DataType::Clob(size) => write!(f, "CLOB({})", size),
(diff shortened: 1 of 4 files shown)


src/ast/data_type.rs
```diff
@@ -98,6 +100,9 @@ impl fmt::Display for DataType {
             DataType::Varchar(size) => {
                 format_type_with_optional_length(f, "CHARACTER VARYING", size, false)
             }
+            DataType::Nvarchar(size) => {
+                format_type_with_optional_length(f, "Numeric CHARACTER VARYING", size, false)
```
alamb, line 104:
```suggestion
                format_type_with_optional_length(f, "NUMERIC CHARACTER VARYING", size, false)
```

I think this should be all capitals to be consistent with the rest of this function

alamb (commented):
Thanks @yuval-illumex  -- I have one small suggestion and I think this PR needs a test. Otherwise it should be good to go

yuval-illumex:
Thanks for the feedback @alamb

alamb:
Thanks @yuval-illumex  -- can you please add a test (so that someone doesn't accidentally break this feature in the future and not know it?)

alamb (approved):
Thanks -- looks good @yuval-illumex

PR #463, opened 2022-04-17
Add Support in the "Position" Function

yuval-illumex:
Add support to the function position as described in the [docs](https://www.postgresql.org/docs/9.1/functions-string.html)

Example:

```
SELECT POSITION('@' in field)
FROM table
```


src/ast/mod.rs
```diff
@@ -252,6 +252,11 @@ pub enum Expr {
         field: DateTimeField,
         expr: Box<Expr>,
     },
+    /// POSITION(<expr> in <expr>)
+    Position {
+        expr: Box<Expr>,
+        from: Box<Expr>,
```
alamb, line 258: I wonder if calling this field `in` (rather than `from`) would be better as it is closer to the original SQL text (which uses `in`)?

alamb (commented):
Thanks @yuval-illumex  -- Does `POSITION('@' in 'foo')` parse in sqlparser without changes today? Or is this PR required to parse the expression?

src/ast/mod.rs
yuval-illumex: @alamb I called it from because I found it very similar to the `substring`  function. I think `in` is a great option

src/parser.rs
```diff
@@ -1201,23 +1210,32 @@ impl<'a> Parser<'a> {
...
         }
-        self.expect_token(&Token::LParen)?;
-        let in_op = if self.parse_keyword(Keyword::SELECT) || self.parse_keyword(Keyword::WITH) {
-            self.prev_token();
-            Expr::InSubquery {
-                expr: Box::new(expr),
-                subquery: Box::new(self.parse_query()?),
-                negated,
-            }
+        if self.peek_token() == Token::LParen {
```
yuval-illumex, line 1213: @alamb This section is a must in order to the `POSITION` function to parse

src/parser.rs
```diff
@@ -423,6 +423,7 @@ impl<'a> Parser<'a> {
                 Keyword::TRY_CAST => self.parse_try_cast_expr(),
                 Keyword::EXISTS => self.parse_exists_expr(),
                 Keyword::EXTRACT => self.parse_extract_expr(),
+                Keyword::POSITION => self.parse_position_expr(),
```
yuval-illumex, line 426: @alamb This section is not a must. If I will remove it, the parsed object will look like:

```
"projection": [
            {
              "UnnamedExpr": {
                "Function": {
                  "name": [
                    {
                      "value": "POSITION",
                      "quote_style": null
                    }
                  ],
                  "args": [
                    {
                      "Unnamed": {
                        "Expr": {
                          "Position": {
                            "expr": {
                              "Value": {
                                "SingleQuotedString": "@"
                              }
                            },
                            "from": {
                              "Identifier": {
                                "value": "field",
                                "quote_style": null
                              }
                            }
                          }
                        }
                      }
                    }
                  ],
                  "over": null,
                  "distinct": false
                }
              }
            }
```

I'm not if position deserves special method in the parser.

alamb: > I'm not if position deserves special method in the parser.

I guess if there is some reason to special case the parser for `POSITION` then it is a good idea. If the parser can handle `POSITION` without any changes, I would be inclined not to include the extra code (and any users of `sqlparser-rs` who want to support `POSITION` can look for the patterns they want to support)

yuval-illumex:
@alamb Updated the PR, changed the POSITION function to be like any other function (But the parsed response must use a struct for POSITION)

src/parser.rs
```diff
@@ -1201,23 +1201,32 @@ impl<'a> Parser<'a> {
...
         }
-        self.expect_token(&Token::LParen)?;
-        let in_op = if self.parse_keyword(Keyword::SELECT) || self.parse_keyword(Keyword::WITH) {
-            self.prev_token();
-            Expr::InSubquery {
-                expr: Box::new(expr),
-                subquery: Box::new(self.parse_query()?),
-                negated,
-            }
+        if self.peek_token() == Token::LParen {
```
alamb, line 1204: As written, this code will now parse `<expr> IN <expr>` *anywhere* in the SQL query, not just in the position function which I don't think is correct

I wrote a test for the expected behavior in https://github.com/sqlparser-rs/sqlparser-rs/pull/468 (and that test does not pass with this branch)

Given that I finally see that `POSITION` does probably need to be special cased as only within `POSITION` would `<expr> IN <expr>` parse successfully.

src/parser.rs
```diff
@@ -1201,23 +1201,32 @@ impl<'a> Parser<'a> {
...
+                    list: self.parse_comma_separated(Parser::parse_expr)?,
+                    negated,
+                }
+            };
+            self.expect_token(&Token::RParen)?;
+            Ok(in_op)
         } else {
-            Expr::InList {
+            let from = self.parse_expr()?;
+            Ok(Expr::Position {
```
alamb, line 1225: Won't this make `Expr::Position` for *any* function, even like `foo(x in y)`?

alamb (commented):
Sorry for the delayed review @yuval-illumex  -- I have been quite busy recently.

I didn't appreciate that `<expr> IN <expr>` was only valid within the context of a `POSITION()` function. Sorry for my confusion

Can you please change this code so that the `POSITION` function is handled specially (I realize you did do that initially and I implied you should change)?

Thanks and sorry again.

yuval-illumex:
@alamb Thanks for the detailed feedback! Missed the bug with IN.
Appreciate new feedback after the fix

alamb (approved):
Looks good to me -- thank you @yuval-illumex  ❤️

alamb:
I wrote up a small negative case here as well: https://github.com/sqlparser-rs/sqlparser-rs/pull/469

PR #466, opened 2022-04-20
add support for postgres composite types

poonai:
This PR adds support for postgres composite values. 
ref: https://www.postgresql.org/docs/current/rowtypes.html#:~:text=A%20composite%20type%20represents%20the,be%20of%20a%20composite%20type

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -516,7 +516,18 @@ impl<'a> Parser<'a> {
                         }
                     };
                 self.expect_token(&Token::RParen)?;
-                Ok(expr)
+                if !self.consume_token(&Token::Period) {
+                    return Ok(expr);
+                }
+                let tok = self.next_token();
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -516,7 +516,18 @@ impl<'a> Parser<'a> {
                         }
                     };
                 self.expect_token(&Token::RParen)?;
-                Ok(expr)
+                if !self.consume_token(&Token::Period) {
```
alamb, line 519: I am thinking if the `CompositeAccess` can only appear after a `)` -- it seems like it should be possible in other places, though i suppose without a parenthesis `.<name>` will appear as a compound identifier.

alamb (approved):
Thanks @poonai  -- I am really sorry for the delay in review.

PR #469, opened 2022-04-25
Add negative tests for POSITION

alamb:
While reviewing https://github.com/sqlparser-rs/sqlparser-rs/pull/463 I noticed the error condition was not covered. Added test coverage here. cc @yuval-illumex

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -4601,3 +4601,20 @@ fn parse_position() {
         expr_from_projection(only(&select.projection))
     );
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


yuval-illumex (approved):

yuval-illumex:
@alamb Thanks! I wanted your feedback on your code before adding tests, so thank you for this PR :)

alamb:
> @alamb Thanks! I wanted your feedback on your code before adding tests, so thank you for this PR :)


No problem -- I didn't want to drag out the PR any longer as we had already had several back and forths. If you have more tests in mind, another PR would be great!

PR #471, opened 2022-04-26
Add Redshift dialect, handle square brackets properly

mskrzypkows:
We need to detect `[` or `"` for Redshift quotes around indentifier and at the same time exclude
treating JSON paths as indentifer

### DIFF
--- a/examples/cli.rs
+++ b/examples/cli.rs
@@ -43,6 +43,7 @@ $ cargo run --feature json_example --example cli FILENAME.sql [--dialectname]
         "--mysql" => Box::new(MySqlDialect {}),
         "--snowflake" => Box::new(SnowflakeDialect {}),
         "--hive" => Box::new(HiveDialect {}),
+        "--redshift" => Box::new(RedshiftSqlDialect {}),
         "--generic" | "" => Box::new(GenericDialect {}),
         s => panic!("Unexpected parameter: {}", s),
     };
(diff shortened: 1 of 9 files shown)


mskrzypkows:
Hi @alamb, could you give me some feedback? May the PR be reviewed?

alamb (commented):
Thank you @mskrzypkows for the contribution. 

I am sorry for the delay.

I am not super familiar with the Redshift SQL dialect, nor have I used it and I don't have access to such a system. I didn't find any mention of using `[` and `]` as quote characters in the redshift docs. Here is where I looked: https://docs.aws.amazon.com/redshift/latest/dg/r_names.html

Can you maybe point me at some examples or reference documentation with this syntax?

I think in general this PR is looking good to merge because:
1. It is a new Dialect (and thus existing dialects are unaffected)
2. It has good test coverate

I left some small style / documentation suggestions

mskrzypkows:
> Can you maybe point me at some examples or reference documentation with this syntax?

You're right, it's hard to find some documentation mentioning support for `[` and `]`. I've tried to find something, but failed. I know it works from tests on Redshift DB. Furthermore, I don't know why it's not present in the documentation, while DB support it. I filled the "Documentation feedback form" so hopefully Amazon will mention support for square brackets soon.
I suppose that Redshift mimics this behaviour from MSSQL: https://docs.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver15

alamb:
I hope to create a new sqlparser release later later this week and publish
to crates.io

PR #473, opened 2022-04-26
insert, merge: make INTO keyword optional

mobuchowski:
In multiple databases, INTO keyword after INSERT and MERGE is optional.

BigQuery: https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#insert_statement
MySQL: https://dev.mysql.com/doc/refman/8.0/en/insert.html
MSSQL: https://docs.microsoft.com/en-us/sql/t-sql/statements/insert-transact-sql?view=sql-server-ver15

This PR makes it so. Also, it records it's existence in a struct - not sure this makes sense, but it allows distinguishing statements that have the keyword from the ones that don't have it. 

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -41,6 +41,9 @@ fn parse_insert_values() {
     let rows1 = vec![row.clone()];
     let rows2 = vec![row.clone(), row];
 
+    let sql = "INSERT customer VALUES (1, 2, 3)";
+    check_one(sql, "customer", &[], &rows1);
+
     let sql = "INSERT INTO customer VALUES (1, 2, 3)";
     check_one(sql, "customer", &[], &rows1);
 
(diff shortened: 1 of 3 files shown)


mobuchowski:
Sorry to bother you @alamb, but hopefully you'll have a bit of time to look at this PR?

tests/sqlparser_common.rs
```diff
@@ -4333,12 +4326,14 @@ fn parse_merge() {
     let sql = "MERGE INTO s.bar AS dest USING (SELECT * FROM s.foo) as stg ON dest.D = stg.D AND dest.E = stg.E WHEN NOT MATCHED THEN INSERT (A, B, C) VALUES (stg.A, stg.B, stg.C) WHEN MATCHED AND dest.A = 'a' THEN UPDATE SET dest.F = stg.F, dest.G = stg.G WHEN MATCHED THEN DELETE";
     match verified_stmt(sql) {
```
alamb, line 4327: Can you please also add a test for `MERGE` without the `INTO`?

alamb:
Thanks @mobuchowski  -- this looks good to me. I think it needs a test to cover `MERGE` without an `INTO` but otherwise it is good to go!

tests/sqlparser_common.rs
alamb: Thanks -- I don't seem to see a new commit on this PR yet. Did you push one?

mobuchowski: Force pushed previous one 🙂

alamb (approved):
Thanks @mobuchowski

PR #474, opened 2022-04-27
Add support in IS boolean filter

yuval-illumex:
The following query is valid and should be supported:
`SELECT f FROM foo WHERE field = true`

Verified on latest postgres version

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -4618,3 +4618,18 @@ fn parse_position_negative() {
         res.unwrap_err()
     );
 }
+
+#[test]
+fn parse_is_boolean() {
+    one_statement_parses_to(
+        "SELECT f from foo where field is true",
+        "SELECT f FROM foo WHERE field = true",
+    );
+
+    let sql = "SELECT f from foo where field is 0";
+    let res = parse_sql_statements(sql);
+    assert_eq!(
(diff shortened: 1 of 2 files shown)


tests/sqlparser_common.rs
```diff
@@ -4618,3 +4618,20 @@ fn parse_position_negative() {
...
+
+#[test]
+fn parse_is_boolean() {
+    one_statement_parses_to(
+        "SELECT f from foo where field is true",
+        "SELECT f FROM foo WHERE field = true",
+    );
+
+    let sql = "SELECT f from foo where field is 0";
+    let res = parse_sql_statements(sql);
```
alamb, line 4630: nice

tests/sqlparser_common.rs
```diff
@@ -4618,3 +4618,20 @@ fn parse_position_negative() {
         res.unwrap_err()
     );
 }
+
+#[test]
+fn parse_is_boolean() {
+    one_statement_parses_to(
```
alamb, line 4624: can you also please add a test for the `false` form -- like `SELECT f from foo where field is false`?

alamb (commented):
👍  looking good

alamb (approved):
Looks good -- thank you @yuval-illumex

PR #475, opened 2022-04-27
Fix redshift specific character

yuval-illumex:
The following query is valid in Redshift:
```
SELECT #_of_values
FROM foo
```

In postgres, when running using pgadmin, the query is not valid and must written as the following:
(I'm not sure what is the behavior in other drivers/apps that can query postgres).
```
SELECT foo."#_of_values"
FROM foo
```

As redshift is based on postgres, for me it makes sense to change the postgres dialect.
I would love for some feedback.

Thanks!

### DIFF
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -20,7 +20,7 @@ impl Dialect for PostgreSqlDialect {
         // See https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
         // We don't yet support identifiers beginning with "letters with
         // diacritical marks and non-Latin letters"
-        ('a'..='z').contains(&ch) || ('A'..='Z').contains(&ch) || ch == '_'
+        ('a'..='z').contains(&ch) || ('A'..='Z').contains(&ch) || ch == '_' || ch == '#'
     }
 
     fn is_identifier_part(&self, ch: char) -> bool {
@@ -29,5 +29,6 @@ impl Dialect for PostgreSqlDialect {
             || ('0'..='9').contains(&ch)
             || ch == '$'
             || ch == '_'
+            || ch == '#'
     }
 }
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -422,7 +422,24 @@ impl<'a> Tokenizer<'a> {
                         s += s2.as_str();
                         return Ok(Some(Token::Number(s, false)));
                     }
-                    Ok(Some(Token::make_word(&s, None)))
+                    if s == "#" {
+                        match chars.peek() {
+                            Some('>') => {
+                                chars.next();
+                                match chars.peek() {
+                                    Some('>') => {
+                                        chars.next();
+                                        return Ok(Some(Token::HashLongArrow))
+                                    }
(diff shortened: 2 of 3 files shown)


yuval-illumex:
@alamb appreciate your feedback on this one, as it's more complicated then other PRs

src/dialect/postgresql.rs
```diff
@@ -20,7 +20,7 @@ impl Dialect for PostgreSqlDialect {
         // See https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
         // We don't yet support identifiers beginning with "letters with
         // diacritical marks and non-Latin letters"
-        ('a'..='z').contains(&ch) || ('A'..='Z').contains(&ch) || ch == '_'
+        ('a'..='z').contains(&ch) || ('A'..='Z').contains(&ch) || ch == '_' || ch == '#'
```
alamb, line 23: Yeah I am torn on this one - I worry about causing trouble for someone who is relaying on `#` not being a valid pg identifiers. I don't really have a great understanding of what people are doing / using sqlparser for and I would say it is not particularly consistent about disallowing constructs that are not valid other dialects of SQL 🤔 

What do you think about adding a `RedshiftSqlDialect` to avoid unintended consquences? I think it could be pretty straightforward, something like:

```rust
    fn is_identifier_start(&self, ch: char) -> bool {
      PostgreSqlDialect::is_identifier_start(ch) || ch == '#'
    }
}
```

src/tokenizer.rs
```diff
@@ -422,7 +441,11 @@ impl<'a> Tokenizer<'a> {
                         s += s2.as_str();
                         return Ok(Some(Token::Number(s, false)));
                     }
-                    Ok(Some(Token::make_word(&s, None)))
+                    if s == "#" {
```
yuval-illumex, line 444: @alamb What do you think about this change?
I'm worried about clash with lines 610:
I didn't found a better way to solve it.

`                '#' if dialect_of!(self is SnowflakeDialect) => {
                    chars.next(); // consume the '#', starting a snowflake single-line comment
                    let comment = self.tokenize_single_line_comment(chars);
                    Ok(Some(Token::Whitespace(Whitespace::SingleLineComment {
                        prefix: "#".to_owned(),
                        comment,
                    })))
                }`

yuval-illumex: Maybe we should check if the dialect is RedshiftSqlDialect?

alamb:
I just noticed an outstanding PR from @mskrzypkows that adds a `RedshiftSqlDialect` as well -- what do you think https://github.com/sqlparser-rs/sqlparser-rs/pull/471

src/tokenizer.rs
alamb: The change doesn't *feel* right to me (because it effectively skips `#`s so wouldn't it mean that things like `#SELECT` will parse as `SELECT` (not just identifiers)?

yuval-illumex:
> I just noticed an outstanding PR from @mskrzypkows that adds a `RedshiftSqlDialect` as well -- what do you think #471

I totally agree. I think that I will wait for #471 be merged, and then I will try to make this PR to work only on Redshift.

PR #477, opened 2022-04-28
feat: Support ANY/ALL operators

ovr:
Hello!

Ref issue: https://github.com/sqlparser-rs/sqlparser-rs/issues/476

I want to support `a = ANY(expr)` in DF, but after reviewing I come to decision that it requires changes in `sql-parser` to simplify code (because hacking when the right side is a function is a super tricky)

Thanks

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -270,6 +270,10 @@ pub enum Expr {
         op: BinaryOperator,
         right: Box<Expr>,
     },
+    /// Any operation e.g. `1 ANY (1)` or `foo > ANY(bar)`, It will be wrapped in the right side of BinaryExpr
+    AnyOp(Box<Expr>),
+    /// ALL operation e.g. `1 ALL (1)` or `foo > ALL(bar)`, It will be wrapped in the right side of BinaryExpr
+    AllOp(Box<Expr>),
     /// Unary operation e.g. `NOT foo`
     UnaryOp {
         op: UnaryOperator,
@@ -433,6 +437,8 @@ impl fmt::Display for Expr {
                 high
             ),
             Expr::BinaryOp { left, op, right } => write!(f, "{} {} {}", left, op, right),
+            Expr::AnyOp(expr) => write!(f, "ANY({})", expr),
+            Expr::AllOp(expr) => write!(f, "ALL({})", expr),
             Expr::UnaryOp { op, expr } => {
                 if op == &UnaryOperator::PGPostfixFactorial {
                     write!(f, "{}{}", expr, op)
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -1000,7 +1000,7 @@ fn parse_bitwise_ops() {
 
 #[test]
 fn parse_binary_any() {
-    let select = verified_only_select(&"SELECT a = ANY(b)");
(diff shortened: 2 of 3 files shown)


ovr:
@alamb WDYT? Should I create a separate issue for supporting AllOperatorSubquery & AnyOperatorSubQuery?

alamb (commented):
Ho @ovr  --- sounds like an exiting feature!

Looking at the postgres documentation,  https://www.postgresql.org/docs/current/functions-subquery.html it seems to me like `ANY` / `ALL` can be combined with any operator that results in a `boolean`, not just `<, <=, =, !=, >=, >`

For example, with `~`:

```sql
alamb=# select x from test where x ~ ANY (select x from test);
  x  
-----
 foo
(1 row)
```

What would you think about adding something like

```rust
 /// `ANY()`
    Any(Box<Expr>),
```

This might be more in keeping with the spirit of sql-parser-rs to parse the query faithfully without applying any implicit semantics to it. 

> @alamb WDYT? Should I create a separate issue for supporting AllOperatorSubquery & AnyOperatorSubQuery?

I don't think separate issues are necessary

alamb:
> @alamb Oh, I see, but anyway what about LIKE/AND/OR operators which are supported in Binary expression but is not supported for ANY & ALL operators? Another question is, whether it will lead to a nonimplicit pattern machine on the DF side?

I am not quite sure what you mean about "`nonimplicit` pattern match on the DF side"


Maybe you could do something like the following with a match guard (totally untested):

```rust
sqlparser::Expr::BinaryExpr {
  left, 
  op
  right
} if is_any_expr(&right) => {...}
  }
...
```


and 
```rust
/// returns true of `expr` represents an `Any` clause
fn is_any_expr(expr: sqlparser::Expr) ->bool)  {
...
}
```

I found this style worked pretty well in simplify_expressions: 

https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/optimizer/simplify_expressions.rs#L498-L508

ovr:
@alamb Oh, I see your idea about simplifying it and wrapping it on the right side of the BinaryExpression expression. I've updated the PR as you suggested. Thanks

src/ast/mod.rs
```diff
@@ -270,6 +270,10 @@ pub enum Expr {
         op: BinaryOperator,
         right: Box<Expr>,
     },
+    /// Any operation e.g. `1 ANY (1)` or `foo > ANY(bar)`, It will be wrapped in the right side of BinaryExpr
+    AnyOp(Box<Expr>),
```
alamb, line 274: 👍

tests/sqlparser_common.rs
```diff
@@ -998,6 +998,32 @@ fn parse_bitwise_ops() {
...
 }
 
+#[test]
+fn parse_binary_any() {
+    let select = verified_only_select("SELECT a = ANY(b)");
+    assert_eq!(
+        SelectItem::UnnamedExpr(Expr::BinaryOp {
+            left: Box::new(Expr::Identifier(Ident::new("a"))),
+            op: BinaryOperator::Eq,
+            right: Box::new(Expr::AnyOp(Box::new(Expr::Identifier(Ident::new("b"))))),
```
alamb, line 1008: 👍  looks good to me

alamb (approved):
Looks great to me @ovr  👍

PR #479, opened 2022-04-28
feat: Support KILL [CONNECTION | QUERY] processlist_id

ovr:


src/parser.rs
```diff
@@ -2866,6 +2867,32 @@ impl<'a> Parser<'a> {
         })
     }
 
+    // KILL [CONNECTION | QUERY] processlist_id
```
alamb, line 2870:
```suggestion
    // KILL [CONNECTION | QUERY | MUTATION] processlist_id
```

alamb (approved):
The syntax for clickhouse's kill is more sophisticated than `kill query id` I think: https://clickhouse.com/docs/en/sql-reference/statements/kill#kill-query-statement

But this looks like a good start to me. Thanks @ovr

PR #480, opened 2022-04-29
feat: Allow ArrayIndex for GenericDialect

ovr:
Hello!

From discussion: https://github.com/apache/arrow-datafusion/pull/2195#issuecomment-1096599374
Looks like it's a good decision to allow ArrayIndex for Generic dialect, but anyway it will be cool to improve dialect with feature flags.

Thanks

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -1137,7 +1137,7 @@ fn parse_array_index_expr() {
         .collect();
 
     let sql = "SELECT foo[0] FROM foos";
-    let select = pg().verified_only_select(sql);
+    let select = pg_and_generic().verified_only_select(sql);
     assert_eq!(
         &Expr::ArrayIndex {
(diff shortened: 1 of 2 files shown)


alamb:
👍  looks good to me -- would it be possible to add a test (so someone doesn't accidentally break this)? 

Also I hope to make a release if sqlparser sometime early next week

tests/sqlparser_postgres.rs
```diff
@@ -1137,7 +1137,7 @@ fn parse_array_index_expr() {
         .collect();
 
     let sql = "SELECT foo[0] FROM foos";
-    let select = pg().verified_only_select(sql);
+    let select = pg_and_generic().verified_only_select(sql);
```
ovr, line 1140: >Just waiting on a test and this one is good to go

@alamb Ops, I missed your message. I added generic dialect to the `parse_array_index_expr` test to prove that it works with generic dialect too.

alamb (approved):
Looks good -- thanks @ovr

PR #484, opened 2022-05-04
set: allow dots in variables by moving to ObjectName

mobuchowski:
Motivation for this change are Hive variables like 

```
SET hive.exec.dynamic.partition = true;
```


alamb (approved):
LGTM -- thank you @mobuchowski

PR #487, opened 2022-05-09
feat: Support `TABLE` keyword with `SELECT INTO`

MazterQyou:
This PR adds support for `TABLE` keyword with `SELECT INTO` queries which is valid as per [PostgreSQL docs](https://www.postgresql.org/docs/current/sql-selectinto.html).
This allows processing queries like `SELECT * INTO TEMPORARY TABLE "tmp" FROM (SELECT 1 AS COL) AS CHECKTEMP` which otherwise would fail as `TABLE` would be parsed as a table name.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -654,6 +654,7 @@ impl fmt::Display for Values {
 pub struct SelectInto {
     pub temporary: bool,
     pub unlogged: bool,
+    pub table: bool,
     pub name: ObjectName,
 }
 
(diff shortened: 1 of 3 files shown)


src/ast/query.rs
```diff
@@ -654,14 +654,16 @@ impl fmt::Display for Values {
 pub struct SelectInto {
     pub temporary: bool,
     pub unlogged: bool,
+    pub table: bool,
```
alamb, line 657: I like that this follows the rest of the crate's design to mirror the input syntax in the crated AST

👍

alamb (approved):
Looks good -- thank you @MazterQyou

PR #489, opened 2022-05-09
Add docs for `MapAccess`

alamb:
I found an undocumented variant when reviewing another PR.

This appears to have been introduced by @hntd187  in #235

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -324,6 +324,8 @@ pub enum Expr {
     /// as well as constants of other types (a non-standard PostgreSQL extension).
     TypedString { data_type: DataType, value: String },
     /// Access a map-like object by field (e.g. `column['field']` or `column[4]`
+    /// Note that depending on the dialect, struct like accesses may be
+    /// parsed as [`ArrayIndex`] or [`MapAccess`] 
     /// <https://clickhouse.com/docs/en/sql-reference/data-types/map/>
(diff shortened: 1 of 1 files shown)


src/ast/mod.rs
```diff
@@ -232,10 +232,7 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// CompositeAccess (postgres) eg: SELECT (information_schema._pg_expandarray(array['i','i'])).n
-    CompositeAccess {
-        expr: Box<Expr>,
-        key: Ident,
-    },
+    CompositeAccess { expr: Box<Expr>, key: Ident },
```
alamb, line 235: I don't understand why `cargo fmt` made these formatting changes in this PR

src/ast/mod.rs
```diff
@@ -331,14 +322,10 @@ pub enum Expr {
...
-        data_type: DataType,
-        value: String,
-    },
-    MapAccess {
-        column: Box<Expr>,
-        keys: Vec<Expr>,
-    },
+    TypedString { data_type: DataType, value: String },
+    /// Access a map-like object by field (e.g. `column['field']` or `column[4]`
+    /// <https://clickhouse.com/docs/en/sql-reference/data-types/map/>
```
ovr, line 327: Maybe it's a good idea to mention that ArrayIndex replaces MapAccess for some dialects.

PR #491, opened 2022-05-10
Support hive `array<>` DataType  syntax

mobuchowski:
This PR adds support for Hive's array<data_type> syntax, used in DDL - for example, `CREATE TABLE`.
 
Note, that this implementation has deficiency that would need to be adressed if `sqlparser-rs` were to support nested Hive complex types - like array of structs. Namely, it has "[C++ problem](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1757.html)" - two closing angle brackets are parsed as shift right. 


alamb (approved):
Thank you @mobuchowski  -- sorry for the delayed review.  I like that this is in the context of a data type (so won't affect how `ARRAY(...)` works when used as a function

alamb:
I merged up from main to resolve a merge conflict

mobuchowski:
Thanks!

PR #495, opened 2022-05-11
set: allow negative ident values

mobuchowski:
Sometimes, SET statement values can start with `Token::Minus`. For example - as in test -  setting Java's properties.


alamb (approved):
LGTM -- thanks @mobuchowski

PR #499, opened 2022-05-17
feat: Convert IS TRUE|FALSE to expression

ovr:
Hello!

Support for `IS TRUE|FALSE` operator was introduced in https://github.com/sqlparser-rs/sqlparser-rs/pull/474 by [yuval-illumex](https://github.com/yuval-illumex), but it has a limitation.

There is a difference between `IS TRUE|false` and `= true|false`, it how it handles nulls:

```sql
CREATE TABLE test_boolean (
    col boolean
);
INSERT INTO test_boolean VALUES (true), (false), (null);
SELECT col, col IS FALSE, col = false, col is true, col = true from test_boolean;
```

image

As you can see from the output, it handles `null` values differently.

Thanks

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -233,6 +233,10 @@ pub enum Expr {
     },
     /// CompositeAccess (postgres) eg: SELECT (information_schema._pg_expandarray(array['i','i'])).n
     CompositeAccess { expr: Box<Expr>, key: Ident },
+    /// `IS FALSE` operator
+    IsFalse(Box<Expr>),
+    /// `IS TRUE` operator
+    IsTrue(Box<Expr>),
     /// `IS NULL` operator
     IsNull(Box<Expr>),
     /// `IS NOT NULL` operator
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -236,6 +236,10 @@ pub enum Expr {
         expr: Box<Expr>,
         key: Ident,
     },
+    /// `IS FALSE` operator
```
alamb, line 239: It is somewhat unfortunate we don't have `Expr::Is(Box<Expr>, Box<Expr>)` which would handle all these cases.

What do you think @ovr ? Should I file a ticket to try and reduce the duplication in `Expr`?

alamb (approved):
I am not a huge fan of the growing number of `Expr::Is*` variants, but that is not something introduced in this PR.

alamb:
Screen Shot 2022-05-22 at 3 41 24 PM

@ovr  github tells me there is a conflict to resolve in this PR -- I haven't been able to fix such issues in the past as I don't have the ability to push commits to your PR

Can you please resolve the conflicts, and then I will merge this PR?

Thank you!

src/ast/mod.rs
ovr: IMHO: It's a breaking change, which doesn't give any advantages over this. Right now, it's really easy to implement a match pattern over this because names are really helpful.

ovr:
Rebased ✅

alamb:
Thanks @ovr

Issue #305, opened 2021-04-01
Stack overflow when parsing SQL consisting of many imbalanced open parenthesis


5225225:
This code segfaults with a stack overflow on both 0.9.0 from crates.io as well as loading from git (Version 35ef0eee3)

```rust
use sqlparser::dialect::GenericDialect;
use sqlparser::parser::Parser;

fn main() {
    let sql = "(".repeat(1024);

    let dialect = GenericDialect {};

    let _ = Parser::parse_sql(&dialect, &sql);
}
```

46bit:
@alamb Hi, pinging you as a maintainer. I want to use this crate, but I'd need to fix this issue to prevent users from triggering stack overflows. Hoping to get some feedback on whether you'd accept a fix.

Right now the `Parser` struct doesn't do any tracking of stack size. What I'm thinking is adding depth-limited variants of each method that take an extra `depth: usize` parameter. Each method would return a `ParserError` if the recursion got deeper than a `maxDepth: Option<usize>` field on `Parser`, preventing a stack overflow. The existing parse methods would then call out to the depth-limited variants supplying `depth: 0`.

Does that sound like a PR you'd be open to?

alamb:
Hi @46bit  -- Thank you for offering.

The idea of adding a `max_depth` parameter to the existing (non `pub`) methods seems reasonable to me. Hopefully the interface that most external users see would remain the same

Also, for what it is worth, I have seen rust stack sizes *significantly* larger in debug code than release code -- so the actual stack limit may be much higher in practice in release builds (though certainly would still exist)

46bit:
@alamb Thanks for such a quick response, I appreciate it. I'll get working on the PR soonish and let you know how it goes :)

5225225:
Might be worth looking at what `serde_json` does, since I know they definitely have a depth limit by default. Their limit is 128.

46bit:
serde-json is using a `check_recursion { ... }` macro each time it recurses deeper: https://github.com/serde-rs/json/blob/master/src/de.rs#L1273-L1304. That works OK for them, they only use it 8 times. But sqlparser branches a lot more and I think we'd use it hundreds of times:

```rust
    pub fn parse_statement(&mut self) -> Result<Statement, ParserError> {
        match self.next_token() {
            Token::Word(w) => match w.keyword {
                Keyword::DESCRIBE => Ok(check_recursion! { self.parse_explain(true)? }),
[...]

PR #501, opened 2022-05-18
Prevent stack overflows by limiting recursion

46bit:
Addresses #305

Until now it has been possible to trigger a stack overflow by passing crafted inputs (e.g. a string of 1000 `(`). This is problematic when using the library to parse user input.

This commit adds a recursion limit so long as the library is compiled with `std`. The limit ensures that malicious inputs cannot trigger a stack overflow, by preventing excessively deep recursion.

Since the SQL parser is so big, I've added this protection to every `Parser` method that returns a `Result`. The actual recursive patterns in the AST are more limited, but with such a large parser it's much harder to identify them and put targeted protections in place.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ path = "src/lib.rs"
 
 [features]
 default = ["std"]
-std = []
+std = ["scopeguard"]
 # Enable JSON output in the `cli` example:
 json_example = ["serde_json", "serde"]
 
@@ -32,6 +32,7 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 # of dev-dependencies because of
 # https://github.com/rust-lang/cargo/issues/1596
 serde_json = { version = "1.0", optional = true }
+scopeguard = { version = "1.1.0", optional = true }
 
 [dev-dependencies]
 simple_logger = "2.1"
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -21,8 +21,14 @@ use alloc::{
     vec::Vec,
 };
 use core::fmt;
+#[cfg(feature = "std")]
+use std::cell::Cell;
+#[cfg(feature = "std")]
+use std::rc::Rc;
 
 use log::debug;
+#[cfg(feature = "std")]
+use scopeguard::defer;
 
 use crate::ast::*;
 use crate::dialect::*;
@@ -33,6 +39,7 @@ use crate::tokenizer::*;
 pub enum ParserError {
     TokenizerError(String),
     ParserError(String),
+    RecursionLimitExceeded,
 }
 
 // Use `Parser::expected` instead, if possible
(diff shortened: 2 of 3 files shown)


examples/parse_select.rs
```diff
@@ -16,10 +16,7 @@ use sqlparser::dialect::GenericDialect;
 use sqlparser::parser::*;
 
 fn main() {
-    let sql = "SELECT a, b, 123, myfunc(b) \
-               FROM table_1 \
-               WHERE a > b AND b < 100 \
-               ORDER BY a DESC, b";
+    let sql = "((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((";
```
alamb, line 19: Why is this example changed? Maybe it is left over from earlier development

alamb (commented):
Thank you @46bit -- This is quite cool and close I think. 

Given how low level the `sqlparser`  crate is, I would like to minimize the dependencies as much as possible. https://crates.io/crates/scopeguard has a bunch of `unsafe` as well .

The use of `Rc` and `unsafe` to check a value that requires annotation everywhere  somewhat overly complex to me

I was wondering how hard it would be to make a macro that avoids the need for `scope_guard` entirely. For example Instead of:

```rust
    pub fn parse_msck(&mut self) -> Result<Statement, ParserError> {
        check_recursion_depth!(self);
```

I wonder if it would be possible to create, via macro functions like

```rust
    pub fn parse_msck(&mut self) -> Result<Statement, ParserError> {
        self.remaining_depth -= 1;
        if self.remaining_depth == 0 {
            return Err(ParserError::RecursionLimitExceeded);
        }
        let res = parse_msck_impl(self)?
        self.remaining_depth += 1;
    }
    pub fn parse_msck_impl(&mut self) -> Result<Statement, ParserError> {
      // existing code here
    }
```

We could make it even more encapsulated perhaps via a RAA struct like:

```rust
struct StackChecker<'a> {
 parser: &'a mut Parser,
} 

impl StackChecker {
  pub fn try_new(&mut parser) -> Result<Self> {
    // stack check here, and decrement depth
  }
}

impl Drop for StackChecker {
  fn drop(&mut self) {
   self.parser.remaining_depth -= 1
}
}
```

 If we truly need scopeguard, I think in this case we should contemplate incorporating the code by value (aka copy it in)

examples/parse_select.rs
46bit: Sorry about that, will fix

46bit:
@alamb Thank you for the detailed review, I agree with pretty much all your points :) The overflow protection seems effective (we've been fuzzing it for a solid week now and haven't found a workaround) yet I definitely agree on finding a nicer implementation. I'll revisit this PR in the next few weeks as soon as time allows and work on getting it into a better place.

alamb:
>  @alamb Thank you for the detailed review, I agree with pretty much all your points :) The overflow protection seems effective (we've been fuzzing it for a solid week now and haven't found a workaround) yet I definitely agree on finding a nicer implementation. I'll revisit this PR in the next few weeks as soon as time allows and work on getting it into a better place.

Awesome -- I can't wait to see it. Thank you for your contribution and sticking with it.

alamb:
Closing as stale -- please reopen if you still plan to work on it

alamb:
I have an proposal, based partly on this PR, here: https://github.com/sqlparser-rs/sqlparser-rs/pull/764

PR #506, opened 2022-05-24
Distinguish between `SELECT * FROM foo` and `SELECT * FROM foo()`

### LINKED ISSUE #405: Cannot distinguish between `SELECT * FROM foo` and `SELECT * FROM foo()`
`sqlparser::ast::TableFactor::Table` just represents the arguments as a `Vec<FunctionArg>` both
`SELECT * FROM foo` and `SELECT * FROM foo()` and are parsed as:

```
Table { name: ObjectName([Ident { value: "foo", quote_style: None }]), alias: None, args: [], with_hints: [] }
```

This is apparently a regression that was introduced in 5652b4676c65f0875fd8c4849dd7cb27c819bef6 (#73).

razzolini-qpq:
Fixes #405. As stated in that issue,

> This is apparently a regression that was introduced in https://github.com/sqlparser-rs/sqlparser-rs/commit/5652b4676c65f0875fd8c4849dd7cb27c819bef6 (https://github.com/sqlparser-rs/sqlparser-rs/pull/73)

To fix it, I reintroduced the `Option` around the vector of arguments, which was removed in 5652b4676c65f0875fd8c4849dd7cb27c819bef6 even though it wasn't actually redundant. I also added a test which ensures that `SELECT * FROM fn()` is not modified after a serialization round-trip, and I added a brief explanation of the different meanings of `None` and `Some` (including `Some(vec![])`) to the documentation.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -337,7 +337,11 @@ pub enum TableFactor {
         /// Arguments of a table-valued function, as supported by Postgres
         /// and MSSQL. Note that deprecated MSSQL `FROM foo (NOLOCK)` syntax
         /// will also be parsed as `args`.
-        args: Vec<FunctionArg>,
+        ///
+        /// This field's value is `Some(v)`, where `v` is a (possibly empty)
+        /// vector of arguments, in the case of a table-valued function call,
+        /// whereas it's `None` in the case of a regular table name.
+        args: Option<Vec<FunctionArg>>,
         /// MSSQL-specific `WITH (...)` hints such as NOLOCK.
         with_hints: Vec<Expr>,
     },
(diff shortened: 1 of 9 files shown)


src/ast/query.rs
```diff
@@ -337,7 +337,11 @@ pub enum TableFactor {
         /// Arguments of a table-valued function, as supported by Postgres
         /// and MSSQL. Note that deprecated MSSQL `FROM foo (NOLOCK)` syntax
         /// will also be parsed as `args`.
-        args: Vec<FunctionArg>,
+        ///
+        /// This field's value is `Some(v)`, where `v` is a (possibly empty)
```
alamb, line 341: 👍

alamb (approved):
LGTM -- thank you @razzolini-qpq

PR #512, opened 2022-05-28
Support `DATETIME` keyword

komukomo:
BigQuery can parse `DATETIME`.
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#datetime_literals

Before this PR, `SELECT DATETIME '1999-01-01 01:23:34.45'` was parsed as DATETIME column with '1999-01-..' alias.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -71,6 +71,8 @@ pub enum DataType {
     Date,
     /// Time
     Time,
+    /// Datetime
+    Datetime,
     /// Timestamp
     Timestamp,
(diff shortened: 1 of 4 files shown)


komukomo:
Also, `DATETIME` was treated as a `Custom` type in `SELECT CAST('1999-01-01 01:23:34.45' AS DATETIME)` for example.

ref. https://github.com/sqlparser-rs/sqlparser-rs/issues/503

alamb (approved):
Looks good to me -- thank you @komukomo

Issue #298, opened 2021-03-13
MySQL allows integer interval values


duvall-eb:
The usual format for an interval is `INTERVAL 'value' unit`, but MySQL allows `'value'` to be an expression that is either a string or a number. For example:
```
SELECT DATE_SUB('2018-05-01',INTERVAL '1' YEAR);
SELECT DATE_SUB('2018-05-01',INTERVAL 1 YEAR);
```
both return `2017-05-01`. Currently, the latter syntax is not supported.

kyle-mccarthy:
I am running into this issue as well. mysql also allows for referencing columns in the unit part of the interval.

46bit:
ClickHouse is the same, it supports `INTERVAL 1 HOUR` without requiring the `1` to be quoted. https://clickhouse.com/docs/en/sql-reference/operators/#interval

LucaCappelletti94:
Closing as resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/517

PR #517, opened 2022-06-07
fix: Support `expr` instead of `string` for argument to `interval`

togami2864:
In Big Query and MySQL, the following query is valid:
```
SELECT DATE_ADD(DATE "2008-12-25", INTERVAL 5 DAY);
```
https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_add
https://www.w3schools.com/mysql/func_mysql_date_add.asp

But now if the expression is number, an error occurred because the type of the value in `interval` is `String`
https://github.com/sqlparser-rs/sqlparser-rs/blob/ca15a4edbf1de4c00304b12e44d3adf9ac729406/src/ast/value.rs#L53

This PR replaces `String` with `Expr` to run correctly in `Interval`

One concern thing is that should I validate when the other dialect like `PostgresSQL`, `sqlite`. I don't know if the query is valid except for Big query and MySQL.

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -11,6 +11,8 @@
 // limitations under the License.
 
 #[cfg(not(feature = "std"))]
+use alloc::boxed::Box;
+#[cfg(not(feature = "std"))]
 use alloc::string::String;
 use core::fmt;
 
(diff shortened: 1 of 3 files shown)


46bit:
Great to see this. There's related discussion in #298.

alamb (approved):
Looks good to me -- thank you @togami2864

alamb:
@togami2864  -- there appear to be some CI failures

togami2864:
@alamb I fixed all CI failures. Thanks.

PR #520, opened 2022-06-09
snowflake column identifier

pawel-big-lebowski:
Signed-off-by: Pawel Leszczynski <leszczynski.pawel@gmail.com>

resolves https://github.com/sqlparser-rs/sqlparser-rs/issues/519

According to documentation, Snowflake allows:
 * [query data staged in files](https://docs.snowflake.com/en/user-guide/querying-stage.html#required-parameters) with an identifier of a form `[<alias>.]$<file_col_num>[.<element>]`
* [copy into table](https://docs.snowflake.com/en/user-guide/data-load-transform.html) with an column identifier containing $ character like `select t.$1,t.$2,t.$3 from @~/datafile.csv.gz t;`
* [get-path](https://docs.snowflake.com/en/sql-reference/functions/get_path.html) like `select v:attr[0].name from vartab;`

Although the feature seems to be snowflake specific, it cannot be implemented within snowflake dialect. It's because tokens starting with `$` character are tokenized into `Token::Placeholder`. That's why parser's change is required.

Initial OpenLineage project issue: https://github.com/OpenLineage/OpenLineage/issues/814.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -217,6 +217,18 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// Multi-part identifier with a column number, e.g. [alias.]$file_col_num[.element] (snowflake)
+    CompoundIdentifierWithColumnNumber {
+        alias: Ident,
+        col_num: String,
+        element: Option<Ident>
+    },
+    /// Operator that extracts a value from semi-structured like v:attr[0].name (snowflake)
+    GetPathExpression {
+        idents: Vec<Ident>,
+        col_num: Option<String>,
+        path: Vec<String>
+    },
     /// JSON access (postgres)  eg: data->'tags'
     JsonAccess {
         left: Box<Expr>,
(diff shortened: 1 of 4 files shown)


alamb:
There appear to be a few CI failures on this PR

alamb (commented):
Thank you for the contribution @pawel-big-lebowski  I will try and find time to review this PR tomorrow.

src/ast/mod.rs
```diff
@@ -217,6 +217,18 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// Multi-part identifier with a column number, e.g. [alias.]$file_col_num[.element] (snowflake)
+    CompoundIdentifierWithColumnNumber {
```
alamb, line 221: In terms of supporting things like `a.$1` -- what would you think about modeling this as a a two part `CompoundIdent` identifier `["a", "$1"]` rather than a new variant?

I think the parser would still have to be changed, but I think that approach would keep the AST and the consumption of the AST smaller with fewer dialect specific structures

tests/sqlparser_snowflake.rs
```diff
@@ -99,6 +99,12 @@ fn test_single_table_in_parenthesis() {
     );
 }
 
+#[test]
+fn test_snowflake_variables() {
```
alamb, line 103: I think it would be good to have a test that the parsed structure is actually what is expected -- for example https://github.com/sqlparser-rs/sqlparser-rs/blob/c884fbc/tests/sqlparser_common.rs#L138-L168

alamb (commented):
Thank you @pawel-big-lebowski  -- I spent some time researching this feature.

First, I think it might help this PR if you split it into 2 pieces so we can discuss each feature (`$identifiers` and `get path`) separately. 


Also, 🤔  it seems as though snowflake also supports identifiers like `column$FILENAME` -- is this something you want to explore?

https://docs.snowflake.com/en/user-guide/querying-metadata.html#metadata-columns

alamb:

AugustoFKL:
@alamb @pawel-big-lebowski is closing this for now because it's almost a year without an update. 

@pawel-big-lebowski, ping me if you will work on this code specifically, if so, I open the PR again :)

alamb:
Thanks for cleaning this up @AugustoFKL

PR #523, opened 2022-06-13
add more consistency in ast

frolovdev:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/472

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -426,7 +426,7 @@ impl<'a> Parser<'a> {
                 Keyword::CASE => self.parse_case_expr(),
                 Keyword::CAST => self.parse_cast_expr(),
                 Keyword::TRY_CAST => self.parse_try_cast_expr(),
-                Keyword::EXISTS => self.parse_exists_expr(),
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -984,6 +984,25 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_not(&mut self) -> Result<Expr, ParserError> {
+        match self.peek_token() {
+            Token::Word(w) => match w.keyword {
+                Keyword::EXISTS => {
+                    let negated = self.parse_keyword(Keyword::EXISTS);
```
alamb, line 991: don't we know here that `negated` is true? I think the code might be clearer if this were something more like

```rust
let negated = true;
self.parse_keyword(Keyword::Exists);
self.parse_exists_expr(negated)
```

What do you think?

alamb (approved):
Thanks @frolovdev -- looks like a nice cleanup.

cc @andygrove

alamb:
Looks good -- thanks @frolovdev

PR #525, opened 2022-06-18
Distinguish between `INT` and `INTEGER` types

### LINKED ISSUE #398: distinguishing between `Int` and `Integer`
Currently the parser will parse `INT` and `INTEGER` into the same AST representation. It will later serialize it back all into `INT`. Unfortunatly, in SQLite `INT` is not recognized as a valid type. This means that primary keys with `INT` as a type do not enforce type checking nor do they support autoincrementing functionality. 
Thus, parsing this file and displaying it back again would introduce a runtime error: 
```sql
CREATE TABLE foo (
   bar INTEGER NOT NULL PRIMARY KEY,
   baz TEXT
); 
INSERT INTO foo (baz) values ('baz'); 
```
I forked the project when I stumbled into this issue (I addressed it by defaulting to `INTEGER` instead of `INT`), but I still wanted to mention it in case you weren't aware of this.

frolovdev:
@alamb (you said that I can ping you :d)

Closes #398

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -59,7 +59,7 @@ pub enum DataType {
     Integer(Option<u64>),
     /// Unsigned integer with optional display width e.g. INT UNSIGNED or INT(11) UNSIGNED
     UnsignedInt(Option<u64>),
-    /// Unsigned integer with optional display width e.g. INT UNSIGNED or INT(11) UNSIGNED
+    /// Unsigned integer with optional display width e.g. INTGER UNSIGNED or INTEGER(11) UNSIGNED
     UnsignedInteger(Option<u64>),
     /// Big integer with optional display width e.g. BIGINT or BIGINT(20)
     BigInt(Option<u64>),
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -138,7 +138,9 @@ impl fmt::Display for DataType {
             DataType::UnsignedInt(zerofill) => {
                 format_type_with_optional_length(f, "INT", zerofill, true)
             }
-            DataType::Integer(zerofill) => format_type_with_optional_length(f, "INTEGER", zerofill, false),
+            DataType::Integer(zerofill) => {
+                format_type_with_optional_length(f, "INTEGER", zerofill, false)
+            }
             DataType::UnsignedInteger(zerofill) => {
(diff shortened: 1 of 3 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -35,7 +35,7 @@ fn parse_create_table_with_defaults() {
             activebool boolean DEFAULT true NOT NULL,
             create_date date DEFAULT now()::text NOT NULL,
             last_update timestamp without time zone DEFAULT now() NOT NULL,
-            active integer NOT NULL
+            active int NOT NULL
```
frolovdev, line 38: I change the test to cover both bariants

src/ast/data_type.rs
```diff
@@ -55,8 +55,12 @@ pub enum DataType {
     UnsignedSmallInt(Option<u64>),
     /// Integer with optional display width e.g. INT or INT(11)
     Int(Option<u64>),
+    /// Integer with optional display width e.g. INTEGER or INTEGER(11)
+    Integer(Option<u64>),
     /// Unsigned integer with optional display width e.g. INT UNSIGNED or INT(11) UNSIGNED
     UnsignedInt(Option<u64>),
+    /// Unsigned integer with optional display width e.g. INT UNSIGNED or INT(11) UNSIGNED
```
alamb, line 62:
```suggestion
    /// Unsigned integer with optional display width e.g. INTGER UNSIGNED or INTEGER(11) UNSIGNED
```

src/parser.rs
```diff
@@ -2859,14 +2859,22 @@ impl<'a> Parser<'a> {
...
-                Keyword::INT | Keyword::INTEGER => {
+                Keyword::INT => {
                     let optional_precision = self.parse_optional_precision();
                     if self.parse_keyword(Keyword::UNSIGNED) {
                         Ok(DataType::UnsignedInt(optional_precision?))
                     } else {
                         Ok(DataType::Int(optional_precision?))
                     }
                 }
+                Keyword::INTEGER => {
```
alamb, line 2870: 👍

frolovdev:
> Thank you @frolovdev  -- other than the docstring this looks great.
> 
> 
> 
> 👍 

Applied

alamb:
Thanks again @frolovdev !

PR #527, opened 2022-06-24
Add support for ClickHouse DDL query syntax (`on cluster`)

andyrichardson:
Add support for ClickHouse DDL query syntax [(source)](https://clickhouse.com/docs/en/sql-reference/distributed-ddl/)

This is an example only on `CREATE TABLE` statements. If you're happy to go ahead with supporting this, I can add this to all `CREATE` `DROP` `ALTER` and `RENAME` queries

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -867,6 +867,8 @@ pub enum Statement {
         default_charset: Option<String>,
         collation: Option<String>,
         on_commit: Option<OnCommit>,
+        /// Click house "ON CLUSTER" clause: 
+        /// <https://clickhouse.com/docs/en/sql-reference/distributed-ddl/>
         on_cluster: Option<String>,
     },
     /// SQLite's `CREATE VIRTUAL TABLE .. USING <module_name> (<module_args>)`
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -867,6 +867,7 @@ pub enum Statement {
         default_charset: Option<String>,
         collation: Option<String>,
         on_commit: Option<OnCommit>,
+        on_cluster: Option<String>,
```
alamb, line 870:
```suggestion
        /// Click house "ON CLUSTER" clause: 
        /// <https://clickhouse.com/docs/en/sql-reference/distributed-ddl/>
        on_cluster: Option<String>,
```

alamb (approved):
Looks great -- thank you @andyrichardson. I think adding support for other DDL statements sounds like a great idea. 

The only possible thing that might be worth considering is to only parse `ON CLUSTER` when using the `generic` and `ClickHouse` dialects

src/tokenizer.rs
```diff
@@ -243,7 +243,7 @@ pub struct Word {
     /// The value of the token, without the enclosing quotes, and with the
     /// escape sequences (if any) processed (TODO: escapes are not handled)
     pub value: String,
-    /// An identifier can be "quoted" (&lt;delimited identifier> in ANSI parlance).
+    /// An oidentifier can be "quoted" (&lt;delimited identifier> in ANSI parlance).
```
alamb, line 246:
```suggestion
    /// An identifier can be "quoted" (&lt;delimited identifier> in ANSI parlance).
```

PR #529, opened 2022-06-26
Count characters instead of bytes

michael-2956:
For queries such as `SELECT "なにか" FROM Y WHERE "なにか" = 'test;` which are using UTF-8 characters as column names, counting lengths as byte counts (which `len()` does) may cause bugs.

For instance, the above query would produce an error `Unterminated string literal at Line: 1, Column 47`, while the correct column number is 35.

So, simply use `.chars().count()` instead. Replacing `len()` with this produces correct output.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -351,11 +351,11 @@ impl<'a> Tokenizer<'a> {
                 }
 
                 Token::Whitespace(Whitespace::Tab) => self.col += 4,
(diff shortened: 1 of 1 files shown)


alamb (commented):
Thank you for the contribution @michael-2956 ! 

Can you please add a test to this PR (so that this feature doesn't potentially get broken by a future change without us knowing)

alamb (approved):
LGTM -- thank you @michael-2956

PR #533, opened 2022-07-07
Add support for multi-part identifiers with colons

UlfarErl:
Snowflake SQL uses colons for multi-part identifiers as part of semi-structured data support, such that a multi-part identifier may include both periods and colons.  I didn't see a way to add this as part of the Snowflake-dialect support, so proposing to add here.  Don't see a big downside.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -442,14 +442,14 @@ impl<'a> Parser<'a> {
                 // Here `w` is a word, check if it's a part of a multi-part
                 // identifier, a function call, or a simple identifier:
                 _ => match self.peek_token() {
-                    Token::LParen | Token::Period => {
+                    Token::LParen | Token::Period | Token::Colon => {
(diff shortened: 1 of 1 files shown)


alamb (commented):
Thank you for the suggestion and contribution @UlfarErl. Supporting `:` seems fine to me -- can you please
1.  provide some tests for this behavior (so that we don't break it accidentally in the future)?
2. Provide a link to the snowflake documentation for this feature?

Thanks so much

UlfarErl:
Thanks for taking a look.

The relevant documentation is here: https://docs.snowflake.com/en/user-guide/querying-semistructured.html

After looking at it further, I'm not sure if this pull request should be merged. It only handles one particular special case of the generic semi-structured support, and it does that in a way that is too permissive.  It solved my problem, but it is really a hack.  Adding full support for semi-structured syntax, and integrating that into the codebase in a way that allows different variants is a pretty big lift which I don't have time for.  I'm going to close this for now.

PR #540, opened 2022-07-13
Boxed Query body to save some stack space

5tan:
Boxing `Query::body` reduced size of `Query` from 1312B to 592B.

In following test of standalone parser runs (7MB stack) allowed subquery nesting jumped from 239 to 286.
```bash
x=286 ; echo 'print("SELECT * FROM {}ttt{}".format("("*'$x',")"*'$x'))' | python3 > test.sql && cargo run --example cli --manifest-path Cargo.toml test.sql --snowflake
```

This makes significant difference when parsing "recursive queries" in environments with reduced stack size (e.g. like [tokio spawned threads](https://durch.github.io/rust-goauth/tokio/runtime/struct.Builder.html#:~:text=The%20default%20stack%20size%20for,to%20change%20in%20the%20future) - 2MB).

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -26,7 +26,7 @@ pub struct Query {
     /// WITH (common table expressions, or CTEs)
     pub with: Option<With>,
     /// SELECT or UNION / EXCEPT / INTERSECT
-    pub body: SetExpr,
+    pub body: Box<SetExpr>,
     /// ORDER BY
     pub order_by: Vec<OrderByExpr>,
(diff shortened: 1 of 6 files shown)


alamb:
Thanks @5tan  -- I think this change looks reasonable to me. For what it is worth, I found that significant difference in rust when compiling with `--release` -- https://github.com/apache/arrow-datafusion/pull/1047 for details

alamb (approved):

PR #541, opened 2022-07-14
delete: add using clause, possibility of using aliases

mobuchowski:
[PostgreSQL](https://www.postgresql.org/docs/current/sql-delete.html) and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/delete.html) have possibility of using aliases in `DELETE` statements.

PostgreSQL and [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/delete.html) have `USING` clause. 

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -253,7 +253,12 @@ fn parse_delete_statement() {
     match verified_stmt(sql) {
         Statement::Delete { table_name, .. } => {
             assert_eq!(
-                ObjectName(vec![Ident::with_quote('"', "table")]),
+                TableFactor::Table {
+                    name: ObjectName(vec![Ident::with_quote('"', "table")]),
+                    alias: None,
+                    args: None,
+                    with_hints: vec![]
+                },
                 table_name
             );
         }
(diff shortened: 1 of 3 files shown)


tests/sqlparser_common.rs
```diff
@@ -269,11 +274,20 @@ fn parse_where_delete_statement() {
...
+                TableFactor::Table {
+                    name: ObjectName(vec![Ident::new("foo")]),
+                    alias: None,
+                    args: None,
+                    with_hints: vec![]
+                },
+                table_name,
+            );
 
+            assert_eq!(None, using);
```
alamb, line 290: Can you please add a positive test for `using` as well (aka a query that has a `USING` clause)?

alamb (commented):
Thank you @mobuchowski  -- looking good. I think this PR just needs one more test

tests/sqlparser_common.rs
alamb: No this is fine -- I just wanted to add some coverage so that we didn't accidentally break the functionality in the future but have all the tests pass

PR #547, opened 2022-07-29
Initial fork to create sqlgen-rs

jonmmease:
This branch is the initial fork of `sqlparser-rs` into the start of `sqlgen-rs`.

### Query only
All of the DDL constructs, other than `Query`, have been removed from the AST and parsing logic. sqlgen is only concerned with queries.

### Single parsing dialect
 All of the dialect-based parsing logic has been removed. The configuration for the generic dialect has been inlined. sqlgen is intended to input a single standard SQL dialect.

### Dialect aware string generation
The existing logic to format the AST into an SQL string has refactored to support configuration with a `Dialect` struct. Currently the only thing the dialect struct configures is how identifiers are to be quoted in the output SQL string.  But the idea is that we'll add more and more configuration options to this dialect struct to handle the customizations required across a range of SQL dialects.


jonmmease:
Apologies, this PR was not intended for the upstream repo!  I'm experimenting with a fork.

PR #548, opened 2022-07-30
[postgres] Add support for custom binary operators

iskakaushik:
More details about operators in general are at: https://www.postgresql.org/docs/current/sql-createoperator.html. This patch attempts to parse `SELECT` queries that reference an operator using `OPERATOR(<optional_schema>.<operator_name>)` syntax. Tests added in the patch illustrate a common use-case of this feature.

This is a PostgreSQL extension. There are no provisions for user-defined operators in the SQL standard.

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -12,6 +12,8 @@
 
 use core::fmt;
 
+#[cfg(not(feature = "std"))]
+use alloc::string::String;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -17,6 +17,8 @@ use alloc::string::String;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+use super::display_separated;
+
 /// Unary operators
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -88,17 +90,11 @@ pub enum BinaryOperator {
     PGRegexIMatch,
     PGRegexNotMatch,
     PGRegexNotIMatch,
-    PGCustomBinaryOperator(PGCustomOperator),
-}
-
-/// PostgreSQL-specific custom operator.
-///
-/// See [CREATE OPERATOR](https://www.postgresql.org/docs/current/sql-createoperator.html) for more information.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct PGCustomOperator {
-    pub schema: Option<String>,
-    pub name: String,
+    /// PostgreSQL-specific custom operator.
+    ///
+    /// See [CREATE OPERATOR](https://www.postgresql.org/docs/current/sql-createoperator.html)
+    /// for more information.
+    PGCustomBinaryOperator(Vec<String>),
 }
 
 impl fmt::Display for BinaryOperator {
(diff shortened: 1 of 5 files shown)


iskakaushik:
@alamb could you please help with triggering the workflow?

src/parser.rs
```diff
@@ -1159,6 +1159,33 @@ impl<'a> Parser<'a> {
                     }
                 }
                 Keyword::XOR => Some(BinaryOperator::Xor),
+                Keyword::OPERATOR if dialect_of!(self is PostgreSqlDialect) => {
+                    debug!("parsing custom operator");
```
alamb, line 1163:
```suggestion
```

Not sure that will add much after the initial implementation

alamb (commented):
Thanks @iskakaushik  -- I'll trigger the tests

src/parser.rs
iskakaushik: Good call, I have removed this.

ovr:
Just an addition about real-life usage:

image

Official `psql` CLI uses it under the hood behind some backslash commands.

src/parser.rs
```diff
@@ -1159,6 +1159,32 @@ impl<'a> Parser<'a> {
                     }
                 }
                 Keyword::XOR => Some(BinaryOperator::Xor),
+                Keyword::OPERATOR if dialect_of!(self is PostgreSqlDialect) => {
```
alamb, line 1162: Given @ovr 's comment here, perhaps we should also add support when parsing in `GenericDialect`? https://github.com/apache/arrow-datafusion/pull/3037#issuecomment-1205609941

src/ast/operator.rs
```diff
@@ -86,41 +88,59 @@ pub enum BinaryOperator {
...
+}
+
+/// PostgreSQL-specific custom operator.
+///
+/// See [CREATE OPERATOR](https://www.postgresql.org/docs/current/sql-createoperator.html) for more information.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct PGCustomOperator {
+    pub schema: Option<String>,
+    pub name: String,
```
alamb, line 101: What would you think about using the pre-existing `Identifier` here (that supports various qualified names)

like:

```suggestion
pub struct PGCustomOperator {
    pub ident: ObjectName
```

Thank you could use `parse_object_name` instead of custom parsing logic. 

https://github.com/sqlparser-rs/sqlparser-rs/blob/076b587bb26f725a9a172c2ca4da373e83937158/src/parser.rs#L3114-L3125

This would likely require less code as well as handle parsing more cases (like `OPERATOR("~")` -- aka quoted strings)

alamb (commented):
Thanks @iskakaushik  -- this is looking close.

alamb:
@iskakaushik I am planning to make a sqlparser release sometime soon -- I think this PR has just a few more items and it could be ready to merge. Let me know what you think.

iskakaushik:
> @iskakaushik I am planning to make a sqlparser release sometime soon -- I think this PR has just a few more items and it could be ready to merge. Let me know what you think.

I will address the feedback today 😊

src/ast/operator.rs
iskakaushik: I tried doing this, but looks like `Identifier` doesn't allow for `Token`s that are un-quoted operators, for e.g: `~` is not a valid identifier. I am not sure if extending the `parse_identifier` to allow for un-quoted operators is the right call here.

IMO we could start with the existing mechanism of schema qualifier operator name (though I would've liked to use the `parse_object_name` logic) and can extend it as needed later. Let me know.

src/parser.rs
iskakaushik: Done! I kept the name as `PGCustomOperator` but happy to change it as well.

src/ast/operator.rs
alamb, line 101: Thank you for trying -- Something about hard coding "schema" seems not right to me here -- at the very least because there can also be a database in identifiers `https://www.postgresql.org/docs/current/ddl-schemas.html`  -- e.g. `database.schema.table`, so maybe it should be a `Vec<String>`

I think given the special set of allowed characters described in https://www.postgresql.org/docs/current/sql-createoperator.html in `The operator name is a sequence of up to NAMEDATALEN-1 (63 by default) characters from the following list:` to properly parse this limitation some extra code would be needed

Let me give it a shot

alamb:
@iskakaushik  here is my proposal: https://github.com/iskakaushik/sqlparser-rs/pull/1

let me know what you think

iskakaushik:
@alamb  I like you approach. The comment you made about converting to `(ident: Vec<Identifier>, op: String)` is a good idea, but I think this might be good enough for now as you mentioned.

Thanks for fixing up my patch, really appreciate it!

alamb:
I will fix CI

alamb (approved):

alamb:
Thanks again @iskakaushik

PR #551, opened 2022-08-01
Support `NestedJoin` with an alias

waitingkuo:
closes https://github.com/sqlparser-rs/sqlparser-rs/issues/537 , https://github.com/apache/arrow-datafusion/issues/2867

origin nestedjoin doesn't have alias. this pr add alias for nestedjoin

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -375,7 +375,10 @@ pub enum TableFactor {
     ///
     /// The parser may also accept non-standard nesting of bare tables for some
     /// dialects, but the information about such nesting is stripped from AST.
-    NestedJoin(Box<TableWithJoins>),
+    NestedJoin {
+        table_with_joins: Box<TableWithJoins>,
+        alias: Option<TableAlias>,
+    },
 }
 
 impl fmt::Display for TableFactor {
@@ -438,7 +441,16 @@ impl fmt::Display for TableFactor {
                 }
                 Ok(())
             }
-            TableFactor::NestedJoin(table_reference) => write!(f, "({})", table_reference),
+            TableFactor::NestedJoin {
+                table_with_joins,
+                alias,
+            } => {
+                write!(f, "({})", table_with_joins)?;
+                if let Some(alias) = alias {
+                    write!(f, " AS {}", alias)?;
+                }
+                Ok(())
+            }
         }
     }
 }
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -3676,7 +3676,7 @@ fn parse_derived_tables() {
     let from = only(select.from);
     assert_eq!(
         from.relation,
-        TableFactor::NestedJoin(Box::new(TableWithJoins {
+        TableFactor::NestedJoin {table_with_joins: Box::new(TableWithJoins {
             relation: TableFactor::Derived {
                 lateral: false,
                 subquery: Box::new(verified_query("(SELECT 1) UNION (SELECT 2)")),
@@ -3694,7 +3694,7 @@ fn parse_derived_tables() {
                 },
                 join_operator: JoinOperator::Inner(JoinConstraint::Natural),
             }],
-        }))
+        }), alias: None}
     );
 }
 
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -3481,7 +3481,7 @@ fn parse_natural_join() {
         vec![natural_join(JoinOperator::FullOuter, None)]
     );
 
-    // natural join with alias
+    // natural join another table with alias
     assert_eq!(
         only(&verified_only_select("SELECT * FROM t1 NATURAL JOIN t2 AS t3").from).joins,
         vec![natural_join(JoinOperator::Inner, table_alias("t3"))]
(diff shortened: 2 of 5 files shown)


waitingkuo:
help wanted.

this pr break [this test](https://github.com/waitingkuo/sqlparser-rs/blob/65755136543b5fb6013ebae0ab500961b3b42817/tests/sqlparser_snowflake.rs#L134)
```rust
    let res = snowflake_and_generic().parse_sql_statements("SELECT * FROM (a NATURAL JOIN b) c");
    assert_eq!(
        ParserError::ParserError("Expected end of statement, found: c".to_string()),
        res.unwrap_err()
    );
```

Could anyone tell me how to deal with this?

thanks~

alamb:
>this pr break this test

That test is validating that there is a parser error, right @waitingkuo ?

It looks like it was added as part of https://github.com/sqlparser-rs/sqlparser-rs/pull/260  from @eyalleshem -- perhaps there is some information there. 

Perhaps we can update the test with the new behavior as the point of this PR is to support aliases 🤔

waitingkuo:
@alamb the original test case is to test whether the parser could generate the right error. Since that error is solved in error, i'll update that case

waitingkuo:
@alamb  fixed snowflake's test case, add 2 new test cases in ```sqlparser_common```

src/ast/query.rs
```diff
@@ -375,7 +375,10 @@ pub enum TableFactor {
     ///
     /// The parser may also accept non-standard nesting of bare tables for some
     /// dialects, but the information about such nesting is stripped from AST.
-    NestedJoin(Box<TableWithJoins>),
+    NestedJoin {
+        table_with_joins: Box<TableWithJoins>,
+        alias: Option<TableAlias>,
```
alamb, line 380: 👍

tests/sqlparser_common.rs
```diff
@@ -3446,32 +3446,45 @@ fn parse_joins_using() {
...
+    // full join explicitly
     assert_eq!(
         only(&verified_only_select("SELECT * FROM t1 NATURAL FULL JOIN t2").from).joins,
-        vec![natural_join(JoinOperator::FullOuter)]
+        vec![natural_join(JoinOperator::FullOuter, None)]
+    );
+
+    // natural join with alias
+    assert_eq!(
+        only(&verified_only_select("SELECT * FROM t1 NATURAL JOIN t2 AS t3").from).joins,
```
alamb, line 3486: this is an interesting one -- it adds the alias to the entire `NaturalJoin` rather than aliasing `t2 to t3` on one side of the input. 

Is that what other implementations do?

tests/sqlparser_common.rs
```diff
@@ -3519,6 +3532,15 @@ fn parse_join_nesting() {
         from.joins,
         vec![join(nest!(nest!(nest!(table("b"), table("c")))))]
     );
+
+    let sql = "SELECT * FROM (a NATURAL JOIN b) AS c";
```
alamb, line 3536: this makes sense to me 👍

tests/sqlparser_snowflake.rs
```diff
@@ -131,10 +131,9 @@ fn test_single_table_in_parenthesis_with_alias() {
         "SELECT * FROM (a AS alias1 NATURAL JOIN b AS c)",
     );
 
-    let res = snowflake_and_generic().parse_sql_statements("SELECT * FROM (a NATURAL JOIN b) c");
-    assert_eq!(
-        ParserError::ParserError("Expected end of statement, found: c".to_string()),
-        res.unwrap_err()
+    snowflake_and_generic().one_statement_parses_to(
```
alamb, line 134: nice

tests/test_utils/mod.rs
```diff
@@ -26,9 +26,19 @@ pub use sqlparser::test_utils::*;
...
+        TableFactor::NestedJoin { table_with_joins: Box::new(TableWithJoins {
             relation: $base,
             joins: vec![$(join($join)),*]
-        }))
+        }), alias: None}
+    };
+}
+
+#[macro_export]
+macro_rules! nest_with_alias {
```
alamb, line 37: Since this macro is only invoked in a single place I can see, what do you think about inlining it instead?

alamb:
Thanks @waitingkuo  -- this is looking good

tests/sqlparser_common.rs
waitingkuo: postgresql has this

```bash
willy=# select * from test1 natural join test2 as test3;
 c1 | c0 | c2 | c3 
----+----+----+----
(0 rows)

willy=# select test3.c1 from test1 natural join test2 as test3;
 c1 
----
(0 rows)
```

waitingkuo: it alias `t2 to t3` but not alias to the entire `NaturalJoin`.

waitingkuo: perhaps i should delete this test? i thought at we need to protect the case that it incorrectly alias the whole natural join to t3

alamb: I think it is a good test and that we should leave it in

alamb (approved):
Thanks again @waitingkuo !

tests/test_utils/mod.rs
waitingkuo: done

PR #558, opened 2022-08-08
feat: Support SET NAMES literal [COLLATE literal]

ovr:
Hello!

https://dev.mysql.com/doc/refman/8.0/en/set-names.html

```
SET NAMES {'charset_name'
    [COLLATE 'collation_name'] | DEFAULT}
```

Thanks


alamb (approved):
Awesome -- thank you @ovr

PR #561, opened 2022-08-09
feat: Parse special keywords as functions (current_user, user, etc)

ovr:
Hello!

> Note: current_catalog, current_schema, current_user, session_user, and user have special syntactic status in SQL: they must be called without trailing parentheses. (In PostgreSQL, parentheses can optionally be used with current_schema, but not with the others.)

https://www.postgresql.org/docs/9.1/functions-info.html

Thanks

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2292,20 +2292,27 @@ pub struct Function {
     pub over: Option<WindowSpec>,
     // aggregate functions may specify eg `COUNT(DISTINCT x)`
     pub distinct: bool,
+    pub special: bool,
 }
 
 impl fmt::Display for Function {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(
-            f,
-            "{}({}{})",
-            self.name,
-            if self.distinct { "DISTINCT " } else { "" },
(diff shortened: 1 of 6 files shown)


src/ast/mod.rs
```diff
@@ -2243,20 +2243,27 @@ pub struct Function {
     pub over: Option<WindowSpec>,
     // aggregate functions may specify eg `COUNT(DISTINCT x)`
     pub distinct: bool,
+    pub special: bool,
```
alamb, line 2246: I wonder why a `special` field is needed ? Wouldn't it be enough to match on the `name`? 

If we do need a `special` field, can you please add a `///` comment explaining what it means and what it is used for?

alamb (commented):
Thank you @ovr  -- looks good to me except for a question about the `special` field

alamb:

src/ast/mod.rs
ovr: I use this field to skip formating `(arguments)`.

image

Right now, Dialect is not available in formating, because we use `fmt::Display` for implementing `to_string()` method.

alamb (approved):

alamb:
Thanks @ovr

PR #565, opened 2022-08-09
Support `USE db`

MazterQyou:
This PR adds support for MySQL-specific `USE db` statement, as well as a related test.


alamb (approved):
LGTM -- thanks @MazterQyou

PR #566, opened 2022-08-10
Support PostgreSQL array subquery constructor

MazterQyou:
This PR adds support for PostgreSQL array subquery constructor (e.g. `SELECT ARRAY(SELECT 1 UNION SELECT 2)`), as well as a related test.

### DIFF
--- a/tests/sqlparser_clickhouse.rs
+++ b/tests/sqlparser_clickhouse.rs
@@ -122,6 +122,25 @@ fn parse_array_expr() {
 }
 
 #[test]
+fn parse_array_fn() {
+    let sql = "SELECT array(x1, x2) FROM foo";
+    let select = clickhouse().verified_only_select(sql);
+    assert_eq!(
+        &Expr::Function(Function {
+            name: ObjectName(vec![Ident::new("array")]),
+            args: vec![
+                FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Identifier(Ident::new("x1")))),
+                FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Identifier(Ident::new("x2")))),
+            ],
+            over: None,
+            distinct: false,
+            special: false,
+        }),
+        expr_from_projection(only(&select.projection))
+    );
+}
+
+#[test]
 fn parse_kill() {
     let stmt = clickhouse().verified_stmt("KILL MUTATION 5");
     assert_eq!(
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -2597,7 +2597,7 @@ fn parse_bad_constraint() {
 
 #[test]
 fn parse_scalar_function_in_projection() {
-    let names = vec!["sqrt", "array", "foo"];
+    let names = vec!["sqrt", "foo"];
 
     for function_name in names {
         // like SELECT sqrt(id) FROM foo
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -376,6 +376,8 @@ pub enum Expr {
     /// A parenthesized subquery `(SELECT ...)`, used in expression like
     /// `SELECT (subquery) AS x` or `WHERE (subquery) = x`
     Subquery(Box<Query>),
+    /// An array subquery constructor, e.g. `SELECT ARRAY(SELECT 1 UNION SELECT 2)`
+    ArraySubquery(Box<Query>),
     /// The `LISTAGG` function `SELECT LISTAGG(...) WITHIN GROUP (ORDER BY ...)`
     ListAgg(ListAgg),
     /// The `GROUPING SETS` expr.
@@ -573,6 +575,7 @@ impl fmt::Display for Expr {
                 subquery
             ),
             Expr::Subquery(s) => write!(f, "({})", s),
+            Expr::ArraySubquery(s) => write!(f, "ARRAY({})", s),
             Expr::ListAgg(listagg) => write!(f, "{}", listagg),
             Expr::GroupingSets(sets) => {
                 write!(f, "GROUPING SETS (")?;
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -449,10 +449,17 @@ impl<'a> Parser<'a> {
                 Keyword::TRIM => self.parse_trim_expr(),
                 Keyword::INTERVAL => self.parse_literal_interval(),
                 Keyword::LISTAGG => self.parse_listagg_expr(),
-                // Treat ARRAY[1,2,3] as an array [1,2,3], otherwise try as function call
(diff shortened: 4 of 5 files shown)


tests/sqlparser_common.rs
```diff
@@ -2518,7 +2518,7 @@ fn parse_bad_constraint() {
 
 #[test]
 fn parse_scalar_function_in_projection() {
-    let names = vec!["sqrt", "array", "foo"];
```
MazterQyou, line 2521: This may be a controversial change for this test. PostgreSQL doesn't have a function-based array constructor (`array(1, 2, 3)`), nor can I find information on other databases using such a syntax. Some, like BigQuery, use the `ARRAY(subquery)` syntax implemented there. It'd be great to hear some thoughts on this.

tests/sqlparser_common.rs
```diff
@@ -2518,7 +2518,7 @@ fn parse_bad_constraint() {
...
+    let names = vec!["sqrt", "foo"];
```
MazterQyou, line 2521: This may be a controversial change for this test. PostgreSQL doesn't have a function-based array constructor (`array(1, 2, 3)`), nor can I find information on other databases using such a syntax. Some, like BigQuery, use the `ARRAY(subquery)` syntax implemented there. It'd be great to hear some thoughts on this.

alamb: This test was added in https://github.com/sqlparser-rs/sqlparser-rs/pull/432

Can you review the description on that PR and let me know what you think?

alamb: Interestingly, postgres doesn't allow the syntax used in the DataFusion test (and referred to in https://github.com/sqlparser-rs/sqlparser-rs/pull/432)

```sql
alamb=# SELECT array(c1, cast(c2 as varchar)) FROM test;
ERROR:  syntax error at or near "c1"
LINE 1: SELECT array(c1, cast(c2 as varchar)) FROM test;
                     ^
```

https://github.com/apache/arrow-datafusion/blame/834924f274d9e04444030b8eb3e07e1035fcd3cf/datafusion/core/tests/sql/functions.rs#L134 doesn't really help me figure out from who the test originated from as it was moved (maybe some more poking around could find the original author)

@ovr  do you remember any of the previous history here?

🤔

alamb (commented):
Thanks @MazterQyou  -- the code looks great but as you have identified I am not sure about the `array` test change. I left some comments

tests/sqlparser_common.rs
MazterQyou: I poked around and found this commit: https://github.com/apache/arrow-datafusion/commit/98b710a32361196f1e4d13251a1ef34a93a8211a

alamb: > I poked around and found this commit: https://github.com/apache/arrow-datafusion/commit/98b710a32361196f1e4d13251a1ef34a93a8211a

Nice -- that came in from https://github.com/apache/arrow/pull/8102 by @jorgecarleitao and approved by @andygrove 

Can either if you remember if/how using `array` as a function (rather than the more specialized SQL support offered by postgres) was important? We are triyng to sort out how important backwards compatibility is in this case

alamb: Maybe we can rename the datafusion function `make_array` or something else that doesn't conflict with the sql keyword 🤔

jorgecarleitao: If I remember correctly, we based it on https://spark.apache.org/docs/latest/api/sql/index.html#array , since at that time we had not aligned it with postgres.

alamb: I filed https://github.com/apache/arrow-datafusion/issues/3115 to track

It turns out that `array[c1, c2]` is the proper postgres syntax for the `array(c1, c2)` in spark

Ideally we would allow `array` as a function name in dialects that support it (apparently Spark does). But I don't have the time now to do an analysis of all dialects and if they support that as a function name. How about we merge this PR as is and if someone needs to support `array` as a function name we'll figure out what dialect they are using and support it

MazterQyou: I support this as long as DataFusion team's okay with that. I haven't been able to find a dialect with built-in `array` function, but at least two of them with built-in array subquery constructor.

alamb (approved):

tests/sqlparser_common.rs
MazterQyou: @alamb do you feel it's a good idea to remove the `GenericDialiect | PostgreSqlDialect` check so later someone can make negated checks for dialects that shouldn't go the subquery route?

alamb: >  I haven't been able to find a dialect with built-in array function, but at least two of them with built-in array subquery constructor.

Here is the spark reference docs: https://spark.apache.org/docs/2.4.1/api/sql/index.html#array (though  perhaps you mean there is no `SparkDialect` in this crate)?

> do you feel it's a good idea to remove the GenericDialiect | PostgreSqlDialect check now so later someone can make negated checks for dialects that shouldn't go the subquery route?

Sounds reasonable to me -- can you also possibly link one or two examples from other dialects that support the array subquery constructor?

MazterQyou: PostgreSQL array subquery constructor (in fact, all of array constructors): https://www.postgresql.org/docs/current/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS

I haven't been able to find info on other **dialects** present in the crate to support array subquery, but at least CockroachDB and BigQuery support array subquery.

However while searching around I found that ClickHouse (which we do have a dialect for) actually has array function constructor built in: https://clickhouse.com/docs/en/sql-reference/functions/array-functions/#arrayx1--operator-x1-

So we can either make a decision for leaving this on for specific dialects (`GenericDialect | PostgreSqlDialect` for instance) or negate this to support `ARRAY(x1, x2, …)` for dialects known to support array function constructor (currently `ClickHouseDialect` only).

alamb: Let's negate it for `ClickHouseDialect` and call it a day

MazterQyou:
@alamb updated the if clause with dialects and added a test for ClickHouse's array function (renaming the file in the process due to a typo 😶). Please re-review

tests/sqlparser_clickhouse.rs
```diff
@@ -121,6 +121,25 @@ fn parse_array_expr() {
     )
 }
 
+#[test]
```
alamb, line 124: 👌  Very nice @MazterQyou

alamb (commented):
Looks great -- thanks for sticking with it @MazterQyou

alamb (approved):

PR #567, opened 2022-08-10
Support expressions in `LIMIT`/`OFFSET`

MazterQyou:
`LIMIT` and `OFFSET` allow usage of complex expressions, such as `1+2`, or placeholders for prepared statements, whereas sqlparser only accepts numbers. This PR adds support for any kind of expression in `LIMIT` and `OFFSET`, and adds related tests.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4402,13 +4402,13 @@ impl<'a> Parser<'a> {
         if self.parse_keyword(Keyword::ALL) {
             Ok(None)
         } else {
-            Ok(Some(Expr::Value(self.parse_number_value()?)))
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -4402,13 +4402,13 @@ impl<'a> Parser<'a> {
         if self.parse_keyword(Keyword::ALL) {
             Ok(None)
         } else {
-            Ok(Some(Expr::Value(self.parse_number_value()?)))
+            Ok(Some(self.parse_expr()?))
```
alamb, line 4405: 👍

alamb (approved):
Thank you @MazterQyou  -- we are on a roll today!

PR #569, opened 2022-08-10
Support for `SIMILAR TO` syntax, change `Like` and `ILike` to `Expr` variants, allow escape char for like/ilike

ayushdg:
This pr adds support for the `similar to` sql operator syntax and also adds support to provide escape char to the `[Not] like/ilike` operators.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1,5 +1,3 @@
-// SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1,3 +1,5 @@
+// SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -268,6 +270,27 @@ pub enum Expr {
         op: BinaryOperator,
         right: Box<Expr>,
     },
+    /// LIKE
+    Like {
+        negated: bool,
+        expr: Box<Expr>,
+        pattern: Box<Value>,
+        escape_char: Option<char>,
+    },
+    /// ILIKE (case-insensitive LIKE)
+    ILike {
+        negated: bool,
+        expr: Box<Expr>,
+        pattern: Box<Value>,
+        escape_char: Option<char>,
+    },
+    /// SIMILAR TO regex
+    SimilarTo {
+        negated: bool,
+        expr: Box<Expr>,
+        pattern: Box<Value>,
+        escape_char: Option<char>,
+    },
     /// Any operation e.g. `1 ANY (1)` or `foo > ANY(bar)`, It will be wrapped in the right side of BinaryExpr
     AnyOp(Box<Expr>),
     /// ALL operation e.g. `1 ALL (1)` or `foo > ALL(bar)`, It will be wrapped in the right side of BinaryExpr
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -1,3 +1,5 @@
+// SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
alamb, line 1: I am not sure what this copyright line means -- if it was an intentional inclusion, can you please explain the rationale for it and if it was not intentional remove it

src/ast/mod.rs
```diff
@@ -438,6 +461,72 @@ impl fmt::Display for Expr {
                 high
             ),
             Expr::BinaryOp { left, op, right } => write!(f, "{} {} {}", left, op, right),
+            Expr::Like {
```
alamb, line 464: Seems to match https://www.postgresql.org/docs/current/functions-matching.html so 👍

src/ast/operator.rs
```diff
@@ -1,3 +1,5 @@
+// SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
alamb, line 1: same copyright notice question here and in several other files of this PR below which I will not explicitly point out

src/ast/operator.rs
```diff
@@ -76,10 +78,6 @@ pub enum BinaryOperator {
     And,
     Or,
     Xor,
-    Like,
```
alamb, line 79: 👍

tests/sqlparser_common.rs
```diff
@@ -889,14 +892,11 @@ fn parse_like() {
...
         assert_eq!(
-            Expr::BinaryOp {
-                left: Box::new(Expr::Identifier(Ident::new("name"))),
-                op: if negated {
-                    BinaryOperator::NotLike
-                } else {
-                    BinaryOperator::Like
-                },
-                right: Box::new(Expr::Value(Value::SingleQuotedString("%a".to_string()))),
+            Expr::Like {
```
alamb, line 895: Could you please also extend the `parse_like` and `parse_ilike` tests to include coverage of `escape_char` for both `LIKE` and `ILIKE`?

tests/sqlparser_common.rs
```diff
@@ -954,14 +948,50 @@ fn parse_ilike() {
...
+            Expr::SimilarTo {
+                expr: Box::new(Expr::Identifier(Ident::new("name"))),
+                negated,
+                pattern: Box::new(Value::SingleQuotedString("%a".to_string())),
+                escape_char: Some('\\')
+            },
+            select.selection.unwrap()
+        );
+
+        // This statement tests that LIKE and NOT LIKE have the same precedence.
```
alamb, line 982: This comment appears to be out of date

tests/sqlparser_common.rs
```diff
@@ -954,14 +948,50 @@ fn parse_ilike() {
...
+        // This was previously mishandled (#81).
+        let sql = &format!(
+            "SELECT * FROM customers WHERE name {}SIMILAR TO '%a' ESCAPE '\\' IS NULL",
+            if negated { "NOT " } else { "" }
+        );
+        let select = verified_only_select(sql);
```
alamb, line 988: Can you also please add a test case of `SIMILAR` without an `ESCAPE` clause? Thanks!

alamb (commented):
Thanks @ayushdg -- this code looks great. Thank you for the contribution.

I think the copyright comments need to be addressed and a few more test cases added, but otherwise this PR is looking ready to go!

alamb:

src/ast/mod.rs
ayushdg: This was an intentional inclusion, primarily marking recognition of ownership for code, due to my affiliation with the organization I work for (Nvidia). 

If this open source project doesn't use/accept such copyright licenses for pr's I'm happy to go ahead and remove it.

alamb: > If this open source project doesn't use/accept such copyright licenses for pr's I'm happy to go ahead and remove it.

Please remove it. The code in this repository is licensed with the Apache Software License 2.0 and by contributing you are also agreeing to license your contribution under the same terms. 

https://github.com/sqlparser-rs/sqlparser-rs/blob/main/LICENSE.TXT

If this is unacceptable we will not be able to accept this contribution

alamb: I am sorry if this was not clear -- I have added some clarification in the main project README: https://github.com/sqlparser-rs/sqlparser-rs/pull/570

ayushdg: Thanks for the clarification. This makes sense to me, I'll go ahead and remove the copyright license from these files.

alamb (approved):
LGTM -- thanks again @ayushdg !

alamb:
I believe a new version of rust has been released so we need to update lint on master (unrelated to this change)

PR #572, opened 2022-08-12
Box `Query` in `Cte`

MazterQyou:
`Cte` is the only struct that has `Query` unboxed; all other instances are `Box`ed. I propose `Box`ing `Query` in `Cte` for the sake of interoperability and keeping things similar, thus the PR contains the proposed change.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -271,7 +271,7 @@ impl fmt::Display for With {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Cte {
     pub alias: TableAlias,
-    pub query: Query,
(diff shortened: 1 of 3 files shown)


alamb (commented):
This change is fine with me, however, there is already some discussion about un `Box`ing the `Query` in `Statement` 🤔 

https://github.com/sqlparser-rs/sqlparser-rs/pull/556

cc @ever0de and @andygrove

alamb:
The conclusion from https://github.com/sqlparser-rs/sqlparser-rs/pull/556 is that `Box`ing is the pattern and we will stick to it.

alamb (approved):

PR #573, opened 2022-08-12
Support `TRIM` from with optional expr and FROM expr

### LINKED ISSUE #568: Trim syntax without trimWhere and From not supported.
Postgresql syntax for trim `trim ( [ LEADING | TRAILING | BOTH ] [ characters text ] FROM string text ) → text` allows the pattern of
`trim(chars from string)`.

sqlparser currently fails with this combo of no leading/trailing/both flag and the `from` keyword specified.

Happy to work on a pr for this fix

ayushdg:
Fixes #568 
The objective of this pr is to allow specifying the chars from expr syntax for the trim expression even without specifying a trim where flag. 

- [x] Update trim expr members, not tying the trimWhere option to the trim_what option
- [x] Update parsing logic to handle cases with no trim flag but trim_char
- [x] Add tests

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -337,7 +337,8 @@ pub enum Expr {
     Trim {
         expr: Box<Expr>,
         // ([BOTH | LEADING | TRAILING], <expr>)
-        trim_where: Option<(TrimWhereField, Box<Expr>)>,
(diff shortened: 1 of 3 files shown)


alamb (approved):
Looks good to me -- thanks @ayushdg

PR #574, opened 2022-08-16
feat: Support generic expressions in SET statement

ovr:
Hello!

MySQL allows to use any kind of expression in `SetVariable` statement, for example:

```
SET sql_mode = CONCAT(@@sql_mode, ',STRICT_TRANS_TABLES')
```

Thanks

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -580,7 +580,7 @@ fn parse_select_count_wildcard() {
 
 #[test]
 fn parse_select_count_distinct() {
-    let sql = "SELECT COUNT(DISTINCT + x) FROM customer";
+    let sql = "SELECT COUNT(DISTINCT +x) FROM customer";
     let select = verified_only_select(sql);
     assert_eq!(
         &Expr::Function(Function {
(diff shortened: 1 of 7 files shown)


tests/sqlparser_common.rs
```diff
@@ -580,7 +580,7 @@ fn parse_select_count_wildcard() {
 
 #[test]
 fn parse_select_count_distinct() {
-    let sql = "SELECT COUNT(DISTINCT + x) FROM customer";
+    let sql = "SELECT COUNT(DISTINCT +x) FROM customer";
```
alamb, line 583: Ah, I was confused about this for a while, but then I see that `+x` is actually a unary op, so it makes sense to display it as `+x` rather than `+ x` 👍

tests/sqlparser_common.rs
```diff
@@ -754,7 +754,7 @@ fn parse_compound_expr_2() {
 #[test]
 fn parse_unary_math() {
     use self::Expr::*;
-    let sql = "- a + - b";
+    let sql = "-a + -b";
```
alamb, line 757: ❤️

alamb (approved):
This looks great -- thanks @ovr

PR #579, opened 2022-08-17
Parse LIKE patterns as Expr not Value

andygrove:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/578

This fixes a regression in parsing LIKE patterns that was introduced in https://github.com/sqlparser-rs/sqlparser-rs/pull/569.

I also made some small improvements in the precedence code by extracting more hard-coded values into consts so we can more easily see the relationship between different operators. I added a link to the Postgres precedence docs for reference as well.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -272,21 +272,21 @@ pub enum Expr {
     Like {
         negated: bool,
         expr: Box<Expr>,
-        pattern: Box<Value>,
+        pattern: Box<Expr>,
         escape_char: Option<char>,
     },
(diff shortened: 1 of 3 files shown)


andygrove:
@ayushdg could you take a look?

ayushdg (approved):
Thanks for catching the regressions and updating.

andygrove:
I'm going to go ahead and merge this based on the approval from @ayushdg since this is blocking a number of issues in DataFusion.

alamb:
Thanks @andygrove

PR #581, opened 2022-08-17
Add ability for dialects to override prefix, infix, and statement parsing

andygrove:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/580

This change adds some new functions to `Dialect`:

- parse_prefix
- parse_infix
- parse_statement

The parser will invoke these methods to see if the dialect wants to override the parsing of these items; otherwise, it will parse them itself. This is a much cleaner way to add dialect-specific parsing, IMO and will allow us to start removing all the `if dialect_of` hacks from the main parser.

I moved the Postgres parsing of comments and the SqlLite parsing of REPLACE into their respective dialects to demonstrate how this works.


src/parser.rs
```diff
@@ -195,13 +200,6 @@ impl<'a> Parser<'a> {
                 Keyword::EXECUTE => Ok(self.parse_execute()?),
                 Keyword::PREPARE => Ok(self.parse_prepare()?),
                 Keyword::MERGE => Ok(self.parse_merge()?),
-                Keyword::REPLACE if dialect_of!(self is SQLiteDialect ) => {
-                    self.prev_token();
-                    Ok(self.parse_insert()?)
-                }
-                Keyword::COMMENT if dialect_of!(self is PostgreSqlDialect) => {
-                    Ok(self.parse_comment()?)
-                }
```
andygrove, line 204: This dialect-specific code has now moved to the appropriate dialects.

src/parser.rs
```diff
@@ -4926,4 +4910,124 @@ mod tests {
...
+                    None
+                }
+            }
+        }
+
+        let dialect = MyDialect {};
+        let sql = "SELECT 1 + 2";
+        let ast = Parser::parse_sql(&dialect, sql)?;
+        let query = &ast[0];
+        assert_eq!("SELECT NULL + 2", &format!("{}", query));
```
alamb, line 4941: 😆

src/parser.rs
```diff
@@ -4926,4 +4910,124 @@ mod tests {
             assert_eq!(ast.to_string(), sql.to_string());
         });
     }
+
+    #[test]
+    fn custom_prefix_parser() -> Result<(), ParserError> {
```
alamb, line 4915: I recommend putting these tests into `tests` somewhere so that we can be sure everything needed is exposed `pub`lically.

Perhaps `tests/sqlparser_custom_dialects.rs`?

alamb (approved):
I think it looks like a great idea -- thank you @andygrove 

I had a small comment on tests but otherwise 👍 🚀

src/parser.rs
andygrove: Good idea. I have moved them.

PR #583, opened 2022-08-17
Add `IS NOT TRUE`, `IS NOT FALSE`, and `IS [NOT] UNKNOWN`

### LINKED ISSUE #582: Support `IS NOT TRUE`, `IS NOT FALSE`, and `IS [NOT] UNKNOWN`
Hi, I would like to be able to parse something like:
```
SELECT
    b IS TRUE AS t,
    b IS FALSE AS f,
    b IS NOT TRUE AS nt,
    b IS NOT FALSE AS nf,
    b IS UNKNOWN AS u,
    b IS NOT UNKNOWN AS nu
FROM data
```
But we are unable to parse `IS NOT TRUE`, `IS NOT FALSE`, and `IS [NOT] UNKNOWN`.

sarahyurick:
Closes #582

Previously, sqlparser did not support "is not true," "is not false," "is unknown," and "is not unknown," so I have added them here.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -227,12 +227,20 @@ pub enum Expr {
     CompositeAccess { expr: Box<Expr>, key: Ident },
     /// `IS FALSE` operator
     IsFalse(Box<Expr>),
+    /// `IS NOT FALSE` operator
+    IsNotFalse(Box<Expr>),
     /// `IS TRUE` operator
     IsTrue(Box<Expr>),
+    /// `IS NOT TRUE` operator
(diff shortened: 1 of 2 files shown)


andygrove (approved):
Thanks @sarahyurick. This seems to follow the pattern we use for other `IS` expressions such as `IsNull` / `IsNotNull`, so LGTM.

andygrove:
Link to Postgres documentation on these expressions for anyone interested:

https://www.postgresql.org/docs/current/functions-comparison.html

PR #586, opened 2022-08-19
fix: mysql ON UPDATE not taking an expression

### LINKED ISSUE #543: MySQL - On Update not taking expression
```sql 
CREATE TABLE foo (`modification_time` DATETIME ON UPDATE CURRENT_TIMESTAMP())
```
does not work

Sibz:
Attempt to resolve the #543, the PR resolves the issue however I was not sure on the test construction so there is currently a pending todo before this can be merged. As per my comments on #543 any assistance/guidance would be appreciated. 

Also not sure the extra enum variant is the best way to do this, but wasn't able to pull the tokens and expr from `DialectSpecific` variant.

closes #543

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3254,9 +3254,9 @@ impl<'a> Parser<'a> {
         } else if self.parse_keywords(&[Keyword::ON, Keyword::UPDATE])
             && dialect_of!(self is MySqlDialect)
         {
-            Ok(Some(ColumnOption::DialectSpecific(vec![
-                Token::make_keyword("ON UPDATE"),
-            ])))
+            let expr = self.parse_expr()?;
+            Self::check_on_update_expr_is_valid(&expr)?;
+            Ok(Some(ColumnOption::OnUpdate(expr)))
         } else {
             Ok(None)
         }
@@ -6317,6 +6317,25 @@ impl<'a> Parser<'a> {
     /// The index of the first unprocessed token.
     pub fn index(&self) -> usize {
         self.index
+    }
+
+    fn check_on_update_expr_is_valid(expr: &Expr) -> Result<(), ParserError> {
+        let valid_object_names: [ObjectName; 4] = [
+            ObjectName(vec!["CURRENT_TIMESTAMP".into()]),
+            ObjectName(vec!["LOCALTIME".into()]),
+            ObjectName(vec!["LOCALTIMESTAMP".into()]),
+            ObjectName(vec!["NOW".into()]),
+        ];
+
+        if let Expr::Function(f) = expr {
+            if valid_object_names.contains(&f.name) {
+                return Ok(());
+            }
+        }
+        Err(ParserError::ParserError(format!(
+            "Expected one of '{}' after ON UPDATE in column definition",
+            valid_object_names.map(|x| x.to_string()).join("', '")
+        )))
     }
 }
 
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -497,9 +497,8 @@ impl<'a> Parser<'a> {
                 | Keyword::CURRENT_TIME
                 | Keyword::CURRENT_DATE
                 | Keyword::LOCALTIME
-                | Keyword::LOCALTIMESTAMP => {
-                    self.parse_time_functions(ObjectName(vec![w.to_ident()]))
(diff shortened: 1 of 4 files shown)


alamb:
I think the enum idea is a good one

alamb:
FYI, I am planning to make a new sqlparser-rs release in the next day or two. This PR has some outstanding comments and appears to have have stalled

alamb:

Sibz:
Commit updated:
Added verification the expr is one of 'CURRENT_TIMESTAMP', 'LOCALTIME', 'LOCALTIMESTAMP', 'NOW'.
Fixed test
Added test for invalid expr parser error

alamb (approved):
Thank you @Sibz

Sibz:
@alamb Just had a quick review and found an improvement could be made, so added a re

src/parser.rs
```diff
@@ -6320,21 +6320,22 @@ impl<'a> Parser<'a> {
...
+        const VALID_FN_NAMES: [&'static str; 4] = [
+            keywords::CURRENT_TIMESTAMP,
+            keywords::LOCALTIME,
+            keywords::LOCALTIMESTAMP,
+            "NOW",
         ];
 
         if let Expr::Function(f) = expr {
-            if valid_object_names.contains(&f.name) {
+            if VALID_FN_NAMES.contains(&f.name.to_string().as_str()) {
```
alamb, line 6331: I think this will now allocate each time to create a `String`

However, that does seem better than 4 allocations :)

Sibz: Yeah I figured 1 was better than four ... but didn't think to use the `Ident` in the `ObjectName` as there may be multiple .. however for this case we are only testing for functions that must only have one `Ident` for the `ObjectName`, other cases would be invalid anyway.

Updated the refactor commit and force pushed.

Sibz:
Added NOW keyword, included that in the parse time functions match arm
Fix check name to be case insensitive
Added test for lowercase `now()` fn use.

src/parser.rs
```diff
@@ -6318,6 +6317,31 @@ impl<'a> Parser<'a> {
     pub fn index(&self) -> usize {
         self.index
     }
+
+    fn check_on_update_expr_is_valid(expr: &Expr) -> Result<(), ParserError> {
```
alamb, line 6321: my preference is to leave this type of "semantic" check downstream -- so I would prefer that sqlparser accepts any expression in this place, and then downstream crates can enforce limits like the expr must be one of the specified time functions. 

@AugustoFKL  and I tried to explain this difference in the readme recently

https://github.com/sqlparser-rs/sqlparser-rs#extensible-sql-lexer-and-parser-for-rust

> This crate provides only a syntax parser, and tries to avoid applying
> any SQL semantics, and accepts queries that specific databases would
> reject, even when using that Database's specific `Dialect`. For
> example, `CREATE TABLE(x int, x int)` is accepted by this crate, even
> though most SQL engines will reject this statement due to the repeated
> column name `x`.

> This crate avoids semantic analysis because it varies drastically
> between dialects and implementations. If you want to do semantic
> analysis, feel free to use this project as a base

src/parser.rs
```diff
@@ -3254,9 +3253,9 @@ impl<'a> Parser<'a> {
         } else if self.parse_keywords(&[Keyword::ON, Keyword::UPDATE])
             && dialect_of!(self is MySqlDialect)
         {
-            Ok(Some(ColumnOption::DialectSpecific(vec![
-                Token::make_keyword("ON UPDATE"),
-            ])))
+            let expr = self.parse_expr()?;
+            Self::check_on_update_expr_is_valid(&expr)?;
```
alamb, line 3257: I recommend just accepting `expr` here

alamb (commented):
Thank you @Sibz  -- this is looking very close.

Can you please remove the validation code? If so I think this PR is good to go.

alamb:
Given https://github.com/sqlparser-rs/sqlparser-rs/pull/602  has something similar, perhaps I can try and push this one through

alamb:
I believe this PR is no longer necessary after https://github.com/sqlparser-rs/sqlparser-rs/pull/685. Can you please confirm @Sibz ?

PR #587, opened 2022-08-20
Support for table column type identity with sequence options, create sequence, drop sequnece.

sam-mmm:
**Support added for** 

- Table column def type   `GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ] |` (https://www.postgresql.org/docs/current/sql-createtable.html)
- Create sequence `CREATE [ TEMPORARY | TEMP ] SEQUENCE [ IF NOT EXISTS ] name
    [ AS data_type ]
    [ INCREMENT [ BY ] increment ]
    [ MINVALUE minvalue | NO MINVALUE ] [ MAXVALUE maxvalue | NO MAXVALUE ]
    [ START [ WITH ] start ] [ CACHE cache ] [ [ NO ] CYCLE ]
    [ OWNED BY { table_name.column_name | NONE } ]` (https://www.postgresql.org/docs/current/sql-createsequence.html)
- Drop sequence `DROP SEQUENCE [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ] ` (https://www.postgresql.org/docs/current/sql-dropsequence.html)

Only tested in regard to PostgreSQL documentation

Test Cases:

- sqlparser_postgres::parse_create_table_generated_always_as_identity
- sqlparser_postgres::parse_create_sequence
- sqlparser_postgres::parse_drop_sequence


andygrove:
Thanks @sam-mmm I will try and make time to review this next week.

sam-mmm:
Fixed issues with check list and added additional test scenarios

src/ast/ddl.rs
```diff
@@ -377,6 +379,26 @@ pub enum ColumnOption {
...
+    /// https://www.postgresql.org/docs/current/sql-createtable.html
+    Generated {
+        always_or_by_default_or_always_as: AlwaysOrByDefaultOrAlwaysAs,
+        sequence_options: Vec<SequenceOptions>,
+    },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum AlwaysOrByDefaultOrAlwaysAs {
```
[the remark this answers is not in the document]
alamb: I wonder if calling this enum `Generated` would better match the SQL it comes from and be more in line with the conventions of the rest of this crate?

src/ast/mod.rs
```diff
@@ -2175,6 +2191,100 @@ impl fmt::Display for Statement {
...
+                if_not_exists,
+                name,
+                data_type,
+                sequence_options,
+                owned_by,
+            } => {
+                let mut as_type: String = String::from("");
+                if let Some(dt) = data_type.as_ref() {
+                    as_type = " AS ".to_string() + &dt.to_string();
+                }
```
alamb, line 2205: As written this looks like it will do a bunch of string creation and copying that is unnecessary. Perhaps something like this would be less code and more efficient

```suggestion
                let as_type =  if let Some(dt) = data_type.as_ref() {
                    format!(" AS {}", dt);
                } else {
                    "".to_string()
                }
```

src/ast/mod.rs
```diff
@@ -2175,6 +2191,100 @@ impl fmt::Display for Statement {
...
+            }
+        }
+    }
+}
+
+/// Will not add to vec when there are no value since all blocks are optional
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// #[non_exhaustive]
+pub enum SequenceOptions {
```
alamb, line 2230: Can you please add docstring comments to this struct explaining what SQL structure it came from? 

I realize it is obvious in the context of this PR, but it is important to make sure the contents of docs.rs is easy to follow

src/ast/mod.rs
```diff
@@ -2175,6 +2191,100 @@ impl fmt::Display for Statement {
...
+    ///MaxValue(maxvalue, 'NO MAXVALUE')
+    MaxValue(Option<Expr>, Option<bool>),
+    ///StartWith(start, with) 'WITH' is optional in PostgreSQL
+    StartWith(Expr, bool),
+    Cache(Expr),
+    Cycle(bool),
+}
+
+/// https://www.postgresql.org/docs/current/sql-createsequence.html
+/// Will not add to vec when there are no value since all blocks are optional
```
alamb, line 2244: Again, I am not sure what this comment means

src/ast/mod.rs
```diff
@@ -2175,6 +2191,100 @@ impl fmt::Display for Statement {
...
+                )
+            }
+            SequenceOptions::MinValue(minvalue, no) => {
+                if let Some(minvalue) = minvalue.as_ref() {
+                    write!(f, " MINVALUE {minvalue}", minvalue = minvalue)
+                } else if no.is_some() {
+                    write!(f, " NO MINVALUE")
+                } else {
+                    write!(f, "")
+                }
```
alamb, line 2263: You can write this type of code more concisely (and have the compiler verify you covered all the cases) like:

```suggestion
                match minvalue.as_ref() {
                    (Some(minvalue), _) =>  write!(f, " MINVALUE {minvalue}", minvalue = minvalue),
                    (None, Some(_)) => write!(f, " NO MINVALUE")
                    (None, _) => write!(f, " NO MINVALUE")
  
```

src/ast/mod.rs
```diff
@@ -2175,6 +2191,100 @@ impl fmt::Display for Statement {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// #[non_exhaustive]
+pub enum SequenceOptions {
+    ///IncrementBy(increment, by) 'BY' is optional in PostgreSQL
+    IncrementBy(Expr, bool),
+    ///MinValue(minvalue, 'NO MINVALUE')
+    MinValue(Option<Expr>, Option<bool>),
+    ///MaxValue(maxvalue, 'NO MAXVALUE')
+    MaxValue(Option<Expr>, Option<bool>),
```
alamb, line 2236: Instead of encoding this as `Option<Expr>` and an `Option<bool> which can have 4 possible states, only three of which are valid, how about encoding it as an enum:

```rust
enum MinMaxValue {
  // clause is not specified
  Empty,
  // NO MINVALUE/NO MAX VALUE
  None, 
  // MINVALUE <expr> / MAXVALUE <expr>
  Some(Expr)
}
```

tests/sqlparser_postgres.rs
```diff
@@ -14,13 +14,266 @@
...
-#[macro_use]
-mod test_utils;
-use test_utils::*;
+// use simple_logger::SimpleLogger;
 
 use sqlparser::ast::*;
+// use sqlparser::ast::Value::Boolean;
 use sqlparser::dialect::{GenericDialect, PostgreSqlDialect};
 use sqlparser::parser::ParserError;
+use test_utils::*;
```
alamb, line 23: ❤️  very nice tests

tests/sqlparser_postgres.rs
```diff
@@ -1247,75 +1500,12 @@ fn parse_array_index_expr() {
...
             elem: vec![],
-            named: true
+            named: true,
         }),
         expr_from_projection(only(&select.projection)),
     );
 }
 
-#[test]
-fn parse_array_subquery_expr() {
```
alamb, line 1257: Can you please explain why this test was deleted?

tests/sqlparser_postgres.rs
```diff
@@ -1618,52 +1808,6 @@ fn parse_declare() {
     pg_and_generic().verified_stmt("DECLARE \"SQL_CUR0x7fa44801bc00\" BINARY INSENSITIVE SCROLL CURSOR WITH HOLD FOR SELECT * FROM table_name LIMIT 2222");
 }
 
-#[test]
-fn parse_current_functions() {
```
alamb, line 1622: likewise, why is this test deleted?

alamb (commented):
Thank you for the contribution @sam-mmm  -- this is a great start. With a few more doc comments and some cleanup I think this PR will be ready to go

I apologize for the belated review

PR #589, opened 2022-08-23
add with/without time zone 

### LINKED ISSUE #464: Not possible to have timezone in postgres
(as ever, thanks for this excellent library!)

It seems that in postgres, specifying a timezone in a string is not enough for it to parse a timezone — [postgres requires `WITH TIMEZONE`](https://www.postgresql.org/docs/current/datatype-datetime.html): 

> PostgreSQL never examines the content of a literal string before determining its type, and therefore will treat both of the above as timestamp without time zone. To ensure that a literal is treated as timestamp with time zone, give it the correct explicit type:
  ```
  TIMESTAMP WITH TIME ZONE '2004-10-19 10:23:54+02'
  ```

And `sqlparser` seems to discard that term:

https://github.com/sqlparser-rs/sqlparser-rs/blob/8f207db0598139310211e808ea6e5f853d3a3a96/src/parser.rs

So IIUC it's not currently possible to correctly parse a postgres `TIMESTAMP WITH TIME ZONE`

### LINKED ISSUE #588: Add Time Zone support
`With/Without Time Zone` is currently parsed but not deal with.

I'd like to add them and TimestampTz

waitingkuo:
close  #588 
close  #464

add datetype `TimestampTz`

add following into parser

`TIMESTAMP [WITHOUT TIME ZONE]` => `Timestamp`
`TIMESTAMP WITH TIME ZONE` and `TIMESTAMPTZ` => `TimestampTz

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -77,8 +77,10 @@ pub enum DataType {
     Time,
     /// Datetime
     Datetime,
-    /// Timestamp
+    /// Timestamp [Without Time Zone]
     Timestamp,
+    /// Timestamp With Time Zone
+    TimestampTz,
     /// Interval
     Interval,
     /// Regclass used in postgresql serial
@@ -157,6 +159,7 @@ impl fmt::Display for DataType {
             DataType::Time => write!(f, "TIME"),
             DataType::Datetime => write!(f, "DATETIME"),
             DataType::Timestamp => write!(f, "TIMESTAMP"),
+            DataType::TimestampTz => write!(f, "TIMESTAMPTZ"),
             DataType::Interval => write!(f, "INTERVAL"),
             DataType::Regclass => write!(f, "REGCLASS"),
             DataType::Text => write!(f, "TEXT"),
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3119,12 +3119,17 @@ impl<'a> Parser<'a> {
                 Keyword::DATE => Ok(DataType::Date),
                 Keyword::DATETIME => Ok(DataType::Datetime),
                 Keyword::TIMESTAMP => {
-                    // TBD: we throw away "with/without timezone" information
-                    if self.parse_keyword(Keyword::WITH) || self.parse_keyword(Keyword::WITHOUT) {
+                    if self.parse_keyword(Keyword::WITH) {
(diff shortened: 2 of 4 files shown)


src/ast/data_type.rs
```diff
@@ -157,6 +159,7 @@ impl fmt::Display for DataType {
             DataType::Time => write!(f, "TIME"),
             DataType::Datetime => write!(f, "DATETIME"),
             DataType::Timestamp => write!(f, "TIMESTAMP"),
+            DataType::TimestampTz => write!(f, "TIMESTAMPTZ"),
```
waitingkuo, line 162: add `DataType::TimestampTz`

src/parser.rs
```diff
@@ -3119,12 +3119,17 @@ impl<'a> Parser<'a> {
...
                         self.expect_keywords(&[Keyword::TIME, Keyword::ZONE])?;
+                        Ok(DataType::TimestampTz)
+                    } else if self.parse_keyword(Keyword::WITHOUT) {
+                        self.expect_keywords(&[Keyword::TIME, Keyword::ZONE])?;
+                        Ok(DataType::Timestamp)
+                    } else {
+                        Ok(DataType::Timestamp)
                     }
-                    Ok(DataType::Timestamp)
                 }
```
waitingkuo, line 3131: enable `with time zone` and `without time zone`

src/parser.rs
```diff
@@ -3119,12 +3119,17 @@ impl<'a> Parser<'a> {
...
+                Keyword::TIMESTAMPTZ => Ok(DataType::TimestampTz),
```
waitingkuo, line 3132: add TimestampTz into parser

waitingkuo:
@liukun4515 @alamb  this is to add time zone support for sql parser

we need this to solve time-zone related issue in apache/arrow-datafusion#3148

liukun4515:
I need some time to get familiar with the context in the datafusion/arrow-rs. @waitingkuo Maybe be slower to reply

andygrove (approved):
LGTM. Thanks @waitingkuo

AugustoFKL:
@waitingkuo @andygrove `TIMESTAMPTZ` and `TIMESTAMP WITH TIME ZONE` aren't the same. Did a fix PR: https://github.com/sqlparser-rs/sqlparser-rs/pull/641

PR #592, opened 2022-08-25
feat: Support `LOCALTIME` and `LOCALTIMESTAMP` time functions

MazterQyou:
This PR adds support for parsing `LOCALTIME` and `LOCALTIMESTAMP` without parentheses as a time function with no arguments (see https://www.postgresql.org/docs/14/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT). It also adds related tests.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -438,7 +438,11 @@ impl<'a> Parser<'a> {
                         special: true,
                     }))
                 }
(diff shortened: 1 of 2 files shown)


andygrove (approved):
LGTM. Will merge next week if there are no objections.

cc @waitingkuo

PR #594, opened 2022-08-26
Add overlay expr

ayushdg:
Fixes: #593
This pr adds support for the overlay syntax See https://www.postgresql.org/docs/current/functions-string.html.

- [X] Add high level overlay expr
- [X] Add parsing logic for overlay
- [X] Add tests

The pr also adds a few tests cases for `is not true/false` and `is [not] unknown` added in #582


andygrove (approved):
LGTM. Thanks @ayushdg

PR #598, opened 2022-09-01
Support CREATE ROLE and DROP ROLE

blx:
Adds some support for parsing CREATE ROLE and DROP ROLE.

CREATE ROLE varies considerably across dialects:
* implemented `CREATE ROLE [IF NOT EXISTS] role_name, role_name, ...` for all dialects
* supports the vast majority of the optional parameters for Postgres, if Postgres dialect is enabled (like `[NO]SUPERUSER`, `[NO]REPLICATION`, `[NO]BYPASSRLS`, `PASSWORD ...`, etc, etc) - following https://www.postgresql.org/docs/current/sql-createrole.html
* supports the one optional parameter used by MSSQL, following https://docs.microsoft.com/en-us/sql/t-sql/statements/create-role-transact-sql?view=sql-server-ver16

Not yet handled:
* optional parameters for Clickhouse or Oracle
* I haven't looked at other dialects beyond that


src/parser.rs
```diff
@@ -2044,17 +2046,223 @@ impl<'a> Parser<'a> {
...
+                    self.expect_keyword(Keyword::UNTIL)?;
+                    if valid_until.is_some() {
+                        parser_err!("Found multiple VALID UNTIL")
+                    } else {
+                        valid_until = Some(Expr::Value(self.parse_value()?));
+                        Ok(())
+                    }
+                }
+                Keyword::IN => {
+                    if self.parse_keyword(Keyword::ROLE) || self.parse_keyword(Keyword::GROUP) {
```
blx, line 2199: > IN GROUP is an obsolete spelling of IN ROLE.

(postgres docs)

src/parser.rs
```diff
@@ -2044,17 +2046,223 @@ impl<'a> Parser<'a> {
...
+                            parser_err!("Found multiple IN ROLE or IN GROUP")
+                        } else {
+                            in_role = self.parse_comma_separated(Parser::parse_identifier)?;
+                            Ok(())
+                        }
+                    } else {
+                        self.expected("ROLE or GROUP after IN", self.peek_token())
+                    }
+                }
+                Keyword::ROLE | Keyword::USER => {
```
blx, line 2210: > The USER clause is an obsolete spelling of the ROLE clause.

(postgres docs)

src/parser.rs
```diff
@@ -2044,17 +2046,223 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_create_role(&mut self) -> Result<Statement, ParserError> {
+        let if_not_exists = self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
+        let names = self.parse_comma_separated(Parser::parse_object_name)?;
+
+        let _ = self.parse_keyword(Keyword::WITH);
```
alamb, line 2053: As written this ignores errors. I think it should be

```suggestion
        self.parse_keyword(Keyword::WITH)?;
```

src/parser.rs
alamb: I think the typical pattern in `sqlparser` is that we parse the sql as provided into an AST that could recover the same pattern rather than applying semantic meaning

So in this case, can you please change `CreateRole` to have an explicit `group: Option<Ident>` rather than reusing `in_role`? 

The same comment applies to `Role/User`

src/ast/mod.rs
```diff
@@ -1047,6 +1054,27 @@ pub enum Statement {
...
+        login: Option<bool>,
+        inherit: Option<bool>,
+        bypassrls: Option<bool>,
+        password: Option<Password>,
+        superuser: Option<bool>,
+        create_db: Option<bool>,
+        create_role: Option<bool>,
+        replication: Option<bool>,
+        connection_limit: Option<Expr>,
+        valid_until: Option<Expr>,
```
alamb, line 1071: One downside with this implementation is that the order of the grants is not preserved and thus the original SQL can not be reconstructed

What do you think about an alternate encoding like:

```rust
...
role_grants: Vec<RoleGrant>
...
```

And then


```rust
enum RoleGrant {
  Login, 
  Inherit,
  Password(Password)
   ..
}
```

However, I see this PR follows the model of GRANT so I think it is also fine as is

tests/sqlparser_common.rs
```diff
@@ -4831,14 +4883,8 @@ fn parse_grant() {
...
                 assert_eq!(vec![Action::Insert { columns: None }], actions);
-                assert_eq!(
-                    vec!["public"],
-                    schemas.iter().map(ToString::to_string).collect::<Vec<_>>()
-                );
-                assert_eq!(
-                    vec!["browser"],
-                    grantees.iter().map(ToString::to_string).collect::<Vec<_>>()
-                );
+                assert_eq_vec(&["public"], &schemas);
```
alamb, line 4886: thank you -- this is much nicer

alamb (approved):
I think this looks great @blx  -- thank you for the contribution; There is one bug I think but I will provide a follow on PR to fix it.

src/parser.rs
alamb: Sorry -- I misunderstood this code. It actually skips the optional `WITH` clause. Added a PR with a test https://github.com/sqlparser-rs/sqlparser-rs/pull/627

src/parser.rs
alamb: Here is my proposal: https://github.com/sqlparser-rs/sqlparser-rs/pull/628

PR #599, opened 2022-09-01
Implement --quiet option for cli example

blx:
Not sure if there's any interest in this, or if this is the best flag name - but this allows `--quiet` to be passed to the `cli` example target so that it can be composed with other programs in a rudimentary way without writing a custom wrapper around sqlparser-rs.

For example, you could pipe the parsed AST into `jq` or some other JSON-expecting program:
```
$ cargo run --release --features json_example --example cli "$filename" --quiet | jq 'map(keys)[][]' -rc
Drop
CreateTable
Insert
Drop
```

### DIFF
--- a/examples/cli.rs
+++ b/examples/cli.rs
@@ -21,22 +21,31 @@ use sqlparser::dialect::*;
 use sqlparser::parser::Parser;
 
 fn main() {
-    SimpleLogger::new().init().unwrap();
-
     let filename = std::env::args().nth(1).expect(
         r#"
 No arguments provided!
 
 Usage:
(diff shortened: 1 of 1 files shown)


andygrove (commented):
Thanks @blx. I like the idea of this. With more than one arg I think we should use `clap` for argument parsing rather than manually parsing the args though. Could you make that change in this PR?

alamb:

AugustoFKL:
@blx @pawel-big-lebowski is closing this for now because it's almost a year without an update.

@blx, ping me if you work on this code specifically; if so, I will open the PR again :)

PR #600, opened 2022-09-04
[postgres] Support `ABORT` and `END`

iskakaushik:
From https://www.postgresql.org/docs/current/sql-abort.html

> `ABORT` rolls back the current transaction and causes all the updates made by the transaction to be discarded. This command is identical in behavior to the standard SQL command `ROLLBACK`, and is present only for historical reasons.

From https://www.postgresql.org/docs/current/sql-end.html

> `END` commits the current transaction. All changes made by the transaction become visible to others and are guaranteed to be durable if a crash occurs. This command is a PostgreSQL extension that is equivalent to `COMMIT`.


alamb (commented):
Thanks @iskakaushik !

I am sorry for the delayed feedback

alamb:

PR #602, opened 2022-09-05
Fix create statement for mysql

RainJoe:
1. Add ON UPDATE CURRENT_TIMESTAMP support.
2. Add KEY/INDEX support

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -229,6 +229,7 @@ pub enum TableConstraint {
         columns: Vec<Ident>,
         /// Whether this is a `PRIMARY KEY` or just a `UNIQUE` constraint
         is_primary: bool,
+        is_mysql_unique_key: bool,
     },
     /// A referential integrity constraint (`[ CONSTRAINT <name> ] FOREIGN KEY (<columns>)
     /// REFERENCES <foreign_table> (<referred_columns>)
@@ -262,13 +263,30 @@ impl fmt::Display for TableConstraint {
                 name,
                 columns,
                 is_primary,
-            } => write!(
-                f,
-                "{}{} ({})",
-                display_constraint_name(name),
-                if *is_primary { "PRIMARY KEY" } else { "UNIQUE" },
-                display_comma_separated(columns)
-            ),
+                is_mysql_unique_key,
+            } => {
+                if *is_mysql_unique_key {
+                    write!(
+                        f,
+                        "{}{}({})",
+                        if *is_primary {
+                            "PRIMARY KEY"
+                        } else {
+                            "UNIQUE "
+                        },
+                        display_key_name(name, Keyword::KEY),
+                        display_comma_separated(columns)
+                    )
+                } else {
+                    write!(
+                        f,
+                        "{}{} ({})",
+                        display_constraint_name(name),
+                        if *is_primary { "PRIMARY KEY" } else { "UNIQUE" },
+                        display_comma_separated(columns)
+                    )
+                }
+            }
             TableConstraint::Key {
                 name,
                 columns,
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -22,6 +22,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::ast::value::escape_single_quote_string;
 use crate::ast::{display_comma_separated, display_separated, DataType, Expr, Ident, ObjectName};
+use crate::keywords::Keyword;
 use crate::tokenizer::Token;
 
 /// An `ALTER TABLE` (`Statement::AlterTable`) operation
@@ -247,6 +248,11 @@ pub enum TableConstraint {
         name: Option<Ident>,
         expr: Box<Expr>,
     },
+    Key {
+        name: Option<Ident>,
+        columns: Vec<Ident>,
+        keyword: Keyword,
+    },
 }
 
 impl fmt::Display for TableConstraint {
@@ -261,6 +267,16 @@ impl fmt::Display for TableConstraint {
                 "{}{} ({})",
                 display_constraint_name(name),
                 if *is_primary { "PRIMARY KEY" } else { "UNIQUE" },
+                display_comma_separated(columns)
+            ),
+            TableConstraint::Key {
+                name,
+                columns,
+                keyword,
+            } => write!(
+                f,
+                "{}({})",
+                display_key_name(name, *keyword),
                 display_comma_separated(columns)
             ),
             TableConstraint::ForeignKey {
@@ -426,6 +442,27 @@ fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
         }
     }
     ConstraintName(name)
+}
+
+fn display_key_name(name: &'_ Option<Ident>, keyword: Keyword) -> impl fmt::Display + '_ {
+    struct KeyName<'a>(&'a Option<Ident>, Keyword);
+    impl<'a> fmt::Display for KeyName<'a> {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            if let Some(name) = self.0 {
+                match self.1 {
+                    Keyword::KEY => {
+                        write!(f, "KEY {} ", name)?;
+                    }
+                    Keyword::INDEX => {
+                        write!(f, "INDEX {} ", name)?;
+                    }
+                    _ => unreachable!(),
+                }
+            }
+            Ok(())
+        }
+    }
+    KeyName(name, keyword)
 }
 
 /// `<referential_action> =
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -22,7 +22,6 @@ use serde::{Deserialize, Serialize};
 
 use crate::ast::value::escape_single_quote_string;
 use crate::ast::{display_comma_separated, display_separated, DataType, Expr, Ident, ObjectName};
-use crate::keywords::Keyword;
 use crate::tokenizer::Token;
 
 /// An `ALTER TABLE` (`Statement::AlterTable`) operation
@@ -218,6 +217,14 @@ impl fmt::Display for AlterColumnOperation {
     }
 }
 
+/// For mysql index, here may be there format for index, there patterns are similar。
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum KeyFormat {
+    Unique,
+    Key,
+    Index,
+}
 /// A table-level constraint, specified in a `CREATE TABLE` or an
 /// `ALTER TABLE ADD <constraint>` statement.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


src/ast/ddl.rs
```diff
@@ -256,11 +263,38 @@ impl fmt::Display for TableConstraint {
                 name,
                 columns,
                 is_primary,
+                is_mysql_unique_key,
```
AugustoFKL, line 266: This is a bad design that can turn this code huge when considering other dialects.

If you `really` need the dialect info here, maybe you could have the `Dialect` itself, not that level of specificity here.

Since there are 3 possible options, couldn't we use an enum variant? Like:

```rust
pub enum UniqueType {
    PrimaryKey,
    Key,
    Simple
} 
```

(ignore the horrible naming, but you got the idea hehe)

RainJoe: Use enum variant would not work, i think using specific dialect is a good solution, but when I add dialect trait in TableConstrait struct, the compiler says: the trait bound `dyn dialect::Dialect: Clone` is not satisfied。
There may not have explict `constrait` key word in mysql create statement, `constraint` keyword is optional, like:
```
CREATE TABLE table_name(
   column_name1 column_definition,
   column_name2 column_definition,
   ...,
   [CONSTRAINT constraint_name]
      UNIQUE(column_name1,column_name2)
);
```
 so use the origin `display_constraint_name` method not work，This is why I use `is_mysql_unique_key` to distinguish mysql dialect specifically.

I haven't been able to find an elegant way to solve it yet, any good idea?

src/ast/ddl.rs
```diff
@@ -247,6 +249,11 @@ pub enum TableConstraint {
         name: Option<Ident>,
         expr: Box<Expr>,
     },
+    Key {
```
alamb, line 252: Can you please add comments here about what this field means / how it maps back to SQL?

src/ast/ddl.rs
```diff
@@ -228,6 +229,7 @@ pub enum TableConstraint {
         columns: Vec<Ident>,
         /// Whether this is a `PRIMARY KEY` or just a `UNIQUE` constraint
         is_primary: bool,
+        is_mysql_unique_key: bool,
```
alamb, line 232: Please add comments to this field explaining what it means (aka how is this different)

src/ast/ddl.rs
alamb: > Use enum variant would not work

Why would this not work?

alamb (commented):
Thank you for your contribution @RainJoe 

I suggest following the idea of @AugustoFKL and using an enum variant to describe the different possible parsed versions of the `CONSTRAINT`s keyword. 

I am not familiar with the differences in syntax -- perhaps if you could summarize the differences in the description of this PR we could offer some more helpful feedback

src/ast/ddl.rs
AugustoFKL: I'm almost sure this is related to MySQL's index creation [1](https://dev.mysql.com/doc/refman/8.0/en/create-table.html). That's something I wanted to have as well.

Conceptually has the same idea of PostgreSQL's CREATE INDEX [2](https://www.postgresql.org/docs/current/sql-createindex.html) statement and MySQL's [3](https://dev.mysql.com/doc/refman/8.0/en/create-index.html) one, but inside the create table.

In MySQL seems better to have it inside the CREATE TABLE, since their CREATE INDEX doesn't support `IF NOT EXISTS`, while the CREATE TABLE does.

[1]: https://dev.mysql.com/doc/refman/8.0/en/create-table.html
[2]: https://www.postgresql.org/docs/current/sql-createindex.html
[3]: https://dev.mysql.com/doc/refman/8.0/en/create-index.html

src/ast/ddl.rs
RainJoe: If I remove `is_mysql_unique_key` and use the original display method, then run the parse_table_colum_option_key test, it would panic, the painc message show the difference:

thread parse_table_colum_option_key panicked at assertion failed: (left == right)
  left: "CREATE TABLE foo (modification_time DATETIME ON UPDATE, **UNIQUE KEY u_m_t (modification_time))**",
 right: "CREATE TABLE foo (modification_time DATETIME ON UPDATE, **CONSTRAINT u_m_t UNIQUE (modification_time))**", src/test_utils.rs:95:13

AugustoFKL: @RainJoe but you can still use an enum or something like that. Using a single boolean makes the expansion difficult, and doesn't make that much sense as it is really specific information added to the general AST. Using an enum there allows the upper levels to match and handle as desired when, for example, they're using the PostgreSQL dialect.

RainJoe: @AugustoFKL Thanks for your suggestion，I will make some modifications and create a now PR later.

RainJoe:
@AugustoFKL PTAL
I made some modifications at your suggestion.

src/ast/ddl.rs
AugustoFKL: @RainJoe please LMK if I'm wrong, but `UNIQUE KEY` isn't the same as a `UNIQUE` constraint, while only `KEY` refers to the same as `CREATE INDEX`?

Wouldn't be better to keep `UNIQUE [KEY | INDEX]` as a `UNIQUE` table constraint?

src/parser.rs
```diff
@@ -2562,10 +2568,38 @@ impl<'a> Parser<'a> {
...
+                        format: KeyFormat::Index,
+                    })),
+                    _ => unreachable!(),
+                }
+            }
             Token::Word(w) if w.keyword == Keyword::PRIMARY || w.keyword == Keyword::UNIQUE => {
                 let is_primary = w.keyword == Keyword::PRIMARY;
                 if is_primary {
                     self.expect_keyword(Keyword::KEY)?;
+                } else if dialect_of!(self is MySqlDialect) {
```
AugustoFKL, line 2595: Please add the `GenericDialect` here. The idea of that, as long as I understood by [this](https://github.com/sqlparser-rs/sqlparser-rs/pull/642#discussion_r985143947) comment, is to support everything possible.

src/parser.rs
```diff
@@ -2562,10 +2568,38 @@ impl<'a> Parser<'a> {
...
+                        name,
+                        columns,
+                        format: KeyFormat::Key,
+                    })),
+                    Keyword::INDEX => Ok(Some(TableConstraint::Key {
+                        name,
+                        columns,
+                        format: KeyFormat::Index,
+                    })),
+                    _ => unreachable!(),
```
AugustoFKL, line 2588: Please replace this with an error. I left a possible approach for cases like that [here](https://github.com/sqlparser-rs/sqlparser-rs/pull/608#discussion_r985164256)

src/parser.rs
```diff
@@ -2526,9 +2526,15 @@ impl<'a> Parser<'a> {
         } else if self.parse_keywords(&[Keyword::ON, Keyword::UPDATE])
             && dialect_of!(self is MySqlDialect)
         {
-            Ok(Some(ColumnOption::DialectSpecific(vec![
-                Token::make_keyword("ON UPDATE"),
-            ])))
+            if self.parse_keyword(Keyword::CURRENT_TIMESTAMP) {
```
AugustoFKL, line 2529: Why did it change?

It's not related to your PR, so shouldn't have changed at all

src/ast/ddl.rs
```diff
@@ -428,6 +474,19 @@ fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
     ConstraintName(name)
 }
 
+fn display_key_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
```
AugustoFKL, line 477: Too much complex, this isn't necessary at all...

Please can you try to make it simple? LMK if you need any help

src/ast/ddl.rs
```diff
@@ -217,6 +217,14 @@ impl fmt::Display for AlterColumnOperation {
     }
 }
 
+/// For mysql index, here may be there format for index, there patterns are similar。
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum KeyFormat {
```
AugustoFKL, line 223: As I stated in another comment, `UNIQUE` is a constraint. Please don't change the current handle structure for it, as this doesn't make sense. You're changing the AST to match only one dialect.

src/ast/ddl.rs
RainJoe: Yes, it's not same. I have fixed it, PTAL.

src/parser.rs
RainJoe: It's hard to cover all dialects now. so using the specific dialect to avoid affecting other dialects.

src/parser.rs
RainJoe: There would be only two cases, and both of it are covered, so it's actually unreachable.

src/ast/ddl.rs
RainJoe: It's similar to display_constraint_name, the name dose not implement display method, so i think it's necessary to do this.

src/ast/ddl.rs
RainJoe: If you don't do this. this libaray would never support for mysql create statement. I think we should make it work first, and if someone have a better solution，we can apply it later.

src/parser.rs
RainJoe: I mentioned it in my [first comment](https://github.com/sqlparser-rs/sqlparser-rs/pull/602#issue-1361406359).

src/parser.rs
AugustoFKL: Fine, but can't you return an error here?

It's better to have an error (like the `self.expected(...)`) than having a unreachable

src/parser.rs
AugustoFKL: @RainJoe THe`GenericDialect` is the one that contains literally EVERY single possible AST structure. Therefore, it should contemplate any possible query

`dialect_of!(self is GenericDialect | MySqlDialect)` should be enough

src/ast/ddl.rs
AugustoFKL: I'll try something up, would you mind waiting a day or two?

AugustoFKL: @RainJoe would you mind if I tried a different approach and you took a look in a day or two?

Sorry for asking that, I know that this has been your work, don't want to look like throwing that away

RainJoe: Of course not. waitting for your approach.

AugustoFKL: @RainJoe @alamb could you please leave your opinion? https://github.com/sqlparser-rs/sqlparser-rs/pull/665

RainJoe: @AugustoFKL Thanks for your work, but it didn't solve my problem, maybe I would just use my fork repository. It's seems hard to consider all possible scenarios.

alamb:
Marking as draft as it seems to be waiting on feedback. Please mark as ready for review if/when it is

src/ast/ddl.rs
AugustoFKL: @RainJoe can you give examples of what you intend about this task?

The index usage itself is being parsed... What is the syntax you need?

alamb:
@RainJoe  can you see if https://github.com/sqlparser-rs/sqlparser-rs/pull/685 works for you?

Issue #597, opened 2022-08-31
Add support for postgresql tablesample command


ayushdg:
Postgresql supports an optional `[ TABLESAMPLE sampling_method ( argument [, ...] ) [ REPEATABLE ( seed ) ] ]` during a select statement to retrieve a subset of rows based on the arguments provided.

LucaCappelletti94:
Completed by #1580.

PR #608, opened 2022-09-09
Add support for parsing tablesample

ayushdg:
Initial attempt at addressing #597.

- [x] TODO: Add tests


src/ast/query.rs
```diff
@@ -455,6 +460,44 @@ impl fmt::Display for TableFactor {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum SamplingMethod {
+    BERNOULLI,
+    SYSTEM,
+}
+
+impl fmt::Display for SamplingMethod {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            SamplingMethod::BERNOULLI => write!(f, " BERNOULLI"),
```
andygrove, line 495:
```suggestion
            SamplingMethod::BERNOULLI => write!(f, "BERNOULLI"),
```

andygrove:
Looking good so far :+1:

src/parser.rs
```diff
@@ -4227,6 +4230,33 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_tablesample_args(&mut self) -> Result<Option<TableSample>, ParserError> {
```
ayushdg, line 4233: There are subtle differences in how tablesample is parsed in postgresql vs other dialects. Eg: mssql expects the sampling_fraction to also specify `rows/percent`. Does it make sense to have this check (and tests) for postgresql only?

src/ast/query.rs
```diff
@@ -455,6 +460,44 @@ impl fmt::Display for TableFactor {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct TableSample {
+    pub sampling_method: SamplingMethod,
+    pub sampling_fraction: Box<Expr>,
```
ayushdg, line 467: SQL Syntax allows for the sampling_fraction and seed value to be a real valued expression but not arbitrary columns. is Expr still the best type for this member?

src/parser.rs
andygrove: Would it make sense to add `parse_tablesample_args` to the `Dialect` trait and have a default implementation that returns `None`? The dialects that want to support it can add their custom parsing. The `TableSample` struct would still need to be a superset of all supported options across all dialects though.

alamb:
FYI I am planning to make a sqlparser-rs release in the next day or two. This PR appears to have stalled

src/parser.rs
augusto-vaultree: Isn't it better to use the dialect locally to handle impl differences, instead of adding it to the `Dialect` trait?

Otherwise the parser would end up with incompatible concepts around the code: some places use the dialect to validate, others use the `Dialect` trait to implement specific concepts. Would be nice to have a standard around where to put specific implementations

ayushdg:
Thanks for the ping @alamb. I got carried away with other thing, but I'm picking this back up now. Hopefully I can have things sorted by this release. If not, we can push to the next one

src/parser.rs
AugustoFKL: Isn't it better to use the dialect locally to handle impl differences, instead of adding it to the `Dialect` trait?

Otherwise the parser would end up with incompatible concepts around the code: some places use the dialect to validate, others use the `Dialect` trait to implement specific concepts. Would be nice to have a standard around where to put specific implementations

src/ast/query.rs
```diff
@@ -344,6 +344,7 @@ pub enum TableFactor {
         args: Option<Vec<FunctionArg>>,
         /// MSSQL-specific `WITH (...)` hints such as NOLOCK.
         with_hints: Vec<Expr>,
+        tablesample: Option<TableSample>,
```
AugustoFKL, line 347: Please add a comment about the source/description

src/ast/query.rs
AugustoFKL: @ayushdg @andygrove as the standard states ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#sample-clause)), we can have a huge amount of expressions there (described as a term ([2](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#term)) and a few additions to it). It can support values, but a lot of additional things (e.g., numeric value functions ([3](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#numeric-value-function))). Is it really reasonable to have a `Value` there?

Maybe have a task for an expansion to all other possible values, that includes but is not restricted to simple values?

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#sample-clause
[2] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#term
[3] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#numeric-value-function

src/keywords.rs
```diff
@@ -98,6 +98,7 @@ define_keywords!(
     BEGIN,
     BEGIN_FRAME,
     BEGIN_PARTITION,
+    BERNOULLI,
```
AugustoFKL, line 101: Bernoulli isn't reserved ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#non-reserved-word)). Can't you match against it without making it a keyword?

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#non-reserved-word

src/parser.rs
```diff
@@ -4227,6 +4230,33 @@ impl<'a> Parser<'a> {
...
+        if self.parse_keyword(Keyword::TABLESAMPLE) {
+            let sampling_method =
+                match self.expect_one_of_keywords(&[Keyword::BERNOULLI, Keyword::SYSTEM])? {
+                    Keyword::BERNOULLI => SamplingMethod::BERNOULLI,
+                    Keyword::SYSTEM => SamplingMethod::SYSTEM,
+                    _ => unreachable!(),
+                };
```
AugustoFKL, line 4240: You're generating an `unreachable` unnecessarily. Since the behavior differs between the variants, you can replace this by:

```suggestion
let sampling_method = if self.parse_keyword(Keyword::BERNOULLI) {
    SamplingMethod::BERNOULLI
}  else if self.parse_keyword(Keyword::SYSTEM) {
    SamplingMethod::SYSTEM
} else {
    return self.expected("BERNOULLI or SYSTEM", self.peek_token()),
}
```

alamb:

PR #609, opened 2022-09-10
Move `Value::Interval` to `Expr::Interval` 

ding-young:
Current ast `Interval` is defined as a variant of enum `Value`, which means a literal value. However, considering that the field `value` of Interval holds wrapped Expr, `Value` becomes somewhat requires endless evaluation... not "literal". Personally, it would be nice to change Value::Interval to Expr::Interval. 
```rust
Interval {
        value: Box<Expr>,
        leading_field: Option<DateTimeField>,
        /// or as `__ TO SECOND(x)`.
    }
```

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -410,6 +410,25 @@ pub enum Expr {
     ArrayIndex { obj: Box<Expr>, indexes: Vec<Expr> },
     /// An array expression e.g. `ARRAY[1, 2]`
     Array(Array),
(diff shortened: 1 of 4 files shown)


alamb (approved):
This makes sense to me @ding-young  -- thank you for the contribution

I think @togami2864  added `Expr` support for Interval in https://github.com/sqlparser-rs/sqlparser-rs/pull/517 whereas it previously was just a literla

PR #618, opened 2022-09-21
Fix parsing CLOB, BINARY, VARBINARY, BLOB data type

### LINKED ISSUE #518: VARBINARY or BLOB columns not parsed
I see the DataType enum has variants for these but it doesn't parse them correctly, or perhaps I am doing something wrong.

Sql:
```mysql
CREATE TABLE IF NOT EXISTS `data` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `name` VARCHAR(255) NOT NULL,
  `key` VARCHAR(255) NOT NULL,
  `secret` BLOB(255) NOT NULL,
  PRIMARY KEY (`id`))

# or

CREATE TABLE IF NOT EXISTS `data` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `name` VARCHAR(255) NOT NULL,
  `key` VARCHAR(255) NOT NULL,
  `secret` VARBINARY(255) NOT NULL,
  PRIMARY KEY (`id`))
```

Error:
```
unexpected end of input, sql parser error: Expected ',' or ')' after column definition, found:
 ( input: 'CREATE TABLE IF NOT EXISTS `data` ( `id` INT UNSIGNED NOT NULL AUTO_INCREMENT, `name` VARCHAR(255) NOT NULL, 
`key` VARCHAR(255) NOT NULL, `secret` BLOB (255) NOT NULL, PRIMARY KEY (`id`))'
```

ding-young:
## Related Issue
Resolves #518 

## Description
Parse error occurs for `CLOB`, `BINARY`, `VARBINARY`, and `BLOB` data types though they are included as variants for data type & keywords in sqlparser. Since all these 4 data types contain (non-optional) u64 value, added `parse_precision` fn.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -1647,6 +1647,46 @@ fn parse_cast() {
         },
         expr_from_projection(only(&select.projection))
     );
+
+    let sql = "SELECT CAST(id AS CLOB(50)) FROM customer";
+    let select = verified_only_select(sql);
+    assert_eq!(
+        &Expr::Cast {
+            expr: Box::new(Expr::Identifier(Ident::new("id"))),
+            data_type: DataType::Clob(50)
+        },
+        expr_from_projection(only(&select.projection))
+    );
+
+    let sql = "SELECT CAST(id AS BINARY(50)) FROM customer";
+    let select = verified_only_select(sql);
+    assert_eq!(
+        &Expr::Cast {
+            expr: Box::new(Expr::Identifier(Ident::new("id"))),
+            data_type: DataType::Binary(50)
+        },
+        expr_from_projection(only(&select.projection))
+    );
+
+    let sql = "SELECT CAST(id AS VARBINARY(50)) FROM customer";
+    let select = verified_only_select(sql);
+    assert_eq!(
+        &Expr::Cast {
+            expr: Box::new(Expr::Identifier(Ident::new("id"))),
+            data_type: DataType::Varbinary(50)
+        },
+        expr_from_projection(only(&select.projection))
+    );
+
+    let sql = "SELECT CAST(id AS BLOB(50)) FROM customer";
+    let select = verified_only_select(sql);
+    assert_eq!(
+        &Expr::Cast {
+            expr: Box::new(Expr::Identifier(Ident::new("id"))),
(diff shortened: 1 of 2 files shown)


alamb (commented):
Thank you for this @ding-young  -- the code looks good. Can you please add regression tests so that this syntax is not accidentally broken in the future? 

Thank you

tests/sqlparser_common.rs
```diff
@@ -1647,6 +1647,46 @@ fn parse_cast() {
...
+
+    let sql = "SELECT CAST(id AS CLOB(50)) FROM customer";
+    let select = verified_only_select(sql);
+    assert_eq!(
+        &Expr::Cast {
+            expr: Box::new(Expr::Identifier(Ident::new("id"))),
+            data_type: DataType::Clob(50)
+        },
+        expr_from_projection(only(&select.projection))
+    );
```
ding-young, line 1659: @alamb I added test cases for `parse_cast` with these data types since I could not find test fn `parse_datatype`.. Thanks for review.

alamb (approved):
LGTM -- thanks @ding-young

augusto-vaultree:
@alamb @ding-young 

`CREATE TABLE tb (id BINARY);` works on MySQL, so the precision is actually optional

The same for `CREATE TABLE tb_3 (id BLOB);`

This seems a regression of support for me.

AugustoFKL:
@alamb @ding-young 

`CREATE TABLE tb (id BINARY);` works on MySQL, so the precision is actually optional

The same for `CREATE TABLE tb_3 (id BLOB);`

This seems a regression of support for me.

AugustoFKL:
image

As the standard shows, BINARY and BLOB have optional precision

alamb:
Thanks for the report @AugustoFKL  -- any chance you can make a PR to fix it?

AugustoFKL:
@alamb sure, I can do it today

I also need to add precision to `TIMESTAMP`, `TIMESTAMPZ`, `DATETIME`, and `TIME`

I don't like to do multiple things in one PR, so I'll possibly open two (or one for each one?)

alamb:
Multiple small PRs would be most preferable. 🙏

PR #628, opened 2022-09-27
Disambiguate CREATE ROLE ... USER and GROUP

alamb:
Rather than assume `ROLE` and `USER` are synonyms in the parser, let's parse them separately and allow upstream crates to treat them the same if they desire

Per comment https://github.com/sqlparser-rs/sqlparser-rs/pull/598/files#r981277941 on contribution from @blx


alamb:
cc @blx

blx (approved):
This seems more sensible and flexible 👍

PR #630, opened 2022-09-27
Adding mediumint support

AugustoFKL:
Adding MySQ's mediumint support([1](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html))

Other INT types are already supported (e.g., MySQL's BIGINT), so adding `MEDIUMINT` should be a problem...

[1] : https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
Resolves: https://github.com/sqlparser-rs/sqlparser-rs/issues/625


alamb (approved):
Thanks @AugustoFKL  -- looks good to me

PR #635, opened 2022-09-28
Support `CEIL(expr TO DateTimeField)` and `FLOOR(expr TO DateTimeField)`

sarahyurick:
Hi, I'd like to be able to parse the following statements:
```
SELECT CEIL(d TO DAY) FROM df
SELECT CEIL(d TO HOUR) FROM df
SELECT CEIL(d TO MINUTE) FROM df
SELECT CEIL(d TO SECOND) FROM df
SELECT CEIL(d TO MILLISECOND) FROM df
```
and
```
SELECT FLOOR(d TO DAY) FROM df
SELECT FLOOR(d TO HOUR) FROM df
SELECT FLOOR(d TO MINUTE) FROM df
SELECT FLOOR(d TO SECOND) FROM df
SELECT FLOOR(d TO MILLISECOND) FROM df
```

I also added `Microsecond` and `Millisecond` in addition to the existing (plural) `Microseconds` and `Milliseconds`.


tests/sqlparser_common.rs
```diff
@@ -1753,18 +1753,72 @@ fn parse_extract() {
...
     verified_stmt("SELECT EXTRACT(QUARTER FROM d)");
     verified_stmt("SELECT EXTRACT(TIMEZONE FROM d)");
     verified_stmt("SELECT EXTRACT(TIMEZONE_HOUR FROM d)");
     verified_stmt("SELECT EXTRACT(TIMEZONE_MINUTE FROM d)");
 
-    let res = parse_sql_statements("SELECT EXTRACT(MILLISECOND FROM d)");
+    let res = parse_sql_statements("SELECT EXTRACT(JIFFY FROM d)");
     assert_eq!(
-        ParserError::ParserError("Expected date/time field, found: MILLISECOND".to_string()),
+        ParserError::ParserError("Expected date/time field, found: JIFFY".to_string()),
```
sarahyurick, line 1769: This existing check implies that not having the singular `Millisecond` and `Microsecond` (as opposed to the plural) was intentional, and I was wondering the reason behind this?

Let me know if you'd like me to revert these changes to support only the plural versions.

alamb (commented):
Thank you for the contribution @sarahyurick  -- what database supports the CEIL(<expr> as date) syntax?

Do you think this will cause issues with existing queries that use `ceil` as a function, for example

```sql

select ceil(float_column) from my_table;
```

?

sarahyurick:
> Thank you for the contribution @sarahyurick -- what database supports the CEIL( as date) syntax?
> 
> Do you think this will cause issues with existing queries that use `ceil` as a function, for example
> 
> ```sql
> select ceil(float_column) from my_table;
> ```
> 
> ?

In terms of other databases that support this syntax, I've had trouble finding examples, but am mostly trying to support existing tests in Dask-SQL that were originally implemented with https://github.com/dask-contrib/dask-sql/pull/91.

I can create some tests with `CEIL(float_column)` to try to gauge possible issues there.

sarahyurick:
Okay @alamb, let me know what you think of this implementation.

I created a special `DateTimeField::NoDateTime` and then, depending on whether or not the `TO` keyword is there, we construct `Expr::Ceil`/`Expr::Floor` accordingly.

Similarly, when we're formatting the expression, we check if we have a `DateTimeField::NoDateTime`, which means that it should be formatted as `CEIL(x)`, else it should be formatted as `CEIL(expr TO DateTimeField)`.

src/parser.rs
```diff
@@ -849,6 +851,56 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_ceil_expr(&mut self) -> Result<Expr, ParserError> {
+        self.expect_token(&Token::LParen)?;
+        let expr = self.parse_expr()?;
+        let keyword_to = self.expect_keyword(Keyword::TO);
```
alamb, line 857: Could you use `self.consume_token` here rather than `expect_token`, as a lack of a `TO` isn't really an error?

https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser.rs#L1743

src/parser.rs
```diff
@@ -849,6 +851,56 @@ impl<'a> Parser<'a> {
...
+                self.expect_token(&Token::RParen)?;
+                Ok(Expr::Ceil {
+                    expr: Box::new(expr),
+                    field,
+                })
+            }
+        }
+    }
+
+    pub fn parse_floor_expr(&mut self) -> Result<Expr, ParserError> {
```
alamb, line 879: This function is almost the same as `parse_ceil_expr` -- would it be possible to refactor them and reduce the duplication?

tests/sqlparser_common.rs
```diff
@@ -1753,18 +1753,84 @@ fn parse_extract() {
...
+        res.unwrap_err()
+    );
+}
+
+#[test]
+fn parse_ceil_number() {
+    verified_stmt("SELECT CEIL(1.5)");
```
alamb, line 1776: 👍

src/ast/mod.rs
```diff
@@ -331,6 +331,16 @@ pub enum Expr {
         field: DateTimeField,
         expr: Box<Expr>,
     },
+    /// CEIL(<expr> TO DateTimeField)
```
alamb, line 334:
```suggestion
    /// CEIL(<expr> [TO DateTimeField])
```

src/ast/mod.rs
```diff
@@ -331,6 +331,16 @@ pub enum Expr {
...
+    Ceil {
+        expr: Box<Expr>,
+        field: DateTimeField,
+    },
+    /// FLOOR(<expr> TO DateTimeField)
```
alamb, line 339:
```suggestion
    /// FLOOR(<expr> [TO DateTimeField])
```

alamb (approved):
Thank you @sarahyurick  -- I think this PR looks almost ready to go. I had a few style suggestions, but I think the basic structure looks good as do the tests. Thank you.

sarahyurick:
OK, I've refactored `parse_ceil_expr` and `parse_floor_expr` into `parse_ceil_floor_expr`. Let me know what you think.

alamb (approved):

PR #642, opened 2022-10-01
Correct order of arguments in  `LIMIT x,y` , restrict to MySql and generic dialects

AugustoFKL:
Fixing MySQL `LIMIT x, y` syntax([1](https://dev.mysql.com/doc/refman/8.0/en/select.html)).

There's no much info to add besides that. MySQL syntax expects `LIMIT x,y` as if it were `LIMIT y OFFSET x`, but we were converting it to `LIMIT x OFFSET y`.

```sql
SELECT id FROM tb LIMIT 5,3

--- is the same as

SELECT id FROM tb LIMIT 3 OFFSET 5
```

[1] : https://dev.mysql.com/doc/refman/8.0/en/select.html
Resolves: https://github.com/sqlparser-rs/sqlparser-rs/pull/613

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3762,9 +3762,25 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if offset.is_none() && self.consume_token(&Token::Comma) {
-                    // mysql style LIMIT 10, offset 5
-                    offset = Some(self.parse_offset()?)
+                if dialect_of!(self is MySqlDialect)
+                    && offset.is_none()
+                    && self.consume_token(&Token::Comma)
+                {
+                    // MySQL style LIMIT x,y => LIMIT y OFFSET x.
+                    // Check <https://dev.mysql.com/doc/refman/8.0/en/select.html> for more details.
+                    match limit {
+                        None => {
+                            offset = None;
+                        }
+                        Some(expr) => {
+                            offset = Some(Offset {
+                                value: expr.clone(),
+                                rows: OffsetRows::None,
+                            })
+                        }
+                    }
+                    // MySQL doesn't support the ALL clause, so I must parse only the expression.
+                    limit = Some(self.parse_expr()?)
                 }
             }
 
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3761,7 +3761,7 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if dialect_of!(self is MySqlDialect)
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -3762,9 +3761,18 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if offset.is_none() && self.consume_token(&Token::Comma) {
-                    // mysql style LIMIT 10, offset 5
-                    offset = Some(self.parse_offset()?)
+                if dialect_of!(self is MySqlDialect)
```
alamb, line 3764: I think the genera pattern in sqlparser is that the `General` dialect supports as much as possible.

Thus can you please also support this syntax with the general as well?

alamb (commented):
Thanks @AugustoFKL  -- this looks good. There is just one small change and I think this will be ready to go

src/parser.rs
AugustoFKL: Done!

tests/sqlparser_mysql.rs
```diff
@@ -1065,6 +1065,14 @@ fn parse_set_names() {
     assert_eq!(stmt, Statement::SetNamesDefault {});
 }
 
+#[test]
+fn parse_limit_my_sql_syntax() {
+    mysql().one_statement_parses_to(
```
alamb: If your usecase is to recover the same syntax (as make sure `LIMIT 5, 10` is serialized as `LIMIT 5,10` you probably need to change this test to use `verified_statement` or `verified_query`: https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/test_utils.rs#L108

AugustoFKL: It doesn't. Should I target that?

The original idea was just to have the correct semantic, not the exact same conversion. But I can work on having the same syntax, which would change the AST for limit and offset

alamb: I don't think you need to expand the scope of the PR to include preserving the syntax on round trip, I just wanted to ensure that was clear

alamb (approved):
Thanks @AugustoFKL

tests/sqlparser_mysql.rs
AugustoFKL: Sure! But you have a point, I'll try to look into it to see if it's possible/meaningful to make it semantically the same as the original query.

alamb: I agree with your assement that you would have to extend the struct somehow to distinguish between `LIMIT x OFFSET y` and `LIMIT y, x`

PR #646, opened 2022-10-02
645 New schema name structure

AugustoFKL:
Adding support for `AUTHORIZATION` on `CREATE SCHEMA` ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#schema-definition), [2](https://www.postgresql.org/docs/current/sql-createschema.html)) using a new enum structure. I used that structure because it makes it easier to validate and handle than Options. The other possible approach was:

```rust
pub enum Statement {
//...
    CreateSchema {
        schema_name: Option<ObjectName>,
        authorization_name: Option<Ident>
        if_not_exists: bool,
    },
//...
}
```

But it has a few problems:
1. It's harder to parse
2. We can't have both false, but that structure would allow that
3. It's harder for the caller to know what's happening.

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#schema-definition
[2] : https://www.postgresql.org/docs/current/sql-createschema.html
Resolves: https://github.com/sqlparser-rs/sqlparser-rs/issues/645

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1299,7 +1299,8 @@ pub enum Statement {
     Rollback { chain: bool },
     /// CREATE SCHEMA
     CreateSchema {
-        schema_name: ObjectName,
+        /// `<schema name> | AUTHORIZATION <schema authorization identifier>  | <schema name>  AUTHORIZATION <schema authorization identifier>`
+        schema_name: SchemaName,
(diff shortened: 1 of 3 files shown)


alamb (approved):
Looks really nice -- thank you @AugustoFKL

PR #648, opened 2022-10-02
Distinguish between `CHAR VARYING`, `CHARACTER VARYING`, and `CHARACTER` data types, differentiating them from `VARCHAR` and `CHAR`

AugustoFKL:
Adding support for `CHAR VARYING`, `CHARACTER VARYING`, and `CHARACTER` data types, differentiating them from `VARCHAR` and `CHAR`, which were mistakenly considered the same ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#character-string-type)).

The precision is optional to support PostgreSQL's syntax ([2](https://www.postgresql.org/docs/current/datatype-character.html)).

Also, adding tests to start resolving https://github.com/sqlparser-rs/sqlparser-rs/issues/2, but that's a WIP.

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#character-string-type
[2] : https://www.postgresql.org/docs/current/datatype-character.html
Resolves: https://github.com/sqlparser-rs/sqlparser-rs/issues/647

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3414,9 +3414,16 @@ impl<'a> Parser<'a> {
                 }
                 Keyword::VARCHAR => Ok(DataType::Varchar(self.parse_optional_precision()?)),
                 Keyword::NVARCHAR => Ok(DataType::Nvarchar(self.parse_optional_precision()?)),
-                Keyword::CHAR | Keyword::CHARACTER => {
+                Keyword::CHARACTER => {
                     if self.parse_keyword(Keyword::VARYING) {
-                        Ok(DataType::Varchar(self.parse_optional_precision()?))
+                        Ok(DataType::CharacterVarying(self.parse_optional_precision()?))
+                    } else {
+                        Ok(DataType::Character(self.parse_optional_precision()?))
+                    }
+                }
+                Keyword::CHAR => {
+                    if self.parse_keyword(Keyword::VARYING) {
+                        Ok(DataType::CharVarying(self.parse_optional_precision()?))
                     } else {
                         Ok(DataType::Char(self.parse_optional_precision()?))
                     }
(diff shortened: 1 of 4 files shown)


AugustoFKL:
@alamb ready for review :D

alamb (approved):

alamb:
Thank you @AugustoFKL

alamb:
This PR appears to have a conflict -- can you please resolve that @AugustoFKL ?

src/parser.rs
```diff
@@ -5288,80 +5295,46 @@ mod tests {
         });
     }
 
-    // TODO add tests for all data types? https://github.com/sqlparser-rs/sqlparser-rs/issues/2
-    // TODO when we have dialect validation by data type parsing, split test
-    #[test]
```
alamb, line 5293: seems like we have lost coverage after the rebase (e.g. the new tests for `TIME WITHOUT TIMEZONE` for example)

AugustoFKL: @alamb that's expected, as I mentioned in my comment above.

I'll add tests along the week, trying to reach 100% coverage for data type parsing. But for now, I'd like to restrict this to the ones I added if possible

alamb: I don't understand why this PR would *remove* coverage for types that were added in other PRs. I am fine with expanding coverage in future PRs

AugustoFKL: Fair enough. It was because I (dumbly) did the structure for tests twice, as it didn't please me. I'll re-add the coverage for those data types, and for now leave as "additional data types" for those that aren't covered by ANSII

AugustoFKL:
@alamb done

alamb (approved):
Thanks @AugustoFKL

PR #652, opened 2022-10-03
Patch for supporting backticks in GenericDialect

### LINKED ISSUE #632: Backticks in ObjectName argument to DESCRIBE fails to parse, dialect behavior specialization
The library cannot currently parse this under the generic dialect:

```
DESCRIBE namespace.`table`;
```

I tried making my own dialect that added backticks to the delimited identifier characters, but it's not actually possible to make that work with the library modified as-is.

The reason is because of code like this that special-cases behavior based on known vs. unknown dialects:

https://github.com/sqlparser-rs/sqlparser-rs/blob/6c8f31c367ac75ea29be030bad244c0c6a2fafee/src/parser.rs#L4388

When I define my own dialect, I hit this fallback case and it fails on _many_ examples that worked fine with the GenericDialect:

https://github.com/sqlparser-rs/sqlparser-rs/blob/6c8f31c367ac75ea29be030bad244c0c6a2fafee/src/parser.rs#L4427

This is deeply problematic. The trait API gives me no way to identify my dialect with a family or type of dialect that I can tell. This means that as soon as I want to do something as simple as extending the GenericDialect with an extra identifier delimiter, all my tests start failing for reasons that were initially mysterious.

Here's the temporary hack I'm using to work around the problem:

https://github.com/bitemyapp/sqlparser-rs/commit/96c9b0df314821e5a285fb090afc629218aa2049#diff-5638cd35d93e9e77be12ed5b86d621439eb7ed698fd2eba0e8a496a9b8ba208cR60

I am willing to submit a patch to rectify this but I am fairly bothered by the `dialect_of!(…)` stuff. Is there a way to make my custom Dialect be identified as a library-defined dialect? I don't think I can override that `impl dyn Dialect` method as `dyn Dialect` is a type and not a class of types with a trait in common. Please let me know if I am wrong.

Would you mind if I submitted an MR to hoist that machinery out of the dyn/Any hack into a trait method with a default that preserves the existing behavior. I'm open to alternatives, especially ones that involve less work, as well.

bitemyapp:
This was needed for the Athena SQL parsing I've been doing.

Fixes #632 632

### DIFF
--- a/tests/sqlparser_hive.rs
+++ b/tests/sqlparser_hive.rs
@@ -29,6 +29,19 @@ fn parse_table_create() {
     hive().verified_stmt(iof);
 }
 
+fn generic() -> TestedDialects {
+    TestedDialects {
+        dialects: vec![Box::new(GenericDialect {})]
+    }
+}
+
+#[test]
+fn parse_describe() {
+    let describe = r#"DESCRIBE namespace.`table`"#;
+    hive().verified_stmt(describe);
+    generic().verified_stmt(describe);
+}
+
 #[test]
 fn parse_insert_overwrite() {
     let insert_partitions = r#"INSERT OVERWRITE TABLE db.new_table PARTITION (a = '1', b) SELECT a, b, c FROM db.table"#;
@@ -258,12 +271,8 @@ fn parse_create_function() {
         _ => unreachable!(),
     }
 
-    let generic = TestedDialects {
-        dialects: vec![Box::new(GenericDialect {})],
-    };
-
     assert_eq!(
-        generic.parse_sql_statements(sql).unwrap_err(),
+        generic().parse_sql_statements(sql).unwrap_err(),
         ParserError::ParserError(
             "Expected an object type after CREATE, found: FUNCTION".to_string()
         )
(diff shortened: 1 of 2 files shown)


bitemyapp:
I've pushed a cleaned up version here that should be fine.

tests/sqlparser_hive.rs
```diff
@@ -29,6 +29,12 @@ fn parse_table_create() {
     hive().verified_stmt(iof);
 }
 
+#[test]
+fn parse_describe() {
```
alamb, line 33: Given this was a change to the generic Dialect (the default impl) can you also please add a test in `parser_generic.rs` as well?

alamb (commented):
Thanks for the contribution @bitemyapp

tests/sqlparser_hive.rs
bitemyapp: Yep, I think that makes sense. Thank you! 3rd kid in as many years born Monday (first daughter!), will get to it in a few days I think.

alamb: > 3rd kid in as many years born Monday (first daughter!

Congratulations and good luck!

alamb:
Marking as draft so I know it isn't waiting on review anymore

tests/sqlparser_hive.rs
bitemyapp: >can you also please add a test in parser_generic.rs as well?

I've added a generic test in the Hive test. parser_common was mostly about "all dialects", this change is just hive and generic. There are other Hive tests testing generic such that I factored out the dialect value creation into a function.

I think the remaining thing here is lifting the backtick out of the trait default implementation out to Hive and Generic's trait impls. What do you think?

PR #653, opened 2022-10-04
Decimal precision and scale enum

AugustoFKL:
Implementation of an enum to handle Decimal ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#exact-numeric-type)) precision and scale, instead of options, for two reasons:
- An enum is easier to access and handle. Options need unwrap or errors even when you KNOW that an item should exist (e.g., `DATETIME (p,s)`, if `s` is some, necessarily `p` is some).
- Remove unwrap from our code. Even if they shouldn't happen, it's not nice to have unwraps as "certainty" handlers (that was happening when we printed stuff).

Since we already didn't accept `DECIMAL` data type only with scale, that doesn't change the behavior of anything.

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#exact-numeric-type

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ Cargo.lock
 # IDEs
 .idea
 .vscode
+*.iml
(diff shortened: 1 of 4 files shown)


AugustoFKL:
@alamb I based this one on https://github.com/sqlparser-rs/sqlparser-rs/pull/648, so it duplicates its changes. Is there a better way to approach situations like that?

I use its structure, so I'm kinda stuck on that. Trying to work on things that don't conflict isn't easy hehe

PR #654, opened 2022-10-04
Enum to handle exact number precisios

AugustoFKL:
Implementation of an enum to handle Decimal ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#exact-numeric-type)) precision and scale, instead of options, for two reasons:
- An enum is easier to access and handle. Options need to unwrap or errors even when you KNOW that an item should exist (e.g., `DATETIME (p,s)`, if `s` is some, necessarily `p` is some).
- Remove unwrap from our code. Even if they shouldn't happen, it's not nice to have unwrapped as "certainty" handlers (that was happening when we printed stuff).

Since we already didn't accept the `DECIMAL` data type only with scale, that doesn't change the behavior.

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#exact-numeric-type

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -60,7 +60,7 @@ pub enum DataType {
     /// [Oracle]: https://docs.oracle.com/javadb/10.8.3.0/ref/rrefblob.html
     Blob(Option<u64>),
     /// Decimal type with optional precision and scale e.g. DECIMAL(10,2)
-    Decimal(Option<u64>, Option<u64>),
+    Decimal(ExactNumberInfo),
     /// Floating point with optional precision e.g. FLOAT(8)
     Float(Option<u64>),
     /// Tiny integer with optional display width e.g. TINYINT or TINYINT(3)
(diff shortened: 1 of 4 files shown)


src/ast/data_type.rs
```diff
@@ -297,3 +293,34 @@ impl fmt::Display for TimezoneInfo {
         }
     }
 }
+
+/// Additional information for `NUMERIC`, `DECIMAL`, and `DEC` data types
+/// following the 2016 [standard].
+///
+/// [standard]: https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#datetime-type
```
alamb, line 300: this link seems incorrect (as it points to datetime-type)?

src/parser.rs
```diff
@@ -3467,10 +3467,9 @@ impl<'a> Parser<'a> {
                 Keyword::STRING => Ok(DataType::String),
                 Keyword::TEXT => Ok(DataType::Text),
                 Keyword::BYTEA => Ok(DataType::Bytea),
-                Keyword::NUMERIC | Keyword::DECIMAL | Keyword::DEC => {
-                    let (precision, scale) = self.parse_optional_precision_scale()?;
```
alamb, line 3471: > parse_optional_precision_scale

is this function still used anywhere? If not shall we remove it? If so, perhaps there are other places that should be migrated over?

alamb (commented):
Thanks @AugustoFKL

src/parser.rs
AugustoFKL: It is used in another place, in the parser as well. TBH I didn't review the logic for the other usage, maybe it ends in the same problem, but I'll take a look in another moment.

I think it can be left like that for now.

src/ast/data_type.rs
AugustoFKL: Fixed, thanks!

alamb (approved):

alamb:
Thanks @AugustoFKL

PR #655, opened 2022-10-05
Support for INTERVAL inside window frames

### LINKED ISSUE #631: Support "date" and "timestamp" inside window frame RANGE Queries
# Support INTERVAL in window RANGE Queries

Currently, we do not support the queries taking `date` or `timestamp` for range inside window frames. Following query

```sql
SELECT 
COUNT(*) OVER (ORDER BY ts RANGE BETWEEN  '1 DAY' PRECEDING AND '1 DAY' FOLLOWING) 
FROM t;
```

produces the error below: 

```sql
ParserError("Expected literal int, found: INTERVAL")
```

Relevant section in the code, where window frame bounds are parsed can be found here.

[sqlparser-rs/parser.rs at 6afd194e947cfd800376f424ff7c300ee385cd9e · sqlparser-rs/sqlparser-rs](https://github.com/sqlparser-rs/sqlparser-rs/blob/6afd194e947cfd800376f424ff7c300ee385cd9e/src/parser.rs#L679)

mustafasrepo:
# Support for INTERVAL Inside Window Frames

****Which issue does this PR close?****
Closes #631 by adding `INTERVAL` support inside window frames. Now we can parse queries such as

```sql
SELECT 
COUNT(*) OVER (ORDER BY ts RANGE BETWEEN  INTERVAL '1 DAY' PRECEDING AND INTERVAL '1 DAY' FOLLOWING) 
FROM t;
```

****Rationale for this change****

[Datafusion](https://github.com/synnada-ai/arrow-datafusion) now supports window frame queries with [PR #3570](https://github.com/apache/arrow-datafusion/pull/3570).  However, since window frame bound only emits `u64` values, we cannot run queries like the one above in Datafusion. If this PR merges, after some minor changes we would be able to run `RANGE` queries involving `INTERVAL` in Datafusion. The roadmap of Datafusion includes a [plan](https://github.com/apache/arrow-datafusion/issues/3571) to support different types for `WindowFrameBound`.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -880,9 +880,9 @@ pub enum WindowFrameBound {
     /// `CURRENT ROW`
     CurrentRow,
     /// `<N> PRECEDING` or `UNBOUNDED PRECEDING`
-    Preceding(Option<RangeBounds>),
+    Preceding(Option<Box<Expr>>),
     /// `<N> FOLLOWING` or `UNBOUNDED FOLLOWING`.
-    Following(Option<RangeBounds>),
+    Following(Option<Box<Expr>>),
 }
 
 impl fmt::Display for WindowFrameBound {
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -880,9 +880,9 @@ pub enum WindowFrameBound {
     /// `CURRENT ROW`
     CurrentRow,
     /// `<N> PRECEDING` or `UNBOUNDED PRECEDING`
-    Preceding(Option<u64>),
+    Preceding(Option<RangeBounds>),
```
alamb, line 883: It seems like this in general can be any Expression (not just a Number of Interval): 

https://www.postgresql.org/docs/current/sql-expressions.html

> In the offset PRECEDING and offset FOLLOWING frame options, the offset must be an expression not containing any variables, aggregate functions, or window functions. The meaning of the offset depends on the frame mode:

What do you think about just parsing it as an Expression rather than special casing Number / Interval?

alamb:
Thank you @mustafasrepo  for the contribution

ozankabak:
Hi Andrew! The SQL:2016 reference seems to expect [an unsigned literal](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#unsigned-value-specification), that's why we opted for this. But you are right, PostgreSQL seems to allow expressions -- it only requires such expressions to evaluate to appropriate unsigned values. We can certainly follow the latter approach, but doing that will require both sqlparser and datafusion to accommodate that logic.

It seems to me a good approach could be two-tiered: We can implement the unsigned-value-only design first (both here and in datafusion), and then subsequently work towards PSQL compatibility. What do you think?

alamb:
> It seems to me a good approach could be two-tiered: We can implement the unsigned-value-only design first (both here and in datafusion), and then subsequently work towards PSQL compatibility. What do you think?

That is fine with me -- I was thinking that using a general Expr in the parser at first would actually make the process easier overall -- the parser / parsed structs would be simpler and DataFusion would have to special case what types it handled. Then over time datafusion could be extended to handle more and more types, but the parser wouldn't need to be changed again

mustafasrepo:
Hi Andrew, we have changed the implementation as you suggested. It improves readability and decreases the changes required. Thanks for the suggestion. In this PR  https://github.com/synnada-ai/arrow-datafusion/pull/6 you can see the corresponding datafusion change as POC implementation to support non-u64 queries inside window frames.

alamb (approved):
Love it -- thank you @mustafasrepo

PR #659, opened 2022-10-06
Create table builder structure

AugustoFKL:
Proposal of usage of a `CreateTableBuilder` structure, instead of the pure create table that we've been using right now.

This has 2 benefits:
- We can reduce calls anywhere the statement isn't meant to be "complete", i.e., when some parts are certain to be a default value.
- Callers can switch between the current statement to this structure, so they don't need to move an enum, a struct instead.

@alamb @nickolay what do you think? 

I know this resembles https://github.com/sqlparser-rs/sqlparser-rs/pull/315#issuecomment-1259597684 but, instead of replacing our current API, we can give the upstream the possibility of creating the struct, moving it with more ease around the code. If you think that's OK, I'll add a `TryFrom<Statement>` for this (or maybe just a convert method), and expand that going from the biggest statement to the smallest.

### DIFF
--- a/src/ast/helpers/stmt_create_table.rs
+++ b/src/ast/helpers/stmt_create_table.rs
@@ -1,11 +1,45 @@
 #[cfg(not(feature = "std"))]
-use alloc::{boxed::Box, string::String, vec, vec::Vec};
+use alloc::{boxed::Box, format, string::String, vec, vec::Vec};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 
 use crate::ast::{
     ColumnDef, FileFormat, HiveDistributionStyle, HiveFormat, ObjectName, OnCommit, Query,
     SqlOption, Statement, TableConstraint,
 };
+use crate::parser::ParserError;
 
+/// Builder for create table statement variant ([1]).
+///
+/// This structure helps building and accessing a create table with more ease, without needing to:
+/// - Match the enum itself a lot of times; or
+/// - Moving a lot of variables around the code.
+///
+/// # Example
+/// ```rust
+/// use sqlparser::ast::helpers::stmt_create_table::CreateTableBuilder;
+/// use sqlparser::ast::{ColumnDef, DataType, Ident, ObjectName};
+/// let builder = CreateTableBuilder::new(ObjectName(vec![Ident::new("table_name")]))
+///    .if_not_exists(true)
+///    .columns(vec![ColumnDef {
+///        name: Ident::new("c1"),
+///        data_type: DataType::Int(None),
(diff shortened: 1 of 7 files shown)


alamb:
I think this makes a lot of sense to me 👍 

I think it needs some docstrings (ideally with an example) but otherwise 🚀

AugustoFKL:
> I think this makes a lot of sense to me 👍
> 
> I think it needs some docstrings (ideally with an example) but otherwise 🚀

Ok! Will be done in 10 :)

AugustoFKL:
@alamb done!

alamb:
Looks like there is a minor issue with rustfmt/clippy

src/ast/helpers/stmt_create_table.rs
```diff
@@ -0,0 +1,325 @@
...
+use crate::ast::{
+    ColumnDef, FileFormat, HiveDistributionStyle, HiveFormat, ObjectName, OnCommit, Query,
+    SqlOption, Statement, TableConstraint,
+};
+use crate::parser::ParserError;
+
+/// Builder for create table statement variant ([1]).
+///
+/// This structure is a facilitator to allow building and accessing a create table with more ease,
+/// without needing to:
```
alamb, line 16:
```suggestion
/// This structure helps building and accessing a `create table` with more ease,
/// without needing to:
```

src/ast/helpers/stmt_create_table.rs
```diff
@@ -0,0 +1,325 @@
...
+///        options: vec![],
+/// }]);
+/// // You can access internal elements with ease
+/// assert!(builder.if_not_exists);
+/// // Convert to a statement
+/// assert_eq!(
+///    builder.build().to_string(),
+///    "CREATE TABLE IF NOT EXISTS table_name (c1 INT)"
+/// )
+///
```
alamb, line 39:
```suggestion
```

src/ast/helpers/stmt_create_table.rs
```diff
@@ -0,0 +1,325 @@
...
+    #[test]
+    pub fn test_from_valid_statement() {
+        let builder = CreateTableBuilder::new(ObjectName(vec![Ident::new("table_name")]));
+
+        let stmt = builder.clone().build();
+
+        assert_eq!(builder, CreateTableBuilder::try_from(stmt).unwrap());
+    }
+
+    #[test]
```
alamb, line 314: 👍

alamb (approved):
LGTM -- thanks @AugustoFKL

alamb:
It just needs to get a clean CI and we can merge it.

PR #662, opened 2022-10-08
Parse ARRAY_AGG for Bigquery and Snowflake

SuperBo:
There are subtle differences of ARRAY_AGG in SQL dialects:

1. https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#array-aggregate-function
2. https://docs.snowflake.com/en/sql-reference/functions/array_agg.html
3. https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions
4. https://www.postgresqltutorial.com/postgresql-aggregate-functions/postgresql-array_agg/

This the very first attempt to support correct ARRAY_AGG function while parsing BigQuery and snowflake.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -416,6 +416,8 @@ pub enum Expr {
     ArraySubquery(Box<Query>),
     /// The `LISTAGG` function `SELECT LISTAGG(...) WITHIN GROUP (ORDER BY ...)`
     ListAgg(ListAgg),
+    /// The `ARRAY_AGG` function `SELECT ARRAY_AGG(... ORDER BY ...)`
+    ArrayAgg(ArrayAgg),
     /// The `GROUPING SETS` expr.
     GroupingSets(Vec<Vec<Expr>>),
     /// The `CUBE` expr.
@@ -655,6 +657,7 @@ impl fmt::Display for Expr {
             Expr::Subquery(s) => write!(f, "({})", s),
             Expr::ArraySubquery(s) => write!(f, "ARRAY({})", s),
             Expr::ListAgg(listagg) => write!(f, "{}", listagg),
+            Expr::ArrayAgg(arrayagg) => write!(f, "{}", arrayagg),
             Expr::GroupingSets(sets) => {
                 write!(f, "GROUPING SETS (")?;
                 let mut sep = "";
@@ -3033,6 +3036,45 @@ impl fmt::Display for ListAggOnOverflow {
                 write!(f, " COUNT")
             }
         }
+    }
+}
+
+/// An `ARRAY_AGG` invocation `ARRAY_AGG( [ DISTINCT ] <expr> [ORDER BY <expr>] [LIMIT <n>] )`
+/// Or `ARRAY_AGG( [ DISTINCT ] <expr> ) [ WITHIN GROUP ( ORDER BY <expr> ) ]`
+/// ORDER BY position is defined differently for BigQuery, Postgres and Snowflake.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ArrayAgg {
+    pub distinct: bool,
+    pub expr: Box<Expr>,
+    pub order_by: Option<Box<OrderByExpr>>,
+    pub limit: Option<Box<Expr>>,
+    pub within_group: bool, // order by is used inside a within group or not
+}
+
+impl fmt::Display for ArrayAgg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "ARRAY_AGG({}{}",
+            if self.distinct { "DISTINCT " } else { "" },
+            self.expr
+        )?;
+        if !self.within_group {
+            if let Some(order_by) = &self.order_by {
+                write!(f, " ORDER BY {}", order_by)?;
+            }
+            if let Some(limit) = &self.limit {
+                write!(f, " LIMIT {}", limit)?;
+            }
+        }
+        write!(f, ")")?;
+        if self.within_group {
+            if let Some(order_by) = &self.order_by {
+                write!(f, " WITHIN GROUP (ORDER BY {})", order_by)?;
+            }
+        }
+        Ok(())
     }
 }
 
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -71,6 +71,12 @@ pub trait Dialect: Debug + Any {
     fn supports_filter_during_aggregation(&self) -> bool {
         false
     }
+    /// Returns true if the dialect supports ARRAY_AGG() [WITHIN GROUP (ORDER BY)] expressions.
+    /// Otherwise, the dialect should expect an `ORDER BY` without the `WITHIN GROUP` clause, e.g. `ANSI` [(1)].
+    /// [(1)]: https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#array-aggregate-function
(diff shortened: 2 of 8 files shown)


alamb:
@AugustoFKL  do you have time to review this PR?

AugustoFKL:
@SuperBo please fix the linter warnings

src/ast/mod.rs
```diff
@@ -2844,6 +2847,41 @@ impl fmt::Display for ListAggOnOverflow {
...
+/// An `ARRAY_AGG` invocation `ARRAY_AGG( [ DISTINCT ] <expr> [ORDER BY <expr>] [LIMIT <n>] )
+/// [ WITHIN GROUP (ORDER BY <within_group>  ]`
+/// ORDERY BY position is defined differently for BigQuery, Postgres and Snowflake
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ArrayAgg {
+    pub distinct: bool,
+    pub expr: Box<Expr>,
+    pub order_by: Option<Box<OrderByExpr>>,
+    pub limit: Option<Box<Expr>>,
```
AugustoFKL, line 2859: @SuperBo any reason why you did the implementation for all dialects the same time?

You didn't complete none, but added parts of different ones. Got me a little confused

src/ast/mod.rs
```diff
@@ -2844,6 +2847,41 @@ impl fmt::Display for ListAggOnOverflow {
     }
 }
 
+/// An `ARRAY_AGG` invocation `ARRAY_AGG( [ DISTINCT ] <expr> [ORDER BY <expr>] [LIMIT <n>] )
+/// [ WITHIN GROUP (ORDER BY <within_group>  ]`
+/// ORDERY BY position is defined differently for BigQuery, Postgres and Snowflake
```
AugustoFKL, line 2852:
```suggestion
/// ORDER BY position is defined differently for BigQuery, Postgres and Snowflake.
```

src/ast/mod.rs
```diff
@@ -2844,6 +2847,41 @@ impl fmt::Display for ListAggOnOverflow {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ArrayAgg {
+    pub distinct: bool,
+    pub expr: Box<Expr>,
+    pub order_by: Option<Box<OrderByExpr>>,
+    pub limit: Option<Box<Expr>>,
+    pub within_group: Option<Box<OrderByExpr>>,
```
AugustoFKL, line 2860: To be honest I'm not quite sure whether is good or not the usage of the order by inside the within group.

Considering the little difference, can't you just have a flag like `within_group` to identify if the order by is used inside a within group or not?

I couldn't find a situation where both are possible. Having both there make it possible to invalid (theoretically impossible) states to be built, which can be avoided in this case.

src/dialect/mod.rs
```diff
@@ -71,6 +71,10 @@ pub trait Dialect: Debug + Any {
     fn supports_filter_during_aggregation(&self) -> bool {
         false
     }
+    /// Does the dialect supports ARRAY_AGG() [WITHIN GROUP (ORDER BY)] or ARRAY_AGG([ORDER BY])
+    fn supports_within_after_array_aggregation(&self) -> bool {
```
AugustoFKL, line 75: The comment is confusing.

The comment make it sounds like if any of both is valid, it will return true.

Could be something like

```rust
/// Returns true if the dialect supports ARRAY_AGG() [WITHIN GROUP (ORDER BY)] expressions.
///
/// Otherwise, the dialect should expect an `ORDER BY` without the `WITHIN GROUP` clause, e.g. `ANSI` [(1)].
/// [(1)]: https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#array-aggregate-function
```

Having the function here is not my usual approach, as it will require maintenance over other dialects individually, but I don't think it's wrong.

AugustoFKL: Also, @SuperBo, one thing:
- Generic dialect should be the most permissive one so, if you actually can have both structures in a query, generic dialect should allow both. But considering my previous comment, I'd allow the most common one, maybe?

@alamb what do you think? This is a divergence from dialects. I'd go for the one with more dialects, so the generic is as generic as possible.

tests/sqlparser_snowflake.rs
```diff
@@ -143,6 +143,26 @@ fn test_single_table_in_parenthesis_with_alias() {
     );
 }
 
+#[test]
+fn test_array_agg_func() {
+    for (sql, canonical) in [
```
AugustoFKL, line 148: @SuperBo the only difference between both is the case, right?

I'd use the `verified_stmt` instead of `one_statement_parses_to` using the same case. You're not testing the case structure, but the ARRAY_AGG parsing. There's no need to have a proof on that.

tests/sqlparser_common.rs
```diff
@@ -1899,6 +1901,36 @@ fn parse_listagg() {
...
+            Box::new(PostgreSqlDialect {}),
+            Box::new(MsSqlDialect {}),
+            Box::new(AnsiDialect {}),
+            Box::new(HiveDialect {}),
+        ]
+    };
+
+    for (sql, canonical) in [
+        (
+            "select array_agg(x order by x) as a from T",
```
AugustoFKL, line 1918: Same as last comment. Please use verified_stmt.

src/parser.rs
```diff
@@ -1067,6 +1068,57 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    pub fn parse_array_agg_expr(&mut self) -> Result<Expr, ParserError> {
+        self.expect_token(&Token::LParen)?;
+        let distinct = self.parse_all_or_distinct()?;
```
AugustoFKL, line 1073: `parse_all_or_distinct` will accept `ALL`, which isn't accepted by Snowflake and BigQuery, although it is by PostgreSQL [(3)].

I'd suggest you to add support for `ALL`, or adjust the structure to sinalize an `ALL`. Having this generates undesired permissiveness.

[1] : https://docs.snowflake.com/en/sql-reference/functions/array_agg.html
[2] : https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#array_agg
[3] : https://www.postgresql.org/docs/9.5/sql-expressions.html#SYNTAX-AGGREGATES

AugustoFKL (changes requested):

src/ast/mod.rs
SuperBo: I agree to your comment. In the beginning, I want to keep it like this for the purpose of easier to rewrite. But this can leads to invalid state. Will work on this.

src/dialect/mod.rs
SuperBo: I think you are right, the ANSI one should be default option.

src/ast/mod.rs
SuperBo: change in latest commit

tests/sqlparser_snowflake.rs
SuperBo: Fixed

src/dialect/mod.rs
SuperBo: @AugustoFKL, are we good to go with this or should we change to more generics behavior?

src/ast/mod.rs
SuperBo: I changed this. You can review again

AugustoFKL (approved):
LGTM

@alamb ready for your eyes 👀

alamb (approved):
Looks good to me -- thank you @SuperBo  and @AugustoFKL

PR #663, opened 2022-10-09
Add support for unit for character string types

AugustoFKL:
Add support for optional `OCTETS` and `CHARACTERS` after the precision. I haven't seen that syntax for any big dialects that I remember but, since ANSI supports it ([1](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#character-string-type)), I wanted to add the possible syntax.

@alamb ready for review :)

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#character-string-type
Related to: https://github.com/sqlparser-rs/sqlparser-rs/issues/661


alamb (approved):
Makes sense to me -- thank you @AugustoFKL

PR #664, opened 2022-10-09
Replace the two booleans from 'SET ROLE' with a single enum `ContextModifier`

AugustoFKL:
This way, we don't rely on the parser to have valid structures for that statement, as the structure itself is valid.

Postgresql doesn't allow both to be present ([1](https://www.postgresql.org/docs/current/sql-set-role.html)), so there's no meaning in having a structure that allows both. Also, ANSI ([2](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#set-role-statement)) and a lot of other dialects support that command, so having the comment PostgreSQL specific doesn't make sense.

@alamb This is ready for review. Also, do you know why the coverage is so weird? Like, a lot of enum matches is not covered, even though their interior is

[1] : https://www.postgresql.org/docs/current/sql-set-role.html
[2] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#set-role-statement

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -11,13 +11,6 @@
 // limitations under the License.
 
 //! SQL Abstract Syntax Tree (AST) types
-mod data_type;
-mod ddl;
-pub mod helpers;
-mod operator;
-mod query;
-mod value;
-
 #[cfg(not(feature = "std"))]
 use alloc::{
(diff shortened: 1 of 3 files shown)


alamb (approved):
Makes sense to me -- thank you @AugustoFKL

PR #665, opened 2022-10-10
Adding MySQL table option {INDEX | KEY} to the CREATE TABLE,  restrict to MySql and generic dialects

AugustoFKL:
Breaking changes:
- Now HASH and BTREE are KEYWORDS, and using them as names can result in errors.
- Now 'KEY' and 'INDEX' column names start the parsing of a table constraint if unquoted for the Generic dialect. This results in possible conflicts if canonical results are compared for all dialects if a column is named 'key' without quotes.

Theoretically, the behavior should be the same as CREATE INDEX, but we cannot make that assumption, so the parse is (almost) identical to the input. MySQL seems to be the only one to support this syntax. ([1](https://dev.mysql.com/doc/refman/8.0/en/create-table.html))

Now we can support statements like:

```sql
CREATE TABLE (id INT, INDEX (id));

CREATE TABLE (id INT, KEY index_name USING BTREE (id))
```

The supported syntax implemented here is:
```
CREATE TABLE (..., {KEY | INDEX} [index_name] [USING {BTREE | HASH}] (<key_part>, ...))

<index_name>: Single identifier
<key_part>: Single identifier
```

[1] : https://dev.mysql.com/doc/refman/8.0/en/create-table.html


src/ast/ddl.rs
```diff
@@ -247,6 +247,24 @@ pub enum TableConstraint {
...
         expr: Box<Expr>,
     },
+    /// MySQLs [index definition][1] for index creation. Not present on ANSI so, for now, the usage
+    /// is restricted to MySQL, as no other dialects that support this syntax were found.
+    ///
+    /// `{INDEX | KEY} [index_name] [index_type] (key_part,...) [index_option]...`
+    ///
+    /// [1]: https://dev.mysql.com/doc/refman/8.0/en/create-table.html
+    Index {
+        /// Whether this index starts with KEY (true) or INDEX (false), to maintain the same syntax.
```
alamb, line 257: 👍

alamb (approved):
Thank you @AugustoFKL

PR #667, opened 2022-10-11
insert: do not always detect parens after table name as projection

mobuchowski:
Before this PR, query

```
INSERT INTO s.table_a (SELECT col1, col2 FROM s_two.table_x WHERE condition IS TRUE AND condition_2 IS FALSE GROUP BY col2, col1)
```

fails because parser always assumes that left paren starts a projection list.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3652,7 +3652,10 @@ impl<'a> Parser<'a> {
         &mut self,
         optional: IsOptional,
     ) -> Result<Vec<Ident>, ParserError> {
-        if self.consume_token(&Token::LParen) {
+        if self.peek_token() == Token::LParen
+            && self.peek_nth_token(1) != Token::make_keyword("SELECT")
+        {
+            self.expect_token(&Token::LParen)?;
             let cols = self.parse_comma_separated(Parser::parse_identifier)?;
             self.expect_token(&Token::RParen)?;
             Ok(cols)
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -3652,7 +3652,10 @@ impl<'a> Parser<'a> {
         &mut self,
         optional: IsOptional,
     ) -> Result<Vec<Ident>, ParserError> {
-        if self.consume_token(&Token::LParen) {
+        if self.peek_token() == Token::LParen
```
AugustoFKL, line 3655: @mobuchowski could you please:
- Add an example of which dialect supports that syntax
- Make sure this doesn't break another syntax. Is using `SELECT` the best approach here? It seems too manual and prone to problems, as nested queries could have `WITH` clauses. Am I missing something?

AugustoFKL (changes requested):

src/parser.rs
alamb: Interestingly, it seems like postgres expects a `column_name` https://www.postgresql.org/docs/current/sql-insert.html

So I echo @AugustoFKL 's question about what dialect supports this construction? It is probably good to mention this in comments

In terms of not breaking other dialects, the coverage seems reasonable 

https://github.com/sqlparser-rs/sqlparser-rs/pull/667/files#diff-ae2dab2c924f4018c1daf0743b8fd912c44c5b63ae7144dd8b951bb3bd254349R64

alamb:
marking as draft as it has feedback and doesn't need a re-review yet

alamb:

PR #670, opened 2022-10-13
Support cache/uncache table syntax

### LINKED ISSUE #669: Support cache/uncache table syntax
1. Cache 

```sql
CACHE [ FLAG ] TABLE <table_name> [ OPTIONS('K1' = 'V1', 'K2' = V2) ] [ AS ] [ <query> ]
```

2. Uncache

```sql
UNCACHE TABLE [ IF EXISTS ]  <table_name>
```

Spark SQL CACHE TABLE: https://docs.databricks.com/spark/latest/spark-sql/language-manual/sql-ref-syntax-aux-cache-cache-table.html

francis-du:
Close #669

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1430,18 +1430,24 @@ pub enum Statement {
         // Specifies the actions to perform when values match or do not match.
         clauses: Vec<MergeClause>,
     },
-    /// CACHE TABLE <table_name> [[AS] [QUERY]]
+    /// CACHE [ FLAG ] TABLE <table_name> [ OPTIONS('K1' = 'V1', 'K2' = V2) ] [ AS ] [ <query> ]
+    /// Based on Spark SQL,see <https://docs.databricks.com/spark/latest/spark-sql/language-manual/sql-ref-syntax-aux-cache-cache-table.html>
     Cache {
+        // Table flag
+        table_flag: Option<ObjectName>,
         // Table name
         table_name: ObjectName,
-        alias: bool,
+        has_as: bool,
+        // Table confs
+        options: Vec<SqlOption>,
         // Cache table as a Query
         query: Option<Query>,
     },
-    /// UNCACHE TABLE <table_name>
+    /// UNCACHE TABLE [ IF EXISTS ]  <table_name>
     UNCache {
         // Table name
         table_name: ObjectName,
+        if_exists: bool,
     },
 }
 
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -1430,6 +1430,19 @@ pub enum Statement {
         // Specifies the actions to perform when values match or do not match.
         clauses: Vec<MergeClause>,
     },
+    /// CACHE TABLE <table_name> [[AS] [QUERY]]
```
alamb, line 1433: What database is this meant to follow? It doesn't seem to follow Spark's syntax

https://docs.databricks.com/spark/latest/spark-sql/language-manual/sql-ref-syntax-aux-cache-cache-table.html

How is it different than a `CREATE VIEW` statement?

tests/sqlparser_common.rs
```diff
@@ -3239,14 +3240,14 @@ fn parse_at_timezone() {
...
-                    quote_style: None
+                    quote_style: None,
                 }]),
                 args: vec![FunctionArg::Unnamed(FunctionArgExpr::Expr(zero.clone()))],
                 over: None,
                 distinct: false,
                 special: false,
             })),
-            time_zone: "UTC-06:00".to_string()
+            time_zone: "UTC-06:00".to_string(),
```
alamb, line 3250: I don't mind the `,` everywhere but it sure makes the diff a lot bigger ;)

francis-du: I think this one is from the latest` cargo fmt`

src/ast/mod.rs
francis-du: Thanks for the doc, I will support the full Cache table syntax.

The difference between `CACHE TABLE` and `CREATE VIEW` is that users can cache small tables in memory before Join to speed up queries, such as dimension tables in data warehouse.The View should not be able to do this if it is not a memory table.

francis-du:
@alamb Hi Andrew, could you review this PR again, if you have time. I have supported the full syntax.

tests/sqlparser_common.rs
```diff
@@ -5810,3 +5813,249 @@ fn parse_show_functions() {
...
+            table_name: ObjectName(vec![Ident::with_quote('\'', cache_table_name)]),
+            has_as: true,
+            options: vec![
+                SqlOption {
+                    name: Ident::with_quote('\'', "K1"),
+                    value: Value::SingleQuotedString("V1".into()),
+                },
+                SqlOption {
+                    name: Ident::with_quote('\'', "K2"),
+                    value: number("0.88"),
```
alamb, line 5931: 👍

tests/sqlparser_common.rs
```diff
@@ -5810,3 +5813,249 @@ fn parse_show_functions() {
         }
     );
 }
+
+#[test]
+fn parse_cache_table() {
```
alamb, line 5818: giphy
👍

alamb (approved):
Looks good to me -- thanks @francis-du !

PR #673, opened 2022-10-14
Support drop sequence statement

sam-mmm:
Drop for sequence `DROP SEQUENCE [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]`


alamb (approved):
Looks great -- thanks @sam-mmm

PR #677, opened 2022-10-18
Add scala function `LOCATE` in MySQL

zmrdltl:
ref: [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_locate)
[examples](https://www.w3schools.com/sql/func_mysql_locate.asp)

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -346,7 +346,7 @@ pub enum Expr {
     },
     /// POSITION(<expr> in <expr>)
     Position { expr: Box<Expr>, r#in: Box<Expr> },
-    /// LOCATE(<expr>, <expr>)
+    /// LOCATE(<expr>, <expr>, <expr>)
     Locate {
         find_expr: Box<Expr>,
         from_expr: Box<Expr>,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -346,6 +346,12 @@ pub enum Expr {
     },
     /// POSITION(<expr> in <expr>)
     Position { expr: Box<Expr>, r#in: Box<Expr> },
+    /// LOCATE(<expr>, <expr>, <expr>)
```
alamb, line 349: Wouldn't the parser parse `LOCATE("foo", "foobar")` already today?

The `POSITION` has some special syntax (`IN`) but `LOCATE` looks like a normal function 

So I am wondering why this PR is needed

alamb (commented):
Thank you for the contribution @zmrdltl

src/ast/mod.rs
zmrdltl: I confirmed that it is parsed as a function because it does not need such a special keyword and is separated only by commas. As you said, the PR seems unnecessary. I will close this PR.

PR #680, opened 2022-10-20
Modifier support for Custom Datatype

sunng87:
In PostgreSQL/PostGIS, when creating tables with `geometry` or `geography` data type, additional modifiers are allowed. In [this example](https://postgis.net/docs/manual-3.3/using_postgis_dbmanagement.html#Create_Geography_Tables), we can specify `geography(POINT, 4269)` as data type, which indicates the column is geography point typed, and uses 4269 as its spatial reference system.

Back to sqlparser, at the moment we don't support `Custom` data type that comes with additional modifiers/arguments in form of `(..)`. This patch add support for that. Since these data can be in any form like an identifier, a number or a string, here I convert all of them into `String`. Suggestions are welcomed.

Edit: In PostGIS this is called type modifiers.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -130,6 +130,8 @@ pub enum DataType {
     Bytea,
     /// Custom type such as enums
     Custom(ObjectName),
+    /// Custom type with arguments
+    CustomWithArgs(ObjectName, Vec<String>),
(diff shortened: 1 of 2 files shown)


alamb (approved):
LGTM -- thanks @sunng87

PR #682, opened 2022-10-21
Support `ALTER TABLE DROP PRIMARY KEY`

### LINKED ISSUE #591: MySQL "ALTER ... DROP PRIMARY KEY" is not parsed correctly
The following alter statement to drop primary key directly is allowed in MySQL:

```SQL
ALTER TABLE table_name DROP PRIMARY KEY;
```

However, it's not parsed correctly in sqlparser, for example:

```Rust
use sqlparser::dialect::MySqlDialect;
use sqlparser::tokenizer::Token;
use sqlparser::tokenizer::Tokenizer;

#[test]
fn test() {
    let dialect = &MySqlDialect {};
    let sql = r"ALTER TABLE my_table DROP PRIMARY KEY;";
    let mut tokenizer = Tokenizer::new(dialect, sql);
    let tokens: Vec<Token> = tokenizer.tokenize().unwrap();
    let mut parser = Parser::new(tokens, dialect);
    parser.next_token();
    let statement = parser.parse_alter();
    println!("{:?}", statement);
}
```

Which prints 

Ok(AlterTable { name: ObjectName([Ident { value: "my_table", quote_style: None }]), operation: DropColumn { column_name: Ident { value: "PRIMARY", quote_style: None }, if_exists: false, cascade: false } })

Looks like the "drop primary key" is interpreted as "drop column 'PRIMARY'".

ding-young:
## Issue
resolves #591

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3136,8 +3136,9 @@ impl<'a> Parser<'a> {
                     name,
                     cascade,
                 }
-            } else if self.parse_keyword(Keyword::PRIMARY) {
-                let _ = self.parse_keyword(Keyword::KEY);
+            } else if self.parse_keywords(&[Keyword::PRIMARY, Keyword::KEY])
+                && dialect_of!(self is MySqlDialect | GenericDialect)
+            {
                 AlterTableOperation::DropPrimaryKey
             } else {
                 let _ = self.parse_keyword(Keyword::COLUMN);
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3136,6 +3136,9 @@ impl<'a> Parser<'a> {
                     name,
                     cascade,
                 }
+            } else if self.parse_keyword(Keyword::PRIMARY) {
+                let _ = self.parse_keyword(Keyword::KEY);
+                AlterTableOperation::DropPrimaryKey
             } else {
                 let _ = self.parse_keyword(Keyword::COLUMN);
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -3136,6 +3136,9 @@ impl<'a> Parser<'a> {
                     name,
                     cascade,
                 }
+            } else if self.parse_keyword(Keyword::PRIMARY) {
```
AugustoFKL, line 3139: @ding-young `ALTER TABLE tb DROP PRIMARY` isn't allowed and, therefore, shouldn't be acceptable. 

The `let _` in the next line is permissive over this.

Also, I think this is a MySQL-specific statement. Could you please add the validation to accept only `MySQL` and `GenericDialect`, please?

AugustoFKL (changes requested):

alamb (commented):
@ding-young  do you have time to implement @AugustoFKL 's suggestion? Otherwise this PR  looks good to me

ding-young:
> @ding-young  do you have time to implement @AugustoFKL 's suggestion? Otherwise this PR  looks good to me

Yes, I'll work on it soon.

ding-young:
@AugustoFKL Thank you for feedback. I changed the code 😃

AugustoFKL (approved):
LGTM

@alamb ready for approval :)

alamb (approved):

alamb:
Thanks @ding-young  and @AugustoFKL !

PR #685, opened 2022-10-22
Support `CREATE  TABLE  ON UPDATE <expr>`  Function

### LINKED ISSUE #684: Support OnUpdate Function
## Description

In MySQl, it supports ON UPDATE CURRENT_TIMESTAMP Function when creating table and declaring DATETIME column. But I wonder if there is non-Token::LParen option in Keyword::CURRENT_TIMESTAMP. "CURRENT_TIMESTAMP()" in Following Task passed but "CURRENT_TIMESTAMP" is not working

## Task

```rust
#[test]
fn parse_table_colum_option_on_update() {
    let sql1 = "CREATE TABLE foo (`modification_time` DATETIME ON UPDATE CURRENT_TIMESTAMP())";
    match mysql().verified_stmt(sql1) {
        Statement::CreateTable { name, columns, .. } => {
            assert_eq!(name.to_string(), "foo");
            assert_eq!(
                vec![ColumnDef {
                    name: Ident::with_quote('`', "modification_time"),
                    data_type: DataType::Datetime,
                    collation: None,
                    options: vec![ColumnOptionDef {
                        name: None,
                        option: ColumnOption::OnUpdate(Expr::Function(Function {
                            name: ObjectName(vec![Ident::new("CURRENT_TIMESTAMP")]),
                            args: vec![],
                            over: None,
                            distinct: false,
                            special: false,
                        })),
                    },],
                }],
                columns
            );
        }
        _ => unreachable!(),
    }
}
```

CEOJINSUNG:
Resolve #684


src/parser.rs
```diff
@@ -2921,9 +2921,8 @@ impl<'a> Parser<'a> {
         } else if self.parse_keywords(&[Keyword::ON, Keyword::UPDATE])
             && dialect_of!(self is MySqlDialect)
```
AugustoFKL, line 2922: @CEOJINSUNG please add the `GenericDialect` here.

src/ast/ddl.rs
```diff
@@ -437,6 +437,7 @@ pub enum ColumnOption {
     DialectSpecific(Vec<Token>),
     CharacterSet(ObjectName),
     Comment(String),
+    OnUpdate(Expr),
```
AugustoFKL, line 440: Why does it take an expression?

`ON UPDATE CURRENT TIMESTAMP` is an expression by itself (1). You can take `TIMESTAMP [(p)]` for the parameter, as long as I'm aware of it.

There's no reason to make it complex to the upstream with expressions, as there's no `ON UPDATE column_1`, for example.

You could make it `OnUpdateCurrentTimestamp(Option<u64>)` to get the precision, maybe?

I don't think there are things like `ON UPDATE CURRENT TIMESTAMP WITH TIMEZONE` exist at all, so only having the precision should be enough.

AugustoFKL (changes requested):

src/ast/ddl.rs
alamb: I tried this with mysql and the syntax is accepted

```sql
mysql> CREATE TABLE foo (`modification_time` DATETIME ON UPDATE CURRENT_TIMESTAMP());
Query OK, 0 rows affected (0.01 sec)

mysql> show create table foo;
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                    |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| foo   | CREATE TABLE `foo` (
  `modification_time` datetime DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

Which is very strange to me as the documentation doesn't seem to support that syntax 🤔 
 https://dev.mysql.com/doc/refman/8.0/en/create-table.html

```

reference_definition:
    REFERENCES tbl_name (key_part,...)
      [MATCH FULL | MATCH PARTIAL | MATCH SIMPLE]
      [ON DELETE reference_option]
      [ON UPDATE reference_option]


reference_option:
    RESTRICT | CASCADE | SET NULL | NO ACTION | SET DEFAULT
```

alamb (approved):
I think this PR is ok to go -- thank you @CEOJINSUNG 

If possible, can you please answer @AugustoFKL 's question?

src/ast/ddl.rs
AugustoFKL: @alamb this is a specific syntax from MySQL [(1)](https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html)

The problem is, the documentation states that the `ON UPDATE CURRENT_TIMESTAMP` is a special expression uniquely. The only possible modification seems to be the precision information `ON UPDATE CURRENT_TIMESTAMP(n)`, which is dependant on the data type for that column.

But this PR seems to expect any expression, which doesn't make sense, and it makes the upstream handle all other possible expressions as errors, which we should do.

[1] : https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

src/parser.rs
CEOJINSUNG: Thank you for advising good comment

src/ast/ddl.rs
CEOJINSUNG: I understand "expression" issue you are raising. `OnUpdateCurrentTimestamp(Option<u64>)` is also good proposal which I was thinking about it. But I am confused in "CurrentTimestamp". 

- https://www.w3schools.com/mysql/func_mysql_current_timestamp.asp
- https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

In this document, CurrentTimestamp belongs to Function and MySQL make it as special expression without using parentheses. Maybe it would be good to make reservation expression function options which has no parentheses and make users write both CURRENT_TIMESTAMP and CURRENTP_TIMESTAMP()

AugustoFKL: @CEOJINSUNG I kinda agree, but the function and the expression are not the same.

Either way, maybe you can have an enum there?

Like

```rust
pub enum OnUpdateCurrentTimestampInfo {
  None,
  Parenthesis,
  ParenthesisAndPrecision
}
```

Other approach would be something like:

```rust
OnUpdateCurrentTimestamp(Option<Option<u64>>)
```

Where the internal option is the precision, and the external the presence of parenthesis. But seems less idiomatic and a little obscure for me.

alamb (approved):
I think this looks good to me and the rationale for a more permissive Expr  makes sense -- @AugustoFKL  are you ok with it?

AugustoFKL (commented):
@alamb actually no. I'm still not convinced about the expression usage.

Would be TOO permissive, making anyone in the upstream that would use it really confused.

src/ast/ddl.rs
CEOJINSUNG: @AugustoFKL Thanks for the suggestion But in sqlparse-rs structure, Function belongs to Expr. And I think it is not good structure because there is no scalability. OnUpdateCurrentTimestampInfo looks like only using for MySQL. 

There are another functions without parenthesis like DIV, SYSDATE Function.

```
SELECT 10 DIV 5; // MySQL
SELECT SYSDATE FROM DUAL; // ORACLE
```

An extensible structure that can be freely added to functions used without parentheses such as DIV and SYSDATE is required like below NoArgsFunction.

```rust
pub enum NoArgsFunction {
    CurrentTimpeStamp,
    DIV,
    SYSDATE
}
```

Since SQL is used in various places such as MySQL, Oracle, PosgreSQL, etc., OnupdateCurrentTimeStamp does not seem to be a very good structure even for the open-close principle. Therefore, I think it would be better to respond so that elements used without parentheses can be freely added to functions without arguments.

AugustoFKL: @CEOJINSUNG but can you say that this is a function per se?

To me it looks like a really special case.

alamb:
Marking as a draft until we get a resolution

alamb:
Given there there are several proposals for this same feature:
https://github.com/sqlparser-rs/sqlparser-rs/pull/586 from @Sibz (apparently MySQL also supports `now()` in addition to `current_timestamp`
https://github.com/sqlparser-rs/sqlparser-rs/pull/602 from @RainJoe

> Would be TOO permissive, making anyone in the upstream that would use it really confused.

While this PR is more general than MySQL would allow I think the distinction of what functions to support in this area should be left to downstream implementations as described in https://github.com/sqlparser-rs/sqlparser-rs#extensible-sql-lexer-and-parser-for-rust

> This crate avoids semantic analysis because it varies drastically between dialects and implementations. If you want to do semantic analysis, feel free to use this project as a base

Thus I plan to merge this PR in

alamb (approved):

PR #686, opened 2022-10-23
Change quote style to enum

### LINKED ISSUE #526: Make Ident enum
Ident now is a struct with a quote style:
```
pub struct Ident {
    pub value: String,
    pub quote_style: Option<char>,
```
Now it is a bit risky, because if the char isn't the expected it panics:
```
       assert!(quote == '\'' || quote == '"' || quote == '`' || quote == '[');
```
Now it is true that it shouldn't never get to it, because when we tokenize we are only allowing those to be the quote style, but it is still risk for panics.

I suggest to move Ident to be an enum:
```
pub enum Ident {
    Literal(String),
    SingleQoute(String),
    DoubleQoute(String)
    ...
}
```
Now the Display will be:
```
impl fmt::Display for Ident {
        match self {
            Literal(value) =>  write!(f, "{}", value),
            SingleQoute(value) => write!(f, "'{}'", value),
            ...
    }
}
```
And we can also implement a method `get_value`:
```
impl Ident {
    fn get_value(&self) -> String {
       match self {
           Literal(value)  |  SingleQoute(value)  ... => return value.clone()
    }
}
```

If it is acceptable I will add a PR with the code

omer-shtivi:
Closes #526

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -89,7 +89,7 @@ pub struct Ident {
     pub value: String,
     /// The starting quote if any. Valid quote characters are the single quote,
     /// double quote, backtick, and opening square bracket.
-    pub quote_style: Option<char>,
+    pub quote_style: QuoteStyle,
 }
 
 impl Ident {
@@ -100,20 +100,19 @@ impl Ident {
     {
         Ident {
             value: value.into(),
-            quote_style: None,
+            quote_style: QuoteStyle::None,
         }
     }
 
     /// Create a new quoted identifier with the given quote and value. This function
     /// panics if the given quote is not a valid quote character.
-    pub fn with_quote<S>(quote: char, value: S) -> Self
+    pub fn with_quote<S>(quote_style: QuoteStyle, value: S) -> Self
     where
         S: Into<String>,
     {
-        assert!(quote == '\'' || quote == '"' || quote == '`' || quote == '[');
         Ident {
             value: value.into(),
-            quote_style: Some(quote),
+            quote_style,
         }
     }
 }
@@ -122,7 +121,7 @@ impl From<&str> for Ident {
     fn from(value: &str) -> Self {
         Ident {
             value: value.to_string(),
-            quote_style: None,
+            quote_style: QuoteStyle::None,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


src/ast/mod.rs
```diff
@@ -122,21 +121,59 @@ impl From<&str> for Ident {
...
+}
+
+impl QuoteStyle {
+    fn as_char(&self) -> char {
+        match self {
+            QuoteStyle::SingleQuote => '\'',
+            QuoteStyle::DoubleQuote => '"',
+            QuoteStyle::BackTick => '`',
+            QuoteStyle::SquareBracket | QuoteStyle::None => {
+                panic!("Not supported for quote style {}", self)
```
AugustoFKL, line 175: @omer-shtivi please make this an error. Panicking here isn't nice at all.

src/ast/mod.rs
```diff
@@ -122,21 +121,59 @@ impl From<&str> for Ident {
...
 
 impl fmt::Display for Ident {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self.quote_style {
-            Some(q) if q == '"' || q == '\'' || q == '`' => {
-                let escaped = value::escape_quoted_string(&self.value, q);
-                write!(f, "{}{}{}", q, escaped, q)
+            QuoteStyle::None => f.write_str(&self.value),
+            QuoteStyle::SingleQuote | QuoteStyle::DoubleQuote | QuoteStyle::BackTick => {
+                let escaped = value::escape_quoted_string(&self.value, self.quote_style.as_char());
```
AugustoFKL, line 134: @omer-shtivi I don't get why you need to escape the value to print it.

Can't you just access the self.value? This seems to turn the code more complex to upstream projects

src/ast/mod.rs
```diff
@@ -122,21 +121,59 @@ impl From<&str> for Ident {
...
+    /// No quote
+    None,
+    /// Quote with `'`
+    SingleQuote,
+    /// Quote with `"`
+    DoubleQuote,
+    /// Quote with ``` (BackTick)
+    BackTick,
+    /// Quote with `[]` (Square bracket)
+    SquareBracket,
```
AugustoFKL, line 155: Never saw this, please add a reference to the comment about dialect(s) that use it.

AugustoFKL (changes requested):
@alamb what are your thoughts on this?

TBH I have mixed feelings. It seems to be too lengthy and a little confusing (mainly the square bracket and none printing) for such a slight gain.

src/ast/mod.rs
omer-shtivi: @AugustoFKL can't because idents like 
```
"SELECT ```quoted identifier```"
```
will fail
please see the following test:
https://github.com/omer-shtivi/sqlparser-rs/blob/914810d36617581c2380ae9c0f501116a7d810b6/tests/sqlparser_mysql.rs#L476-L477

src/ast/mod.rs
omer-shtivi: Done

src/ast/mod.rs
omer-shtivi: Done

AugustoFKL (approved):
LGTM

@alamb what do you think?

src/ast/mod.rs
AugustoFKL: Never saw this syntax, really interesting.

One question. If the column name is `c1` (with the quotes, same as your example), why escape it?

omer-shtivi: Because the column name might be like this:
```
"my_column_""_name"
```
or
```
"my_column_\"_name"
```

AugustoFKL (approved):
LGTM I guess

src/tokenizer.rs
```diff
@@ -258,11 +258,11 @@ pub struct Word {
 impl fmt::Display for Word {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self.quote_style {
-            Some(s) if s == '"' || s == '[' || s == '`' => {
-                write!(f, "{}{}{}", s, self.value, Word::matching_end_quote(s))
+            QuoteStyle::None => write!(f, "{}", self.value),
+            QuoteStyle::SingleQuote | QuoteStyle::DoubleQuote | QuoteStyle::BackTick => {
```
alamb, line 262: I wonder if it would make sense to put the logic for printing quotes into `QuoteStyle` rather than partially there are partially here. It would make the `""` returned by `SquareBrack` easier to understand I think

omer-shtivi:
Thanks @alamb for the feedback, I'm learning from it (:
I change quote style to return a Display object, I hope it what you had in mind, if no let me know when I went wrong.
This is just the general idea, if it is indeed the way forward I will change all of those:
``` rust
             QuoteStyle::SingleQuote => {
                let escaped = value::escape_quoted_string(&self.value, '\'');
                let quoted = self.quote_style.quote(escaped.to_string());
                write!(f, "{}", quoted)
            }
            QuoteStyle::DoubleQuote => {
                let escaped = value::escape_quoted_string(&self.value, '"');
                let quoted = self.quote_style.quote(escaped.to_string());
                write!(f, "{}", quoted)
            }
            QuoteStyle::BackTick => {
                let escaped = value::escape_quoted_string(&self.value, '`');
                let quoted = self.quote_style.quote(escaped.to_string());
                write!(f, "{}", quoted)
            }
```
to an helper function

src/tokenizer.rs
```diff
@@ -476,7 +470,18 @@ impl<'a> Tokenizer<'a> {
...
 
                     if last_char == Some(quote_end) {
-                        Ok(Some(Token::make_word(&s, Some(quote_start))))
+                        let quote_style = match quote_start {
+                            '\'' => QuoteStyle::SingleQuote,
+                            '[' => QuoteStyle::SquareBracket,
+                            '"' => QuoteStyle::DoubleQuote,
+                            '`' => QuoteStyle::BackTick,
+                            _ => {
+                                return self.tokenizer_error(format!(
```
alamb, line 479: 👍

src/ast/mod.rs
```diff
@@ -122,25 +121,69 @@ impl From<&str> for Ident {
...
+    DoubleQuote,
+    /// Quote with ``` (BackTick)
+    /// Used by MySql and BigQuery
+    BackTick,
+    /// Quote with `[]` (Square bracket)
+    /// Used by Redshift and SqLite
+    SquareBracket,
+}
+impl QuoteStyle {
+    pub fn quote(self, value: String) -> impl Display {
```
alamb, line 173: 👍

alamb (approved):
Looks great -- thank you @omer-shtivi  and @AugustoFKL

alamb:
This PR appears to have some issues in tests now

omer-shtivi:
Sorry for that, fixed.

alamb:
Marking as draft as the CI is still failing -- please mark as ready to review when the conflicts are resolved and the CI is ready

alamb:

PR #689, opened 2022-10-26
Support `DISTINCT` for SetOperator

### LINKED ISSUE #688: Support `DISTINCT` for SetOperator
I’d like to parse `UNION DISTINCT` using BigQuery and MySQL dialect.
This is not included in standard.

BigQuery Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax?hl=ja#set_operators
MySQL Reference: https://dev.mysql.com/doc/refman/8.0/en/union.html

I'm not familiar with other dialects, if other dialects allow `UNION DISTINCT`, please comment.

unvalley:
closes #688

This supports following  for GenericDialect ~BigQuery and MySql dialects~.
- [x] UNION DISTINCT
- [x] INTERSECT DISTINCT
- [x] EXCEPT DISTINCT

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -67,18 +67,18 @@ impl fmt::Display for Query {
 /// Options for SetOperator
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum SetQualifier {
+pub enum SetQuantifier {
     All,
     Distinct,
     None,
 }
 
-impl fmt::Display for SetQualifier {
+impl fmt::Display for SetQuantifier {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            SetQualifier::All => write!(f, "ALL"),
-            SetQualifier::Distinct => write!(f, "DISTINCT"),
-            SetQualifier::None => write!(f, ""),
+            SetQuantifier::All => write!(f, "ALL"),
+            SetQuantifier::Distinct => write!(f, "DISTINCT"),
+            SetQuantifier::None => write!(f, ""),
         }
     }
 }
@@ -97,7 +97,7 @@ pub enum SetExpr {
     /// UNION/EXCEPT/INTERSECT of two queries
     SetOperation {
         op: SetOperator,
-        set_qualifier: SetQualifier,
+        set_quantifier: SetQuantifier,
         left: Box<SetExpr>,
         right: Box<SetExpr>,
     },
@@ -117,12 +117,12 @@ impl fmt::Display for SetExpr {
                 left,
                 right,
                 op,
-                set_qualifier,
+                set_quantifier,
             } => {
                 write!(f, "{} {}", left, op)?;
-                match set_qualifier {
-                    SetQualifier::All | SetQualifier::Distinct => write!(f, " {}", set_qualifier)?,
-                    SetQualifier::None => write!(f, "{}", set_qualifier)?,
+                match set_quantifier {
+                    SetQuantifier::All | SetQuantifier::Distinct => write!(f, " {}", set_quantifier)?,
+                    SetQuantifier::None => write!(f, "{}", set_quantifier)?,
                 }
                 write!(f, " {}", right)?;
                 Ok(())
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -121,7 +121,9 @@ impl fmt::Display for SetExpr {
             } => {
                 write!(f, "{} {}", left, op)?;
                 match set_quantifier {
-                    SetQuantifier::All | SetQuantifier::Distinct => write!(f, " {}", set_quantifier)?,
+                    SetQuantifier::All | SetQuantifier::Distinct => {
+                        write!(f, " {}", set_quantifier)?
+                    }
                     SetQuantifier::None => write!(f, "{}", set_quantifier)?,
                 }
                 write!(f, " {}", right)?;
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -133,7 +133,7 @@ impl fmt::Display for SetOperator {
 }
 
 /// A quantifier for [SetOperator].
-// TODO: Restrict some dialects to parse, specific SetQuantifier.
+// TODO: Restrict parsing specific SetQuantifier in some specific dialects.
 // For example, BigQuery does not support `DISTINCT` for `EXCEPT` and `INTERSECT`
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -64,25 +64,6 @@ impl fmt::Display for Query {
     }
 }
 
-/// Options for SetOperator
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum SetQuantifier {
-    All,
-    Distinct,
-    None,
-}
-
-impl fmt::Display for SetQuantifier {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            SetQuantifier::All => write!(f, "ALL"),
-            SetQuantifier::Distinct => write!(f, "DISTINCT"),
-            SetQuantifier::None => write!(f, ""),
-        }
-    }
-}
-
 /// A node in a tree, representing a "query body" expression, roughly:
 /// `SELECT ... [ {UNION|EXCEPT|INTERSECT} SELECT ...]`
 #[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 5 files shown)


tests/sqlparser_common.rs
```diff
@@ -4096,6 +4096,16 @@ fn parse_union() {
...
 }
 
+#[test]
+fn parse_union_distinct() {
+    let dialects = TestedDialects {
+        dialects: vec![Box::new(BigQueryDialect {}), Box::new(MySqlDialect {})],
+    };
+    dialects.verified_stmt("SELECT 1 UNION DISTINCT SELECT 2");
+    dialects.verified_stmt("SELECT 1 UNION DISTINCT SELECT 2 INTERSECT ALL SELECT 3");
+    dialects.verified_stmt("SELECT 1 EXCEPT (SELECT 2 UNION DISTINCT SELECT 3)");
```
unvalley, line 4106: If test cases aren't enough, please let me know.

src/ast/query.rs
```diff
@@ -78,7 +95,7 @@ pub enum SetExpr {
     /// UNION/EXCEPT/INTERSECT of two queries
     SetOperation {
         op: SetOperator,
-        all: bool,
+        op_option: Option<SetOperatorOption>,
```
unvalley, line 98: This change make users change their code to use `Option<SetOperatorOption>`.

src/ast/query.rs
```diff
@@ -64,6 +64,23 @@ impl fmt::Display for Query {
     }
 }
 
+/// Options for SetOperator
```
AugustoFKL, line 67: Would be nice having more description about the structure. Maybe a reference of which big dialects use it...

src/ast/query.rs
AugustoFKL: @unvalley I'd name it set_quantifier, as it is the name used in ANSI [(1)](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#set-quantifier).

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#set-quantifier

src/ast/query.rs
```diff
@@ -64,6 +64,23 @@ impl fmt::Display for Query {
...
+pub enum SetOperatorOption {
+    All,
+    Distinct,
+}
+
+impl fmt::Display for SetOperatorOption {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            SetOperatorOption::All => write!(f, " ALL"),
+            SetOperatorOption::Distinct => write!(f, " DISTINCT"),
```
AugustoFKL, line 79: Printing spaces inside the enum structure doesn't make much sense.

If there's a space or not, the structure that holds it should handle it (in this case, the SetExpr).

src/parser.rs
```diff
@@ -4193,6 +4194,40 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_set_operator_option(
```
AugustoFKL, line 4197: I'd be cautious about 2 things:
1. Are you really sure that the `ALL` and `DISTINCT` change between operations? 
2. You're being restrictive here. Please add `GenericDialect` and be aware of one thing: unless you're REALLY sure that only those 2 dialects support that feature, please be permissive and do not validate dialects.

I found that at least `ANSI` supports this. Please, or be permissive or validate perfectly (which is really hard)

AugustoFKL (changes requested):

src/ast/query.rs
alamb: I agree set_quantifier sounds good (along with `SetOperatorOption` --> `SetQuantifier`)

src/ast/query.rs
```diff
@@ -64,6 +64,23 @@ impl fmt::Display for Query {
     }
 }
 
+/// Options for SetOperator
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum SetOperatorOption {
+    All,
```
alamb, line 71: In general it should be possible to round trip structures in sqlparser -- so it should be possible to distinguish between `ALL`, `DISTINCT` or nothing was specified in the query

src/parser.rs
alamb: I think ALL / DISTINCT are ANSI SQL and thus should be handled by all dialects (unless we have specific evidence that some dialects don't support them there is no reason to prevent them being parsed)

alamb (commented):
Thank you @unvalley  -- this is looking pretty close1

src/parser.rs
unvalley: Thank you for commenting.

> Are you really sure that the ALL and DISTINCT change between operations?
sorry, could you elaborate this? 

> You're being restrictive here. Please add GenericDialect and be aware of one thing: unless you're REALLY sure that only those 2 dialects support that feature, please be permissive and do not validate dialects.

> I think ALL / DISTINCT are ANSI SQL and thus should be handled by all dialects (unless we have specific evidence that some dialects don't support them there is no reason to prevent them being parsed)

OK, I'll remove the specific dialects handling (bq, mysql) and make it parse by GenericDialect.

src/ast/query.rs
unvalley: I'll add the `None` to the enum `SetOperatorOption`, and remove `Option<T>` for `SetOperatorOption`.

src/parser.rs
AugustoFKL: @unvalley you're parsing them differently. `Union` is being parsed with both `ALL` and `DISTINCT`, while `Except` and `Intersect` not.

Is there any reason why you can't do something like:
```
match op {
            Some(SetOperator::Union) => {
                    Some(SetOperatorOption::All)
                } else if self.parse_keyword(Keyword::DISTINCT)
                    && dialect_of!(self is MySqlDialect | BigQueryDialect)
                {
                    Some(SetOperatorOption::Distinct)
                } else {
                    None
                }
            }
            Some(SetOperator::Except) | Some(SetOperator::Intersect) => {
                    Some(SetOperatorOption::All)
                } else {
                    None
                }
            }
            _ => None,
        }
```

Also, FYI (not saying you must do it in this PR, besides the second one):
- ANSI supports both [(1)](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#query-expression-body). 
- The `GenericDialect` should be added to parse it as well. Unless the syntax is REALLY different between dialects, I'd say to add every syntax to the generic because, as the name states, it tries to be as generic as possible.
- MySQL also accepts it in `INTERSECT` [(2)](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#query-expression-body) and `EXCEPT` [(3)](https://dev.mysql.com/doc/refman/8.0/en/except.html).
- BigQuery, following your implementation, support both only for `UNION` [(4)](https://dev.mysql.com/doc/refman/8.0/en/except.html)

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#query-expression-body
[2] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#query-expression-body
[3] : https://dev.mysql.com/doc/refman/8.0/en/except.html
[4] : https://dev.mysql.com/doc/refman/8.0/en/except.html

unvalley: Thanks, I got it. I should implement `ALL` and `DISTINCT` for `INTERSECT` and `EXCEPT` as I do for `UNION`. And, I should focus mainly on `GenericDialect`.
In this pr, I'll only implement the [1] and [2] for `Generic Dialect`.

I thought the BigQuery dialect should not allow `ALL` to be attached to `INTERSECT` and `EXCEPT`, but I see that is not correct. Because ANSI supports it, there is no trouble with parsing being done. And it is difficult to test that it cannot be parsed as you said.

AugustoFKL: @unvalley one option was to parse it conditionally for all of them, the same you do for `UNION`. But IMHO I think this can be scoped in another PR.

Your PR is to support `UNION DISTINCT` for both dialects and, parsing with `Generic`, you can have it.

If you want the EXACT same support for a dialect (being it permissive or restrictive), I'd say to do it separately, working from a generic scope to a restrict one. Trying to do all at once make the code confusing, and requires reading all documentations at once.

src/ast/query.rs
unvalley: I fiexed at [f6e8305](https://github.com/sqlparser-rs/sqlparser-rs/pull/689/commits/f6e83058b0b5a37de5d2c9b0ce473186a1294ed4)

src/ast/query.rs
unvalley: I fixed this at [f6e8305](https://github.com/sqlparser-rs/sqlparser-rs/pull/689/commits/f6e83058b0b5a37de5d2c9b0ce473186a1294ed4)

src/ast/query.rs
unvalley: I fixed this at [ac41a8e](https://github.com/sqlparser-rs/sqlparser-rs/pull/689/commits/ac41a8e839e1a058c9957d6676ffa269cfd82366)

src/ast/query.rs
```diff
@@ -78,7 +97,7 @@ pub enum SetExpr {
     /// UNION/EXCEPT/INTERSECT of two queries
     SetOperation {
         op: SetOperator,
-        all: bool,
+        set_quantifier: SetQuantifier,
```
unvalley, line 100: Should I rename `set_quantifier` `qu`(or something)?

src/ast/query.rs
unvalley: Okay, I added some comments at [ae6fb81](https://github.com/sqlparser-rs/sqlparser-rs/pull/689/commits/ae6fb81c13e74ae8240016c0203a146462b9ba42) .
But, I wasn't quite sure how to describe this enum...

src/ast/query.rs
```diff
@@ -125,6 +132,26 @@ impl fmt::Display for SetOperator {
     }
 }
 
+/// A quantifier for [SetOperator].
+// TODO: Restrict parsing specific SetQuantifier in some specific dialects.
+// For example, BigQuery does not support `DISTINCT` for `EXCEPT` and `INTERSECT`
```
unvalley, line 137: Okay, I added some comments at [ae6fb81](https://github.com/sqlparser-rs/sqlparser-rs/pull/689/commits/ae6fb81c13e74ae8240016c0203a146462b9ba42) .
But, I wasn't quite sure how to describe this enum🥲

unvalley:
okay, sorry for notification.. I didn't know I can set one reviewer.

alamb:
> sorry for the notification.. I didn't know I can set one reviewer only.

No worries -- without these notifications it takes me a while to find PRs to review!

alamb (approved):
I think this looks great @unvalley  -- thank you!

Any concerns @AugustoFKL ?

AugustoFKL (approved):
LGTM

alamb:
Thanks again @unvalley !

PR #691, opened 2022-10-26
Support `DATE` keyword

sarahyurick:
Hi, I'd like to be able to parse something like `SELECT EXTRACT(DATE FROM d)`, where d is a timestamp and the expected output would be something like YYYY/MM/DD.


tests/sqlparser_common.rs
```diff
@@ -1755,6 +1755,7 @@ fn parse_extract() {
     verified_stmt("SELECT EXTRACT(MONTH FROM d)");
     verified_stmt("SELECT EXTRACT(WEEK FROM d)");
     verified_stmt("SELECT EXTRACT(DAY FROM d)");
+    verified_stmt("SELECT EXTRACT(DATE FROM d)");
```
AugustoFKL, line 1758: @sarahyurick could you please add a reference from which dialect you're expecting this expansion?

The ANSI extract expression doesn't take `DATE` [(1)](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#extract-field), so I assume this is dialect-specific behavior. I'm not saying that you must turn this feature dialect-specific, but a source would be nice.

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#extract-field

AugustoFKL (changes requested):

tests/sqlparser_common.rs
sarahyurick: Hi, we wanted to add support for the query in [Dask-SQL](https://github.com/dask-contrib/dask-sql).

The closest outside example with a `DATE` keyword was here: (https://hevodata.com/learn/postgresql-convert-datetime-to-date/#:~:text=One%20other%20way%20using%20which,now()%20function)%20to%20date, although it's not paired with the `EXTRACT` expression.

alamb (approved):

alamb:
Thanks @sarahyurick  and @AugustoFKL !

PR #693, opened 2022-10-27
Snowflake: Support semi-structured data

yuval-illumex:
Per snowflake [documentation](https://docs.snowflake.com/en/user-guide/querying-semistructured.html#traversing-semi-structured-data), Colon can be used to query semi-structured data (JSON).

I decided to use the `JsonAccess` expression (which is used by postgres), but if you think a new expression is needed - I will change.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1429,7 +1429,7 @@ impl<'a> Parser<'a> {
         } else if Token::Colon == tok {
             Ok(Expr::JsonAccess {
                 left: Box::new(expr),
-                operator: JsonOperator::LongArrow,
+                operator: JsonOperator::Colon,
                 right: Box::new(self.parse_subexpr(51)?),
             })
         } else if Token::Arrow == tok
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3448,7 +3448,8 @@ impl<'a> Parser<'a> {
                     Some('\'') => Ok(Value::SingleQuotedString(w.value)),
                     _ => self.expected("A value?", Token::Word(w))?,
                 },
-                Keyword::NoKeyword if dialect_of!(self is SnowflakeDialect | GenericDialect) => {
+                // Case when Snowflake Semi-structured data like key:value
+                Keyword::NoKeyword | Keyword::LOCATION if dialect_of!(self is SnowflakeDialect | GenericDialect) => {
                     Ok(Value::UnQuotedString(w.value))
                 }
                 _ => self.expected("a concrete value", Token::Word(w)),
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1430,7 +1430,7 @@ impl<'a> Parser<'a> {
             Ok(Expr::JsonAccess {
                 left: Box::new(expr),
                 operator: JsonOperator::Colon,
-                right: Box::new(self.parse_subexpr(51)?),
+                right: Box::new(Expr::Value(self.parse_value()?)),
             })
         } else if Token::Arrow == tok
             || Token::LongArrow == tok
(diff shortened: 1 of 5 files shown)


tests/sqlparser_common.rs
```diff
@@ -1594,6 +1594,15 @@ fn parse_limit_accepts_all() {
     );
 }
 
+#[test]
+fn parse_json_using_colon() {
+    one_statement_parses_to("SELECT field:key FROM t", "SELECT field ->> key FROM t");
```
AugustoFKL, line 1599: @yuval-illumex would it be possible to create an expression to maintain the original syntax?

This transformation seems to be using semantic analysis (converting `:` to `->>`) which, even though could be considered right depending on the dialect, doesn't maintain the original expression syntax.

AugustoFKL (changes requested):

yuval-illumex:
@AugustoFKL Thank you for your feedback! I change the code to still use the JsonAccess expression (because it's still access to JSON) But with Snowflake syntax support. Appreciate your feedback!

AugustoFKL (approved):
LGTM

@alamb ready for review :)

yuval-illumex:
@AugustoFKL @alamb I have a question about `JsonAccess`:
The right expression is `Identifier`, however according to the postgres [documentation (https://www.postgresql.org/docs/9.3/functions-json.html) the value can be int or text (in snowflake it must be text as I understand it).

Maybe instead of identifier it should be `Value` and not `Identifier`? What do you think?

yuval-illumex:
I found a bug on JsonAccess, the right value should be Value and not Identifier.
However, I can't find a way for the `parse_value` function to check if the dialect is Snowflake - only in the test. it works fine from the cli.

@AugustoFKL WDYT? If it seems right, any change for some assistance with the test?

AugustoFKL:
@yuval-illumex you can get the current dialect by using `dialect_of!(self is SnowflakeDialect)`

Please try this. The parser is trying to parse a value but is matching it to another expression. TBH there's 2 possible problems:
1. It's not using your new value support (maybe conflicting with identifiers).
2. Using `KEY` is conflicting with our current keyworks.

I'd try to test an expression without possible keywords and with this dialect validation adjust :)

LMK if it goes ok

src/parser.rs
```diff
@@ -3441,6 +3448,9 @@ impl<'a> Parser<'a> {
                     Some('\'') => Ok(Value::SingleQuotedString(w.value)),
                     _ => self.expected("A value?", Token::Word(w))?,
                 },
+                Keyword::NoKeyword if dialect_of!(self is SnowflakeDialect) => {
```
yuval-illumex, line 3451: @AugustoFKL Here is how I tried. For some reason, this if clause works when using the cli, but when running tests it fails

yuval-illumex:
@AugustoFKL in parse_value I have the following if clause:
 ` Keyword::NoKeyword if dialect_of!(self is SnowflakeDialect) => {
                    Ok(Value::UnQuotedString(w.value))
 }`

For some weird reason, the if clause which check dialect is working when running via cli (or python binding) but when running the snowflake tests, the if clause not working. I really have no idea why.

yuval-illumex:
@AugustoFKL Thank you for your response! your #2nd point was correct!

Appreciate your feedback on the current PR.

src/parser.rs
AugustoFKL, line 3451: @yuval-illumex adding the generic here is feasible?

I'd suggest adding it but, as identifiers have the same exact syntax, I don't know how that would work.

yuval-illumex: @AugustoFKL added :)

AugustoFKL (approved):
LGTM

@alamb ready for your look :)

yuval-illumex:
@alamb @AugustoFKL I added support when the key the user tries to extract from the json is "location". I have a user who tried it.

AugustoFKL (approved):
Not sure whether is reasonable or not to nitpick keywords when parsing.

But If no tests break, don't think it's a problem for now...

alamb (approved):
LGTM -- thank you @yuval-illumex  and @AugustoFKL

PR #699, opened 2022-11-02
Support the ARRAY type of Snowflake

yuval-illumex:
According snowflake [documentation](https://docs.snowflake.com/en/sql-reference/data-types-semistructured.html#array), There is an ARRAY type.
Behind the scenes  it contains the `VARIANT` Data type, but the syntax is different. Snowflake don't support `VARIANT[]`.
The query that I want to support is: `SELECT CAST (a as array)`
I started with creating a new data type called `SnowflakeArray`, But I think it's better to use the existing `ARRAY` data type (which created for the HIVE dialect)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3650,13 +3650,17 @@ impl<'a> Parser<'a> {
                 Keyword::ENUM => Ok(DataType::Enum(self.parse_string_values()?)),
                 Keyword::SET => Ok(DataType::Set(self.parse_string_values()?)),
                 Keyword::ARRAY => {
-                    // Hive array syntax. Note that nesting arrays - or other Hive syntax
-                    // that ends with > will fail due to "C++" problem - >> is parsed as
-                    // Token::ShiftRight
-                    self.expect_token(&Token::Lt)?;
-                    let inside_type = self.parse_data_type()?;
-                    self.expect_token(&Token::Gt)?;
-                    Ok(DataType::Array(Box::new(inside_type)))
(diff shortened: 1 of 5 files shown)


src/parser.rs
```diff
@@ -3672,13 +3672,20 @@ impl<'a> Parser<'a> {
...
-                    // Hive array syntax. Note that nesting arrays - or other Hive syntax
-                    // that ends with > will fail due to "C++" problem - >> is parsed as
-                    // Token::ShiftRight
-                    self.expect_token(&Token::Lt)?;
-                    let inside_type = self.parse_data_type()?;
-                    self.expect_token(&Token::Gt)?;
-                    Ok(DataType::Array(Box::new(inside_type)))
+                    if dialect_of!(self is SnowflakeDialect) {
+                        Ok(DataType::Array(Box::new(DataType::Custom(
+                            ObjectName(vec!["VARAINT".into()]),
```
alamb, line 3677: I would prefer if this didn't add VARAINT into the parse tree when it wasn't in the original syntax. Can we model this like `DataType::Array(None)` perhaps?

alamb (commented):
Thank you for the contribution @yuval-illumex

yuval-illumex:
@alamb I change the data type to be `Array(Option<Box<DataType>>)` to accept None.
I don't think it breaks anything but waiting to your feedback :) (I'm new to Rust).

Thank you!

alamb (approved):
Looks good to me -- thanks @yuval-illumex

PR #701, opened 2022-11-06
Add precision for TIME, DATETIME, and TIMESTAMP

AugustoFKL:
feat: add precision for TIME, DATETIME, and TIMESTAMP data types

Now all those statements are both parsed and displayed with precision and timezone info. Tests were added to the ones presented in the ANSI standard [(1)](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#datetime-type).

[1] : https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#datetime-type

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -11,7 +11,7 @@
 // limitations under the License.
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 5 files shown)


alamb (approved):
LGTM -- thanks @AugustoFKL

PR #702, opened 2022-11-06
Add FULLTEXT support for MySQL and Generic dialects

AugustoFKL:
feat: add FULLTEXT [(1)](https://dev.mysql.com/doc/refman/8.0/en/create-table.html), [(2)](https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html), option on create table for MySQL and Generic dialects.

[1] : https://dev.mysql.com/doc/refman/8.0/en/create-table.html
[2] : https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html


src/ast/ddl.rs
```diff
@@ -270,6 +270,29 @@ pub enum TableConstraint {
...
+    /// key_part: col_name
+    /// ```
+    ///
+    /// [1]: https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html
+    FulltextOrSpatial {
+        /// Whether this is a `FULLTEXT` (true) or `SPATIAL` (false) definition.
+        fulltext: bool,
+        /// Whether the type is followed by the keyword `KEY` (Some(true)), `INDEX` (Some(false)),
+        /// or no keyword at all (None).
+        opt_key: Option<bool>,
```
alamb, line 290: Maybe an enum would be easier to use here

```rust
enum SpatialKey {
  Key,
  Index,
  None,
}
```

🤔

alamb (approved):

src/ast/ddl.rs
AugustoFKL: Will do it asap!

AugustoFKL: @alamb done!

src/ast/ddl.rs
```diff
@@ -330,6 +352,64 @@ impl fmt::Display for TableConstraint {
...
+            } => {
+                if *fulltext {
+                    write!(f, "FULLTEXT")?;
+                } else {
+                    write!(f, "SPATIAL")?;
+                }
+
+                if !matches!(index_type_display, KeyOrIndexDisplay::None) {
+                    write!(f, " {}", index_type_display)?;
+                }
```
alamb, line 369: Since `KeyOrIndexDisplay::None` is displayed as `""` I suspect you don't even have to have this test

```suggestion
                write!(f, " {}", index_type_display)?;
```

alamb (approved):
LGTM -- thanks @AugustoFKL

src/ast/ddl.rs
AugustoFKL: @alamb the problem is that i'd have an additional space compared to the input string. Although it isn't incorrect, IMHO would be weird to print something like:

```sql
CREATE TABLE tb (c1 TEXT, FULLTEXT  (c1, c2);
```

PR #703, opened 2022-11-08
Fix broken DataFusion link

jmg-duarte:


alamb (approved):
Thank you @jmg-duarte !

PR #704, opened 2022-11-08
Support `CREATE TABLE x AS TABLE y`

### LINKED ISSUE #614: Support `CREATE TABLE x AS TABLE y`
Hi, I'd like to be able to parse a statement like `CREATE TABLE films2 AS TABLE films` as from the second Postgres example found at [this source](https://www.postgresql.org/docs/current/sql-createtableas.html). I currently tried to run this command with DataFusion but got a parser error, which is why I'm submitting the issue in this repo.

@andygrove I'm currently working on `test_table_schema` in [dask-sql/test_schema.py](https://github.com/dask-contrib/dask-sql/blob/datafusion-sql-planner/tests/integration/test_schema.py) which has a statement `CREATE TABLE bar AS TABLE root.df`, which is where the issue came from.

sarahyurick:
Closes #614 

My solution was to parse `CREATE TABLE x AS TABLE y` the same as `CREATE TABLE x AS SELECT * FROM y`.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4575,7 +4575,12 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse `CREATE TABLE x AS TABLE y`
-    pub fn parse_as_table(&self, token1: Token, token2: Token, token3: Token) -> Result<Table, ParserError> {
+    pub fn parse_as_table(
+        &self,
+        token1: Token,
+        token2: Token,
+        token3: Token,
+    ) -> Result<Table, ParserError> {
         let table_name;
         let schema_name;
         if token2 == Token::Period {
@@ -4596,8 +4601,8 @@ impl<'a> Parser<'a> {
                 }
             }
             Ok(Table {
-                table_name: Some(table_name.to_string()),
-                schema_name: Some(schema_name.to_string()),
+                table_name: Some(table_name),
+                schema_name: Some(schema_name),
             })
         } else {
             match token1 {
@@ -4609,7 +4614,7 @@ impl<'a> Parser<'a> {
                 }
             }
             Ok(Table {
-                table_name: Some(table_name.to_string()),
+                table_name: Some(table_name),
                 schema_name: None,
             })
         }


src/parser.rs
```diff
@@ -4378,6 +4395,161 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse `CREATE TABLE x AS TABLE y`
+    /// the same as `CREATE TABLE x AS SELECT * FROM y`
+    /// Mostly the same logic as parse_select,
+    /// except for the "projection" and "from" variables
+    pub fn parse_as_table(&mut self) -> Result<Select, ParserError> {
```
sarahyurick, line 4402: Probably doesn't need all of the same logic as from `parse_select`.

tests/sqlparser_common.rs
```diff
@@ -2153,6 +2153,17 @@ fn parse_create_table_as() {
     }
 }
 
+#[test]
+fn parse_create_table_as_table() {
+    let sql1 = "CREATE TABLE new_table AS TABLE old_table";
```
alamb, line 2158: In general I think we try to have the sqlparser crate parse into an AST that reflects the parsed SQL -- so rather that assuming some sematics (like `CREATE TABLE x AS y` means the same as `CREATE TABLE AS SELECT * ..` it would be more consistent to parse the create table differently.

Does that make sense?

sarahyurick: Yes, I can edit accordingly. Thanks!

src/ast/query.rs
```diff
@@ -152,6 +153,21 @@ impl fmt::Display for SetQuantifier {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Table {
+    pub table_name: String,
+    // pub schema_name: String,
```
sarahyurick, line 161: I was having a bit of trouble parsing something like `my_schema_name.my_table_name`, but I imagine that we'd ideally want the schema name to be a part of the `Table` struct.

sarahyurick:
Hi @alamb, let me know if these updates look reasonable 🙂

src/ast/query.rs
```diff
@@ -84,7 +84,7 @@ pub enum SetExpr {
     },
     Values(Values),
     Insert(Statement),
-    // TODO: ANSI SQL supports `TABLE` here.
+    Table(Box<Table>),
```
alamb, line 87: 👍

src/ast/query.rs
```diff
@@ -152,6 +153,30 @@ impl fmt::Display for SetQuantifier {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Table {
```
alamb, line 159:
```suggestion
/// A [`TABLE` command]( https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE)
pub struct Table {
```

alamb: I had to look up that this indeed is part of postgres syntax: https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE

src/parser.rs
```diff
@@ -4180,6 +4180,25 @@ impl<'a> Parser<'a> {
             SetExpr::Query(Box::new(subquery))
         } else if self.parse_keyword(Keyword::VALUES) {
             SetExpr::Values(self.parse_values()?)
+        } else if self.parse_keyword(Keyword::TABLE) {
+            self.next_token();
+            self.next_token();
+            self.next_token();
+            match &self.tokens[0] {
```
alamb, line 4187: I think `self.peek_token()` would be more appropriate here.

tests/sqlparser_common.rs
```diff
@@ -2153,6 +2153,53 @@ fn parse_create_table_as() {
...
+        fetch: None,
+        lock: None,
+    });
+
+    match verified_stmt(sql1) {
+        Statement::CreateTable { query, .. } => {
+            assert_eq!(query.unwrap(), expected_query1);
+        }
+        _ => unreachable!(),
+    }
```
alamb, line 2178:
```suggestion
    match verified_stmt(sql1) {
        Statement::CreateTable { query, name, .. } => {
            assert_eq!(name, ObjectName(vec![Ident::new("new_table")]));
            assert_eq!(query.unwrap(), expected_query1);
        }
        _ => unreachable!(),
    }
```

I think it would be good to verify the new name as well

tests/sqlparser_common.rs
```diff
@@ -2153,6 +2153,53 @@ fn parse_create_table_as() {
...
+        fetch: None,
+        lock: None,
+    });
+
+    match verified_stmt(sql2) {
+        Statement::CreateTable { query, .. } => {
+            assert_eq!(query.unwrap(), expected_query2);
+        }
+        _ => unreachable!(),
+    }
```
alamb, line 2200:
```suggestion
    match verified_stmt(sql2) {
        Statement::CreateTable { query, name, .. } => {
            assert_eq!(name, ObjectName(vec![Ident::new("new_table")]));
            assert_eq!(query.unwrap(), expected_query2);
        }
        _ => unreachable!(),
    }
```

src/parser.rs
```diff
@@ -4378,6 +4397,50 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// Parse `CREATE TABLE x AS TABLE y`
+    pub fn parse_as_table(&mut self) -> Result<Table, ParserError> {
+        let mut as_flag = false;
+        let mut table_flag = false;
+        let mut table_name = "";
+        let mut schema_name = "";
+
+        let mut tokens_iter = self.tokens.iter().enumerate().peekable();
```
alamb, line 4407: I find this somewhat confusing -- it parses the statement directly starting at the first token -- what if the first token was whitespace?

In fact, when I change the test to start with whitespace it failed:

```
#[test]
fn parse_create_table_as_table() {
    let sql1 = "   CREATE TABLE new_table AS TABLE old_table";
```

```
---- parse_create_table_as_table stdout ----
thread 'parse_create_table_as_table' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected CREATE TABLE ... AS TABLE ..., found:  ")', src/test_utils.rs:86:61
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
```

src/parser.rs
```diff
@@ -4180,6 +4180,25 @@ impl<'a> Parser<'a> {
             SetExpr::Query(Box::new(subquery))
         } else if self.parse_keyword(Keyword::VALUES) {
             SetExpr::Values(self.parse_values()?)
+        } else if self.parse_keyword(Keyword::TABLE) {
+            self.next_token();
+            self.next_token();
+            self.next_token();
```
alamb, line 4186: Rather than consuming these tokens and the starting all the way back at `tokens[0]` the typical pattern in this codebase would be to call `self.parse_as_table` which would assume the token stream was already at `CREATE TABLE name AS ...` and only parse the name.

I wonder if you could use `parse_identifier` too rather than parsing schema / table name directly

alamb (commented):
Sorry for the delay in review @sarahyurick  -- this is looking better but not quite ready yet I think

sarahyurick:
Thanks @alamb ! Updated to use `peek_token()` (and also `peek_nth_token()`) and does not fail on your whitespace example anymore. I wasn't exactly sure how to use `parse_identifier()` for this, but hopefully the logic for `parse_as_table()` should be a lot cleaner now.

alamb (approved):
Thanks @sarahyurick  -- I think this PR now looks good. I have some ideas of how to simplify the code a little more but I'll make a follow on PR with that change.

Thanks again for sticking with this

PR #705, opened 2022-11-10
Fix interval parsing logic and precedence

sarahyurick:
I originally opened an issue for this on the DataFusion side: https://github.com/apache/arrow-datafusion/issues/3944, but found the error to lie in how the queries were being parsed.

These changes allow something like `"SELECT col FROM test WHERE d3_date > d1_date + INTERVAL '5 days' AND d2_date > d1_date + INTERVAL '3 days'"` to be parsed correctly. Previously, the '5 days' was getting separated from the `INTERVAL` keyword and lumped into the `AND` operator.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -363,6 +363,36 @@ impl<'a> Parser<'a> {
         Ok(expr)
     }
 
+    pub fn parse_interval_expr(&mut self) -> Result<Expr, ParserError> {
+        let precedence = 0;
+        let mut expr = self.parse_prefix()?;
+
+        loop {
+            let next_precedence = self.get_next_interval_precedence()?;
+
+            if precedence >= next_precedence {
+                break;
+            }
+
+            expr = self.parse_infix(expr, next_precedence)?;
+        }
+
+        Ok(expr)
+    }
+
+    /// Get the precedence of the next token
+    /// With AND, OR, and XOR
+    pub fn get_next_interval_precedence(&self) -> Result<u8, ParserError> {
+        let token = self.peek_token();
+
+        match token {
+            Token::Word(w) if w.keyword == Keyword::AND => Ok(0),
+            Token::Word(w) if w.keyword == Keyword::OR => Ok(0),
+            Token::Word(w) if w.keyword == Keyword::XOR => Ok(0),
+            _ => self.get_next_precedence()
+        }
+    }
+
     pub fn parse_assert(&mut self) -> Result<Statement, ParserError> {
         let condition = self.parse_expr()?;
         let message = if self.parse_keyword(Keyword::AS) {
@@ -1151,7 +1181,7 @@ impl<'a> Parser<'a> {
 
         // The first token in an interval is a string literal which specifies
         // the duration of the interval.
-        let value = self.parse_expr()?;
+        let value = self.parse_interval_expr()?;
 
         // Following the string literal is a qualifier which indicates the units
         // of the duration specified in the string literal.
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -389,7 +389,7 @@ impl<'a> Parser<'a> {
             Token::Word(w) if w.keyword == Keyword::AND => Ok(0),
             Token::Word(w) if w.keyword == Keyword::OR => Ok(0),
             Token::Word(w) if w.keyword == Keyword::XOR => Ok(0),
-            _ => self.get_next_precedence()
+            _ => self.get_next_precedence(),
         }
     }
 
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -3216,19 +3216,22 @@ fn parse_interval_and_or_xor() {
 
     assert_eq!(actual_ast, expected_ast);
 
-    verified_stmt("SELECT col FROM test \
+    verified_stmt(
+        "SELECT col FROM test \
         WHERE d3_date > d1_date + INTERVAL '5 days' \
-        AND d2_date > d1_date + INTERVAL '3 days'"
+        AND d2_date > d1_date + INTERVAL '3 days'",
     );
 
-    verified_stmt("SELECT col FROM test \
+    verified_stmt(
+        "SELECT col FROM test \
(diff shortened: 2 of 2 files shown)


sarahyurick:
I'm having trouble creating a test for this. Right now I have something like:

```
#[test]
fn parse_interval_and_or_xor() {
    let sql = "SELECT col FROM test \
        WHERE d3_date > d1_date + INTERVAL '5 days' \
        AND d2_date > d1_date + INTERVAL '3 days'";

    let ast = Parser::parse_sql(&GenericDialect {}, sql).unwrap();

    let expected_ast = vec![Query(Query {
        with: None,
        body: Select(Select {
            distinct: false,
            top: None,
            projection: vec![UnnamedExpr(Identifier(Ident { value: "col".to_string(), quote_style: None }))],
            into: None,
            from: vec![TableWithJoins {
                relation: Table {
                    name: ObjectName(vec![Ident { value: "test".to_string(), quote_style: None }]),
                    alias: None,
                    args: None,
                    with_hints: vec![] },
                joins: vec![] }],
            lateral_views: vec![],
            selection: Some(BinaryOp {
                left: BinaryOp {
                    left: Identifier(Ident { value: "d3_date".to_string(), quote_style: None }),
                    op: Gt,
                    right: BinaryOp {
                        left: Identifier(Ident { value: "d1_date".to_string(), quote_style: None }),
                        op: Plus,
                        right: Interval {
                            value: Value(SingleQuotedString("5 days")),
                            leading_field: None,
                            leading_precision: None,
                            last_field: None,
                            fractional_seconds_precision: None } } },
                op: And,
                right: BinaryOp {
                    left: Identifier(Ident { value: "d2_date".to_string(), quote_style: None }),
                    op: Gt,
                    right: BinaryOp {
                        left: Identifier(Ident { value: "d1_date".to_string(), quote_style: None }),
                        op: Plus,
                        right: Interval {
                            value: Value(SingleQuotedString("3 days")),
                            leading_field: None,
                            leading_precision: None,
                            last_field: None,
                            fractional_seconds_precision: None } } } }),
            group_by: vec![],
            cluster_by: vec![],
            distribute_by: vec![],
            sort_by: vec![],
            having: None,
            qualify: None }),
        order_by: vec![],
        limit: None,
        offset: None,
        fetch: None,
        lock: None })];

    assert_eq!(actual_ast, expected_ast);
}
```

But this is pretty clunky, plus the main thing we want to check is that the `selection` value (with all the nested BinaryOps) is correct. I was wondering what the best way to do this would be?

src/parser.rs
```diff
@@ -363,6 +363,36 @@ impl<'a> Parser<'a> {
...
+
+    /// Get the precedence of the next token
+    /// With AND, OR, and XOR
+    pub fn get_next_interval_precedence(&self) -> Result<u8, ParserError> {
+        let token = self.peek_token();
+
+        match token {
+            Token::Word(w) if w.keyword == Keyword::AND => Ok(0),
+            Token::Word(w) if w.keyword == Keyword::OR => Ok(0),
+            Token::Word(w) if w.keyword == Keyword::XOR => Ok(0),
```
alamb, line 391: why XOR here? 

Also OR and XOR do not appear to be covered in the tests

src/parser.rs
```diff
@@ -363,6 +363,36 @@ impl<'a> Parser<'a> {
         Ok(expr)
     }
 
+    pub fn parse_interval_expr(&mut self) -> Result<Expr, ParserError> {
```
alamb, line 366: I wonder if you investigated following a similar pattern to `NOT` which adjusts the precedence based on peeking in the token stream? 

https://github.com/sqlparser-rs/sqlparser-rs/blob/87b4a168cbcd83439c4e8729dd992d150eebd15c/src/parser.rs#L1648-L1660

I couldn't find any example of a similar partten in the codebase

https://github.com/search?q=repo%3Asqlparser-rs%2Fsqlparser-rs%20parse_infix&type=code

What do you think @AugustoFKL ?

alamb:
Thank you for the contribution @sarahyurick

src/parser.rs
AugustoFKL: @alamb I think we may need to refactor the expression parsing... 

With the current structure, we end up with no such good expansion of expressions to parse. And there's still the (big) problem with precedence analysis.

But, for now, and considering the scope of this PR, I think that's OK. Doing it for `NOT` would be better in another PR, right?

tests/sqlparser_common.rs
```diff
@@ -3120,6 +3120,103 @@ fn parse_interval() {
     );
 }
 
+#[test]
+fn parse_interval_and_or_xor() {
```
AugustoFKL, line 3124: @sarahyurick please try to add tests considering the `FROM -> TO` being equal as well, if possible. It's nice to have a known about which queries are parsed the same as the original one.

src/parser.rs
sarahyurick: I mostly just included XOR since I figured precedence rules in this case would be similar to AND and OR, and they're also lumped together at https://github.com/sqlparser-rs/sqlparser-rs/blob/4b1dc1abf7138cb1576707c194ec89ad89049d27/src/parser.rs#L1635.

Since the AND test looks ok, I can add OR and XOR tests as well.

tests/sqlparser_common.rs
sarahyurick: Sure, just to clarify do you mean tests with something like `INTERVAL '1-1' YEAR TO MONTH` in them?

sarahyurick: Got it, thanks!

sarahyurick:
Thanks @alamb @AugustoFKL ! Let me know what you think. Should I update anything with the precedence logic, or just keep as is for now?

AugustoFKL (commented):
lgtm

@sarahyurick IMHO that's ok

alamb (approved):

alamb:
Thanks again @sarahyurick  and @AugustoFKL

PR #708, opened 2022-11-13
Support `MATCH AGAINST`

AugustoFKL:
Support was added for both MySQL and Generic dialects.

Reference: https://dev.mysql.com/doc/refman/8.0/en/fulltext-search.html#function_match


alamb (approved):
Looks great to me -- thank you @AugustoFKL

PR #288, opened 2021-01-08
Add Tokenizer::tokenize_with_location() API

segoon:
* add TokenWithLocation to track Token location
* implement Tokenizer::tokenize_with_location() -> ``Result<Vec<TokenWithLocation>, ..>``
* refactor line/column tracking to alter current position along with peekable.next() (the old approach with duplicate position tracking was error-prone and contained a bug in multiline word parsing, see the test)
* eliminate String copying for Tokenizer, use ``&'a str`` instead
* minor TokenizerError column fixes, now it shows the missing symbol position and not the expected token start location

Issue #524, opened 2022-06-16
Add support of locations in AST


frolovdev:
**Proposal**

I go through the sources of the parser and find out that there is no such thing as a location in AST

I have in mind something like this:

lexer layer

```
pub enum Token {
   EOF { loc: Option<Loc> },
   Word { value: Word, loc: Option<Loc> }
  // and so on
}

struct Loc {
   start: i32
   end: i32
   line: i32
   column: i32
}
```

parser layer

```
pub enum Expr {
   Identifier{ value: Ident, loc: Option<Loc> },
 }

struct Loc {
   start: i32
   end: i32
   startToken: Token,
   endToken: Token
}

```

**Behind motivation**
Much better error reporting

**Proposal design highlights**

Why `Option<Loc>`? 

We can allow to run the CLI in different modes, with emitting and not emitting locations

**Plan of migration**

Doesn't cover it right now ( but I have in mind high-level plan of gradual migration). Just want to discuss the idea first of all

---

@alamb what do you think? I can drive this migration if we discuss details and make a decision about it

olalonde:
Some more prior art: https://github.com/sqlparser-rs/sqlparser-rs/pull/189

I'd be very interested in a lossless AST. I wanted to use this library to collect comments from a SQL file but that use case isn't supported right now since comments are discarded.

ankrgyl:
If it's still of interest, I was able to revive #288 and get it working w/ fewer merge conflicts + passing tests. Happy to send a PR if interesting.

alamb:
@ankrgyl  would love to see a PR -- basically if adding support won't cause significant maintenance burden and a reasonable (but not pain free of course) upgrade path.

PR #710, opened 2022-11-15
Introduce location tracking in the tokenizer and parser

ankrgyl:
This revives PR #288 and #514, rebased against the latest and done with fewer variable renames (limiting merge conflicts).

Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/524

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -6751,4 +6751,31 @@ mod tests {
             r#"UPDATE test SET name = $1, value = $2, where = $3, create = $4, is_default = $5, classification = $6, sort = $7 WHERE id = $8"#
         );
     }
+
+    #[test]
+    fn test_tokenizer_error_loc() {
+        let sql = "foo '";
+        let ast = Parser::parse_sql(&GenericDialect, sql);
+        assert_eq!(
+            ast,
+            Err(ParserError::TokenizerError(
+                "Unterminated string literal at Line: 1, Column 5".to_string()
+            ))
+        );
+    }
+
+    #[test]
+    fn test_parser_error_loc() {
+        // TODO: Once we thread token locations through the parser, we should update this
+        // test to assert the locations of the referenced token
+        let sql = "SELECT this is a syntax error";
+        let ast = Parser::parse_sql(&GenericDialect, sql);
+        assert_eq!(
+            ast,
+            Err(ParserError::ParserError(
+                "Expected [NOT] NULL or TRUE|FALSE or [NOT] DISTINCT FROM after IS, found: a"
+                    .to_string()
+            ))
+        );
+    }
 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4769,7 +4769,7 @@ impl<'a> Parser<'a> {
         let table_name;
         let schema_name;
         if token2 == Token::Period {
-            match token1 {
+            match token1.token {
                 Token::Word(w) => {
                     schema_name = w.value;
                 }
@@ -4777,7 +4777,7 @@ impl<'a> Parser<'a> {
                     return self.expected("Schema name", token1);
                 }
             }
-            match token3 {
+            match token3.token {
                 Token::Word(w) => {
                     table_name = w.value;
                 }
@@ -4790,7 +4790,7 @@ impl<'a> Parser<'a> {
                 schema_name: Some(schema_name),
             })
         } else {
-            match token1 {
+            match token1.token {
                 Token::Word(w) => {
                     table_name = w.value;
                 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1849,7 +1849,7 @@ impl<'a> Parser<'a> {
                 token => {
                     return token
                         .cloned()
-                        .unwrap_or(TokenWithLocation::wrap(Token::EOF))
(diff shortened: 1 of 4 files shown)


ankrgyl:
@alamb @AugustoFKL this PR addresses #524.

This change is reasonably contained, but one thing that could make it even smaller would be maintaining the interface (i.e.`next_token()` returns a `Token` not a `TokenWithLocation`) and maybe adding something like `last_location()` to access the location just when you need it (e.g. for error messages).

alamb:
Thanks @ankrgyl  -- I'll try to look at this PR more carefully later this week

AugustoFKL:
@ankrgyl @alamb sorry for the delay. This has been a hectic week for me. I'm on vacation next week, and will be able to review this properly

src/tokenizer.rs
```diff
@@ -300,6 +300,41 @@ impl fmt::Display for Whitespace {
...
+/// Location in input string
+#[derive(Debug, PartialEq, Clone)]
+pub struct Location {
+    /// Line number, starting from 1
+    pub line: u64,
+    /// Line column, starting from 1
+    pub column: u64,
+}
+
+/// A [Token] with [Location] attached to it
```
alamb, line 312: We can probably avoid many of the changes in this PR by implementing `PartialEq` between Token and TokenWithLocation:

```rust
impl PartialEq<Token>` for TokenWithLocation
```

Which would allow things like

```rust
        if self.peek_nth_token(1).token == Token::RArrow {
```

to keep working

src/tokenizer.rs
```diff
@@ -321,58 +356,88 @@ impl fmt::Display for TokenizerError {
 #[cfg(feature = "std")]
 impl std::error::Error for TokenizerError {}
 
+struct State<'a> {
```
alamb, line 359: 👍

alamb (commented):
This PR looks great @ankrgyl  -- thank you for proposing it; I apologize for the delay in finding time to review it

I think we can avoid many of the changes with an `impl PartialEq` as I mentioned

Otherwise, all this PR needs is some tests and I think it would be ready to merge.

Thanks again!

ankrgyl:
@alamb updated with the `PartialEq` suggestion (great idea!). For tests -- do you have a suggestion on what would be best? I can tweak a few of the tests to tokenize w/ location and hardcode them in, or do something else?

alamb:
> For tests -- do you have a suggestion on what would be best? I can tweak a few of the tests to tokenize w/ location and hardcode them in, or do something else?

I think the goal should be that if we accidentally break the location parsing code some tests will fail. I think ensuring the locations are set on parse errors would be particularly helpful. 

So perhaps you could add a few tests that assert locations of parsed tokens, and then a test that has a parsing error that retrieves the location of the error (if this last test needs more code changes, we could do it as a follow on PR)


Thank you so much for pushing this along @ankrgyl  -- this is going to be a great addition (and is a long asked for feature)

ankrgyl:
Added a few tests. We'll need to thread parser locations through as a follow up PR, but I added a test that fails in the parser due to a tokenizer error (and asserts the location) + a test that we can fill in with locations once the parser carries along that info.

src/parser.rs
```diff
@@ -6592,4 +6592,31 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_tokenizer_error_loc() {
+        let sql = "foo '";
+        let ast = Parser::parse_sql(&GenericDialect, sql);
+        assert_eq!(
+            ast,
+            Err(ParserError::TokenizerError(
+                "Unterminated string literal at Line: 1, Column 5".to_string()
```
alamb, line 6603: ❤️

alamb (approved):
I think this looks great -- thank you @ankrgyl

Would you like a chance to review @AugustoFKL ?

alamb:
This appears to have a logical conflict with master.

I am planning on cutting a release soon -- maybe given the nature of this change I will make release right before this change, and then merge this one. That way people who use the tip of this repository can effectively beta test the changes before we make a release for everyone 🤔

ankrgyl:
That seems wise to me. We may even want to flesh out the (potentially breaking) parser changes so that users who upgrade only need to adjust to breakages once. Happy to contribute those and collaborate with you on that.

ankrgyl:
Additionally, I just rebased and updated the merge conflicts.

AugustoFKL (approved):

alamb:
> That seems wise to me. We may even want to flesh out the (potentially breaking) parser changes so that users who upgrade only need to adjust to breakages once. Happy to contribute those and collaborate with you on that.


That would be awesome. I plan to make a release soon (TM) -- either later today or over the weekend sometime. Then let's merge this one in and iterate.

alamb:
@ankrgyl  I took the liberty of fixing the clippy CI error and merging up from main to this branch.

alamb:
>  That seems wise to me. We may even want to flesh out the (potentially breaking) parser changes so that users who upgrade only need to adjust to breakages once. Happy to contribute those and collaborate with you on that.


This is merged and should be included in the 0.29.0 release (eta in approximately 1 months time) I would love to start collaborating to fixup the parser. The first step might be to list / sketch out the changes that are needed.

Thanks again -- this is super exciting!

PR #717, opened 2022-11-19
refactor: add PartialOrd, Ord, and Copy whenever possible

AugustoFKL:
This allows other projects to use our structures inside others that need those.

Based on Rust's API Guidelines [(1)](https://rust-lang.github.io/api-guidelines/interoperability.html)

[1] : https://rust-lang.github.io/api-guidelines/interoperability.html


alamb (approved):
Makes sense to me -- thanks @AugustoFKL

Issue #716, opened 2022-11-18
Support `EXCLUDE` and `RENAME` in Snowflake


CorentinLimier:
Hello,

Snowflake added two SQL keywords : `EXCLUDE` and `RENAME`

ex of query : 

```sql
SELECT 
    *
    EXCLUDE (col_a)
    RENAME col_b as b
FROM data
```

Link to documentation : https://docs.snowflake.com/en/sql-reference/sql/select.html

Thank you

AugustoFKL:
@CorentinLimier https://github.com/sqlparser-rs/sqlparser-rs/pull/721 exclude PR

PR #721, opened 2022-11-20
feat: add snowflake and generic exclude support

AugustoFKL:
The exclude clause can be used after a possibly qualified on SELECT. [(1)](https://docs.snowflake.com/en/sql-reference/sql/select.html)

Refs: #716


alamb (approved):
LGTM -- thank you @AugustoFKL

PR #496, opened 2022-05-11
Support hive `CREATE FUNCTION` syntax

mobuchowski:
Add `CreateFunction` syntax based on Hive: https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl#LanguageManualDDL-Create/Drop/ReloadFunction

Signed-off-by: Maciej Obuchowski <obuchowski.maciej@gmail.com>

PR #722, opened 2022-11-21
Support postgres `CREATE FUNCTION` 

wangrunji0408:
This PR adds some support for postgres `CREATE FUNCTION` based on:
https://www.postgresql.org/docs/15/sql-createfunction.html

To live with the hive syntax introduced by #496, we move its field `class_name` and `using` into the new `CreateFunctionBody` structure.


src/ast/mod.rs
```diff
@@ -1363,11 +1363,14 @@ pub enum Statement {
...
+    /// Postgres: https://www.postgresql.org/docs/15/sql-createfunction.html
     CreateFunction {
+        or_replace: bool,
         temporary: bool,
         name: ObjectName,
-        class_name: String,
-        using: Option<CreateFunctionUsing>,
+        args: Option<Vec<CreateFunctionArg>>,
+        return_type: Option<DataType>,
+        bodies: Vec<CreateFunctionBody>,
```
alamb, line 1373: I find the Vec<CreateFunctionBody>` somewhat confusing as it implies that some of them can be repeated. For example, can there be two `CreateFunctionBody::Returning` in this list? I don't think it is possible

Did you consider creating optional fields to encode this in the type system itself?

So instead of 

```rust
    CreateFunction {
        or_replace: bool,
        temporary: bool,
        name: ObjectName,
        args: Option<Vec<CreateFunctionArg>>,
        return_type: Option<DataType>,
        bodies: Vec<CreateFunctionBody>,
```

Something like

```rust

```rust
    CreateFunction {
        or_replace: bool,
        temporary: bool,
        name: ObjectName,
        args: Option<Vec<CreateFunctionArg>>,
        return_type: Option<DataType>,
        as: Option<String>.
        /// LANGUAGE lang_name
        language: Option<Ident>,
...
}
```

?

tests/sqlparser_hive.rs
```diff
@@ -244,17 +244,19 @@ fn parse_create_function() {
...
             name,
-            class_name,
-            using,
+            bodies,
+            ..
         } => {
             assert!(temporary);
-            assert_eq!("mydb.myfunc", name.to_string());
-            assert_eq!("org.random.class.Name", class_name);
+            assert_eq!(name.to_string(), "mydb.myfunc");
```
alamb, line 251: these changes look good to me -- what do you think @mobuchowski ?

alamb (commented):
Thank you very much for the contribution @wangrunji0408 

I had a question about the use of `Vec<CreateFunctionBody>` but otherwise this PR is looking good. Thank you!

cc @AugustoFKL 

cc @mobuchowski as you contributed the original `CREATE FUNCTION` https://github.com/sqlparser-rs/sqlparser-rs/pull/496

src/ast/mod.rs
AugustoFKL: @alamb @wangrunji0408 I have some time, but will not be able to review the whole PR RN, sorry.
- The return type does not match the PostgreSQL syntax. @alamb, do you think it's reasonable to add something like a `SUPPORTED SYNTAX` field to the header of the structure? Here's an [example]( https://github.com/AugustoFKL/sql-helper/blob/dev/src/ansi/ast/common.rs) that I'm implementing for myself.

- @wangrunji0408, for the body structure, it's a problem that we can repeat. You're creating a syntax that can be built wrong for a single dialect. Please let us know if you have a good reason (something like it being repeatable on Snowflake or something like that). However, @alamb, I think having all of them, as you showed, makes building it complex and moving it as well. Sometimes the upstream only wants to handle one or two scenarios, not the whole object. My suggestion is to have a structure that encapsulates them all:
```rust
CREATE FUNCTION {
...,
or_replace: bool, 
temporary: bool, 
name: ObjectName, 
args: Option<Vec<CreateFunctionArg>>, 
return_type: Option<DataType>, 
opt_body: Option<FunctionBody>,
}

pub enum FunctionBody {
Language...
}
```

src/ast/mod.rs
```diff
@@ -3598,6 +3608,121 @@ impl fmt::Display for ContextModifier {
...
+        }
+    }
+}
+
+/// Postgres: https://www.postgresql.org/docs/15/sql-createfunction.html
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[non_exhaustive]
+pub enum CreateFunctionBody {
+    /// AS 'definition'
```
mobuchowski, line 3702: Would add in a comment that Hive's `AS class_name` parses too here.

src/ast/mod.rs
mobuchowski: I agree with @AugustoFKL and @alamb regarding repeatable structures. 

From my side, I can still express Hive's `CREATE FUNCTION` syntax so it looks good for me - for me it's okay to change my code when bumping sqlparser-rs dependency.

tests/sqlparser_hive.rs
mobuchowski: I agree.

src/ast/mod.rs
wangrunji0408: Thanks for this catch. I didn't realize the problem with duplicate items. It has been changed to the way @AugustoFKL mentioned above.

src/ast/mod.rs
```diff
@@ -3679,6 +3690,131 @@ impl fmt::Display for ContextModifier {
...
+#[derive(Debug, Default, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct CreateFunctionBody {
+    /// LANGUAGE lang_name
+    pub language: Option<Ident>,
+    /// IMMUTABLE | STABLE | VOLATILE
+    pub behavior: Option<FunctionBehavior>,
+    /// AS 'definition'
+    ///
+    /// Note that Hive's `AS class_name` is also parsed here.
```
alamb, line 3789: 👍

alamb (approved):
Looks good to me -- thank you @wangrunji0408

PR #723, opened 2022-11-21
SUPPORT SEMI/ANTI JOIN syntax

mingmwang:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/712

Some of SQL engines support SEMI/ANTI JOIN syntax explicitly. This PR adds this support to sqlparser-rs.


src/parser.rs
```diff
@@ -4746,16 +4746,57 @@ impl<'a> Parser<'a> {
...
+                            Keyword::ANTI,
+                            Keyword::JOIN,
+                        ]);
+                        match join_type {
+                            Some(Keyword::OUTER) => {
+                                self.expect_keyword(Keyword::JOIN)?;
+                                match kw {
+                                    Keyword::LEFT => JoinOperator::LeftOuter,
+                                    Keyword::RIGHT => JoinOperator::RightOuter,
+                                    _ => unreachable!(),
```
AugustoFKL, line 4763: @mingmwang for this and all the next `unreachable!()`. Even if this is really unreachable, please try to add an error here, or handle it in a different way. Being unsafe is not really good.

AugustoFKL (changes requested):

src/parser.rs
alamb: Since it can have only two values, we can also make it a boolean value. I plan to merge this PR and then I will submit a follow on to remove the use of `unreachable!`

alamb (approved):

Issue #615, opened 2022-09-17
UPDATE doesn't support valid column name "value"


justhaug:
This sql:

```sql
UPDATE a SET value = 1
```

Fails to parse, it used to in previous versions and I suppose should be considered a regression.

The reason this fails is because `VALUE` is now considered a keyword and update doesn't allow non-quoted keywords.

For reference SQLite, Postgres and MySQL parse and run the statement (or any that has `VALUE` as a column name.)

I would make a PR, but I'm not sure if making a special case for `VALUE` keyword on update columns is a good idea.

AugustoFKL:
TBH I think that accepting dialect-specific word patterns would require a good refactor, even though it should be acceptable, like accepting a table named "table".

Maybe being a little less permissive than the usual is fine?

nickolay:
I would just revert the part of PR that changed this ( https://github.com/sqlparser-rs/sqlparser-rs/pull/375/files#diff-4a04259da480a6b794a2e947e4cc03eff4d1aa9330836f5b91cac68c5398193fR3168 introducing `parse_identifiers_non_keywords`).

If there's a need to reject certain keywords after SET, one must decide which keywords to reject list them in a const a la `RESERVED_FOR_COLUMN_ALIAS`.

AugustoFKL:
@nickolay isn't this too specific? Each dialect has its own keyword, with a  different scope from the ANSII. If we try to handle each possible minimum difference, we'll end up a lot of dialect-specific structures, which will be hard to maintain.

nickolay:
I didn't say I was in favor of trying to reject certain keywords after SET. I said that _even if_ we wanted to do that, relying on the [global keywords list is an error](https://github.com/sqlparser-rs/sqlparser-rs/blob/3beecc0a7a06c8dae47a4440129d01553c28c2ee/src/keywords.rs#L17-L23).

step-baby:
UPDATE also doesn't support valid column name "sort"

alamb:
Closed by https://github.com/sqlparser-rs/sqlparser-rs/pull/725

PR #725, opened 2022-11-22
fix: Supports updating valid column names whose names are the same as…

step-baby:
… keywords

#615

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3886,41 +3886,19 @@ impl<'a> Parser<'a> {
         Ok(ObjectName(idents))
     }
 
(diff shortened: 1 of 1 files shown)


AugustoFKL (approved):
For me, if the tests didn't break, looks OK... Not sure how this can affect other projects, though.

@alamb

alamb (approved):
I verified this behavior is consistent with postgres as well:

```sql
postgres=# create table foo(value int);
CREATE TABLE
postgres=# update foo set value = 5;
UPDATE 0
postgres=# 
```

alamb:
Thanks @step-baby

PR #726, opened 2022-11-22
fix: handle nested comments

yang-han:
Make the tokenize_multiline_comment() able to handle nested comments


AugustoFKL (approved):
lgtm

alamb (approved):
Thank you @yang-han and @AugustoFKL  for the review

PR #728, opened 2022-11-28
feat: add method to get current parsing index

neverchanje:
We need to know where the invalid token is so that we can highlight it, and probably color the word or the clause to make it more obvious what's wrong. It'd be better as a higher-level utility than as part of the parser, so I expose the `index` field.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -5775,6 +5775,11 @@ impl<'a> Parser<'a> {
         }
         Ok(sequence_options)
     }
+
+    /// The index of the first unprocessed token.
+    pub fn index(&self) -> usize {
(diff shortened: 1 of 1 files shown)


alamb (approved):

alamb:
Thanks @neverchanje

alamb:
@neverchanje what project are you working on ? There have been several requests over time for more support to show where the tokens came from -- e.g. https://github.com/sqlparser-rs/sqlparser-rs/pull/710 from @ankrgyl

ankrgyl:
@alamb FWIW I had to do the same for the project I'm working on (although did not include it in #710).

PR #731, opened 2022-11-28
[USING method] is supported when creating indexes.

### LINKED ISSUE #729: Postgres: [USING method] does not support index creation.
```rust
#[test]
fn test_create_index() {
     let pg_dialect = PostgreSqlDialect {};
     let sql = r#"
     CREATE INDEX "test_indexes" ON "public"."test" USING btree ("id","user_id","action","result");"#;
     let mut ast = Parser::parse_sql(&pg_dialect, sql).unwrap();
}
```

```text
called `Result::unwrap()` on an `Err` value: ParserError("Expected (, found: USING")
thread 'ast::tests::test_create_index' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected (, found: USING")',
```

step-baby:
closes #729

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2695,12 +2695,18 @@ impl<'a> Parser<'a> {
         let index_name = self.parse_object_name()?;
         self.expect_keyword(Keyword::ON)?;
         let table_name = self.parse_object_name()?;
+        let using = if self.expect_keyword(Keyword::USING).is_ok() {
+            Some(self.parse_identifier()?)
+        }else {
+            None
+        };
         self.expect_token(&Token::LParen)?;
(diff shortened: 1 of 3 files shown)


alamb (approved):
Thank you @step-baby

src/parser.rs
```diff
@@ -2695,12 +2695,18 @@ impl<'a> Parser<'a> {
         let index_name = self.parse_object_name()?;
         self.expect_keyword(Keyword::ON)?;
         let table_name = self.parse_object_name()?;
+        let using = if self.expect_keyword(Keyword::USING).is_ok() {
```
alamb, line 2698: I think we can avoid `is_ok`: here https://github.com/sqlparser-rs/sqlparser-rs/pull/731

PR #732, opened 2022-11-29
fix: unnest join constraint with alias parsing for BigQuery dialect

Ziinc:
This PR fixes a parsing bug for the BigQuery Dialect. 
The expected outcome is for the following statement to be parsed with join constraints:
```sql
select * from t1 join unnest(t1.col) as f1 on  c1 = 123
```
However, this statement would result in the following error:
```
ParserError("Expected end of statement, found: c1")
```
Closer inspection revealed that the following ast was produced for the join table factor 

```
UNNEST { 
  alias: Some(TableAlias { name: Ident { value: "f", quote_style: None }, columns: [] }), 
  array_expr: CompoundIdentifier([Ident { value: "t1", quote_style: None }, Ident { value: "a", quote_style: None }]), 
  with_offset: false, 
  with_offset_alias: Some(Ident { value: "ON", quote_style: None }) 
}
```
As you can see, the `with_offset_alias` parses the `ON` keyword despite the `with_offset` being false.

To replicate this behaviour, I added a failing test that verifies this behaviour difference
```
thread 'parse_join_constraint_unnest' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected end of statement, found: c1")', src/test_utils.rs:86:61
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test parse_join_constraint_unnest ... FAILED
```

This fix uses the boolean check to determine if the subsequent optional alias should be parsed.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4913,12 +4913,15 @@ impl<'a> Parser<'a> {
                 Err(_) => false,
             };
 
-            let with_offset_alias =
+            let with_offset_alias = if with_offset {
                 match self.parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS) {
                     Ok(Some(alias)) => Some(alias),
                     Ok(None) => None,
                     Err(e) => return Err(e),
-                };
+                }
+            } else {
+                None
+            };
 
             Ok(TableFactor::UNNEST {
                 alias,
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -99,7 +99,7 @@ fn parse_join_constraint_unnest_alias() {
     assert_eq!(
         only(
             bigquery()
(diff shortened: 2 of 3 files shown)


alamb (approved):
Looks great -- thank you @Ziinc

alamb:
@Ziinc  there are some CI failures on this PR -- can you please resolve them?

Ziinc:
@alamb I've fixed the lint error. For the test case, it seems like the inline comments for the tested cases were off, and the last case that I removed should actually raise a ParserError, since an alias + WITH OFFSET alias without a WITH OFFSET clause does not make sense.

alamb (approved):
Makes sense to me -- thank you @Ziinc

alamb (approved):

alamb:
Thanks again @Ziinc

PR #735, opened 2022-11-29
update on conflict with condition

zidaye:
close #734 
refer to：
https://www.postgresql.org/docs/current/sql-insert.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2655,7 +2655,16 @@ pub struct OnConflict {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum OnConflictAction {
     DoNothing,
-    DoUpdate(Vec<Assignment>),
+    DoUpdate(DoUpdate),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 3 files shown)


baoyachi:
LGTM

src/ast/mod.rs
```diff
@@ -2655,7 +2655,16 @@ pub struct OnConflict {
...
+    DoUpdate(DoUpdate),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DoUpdate {
+    /// Column assignments
+    pub assignments: Vec<Assignment>,
+    /// WHERE
+    pub selection: Option<Expr>,
```
alamb, line 2667: My preference is to call this `where` to match the SQL more closely -- I will make this change in a follow on PR

alamb (approved):
Thank you @zidaye

alamb:
Thanks again @zidaye

src/ast/mod.rs
alamb: Actually, now I see the name `selection` is consistent with the rest of this crate (e.g.  https://github.com/sqlparser-rs/sqlparser-rs/blob/faf75b71613431e66ec22c82e3c9ad6157474458/src/ast/mod.rs#L1139)

Thank you

PR #736, opened 2022-11-29
Snowflake - Add support in CLUSTER BY when creating Materialised View

yuval-illumex:
According to snowflake [docs](https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view.html), The last optional parameter is CLUSTER BY

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1114,6 +1114,7 @@ pub enum Statement {
         columns: Vec<Ident>,
         query: Box<Query>,
         with_options: Vec<SqlOption>,
+        cluster_by: Vec<Ident>,
     },
     /// CREATE TABLE
     CreateTable {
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -1114,6 +1114,7 @@ pub enum Statement {
         columns: Vec<Ident>,
         query: Box<Query>,
         with_options: Vec<SqlOption>,
+        cluster_by: Vec<Ident>,
```
alamb, line 1117: https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view.html says

```
  [ CLUSTER BY ( <expr1> [, <expr2> ... ] ) ]
```

But this PR has cluster by as only columns (not expressions).

I am fine with merging this as is but I wonder if you want to support the more general syntax?

alamb (commented):
As always, thank you @yuval-illumex 👍  Really appreciate all the help here

alamb (approved):

alamb:
We can make the "Expr" change in a follow on PR if desired. For now I think this is an improvement over main so I am merging it in

PR #737, opened 2022-11-30
Support  MySQL `ROWS` syntax for `VALUES`

aljazerzen:
MySQL does not support following query:

```sql
VALUES (1)
```
... but supports ...

```sql
VALUES ROW(1)
```
https://dev.mysql.com/doc/refman/8.0/en/values.html

Note that is does support:
```sql
INSERT INTO my_table VALUES (1)
```

Quality of my code reflects banality of this detail.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -710,6 +710,8 @@ impl fmt::Display for Top {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Values {
+    /// Was there an explict ROWs keyword (MySQL)?
+    /// <https://dev.mysql.com/doc/refman/8.0/en/values.html>
     pub explicit_row: bool,
     pub rows: Vec<Vec<Expr>>,
 }
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -709,16 +709,20 @@ impl fmt::Display for Top {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Values(pub Vec<Vec<Expr>>);
+pub struct Values {
+    pub explicit_row: bool,
+    pub rows: Vec<Vec<Expr>>,
+}
 
 impl fmt::Display for Values {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "VALUES ")?;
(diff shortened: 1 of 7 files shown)


src/ast/query.rs
```diff
@@ -709,16 +709,20 @@ impl fmt::Display for Top {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Values(pub Vec<Vec<Expr>>);
+pub struct Values {
+    pub explicit_row: bool,
```
alamb, line 713:
```suggestion
    /// Was there an explict ROWs keyword (MySQL)?
    /// <https://dev.mysql.com/doc/refman/8.0/en/values.html>
    pub explicit_row: bool,
```

src/parser.rs
```diff
@@ -5482,14 +5490,25 @@ impl<'a> Parser<'a> {
...
-    pub fn parse_values(&mut self) -> Result<Values, ParserError> {
-        let values = self.parse_comma_separated(|parser| {
+    pub fn parse_values(&mut self, within_insert: bool) -> Result<Values, ParserError> {
+        let mut explicit_row = false;
+
+        let rows = self.parse_comma_separated(|parser| {
+            if parser.parse_keyword(Keyword::ROW) {
+                explicit_row = true;
+            } else {
+                if !within_insert && parser.dialect.values_require_row_in_select() {
```
alamb, line 5500: How important is it to reject the ROWS keyword during inserts?  I see that mySQL is not consistent in that ROW is not allowed in INSERT statements https://dev.mysql.com/doc/refman/8.0/en/insert.html

However, I also don't see any (negative) tests for this case in this PR, though maybe I have missed them

My preference is to remove the special case handling for in select and make the parser permissive and support an optional ROW whenever a VALUES list is specified.

Downstream crates that care can detect `explicit_row` and error if they don't want to allow that syntax.

We tried to make the philosophy of "parse all the things" clearer recently in the introduction here https://github.com/sqlparser-rs/sqlparser-rs#extensible-sql-lexer-and-parser-for-rust

alamb (commented):
Thank you @aljazerzen 

The code and tests look good except for the `within_insert` special case

src/parser.rs
alamb: BTW I found

```
Quality of my code reflects banality of this detail.
```

😆

aljazerzen: Yeah that makes sense - there is no need to inforce explict row. And it simplifies the code a lot!

alamb (approved):
Looks good -- thank you @aljazerzen

PR #740, opened 2022-11-30
minor cleanup to avoid `is_ok()`

alamb:
Minor style cleanup I noticed while reviewing https://github.com/sqlparser-rs/sqlparser-rs/pull/731 from @step-baby

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2749,7 +2749,7 @@ impl<'a> Parser<'a> {
         let index_name = self.parse_object_name()?;
         self.expect_keyword(Keyword::ON)?;
         let table_name = self.parse_object_name()?;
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -2749,7 +2749,7 @@ impl<'a> Parser<'a> {
         let index_name = self.parse_object_name()?;
         self.expect_keyword(Keyword::ON)?;
         let table_name = self.parse_object_name()?;
-        let using = if self.expect_keyword(Keyword::USING).is_ok() {
+        let using = if self.parse_keyword(Keyword::USING) {
```
alamb, line 2752: no need to call the error producing variant and the ignore the error

Dandandan (approved):

PR #741, opened 2022-12-01
Clean up some redundant code in parser

alamb:
This is a  small code cleanup I noticed while reviewing https://github.com/sqlparser-rs/sqlparser-rs/pull/704 from @sarahyurick . Rather than go back and forth again for style on what was a long running PR I figured I would propose the change in a small follow on PR


sarahyurick:
Looks good to me, thanks!

Issue #687, opened 2022-10-25
Support `SELECT * EXCEPT(col1, col2, ...)` for BigQueryDialect


unvalley:
I’d like to parse `SELECT [* | column.*] EXCEPT(col1, col2, ...) FROM table` using BigQueryDialect.

BigQuery Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_except

The query `SELECT * EXCEPT (col1, col2) FROM table` causes ParserError as I described below.

```rust
use sqlparser::dialect::BigQueryDialect;
use sqlparser::parser::*;

fn main() {
    let sql = "SELECT * EXCEPT (col1, col2) FROM table";
    let dialect = BigQueryDialect {};
    let ast = Parser::parse_sql(&dialect, sql).unwrap();
    println!("AST: {:?}", ast);
}

// output: 
// thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected SELECT, VALUES, or a subquery in the query body, found: col1")', examples/parse_select.rs:21:48...
```

AugustoFKL:
@unvalley are you working on this already?

I just worked on `EXCLUDE` from snowflake on #721, and it seems like a really similar syntax. I can work on that as well.

unvalley:
@AugustoFKL No, I'm not working on this. 

> I just worked on EXCLUDE from snowflake on https://github.com/sqlparser-rs/sqlparser-rs/pull/721, and it seems like a really similar syntax. I can work on that as well.

Thanks! That helps a lot.

AugustoFKL:
Sure, I'll do that then. I just need to have my 5 PRs merged hahaha, otherwise I'll start pilling up conflicts, which I don't have the love of solving.

Once 2 or 3 are merged, I can go on that. Does that work for you?

unvalley:
@AugustoFKL No problem at all!

PR #745, opened 2022-12-02
feat: add support for except clause on wildcards

AugustoFKL:
Closes: #687


alamb:
Proposed follow on: https://github.com/AugustoFKL

PR #749, opened 2022-12-06
add keyword NANOSECOND

### LINKED ISSUE #748: support keyword `NANOSECOND`
i'd like to add a new keyword `NANOSECOND` so that we could parse something likes

`SELECT EXTRACT(NANOSECOND FROM d)`

waitingkuo:
close #748 

```rust
let ast = Parser::parse_sql(&dialect, "SELECT EXTRACT(NANOSECOND FROM NOW());").unwrap();
println!("AST: {:?}", ast);
```

AST: [Query(Query { with: None, body: Select(Select { distinct: false, top: None, projection: [UnnamedExpr(Extract { `field: Nanosecond`, expr: Function(Function { name: ObjectName([Ident { value: "NOW", quote_style: None }]), args: [], over: None, distinct: false, special: false }) })], into: None, from: [], lateral_views: [], selection: None, group_by: [], cluster_by: [], distribute_by: [], sort_by: [], having: None, qualify: None }), order_by: [], limit: None, offset: None, fetch: None, lock: None })]


src/keywords.rs
```diff
@@ -361,6 +361,8 @@ define_keywords!(
     MSCK,
     MULTISET,
     MUTATION,
+    NANOSECOND,
+    NANOSECONDS,
```
waitingkuo, line 365: add new keywords (follows the ascending order)

src/ast/value.rs
```diff
@@ -91,6 +91,8 @@ pub enum DateTimeField {
     Millennium,
     Millisecond,
     Milliseconds,
+    Nanosecond,
+    Nanoseconds,
```
waitingkuo, line 95: add new `DateTimeField`s

tests/sqlparser_common.rs
```diff
@@ -1740,6 +1740,9 @@ fn parse_extract() {
     verified_stmt("SELECT EXTRACT(HOUR FROM d)");
     verified_stmt("SELECT EXTRACT(MINUTE FROM d)");
     verified_stmt("SELECT EXTRACT(SECOND FROM d)");
+    verified_stmt("SELECT EXTRACT(MILLISECOND FROM d)");
+    verified_stmt("SELECT EXTRACT(MICROSECOND FROM d)");
+    verified_stmt("SELECT EXTRACT(NANOSECOND FROM d)");
```
waitingkuo, line 1745: add test cases

PR #750, opened 2022-12-06
Snowflake - allow the "type" key

yuval-illumex:
In continue of [693](https://github.com/sqlparser-rs/sqlparser-rs/pull/693) , the parser parse key:value syntax of snowflake. this pr adds support when they key is "type".


alamb (approved):
Thanks @yuval-illumex

PR #751, opened 2022-12-08
fix: stack overflow

### LINKED ISSUE #305: Stack overflow when parsing SQL consisting of many imbalanced open parenthesis

step-baby:
Closes #305


alamb:
Stacker looks quite impressive: https://crates.io/crates/stacker

AugustoFKL (approved):
@alamb really awesome. Never thought about that.

Althought, stackoverflow is dependant on the current architecture support, right? A machine with 2 GB of RAM would have this problem more usually than one with 32GB. Am I missing something?

Not sure if the depth being static is perfect, but seems ok for now IMHO.

src/ast/mod.rs
```diff
@@ -1534,1084 +1534,1086 @@ impl fmt::Display for Statement {
...
-            Statement::ExplainTable {
-                describe_alias,
-                table_name,
-            } => {
-                if *describe_alias {
-                    write!(f, "DESCRIBE ")?;
-                } else {
-                    write!(f, "EXPLAIN ")?;
-                }
+        stacker::maybe_grow(2 * 1024 * 1024, 5 * 1024 * 1024, || {
```
alamb: I was worried about this indent change because it will conflict with every other PR. 

So I pushed 853a890 as a way to make this diff smaller and shrink the merge conflict problem

alamb:
There is some issue building with no-std which I will try and resolve later this week

alamb:
> Althought, stackoverflow is dependant on the current architecture support, right? A machine with 2 GB of RAM would have this problem more usually than one with 32GB. Am I missing something?

stacks are normally much smaller than available memory -- specifically I think the default is like 8MB or so on Linux.

src/parser.rs
```diff
@@ -161,7 +161,9 @@ impl<'a> Parser<'a> {
                 return parser.expected("end of statement", parser.peek_token());
             }
 
-            let statement = parser.parse_statement()?;
+            let statement = stacker::maybe_grow(2 * 1024 * 1024, 5 * 1024 * 1024, || {
```
alamb, line 164: So looking at this code more carefully, I am not sure exactly how it is working -- specifically this call to `maybe_grow` grows the stack if there is less then 2MB remaining

In fact the test passes even if I remove the `maybe_grow` around this call (it fails if i remove the check on the `impl Display`)

alamb:
```diff
diff --git a/src/parser.rs b/src/parser.rs
index 5fd7af2..cff2281 100644
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -161,9 +161,7 @@ impl<'a> Parser<'a> {
                 return parser.expected("end of statement", parser.peek_token());
             }
 
-            let statement = stacker::maybe_grow(2 * 1024 * 1024, 5 * 1024 * 1024, || {
-                parser.parse_statement()
-            })?;
+            let statement = parser.parse_statement()?;
             stmts.push(statement);
             expecting_statement_delimiter = true;
         }
```

And then

```shell
cd /Users/alamb/Software/sqlparser-rs/ && RUST_BACKTRACE=1 CARGO_TARGET_DIR=/Users/alamb/Software/target-df cargo test -- overflow
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests src/lib.rs (/Users/alamb/Software/target-df/debug/deps/sqlparser-d27a7d1b89bd35ed)

running 1 test
test parser::tests::test_stack_overflow ... ok
```

🤔

alamb (changes requested):
I think we need to check the stack size at each expr boundary or something -- I don't think this PR implements overflow checking in the way intended, sadly

src/parser.rs
baoyachi: +1

alamb:
> Hi @alamb . Is there any better proposal? We urgently need this function.

hi @baoyachi, I think the approach sketched out by @46bit in  https://github.com/sqlparser-rs/sqlparser-rs/pull/501 is the best way to go. I will try and revive the code in that PR today. 

Perhaps you can give it a look and see if it would work for your purposes when ready

alamb:
@baoyachi  and @step-baby  -- here is my alternate proposal:  https://github.com/sqlparser-rs/sqlparser-rs/pull/764

alamb:
merged alternate proposal https://github.com/sqlparser-rs/sqlparser-rs/pull/764 - thanks for sparking this discussion @step-baby

PR #752, opened 2022-12-08
support `DROP FUNCTION` syntax

### LINKED ISSUE #753: Postgres: does not support DROP FUNCTION
```rust
fn test() {
    let sql = "DROP FUNCTION IF EXISTS test_func";

    println!("AST: {:?}", ast);
}
```
```
panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected TABLE, VIEW, INDEX, ROLE, SCHEMA, or SEQUENCE after DROP, found: FUNCTION")', examples/parse_select.rs:23:48
```

zidaye:
close #753 
refer to :https://www.postgresql.org/docs/current/sql-dropfunction.html


src/ast/mod.rs
```diff
@@ -1247,6 +1247,14 @@ pub enum Statement {
         /// deleted along with the dropped table
         purge: bool,
     },
+    /// DROP Function
+    DropFunction {
+        if_exists: bool,
+        /// One or more function to drop
+        func_desc: Vec<DropFunctionDesc>,
+        cascade: bool,
```
AugustoFKL, line 1255: @zidaye maybe an enum here would be better (or an option with a boolean inside).

Here you allow the state "cascade + restrict", which doesn't make sense at all.

AugustoFKL (changes requested):

src/ast/mod.rs
zidaye: @AugustoFKL done. very detailed comment

src/ast/mod.rs
```diff
@@ -1247,6 +1247,14 @@ pub enum Statement {
         /// deleted along with the dropped table
         purge: bool,
     },
+    /// DROP Function
+    DropFunction {
+        if_exists: bool,
+        /// One or more function to drop
+        func_desc: Vec<DropFunctionDesc>,
```
alamb, line 1254: What do you think about reusing the `CreateFunctionArg` structure used in CREATE FUNCTION

https://github.com/sqlparser-rs/sqlparser-rs/blob/d420001c37eeb38aac26a85dd5661755d6ef4228/src/ast/mod.rs#L1410-L1417

?

I can also see the argument for using a separate struct. If we stay with a separate struct, I suggest we name if `DropFunctionArg` to mirror the `CreateFunctionArg`

alamb (commented):
Thank you for the contribution @zidaye 

I have some suggestions, but overall this PR is looking close

src/ast/mod.rs
```diff
@@ -1247,6 +1247,14 @@ pub enum Statement {
...
+        /// `CASCADE` or `RESTRICT`
+        option: Option<DropFunctionOption>,
```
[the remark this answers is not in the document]
zidaye: I will change it later, thanks for your suggestion

src/ast/mod.rs
zidaye: I'll try to combine these structures later

alamb:
marking as draft -- please mark as ready for review when you have made the changes you plan. Thank you

src/ast/mod.rs
zidaye: sorry for my late reply。I've reconsidered your suggestion, and I think the structure might be acceptable. The `DropFunctionDesc` contains the function name and function parameters. The `CreateFunctionArg` only describes function parameters. Also, `CreateFunctionArg` has been reused and renamed `OperateFunctionArg`

src/ast/mod.rs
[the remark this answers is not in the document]
zidaye: > It might be possible to use https://github.com/sqlparser-rs/sqlparser-rs/blob/d420001c37eeb38aac26a85dd5661755d6ef4228/src/ast/ddl.rs#LL602C16-L602C16

done

src/ast/mod.rs
zidaye: > What do you think about reusing the `CreateFunctionArg` structure used in CREATE FUNCTION
> 
> https://github.com/sqlparser-rs/sqlparser-rs/blob/d420001c37eeb38aac26a85dd5661755d6ef4228/src/ast/mod.rs#L1410-L1417
> 
> ?
> 
> I can also see the argument for using a separate struct. If we stay with a separate struct, I suggest we name if `DropFunctionArg` to mirror the `CreateFunctionArg`

sorry for my late reply。I've reconsidered your suggestion, and I think the structure might be acceptable. The `DropFunctionDesc` contains the function name and function parameters. The `CreateFunctionArg` only describes function parameters. Also, `CreateFunctionArg` has been reused and renamed `OperateFunctionArg`

alamb (approved):
Thank you  @zidaye

PR #755, opened 2022-12-08
support create function definition with `$$`

### LINKED ISSUE #754: [Postgres Create Function]: does not support `$$`
```rust
fn test() {
    let sql = "CREATE OR REPLACE FUNCTION increment(i INTEGER) RETURNS INTEGER  LANGUAGE plpgsql AS $$ BEGIN RETURN i + 1; END; $$;";

    println!("AST: {:?}", ast);
}
```
```
panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected literal string, found: $")', examples/parse_select.rs:23:48
```

zidaye:
close #754

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3777,6 +3777,23 @@ impl fmt::Display for FunctionBehavior {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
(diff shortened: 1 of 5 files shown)


alamb (approved):
LGTM -- thank you @zidaye 

I verified this is in the postgresql dialect: https://www.postgresql.org/docs/current/sql-createfunction.html 👍

PR #756, opened 2022-12-09
support create trigger

zidaye:
refer to ：https://www.postgresql.org/docs/current/sql-createtrigger.html


src/parser.rs
```diff
@@ -2302,6 +2323,166 @@ impl<'a> Parser<'a> {
...
+                        if lparen_count == 0 {
+                            break;
+                        }
+                        condition_str.push_str(next_token.token.to_string().as_str());
+                    }
+                }
+                Some(condition_str)
+            } else {
+                None
+            };
```
alamb, line 2384: I think this code would become much simpler when `condition` is an `Expr`

Something like `let condition = self.parse_expr()`

alamb (commented):
Thank you @zidaye  -- this PR appears to have some conflicts. Can you please resolve them?

I also think `condition` should be an Expr

alamb:
This looks very good @zidaye  -- thank you. Please mark this as ready for review when it is ready for another look

alamb:

LucaCappelletti94:
Hi - I need this feature, what is the status on this?

alamb:
I don't think there is any update to report

PR #758, opened 2022-12-13
Add manually implemented AST visitor

tustvold:
Related to #78, #114, #601

This implements a basic AST visitor. I'm not convinced this is the best way to implement this, as it is relatively fragile, but I wanted to get something up to acts as a comparison point

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -16,12 +16,15 @@
 #[cfg(not(feature = "std"))]
 use alloc::{boxed::Box, string::String, vec::Vec};
 use core::fmt;
+use std::ops::ControlFlow;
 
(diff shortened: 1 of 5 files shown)


alamb:
So what I would *really* like to do is to figure out some way to automatically generate this code (like using `rustc_parse` or something) and then automatically generate visitors (and mutators) as part of the build.rs system somehow

tustvold:
> So what I would really like to do is to figure out some way to automatically generate this code

I'll have a play with what a proc-macro / codegen setup looks like

tustvold:
I think the approach in #765 is much more maintainable that this, which I think would be non-trivial to maintain and would act as a potential barrier to contribution

PR #759, opened 2022-12-13
Generalize locking clause

audunska:
Postgres supports more generalized locking clauses, for example: FOR UPDATE OF <table_name> SKIP LOCKED

also, multiple locking clauses. Generalize the parser to support these.

This _might_ be a breaking change, since we're adding new keywords `SKIP` , `LOCKED` and `NOWAIT`. What's the policy on that?


alamb (approved):
Thank you for the contribution @audunska  -- I looked at

https://www.postgresql.org/docs/current/sql-select.html

And this PR appears to implement that grammar 👍 


> This might be a breaking change, since we're adding new keywords SKIP , LOCKED and NOWAIT. What's the policy on that?

It is fine and expected -- we release new versions that do not get automatically semver updated.

alamb:
For some reason the CI checks did not run on this PR . I took the liberty of merging up from main to see if I can get them to run

PR #762, opened 2022-12-14
Support `ON CONFLICT ON CONSTRAINT`

audunska:
Postgres supports `ON CONFLICT ON CONSTRAINT <constraint_name>` to explicitly name the constraint that fails. Support this.


alamb (approved):
Thank you @audunska

PR #763, opened 2022-12-14
Support json operators `@>` `<@` `#-` `@?` and `@@`

audunska:
postgres supports a bunch more json operators. See https://www.postgresql.org/docs/15/functions-json.html

Skipping operators starting with a question mark for now, since those are hard to distinguish from placeholders without more context.

The precedence is chosen a bit at random. Maybe the `#-` operator should have precedence 30 like plus and minus. But these json-specific operators also shouldn't mix with numeric operators, so I don't know if it's a problem.


alamb (approved):
Looks good to me -- thank you @audunska

PR #764, opened 2022-12-14
Avoid stack overflows via configurable `with_recursion_limit`

alamb:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/305. Alternative to https://github.com/sqlparser-rs/sqlparser-rs/pull/751

This PR is based off the approach from  @46bit in #501 and includes some code and tests from there. While it may look like a large change, it is mostly docstring changes and tests. The code changes are very small

# Rationale
Given a deeply nested input, `sqlparser` will overflow the stack, resulting in a panic/segfault.

# Changes
1. Add a configurable maximum recursion limit and then check that limit during each call to `parse_expr`,  `parse_statement` and `parse_query`
2. Changes how the Parser is configured to a builder style (and adds docstrings to make it easier to figure out)
3. Improve docstrings (with several examples)
4. Add tests for recursion limit

# API Changes
The `Parser::new()` signature is change, but it is now documented. All other changes are to unreleased code added in https://github.com/sqlparser-rs/sqlparser-rs/pull/710 by @ankrgyl 

# Notes
The only recursive nested structures are `Expr`, `Statement`, and `Query` (as they are the places that are `Box`ed in the AST). Thus it is sufficient to check recursion depth at each of these points

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -37,6 +37,7 @@ use crate::tokenizer::*;
 pub enum ParserError {
     TokenizerError(String),
     ParserError(String),
+    RecursionLimitExceeded,
 }
 
 // Use `Parser::expected` instead, if possible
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -55,6 +56,92 @@ macro_rules! return_ok_if_some {
     }};
 }
 
+#[cfg(feature = "std")]
+/// Implemenation [`RecursionCounter`] if std is available
```
alamb, line 60: This is the key change for actually limiting recursion depth

src/parser.rs
```diff
@@ -55,6 +56,92 @@ macro_rules! return_ok_if_some {
...
+    use core::sync::atomic::{AtomicUsize, Ordering};
+    use std::rc::Rc;
+
+    use super::ParserError;
+
+    /// Tracks remaining recursion depth. This value is decremented on
+    /// each call to `try_decrease()`, when it reaches 0 an error will
+    /// be returned.
+    ///
+    /// Note: Uses an Rc and AtomicUsize in order to satisfy the Rust
```
alamb, line 71: The Rc approach is from @46bit -- if anyone has ideas about how to avoid it, I would love a PR to help.

src/parser.rs
```diff
@@ -55,6 +56,92 @@ macro_rules! return_ok_if_some {
...
+            // ran out of space
+            if old_value == 0 {
+                Err(ParserError::RecursionLimitExceeded)
+            } else {
+                Ok(DepthGuard::new(Rc::clone(&self.remaining_depth)))
+            }
+        }
+    }
+
+    /// Guard that increass the remaining depth by 1 on drop
```
alamb, line 105: The `DepthGuard` is used to automatically ensure the recursion depth is  restored upon function return

src/parser.rs
```diff
@@ -104,22 +192,78 @@ impl fmt::Display for ParserError {
...
+    /// # use sqlparser::{parser::{Parser, ParserError}, dialect::GenericDialect};
+    /// # fn main() -> Result<(), ParserError> {
+    /// let dialect = GenericDialect{};
+    /// let statements = Parser::new(&dialect)
+    ///   .try_with_sql("SELECT * FROM foo")?
+    ///   .parse_statements()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn new(dialect: &'a dyn Dialect) -> Self {
```
alamb, line 224: the signature here has changed - I tried to illustrate the intended usage with doc comments

src/parser.rs
```diff
@@ -128,49 +272,84 @@ impl<'a> Parser<'a> {
...
+    /// See example on [`Parser::new()`] for an example
+    pub fn try_with_sql(self, sql: &str) -> Result<Self, ParserError> {
+        debug!("Parsing sql '{}'...", sql);
+        let mut tokenizer = Tokenizer::new(self.dialect, sql);
+        let tokens = tokenizer.tokenize()?;
+        Ok(self.with_tokens(tokens))
     }
 
-    /// Parse a SQL statement and produce an Abstract Syntax Tree (AST)
-    pub fn parse_sql(dialect: &dyn Dialect, sql: &str) -> Result<Vec<Statement>, ParserError> {
```
alamb, line 144: github mangles the diff -- this function still exists with the same signature. It now also has docstrings

src/parser.rs
```diff
@@ -128,49 +272,84 @@ impl<'a> Parser<'a> {
...
+    /// let dialect = GenericDialect{};
+    /// let statements = Parser::new(&dialect)
+    ///   // Parse a SQL string with 2 separate statements
+    ///   .try_with_sql("SELECT * FROM foo; SELECT * FROM bar;")?
+    ///   .parse_statements()?;
+    /// assert_eq!(statements.len(), 2);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn parse_statements(&mut self) -> Result<Vec<Statement>, ParserError> {
```
alamb, line 306: This code was factored out of `parse_sql` so that people directly using `Parser::new()` could also have access to that logic

alamb:
cc @AugustoFKL @andygrove and @Dandandan

AugustoFKL (approved):

PR #765, opened 2022-12-15
Add derive based AST visitor

tustvold:
Reimplements #758 but using a proc-macro.

One issue is that currently this will not dispatch to `visit_table_name` for statements where the `ObjectType` is parameterised, this is `Drop`, `ShowCreate`, `Comment`. We could theoretically add a `skip_if` option down the line, but I wanted to keep things simple for now.


alamb:
Thanks @tustvold  -- I will check this one out carefully, but probably not until tomorrow

alamb:
Thank you -- I plan to review this carefully later today

Cargo.toml
```diff
@@ -32,6 +33,7 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 # of dev-dependencies because of
 # https://github.com/rust-lang/cargo/issues/1596
 serde_json = { version = "1.0", optional = true }
+sqlparser_derive = { version = "0.1", path = "derive", optional = true }
```
alamb, line 36: Note to myself: if we merge this PR I think we should document this feature flag -- it seems documentation is missing for the existing feature flags

derive/README.md
```diff
@@ -0,0 +1,79 @@
+# SQL Parser Derive Macro
+
+## Visit
+
+This crate contains a procedural macro that can automatically derive implementations of the `Visit` trait
```
alamb, line 5: Note to self -- this should have a doc link back to the sqlparser crate

derive/src/lib.rs
```diff
@@ -0,0 +1,169 @@
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote, quote_spanned, ToTokens};
+use syn::spanned::Spanned;
+use syn::{
+    parse_macro_input, parse_quote, Attribute, Data, DeriveInput, Fields, GenericParam, Generics,
+    Ident, Index, Lit, Meta, MetaNameValue, NestedMeta,
+};
+
+#[proc_macro_derive(Visit, attributes(visit))]
```
alamb, line 9:
```suggestion
/// Implementation of `[#derive(Visit)]`
#[proc_macro_derive(Visit, attributes(visit))]
```

src/ast/data_type.rs
```diff
@@ -337,6 +341,7 @@ fn format_datetime_precision_and_tz(
 /// guarantee compatibility with the input query we must maintain its exact information.
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit))]
```
alamb, line 344: I am a big fan of this method as a way to minimize maintenance burden on new contributions

I picked a random struct and removed the `#[cfg_attr(feature = "visitor", derive(Visit))]` to see what would happen if it was forgotten

I think the error is fairly clear 👍 

```
   Compiling sqlparser v0.28.0 (/Users/alamb/Software/sqlparser-rs)
error[E0277]: the trait bound `ast::Password: visitor::Visit` is not satisfied
    --> src/ast/mod.rs:1262:9
     |
1262 |         password: Option<Password>,
     |         ^^^^^^^^ the trait `visitor::Visit` is not implemented for `ast::Password`
     |
     = help: the following other types implement trait `visitor::Visit`:
               Box<T>
               CommentObject
               CreateTableBuilder
               Keyword
               Option<T>
               String
               Vec<T>
               ast::Action
             and 120 others
note: required for `Option<ast::Password>` to implement `visitor::Visit`
    --> src/ast/visitor.rs:21:16
     |
21   | impl<T: Visit> Visit for Option<T> {
     |                ^^^^^     ^^^^^^^^^
```

src/ast/visitor.rs
```diff
@@ -0,0 +1,242 @@
...
+visit_noop!(u8, u16, u32, u64, i8, i16, i32, i64, char, bool, String);
+
+#[cfg(feature = "bigdecimal")]
+visit_noop!(bigdecimal::BigDecimal);
+
+/// A visitor that can be used to walk an AST tree
+pub trait Visitor {
+    type Break;
+
+    /// Invoked for any tables, virtual or otherwise that appear in the AST
```
alamb, line 64: Can we please rename this to `visit_relation` (which is a more general term for table / view / CTE, etc)?

alamb (commented):
Thank you @tustvold  -- I just had a chance to go through this code in more detail 👍 

I think it works well for the usecase of visiting parts 

I also spent some time comparing it to the approach in https://github.com/sqlparser-rs/sqlparser-rs/pull/601 and I had a queston:

How hard do you think it would be to add the following features (in the future):

1. A `Rewriter` / mutator for performing SQL rewrites in addition to visiting,
2. A `previsit()` and `postvisit()` to specify the order in which children were visted (DFS or BFS)? 
3. Additional "visit" methods (e.g. if I wanted to visit all `Ident`s)? Could this be done in the proc macro itself? It seems like [Drive](https://docs.rs/derive-visitor/0.1.1/derive_visitor/trait.Drive.html) proposed in https://github.com/sqlparser-rs/sqlparser-rs/pull/601 can do so. 

I think 3. would be particularly compelling for other users of this crate in the near term.

cc @lovasoa -- I wonder if you have thoughts on this API  

cc @AugustoFKL  for your guidance

.gitignore
```diff
@@ -2,6 +2,7 @@
 # will have compiled files and executables
 /target/
 /sqlparser_bench/target/
+/derive/target/
```
alamb, line 5: Also note to self

I tested using

```shell
cargo test --all --features=visitor
```

We need to add this feature to the CI tests

tustvold:
> Additional "visit" methods (e.g. if I wanted to visit all Idents)? Could this be done in the proc macro itself? It seems like [Drive](https://docs.rs/derive-visitor/0.1.1/derive_visitor/trait.Drive.html) proposed in https://github.com/sqlparser-rs/sqlparser-rs/pull/601 can do so.

Very easy, it would simply be a case of adding the method to the `Visitor` and then potentially annotating a couple of fields and/or structs with `visit(with = "my_amazing_visit_method")`

> A previsit() and postvisit() to specify the order in which children were visted (DFS or BFS)?

Currently the code implements previsit, it would be relatively straightforward to add an additional annotation to support post visit.

> A Rewriter / mutator for performing SQL rewrites in addition to visiting,

You could definitely apply the same broad approach to this problem, but without having a good idea as to what the API might look like, or what the precise use-case would be, I'm not sure how easy this would be in practice. I suspect rewriting SQL AST directly would be relatively non-trivial to do correctly, not to mention verbose, with a far better approach to first lower to some sort of logical/intermediate representation.

src/ast/data_type.rs
tustvold: It's also making it easier to keep up with the merge conflicts :laughing:

lovasoa:
I would love this to be merged! 

This re-implements the entire derive-visitor crate inside sqlparser-rs, as opposed to my approach that just added it as a dependency. But I would still be very happy if this were merged and we finally had a way to visit sql statements !

derive/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "sqlparser_derive"
+description = "proc macro for sqlparser"
+version = "0.1.0"
+authors = ["Andy Grove <andygrove73@gmail.com>"]
```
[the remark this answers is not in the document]
AugustoFKL: TBH I'm not sure as well. Theoretically I'd say that the owner/creator should be the author, but since this is an expansion, not sure whether put both of you or just the @tustvold

AugustoFKL (commented):
@tustvold any reasons why the coverage is so small?

.gitignore
AugustoFKL: Oh, that explains the poor coverage I guess

derive/Cargo.toml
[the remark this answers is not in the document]
alamb: I think it is ok as is

alamb (approved):
I think we should do it. 

I'll plan to resolve outstanding conflicts and then add additional documentation / tests as a follow on PR. 

Thank you @tustvold @AugustoFKL and @lovasoa

alamb:
Thanks again all@

Cargo.toml
alamb: Docs in https://github.com/sqlparser-rs/sqlparser-rs/pull/774

derive/Cargo.toml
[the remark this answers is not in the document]
alamb: I changed it to "sqlparser authors" in https://github.com/sqlparser-rs/sqlparser-rs/pull/775

.gitignore
alamb: I verified that the CI does run with `--all-features`: https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/3794845435/jobs/6453351547

https://github.com/sqlparser-rs/sqlparser-rs/blob/dec3c2b818bdedcc81f0e3990e2dd2fefcf9e872/.github/workflows/rust.yml#L67

alamb:
Added Docs: https://github.com/sqlparser-rs/sqlparser-rs/pull/778

derive/README.md
alamb: in https://github.com/sqlparser-rs/sqlparser-rs/pull/779

PR #767, opened 2022-12-19
Support `ALTER INDEX {INDEX_NAME} RENAME TO {NEW_INDEX_NAME}`

### LINKED ISSUE #714: Support `ALTER INDEX` statement
Currently, It looks like `ALTER` keyword is supported for table only.

Can i try to implement `ALTER INDEX`?

## Goal
```sql
ALTER INDEX [ IF EXISTS ] name RENAME TO new_name
ALTER INDEX [ IF EXISTS ] name SET TABLESPACE tablespace_name
ALTER INDEX name ATTACH PARTITION index_name
ALTER INDEX name [ NO ] DEPENDS ON EXTENSION extension_name
ALTER INDEX [ IF EXISTS ] name SET ( storage_parameter [= value] [, ... ] )
ALTER INDEX [ IF EXISTS ] name RESET ( storage_parameter [, ... ] )
ALTER INDEX [ IF EXISTS ] name ALTER [ COLUMN ] column_number
    SET STATISTICS integer
ALTER INDEX ALL IN TABLESPACE name [ OWNED BY role_name [, ... ] ]
    SET TABLESPACE new_tablespace [ NOWAIT ]
```

## Ref
https://www.postgresql.org/docs/current/sql-alterindex.html

devgony:
## Goal
To resolve #714, this PR starts with `ALTER INDEX .. RENAME ..` to implemet ALTER INDEX base

## Todo
- [x] add AlterIndex match arm
- [x] fix unreachable with `Parser.expected`

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3618,10 +3618,10 @@ impl<'a> Parser<'a> {
                         let index_name = self.parse_object_name()?;
                         AlterIndexOperation::RenameIndex { index_name }
                     } else {
-                        todo!()
+                        return self.expected("TO after RENAME", self.peek_token());
                     }
                 } else {
-                    todo!()
+                    return self.expected("RENAME after ALTER INDEX", self.peek_token());
                 };
 
                 Ok(Statement::AlterIndex {
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3438,173 +3438,199 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_alter(&mut self) -> Result<Statement, ParserError> {
-        self.expect_keyword(Keyword::TABLE)?;
-        let _ = self.parse_keyword(Keyword::ONLY);
-        let table_name = self.parse_object_name()?;
-        let operation = if self.parse_keyword(Keyword::ADD) {
-            if let Some(constraint) = self.parse_optional_table_constraint()? {
-                AlterTableOperation::AddConstraint(constraint)
-            } else {
-                let if_not_exists =
-                    self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
-                if self.parse_keyword(Keyword::PARTITION) {
-                    self.expect_token(&Token::LParen)?;
-                    let partitions = self.parse_comma_separated(Parser::parse_expr)?;
-                    self.expect_token(&Token::RParen)?;
-                    AlterTableOperation::AddPartitions {
-                        if_not_exists,
-                        new_partitions: partitions,
-                    }
-                } else {
-                    let column_keyword = self.parse_keyword(Keyword::COLUMN);
-
-                    let if_not_exists = if dialect_of!(self is PostgreSqlDialect | BigQueryDialect | GenericDialect)
-                    {
-                        self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS])
-                            || if_not_exists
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -3438,173 +3438,199 @@ impl<'a> Parser<'a> {
...
-            name: table_name,
-            operation,
-        })
+
+                Ok(Statement::AlterIndex {
+                    name: index_name,
+                    operation,
+                })
+            }
+            _ => unreachable!(),
```
devgony, line 3632: How should we handle this unreachable? 🤔 
Is there a way to get fit enum of keywords that i only need (TABLE, INDEX)?

src/parser.rs
```diff
@@ -3438,173 +3438,199 @@ impl<'a> Parser<'a> {
...
-        let _ = self.parse_keyword(Keyword::ONLY);
-        let table_name = self.parse_object_name()?;
-        let operation = if self.parse_keyword(Keyword::ADD) {
-            if let Some(constraint) = self.parse_optional_table_constraint()? {
-                AlterTableOperation::AddConstraint(constraint)
-            } else {
-                let if_not_exists =
-                    self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
-                if self.parse_keyword(Keyword::PARTITION) {
+        let object_type = self.expect_one_of_keywords(&[Keyword::TABLE, Keyword::INDEX])?;
```
alamb, line 3441: reviewing with https://github.com/sqlparser-rs/sqlparser-rs/pull/767/files?w=1 ignoring whitespace made the changes much easier for me to see

src/parser.rs
```diff
@@ -3438,173 +3438,199 @@ impl<'a> Parser<'a> {
...
-            } else if self.parse_keywords(&[Keyword::DROP, Keyword::DEFAULT]) {
-                AlterColumnOperation::DropDefault {}
-            } else if self.parse_keywords(&[Keyword::SET, Keyword::DATA, Keyword::TYPE])
-                || (is_postgresql && self.parse_keyword(Keyword::TYPE))
-            {
-                let data_type = self.parse_data_type()?;
-                let using = if is_postgresql && self.parse_keyword(Keyword::USING) {
-                    Some(self.parse_expr()?)
+            Keyword::INDEX => {
+                let _ = self.parse_keyword(Keyword::ONLY);
```
alamb, line 3614: I think this is ignoring the error -- it should be something like
```suggestion
                self.parse_keyword(Keyword::ONLY)?;
```

However, it isn't clear to me why it is trying to parse an `ONLY` as part of an alter index statement

src/parser.rs
alamb: I think this is ok and a consequence of using `expect_one_of_keywords`

Perhaps a comment

```suggestion
            // unreachable because expect_one_of_keywords used above
            _ => unreachable!(),
```

alamb (approved):
Thank you @devgony  -- this looks good to me.

src/parser.rs
alamb: I see this is the same in some statements above. I'll clean it up in a follow on PR

alamb:
Thanks again @devgony

src/parser.rs
alamb: Removed in https://github.com/sqlparser-rs/sqlparser-rs/pull/773

PR #768, opened 2022-12-22
Parse exponent literal as number

Jefffrey:
Resolve part of https://github.com/sqlparser-rs/sqlparser-rs/issues/610

Allows parsing of exponent literals as numbers, provided the dialect does not allow an identifier to being with a number (Hive dialect). Will need more thinking on how to handle that specific case.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -144,6 +144,7 @@ pub fn all_dialects() -> TestedDialects {
             Box::new(RedshiftSqlDialect {}),
             Box::new(MySqlDialect {}),
             Box::new(BigQueryDialect {}),
+            Box::new(SQLiteDialect {}),
         ],
     }
 }
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -541,6 +541,7 @@ impl<'a> Tokenizer<'a> {
                     chars.next(); // consume the first char
                     let s = self.tokenize_word(ch, chars);
 
+                    // TODO: implement parsing of exponent here
                     if s.chars().all(|x| ('0'..='9').contains(&x) || x == '.') {
                         let mut inner_state = State {
                             peekable: s.chars().peekable(),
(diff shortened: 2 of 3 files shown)


src/test_utils.rs
```diff
@@ -144,6 +144,7 @@ pub fn all_dialects() -> TestedDialects {
             Box::new(RedshiftSqlDialect {}),
             Box::new(MySqlDialect {}),
             Box::new(BigQueryDialect {}),
+            Box::new(SQLiteDialect {}),
```
alamb, line 147: 👍

src/tokenizer.rs
```diff
@@ -617,6 +618,36 @@ impl<'a> Tokenizer<'a> {
                         return Ok(Some(Token::Period));
                     }
 
+                    // Parse exponent as number
+                    if chars.peek() == Some(&'e') || chars.peek() == Some(&'E') {
+                        let mut char_clone = chars.peekable.clone();
```
alamb, line 623: Why is this copy needed?

Given `chars` is already peekable I don't see why it can't be used directly

src/tokenizer.rs
```diff
@@ -1091,6 +1122,41 @@ mod tests {
...
+            Token::Whitespace(Whitespace::Space),
+            Token::Number(String::from("1e+10"), false),
+            Token::Comma,
+            Token::Whitespace(Whitespace::Space),
+            Token::Number(String::from("1"), false),
+            Token::make_word("ea", None),
+            Token::Comma,
+            Token::Whitespace(Whitespace::Space),
+            Token::Number(String::from("1e-10"), false),
+            Token::make_word("a", None),
```
alamb, line 1149: I found this very strange that a new token is formed without whitespace after a number. I expected that this is a token error but this implementation agrees with postgres 🤯 

```sql
postgres=# select 12e-10a;
      a       
--------------
 0.0000000012
(1 row)

postgres=# select 12e-10 a;
      a       
--------------
 0.0000000012
(1 row)
```

alamb: Likewise 

```sql
postgres=# select 1e-10-10;
   ?column?    
---------------
 -9.9999999999
(1 row)

postgres=# select 1e-10 -10;
   ?column?    
---------------
 -9.9999999999
(1 row)
```

🤯

alamb (approved):

alamb:
Thank you @Jefffrey

src/tokenizer.rs
Jefffrey: I needed a way to peek more than just the next char, since is only valid exponent if `e` followed by optional sign and an actual number. Found easiest way was to simply clone the iter and use that, and if found not to be an exponent and safely discard it and continue regular behaviour with original iter.

src/tokenizer.rs
Jefffrey: I believe this behaviour is part of what bit me when trying to implement for Hive dialect 😅

PR #769, opened 2022-12-23
Support redshift's columns definition list for system information functions

mskrzypkows:
pg_get_late_binding_view_cols
pg_get_cols
pg_get_grantee_by_iam_role
pg_get_iam_role_by_user


src/ast/mod.rs
```diff
@@ -3954,6 +3954,33 @@ impl fmt::Display for SearchModifier {
     }
 }
 
+/// A cols definition (i.e. `cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int)`
+/// when used with redshift pg_get_late_binding_view_cols/pg_get_cols)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ColsDefinition {
```
alamb, line 3961: Is there any reason not to use the existing `ColumnDef`? 

https://github.com/sqlparser-rs/sqlparser-rs/blob/61c661c234518ae02ae2c50d317e4640e4b90e26/src/ast/ddl.rs#L479-L486

I think that would make this PR significantly shorter as well as supporting other options

What do you think?

alamb (commented):
Thank you for the contribution @mskrzypkows !

src/ast/mod.rs
mskrzypkows: I think that the name is misleading. I should have named it e.g. `TableAliasDefinition`, because it defines what will be a result table, what column names it will have.
Like in the example:
```
select * from pg_get_late_binding_view_cols() cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);
view_schema | view_name | col_name   | col_type                    | col_num
------------+-----------+------------+-----------------------------+--------
public      | sales_lbv | salesid    | integer                     |       1
public      | sales_lbv | listid     | integer                     |       2
...
```
But I think it's a good idea to use `ColumnDef` inside the vector, instead of vector of `IdentPair`s.

What's your opinion?

alamb: > I think that the name is misleading. I should have named it e.g. TableAliasDefinition, because it defines what will be a result table, what column names it will have.

Sounds like a good idea to me

> But I think it's a good idea to use ColumnDef inside the vector, instead of vector of IdentPairs.


👍

mskrzypkows: I've tried using `ColumnDef` and failed because `name` is not a column type. It's strange, redshift doesn't define it as a keyword nor a type, but it's used in such a case. I think it's not a good idea to add it as additional type, so I'll rather back to `IdentPair`, or is there a better way to handle it?

src/parser.rs
```diff
@@ -5372,11 +5373,57 @@ impl<'a> Parser<'a> {
...
                 with_hints,
             })
         }
     }
 
+    fn parse_redshift_columns_definition_list(
+        &mut self,
+        name: &ObjectName,
+    ) -> Result<Option<ColsDefinition>, ParserError> {
+        if !dialect_of!(self is RedshiftSqlDialect) {
```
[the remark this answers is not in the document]
mskrzypkows: These functions are specific for the Redshift: https://docs.aws.amazon.com/redshift/latest/dg/r_System_information_functions.html so I think it's not needed for GenericDialect.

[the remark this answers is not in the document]
AugustoFKL: I know, but the Generic dialect is supposed to be the most permissive one. Unless there are syntax conflicts (e.g., accepting those would break another part of the code), I can't see why not.

AugustoFKL (commented):
@mskrzypkows there are some actions failing, could you please take a look?

alamb:

mskrzypkows:
> @mskrzypkows there are some actions failing, could you please take a look?

Sure, I'll fix it.

src/parser.rs
```diff
@@ -5372,11 +5373,57 @@ impl<'a> Parser<'a> {
...
+        let fname = name
+            .0
+            .last()
+            .ok_or_else(|| {
+                ParserError::ParserError("Empty identifier vector for ObjectName".to_string())
+            })?
+            .value
+            .to_lowercase();
+
+        if fname == "pg_get_late_binding_view_cols"
```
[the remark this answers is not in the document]
alamb: Yes I agree -- I would not expect this list to be hard coded. Instead I would expect any identifier to be accepted here and then the downstream crate would check against a list if it wanted. 

This design goal is explained in https://github.com/sqlparser-rs/sqlparser-rs#syntax-vs-semantics

src/ast/mod.rs
alamb: So I looked at the docs for

https://docs.aws.amazon.com/redshift/latest/dg/PG_GET_GRANTEE_BY_IAMROLE.html

I see in the example it has something like

```sql
select grantee, grantee_type, cmd_type 
FROM 
  pg_get_grantee_by_iam_role('arn:aws:iam::123456789012:role/Redshift-S3-Write') 
  res_grantee(grantee text, grantee_type text, cmd_type text) 
ORDER BY
 1,2,3;
```

I think this PR is related to this bit of the query:

```
  res_grantee(grantee text, grantee_type text, cmd_type text) 
```

I have several confusions:
1. Are you sure this construct is specific to the system information functions, or is a more general mechanism (it look like a more general mechanism that expands out a tuple to a table to me)
2. If it is a view definition, I would expect it to be parsed as ident, variable type not as just generic idents


Maybe someone could do some more research into what exactly this syntax construct in Redshift is called and what its rules are

mskrzypkows: 1. You're right, it looks like a more general mechanism. I've checked the documentation, but haven't found any general syntax description for it.
2. So if we have an additional `name` type, then it has to be added to the `DataType`, right?

alamb (approved):
I think this looks good to me now. Thank you @mskrzypkows 

This is one of the more unique syntaxes I have seen in SQL 🤯

alamb:
I took the liberty of merging up from main and fixing a (new) clippy lint on this PR

alamb:
This change seems to have caused a regression in 0.31.0 -- see https://github.com/sqlparser-rs/sqlparser-rs/issues/826. Here is a PR to back it out: https://github.com/sqlparser-rs/sqlparser-rs/pull/827

Looking more at the example https://docs.aws.amazon.com/redshift/latest/dg/PG_GET_GRANTEE_BY_IAMROLE.html

```sql
select grantee, grantee_type, cmd_type 
FROM 
  pg_get_grantee_by_iam_role('arn:aws:iam::123456789012:role/Redshift-S3-Write') 
  res_grantee(grantee text, grantee_type text, cmd_type text) 
ORDER BY
 1,2,3;
```

Rather than a special table definition syntax I think this is actually

```sql
FROM <table_function> <table_alias>(<col_alias1>, <col_alias2>, ...)
```

So 

```sql
select grantee, grantee_type, cmd_type 
FROM 
  pg_get_grantee_by_iam_role('arn:aws:iam::123456789012:role/Redshift-S3-Write') <-- this is "table function"
  res_grantee(grantee text, grantee_type text, cmd_type text) <-- `res_grantee` is the table alias, and the other things are column aliases
ORDER BY
 1,2,3;
```

I see perhaps the sqlparser column alias syntax doesn't support datatypes, maybe that could be extended

PR #771, opened 2022-12-25
Support grouping sets parsing for `GenericDialect`

Jefffrey:
Part of https://github.com/apache/arrow-datafusion/issues/2469

To allow GenericDialect to parse Grouping sets.

Note that will also now parse rollup & cube as their Expr enum variant now, instead of as Expr::Function


alamb (approved):
Thank you @Jefffrey

PR #772, opened 2022-12-27
feat: dollar-quoted strings support

### LINKED ISSUE #698: Support Postgresql dollar-quoted string $$constant$$
Hello,

I noticed that the double dollar quoted string is not supported by the parser. Only the single $ identifier is supported, thus the parsing fails all the time for sql texts with dollar quoted strings.
As it is a very convenient way to write complicated strings, dollar-quoting is largely used in the Postgres community. Could you please support it in the Postgres dialect? 

Thanks in advance ;)

vasilev-alex:
Previously, there was the `DoubleDollarQuoting` token, and it was used only in the context of functions. But technically, dollar-quoting is another way to present strings. 
e.g.

```sql
SELECT $$Hello, World!$$;
SELECT $tag_name$Hello, World!$tag_name$;
```

That's why I think it's reasonable to introduce the `DollarQuotedString` token, which can hold such string values and their optional tag names.

I have not added any tests yet but will do it if otherwise, it looks ok.

Closes #698

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -155,8 +155,6 @@ pub enum Token {
     PGCubeRoot,
     /// `?` or `$` , a prepared statement arg placeholder
     Placeholder(String),
-    // todo: remove
-    /// `$$`, used for PostgreSQL create function definition
     // DoubleDollarQuoting,
     /// ->, used as a operator to extract json field in PostgreSQL
     Arrow,
@@ -244,7 +242,6 @@ impl fmt::Display for Token {
             Token::HashArrow => write!(f, "#>"),
             Token::HashLongArrow => write!(f, "#>>"),
             Token::AtArrow => write!(f, "@>"),
-            // Token::DoubleDollarQuoting => write!(f, "$$"),
             Token::ArrowAt => write!(f, "<@"),
             Token::HashMinus => write!(f, "#-"),
             Token::AtQuestion => write!(f, "@?"),
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -471,7 +471,6 @@ impl<'a> Tokenizer<'a> {
 
         let mut location = state.location();
         while let Some(token) = self.next_token(&mut state)? {
-            println!("{:?}", token);
             tokens.push(TokenWithLocation {
                 token,
                 location: location.clone(),
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -34,6 +34,10 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "visitor")]
 use sqlparser_derive::Visit;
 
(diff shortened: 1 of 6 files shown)


alamb (commented):
Thank you @vasilev-alex 
The basic idea looks good to me. I think with some tests it would be 👍

alamb:
Marking as draft to signify it is waiting on tests -- please mark as ready to review when it is

tests/sqlparser_postgres.rs
```diff
@@ -2507,3 +2507,34 @@ fn parse_drop_function() {
         }
     );
 }
+
+#[test]
+fn parse_dollar_quoted_string() {
+    let sql = "SELECT $$hello$$, $tag_name$world$tag_name$";
+    let select = pg().verified_only_select(sql);
```
alamb, line 2514: Can you please add a test for something like `$$Foo$Bar$$`?

```sql
postgres=# select $$Foo$Bar$$;
 ?column? 
----------
 Foo$Bar
(1 row)
```

Also, 

```sql
postgres=# select $$Foo$Bar$$Baz;
   baz   
---------
 Foo$Bar
(1 row)
```

src/tokenizer.rs
```diff
@@ -149,8 +152,7 @@ pub enum Token {
     PGCubeRoot,
     /// `?` or `$` , a prepared statement arg placeholder
     Placeholder(String),
-    /// `$$`, used for PostgreSQL create function definition
-    DoubleDollarQuoting,
+    // DoubleDollarQuoting,
```
alamb, line 155: I wonder if this should be removed?

alamb (commented):
Thanks @vasilev-alex -- this is looking close

src/tokenizer.rs
```diff
@@ -860,17 +859,35 @@ impl<'a> Tokenizer<'a> {
...
 
-            return if let Some('$') = chars.peek() {
+            let mut is_terminated = false;
+            let mut prev: Option<char> = None;
+
+            while let Some(&ch) = chars.peek() {
+                if prev == Some('$') {
+                    if ch == '$' {
+                        chars.next();
+                        is_terminated = true;
```
alamb, line 870: 👍

tests/sqlparser_postgres.rs
```diff
@@ -2507,3 +2507,84 @@ fn parse_drop_function() {
...
+    );
+
+    assert_eq!(
+        projection[3],
+        SelectItem::ExprWithAlias {
+            expr: Expr::Value(Value::DollarQuotedString(DollarQuotedString {
+                tag: None,
+                value: "Foo$Bar".into(),
+            })),
+            alias: Ident {
```
alamb, line 2556: 👍

alamb (approved):
Awsome -- thank you @vasilev-alex

PR #777, opened 2022-12-28
Fix `cargo docs / warnings` and add CI check

alamb:
This was bothering me when working on tests / docs for visitor

Rationale:
Fix all `cargo doc` warning, and add CI check to avoid errors in the future

Example CI test failure: https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/3795223499/jobs/6454127069

### DIFF
--- a/src/keywords.rs
+++ b/src/keywords.rs
@@ -11,9 +11,7 @@
 // limitations under the License.
 
 //! This module defines
-//! 1) a list of constants for every keyword that
-//! can appear in [Word::keyword]:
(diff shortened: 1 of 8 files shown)


src/keywords.rs
```diff
@@ -11,9 +11,7 @@
 // limitations under the License.
 
 //! This module defines
-//! 1) a list of constants for every keyword that
-//! can appear in [Word::keyword]:
```
alamb, line 15: This appears to be referring to some older version of the code

PR #781, opened 2022-12-29
ParserError if nested explain

### LINKED ISSUE #744: Return an error when the SQL query contains nested `Explain`
In datafusion, we met a bug (https://github.com/apache/arrow-datafusion/issues/4378) which the datafusion-CLI will panic when the query contains nested `EXPLAIN`.

We fixed this bug using a workaround way, and currently the datafusion will return an `Internal` error when meeting these invalid `explain` queries. However the error is generated when converting a logical plan to a physical plan, which is too late.

We'd like the sqlparser to return a parsing error when parsing these invalid query string.
An ideal example is:
```
> explain explain select 1;
ParseError ...

Jefffrey:
Closes #744

Return ParserError during parsing if nested explain found

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4693,21 +4693,24 @@ impl<'a> Parser<'a> {
             format = Some(self.parse_analyze_format()?);
         }
 
-        if let Some(statement) = self.maybe_parse(|parser| parser.parse_statement()) {
-            Ok(Statement::Explain {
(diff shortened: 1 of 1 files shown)


alamb (approved):
LGTM

alamb:
Thank you for the contribution @Jefffrey

PR #782, opened 2022-12-31
Add a mutable visitor

lovasoa:
Happy new year 🥳 

This is a follow-up on #765 .

This adds the ability to mutate the AST as it is visited. I didn't remove the old immutable-only visitor trait.
Previously, only visitors taking an immutable reference to the visited structures were allowed.

### DIFF
--- a/src/ast/visitor.rs
+++ b/src/ast/visitor.rs
@@ -298,6 +298,14 @@ impl<E, F: FnMut(&ObjectName) -> ControlFlow<E>> Visitor for RelationVisitor<F> 
     }
 }
 
+impl<E, F: FnMut(&mut ObjectName) -> ControlFlow<E>> VisitorMut for RelationVisitor<F> {
+    type Break = E;
+
+    fn pre_visit_relation(&mut self, relation: &mut ObjectName) -> ControlFlow<Self::Break> {
+        self.0(relation)
+    }
+}
+
 /// Invokes the provided closure on all relations (e.g. table names) present in `v`
 ///
 /// # Example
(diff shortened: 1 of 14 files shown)


src/ast/visitor.rs
```diff
@@ -262,6 +430,36 @@ where
...
+/// # Example
+/// ```
+/// # use sqlparser::parser::Parser;
+/// # use sqlparser::dialect::GenericDialect;
+/// # use sqlparser::ast::{Expr, visit_expressions_mut, visit_statements_mut};
+/// # use core::ops::ControlFlow;
+/// let sql = "SELECT (SELECT y FROM z LIMIT 9) FROM t LIMIT 3";
+/// let mut statements = Parser::parse_sql(&GenericDialect{}, sql).unwrap();
+///
+/// // Remove all select limits in sub-queries
```
alamb, line 444: 👍

src/ast/visitor.rs
```diff
@@ -166,6 +210,84 @@ pub trait Visitor {
     }
 }
 
+/// A visitor that can be used to mutate an AST tree.
```
alamb, line 213: One thing that might be worth considering is to change this from mutating to rewriting by taking ownership.

Having such a signature  could be more efficient in cases where an owned value is needed, for example, rather than

```rust
impl VisitorMut for Replacer {
  type Break = ();
  fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<Self::Break> {
    if let Expr::Identifier(Ident{ value, ..}) = expr {
      // rewrite the expr to function call
      *expr = Expr::FunctionCall(name: value.clone(), ...)
    }
    ControlFlow::Continue(())
  }
}
```

One could write something like the following (reusing the allocated value):

```rust
impl VisitorRewrite for Rewriter {
  fn post_visit_expr(&mut self, mut expr: Expr) -> Result<Expr, Break> {
    if let Expr::Identifier(Ident{ value, ..}) = expr {
      Ok(Expr::FunctionCall(name: value, ...)
    } else {
      Ok(expr)
    }
}
```

alamb (approved):
Looks good to me -- thank you @lovasoa 

Happy new year to you as well 🌟 

cc @tustvold and @AugustoFKL 

Prior to releasing this code, I think we need to bump the version of the derive crate (semantically compatible), and make sqlparser depend on that higher version (to get the new derive_mut visitor)

src/ast/visitor.rs
lovasoa: I think copying all the potentially large elements in the AST multiple times in an out of functions would actually be less performant. And that would require to put something inside the AST while the function has the ownership of the elements, requiring everything to be `Default`, taking time to instantiate values that will never be read, and leaving the statement in an unexpected state if the function panics.

lovasoa: In the example you give, you don't need to clone the ident value, you can just [`take`](https://doc.rust-lang.org/std/mem/fn.take.html) it.

lovasoa:
@alamb, I think taking the values by ownership would require a significant amount of work, and not bring performance gains. 
For the version bumps, I'll make a new commit.

src/ast/visitor.rs
tustvold: FWIW the sizes are

```
#[test]
fn test() {
    assert_eq!(std::mem::size_of::<Statement>(), 712);
    assert_eq!(std::mem::size_of::<Query>(), 536);
    assert_eq!(std::mem::size_of::<Expr>(), 144);
}
```

These are likely large enough that moving them in and out of functions may have a non-trivial overhead.

I also confirmed that Rust isn't able to elide these copies in the general case - https://rust.godbolt.org/z/3W8YfWM6q

That being said Rust inlines very aggressively and inlining the visitor body may eliminate this.

I don't feel strongly about either solution, particularly in the absence of any benchmarks either way. In either case a rewrite will always result in a memcpy, passing by reference avoids needing the compiler to inline and elide the move in the non-mutating case, but passing by value may help it better optimise other cases, I don't really know :sweat_smile:

lovasoa: I made a small toy version to compare the two, the by-value version is 4x slower even with a small struct: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2abc5db67462f0516fd37e5d865a9563

alamb:
> @alamb, I think taking the values by ownership would require a significant amount of work, and not bring performance gains.

I don't think I did a good job explaining the alternative

The idea is to use something like  https://github.com/apache/arrow-datafusion/blob/master/datafusion/expr/src/expr_rewriter.rs

However, I don't feel strongly about it -- the mutation approach looks good to me. 

> For the version bumps, I'll make a new commit.

Thank you https://github.com/sqlparser-rs/sqlparser-rs/pull/782/commits/3ac3c677ad7c1f1c8d9720ee4111efae93f1210c looks good

alamb:
> I made a small toy version to compare the two, the by-value version is 4x slower even with a small struct:

BTW I think the more appropriate comparison is with release mode, but even then the example is 2x slower with ownership

https://play.rust-lang.org/?version=stable&mode=release&edition=2021

lovasoa:
Great, thanks for merging! I'll use that in sqlpage. Do you think you can release a new version on crates.io soon?

alamb:
>  Great, thanks for merging! I'll use that in sqlpage. Do you think you can release a new version on crates.io soon?


Yes I will go make one now

alamb:
Preparing one now: https://github.com/sqlparser-rs/sqlparser-rs/pull/786

alamb:
https://crates.io/crates/sqlparser/0.30.0 ✅

PR #783, opened 2023-01-01
Allow parsing of mysql empty row inserts

### LINKED ISSUE #624: Support MySQL's empty row insert
MySQL supports default row inserts [1], which we don't, like:

```sql
INSERT INTO tb () VALUES ();
```

This returns: `ParserError("Expected identifier, found: )")`

Maybe the columns list could be an option, instead of a vector? This way we can differentiate between non-existent lists (`INSERT INTO tb VALUES ...`) and empty lists (`INSERT INTO tb () ...`).

[1]: https://dev.mysql.com/doc/refman/8.0/en/insert.html

Jefffrey:
Closes #624

Allows parsing of

```sql
INSERT INTO tb () VALUES ();
```

for MySql dialect

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1433,7 +1433,7 @@ impl<'a> Parser<'a> {
     ///
     /// [(1)]: Expr::MatchAgainst
     pub fn parse_match_against(&mut self) -> Result<Expr, ParserError> {
-        let columns = self.parse_parenthesized_column_list(Mandatory)?;
+        let columns = self.parse_parenthesized_column_list(Mandatory, false)?;
 
         self.expect_keyword(Keyword::AGAINST)?;
 
@@ -2419,7 +2419,7 @@ impl<'a> Parser<'a> {
         // general that the arguments can be made to appear as column
         // definitions in a traditional CREATE TABLE statement", but
         // we don't implement that.
-        let module_args = self.parse_parenthesized_column_list(Optional)?;
+        let module_args = self.parse_parenthesized_column_list(Optional, false)?;
         Ok(Statement::CreateVirtualTable {
             name: table_name,
             if_not_exists,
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -4505,11 +4505,17 @@ impl<'a> Parser<'a> {
     pub fn parse_parenthesized_column_list(
         &mut self,
         optional: IsOptional,
+        allow_empty: bool,
     ) -> Result<Vec<Ident>, ParserError> {
         if self.consume_token(&Token::LParen) {
-            let cols = self.parse_comma_separated(Parser::parse_identifier)?;
-            self.expect_token(&Token::RParen)?;
-            Ok(cols)
+            if allow_empty && self.peek_token().token == Token::RParen {
```
alamb, line 4511: 👍

alamb (approved):
Looks good to me - thanks @Jefffrey 

I actually think we could allow the empty values list for all dialects (not just MySQL) per the https://github.com/sqlparser-rs/sqlparser-rs#syntax-vs-semantics section

What do you think?

Jefffrey:
One thing I noticed however, was for postgresql it would be a parser exception. For example, when running:

```sql
insert into tb () values ();
```

Get exception

```
Caused by: org.postgresql.util.PSQLException: ERROR: syntax error at or near ")"
  Position: 17
```

Similarly, for the same in MSSql:

```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near ')'.
```

So seems to suggest that for these dialects, it shouldn't parse at all. Thus, I'm a bit hesitant on enabling for all dialects. Not to mention, that the methods modified (`parse_values()` and `parse_parenthesized_column_list()`) are used in more places than just for parsing `INSERT INTO` statements (e.g. I believe `parse_values()` is used for parsing `EXCEPT ()` for bigquery which requires it to be non-empty:

https://github.com/sqlparser-rs/sqlparser-rs/blob/86d71f21097c0a44e67cc1513466c573ac1f763c/tests/sqlparser_bigquery.rs#L314-L320

So those methods would still likely require the new `allow_empty` parameter so code might not be too simplified even if enabling the empty row insert for all dialects (unless refactor to a separate method specifically for `INSERT INTO () tb VALUES ()` statements, if possible)

PR #787, opened 2023-01-02
Add User to README

joocer:
Existing README welcomes PRs with additional users of this library - this PR adds a reference to Opteryx.

I also added a comma, I'm not sure if it was missing because of style or a typo but commas save lives so I added one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ $ cargo run --features json_example --example cli FILENAME.sql [--dialectname]
 ## Users
 
 This parser is currently being used by the [DataFusion] query engine,
-[LocustDB], [Ballista] and [GlueSQL].
+[LocustDB], [Ballista], [GlueSQL], and [Opteryx].
 
 If your project is using sqlparser-rs feel free to make a PR to add it
 to this list.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,7 +93,7 @@ $ cargo run --features json_example --example cli FILENAME.sql [--dialectname]
 ## Users
 
 This parser is currently being used by the [DataFusion] query engine,
-[LocustDB], [Ballista] and [GlueSQL].
+[LocustDB], [Ballista], [GlueSQL], and [Opteryx].
```
AugustoFKL, line 96: @joocer thanks! 

@alamb I don't think there are any standards around projects that use others, so I'm not sure if that's ok.

Anyway, maybe this session could be renamed to 'Acknowledgements'? Users seems a bit weird IMO

alamb: @AugustoFKL  I don't think "Acknowledgements" is quite right as we are not really acknowledging assistance from other projects. 

The point of this section, as I understand, is basically to encourage more use of the sqlparser crate by offering evidence that other projects have found it valuable enough to use. 

Perhaps we could change the heading (in a different PR) to something like 'Powered By' to follow https://arrow.apache.org/powered_by/ ?

alamb (approved):
Thank you @joocer  -- this looks great to me. 


>  but commas save lives so I added one.

😆

PR #788, opened 2023-01-04
Support MySQL Character Set Introducers

mskrzypkows:
https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html

Characters set introducers: _latin1, _binary, _utf8, _utf8mb4

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -437,7 +437,7 @@ pub enum Expr {
     Nested(Box<Expr>),
     /// A literal value, such as string, number, date or NULL
     Value(Value),
-    /// https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+    /// <https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html>
     IntroducedString { introducer: String, value: Value },
     /// A constant of form `<data_type> 'value'`.
     /// This can represent ANSI SQL `DATE`, `TIME`, and `TIMESTAMP` literals (such as `DATE '2020-01-01'`),
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -404,6 +404,8 @@ pub enum Expr {
     Nested(Box<Expr>),
     /// A literal value, such as string, number, date or NULL
     Value(Value),
+    /// https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+    IntroducedString { introducer: String, value: Value },
     /// A constant of form `<data_type> 'value'`.
     /// This can represent ANSI SQL `DATE`, `TIME`, and `TIMESTAMP` literals (such as `DATE '2020-01-01'`),
     /// as well as constants of other types (a non-standard PostgreSQL extension).
@@ -663,6 +665,7 @@ impl fmt::Display for Expr {
             Expr::Collate { expr, collation } => write!(f, "{} COLLATE {}", expr, collation),
             Expr::Nested(ast) => write!(f, "({})", ast),
             Expr::Value(v) => write!(f, "{}", v),
+            Expr::IntroducedString { introducer, value } => write!(f, "{} {}", introducer, value),
             Expr::TypedString { data_type, value } => {
                 write!(f, "{}", data_type)?;
                 write!(f, " '{}'", &value::escape_single_quote_string(value))
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -698,7 +698,7 @@ impl fmt::Display for Expr {
             Expr::Collate { expr, collation } => write!(f, "{expr} COLLATE {collation}"),
             Expr::Nested(ast) => write!(f, "({ast})"),
             Expr::Value(v) => write!(f, "{v}"),
-            Expr::IntroducedString { introducer, value } => write!(f, "{} {}", introducer, value),
+            Expr::IntroducedString { introducer, value } => write!(f, "{introducer} {value}"),
             Expr::TypedString { data_type, value } => {
                 write!(f, "{data_type}")?;
                 write!(f, " '{}'", &value::escape_single_quote_string(value))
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -696,6 +698,7 @@ impl fmt::Display for Expr {
             Expr::Collate { expr, collation } => write!(f, "{} COLLATE {}", expr, collation),
             Expr::Nested(ast) => write!(f, "({})", ast),
             Expr::Value(v) => write!(f, "{}", v),
+            Expr::IntroducedString { introducer, value } => write!(f, "{} {}", introducer, value),
```
AugustoFKL, line 701: @mskrzypkows is this supposed to always be spaced?

In simple strings, it seems that there's no space:

```sql
SELECT _latin1'abc';
```

While in strings like a binary string does have spaces:

```sql
SELECT _latin1 X'4D7953514C';
```

Am I missing something?

**OBS:** both examples are from your reference: https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html

mskrzypkows: The space doesn't matter, it may be both with or without.

mskrzypkows: Testing both versions here: https://github.com/sqlparser-rs/sqlparser-rs/pull/788/files#diff-c8a9c2f5aa6b5c0f71fde332e59342930b2d67697d6fa11b538a7fadbbe1fe78R1302

AugustoFKL: Oh, I see, sorry the confusion.

AugustoFKL (approved):
LGTM

@alamb would it be possible to access the reports of coverage for this and other PR's?

It feels really confusing sometimes, since we have only the percentage and files, but not exactly what is not covered. I know the report contains the exact uncovered changed lines.

alamb:
> @alamb would it be possible to access the reports of coverage for this and other PR's?



It would be fine with me -- how can I do it? Can you go to https://coveralls.io/builds/55632033 ?

alamb:

alamb:

AugustoFKL:
@alamb not perfectly sure. For some reason for me, the coverage is like this:

image

Is it the same for you?

alamb:
@AugustoFKL here is what https://coveralls.io/builds/55632033  looks like to me:

Screenshot 2023-01-13 at 10 01 10 AM

I wonder if it just took a while to update? Can you check again?

The code coverage job / infrastructure predates me in sqlparser

AugustoFKL:
@alamb try to open one of the files, please.

Here's an example of one coverage of one of my projects
image

alamb:
https://coveralls.io/builds/55632033/source?filename=src%2Fparser.rs

Appears to say "source not available"

Screenshot 2023-01-15 at 5 33 41 AM

src/tokenizer.rs
```diff
@@ -562,7 +566,16 @@ impl<'a> Tokenizer<'a> {
                         s += s2.as_str();
                         return Ok(Some(Token::Number(s, false)));
                     }
-                    Ok(Some(Token::make_word(&s, None)))
+
+                    // https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+                    if ch == '_' && dialect_of!(self is MySqlDialect) {
```
alamb, line 571: In general, I think it would be nice to support this also in the GenericDialect for consistency with other vendor specific features, but I don't think it is required.

src/tokenizer.rs
```diff
@@ -562,7 +566,16 @@ impl<'a> Tokenizer<'a> {
...
+                        const INTRODUCERS: [&str; 4] = ["_latin1", "_binary", "_utf8", "_utf8mb4"];
```
alamb, line 572: My reading of https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html suggests the value for the charset introducer can be any character set supported; Why are these 4 hard coded?

alamb (commented):
Thank you for the contribution -- this is looking close. My major question is about the hard coded list of introducers.

src/tokenizer.rs
mskrzypkows: That's good point! I'll fix it.

src/parser.rs
```diff
@@ -734,6 +734,17 @@ impl<'a> Parser<'a> {
                             Ok(Expr::CompoundIdentifier(id_parts))
                         }
                     }
+                    // string introducer https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+                    Token::SingleQuotedString(_)
+                    | Token::DoubleQuotedString(_)
+                    | Token::HexStringLiteral(_)
+                        if w.value.starts_with('_') =>
```
mskrzypkows, line 741: @alamb Moved detection of Introduced String here. So I check it form `Token::Word` this seems to be the most generic solution. What do you think?

mskrzypkows: Unfortunately, it turned out to be the wrong solution. In MySQL, it's possible to run a query like this one:
```
select _col1 'str' from test_table; 
```
In this case, `'str'` is just a column alias...
So, I need to check if the string introduces is one of the following stings: https://dev.mysql.com/doc/refman/8.0/en/charset-charsets.html with `_` prefix.

The question is where to check it. I may bring back `Token::StringIntroducer` or check it here, only when we have string token in front of the introducer. Personally, I would stick with the second option.

mskrzypkows: @alamb What do you think, does it fit to the sqlparser-rs, to have such a check for mysql string introducers?

alamb: I am sorry -- I am behind on sqlparser-rs reviews. I hope to catch up later this week

alamb (approved):
LGTM -- thank you @mskrzypkows and @AugustoFKL

alamb:
I took the liberty of merging up from main to resolve a conflict and fixing the clippy failure

alamb:
Thanks again @mskrzypkows  and sorry for the delay

PR #789, opened 2023-01-04
use post_* visitors for mutable visits

lovasoa:
My initial implementation of mutable visitors closely matched the existing immutable visitor implementations,
and used pre_* visitors (a mutable expression was first passed to the visitor, then its children were visited).

After some initial usage, I realize this is actually impractical, and can easily lead to infinite recursion when one isn't careful.

For instance a mutable visitor would replace x by f(x), then would be called again on the nested x and result in f(f(x)), then again resulting in f(f(f(x))), and so on.

So, this commit changes the behavior of the visit_*_mut functions to call the visitor on an expression only once all of its children have been visited.

It also makes the documentation more explicit and adds an example.

### DIFF
--- a/src/ast/visitor.rs
+++ b/src/ast/visitor.rs
@@ -301,7 +301,7 @@ impl<E, F: FnMut(&ObjectName) -> ControlFlow<E>> Visitor for RelationVisitor<F> 
 impl<E, F: FnMut(&mut ObjectName) -> ControlFlow<E>> VisitorMut for RelationVisitor<F> {
     type Break = E;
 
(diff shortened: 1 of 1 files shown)


AugustoFKL (approved):

alamb (approved):
Makes sense to me @lovasoa

alamb:
Thank you for the contribution

lovasoa:
Thanks for merging, @alamb !

alamb:
Thank *you* for the contribution @lovasoa

PR #794, opened 2023-01-10
fix(Dialect): fix a typo (precendence -> precedence)

SARDONYX-sard:


AugustoFKL (approved):
lgtm

TBH not sure wether we should remove this variable. Is it used anywhere else? The underscore before the name is quite a bother

SARDONYX-sard:
> lgtm
> 
> TBH not sure wether we should remove this variable. Is it used anywhere else? The underscore before the name is quite a bother

Thanks for the reply.

VSCode told us.
It appears to be used only for testing.
 - https://github.com/sqlparser-rs/sqlparser-rs/blob/main/tests/sqlparser_custom_dialect.rs#L72

AugustoFKL:
@SARDONYX-sard could you please remove it?

alamb (approved):

alamb:
> TBH not sure wether we should remove this variable. Is it used anywhere else? The underscore before the name is quite a bother

I think `_precedence` is part of the `Dialect` trait definition. As you say I don't think it is actually used but it is passed to the trait in the parser, and users may have implemented the `Dialect` trait in their own projects and used the `precedence` argument.

https://github.com/sqlparser-rs/sqlparser-rs/blob/3f874f4ab8ad17f5149b57183436a4cdf4cf0704/src/parser.rs#L1583

Here are the other uses
https://github.com/search?q=repo%3Asqlparser-rs%2Fsqlparser-rs%20parse_infix&type=code


Thus I think we should keep it in the trait

alamb:
Thank you @SARDONYX-sard  and @AugustoFKL

PR #798, opened 2023-01-23
feat: `SELECT * REPLACE <Expr> AS <Identifier>` for bigquery

togami2864:
Support: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_replace

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -395,7 +395,7 @@ pub struct WildcardAdditionalOptions {
     /// `[RENAME ...]`.
     pub opt_rename: Option<RenameSelectItem>,
     /// `[REPLACE]`
-    pub opt_replace:Option<ReplaceSelectItem>
+    pub opt_replace: Option<ReplaceSelectItem>,
 }
 
 impl fmt::Display for WildcardAdditionalOptions {
@@ -408,6 +408,9 @@ impl fmt::Display for WildcardAdditionalOptions {
         }
         if let Some(rename) = &self.opt_rename {
             write!(f, " {rename}")?;
+        }
+        if let Some(replace) = &self.opt_replace {
+            write!(f, " {replace}")?;
         }
         Ok(())
     }
@@ -544,13 +547,13 @@ pub enum ReplaceSelectItem {
     /// ```plaintext
     /// <col_name> AS <col_alias>
     /// ```
-    Single(IdentWithAlias),
+    Single(Box<SelectItem>),
     /// Multiple column names with aliases inside parenthesis.
     /// # Syntax
     /// ```plaintext
     /// (<col_name> AS <col_alias>, <col_name> AS <col_alias>, ...)
     /// ```
-    Multiple(Vec<IdentWithAlias>),
+    Multiple(Vec<Box<SelectItem>>),
 }
 
 impl fmt::Display for ReplaceSelectItem {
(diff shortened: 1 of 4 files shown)


AugustoFKL:
I'll review in a few hours, really sorry for the delay @togami2864. You're on fire the last weeks hehe

src/ast/query.rs
```diff
@@ -528,6 +533,46 @@ impl fmt::Display for ExceptSelectItem {
...
+/// Bigquery `REPLACE` information.
+///
+/// # Syntax
+/// ```plaintext
+/// REPLACE <new_expr> AS <col_name>
+/// ```
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum ReplaceSelectItem {
```
AugustoFKL, line 545: Two questions about this structure.

1. Why don't you have an enum with two fields, an expression (the <expression> field) and another with an identifier? 
2. Any reason why you put the `AS` as mandatory? BigQuery has it as non-mandatory, but this syntax you displayed here says otherwise.

If you have a select item, the user will have to validate that the select item is not an expression or wildcard, which isn't excellent as we are such that it isn't.
- You'll have to validate the column name expression, as it MUST be an identifier.

togami2864: 1. That make sense. It's true that having a select item needs additional validation. I'll modify the implementation.
2. I think `AS` is required in `SELECT * Replace` statement. This is the special case.
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_replace

AugustoFKL: @togami2864 check the syntax structure for the select statement: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax?hl=pt-br#select_list

```
select_all:
    [ expression. ]*
    [ EXCEPT ( column_name [, ...] ) ]
    [ REPLACE ( expression [ AS ] column_name [, ...] ) ]
```

`AS` is optional between the expression and its column, if I understood it correctly.

togami2864: Sorry, my bad. Required element is `expr` and `column_name`.

alamb:
It looks to me like this PR has some feedback that needs implementing, so marking it as draft. Please mark it as ready for review when it is next ready.

src/ast/query.rs
```diff
@@ -394,6 +394,8 @@ pub struct WildcardAdditionalOptions {
     pub opt_except: Option<ExceptSelectItem>,
     /// `[RENAME ...]`.
     pub opt_rename: Option<RenameSelectItem>,
+    /// `[REPLACE]`
```
alamb, line 397:
```suggestion
    /// `[REPLACE]`
    ///  BigQuery syntax: <https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_replace>
```

alamb (approved):
LGTM -- thank you @togami2864

ankrgyl:
(CC @alamb ) I noticed a small typo in this change while rebasing my code: the field is named `colum_name` instead of `column_name` in `ReplaceSelectElement`.

@alamb would it be appropriate to patch and fix this before a bunch of folks upgrade to the new release, or leave it as is?

alamb:
> @alamb would it be appropriate to patch and fix this before a bunch of folks upgrade to the new release, or leave it as is?

If someone wants to fix it up that would be great. I am would be happy to make another quick follow on release.

togami2864:
I made #822 to followup.

PR #802, opened 2023-02-03
feat: support byte string literal in bq

togami2864:
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#string_and_bytes_literals


src/tokenizer.rs
```diff
@@ -493,6 +499,25 @@ impl<'a> Tokenizer<'a> {
                     }
                     Ok(Some(Token::Whitespace(Whitespace::Newline)))
                 }
+                // BigQuery uses b or B for byte string literal
```
AugustoFKL, line 502: This is something related to BigQuery, as I couldn't find this on MySQL. Maybe we could make this dialect-specific so we avoid permissive syntax on other dialects...?

@alamb

AugustoFKL (changes requested):
Left a comment.

One thing that bothers me is that you didn't test the 'b' literal. Please add some tests for it, if possible.

src/tokenizer.rs
togami2864: How about use `dialect_of!` like this.
```rs
b @ 'B' | b @ 'b' if dialect_of!(self is BigQueryDialect | GenericDialect) => { ..
```

togami2864:
We can only choose between 'b' or 'B' when converting AST into string result for test. So, I didn't test it. Is there any good approach??

src/tokenizer.rs
AugustoFKL: Seems reasonable

AugustoFKL:
@togami2864 wym?

I'd test it by using 'verified statement' usually... Considering the select, you put as an example:

```sql
SELECT B'abc', B"abc"
```

The resulting query is equals to the input? If so, use the `verified_stmt` method to show that. This way we can assure how the output is based on the input

alamb (commented):
This PR is looking good to me - thank you for the work @togami2864  and your help getting this ready for merge @AugustoFKL 

It looks like CI is failing (though that may be resolved with a merge from main to get the clippy fixes). Is there anything else we are waiting on?

togami2864:
All tasks are done. There is nothing else we waiting on.

alamb (approved):
Looks good to me -- thank you @togami2864  and @AugustoFKL

PR #807, opened 2023-02-14
snowflake: add support for TRANSIENT keyword

mobuchowski:
This PR adds support for Snowflake's TRANSIENT tables: https://docs.snowflake.com/en/user-guide/tables-temp-transient#transient-tables.

Snowflake grammar: https://docs.snowflake.com/en/sql-reference/sql/create-table#syntax

### DIFF
--- a/src/ast/helpers/stmt_create_table.rs
+++ b/src/ast/helpers/stmt_create_table.rs
@@ -50,6 +50,7 @@ pub struct CreateTableBuilder {
     pub external: bool,
     pub global: Option<bool>,
     pub if_not_exists: bool,
+    pub transient: bool,
     pub name: ObjectName,
     pub columns: Vec<ColumnDef>,
     pub constraints: Vec<TableConstraint>,
@@ -78,6 +79,7 @@ impl CreateTableBuilder {
             external: false,
             global: None,
             if_not_exists: false,
+            transient: false,
             name,
             columns: vec![],
             constraints: vec![],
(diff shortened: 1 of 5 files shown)


AugustoFKL (commented):
Couldn't think of a better place to put this comment, so I'll leave it as general.

1. Can you please take a look into the linter? The CI is failing.
2. Considering the documentation you presented, the transient is exclusive to other keywords before the 'TABLE' keyword:
```sql
CREATE [ OR REPLACE ]
    [ { [ LOCAL | GLOBAL ] TEMP[ORARY] | VOLATILE } | TRANSIENT ]
```

Therefore, should the parse and print of it be flexible regarding both being present? You shouldn't accept `CREATE LOCAL TRANSIENT TABLE,` right...?

Finally, since this is something specific, what do you think about adding validation to check if the dialect is the SnoflakeDialect or the GenericDialect?

Thanks for the PR!

mobuchowski:
@AugustoFKL the lint failures seem to be unrelated to my PR at all and I think https://github.com/sqlparser-rs/sqlparser-rs/pull/803 fixes them. If you want to see full green before approving or merging I'd wait until it gets merged and I'll rebase on top of it.

Regarding validation, I've thought it's kind of to the contrary to project goals:

```
This crate provides only a syntax parser, and tries to avoid applying any SQL semantics, and accepts queries that specific databases would reject, even when using that Database's specific Dialect.
```

I, of course, can add it if we agree that validation and accepting keywords based on dialect is what we generally want to pursue.

alamb (approved):

mobuchowski:
Thanks @alamb!

PR #810, opened 2023-02-21
Add support for trailing commas

ankrgyl:
See context for the proposal in https://github.com/sqlparser-rs/sqlparser-rs/issues/809.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2243,7 +2243,11 @@ impl<'a> Parser<'a> {
                     {
                         break;
                     }
-                    Token::RParen | Token::SemiColon | Token::EOF => break,
+                    Token::RParen
+                    | Token::SemiColon
+                    | Token::EOF
+                    | Token::RBracket
+                    | Token::RBrace => break,
                     _ => continue,
                 }
             }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -230,16 +230,12 @@ impl<'a> Parser<'a> {
     /// # }
     /// ```
     pub fn new(dialect: &'a dyn Dialect) -> Self {
-        Self::new_with_options(dialect, ParserOptions::default())
-    }
-
-    pub fn new_with_options(dialect: &'a dyn Dialect, options: ParserOptions) -> Self {
         Self {
             tokens: vec![],
             index: 0,
             dialect,
             recursion_counter: RecursionCounter::new(DEFAULT_REMAINING_DEPTH),
-            options,
+            options: ParserOptions::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


ankrgyl:
I haven't added tests or docstrings yet but of course happy to after some initial feedback on the idea

AugustoFKL:
@ankrgyl, sorry for the delay. I'll take a look this weekend.

src/parser.rs
```diff
@@ -195,12 +195,20 @@ impl std::error::Error for ParserError {}
...
+    pub trailing_commas: bool,
+}
+
 pub struct Parser<'a> {
     tokens: Vec<TokenWithLocation>,
     /// The index of the first unprocessed token in `self.tokens`
     index: usize,
     /// The current dialect to use
     dialect: &'a dyn Dialect,
+    /// Additional options that allow you to mix & match behavior otherwise
```
alamb, line 209: I like this idea

src/parser.rs
```diff
@@ -222,11 +230,16 @@ impl<'a> Parser<'a> {
     /// # }
     /// ```
     pub fn new(dialect: &'a dyn Dialect) -> Self {
+        Self::new_with_options(dialect, ParserOptions::default())
+    }
+
+    pub fn new_with_options(dialect: &'a dyn Dialect, options: ParserOptions) -> Self {
```
alamb, line 236: What would you think about following the `with_recursion_limit` builder style pattern instead of adding another `new` function. Like this:

```rust
fn with_options(mut self, options: ParserOptions) -> Self {
...
}
```
?

src/parser.rs
```diff
@@ -2195,15 +2208,32 @@ impl<'a> Parser<'a> {
...
     /// Parse a comma-separated list of 1+ SelectItem
     pub fn parse_projection(&mut self) -> Result<Vec<SelectItem>, ParserError> {
+        // BigQuery allows trailing commas, but only in project lists
+        // e.g. `SELECT 1, 2, FROM t`
+        // https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#trailing_commas
+        //
+        // This pattern could be captured better with RAII type semantics, but it's quite a bit of
+        // code to add for just one case, so we'll just do it manually here.
+        let old_value = self.options.trailing_commas;
+        self.options.trailing_commas |= dialect_of!(self is BigQueryDialect);
```
alamb, line 2218: I agree a RAAI style thing would be better but this is fine too for now

alamb (commented):
Thanks @ankrgyl  -- I really like this feature. I left some comments but I think the idea is  💯

src/parser.rs
ankrgyl: 😄

src/parser.rs
ankrgyl: Great idea

src/parser.rs
ankrgyl: Cool

src/parser.rs
```diff
@@ -255,6 +264,31 @@ impl<'a> Parser<'a> {
         self
     }
 
+    /// Specify additional parser options
```
alamb, line 267: ❤️

alamb (approved):
looks good to me -- thank you @ankrgyl

PR #812, opened 2023-02-22
feat: support raw string literal of BigQuery

togami2864:
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_literals

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -45,6 +45,10 @@ pub enum Value {
     SingleQuotedByteStringLiteral(String),
     /// B"string value"
     DoubleQuotedByteStringLiteral(String),
+    /// R'string value' or r'string value'
+    SingleQuotedRawStringLiteral(String),
+    /// R"string value" or r"string value"
+    DoubleQuotedRawStringLiteral(String),
     /// N'string value'
     NationalStringLiteral(String),
     /// X'hex value'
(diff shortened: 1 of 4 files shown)


src/ast/value.rs
```diff
@@ -45,6 +45,10 @@ pub enum Value {
     SingleQuotedByteStringLiteral(String),
     /// B"string value"
     DoubleQuotedByteStringLiteral(String),
+    /// R'string value' or r'string value'
+    SingleQuotedRawStringLiteral(String),
+    /// R"string value" or r"string value"
+    DoubleQuotedRawStringLiteral(String),
```
AugustoFKL, line 51: @togami2864, this seems ok, but I want to raise a question.

Is the quote style of enormous importance here? Does it change the context to a level that the two variants will be handled individually? If `r'abc'` is different, semantically, from r"abc", that's fine. Otherwise, could only be held within a single variant, with an enum of quote styles? (It's a sincere question, if you don't think so, then we can keep it this way).

togami2864: The disadvantage is that the test becomes more complicated(due to canonical), but since there are no other disadvantages, I have merged the two values `RawStringLiteral`.

src/ast/value.rs
```diff
@@ -45,6 +45,8 @@ pub enum Value {
     SingleQuotedByteStringLiteral(String),
     /// B"string value"
     DoubleQuotedByteStringLiteral(String),
+    /// R'string value' or r'string value' or r"string value"
+    RawStringLiteral(String),
```
alamb, line 49:
```suggestion
    /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_literals>
    RawStringLiteral(String),
```

PR #827, opened 2023-03-06
Fix table alias parsing regression in 0.31.0 by backing out redshift column definition list

alamb:
Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/826

This fixes a table alias parsing regression by reverting the change from  https://github.com/sqlparser-rs/sqlparser-rs/pull/769 by @mskrzypkows and adding a regression test to protect against regressions in the future.

I will leave some additional notes on https://github.com/sqlparser-rs/sqlparser-rs/pull/769 about the cause


alamb:
cc @ankrgyl  and @mskrzypkows  I wonder if you have some time to review this PR

mskrzypkows (approved):
I think I have a better solution to this problem, so it's even better that it's reverted :-)

alamb:
Thanks for the quick turnaround @ankrgyl  and @mskrzypkows

PR #831, opened 2023-03-08
support IF EXISTS in COMMENT statements

pawel-big-lebowski:
SqlParser does already support `COMMENT` statements which were implemented within a postgres dialect. 
`Comment` statements are also supported within [Snowflake (see docs)](https://docs.snowflake.com/en/sql-reference/sql/comment) with an optional `IF EXISTS`. 

This PR allows clause `IF EXISTS` within currently supported `COMMENT` statements.


ankrgyl (approved):
The change LGTM. @alamb I think you're up next

src/ast/mod.rs
```diff
@@ -1487,6 +1487,8 @@ pub enum Statement {
         object_type: CommentObject,
         object_name: ObjectName,
         comment: Option<String>,
+        /// An optional `IF EXISTS` clause. (Non-standard.)
+        if_exists: bool,
```
alamb, line 1491:
```suggestion
        /// See <https://docs.snowflake.com/en/sql-reference/sql/comment>
        if_exists: bool,
```

alamb (approved):
Thanks @pawel-big-lebowski  and @ankrgyl !

PR #834, opened 2023-03-13
fix: merge byte string literal

togami2864:
As the same as https://github.com/sqlparser-rs/sqlparser-rs/pull/812#discussion_r1117987296, quote style isn't important.
This PR merges `SingleQuotedByteStringLiteral` and `DoubleQuotedByteStringLiteral` to `ByteStringLiteral`.

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -41,10 +41,8 @@ pub enum Value {
     /// See [Postgres docs](https://www.postgresql.org/docs/8.3/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS)
     /// for more details.
     EscapedStringLiteral(String),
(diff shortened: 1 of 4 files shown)


ankrgyl:
@alamb do you have thoughts on changes like this? I generally agree it's probably fine to only have one type of byte string literal, and the two new types are new (2/20/23) but there's a small risk of regressing someone's existing workflow.

alamb:
I don't see the harm in keeping these two variants separate -- if downstream crates want to treat them the same they can do so, and if they want to handle them separately they can with the current forumulation

Thus I would not be in favor of merging this unless there is some compelling usecase

togami2864:
That makes sense to me. I'll close this PR.

Issue #757, opened 2022-12-12
Plumb location info through the parser


ankrgyl:
This is a follow up to our discussion in https://github.com/sqlparser-rs/sqlparser-rs/pull/710#issuecomment-1338068455 (and also referenced in #301, #179, and #524.

Now that location info is plumbed through the tokenizer, we can leverage these locations to:

1. Produce better error messages in the parser (by referencing the offending token, and its location)
2. Propagate location information outside of the parser, so that libraries that consume the parse tree can reference the original location of an AST node.

From my understanding, the first task should be easier, since we wouldn't need to update the parse tree data structure (perhaps outside of a few edge cases). The main change should be updating the parser to use `TokenWithLocation` and then improving the `ParserError` messages to reference these locations. I would suggest changing the type of `TokenizerError` and `ParserError` to be richer than a string, and pass along the location information directly, so that consumers of the error can access it directly.

The second task will be more difficult, however, since we'll need to either change all of the tree nodes to be structs instead of enums (to host the location information), or change each enum branch to be a struct with location information. I think the second approach will likely be easier to change incrementally, and for users to refactor against, and we can add a trait that homogenizes how users of the tree access the location info.

@alamb let me know your thoughts on the above.

alamb:
Thank you @ankrgyl -- I really appreciate you pushing this project along. 

>  I would suggest changing the type of TokenizerError and ParserError to be richer than a string, and pass along the location information directly, so that consumers of the error can access it directly.

I agree this sound like a good plan

>  I think the second approach will likely be easier to change incrementally

I am not quite sure what this would look like -- could you share a draft PR or rust pseudo code?


cc @AugustoFKL

ankrgyl:
> I am not quite sure what this would look like -- could you share a draft PR or rust pseudo code?

Absolutely, here is a quick patch which shows what that would look like for `Ident` (note: this does not compile -- I did not change the parser yet):

```
ankur-m1:~/projects/sqlparser-rs ankur$ git diff
diff --git a/src/ast/mod.rs b/src/ast/mod.rs
index 7f3d154..14b2bc6 100644
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -22,6 +22,8 @@ use core::fmt;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};

+use crate::tokenizer::Location;
+
 pub use self::data_type::{
[...]

PR #839, opened 2023-03-19
Add Location/Spans in to the AST/parse layer

jakeswenson:
This is an attempt at starting to plumb a location (or in this case a `Span`) through to the parsing/AST layer: #757.

# Changes

1. Add a new wrapper type `WithSpan<T>` which can be used to wrap and attach a span to any AST node. Not sure about this approach yet, but it seemed to work OK. I think this might need to evolve in to a type called `WithTrivia<T>` where `Span` and "trivia items" like leading and trailing `Token`'s to an AST node are attached. (i.e. the `SELECT` keyword for a query, to get a span for the entire select query.) This "trivia" based approach is something I've learned from [dotnet/roslyn (the C# compiler)](https://github.com/dotnet/roslyn/blob/e47b15b2485bee978bafb5c7dcce99d8bdc2f8fd/src/Compilers/Core/Portable/Syntax/SyntaxTrivia.cs#L22). (This trivia based approach was also taken with typescript.)
2. Add a `Span` type which is a struct with a start (inclusive) and end (exclusive) `Location`

# ToDo still
- [ ] Finish wiring spans through the tests. This is going to be the most annoying part about this work. Would love to hear folks thoughts on this.

# Open Questions
- Should this work just align to using [zkat/miette](https://github.com/zkat/miette) `Diagnostic` and `SourceSpan`.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -21,6 +21,7 @@ use alloc::{
     vec::Vec,
 };
 use core::fmt;
+use std::cmp::max;
 
 use log::debug;
 
@@ -303,7 +304,7 @@ impl<'a> Parser<'a> {
             .into_iter()
             .map(|token| TokenWithLocation {
                 token,
-                location: Location { line: 0, column: 0 },
+                span: Span::default(),
             })
             .collect();
         self.with_tokens_with_locations(tokens_with_locations)
@@ -318,8 +319,8 @@ impl<'a> Parser<'a> {
     pub fn try_with_sql(self, sql: &str) -> Result<Self, ParserError> {
         debug!("Parsing sql '{}'...", sql);
         let mut tokenizer = Tokenizer::new(self.dialect, sql);
-        let tokens = tokenizer.tokenize()?;
-        Ok(self.with_tokens(tokens))
+        let tokens = tokenizer.tokenize_with_location()?;
+        Ok(self.with_tokens_with_locations(tokens))
     }
 
     /// Parse potentially multiple statements
@@ -495,7 +496,7 @@ impl<'a> Parser<'a> {
         let mut noscan = false;
         let mut partitions = None;
         let mut compute_statistics = false;
-        let mut columns = vec![];
+        let mut columns: Vec<Ident> = vec![];
         loop {
             match self.parse_one_of_keywords(&[
                 Keyword::PARTITION,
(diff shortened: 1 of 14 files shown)


ankrgyl:
@jakeswenson thanks for putting this up! Can you talk about how it's different than the approach in https://github.com/sqlparser-rs/sqlparser-rs/pull/790?

The key problem we're trying to work through right now is how to avoid breaking downstream users of sqlparser by introducing locations.

src/parser.rs
```diff
@@ -5842,14 +5875,17 @@ impl<'a> Parser<'a> {
...
-        let grantees = self.parse_comma_separated(Parser::parse_identifier)?;
+        let grantees =
+            self.parse_comma_separated(|p| p.parse_identifier().map(WithSpan::unwrap))?;
 
         let with_grant_option =
             self.parse_keywords(&[Keyword::WITH, Keyword::GRANT, Keyword::OPTION]);
 
         let granted_by = self
             .parse_keywords(&[Keyword::GRANTED, Keyword::BY])
-            .then(|| self.parse_identifier().unwrap());
```
jakeswenson, line 5852: I've fixed this previous possible `panic!` in the parser for these `GRANT` statement.

src/parser.rs
```diff
@@ -5973,11 +6009,14 @@ impl<'a> Parser<'a> {
         let (privileges, objects) = self.parse_grant_revoke_privileges_objects()?;
 
         self.expect_keyword(Keyword::FROM)?;
-        let grantees = self.parse_comma_separated(Parser::parse_identifier)?;
+        let grantees =
+            self.parse_comma_separated(|p| p.parse_identifier().map(WithSpan::unwrap))?;
 
         let granted_by = self
             .parse_keywords(&[Keyword::GRANTED, Keyword::BY])
-            .then(|| self.parse_identifier().unwrap());
```
jakeswenson, line 5980: another `panic!` fix in `REVOKE`

alamb:
cc @ankrgyl  who I believe was doing something similar

ankrgyl:
@alamb @jakeswenson yes indeed! I posted this above too, but it'd be great to understand the differences between this and #790. In general, I'd prefer to collaborate on that PR vs. work on two separate ones.

lustefaniak:
I used this PR in our forked https://github.com/getsynq/sqlparser-rs and a tool built on top of that, and it works amazingly well. Adding `empty_span()` all over the place is painful initially, but in the end, all works as it should and is not much of a burden. 

Since I don't have much Rust experience, I can't comment on any technical downsides(if any) of that design, but the "client" code required minimal changes to make it fully work.

lustefaniak:
After working on column-level lineage using sqlparser-rs even more, I think we must add location to every instance of `Ident` and maybe also `ObjectName` as otherwise it is super annoying to mark every Ident as `WithSpan<Ident>` and then patching all the tests 🤔

yuyang-ok:
Is this going to merge or not?

alamb:
> Is this going to merge or not?

I don't think this PR is mergeable as is (it has conflicts and doesn't pass CI)

As the Location/Span thing keeps coming up it would be awesome for someone to take a stab at implementing it. The core challenge, in my mind, of adding this feature is minimizing downstream crate impacts (sqlparser is used quite widely). I don't think we'll be able to avoid all impacts but being able to minimize the impact is important

PR #840, opened 2023-03-21
Non-Latin characters support

mskrzypkows:


alamb:
> This approach seems fine to me. @alamb is there any pre-existing context about the use of ASCII-only that could be relevant?

I do not have any such context. Maybe @andygrove @nickolay  @Dandandan  or @maxcountryman  would like to weigh in

alamb (approved):
Makes sense to me -- thank you @mskrzypkows  and @ankrgyl  for the review

nickolay:
For the MSSQL dialect I just didn’t need it personally. Figured it was better to leave it unimplemented, rather than produce a possibly buggy solution or read up on Unicode character classes in Rust. Thanks for figuring it out!

PR #843, opened 2023-04-05
snowflake: add support for LATERAL FLATTEN and similar

lustefaniak:
Add support for `LATERAL FLATTEN` without `TABLE()`

```sql
SELECT emp.employee_ID, emp.last_name, index, value AS project_name 
FROM employees AS emp, LATERAL FLATTEN(INPUT => emp.project_names) AS proj_names
```

https://docs.snowflake.com/en/sql-reference/constructs/join-lateral#example-of-using-lateral-with-flatten

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -5638,11 +5638,16 @@ impl<'a> Parser<'a> {
     /// A table name or a parenthesized subquery, followed by optional `[AS] alias`
     pub fn parse_table_factor(&mut self) -> Result<TableFactor, ParserError> {
         if self.parse_keyword(Keyword::LATERAL) {
-            // LATERAL must always be followed by a subquery.
-            if !self.consume_token(&Token::LParen) {
-                self.expected("subquery after LATERAL", self.peek_token())?;
+            // LATERAL must always be followed by a subquery or table function.
+            if self.consume_token(&Token::LParen) {
+                self.parse_derived_table_factor(Lateral)
+            } else {
+                let name = self.parse_object_name()?;
+                self.expect_token(&Token::LParen)?;
+                let args = self.parse_optional_args()?;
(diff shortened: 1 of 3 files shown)


lustefaniak:
I was thinking about this approach or adding a flag on TableFactor::Table to know if it is decorated with `TABLE( fn() )`.

src/parser.rs
```diff
@@ -5638,11 +5638,16 @@ impl<'a> Parser<'a> {
     /// A table name or a parenthesized subquery, followed by optional `[AS] alias`
     pub fn parse_table_factor(&mut self) -> Result<TableFactor, ParserError> {
         if self.parse_keyword(Keyword::LATERAL) {
-            // LATERAL must always be followed by a subquery.
-            if !self.consume_token(&Token::LParen) {
-                self.expected("subquery after LATERAL", self.peek_token())?;
+            // LATERAL must always be followed by a subquery or table function.
```
alamb, line 5641: Since this is a snowflake specific syntax, could you please add a check like

```
                if dialect_of!(self is SnowflakeDialect) {
```

?

alamb (commented):
Thank you for the contribution @lustefaniak

> I was thinking about this approach or adding a flag on TableFactor::Table to know if it is decorated with TABLE( fn() ).

I wonder if we can use  the existing LATERAL view https://github.com/sqlparser-rs/sqlparser-rs/blob/29dea5d017b03ea57bad0515356721b94d209341/src/ast/query.rs#L278-L286 structure


I would expect this 
```sql
LATERAL FLATTEN(INPUT => emp.project_names) AS proj_names
```

Would parse to a 
```rust
LateralView {
  lateral_view: ..., // <--- this field would have the FLATTEN function
  ...
}
```

I don't understand the need for a specific function in TableFactor 🤔

lustefaniak:
Thank you, @alamb, for the feedback. It looks reasonable, and I will try implementing it as you suggested.

Issue #800, opened 2023-01-24
Can't parse escaped double quotes


DeadNumbers:
```thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TokenizerError("Expected close delimiter '\"' before EOF. at Line: 12920, Column 429")', src/main.rs:8:54```

The line contains escaped double quotes:

`'S-130, 190      , 290, 390 in 2005, 491 in 2012, 490 in 2017, L-180, 280, 380, WIMS Course, Wildfire Analyst Training, Misc. Ops and Plans Related Courses,       Misc. \"I\" Courses.'`

**Comment by AugustoFKL** [comment]
@DeadNumbers can you please post the query you're trying to run, please?

**Comment by DeadNumbers** [comment]
@AugustoFKL have no query, i'm read .sql file and try to parse from string. my goal is to convert all tables data from sql dump to csv.

**Comment by Jefffrey** [comment]
Will need more info on this to reproduce, as `\"` seems to parse just fine within a literal, e.g.

```sql
SELECT '\"test\"'
```

AugustoFKL:
@DeadNumbers, can you give us the '.sql' you're trying to read? 

Is that the line you're trying to parse?
```sql
'S-130, 190      , 290, 390 in 2005, 491 in 2012, 490 in 2017, L-180, 280, 380, WIMS Course, Wildfire Analyst Training, Misc. Ops and Plans Related Courses,       Misc. \"I\" Courses.'
```

DeadNumbers:
@AugustoFKL yes, this line

appletreeisyellow:
Just curious, @cobyge added support for backslash escape for MySQL in https://github.com/sqlparser-rs/sqlparser-rs/pull/844. Why not for all other SQL?

cobyge:
Honestly I just had my specific use-case which was blocking my use, and I didn't have the time or setup to check and test every single database supported.  

Each database needs it's own escape implementation.

LucaCappelletti94:
The line from this report parses on current `main` with `MySqlDialect`, via the backslash escape support added in #844, and `GenericDialect` accepts it too. Closing, please reopen with a failing statement if you still hit this.

PR #844, opened 2023-04-06
Added support for Mysql Backslash escapes

cobyge:
Hi,
This fixes issue #800 

This adds proper support for MySQL backslash escaping (As described [here](https://dev.mysql.com/doc/refman/8.0/en/string-literals.html#character-escape-sequences))
Before now, the escaping support would only work for escaping single-quotes or backslashes. This adds full support for all Special Character Escape Sequences supported by MySQL at the current time (as defined in the table in the link above).

I also added some tests to make sure that the special characters are parsed as they should be.

Let me know if there is anything else I can do in order for this to get merged.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1094,16 +1094,11 @@ impl<'a> Tokenizer<'a> {
 
         chars.next(); // consume the opening quote
 
-        // slash escaping is specific to MySQL dialect
-        let mut is_escaped = false;
         while let Some(&ch) = chars.peek() {
             match ch {
                 char if char == quote_style => {
                     chars.next(); // consume
-                    if is_escaped {
-                        s.push(ch);
-                        is_escaped = false;
-                    } else if chars.peek().map(|c| *c == quote_style).unwrap_or(false) {
+                    if chars.peek().map(|c| *c == quote_style).unwrap_or(false) {
                         s.push(ch);
                         chars.next();
                     } else {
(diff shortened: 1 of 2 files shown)


cobyge:
Now that I think about it, I realize that serializing special types back into a string will not end well.
I'm debating if I should add the correct implementation to `Display` for `Token` (specifically, doing the reverse translation for `Token::SingleQuotedString` and `Token::DoublQuotedString` as those are the only relevant ones according to the spec.)
Alternatively, I could just not convert special characters, and leave them as "raw" escape codes. (I'd still have to take care of escaping backslashes and quotes)
I'd appreciate maintainer input on this (I might be missing something).
Either way, tomorrow I'll add tests to make sure the sql is Serialized back to a query correctly

cobyge:
As far as I can tell, in order to correctly convert characters back into escape characters, (for example - converting a null byte back to `\0`), the implementation of `fmt::Display for Value` (specifically only `Value::SingleQuotedString` and `Value::DoubleQuotedString` are relevant here) would need to be dialect-aware (Which as far as I can tell, it is not).

This presents a few problems:
* When deserializing a value to a string, what is the correct value of a null byte (or \Z, or other)? Is it always `\0`? I would assume not every SQL dialect supports that.
    * This leads me to prefer NOT converting the special characters, so upon parsing the following quote - `SELECT '1 \b'`, the value of `Value::SingleQuotedString` would be `'1 \b'` and not `'1 \u{8}'`. This leads to sorta-incorrect and possibly unexpected behavior.
* When deserializing `\%` or `\_`, the 'correct' implementation would be to parse it into `%` or `_` respectively, but as stated in the docs - `The \% and \_ sequences are used to search for literal instances of % and _ in pattern-matching contexts where they would otherwise be interpreted as wildcard characters.  If you use \% or \_ outside of pattern-matching contexts, they evaluate to the strings \% and \_, not to % and _.`. So in order to preserve query-semantics, the correct thing to do would be to keep the backslash when parsing those two characters. Like the previous point, this is not obvious behavior.
* Some cases are not escaped properly, I will attempt to fix that in the following commit.

cobyge:
Another issue - 
Assuming that we keep `\b` (or others) as it is, without converting it to `\u{8}`. How can we differentiate between the following two queries:
* `SELECT '\b'`
* `SELECT '\\b'`

Both of them would be serialized to `SELECT '\b'`, and when deserializing them, there is no way to know what the original behavior is.
The 'easy' fix is to store a raw '\' as '\\', but then we are losing most of the semantics of parsing the value.

alamb (approved):
Looks good to me -- thank you @cobyge

alamb:
> I'd love to hear any thoughts about the correct way these things should work, and I'll be happy to implement them

In general I would prefer we be usecase driven. The issues you highlight seem like issues, but I am not sure how important it is to ensure full round trip display compatibility at this time, especially when there are multiple potential ways to escape/encode a particular character.

Perhaps you can write some tests cases that show the behavior (or lack thereof) that you are worried about and we can use those to drive the discussion of what the expected behavior is. 

> Both of them would be serialized to SELECT '\b', and when deserializing them, there is no way to know what the original behavior is.

I am not sure it is critical to recover the original query byte for byte. Thing such as whitespace / newlines are not preserved either

PR #845, opened 2023-04-09
Move tests from parser.rs to appropriate parse_XX tests

alamb:
Follow on to @ankrgyl 's suggestion in https://github.com/sqlparser-rs/sqlparser-rs/pull/842/files#r1149575835

I tried to clean up the tests a little both to make the current code a little nicer but also to help future contributors find the right place (and the right pattern)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -7406,11 +7406,4 @@ mod tests {
             ))
         );
     }
-
-    #[test]
-    fn test_update_in_with_subquery() {
-        let sql = r#"WITH "result" AS (UPDATE "Hero" SET "name" = 'Captain America', "number_of_movies" = "number_of_movies" + 1 WHERE "secret_identity" = 'Sam Wilson' RETURNING "id", "name", "secret_identity", "number_of_movies") SELECT * FROM "result""#;
-        let ast = Parser::parse_sql(&GenericDialect, sql).unwrap();
-        assert_eq!(ast[0].to_string(), sql);
-    }
 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -6912,40 +6912,6 @@ mod tests {
         });
     }
 
-    #[test]
-    fn test_parse_limit() {
-        let sql = "SELECT * FROM user LIMIT 1";
-        all_dialects().run_parser_method(sql, |parser| {
-            let ast = parser.parse_query().unwrap();
-            assert_eq!(ast.to_string(), sql.to_string());
-        });
-
-        let sql = "SELECT * FROM user LIMIT $1 OFFSET $2";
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -6912,40 +6912,6 @@ mod tests {
         });
     }
 
-    #[test]
-    fn test_parse_limit() {
```
alamb, line 6916: I moved this code to sqlparser_common and sqlparser_mysql as appropriate, and used more concise functions to test these dialects

src/parser.rs
```diff
@@ -7402,24 +7368,6 @@ mod tests {
         );
     }
 
-    #[test]
```
alamb, line 7405: this is postgres specific so moved to sqlparser_postgres

src/parser.rs
```diff
@@ -7458,11 +7406,4 @@ mod tests {
             ))
         );
     }
-
-    #[test]
-    fn test_update_in_with_subquery() {
```
alamb, line 7463: moved to sqlparser_postgres

tests/sqlparser_common.rs
```diff
@@ -1559,65 +1564,6 @@ fn parse_select_group_by() {
     );
 }
 
-#[test]
-fn parse_select_group_by_grouping_sets() {
```
alamb, line 1563: these tests are postgres specific (test with generic dialect and postgres) so moved to sqlparser_postgres

src/parser.rs
```diff
@@ -6912,40 +6912,6 @@ mod tests {
...
 
-    #[test]
-    fn test_parse_limit() {
-        let sql = "SELECT * FROM user LIMIT 1";
-        all_dialects().run_parser_method(sql, |parser| {
-            let ast = parser.parse_query().unwrap();
-            assert_eq!(ast.to_string(), sql.to_string());
-        });
-
-        let sql = "SELECT * FROM user LIMIT $1 OFFSET $2";
```
ankrgyl, line 6923: Did we lose these tests? I see a mysql dialect test for offset but not one that touches the other dialects. Apologies if I missed it!

ankrgyl (commented):
lgtm pending one question!

src/parser.rs
alamb: Good spot. I should have called this out explicitly -- this test was replicated and therefore redundant. 

The pre-existing test in sqlparser_common.rs is almost identical: 

https://github.com/sqlparser-rs/sqlparser-rs/blob/784a19138f269dab4ea3e1ad3c9da565d20c9155/tests/sqlparser_common.rs#L6092-L6104

alamb:
Thanks for the review @ankrgyl

PR #848, opened 2023-04-10
Add `dialect_from_str` and improve `Dialect` documentation

alamb:
In DataFusion we wanted to be able to dynamically pick the dialect. @yjshen added code to do in  https://github.com/apache/arrow-datafusion/pull/5868

This PR ports the code from DataFusion upstream to sqlparser-rs as well as improves the documentation of `Dialect` in general

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -42,14 +42,44 @@ pub use self::sqlite::SQLiteDialect;
 pub use crate::keywords;
 use crate::parser::{Parser, ParserError};
 
-/// `dialect_of!(parser is SQLiteDialect |  GenericDialect)` evaluates
-/// to `true` if `parser.dialect` is one of the `Dialect`s specified.
+/// `dialect_of!(parser Is SQLiteDialect |  GenericDialect)` evaluates
+/// to `true` if `parser.dialect` is one of the [`Dialect`]s specified.
+///
+///
 macro_rules! dialect_of {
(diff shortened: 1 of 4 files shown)


src/dialect/mod.rs
```diff
@@ -141,4 +192,32 @@ mod tests {
...
+        assert!(parse_dialect("sqlite").is::<SQLiteDialect>());
+        assert!(parse_dialect("snowflake").is::<SnowflakeDialect>());
+        assert!(parse_dialect("SnowFlake").is::<SnowflakeDialect>());
+        assert!(parse_dialect("MsSql").is::<MsSqlDialect>());
+        assert!(parse_dialect("clickhouse").is::<ClickHouseDialect>());
+        assert!(parse_dialect("ClickHouse").is::<ClickHouseDialect>());
+        assert!(parse_dialect("bigquery").is::<BigQueryDialect>());
+        assert!(parse_dialect("BigQuery").is::<BigQueryDialect>());
+        assert!(parse_dialect("ansi").is::<AnsiDialect>());
+        assert!(parse_dialect("ANSI").is::<AnsiDialect>());
```
ankrgyl, line 213: I wish there were a way for us to throw a compile error if one of the dialects isn't tested here (or alternatively, loop through a list of dialects and roundtrip them or something to make sure each one is represented).

alamb: I agree that would be great. 

The only thing I can think of is to make an enum or something with all the dialects 🤔

PR #850, opened 2023-04-11
Baseline dbt support

callum-mcdata:
Adding support for refs and sources


alamb:
I wonder if you can offer any hints about why this PR got closed :)

callum-mcdata:
Oh I was mainly messing around with a fork of sqlparser-rs for a side project I'm working on (dbt-sql parser that can use the Dialect construct). I had meant to try and open the PR against my fork 🤦 . If there's interest in getting something like this merged into the project I'm happy to chat about it!

I had thought about implementing a DbtDialect but realized that wouldn't then take advantage of the Dialects which can coexist with dbt. So I decided to implement support for refs and sources into the base functionality. That way (in this ideal future where this thing in my head is built) the user can take advantage of the dialects while also parsing their dbt-sql

PR #852, opened 2023-04-13
Support `DISTINCT ON (...)`

### LINKED ISSUE #806: In correct serialization of `DISTINCT ON()` syntax with postgres
Here is the failing sql: 

```sql
SELECT
   DISTINCT ON (column1) column_alias,
   column2
FROM
   table_name
ORDER BY
   column1,
   column2;
```

Note that this parses without errors because it parses the `ON (column1)` as a function.

When serializing back to string from ast, it produces: 

```sql
SELECT DISTINCT ON(column1) AS column_alias, column2 FROM table_name ORDER BY column1, column2
```

The `AS` should not be there (it is an error in Postgres)

aljazerzen:
Closes #806


src/parser.rs
```diff
@@ -2300,16 +2300,31 @@ impl<'a> Parser<'a> {
...
+        if all {
+            return parser_err!("Cannot specify both ALL and DISTINCT".to_string());
+        }
+        let on = self.parse_keyword(Keyword::ON);
+        if !on {
+            return Ok(Some(Distinct::Distinct));
+        }
+
+        self.expect_token(&Token::LParen)?;
+        let col_names = if self.consume_token(&Token::RParen) {
```
alamb, line 2320: This looks like special case code to handle `SELECT DISTINCT ON() ` - can you please add a test for this case?

alamb (commented):
Thank you @aljazerzen  -- this looks very nice. I think it just needs one more test and we could merge it in

src/parser.rs
aljazerzen: :+1:

alamb (approved):
Looks great -- thank you again @aljazerzen

PR #856, opened 2023-04-24
Support identifiers beginning with digits in MySQL

### LINKED ISSUE #804: Incorrect parsing literals that consisting of digits and letters, beginning with digits
It seems when parsing a literal beginning with digits, the sql parser simply ends this token when meets a non-numeric character

see https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/tokenizer.rs#L547

For example, `123foobar` would be parsed into a number token `123` and a word token `foobar` instead of one word token `123foobar`.

And `123foobar456` would be parsed into `123` and `foobar456`

AviRaboah:
Closes #804.

### DIFF
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -18,8 +18,8 @@ pub struct MySqlDialect {}
 impl Dialect for MySqlDialect {
     fn is_identifier_start(&self, ch: char) -> bool {
         // See https://dev.mysql.com/doc/refman/8.0/en/identifiers.html.
-        // We don't yet support identifiers beginning with numbers, as that
-        // makes it hard to distinguish numeric literals.
+        // Identifiers which begin with a digit are recognized while tokenizing numbers,
+        // so they can be distinguished from exponent numeric literals.
         ch.is_alphabetic()
             || ch == '_'
             || ch == '$'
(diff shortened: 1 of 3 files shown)


alamb (commented):
Thank you @AviRaboah  and @ankrgyl  -- the only other thing I suggest is a negative case for the exponent case

For example, can we add a test case to make sure `123e4`  can not be used as an identifier? (at least that is my understanding of the limitiation)

alamb (approved):

AviRaboah:
> Thank you @AviRaboah and @ankrgyl -- the only other thing I suggest is a negative case for the exponent case
> 
> For example, can we add a test case to make sure `123e4` can not be used as an identifier? (at least that is my understanding of the limitation)

It can't be used  as an identifier, because I added a condition which check that the exponent part is not empty in order to check if it's an identifier. in your test case the exponent part will be > 0 and because of that it will end as an exponent number. but i'll add a test case to cover this.

AviRaboah:
Hi @alamb, 
I added the test :)

alamb (commented):
Thank you @AviRaboah and @ankrgyl

alamb (approved):

PR #858, opened 2023-04-27
Add support for query source in COPY .. TO statement

### LINKED ISSUE #857: Add support for query source in COPY .. TO statement
The current implementation of COPY .. TO statement only supports table (with optional column names) but doesn't support arbitrary query according to PostgreSQL specification. It needs to support query source.

Reference:
https://www.postgresql.org/docs/current/sql-copy.html

aprimadi:
Closes #857

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4000,13 +4000,34 @@ impl<'a> Parser<'a> {
 
     /// Parse a copy statement
     pub fn parse_copy(&mut self) -> Result<Statement, ParserError> {
-        let table_name = self.parse_object_name()?;
-        let columns = self.parse_parenthesized_column_list(Optional, false)?;
+        let source;
+        if self.consume_token(&Token::LParen) {
+            source = CopySource::Query(Box::new(self.parse_query()?));
+            self.expect_token(&Token::RParen)?;
+        } else {
+            let table_name = self.parse_object_name()?;
+            let columns = self.parse_parenthesized_column_list(Optional, false)?;
+            source = CopySource::Table {
+                table_name,
+                columns,
+            };
+        }
         let to = match self.parse_one_of_keywords(&[Keyword::FROM, Keyword::TO]) {
             Some(Keyword::FROM) => false,
             Some(Keyword::TO) => true,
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -4000,13 +4000,34 @@ impl<'a> Parser<'a> {
...
+                table_name,
+                columns,
+            };
+        }
         let to = match self.parse_one_of_keywords(&[Keyword::FROM, Keyword::TO]) {
             Some(Keyword::FROM) => false,
             Some(Keyword::TO) => true,
             _ => self.expected("FROM or TO", self.peek_token())?,
         };
+        if !to {
```
alamb, line 4020: I think we should  remove this error in the parser per https://github.com/sqlparser-rs/sqlparser-rs#syntax-vs-semantics (if downstream crates want to error that is fine). 

I don't feel super strongly on this point

what do you think @ankrgyl ?

alamb (approved):
Looks great to me -- thank you @aprimadi 

I had just one question related to the error case

aprimadi:
Ah yes working on fixing it

src/parser.rs
aprimadi: I argue that this is syntactic as the syntax is described here and is different for `COPY .. FROM` vs `COPY .. TO`. https://www.postgresql.org/docs/current/sql-copy.html.

Looking at DuckDB syntax graph also there seems to be difference syntactically between `COPY .. FROM` vs `COPY .. TO`. https://duckdb.org/docs/sql/statements/copy

But yeah let's wait on what @ankrgyl think about it. I'll go with whatever the consensus decide since I'm not an expert in various SQL dialects.

alamb: As I said above, I think this way is fine (and kudos to you for covering it with tests). I agree let's see if @ankrgyl  has any thoughts

aprimadi: Thank you for the really fast review @alamb -- must be tough reviewing so many repos ❤️

alamb: 🤯  most days

alamb:
Thanks again @aprimadi

PR #860, opened 2023-04-29
Add parse_multipart_identifier function to parser

### LINKED ISSUE #805: Parse identifiers function
Extract function from datafusion to here

https://github.com/apache/arrow-datafusion/blob/d5077b55ef39e5376ace93d4e9068b94b678e7f1/datafusion/common/src/table_reference.rs#L228-L279

Allowing use of this function to parse a string containing identifiers (not a SQL string)

Jefffrey:
Closes #805

To be used for parsing identifiers which when could be quoted, can get quite complex (rather than relying on simple split on dot)

Pulled from DataFusion with one change: allow whitespaces between parts in the identifier, as seems this is supported by Spark SQL and Postgres

So following idents would be parsed with the same result:

```sql
catalog.database.table

catalog . database . table
```

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4687,6 +4687,54 @@ impl<'a> Parser<'a> {
         Ok(idents)
     }
 
+    /// Parse identifiers of form ident1[.identN]*
+    pub fn parse_multipart_identifier(&mut self) -> Result<Vec<Ident>, ParserError> {
+        let mut idents = vec![];
+
+        // expecting at least one word for identifier
+        match self.next_token().token {
+            Token::Word(w) => idents.push(w.to_ident()),
+            Token::EOF => {
+                return Err(ParserError::ParserError(
+                    "Empty input when parsing identifier".to_string(),
+                ))?
+            }
+            token => {
+                return Err(ParserError::ParserError(format!(
+                    "Unexpected token in identifier: {token}"
+                )))?
+            }
+        };
+
+        // parse optional next parts if exist
+        loop {
+            match self.next_token().token {
+                // ensure that optional period is succeeded by another identifier
+                Token::Period => match self.next_token().token {
+                    Token::Word(w) => idents.push(w.to_ident()),
+                    Token::EOF => {
+                        return Err(ParserError::ParserError(
+                            "Trailing period in identifier".to_string(),
+                        ))?
+                    }
+                    token => {
+                        return Err(ParserError::ParserError(format!(
+                            "Unexpected token following period in identifier: {token}"
+                        )))?
+                    }
+                },
+                Token::EOF => break,
+                token => {
+                    return Err(ParserError::ParserError(format!(
+                        "Unexpected token in identifier: {token}"
+                    )))?
+                }
+            }
+        }
+
+        Ok(idents)
+    }
+
     /// Parse a simple one-word identifier (possibly quoted, possibly a keyword)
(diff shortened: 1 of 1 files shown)


ankrgyl (commented):
From what I can tell this can only be invoked directly (by calling `parse_multipart_identifier()`) and not automatically while parsing a query. Is that correct/intended?

Jefffrey:
Yes that's correct @ankrgyl 

It was partly inspired by how Spark has a similar method: https://github.com/apache/spark/blob/f8604ad14b24e8c657a0305b4fb8ad7efcb84060/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/ParseDriver.scala#L65-L70

To allow downstream to parse identifiers by themselves (not necessarily as part of a query)

ankrgyl:
Ah got it. Personally I'd prefer to integrate it into the main parser so that we increase our overall support for Spark SQL and Postgres. @alamb wdyt?

alamb:
> Ah got it. Personally I'd prefer to integrate it into the main parser so that we increase our overall support for Spark SQL and Postgres. @alamb wdyt?

Having it integrated makes sense to me in theory. Do you have any suggestion about how we might do this specifically @ankrgyl  ?

src/parser.rs
```diff
@@ -4687,6 +4687,54 @@ impl<'a> Parser<'a> {
         Ok(idents)
     }
 
+    /// Parse identifiers of form ident1[.identN]*
+    pub fn parse_multipart_identifier(&mut self) -> Result<Vec<Ident>, ParserError> {
```
alamb, line 4691: Looking at this function more closely, I wonder how if we can't reuse `parse_identifiers` from above?

Or maybe we could add some more documentation (like docstrings / tests) for this function to make it clearer it is designed to parse strings into identifiers 🤔

Jefffrey: I think `parse_identifiers` above is the odd one out, as it accepts any delimiter between words/idents (except equal sign) which means it would parse `test + one + two` as a valid sequence of identifiers. This is why I introduced the new `parse_multipart_identifier` to be more strict about what constitutes a compound/multipart identifier.

Though I did base `parse_multipart_identifier` primarily on Spark SQL/Postgres syntax, so maybe `parse_identifiers` is generic to accommodate some other type of syntax

alamb: Upon some more reflection, I think we should accept this code and update the documentation strings on `parse_multipart_identifier` and  `parse_identifiers` to explain in more detail what they do and how they are different.

I can try and find time to update the documentation maybe next week -- or @Jefffrey  do you have time to do so?

alamb (approved):

src/parser.rs
Jefffrey: I've update the doc a bit, let me know any further suggestions

src/parser.rs
```diff
@@ -4705,6 +4705,92 @@ impl<'a> Parser<'a> {
         Ok(idents)
     }
 
+    /// Parse identifiers of form ident1[.identN]*
+    ///
+    /// Similar in functionality to [parse_identifiers], with difference
```
alamb, line 4710: This is great. Thank you

alamb (approved):
I think this is a great addition -- thanks again @Jefffrey 

I'll plan to merge (and release) this next week unless there are any more comments. cc @ankrgyl

Issue #861, opened 2023-04-29
feat: no-escape mode for MySQL


canalun:
Hi, thank you for developing this amazing parser...!
I use it for building a MySQL formatter and it help me so much😊

I'm not 100% sure if the idea I write below is good one, but I would like to share it just in case.

### idea
The implementation of no-escape mode.
It means that the info of escape (=backslashes) is kept through parsing queries.

### current situation and details of idea
Currently, this library works as below. In short, backslashes disappear.

```
-- before parse
INSERT INTO `table` ('id', 'text') VALUES (1, 'they say \"Hello\".');
INSERT INTO `table` ('id', 'text') VALUES (1, 'new\nline');

-- after parse and conversion into query
INSERT INTO `table` ('id', 'text') VALUES (1, 'they say Hello".');
INSERT INTO `table` ('id', 'text') VALUES (1, 'new
line');
```

That can be a problem when using this library for developing tools,
because some tools (e.g. formatter) are expected not to add any change to the queries.

So what about the idea that users can choose if they keep backslashes or not?

### blueprint
Actually, I created a small patch to realize the above feature in my forked repo.
And, just for clarification, I opened a draft of PR.

**NOTE**: I'm pretty sure the patch is not for implementation for this library. It is just an emergency patch for my formatter :)

Thank you for reading!😊

canalun:
if it's worth implementing, i'd like to work on it😊

alamb:
I can see the rationale for trying to keep the original escaping (as opposed to processing the escapes). 

What do you think about maybe trying to add this feature as some sort of Dialect (which is the current way we have to control parsing changed)?

canalun:
thanks, i think that's a really good way!
i'll work on it🚀

canalun:
I'm working on #870 , and let me notice when the pr get ready.

canalun:
#870 is ready for review. Thank you for waiting!

alamb:
I believe we released this in 0.36.0 - so closing. Please reopen if I got that wrong

PR #862, opened 2023-04-29
feat: non-escape mode (draft for discussion)

canalun:
this is for explanation of the discussion #861


canalun:
moved to #870

alamb:
Since the branch is on your fork, there is no need to remove it nor any reason for the sqlparser-rs maintainers to do so either

canalun:
got it!! thank you for letting me know that!

Issue #863, opened 2023-05-02
Representing infix `REGEXP`


max-sixty:
[MySQL](https://dev.mysql.com/doc/refman/8.0/en/regexp.html) & Sqlite have unusual regex operators:

```sql
 SELECT 'a' REGEXP '^[a-d]';
```

i.e. it's an infix function.

We are big users (and big fans!) of `sqlparser-rs` in PRQL We have a PR https://github.com/PRQL/prql/pull/2458 for implementing regex in the other dialects.

Could I ask — is there some way of representing an infix function in `sqlparser-rs` AST?

max-sixty:
Thanks for the swift response @alamb 

> 👋 How about https://docs.rs/sqlparser/0.33.0/sqlparser/ast/enum.Expr.html#variant.BinaryOp ?

The only one there which is custom is [`PGCustomBinaryOperator`](https://docs.rs/sqlparser/0.33.0/sqlparser/ast/enum.BinaryOperator.html#variant.PGCustomBinaryOperator). The official [docs](https://www.postgresql.org/docs/current/sql-createoperator.html) of that don't allow for words (i.e. `REGEXP`):

> The operator name is a sequence of up to NAMEDATALEN-1 (63 by default) characters from the following list:
> + - * / < > = ~ ! @ # % ^ & | ` ?

...though maybe `sqlparser-rs` don't enforce that and we can squeeze it in?

alamb:
> ...though maybe sqlparser-rs don't enforce that and we can squeeze it in?

Since sqlparser-rs parses many different dialects I don't think there is any reason to limit sqlparser to posgres's limitations

I think https://github.com/sqlparser-rs/sqlparser-rs#syntax-vs-semantics may be relevant here

max-sixty:
> I think https://github.com/sqlparser-rs/sqlparser-rs#syntax-vs-semantics may be relevant here

OK — I had thought that allowed characters might be syntax rather than semantics, but very happy to take it!

Thanks @alamb , I'll close

max-sixty:
Sorry, actually I spoke too soon — that seems to [write `OPERATOR` when writing it out](https://github.com/sqlparser-rs/sqlparser-rs/blob/1cf913e7173a27897c0d198782e385de806dbcbd/src/ast/operator.rs#L136).

This would be the diff I need — happy to put a PR in — but:
[...]

PR #868, opened 2023-05-07
feat: Add custom operator

max-sixty:
From #863

- It doesn't parse anything — I'm not sure how to parse ` SELECT 'a' REGEXP '^[a-d]';` with `REGEXP` as the operator... (but fine for my narrow purpose)
- If we need tests, where would I add them?


ankrgyl:
Apologies if I'm missing some context, but what is the purpose of this new operator? Is it so you can build up a tree manually with a custom string at a certain position?

max-sixty:
Yes, we'd like to write:

```
 SELECT 'a' REGEXP '^[a-d]';
```

...which is how MySQL & Sqllite do regex...

ankrgyl:
I'll add some comments on the task!

alamb (approved):
I think this looks good to me -- thank you @max-sixty 
I took the liberty of adding a docstring to try and make the use case clearer (and avoid it being deleted by accident in the future

max-sixty:
Thanks @alamb !

alamb:
BTW I expect to release the next version of sqlparser to crates.io today, which will include this fix

PR #870, opened 2023-05-08
feat: mysql no-escape mode

canalun:
for #861


src/test_utils.rs
```diff
@@ -166,13 +166,25 @@ pub fn all_dialects() -> TestedDialects {
...
             Box::new(MySqlDialect {}),
+            Box::new(MySqlNoEscapeDialect {}),
             Box::new(BigQueryDialect {}),
             Box::new(SQLiteDialect {}),
         ],
         options: None,
     }
 }
 
+pub fn all_dialects_other_than_mysqlnoescape() -> TestedDialects {
```
canalun, line 177: the functions, "XX_other_than_mysqlnoescape", including this are supposed to be used for tests regarding escape.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 # https://github.com/rust-lang/cargo/issues/1596
 serde_json = { version = "1.0", optional = true }
 sqlparser_derive = { version = "0.1.1", path = "derive", optional = true }
+duplicate = "1.0.0"
```
alamb, line 37: in general we try and keep the dependencies of this crate as minimal as possible -- can you please remove this dependency?

alamb (commented):
Thanks @canalun  -- I'll try and find time to review this PR over the next few days

Cargo.toml
canalun: got it!
please refer to it: [598d9e9](https://github.com/sqlparser-rs/sqlparser-rs/pull/870/commits/598d9e97ada700d1cf57b2ea17b383365dd2b542)

alamb (commented):
Thank you @canalun for the contribution and the well worded description on https://github.com/sqlparser-rs/sqlparser-rs/issues/861

After reviewing this PR, I feel like "dialect" isn't quite the right abstraction for this behavior as it doesn't change the SQL that is supported, but rather it changes the internal representation of how the data is stored after parsing

Thus, perhaps it could be added as a flag on `ParserOptions` might be better: https://docs.rs/sqlparser/0.33.0/sqlparser/parser/struct.ParserOptions.html


In terms of encoding the value, I wonder if we could add another variant to `Value` that could store the original, unescaped, characters. Perhaps something like the following



```Rust
/// Primitive SQL values such as number and string
pub enum Value {
// ...
    /// 'string value'
    SingleQuotedString(String),
    /// Value and its original (escaped) form
    WithOriginal {
      /// The parsed / unescaped value (so "foo\nbar" would have a newline in it)
       value: Box<Value>, 
       /// characters in the original string (so "foo\nbar" would still have the slash)
       original: String
    }
....
```

This would then let you isolate the parser changes to the parsing of values, as well as make it clear how to format each variant. This might make the implementation and testing more straightforward 

What do you think ?

canalun:
@alamb Thank you for reviewing it!
I agree with your idea👶 handling this situation by using dialects is a bit tricky, but the alternative is straight forward.
I will work on it, but it may take some time...! (thank you for your patience!)

alamb:
No worries ! Thank you @canalun 

FYI I don't plan the next sqlparser release until June 2023 so there is no time pressure from my end https://github.com/sqlparser-rs/sqlparser-rs/issues/887

canalun:
I'm working on it...! Sorry for late response.
I'll update the situation tomorrow!

canalun:
Hi @alamb, sorry for late update.

I've started to work on it and found few things,
and then I'd like to let you know these points beforehand implementing it fully, **just in case**...!
(btw, **It would be useful for following the points to refer to the PR #894**, that is what I prototyped these days👶)

Firstly, I wanna share what I discovered about the details of implementation of [the current strategy](https://github.com/sqlparser-rs/sqlparser-rs/pull/870/#pullrequestreview-1431255108).
- First of all, when parsing sql queries, the point where the information of escape (e.g. backslashes) vanishes is inside `next_tokens()`. Specifically, some functions such as `tokenize_quoted_string()` omit the info such as backslashes.
- So, we have to define and use new functions which keep the info of escape, instead of those escape-vanishing functions.
- Adding to that, tokenizer will start to receive options, a new enum of Value and Token will be defined (as you said😊).

Considering this consequence, I think the current strategy also has some costs.
  1. When using the option for no-escape, the AST will not have the information of type of each token. This is because, for example, both `Token::SingleQuotedString` and `Token::DoubleQuotedString` will become the same `Token::OriginalString` under the no-escape mode.
  Of course, in the reality, `Token::OriginalString` will be like `Token::OriginalString(""example"")` or `Token::OriginalString("'example'")`. So, it's sure that you'll never lost the info of type of each token, but i think it will get a bit difficult to analyze the AST by using programs.
  2. The knowledge about each token's prefix will spillover to `next_tokens()`.
  For example, please see [this line](https://github.com/sqlparser-rs/sqlparser-rs/pull/894/files#diff-1ae3fbde6da7fe7f3b7aec1740bedbd7b52b471ab36c62ea8a37a53f22b6f24bR527) of the prototype.
  3. As you can see from the point 2, the calc of `next_token()`will get messy...! I don't have any good idea, here.TT

If I may, I still think the current strategy is better than the previous one, because the previous one was a bit tricky.
The fact that the same type of Value can have the different escape-mode string would be really confusing,
especially when it comes to the implementation of `fmt::Display()` for each enum of Value. (https://github.com/sqlparser-rs/sqlparser-rs/pull/870/files#diff-9bcd0911924444fc7dd51971f9114291f2e6b619da2ce64ad48e91d6dd1abf8eR187)
This difficulty would affect the development almost permanently.

However, I think it's a bit difficult problem. Both have some pricy points.
I wanna hear your opinion🐤

Thank you for reading the long message...!

alamb:
> Thank you for reading the long message...!

Thanks @canalun  -- I agree it is a tricky problem. Maybe @ankrgyl  has some other thoughts. 

Basically my thinking is that if we are going to update the tokenizer in some backwards compatible way, we should also consider adding original line/offset information as well. 

> both Token::SingleQuotedString and Token::DoubleQuotedString will become the same Token::OriginalString under the no-escape mode.

What if we kept Token::SingleQuotedString and Token::DoubleQuotedString but made them something like

```
SingleQuotedString(Token)
```

Or something ?

I realize that would be a (very) disruptive API change 🤔

ankrgyl:
👋 This may be a very dumb idea — but if no-escape is a "mode" that we pass to the tokenizer, do we need to represent the "escaped" version too? If not, why can't we use the existing structs/enums and just change how the tokenizer populates the string representations inside of them? Especially if it doesn't impact (downstream) how the parser works?

In terms of making backwards incompatible changes, I do have a specific idea on how to do this that I just haven't had time to implement, which is to use features (so users can opt into backwards incompatibility) and then eventually deprecate the old modes.

ankrgyl:
Yes that's exactly right. I think all that matters is that you can get back strings with the same semantics as mysql's no-escape mode. It's basically _like_ a dialect for the tokenizer.

canalun:
@alamb @ankrgyl 
Thank you all for giving the idea! I agree with it :)

I understood queries are parsed and serialized like this.

| phrase of original query  | mode | node of AST | serialized |
| ------------- | ------------- |------------- |
| "A""B""A"  | no-escape | DoubleQuotedString(String::from("A\"\"B\"\"A"))  | "A""B""A"  |
| "A""B""A"  | default | DoubleQuotedString(String::from("A\"B\"A"))  | "A""B""A"  |
| "A\"B\"A"  | no-escape | DoubleQuotedString(String::from("A\\\"B\\\"A"))  | "A\"B\"A"  |
| "A\"B\"A"  | default | DoubleQuotedString(String::from("A\"B\"A"))  | "A""B""A"  |

assumptions
1. in no-escape mode, each node of AST has the exact original string.
2. in default mode, we don't have to care about whether the "\"" comes from \" or "" in the original query (this is the current spec of this library, i think.)

canalun:
I will work on it!
fortunately, I have to impl few modifications on the present PR :)

canalun:
I've added all the needed change and could you please see it? :)

alamb:
Thanks @canalun  --  I am a behind on reviews. I plan to review this tomorrow. Sorry for the delay

canalun:
> Thanks @canalun -- I am a behind on reviews. I plan to review this tomorrow. Sorry for the delay

absolutely no problem :)
please take your time and tell me if there was any misunderstanding between us😮

src/tokenizer.rs
```diff
@@ -447,12 +447,21 @@ impl<'a> State<'a> {
 pub struct Tokenizer<'a> {
     dialect: &'a dyn Dialect,
     query: &'a str,
+    options: &'a TokenizerOptions,
+}
+
+pub struct TokenizerOptions {
+    pub no_escape: bool,
```
alamb, line 454: perhaps we can add some comments here explaining what this is doing

src/tokenizer.rs
```diff
@@ -447,12 +447,21 @@ impl<'a> State<'a> {
...
 }
 
 impl<'a> Tokenizer<'a> {
     /// Create a new SQL tokenizer for the specified SQL statement
-    pub fn new(dialect: &'a dyn Dialect, query: &'a str) -> Self {
-        Self { dialect, query }
+    pub fn new(dialect: &'a dyn Dialect, query: &'a str, options: &'a TokenizerOptions) -> Self {
```
alamb, line 459: Since this is one of the main interfaces to SqlParser, I think changing the signature will be a fairly large change. If we are going to do so I would prefer to combine the dialect and options into a single structure

```rust
pub struct TokenizerOptions<'a> {
    pub no_escape: bool,
    dialect: &'a dyn Dialect,
}
...

impl<'a> Tokenizer<'a> {
    pub fn new(options: TokenizerOptions<'a>, query: &'a str,) -> Self {
...
}
```

That way the API gets more general and we won't have to change it again if we add new parameters

src/tokenizer.rs
```diff
@@ -1259,7 +1283,7 @@ mod tests {
     fn tokenize_select_1() {
         let sql = String::from("SELECT 1");
         let dialect = GenericDialect {};
-        let mut tokenizer = Tokenizer::new(&dialect, &sql);
+        let mut tokenizer = Tokenizer::new(&dialect, &sql, &TokenizerOptions { no_escape: false });
```
alamb, line 1286: Could we make this look more like:

```suggestion
        let mut tokenizer = Tokenizer::new(TokenizerOptions::new(&dialect), &sql);
```

src/tokenizer.rs
```diff
@@ -1876,11 +1900,29 @@ mod tests {
         compare(expected, tokens);
     }
 
+    #[test]
+    fn tokenize_quoted_identifier_with_no_escape() {
+        let sql = r#" "a "" b" "a """ "c """"" "#;
+        let dialect = GenericDialect {};
+        let mut tokenizer = Tokenizer::new(&dialect, sql, &TokenizerOptions { no_escape: true });
```
alamb, line 1907:
```suggestion
        let opts = TokenizerOptions::new(&dialect).with_no_escape(true);
        let mut tokenizer = Tokenizer::new(opts, sql, &TokenizerOptions { no_escape: true });
```

src/ast/value.rs
```diff
@@ -184,12 +184,31 @@ pub struct EscapeQuotedString<'a> {
 
 impl<'a> fmt::Display for EscapeQuotedString<'a> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        for c in self.string.chars() {
-            if c == self.quote {
-                write!(f, "{q}{q}", q = self.quote)?;
-            } else {
-                write!(f, "{c}")?;
+        let quote = self.quote;
```
alamb, line 187: Can you explain what this change is for?

alamb (commented):
Thanks @canalun  -- I think this approach is looking good.

The CI seems to be failing

I had some suggestions related to API design, and I think the new options need some docstrings (which I can help with) but I think this PR is quite close

alamb:
> absolutely no problem, thank you for reviewing it :)
> please take your time and tell me if there was any misunderstanding between us😮

Not at all ! I just feel bad I spend so little time maintaining this crate

src/tokenizer.rs
canalun: @alamb
> Since this is one of the main interfaces to SqlParser, I think changing the signature will be a fairly large change.

While I strongly agree with this idea, I think `dialect` and other options such as `no_escape` would be better to be separated. :)
There are two reasons for this.
1. If we separate `dialect` and other options(`TokenizerOptions`), you won't have to change the interface of `Tokenizer` even when you want to add new parameters. This is because all you have to do will be just adding the new option to `TokenizerOptions`. So the interface of `Tokenizer::new()` itself will not change, i think.
2. `Parser` has the below interface now, and `dialect` and `options` are separated here.
https://github.com/canalun/sqlparser-rs/blob/2e6502e40b9b777e0ee590e5aa172a3c1bc2a42d/src/parser.rs#L204
And this separation is assumed in the logic of `dialect_of`, which is called against both `Parser` and `Tokenizer`.
https://github.com/canalun/sqlparser-rs/blob/2e6502e40b9b777e0ee590e5aa172a3c1bc2a42d/src/dialect/mod.rs#L54
So we have to align the two interfaces of `Parser` and `Tokenizer`. And if point 1 is true, I think it would be better to set the same interface as `Parser` on `Tokenizer`.

How about it?? :)

src/tokenizer.rs
canalun: Thanks! I added the comment. 😊
29cb1e2

src/ast/value.rs
canalun: I should've explained this part, thank you!

In short, the change is necessary to correctly re-serialize value which can be either escaped or non-escaped.
Let me explain in order. :)

1. We are now using the same type of `Token` in both the `escaped` and the `no-escape` modes.
This means tokens such as `DoubleQuotedString` can have both escaped and non-escaped strings.

2. On the other hand, `Value` doesn't know which mode of escape is chosen by the user. So `fmt()` for `Value` has to be able to correctly display strings without knowing the strings are escaped or not.

3. The below table explains how `fmt()` should work.
According to it, we can say as follows:
"if the quote symbol in the string is repeated twice, OR, if the quote symbol is after backslash, `fmt()` should display all the chars without any escape. And if the quote symbol is used just between usual chars, `fmt()` should display it twice."

| phrase of original query  | mode | node of AST | serialized |
| ------------- | ------------- | ------------- | ------------- |
| `"A""B""A"`  | no-escape | `DoubleQuotedString(String::from("A\"\"B\"\"A"))`  | `"A""B""A"`  |
| `"A""B""A"`  | default | `DoubleQuotedString(String::from("A\"B\"A"))`  | `"A""B""A"`  |
| `"A\"B\"A"`  | no-escape | `DoubleQuotedString(String::from("A\\\"B\\\"A"))`  | `"A\"B\"A"`  |
| `"A\"B\"A"`  | default | `DoubleQuotedString(String::from("A\"B\"A"))` | `"A""B""A"`  |

This is a bit tricky, but i thought it was the best way for realizing what we discussed...!
would like to hear your frank opinion. :)

src/tokenizer.rs
canalun: Thanks!!!
I implemented and used `TokenizerOptions::default()` and `TokenizerOptions::with_no_escape(bool)`. :)

[29cb1e2](https://github.com/sqlparser-rs/sqlparser-rs/pull/870/commits/29cb1e20f618b45ea9e9d9418723f8800bcaef8d)

src/tokenizer.rs
canalun: > [29cb1e2](https://github.com/sqlparser-rs/sqlparser-rs/pull/870/commits/29cb1e20f618b45ea9e9d9418723f8800bcaef8d)

I think this commit mentioned above captures this point. ;)

canalun:
now, `cargo test`, `cargo fmt`, and `cargo clippy` passed on local😊

canalun:
Thank you for running CI, and found it failed!!
the command i ran was not correct, so I'm fixing it!

canalun:
fixed :)

alamb (commented):
@canalun  thank you for bearing with me

I think the core escaping / unescaping logic and tests in this PR are great. I do have concerns about the API, but rather than going back and forth with you again  and delaying this PR more, I made a PR with some proposed API changes that target this branch

https://github.com/canalun/sqlparser-rs/pull/1

Please let me know what you think of them

canalun:
@alamb 
Thank you so much...!!! I think your modification is really good idea.
I commented on your PR :)

canalun:
@alamb merged your PR, thanks!!

alamb (approved):
Thanks @canalun !

alamb:
😅  -- thanks again for bearing with this @canalun  -- I think we got to a good place in the end

canalun:
@alamb Thank you so much for continuous support...!!! Hope this feature helps someone in the future :)

PR #872, opened 2023-05-09
Make Expr::Interval its own struct

### LINKED ISSUE #869: Make Expr::Interval its own struct
Currently Expr::Interval is part of Expr enum. Instead, we should create its own struct because it has a lot of fields.

```rust
pub enum Expr {
    ...

    /// INTERVAL literals, roughly in the following format:
    /// `INTERVAL '<value>' [ <leading_field> [ (<leading_precision>) ] ]
    /// [ TO <last_field> [ (<fractional_seconds_precision>) ] ]`,
    /// e.g. `INTERVAL '123:45.67' MINUTE(3) TO SECOND(2)`.
    ///
    /// The parser does not validate the `<value>`, nor does it ensure
    /// that the `<leading_field>` units >= the units in `<last_field>`,
    /// so the user will have to reject intervals like `HOUR TO YEAR`.
    Interval {
        value: Box<Expr>,
        leading_field: Option<DateTimeField>,
        /// or as `__ TO SECOND(x)`.
    },

    ...
}
```

I think it is better this way:

```rust
pub enum Expr {
    ...

    /// INTERVAL expression
    Interval(Interval),

    ...
}
```

aprimadi:
Fix #869

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4321,13 +4321,18 @@ mod tests {
     #[test]
     fn test_interval_display() {
         let interval = Expr::Interval(Interval {
-            value: Box::new(Expr::Value(Value::SingleQuotedString(String::from("123:45.67")))),
+            value: Box::new(Expr::Value(Value::SingleQuotedString(String::from(
+                "123:45.67",
+            )))),
             leading_field: Some(DateTimeField::Minute),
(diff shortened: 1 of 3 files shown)


alamb (approved):
Looks good to me -- thanks @aprimadi

PR #874, opened 2023-05-09
Add REGEXP infix operator for MySQL

max-sixty:
Would replace #868 

But couple of issues:
- I'm getting a test failure. What's the best test for this?
- Should `REGEXP` be a keyword? If not, what would we parse?
- Anything else I'm getting wrong?

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -85,6 +85,8 @@ pub enum BinaryOperator {
     BitwiseOr,
     BitwiseAnd,
     BitwiseXor,
+    // MySQL & Sqlite use `REGEXP` as an infix operator
+    Regexp,
     PGBitwiseXor,
     PGBitwiseShiftLeft,
     PGBitwiseShiftRight,
@@ -122,6 +124,7 @@ impl fmt::Display for BinaryOperator {
             BinaryOperator::BitwiseOr => f.write_str("|"),
             BinaryOperator::BitwiseAnd => f.write_str("&"),
             BinaryOperator::BitwiseXor => f.write_str("^"),
+            BinaryOperator::Regexp => f.write_str("REGEXP"),
             BinaryOperator::PGBitwiseXor => f.write_str("#"),
             BinaryOperator::PGBitwiseShiftLeft => f.write_str("<<"),
             BinaryOperator::PGBitwiseShiftRight => f.write_str(">>"),
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -10,6 +10,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(not(feature = "std"))]
+use alloc::boxed::Box;
+
 use crate::{
     ast::{BinaryOperator, Expr},
     dialect::Dialect,
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -10,7 +10,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::dialect::Dialect;
+use crate::{
+    ast::{BinaryOperator, Expr},
+    dialect::Dialect,
+    keywords::Keyword,
+};
 
 /// [MySQL](https://www.mysql.com/)
 #[derive(Debug)]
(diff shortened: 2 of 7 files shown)


ankrgyl:
> I'm getting a test failure. What's the best test for this?

Looks like your code doesn't compile with no-std? https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/4928867591/jobs/8807772910?pr=874

> Should REGEXP be a keyword? If not, what would we parse?

Adding a keyword should work (they are not automatically reserved)

max-sixty:
> >  I'm getting a test failure. What's the best test for this?

> Looks like your code doesn't compile with no-std? https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/4928867591/jobs/8807772910?pr=874

I think I can solve that one; but https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/4928867591/jobs/8807773893?pr=874#step:5:504 needs some crate-specific guidance...

ankrgyl:
It looks like the new test you wrote is failing, and it reproduces locally with:

```
cargo test --test sqlparser_mysql
```

Can you be more specific about what kind of guidance you need?

max-sixty:
I'm fine reproducing the test. But I'm not sure what the best test for this is — is it `verified_stmt`? I don't see any tests on dialects' `parse_infix` (though there's a more general test of the functionality [here](https://github.com/sqlparser-rs/sqlparser-rs/blob/a8a8e65b7c7dddaba79d1d1553aabd88ae9c2568/tests/sqlparser_custom_dialect.rs#L88)).

Any thoughts why the test is failing? It looks like it's expecting the expression to finish after `REGEXP`...

max-sixty:
OK, one reason is that it never gets to `parse_infix`; this diff produces the same error:

```diff
diff --git a/src/dialect/mysql.rs b/src/dialect/mysql.rs
index 9e82702..62d2624 100644
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -49,6 +49,7 @@ fn parse_infix(
         _expr: &crate::ast::Expr,
         _precedence: u8,
     ) -> Option<Result<crate::ast::Expr, crate::parser::ParserError>> {
+        panic!();
         // Parse REGEXP as an operator
         if _parser.parse_keyword(Keyword::REGEXP) {
             Some(Ok(Expr::BinaryOp {

```

...so this would benefit from some guidance from someone who knows this — thank you

ankrgyl:
Have you verified that the change works outside of the test?

ankrgyl:
> OK, one reason is that it never gets to `parse_infix`; this diff produces the same error:
> 
> ```diff
> diff --git a/src/dialect/mysql.rs b/src/dialect/mysql.rs
> index 9e82702..62d2624 100644
> --- a/src/dialect/mysql.rs
> +++ b/src/dialect/mysql.rs
> @@ -49,6 +49,7 @@ fn parse_infix(
>          _expr: &crate::ast::Expr,
>          _precedence: u8,
>      ) -> Option<Result<crate::ast::Expr, crate::parser::ParserError>> {
> +        panic!();
>          // Parse REGEXP as an operator
>          if _parser.parse_keyword(Keyword::REGEXP) {
>              Some(Ok(Expr::BinaryOp {
> ```
> 
> ...so this would benefit from some guidance from someone who knows this — thank you

I'm not particularly familiar with this code, but it looks like you need to set a precedence in `get_next_precedence`.

max-sixty:
> Have you verified that the change works outside of the test?

No — I'm using the test to verify it!

To be clear — and I'm hoping this is a reasonable request as part of a contribution — I'm hoping someone who is familiar with the library can offer some specific guidance on the proposed code — is this how they would do it / where is it likely incorrect?

max-sixty:
> I'm not particularly familiar with this code, but it looks like you need to set a precedence in `get_next_precedence`.

Perfect, thanks!

This is working on my end now; let's see whether CI passes

alamb:
I feel like https://github.com/sqlparser-rs/sqlparser-rs/pull/868 is a better (simpler) approach -- given that do we still want to try and push this one forward?

alamb:
> I'm totally fine with https://github.com/sqlparser-rs/sqlparser-rs/pull/868! Tbh I was doing this to be a good citizen, as we're such heavy users of sqlparser-rs, rather than needing it per-se.

Thank you for the kind works and the sense of civic responsibility ❤️

alamb:
I believe https://github.com/sqlparser-rs/sqlparser-rs/pull/868 has superseded this PR

PR #878, opened 2023-05-10
feat: add DuckDB dialect

### LINKED ISSUE #873: Add DuckDB's integer devision operator `//`
Maybe related to #863 and #868

DuckDB introduced integer division operator `//` in the next release 0.8.0. (duckdb/duckdb#7082)
Is it worth supporting this here?

eitsupi:
Close #873

Add DuckDB dialect and support DuckDB's `//` operator.


eitsupi:
Perhaps there is quite a lack of testing for dialects here.

I am not very experienced with this repository and DuckDB, so it would be great if someone could help me with this.

src/ast/operator.rs
```diff
@@ -85,6 +85,7 @@ pub enum BinaryOperator {
     BitwiseOr,
     BitwiseAnd,
     BitwiseXor,
+    DuckIntegerDivide,
```
alamb, line 88:
```suggestion
    /// Integer division operator `//` in DuckDB
    DuckIntegerDivide,
```

alamb (commented):
Thank you for the contribution @eitsupi !

>  Perhaps there is quite a lack of testing for dialects here.

I think you need to add `DuckDbDialect` to the list in the following location and the new `DuckDbDialect` will be well covered by tests/sqlparser_common.rs

https://github.com/sqlparser-rs/sqlparser-rs/blob/4559d87a824a6a268a350d02e63c51b77e41bd34/src/test_utils.rs#L158-L173

src/tokenizer.rs
```diff
@@ -768,6 +771,9 @@ impl<'a> Tokenizer<'a> {
                                 comment,
                             })))
                         }
+                        Some('/') if dialect_of!(self is DuckDbDialect) => {
```
alamb, line 774: Can you also check here for `GenericDialect` as well?

alamb:
This is looking pretty good @eitsupi  --  it has a conflict that needs to be resolved and I think the test needs a little updating but otherwise it is ready to go

alamb (approved):
Looks great to me -- thank you @eitsupi

Issue #875, opened 2023-05-10
Add support for multiple expressions in order by clause for aggregate functions


mustafasrepo:
sqlparser can parse query below successfully,
```sql
SELECT ARRAY_AGG(x ORDER BY x) AS a FROM T
```
However, it cannot parse query below
```sql
SELECT ARRAY_AGG(x ORDER BY x, y) AS a FROM T
```
It assumes that order by will contain single entry. However, second form is valid also. Postgres can run queries in the second form.
Add support for multiple expression in the order by clause for aggregate functions.

PR #879, opened 2023-05-10
Add support for multiple expressions, order by in aggregations

mustafasrepo:
Closes [#875](https://github.com/sqlparser-rs/sqlparser-rs/issues/875).
With this Pr sqlparser supports array aggregation with lexicographical ordering such as below. 
```sql
SELECT ARRAY_AGG(x ORDER BY x, y) AS a FROM T
```
Previously only single expression were supported as in below.
```sql
SELECT ARRAY_AGG(x ORDER BY x) AS a FROM T
```


alamb (approved):
Thank you @mustafasrepo

Issue #877, opened 2023-05-10
Add support for FIRST, LAST aggregation parsing.


mustafasrepo:
When we run the query below currently, sqlparser cannot parse it
```sql
SELECT FIRST(x ORDER BY x) AS a FROM T
```
similarly, it cannot parse corresponding query with `LAST` either.
although it is not a standard feature, this function is very common and helpful. I would like to have support for these functions.

PR #882, opened 2023-05-18
Add support for first, last aggregate function parsing

mustafasrepo:
Closes [#877](https://github.com/sqlparser-rs/sqlparser-rs/issues/877)

This PR adds support for order by expressions (they may be unnecessary though such as for `SUM`, `MIN`, etc.) inside aggregate functions.
As an example, see queries below
```sql
SELECT SUM(x ORDER BY x) AS a FROM T
```
or 
```sql
SELECT SUM(x ORDER BY x, y) AS a FROM T
```
With this PR we can parse above queries successfully.


alamb (approved):
Looks good to me -- thank you @mustafasrepo

PR #883, opened 2023-05-18
truncate: table as optional keyword

mobuchowski:
`TABLE` is an optional keyword in Postgres: https://www.postgresql.org/docs/current/sql-truncate.html


alamb (approved):
Makes sense to me -- thank you @mobuchowski

Issue #895, opened 2023-06-13
Modify `Dialect` trait to include `supports_group_by_expr` function


jdye64:
The current `Dialect` trait has no mechanism for determining if the implementing instance supports `Grouping Exprs`. This means that any implementation of the `Dialect` trait has no way to inform sqlparser-rs that it supports the `Rollup` clause for example.

Adding a function to `Dialect` like already existing functions, `supports_filter_during_aggregation`, for example would allow the implementation to inform the parser if it does in fact support grouping exprs or not.

The list is currently hardcoded in `parser.rs` to `if dialect_of!(self is PostgreSqlDialect | DuckDbDialect | GenericDialect)`

PR #896, opened 2023-06-13
Add fn support_group_by_expr to Dialect trait

jdye64:
This closes: #895

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -975,7 +975,7 @@ impl<'a> Parser<'a> {
     /// parse a group by expr. a group by expr can be one of group sets, roll up, cube, or simple
     /// expr.
     fn parse_group_by_expr(&mut self) -> Result<Expr, ParserError> {
-        if self.dialect.supports_filter_during_aggregation() {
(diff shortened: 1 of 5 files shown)


src/parser.rs
```diff
@@ -975,7 +975,7 @@ impl<'a> Parser<'a> {
     /// parse a group by expr. a group by expr can be one of group sets, roll up, cube, or simple
     /// expr.
     fn parse_group_by_expr(&mut self) -> Result<Expr, ParserError> {
-        if dialect_of!(self is PostgreSqlDialect | DuckDbDialect | GenericDialect) {
+        if self.dialect.supports_filter_during_aggregation() {
```
andygrove, line 978: I think this should be:

```suggestion
        if self.dialect.supports_group_by_expr() {
```

jdye64: Thanks for catching that Andy. Seems my autocomplete got a little excited there.

andygrove (approved):
LGTM

andygrove:
cc @Jimexist @alamb

alamb (approved):
Thank you @jdye64

PR #897, opened 2023-06-13
Add support for DuckDB's CREATE MACRO statements

MartinNowak:
- support scalar and table macros
- separate from related but different and already complex functions
  parsing to avoid plenty of runtime dialect switches

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1580,6 +1580,32 @@ pub enum Statement {
         params: CreateFunctionBody,
     },
     /// ```sql
+    /// CREATE MACRO
+    /// ```
+    ///
+    /// Supported variants:
+    /// 1. [DuckDB](https://duckdb.org/docs/sql/statements/create_macro)
+    CreateMacro {
+        or_replace: bool,
+        temporary: bool,
+        name: ObjectName,
+        args: Option<Vec<OperateMacroArg>>,
+        expr: Expr,
+    },
+    /// ```sql
+    /// CREATE MACRO
+    /// ```
+    ///
+    /// Supported variants:
+    /// 1. [DuckDB](https://duckdb.org/docs/sql/statements/create_macro)
+    CreateTableMacro {
+        or_replace: bool,
+        temporary: bool,
+        name: ObjectName,
+        args: Option<Vec<OperateMacroArg>>,
+        query: Query,
+    },
+    /// ```sql
     /// CREATE STAGE
     /// ```
     /// See <https://docs.snowflake.com/en/sql-reference/sql/create-stage>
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -1580,6 +1580,32 @@ pub enum Statement {
...
+    /// ```
+    ///
+    /// Supported variants:
+    /// 1. [DuckDB](https://duckdb.org/docs/sql/statements/create_macro)
+    CreateMacro {
+        or_replace: bool,
+        temporary: bool,
+        name: ObjectName,
+        args: Option<Vec<OperateMacroArg>>,
+        expr: Expr,
```
alamb, line 1593: I wonder if you considered combining the structs to reduce duplication like:

```
enum MacroDefinition {
  Expr(Expr),
  Table(Query)
}

pub enum Statement {
...
 CreateMacro {
        or_replace: bool,
        temporary: bool,
        name: ObjectName,
        args: Option<Vec<OperateMacroArg>>,
        definition: MacroDefinition,
  }
...
```

src/ast/mod.rs
```diff
@@ -4304,6 +4368,39 @@ impl fmt::Display for CreateFunctionUsing {
     }
 }
 
+/// DuckDB specific feature.
```
alamb, line 4371:
```suggestion
/// `NAME = <EXPR>` arguments for DuckDB macros
```

src/ast/mod.rs
```diff
@@ -4304,6 +4368,39 @@ impl fmt::Display for CreateFunctionUsing {
...
+///
+/// See [Create Macro - DuckDB](https://duckdb.org/docs/sql/statements/create_macro)
+/// for more details
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct OperateMacroArg {
```
alamb, line 4378: What is the meaning of the `Operate` prefix? Maybe this could be shortened to `MacroArg`?

src/parser.rs
```diff
@@ -2624,6 +2626,8 @@ impl<'a> Parser<'a> {
                 return_type,
                 params,
             })
+        } else if dialect_of!(self is DuckDbDialect) {
```
alamb, line 2629: Can you please also allow `GenericDialect` as well (which typically allows the superset of everything supported by sqlparser)?

tests/sqlparser_duckdb.rs
```diff
@@ -68,3 +68,53 @@ fn test_select_wildcard_with_exclude() {
 fn parse_div_infix() {
     duckdb_and_generic().verified_stmt(r#"SELECT 5 // 2"#);
 }
+
+#[test]
+fn test_create_macro() {
+    let _macro = duckdb().verified_stmt("CREATE MACRO schema.add(a, b) AS a + b");
```
alamb, line 74: typically the `_` prefix is used in rust for variables that are not used to suppress the compiler from warning. I think it should be avoided

Maybe you could call this `stmt` or something to avoid clashing with the rust `macro` keyword but also not starting your variable with `_`

alamb (commented):
Thank you @MartinNowak  -- this is looking pretty close. I left some comments let me know what you think

src/ast/mod.rs
MartinNowak: Got it from OperateFunctionArg.

src/parser.rs
MartinNowak: I allowed GenericDialect for `CREATE MACRO` but not for their `CREATE FUNCTION` alias.
Mostly because duckdb's macros don't seem like a common denominator for SQL functions (if there is any).
More specifically because SQL functions are typically typed while duckdb macros are not.

I could adjust the failing test https://github.com/MartinNowak/sqlparser-rs/blob/088b2fe0759311801740d5ba490d7acb376a4e4a/tests/sqlparser_hive.rs#L268-L278 and try to merge with the existing FunctionArg (using an optional type), but presumably the AST is going the end up very messy and code will need plenty of dialect switches/config options.

MartinNowak:
> Thank you @MartinNowak -- this is looking pretty close. I left some comments let me know what you think

Thanks for the detailed review @alamb, resolved all points except [this](https://github.com/sqlparser-rs/sqlparser-rs/pull/897#discussion_r1232636866) discussion.

PR #900, opened 2023-06-17
Support basic CREATE PROCEDURE of MSSQL

delsehi:
Hi! This is my first time trying to contribute to open-source :) 

This PR adds support for creating procedures as in MSSQL. 

Example 

"CREATE PROCEDURE foo (@someParam VARCHAR(50)) AS BEGIN SELECT * FROM test END"

It's not extensive and some features are not implemented yet as per the tsql spec, but the basics that I use at work seem to work fine.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -980,7 +980,7 @@ impl<'a> Parser<'a> {
     /// parse a group by expr. a group by expr can be one of group sets, roll up, cube, or simple
     /// expr.
     fn parse_group_by_expr(&mut self) -> Result<Expr, ParserError> {
-        if dialect_of!(self is PostgreSqlDialect | DuckDbDialect | GenericDialect) {
+        if self.dialect.supports_group_by_expr() {
             if self.parse_keywords(&[Keyword::GROUPING, Keyword::SETS]) {
                 self.expect_token(&Token::LParen)?;
                 let result = self.parse_comma_separated(|p| p.parse_tuple(false, true))?;
@@ -2352,6 +2352,8 @@ impl<'a> Parser<'a> {
             self.parse_create_external_table(or_replace)
         } else if self.parse_keyword(Keyword::FUNCTION) {
             self.parse_create_function(or_replace, temporary)
+        } else if self.parse_keyword(Keyword::MACRO) {
+            self.parse_create_macro(or_replace, temporary)
         } else if or_replace {
             self.expected(
                 "[EXTERNAL] TABLE or [MATERIALIZED] VIEW or FUNCTION after CREATE OR REPLACE",
@@ -2632,6 +2634,8 @@ impl<'a> Parser<'a> {
                 return_type,
                 params,
             })
+        } else if dialect_of!(self is DuckDbDialect) {
+            self.parse_create_macro(or_replace, temporary)
         } else {
             self.prev_token();
             self.expected("an object type after CREATE", self.peek_token())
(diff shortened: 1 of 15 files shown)


delsehi:
I maybe got ahead of myself here. 

I added "END" as a reserved keyword, but it seems it's not reserved in MySQL for example (maybe some tests should be added for that). 

Would it be ok if in parse_comma_separated, the RESERVED_FOR_COLUMN_ALIAS could be joined with a dialect specific list of keywords?

Or is there a better way to handle the statements within the BEGIN-END block in my code?

src/parser.rs
```diff
@@ -345,6 +345,15 @@ impl<'a> Parser<'a> {
             while self.consume_token(&Token::SemiColon) {
                 expecting_statement_delimiter = false;
             }
+            if self.peek_token()
```
alamb, line 348: I don't understand the need for this code -- it seems like breaking parsing on `END` should only happen in the context of parsing a procedure.

A pattern used elsewhere in the code might be to conditionalize based on a parameter. Something like

```rust
    pub fn parse_statements(&mut self, in_procedure: bool) -> Result<Vec<Statement>, ParserError> {
```

For example

src/parser.rs
```diff
@@ -7031,6 +7071,21 @@ impl<'a> Parser<'a> {
         Ok(NamedWindowDefinition(ident, window_spec))
     }
 
+    pub fn parse_create_procedure(&mut self, or_alter: bool) -> Result<Statement, ParserError> {
+        let name = self.parse_object_name()?;
+        let params = self.parse_optional_procedure_parameters()?;
+        self.expect_keyword(Keyword::AS)?;
+        self.expect_keyword(Keyword::BEGIN)?;
+        let statements = self.parse_statements()?;
```
alamb, line 7079: If you need special behavior of parse_statements, I think you may have to pass in the parameter here

alamb:
Hi @delsehi  -- this is looking very cool -- thank you for the contribution


> I added "END" as a reserved keyword, but it seems it's not reserved in MySQL for example (maybe some tests should be added for that).

I am not quite sure what problem you are hitting -- perhaps adding some tests, as you suggest, would make the problem clearer. 


> Or is there a better way to handle the statements within the BEGIN-END block in my code?

I left some comments -- not sure if they are helpful or not

delsehi:
> I am not quite sure what problem you are hitting -- perhaps adding some tests, as you suggest, would make the problem clearer.
> 
It wasn't really a problem for me, but "END" is not a reserved keyword in MySQL and if a MySQL user would try to use END as a column or table alias, my change would a cause a bug for them. 

I've made another branch where I added an extension of the reserved keywords that are specific to TSql and appended them only if the current dialect is set to MsSQL. That didn't not break things for MySQL. 

However, maybe it should be the other way around. END is a reserved keyword in the ANSI standard 2023, so maybe the MySQL users should be the ones implementing an opt-out for that as a reserved keyword if they'd want to use END as an alias? 


An actual issue with my implementation though is how I peek at the next token for the END keyword. I make a Word struct, but it's not case insensitive, so if END is in lowercase it wouldn't be parsed as the end of the BEGIN-END.

delsehi:
@alamb I added the parameter as you suggested, and found a more idiomatic way to check if the keyword is END (that works with lowercase too). 

Anything else I should consider to make this more mergeable?

alamb:
Thanks @delsehi  -- I am a behind on reviews. I plan to review this tomorrow. Sorry for the delay

delsehi:
@alamb No hurry! I'm gonna celebrate swedish midsummer this weekend in any case ❤️

Sorry for asking for feedback this early but I was certain I'd get something wrong and thought it'd be better to solve it early on. 

Unless you see something truly crazy you can delay any feedback until I find a way to make the no-std checks pass

alamb:
> @alamb No hurry! I'm gonna celebrate swedish midsummer this weekend in any case ❤️

Awsome!

> Sorry for asking for feedback this early but I was certain I'd get something wrong and thought it'd be better to solve it early on.

No worries -- that was a good thing to do. I just feel bad I don't have as much time as I would like for reviewing PRs in this crate

tests/sqlparser_mssql.rs
```diff
@@ -55,6 +55,80 @@ fn parse_mssql_delimited_identifiers() {
...
+            name: ObjectName(vec![Ident {
+                value: "test".into(),
+                quote_style: None
+            }])
+        }
+    )
+}
+
+#[test]
+fn parse_mssql_create_procedure() {
```
alamb, line 120: Can you also add a test for multiple statements? All these tests seem to have just a single one

alamb (commented):
Thank you @delsehi - I made https://github.com/delsehi/sqlparser-rs/pull/1 that targets your PR with some proposed changes which I think simplify the API and help resolve some of the CI failures. Give it a look at let me know what you think

Thanks again for the contribution

alamb:
FYI I plan to release a new version of sqlparser-rs tomorrow -- it would be awesome if we can get this PR in

I think it is missing:
1. Some multi-statement tests
2. Fix a conflict

Is there any chance you will be able to work on that for tomorrow @delsehi ?

delsehi:
@alamb I stressed together this thing. I hope that it works but I won't have any more time today or tomorrow I'm afraid

tests/sqlparser_mssql.rs
```diff
@@ -55,6 +55,88 @@ fn parse_mssql_delimited_identifiers() {
...
+        "CREATE PROCEDURE foo AS BEGIN SELECT [myColumn] FROM [myschema].[mytable] END",
+    );
+    let _ = ms_and_generic().verified_stmt(
+        "CREATE PROCEDURE foo (@CustomerName NVARCHAR(50)) AS BEGIN SELECT * FROM DEV END",
+    );
+    let _ = ms().verified_stmt("CREATE PROCEDURE [foo] AS BEGIN UPDATE bar SET col = 'test' END");
+    // Test a statement with END in it
+    let _ = ms().verified_stmt("CREATE PROCEDURE [foo] AS BEGIN SELECT [foo], CASE WHEN [foo] IS NULL THEN 'empty' ELSE 'notempty' END AS [foo] END");
+    // Multiple statements
+    let _ = ms().verified_stmt("CREATE PROCEDURE [foo] AS BEGIN UPDATE bar SET col = 'test'; SELECT [foo] FROM BAR WHERE [FOO] > 10 END");
```
alamb, line 137: ❤️

alamb (approved):
Looks great -- thank you @delsehi

alamb:
I really appreciate pushing to get it done before the release

PR #902, opened 2023-06-22
fix: unary negation operator with operators: `Mul`, `Div` and `Mod`

izveigor:
One part of https://github.com/apache/arrow-datafusion/issues/6649.
Unary operator does not work correctly with operators: `Mul`, `Div` and `Mod`.
For example (Arrow-Datafusion):
```
❯ select - 1234 % 12 % 3 * 1545 % 23;
+------------------------------------------------------------------+
| (- Int64(1234) % Int64(12) % Int64(3) * Int64(1545) % Int64(23)) |
+------------------------------------------------------------------+
| -4                                                               |
+------------------------------------------------------------------+
1 row in set. Query took 0.023 seconds.
```
But should be:
```
❯ select - 1234 % 12 % 3 * 1545 % 23;
+------------------------------------------------------------------+
| (- Int64(1234) % Int64(12) % Int64(3) * Int64(1545) % Int64(23)) |
+------------------------------------------------------------------+
| -17                                                              |
+------------------------------------------------------------------+
1 row in set. Query took 0.023 seconds.
```

The main problem is the wrong precedence (See https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser.rs#L797)

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -1074,7 +1074,7 @@ fn parse_compound_expr_2() {
 }
 
 #[test]
-fn parse_unary_math() {
+fn parse_unary_math_with_plus() {
     use self::Expr::*;
     let sql = "-a + -b";
     assert_eq!(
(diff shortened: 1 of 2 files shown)


tests/sqlparser_common.rs
```diff
@@ -1093,6 +1093,26 @@ fn parse_unary_math() {
     );
 }
 
+#[test]
+fn parse_unary_math_with_multiply() {
+    use self::Expr::*;
+    let sql = "-a * -b";
```
alamb, line 1099: I ran this case without any of the code changes in this PR and it fails like the following (aka it is parased as `-(a * -b)` 

```
---- parse_unary_math_with_multiply stdout ----
thread 'parse_unary_math_with_multiply' panicked at 'assertion failed: `(left == right)`
  left: `BinaryOp { left: UnaryOp { op: Minus, expr: Identifier(Ident { value: "a", quote_style: None }) }, op: Multiply, right: UnaryOp { op: Minus, expr: Identifier(Ident { value: "b", quote_style: None }) } }`,
 right: `UnaryOp { op: Minus, expr: BinaryOp { left: Identifier(Ident { value: "a", quote_style: None }), op: Multiply, right: UnaryOp { op: Minus, expr: Identifier(Ident { value: "b", quote_style: None }) } } }`', tests/sqlparser_common.rs:1100:5
stack backtrace:
```

alamb (approved):
Thank you @izveigor  -- this makes sense to me

Issue #886, opened 2023-05-19
Reques Support for `ALTER VIEW` parsing 


eNipu:
Currently there to no support for `ALTER VIEW` DDL.

parkma99:
Hello , I would like to work it. Which SQL syntax need to support?

PostgreSQL syntax or other SQL syntax?

eNipu:
Hi @parkma99, great! Support for PostgreSQL will be a great help!

alamb:
Note https://github.com/sqlparser-rs/sqlparser-rs/pull/907 adds support for mysql syntax -- it would likely be nice to support posgres syntax too

PR #907, opened 2023-06-25
Add support for MySQL ALTER VIEW

liadgiladi:
Hi,

This PR adds support for MySQL ALTER VIEW syntax.

#886 is relevant here, but I now see that OP replied yesterday asking for PostgreSQL syntax, which is different. The use case I'm interested in is MySQL.


alamb (approved):
Thank you very much for the contribution @liadgiladi 

I took the liberty of running `cargo fmt` and pushing to this branch to get the CI checks to pass.

PR #909, opened 2023-06-28
Support multi args for unnest

### LINKED ISSUE #908: Support unnest with multiple arrays
Ref https://www.postgresql.org/docs/current/functions-array.html

`select * from unnest(ARRAY[1,2], ARRAY['foo','bar','baz'])`
`select * from unnest(make_array(1,2),make_array('a','b','c')) as data(num,str)`

We need to parse those kind of SQL for https://github.com/apache/arrow-datafusion/issues/6555

Current Error:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected ), found: ,")', src/main.rs:14:48

Expected behavior:
No Error and probably parse unnest as function or table?

I will work on this

jayzhan211:
Close #908

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -677,7 +677,7 @@ pub enum TableFactor {
     /// ```
     UNNEST {
         alias: Option<TableAlias>,
-        array_expr: Box<Expr>,
+        array_exprs: Vec<Box<Expr>>,
         with_offset: bool,
         with_offset_alias: Option<Ident>,
     },
@@ -749,11 +749,16 @@ impl fmt::Display for TableFactor {
             }
             TableFactor::UNNEST {
                 alias,
-                array_expr,
+                array_exprs,
                 with_offset,
                 with_offset_alias,
             } => {
-                write!(f, "UNNEST({array_expr})")?;
+                match array_exprs.len() {
+                    0 => write!(f, "UNNEST()")?,
+                    1 => write!(f, "UNNEST({})", array_exprs[0])?,
+                    _ => write!(f, "UNNEST({})", display_comma_separated(array_exprs))?,
+                }
+
                 if let Some(alias) = alias {
                     write!(f, " AS {alias}")?;
                 }
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -753,11 +753,7 @@ impl fmt::Display for TableFactor {
                 with_offset,
                 with_offset_alias,
             } => {
-                match array_exprs.len() {
-                    0 => write!(f, "UNNEST()")?,
-                    1 => write!(f, "UNNEST({})", array_exprs[0])?,
-                    _ => write!(f, "UNNEST({})", display_comma_separated(array_exprs))?,
-                }
+                write!(f, "UNNEST({})", display_comma_separated(array_exprs))?;
 
                 if let Some(alias) = alias {
                     write!(f, " AS {alias}")?;
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -677,7 +677,7 @@ pub enum TableFactor {
     /// ```
     UNNEST {
         alias: Option<TableAlias>,
-        array_exprs: Vec<Box<Expr>>,
+        array_exprs: Vec<Expr>,
         with_offset: bool,
         with_offset_alias: Option<Ident>,
     },
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -5980,8 +5980,16 @@ impl<'a> Parser<'a> {
             && self.parse_keyword(Keyword::UNNEST)
         {
             self.expect_token(&Token::LParen)?;
-            let expr = self.parse_expr()?;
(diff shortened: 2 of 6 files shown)


jayzhan211:
@alamb Ready for review, thanks.

src/ast/query.rs
```diff
@@ -677,7 +677,7 @@ pub enum TableFactor {
     /// ```
     UNNEST {
         alias: Option<TableAlias>,
-        array_exprs: Vec<Box<Expr>>,
```
alamb: 👍

src/ast/query.rs
```diff
@@ -749,11 +749,16 @@ impl fmt::Display for TableFactor {
...
+                array_exprs,
                 with_offset,
                 with_offset_alias,
             } => {
-                write!(f, "UNNEST({array_expr})")?;
+                match array_exprs.len() {
+                    0 => write!(f, "UNNEST()")?,
+                    1 => write!(f, "UNNEST({})", array_exprs[0])?,
+                    _ => write!(f, "UNNEST({})", display_comma_separated(array_exprs))?,
+                }
```
alamb, line 760: `display_comma_separated ` already handles 0,1,many. This code can be simplified like this:

```suggestion
                write!(f, "UNNEST({})", display_comma_separated(array_exprs))?;
```

src/parser.rs
```diff
@@ -5979,9 +5991,7 @@ impl<'a> Parser<'a> {
         } else if dialect_of!(self is BigQueryDialect | GenericDialect)
             && self.parse_keyword(Keyword::UNNEST)
         {
-            self.expect_token(&Token::LParen)?;
-            let expr = self.parse_expr()?;
-            self.expect_token(&Token::RParen)?;
+            let exprs = self.parse_unnest_exprs()?;
```
alamb, line 5994: I think the standard pattern for parsing exprs in sqlparser is like this https://github.com/sqlparser-rs/sqlparser-rs/blob/efd8cb7fd1c4e6654b101868f7a76d80f66da74b/src/parser.rs#L484-L488

So in this case like this (and can remove the `parse_unnest_exprs`)

```suggestion
            self.expect_token(&Token::LParen)?;
            let exprs = self.parse_comma_separated(Parser::parse_expr)?;
            self.expect_token(&Token::RParen)?;
```

src/parser.rs
```diff
@@ -7710,6 +7720,22 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_unnest() {
```
alamb, line 7724: Can you please move these tests to the other parser tests in https://github.com/sqlparser-rs/sqlparser-rs/blob/efd8cb7fd1c4e6654b101868f7a76d80f66da74b/tests/sqlparser_common.rs#L1361-L1380

Using `one_statement_parses_to` might be good to verify the round trip. Something like

https://github.com/sqlparser-rs/sqlparser-rs/blob/efd8cb7fd1c4e6654b101868f7a76d80f66da74b/tests/sqlparser_common.rs#L1805-L1808

alamb (commented):
Thank you @jayzhan211 -- this is a nice improvement.  👌 

I left some comments for making this PR more closely follow the rest of the codebase. Let me know what you think ❤️

src/parser.rs
jayzhan211: Unnest does not support for all dialect so I cant test with `one_statement_parses_to`.
Can we change remove #L5998 to support all dialect?

https://github.com/sqlparser-rs/sqlparser-rs/blob/f05f71e20dc54f14cac80a638e2ed252b6fca91c/src/parser.rs#L5998-L5999

jayzhan211: I can't easily add a test with `one_statement_parses_to`.
I got the error `Parse results with GenericDialect are different from PostgreSqlDialect`.

src/parser.rs
```diff
@@ -5980,7 +5980,7 @@ impl<'a> Parser<'a> {
             && self.parse_keyword(Keyword::UNNEST)
         {
             self.expect_token(&Token::LParen)?;
-            let expr = self.parse_expr()?;
+            let array_exprs = self.parse_comma_separated(Parser::parse_expr)?;
```
alamb, line 5983: ❤️

alamb (approved):
Thank you @jayzhan211  -- this looks great!

src/parser.rs
alamb: I don't understand this comment -- I believe the PR looks really nice and well tested now. Let me know if I missed anything

alamb:
@jayzhan211  I took the liberty to fix the tests on this PR when run with `--all-features`

alamb:
Thanks again @jayzhan211

src/parser.rs
jayzhan211: I just want to show that these two test cases failed with `one_statement_parses_to`, so I did not add them to the test. I am not sure if these are expected behavior.

PR #910, opened 2023-06-30
Remove most instances of `#[cfg(feature(bigdecimal))]` in tests

alamb:
# Rationale
Inspired by errors on https://github.com/sqlparser-rs/sqlparser-rs/pull/909 and my own struggles creating `Value::Number` I dug around in the code and found a better way to create them -- `test_utils::number`

It turns out that there were quite a few tests that used `cfg` which made them harder to understand and more brittle (you have to run the tests with different options to cover all the code)

# Changes
Remove almost all uses of `cfg` in the tests and use `test_utils::number` instead


alamb:
cc @ankrgyl  in case this is interesting to you

src/ast/value.rs
```diff
@@ -32,6 +32,9 @@ pub enum Value {
     #[cfg(not(feature = "bigdecimal"))]
     Number(String, bool),
     #[cfg(feature = "bigdecimal")]
+    // HINT: use `test_utils::number` to make an instance of
```
alamb, line 35: I figured I would leave a breadcrumb to help future users find `test_utils:::number`

tests/sqlparser_duckdb.rs
```diff
@@ -97,13 +97,7 @@ fn test_create_macro_default_args() {
...
             MacroArg {
                 name: Ident::new("b"),
-                default_expr: Some(Expr::Value(Value::Number(
-                    #[cfg(not(feature = "bigdecimal"))]
-                    5.to_string(),
-                    #[cfg(feature = "bigdecimal")]
-                    bigdecimal::BigDecimal::from(5),
-                    false,
-                ))),
+                default_expr: Some(Expr::Value(number("5"))),
```
alamb, line 100: I think this style is both more concise and readable

ankrgyl:
This is awesomeness

PR #915, opened 2023-07-12
Support UNION (ALL) BY NAME syntax

### LINKED ISSUE #899: Support `UNION (ALL) BY NAME` syntax (duckdb, spark)
This is a very useful form of `UNION` that has the backend align the columns on either side of the union: https://duckdb.org/docs/sql/query_syntax/setops.html#union-all-by-name.

Would be wonderful to have parsing support :)
Many thanks!

---
```sql
SELECT * FROM table_x UNION BY NAME SELECT * FROM table_y;
```

parkma99:
close #899 

### have testing:

Yes, add two test cases in tests/sqlparser_duckdb.rs

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -110,9 +110,10 @@ impl fmt::Display for SetExpr {
             } => {
                 write!(f, "{left} {op}")?;
                 match set_quantifier {
-                    SetQuantifier::All | SetQuantifier::Distinct | SetQuantifier::ByName(_) => {
-                        write!(f, " {set_quantifier}")?
-                    }
+                    SetQuantifier::All
+                    | SetQuantifier::Distinct
+                    | SetQuantifier::ByName
+                    | SetQuantifier::AllByName => write!(f, " {set_quantifier}")?,
                     SetQuantifier::None => write!(f, "{set_quantifier}")?,
                 }
                 write!(f, " {right}")?;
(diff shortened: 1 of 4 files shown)


alamb (commented):
Thank you @parkma99  for this PR 🙏 

Did you consider using the existing parser structures for UNION / UNION ALL , which I think would take less code and also be more consistent with the rest of this codebase

The parsing happens here:

https://github.com/sqlparser-rs/sqlparser-rs/blob/653346c4d6e67001e492a41920f3c8903267a547/src/parser.rs#L5265-L5320


For example, perhaps `SetQuantifier` could be extended

https://github.com/sqlparser-rs/sqlparser-rs/blob/d6ebb58b969e0eb62fd34ed228ff3d51013e5ea5/src/ast/query.rs#L148-L152

with a `AllByName` variant

🤔

parkma99:
Thanks @alamb , I will have a look how using the existing parser structures for UNION / UNION ALL.

src/ast/query.rs
```diff
@@ -148,6 +150,7 @@ impl fmt::Display for SetOperator {
 pub enum SetQuantifier {
     All,
     Distinct,
+    ByName(bool),
```
alamb, line 153: I suggest making two different enums so it is more self-describing: `ByName` and `AllByName` but I don't it is required. If we don't add different enums then I think we should at least add some comments:

```suggestion
    /// if true, `ALL BY NAME`, if false `BY NAME`
    ByName(bool),
```

alamb (approved):
Looks great @parkma99  -- thank you very much 🙏 

I have only one more suggestion related to the enum names. But I can do it as a follow on PR as well if you prefer

alamb (approved):
Thanks @parkma99

PR #916, opened 2023-07-13
Support `DROP TEMPORARY TABLE`, MySQL syntax

liadgiladi:
Hi,

This PR adds support for MySQL DROP TEMPORARY TABLE syntax.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1401,6 +1401,8 @@ pub enum Statement {
         /// Hive allows you specify whether the table's stored data will be
         /// deleted along with the dropped table
         purge: bool,
+        /// MySQL-specific "TEMPORARY" keyword
+        temporary: bool,
     },
     /// DROP Function
     DropFunction {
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -1401,6 +1401,8 @@ pub enum Statement {
         /// Hive allows you specify whether the table's stored data will be
         /// deleted along with the dropped table
         purge: bool,
+        /// MySQL-specific "TEMPORARY" keyword
+        temporary: Option<bool>,
```
alamb, line 1405: I recommend making this `temporary: bool` rather than trying to distinguish between when it was  supported and not (Some/None)

I think this change would make writing the tests easier and it would make the `temporary` flag more consistent with the rest of the codebase

src/parser.rs
```diff
@@ -3078,6 +3078,13 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_drop(&mut self) -> Result<Statement, ParserError> {
+        // MySQL dialect supports `TEMPORARY`
+        let temporary = if dialect_of!(self is MySqlDialect) {
```
alamb, line 3082: Can we also please parse if it is a dialect of `GenericDialect`? Generic Dialect is sort of a catchall dialect that supports everything

liadgiladi:
Hi @alamb,

Thanks for your feedback. I've updated the PR accordingly.

alamb (approved):
looks good to me -- thank you @liadgiladi

PR #927, opened 2023-07-20
fix parsing of identifiers after `%` symbol

alamb:
Fixes a bug in parsing 'c%d1` where the `d` was accidentally consumed. 

Introduced in https://github.com/sqlparser-rs/sqlparser-rs/pull/913 released in sqlparser-rs 036.0

Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/926

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -857,7 +857,7 @@ impl<'a> Tokenizer<'a> {
                         Some(sch) if self.dialect.is_identifier_start('%') => {
                             self.tokenize_identifier_or_keyword([ch, *sch], chars)
                         }
-                        _ => Ok(Some(Token::Mod))
+                        _ => Ok(Some(Token::Mod)),
                     }
                 }
                 '|' => {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -424,6 +424,7 @@ struct State<'a> {
 }
 
 impl<'a> State<'a> {
+    /// return the next character and advance the stream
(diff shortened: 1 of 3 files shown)


alamb:
If/when this is merged, I plan to make a 0.36.1 release that includes this fix

src/tokenizer.rs
```diff
@@ -849,13 +851,13 @@ impl<'a> Tokenizer<'a> {
...
                 '%' => {
-                    chars.next();
+                    chars.next(); // advance past '%'
                     match chars.peek() {
-                        Some(' ') => self.consume_and_return(chars, Token::Mod),
+                        Some(' ') => Ok(Some(Token::Mod)),
                         Some(sch) if self.dialect.is_identifier_start('%') => {
                             self.tokenize_identifier_or_keyword([ch, *sch], chars)
                         }
-                        _ => self.consume_and_return(chars, Token::Mod),
```
alamb, line 858: This is the bug fix. At is at this point the `%` has been consumed by `chars.next() so calling `consume_and_return` skips another character by accident

It probably doesn't matter for `' '` but it does matter for everything else in this match arm

alamb:
@izveigor  or @ankrgyl  I wonder if you have a few moments to review this PR? It is a 2 line code change with some tests

crepererum (approved):

alamb:
Thanks for the reviews @izveigor  and @crepererum -- I should point out that I studied the rest of the change carefully and I did not find similar bugs in the handling of `@` or `#`.

PR #930, opened 2023-07-22
Fix parsing of datetime functions without parenthesis

### LINKED ISSUE #918: sqlparser breaks CURRENT_TIMESTAMP for sqlite
In [SQLPage](https://github.com/lovasoa/SQLpage), we parse SQL queries using sqlparser in order to be able to modify them, and then dump them as text to run them on the database. This requires the process of parsing and then stringifying back not to break the semantics nor the syntax of the query.

However, the following SQLite query:

```sql
SELECT CURRENT_TIMESTAMP;
```

when parsed and then dumped back, gives:

```sql
SELECT CURRENT_TIMESTAMP();
```

This breaks the syntax for SQLite, which gives

```
sqlite> SELECT CURRENT_TIMESTAMP();
Error: in prepare, near "(": syntax error (1)
```

lovasoa:
The "special" flag was missing when parsing datetime functions as special functions

Fixes #918

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -957,17 +957,18 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_time_functions(&mut self, name: ObjectName) -> Result<Expr, ParserError> {
-        let (args, order_by) = if self.consume_token(&Token::LParen) {
-            self.parse_optional_args_with_orderby()?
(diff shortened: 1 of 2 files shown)


alamb (approved):
Thank you @lovasoa  -- this looks great

lovasoa:
Thanks @alamb !

PR #935, opened 2023-07-25
Fix "BEGIN TRANSACTION" being serialized as "START TRANSACTION"

### LINKED ISSUE #919: "BEGIN TRANSACTION" gets stringified as "START TRANSACTION"
Parsing the following:

```sql
BEGIN;
```

and then calling `.to_string()` on the parsed statement, results in 

```sql
START TRANSACTION;
```

The `START` syntax in not supported in SQLite. But ideally, the parser would keep track of which variant of the syntax was used, and be able to use it in the `Format` implementation.

lovasoa:
"START TRANSACTION" is not valid in SQLite, which means that valid
SQLite SQL code, when parsed and re-serialized by sqlparser, would become invalid.

Fixes #919

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1541,7 +1541,10 @@ pub enum Statement {
     /// Note: This is a MySQL-specific statement.
     Use { db_name: Ident },
     /// `{ BEGIN [ TRANSACTION | WORK ] | START TRANSACTION } ...`
-    StartTransaction { modes: Vec<TransactionMode> },
+    StartTransaction {
+        modes: Vec<TransactionMode>,
+        syntax_begin: bool,
+    },
     /// `SET TRANSACTION ...`
     SetTransaction {
         modes: Vec<TransactionMode>,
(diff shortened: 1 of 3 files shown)


lovasoa:
@alamb , is that okay for you? I need this for SQLPage.

parkma99:
LGTM @lovasoa Thanks

src/ast/mod.rs
```diff
@@ -1541,7 +1541,10 @@ pub enum Statement {
     /// Note: This is a MySQL-specific statement.
     Use { db_name: Ident },
     /// `{ BEGIN [ TRANSACTION | WORK ] | START TRANSACTION } ...`
-    StartTransaction { modes: Vec<TransactionMode> },
+    StartTransaction {
+        modes: Vec<TransactionMode>,
+        syntax_begin: bool,
```
alamb, line 1546: Could we perhaps add a comment explaining what `syntax_begin` means (so that people don't have to look at the source code?)

Maybe 
```
    /// `START  [ TRANSACTION | WORK ] | START TRANSACTION } ...`
    /// If `syntax_begin` is false.
    ///
    /// `BEGIN  [ TRANSACTION | WORK ] | START TRANSACTION } ...`
    /// If `syntax_begin` is true
````

?

alamb: Also, calling this field `begin` might be more consistent with the rest of this crate but `syntax_begin` is fine as well

alamb (approved):
Looks good to me @lovasoa  -- thank you.

I had one request re documentation and a naming suggestion  but otherwise  this looks great

lovasoa:
thanks @alamb . I implemented your suggestions.

alamb:
Thanks @lovasoa

PR #939, opened 2023-07-31
Fix: use Rust idiomatic capitalization for newly added `DataType` enums

Kikkon:
In my previous commit https://github.com/sqlparser-rs/sqlparser-rs/issues/931 , I think the handling of case for the PostgreSQL type aliases is incorrect. We should modify it to  first letter capitalized


Kikkon:
This change is related to PostgreSQL alias types.  ： https://github.com/apache/arrow-datafusion/pull/7156
Maybe we can merge this, then publish a release version. @alamb

alamb (approved):
Makes sense to me -- thank you @Kikkon

PR #943, opened 2023-08-09
feat: support more Postgres index syntax

ForbesLindesay:
Postgres `CREATE INDEX` statements have the following syntax:

```
CREATE [ UNIQUE ] INDEX [ CONCURRENTLY ] [ [ IF NOT EXISTS ] name ] ON [ ONLY ] table_name [ USING method ]
    ( { column_name | ( expression ) } [ COLLATE collation ] [ opclass [ ( opclass_parameter = value [, ... ] ) ] ] [ ASC | DESC ] [ NULLS { FIRST | LAST } ] [, ...] )
    [ INCLUDE ( column_name [, ...] ) ]
    [ NULLS [ NOT ] DISTINCT ]
    [ WITH ( storage_parameter [= value] [, ... ] ) ]
    [ TABLESPACE tablespace_name ]
    [ WHERE predicate ]
```

> source: https://www.postgresql.org/docs/15/sql-createindex.html

This PR adds support for:

- `CONCURRENTLY` - When this option is used, PostgreSQL will build the index without taking any locks that prevent concurrent inserts, updates, or deletes on the table.
- Anonymous indexes - if you do not have `IF NOT EXISTS`, the index name is optional.
- `INCLUDE ( column_name [, ...] )`- The optional INCLUDE clause specifies a list of columns which will be included in the index as non-key columns.
- `NULLS NOT DISTINCT` - Specifies whether for a unique index, null values should be considered distinct (not equal). The default is that they are distinct, so that a unique index could contain multiple null values in a column.
- `WHERE predicate` - The constraint expression for a partial index.

Questions:

1. For Postgres, the default is `NULLS DISTINCT`, so I made that implicit and did not include it in the generated SQL. An alternative would be to make it `nulls_distinct: Option<bool>` and have it set to `None` by default and `Some(true)`/`Some(false)` for `NULLS DISTINCT`/`NULLS NOT DISTINCT` respectively.
2. Many db engines don't support all these features. Should these be made Postgres specific in some way, or is it ok that you can parse a statement like `CREATE INDEX CONCURRENTLY ON my_table(my_col)` even when using SQLite as your dialect.

P.S. I ran `cargo test && cargo fmt && cargo clippy` and found no issues.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1354,6 +1354,7 @@ pub enum Statement {
         using: Option<Ident>,
         columns: Vec<OrderByExpr>,
         unique: bool,
+        concurrently: bool,
         if_not_exists: bool,
     },
     /// CREATE ROLE
@@ -2464,12 +2465,14 @@ impl fmt::Display for Statement {
                 using,
                 columns,
                 unique,
+                concurrently,
                 if_not_exists,
             } => {
                 write!(
                     f,
-                    "CREATE {unique}INDEX {if_not_exists}{name} ON {table_name}",
+                    "CREATE {unique}INDEX {concurrently}{if_not_exists}{name} ON {table_name}",
                     unique = if *unique { "UNIQUE " } else { "" },
+                    concurrently = if *concurrently { "CONCURRENTLY " } else { "" },
                     if_not_exists = if *if_not_exists { "IF NOT EXISTS " } else { "" },
                     name = name,
                     table_name = table_name
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2493,8 +2493,8 @@ impl fmt::Display for Statement {
                 if !include.is_empty() {
                     write!(f, " INCLUDE ({})", display_separated(include, ","))?;
                 }
-                if *nulls_distinct {
-                    write!(f, " NULLS DISTINCT")?;
+                if !nulls_distinct {
+                    write!(f, " NULLS NOT DISTINCT")?;
                 }
                 if let Some(predicate) = predicate {
                     write!(f, " WHERE {predicate}")?;
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2476,16 +2476,15 @@ impl fmt::Display for Statement {
             } => {
                 write!(
                     f,
-                    "CREATE {unique}INDEX {concurrently}{if_not_exists}{name}ON {table_name}",
+                    "CREATE {unique}INDEX {concurrently}{if_not_exists}",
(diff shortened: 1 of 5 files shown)


alamb (commented):
Thank you for the contribution @ForbesLindesay -- this is looking very close. 

The CI is failing here: https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/5807640477/job/15973997516?pr=943

The issue is that when people use sqlparser in some places like web assembly `format!` is not available. 

> Questions:

> For Postgres, the default is NULLS DISTINCT, so I made that implicit and did not include it in the generated SQL. An alternative would be to make it nulls_distinct: Option<bool> and have it set to None by default and Some(true)/Some(false) for NULLS DISTINCT/NULLS NOT DISTINCT respectively.

The basic philosophy of sqlparser-rs is to:
1. Parse the SQL that is present and don't apply any semantics to it
2. round tripping parsed SQL back to a string should produce the same output as the input (except for whitespace)

It is not entirely consistent in these areas

So for your specific question, I think the alternative you list would adhere the most closely to the rest of the code:

> An alternative would be to make it nulls_distinct: Option<bool> and have it set to None by default and Some(true)/Some(false) for NULLS DISTINCT/NULLS NOT DISTINCT respectively.

> Many db engines don't support all these features. Should these be made Postgres specific in some way, or is it ok that you can parse a statement like CREATE INDEX CONCURRENTLY ON my_table(my_col) even when using SQLite as your dialect.

Ideally you would conditionalize it so that they were supported for "GenericDialect" and "PostgresDialect", but not for any others. Here s an example

https://github.com/sqlparser-rs/sqlparser-rs/blob/8bbb85356ceb50e2d45545da0e8a15da3680353e/src/parser.rs#L1724

However I don't think that is required, only a nice to have

ForbesLindesay:
@alamb Thanks for the feedback.

1. I've removed the call to `format!` so hopefully that fixes the CI failure. I can't seem to run that particular case locally - probably something odd with my local environment as I'm fairly new to Rust and working on apple silicon.
2. I've updated `nulls_distinct` to be `Option<bool>` rather than `bool` so that now `NULLS DISTINCT` and `NULLS NOT DISTINCT` are both preserved if they are present in the original query.
3. If it's ok with you, I'll leave all features enabled regardless of the dialect as I don't think there's any valid SQL that would be parsed incorrectly as a result, and some of these features are supported in some other engines, e.g. SQLite has the same syntax for adding a `predicate` to create a conditional index. Going through each db engine to check what they support seems pretty time consuming.

alamb (approved):
Looks great to me -- thank you again @ForbesLindesay

PR #946, opened 2023-08-14
Add support for table-level comments

ehoeve:
MySQL allows adding comments on table level. This PR add support for this


alamb (approved):
Thank you @ehoeve  -- this is a great contribution. I have added a test for the optional `=`. 

Note that the current construction will have the property that the output SQL will never have the equal even when the input did, as shown in the test

```sql
CREATE TABLE foo (bar INT) COMMENT = 'baz'
```

Will be printed as 
```sql
CREATE TABLE foo (bar INT) COMMENT = 'baz'
```

PR #947, opened 2023-08-14
Fix SUBSTRING from/to argument construction for mssql

jmaness:
PR for https://github.com/sqlparser-rs/sqlparser-rs/issues/914. This modifies the parsing of `SUBSTRING` for dialects like MsSqlServer which doesn't support the `SUBSTRING(expr [FROM start] [FOR len])` syntax.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1223,25 +1223,47 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_substring_expr(&mut self) -> Result<Expr, ParserError> {
-        // PARSE SUBSTRING (EXPR [FROM 1] [FOR 3])
-        self.expect_token(&Token::LParen)?;
-        let expr = self.parse_expr()?;
-        let mut from_expr = None;
-        if self.parse_keyword(Keyword::FROM) || self.consume_token(&Token::Comma) {
-            from_expr = Some(self.parse_expr()?);
-        }
+        if self.dialect.supports_substring_from_for_expr() {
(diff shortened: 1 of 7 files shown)


tests/sqlparser_common.rs
```diff
@@ -5084,19 +5084,45 @@ fn parse_scalar_subqueries() {
 
 #[test]
 fn parse_substring() {
-    one_statement_parses_to("SELECT SUBSTRING('1')", "SELECT SUBSTRING('1')");
+    let from_for_supported_dialects = TestedDialects {
```
alamb, line 5087: 👍  this is very nice

src/parser.rs
```diff
@@ -1223,25 +1223,47 @@ impl<'a> Parser<'a> {
...
-        Ok(Expr::Substring {
-            expr: Box::new(expr),
-            substring_from: from_expr.map(Box::new),
-            substring_for: to_expr.map(Box::new),
-        })
+            Ok(Expr::Substring {
+                expr: Box::new(expr),
+                substring_from: from_expr.map(Box::new),
+                substring_for: to_expr.map(Box::new),
+                special: !self.dialect.supports_substring_from_for_expr(),
```
alamb, line 1245: As a minor nit hard coding this to `false` I think would be clearer for me:

```suggestion
                special: false,
```

Likewise below. 

No changes are needed in this PR -- I'll make a follow on one with the proposed change

alamb (approved):
This is a really nicely done PR @jmaness -- thank you very much for the contribution. The comments, are clear, as is the code, and the tests thorough. 🏆

PR #949, opened 2023-08-17
Break test and coverage test into separate jobs

alamb:
Rationale: coverage takes a non trivial amount of time and does not reuse the build artifacts from normal test. Thus running them in parallel will improve the speed of CI jobs

Before this PR
* the [test job](https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/5890637664/job/15976156114) on stable takes 2m57s

With this PR the total wallclock time is less than 2m:
* the [test job](https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/5893101296/job/15983893926?pr=949) on stable takes 1m7s
* the [coverage job](https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/5893101296/job/15983894837?pr=949) takes: 1m44s

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -91,8 +91,6 @@ jobs:
         use-tool-cache: true
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Test
-      run: cargo test --all-features
     - name: Coverage
       run: cargo tarpaulin -o Lcov --output-dir ./coverage
     - name: Coveralls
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -75,11 +75,27 @@ jobs:
       uses: actions/checkout@v2
     - name: Test
       run: cargo test --all-features
+
```
alamb, line 78: The diff makes is somewhat hard to see, but I simply made a second job that runs coverage

alamb:
Merging this one in to see how it works during the 0.38 dev cycle

PR #950, opened 2023-08-17
Add support for MySQL auto_increment offset

### LINKED ISSUE #920: An exception occurred while parsing the MySQL CREATE TABLE statement. ParserError("Expected end of statement, found: AUTO_INCREMENT")
image

image

ehoeve:
This should also fix #920.

### DIFF
--- a/src/ast/helpers/stmt_create_table.rs
+++ b/src/ast/helpers/stmt_create_table.rs
@@ -66,7 +66,7 @@ pub struct CreateTableBuilder {
     pub clone: Option<ObjectName>,
     pub engine: Option<String>,
     pub comment: Option<String>,
-    pub autoincrement_offset: Option<u32>,
+    pub auto_increment_offset: Option<u32>,
     pub default_charset: Option<String>,
     pub collation: Option<String>,
     pub on_commit: Option<OnCommit>,
@@ -99,7 +99,7 @@ impl CreateTableBuilder {
             clone: None,
             engine: None,
             comment: None,
-            autoincrement_offset: None,
+            auto_increment_offset: None,
             default_charset: None,
             collation: None,
             on_commit: None,
@@ -206,8 +206,8 @@ impl CreateTableBuilder {
         self
     }
 
-    pub fn autoincrement_offset(mut self, offset: Option<u32>) -> Self {
-        self.autoincrement_offset = offset;
+    pub fn auto_increment_offset(mut self, offset: Option<u32>) -> Self {
+        self.auto_increment_offset = offset;
         self
     }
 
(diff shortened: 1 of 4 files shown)


src/ast/helpers/stmt_create_table.rs
```diff
@@ -66,6 +66,7 @@ pub struct CreateTableBuilder {
     pub clone: Option<ObjectName>,
     pub engine: Option<String>,
     pub comment: Option<String>,
+    pub autoincrement_offset: Option<u32>,
```
alamb, line 69: is there any reason to use a different style than the original `auto_increment`? I wonder if you considered:

```suggestion
    pub auto_increment_offset: Option<u32>,
```
?

If so can you explain why you chose this way?

I was thinking that following the exisiting convention would make the relevant code / field easier to search for (aka you can search for `auto_increment` in the code to see how `auto_increment` in `SQL` would be parsed.

src/parser/mod.rs
```diff
@@ -3550,6 +3550,17 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let autoincrement_offset = if self.parse_keyword(Keyword::AUTO_INCREMENT) {
```
alamb, line 3553: https://dev.mysql.com/doc/refman/8.0/en/create-table.html

I got super confused for a moment because `AUTO_INCREMENT` can occur either as part of the table options (as this PR correctly does)

```

table_options:
    table_option [[,] table_option] ...

table_option: {
    AUTOEXTEND_SIZE [=] value
  | AUTO_INCREMENT [=] value
  | AVG_ROW_LENGTH [=] value
...
```

Or also as part of the column definition

```

column_definition: {
    data_type [NOT NULL | NULL] [DEFAULT {literal | (expr)} ]
      [VISIBLE | INVISIBLE]
      [AUTO_INCREMENT] [UNIQUE [KEY]] [[PRIMARY] KEY]
...
```



Which SQLParser already handles
https://github.com/sqlparser-rs/sqlparser-rs/blob/41e47cc0136c705a3335b1504d50ae8b22711b86/src/ast/ddl.rs#L587

alamb (approved):
Nice work, thank you @ehoeve 

My only suggestion is to have the name more closely match the sql syntax, though I may have missed something.

Let me know and I'll merge this PR in!

Thanks again

src/ast/helpers/stmt_create_table.rs
ehoeve: No specific reason, your suggestion of more closely following convention is a better option I think, so I updated the PR

alamb (approved):

alamb:
Thanks @ehoeve

PR #951, opened 2023-08-18
Table time travel clause support, add `visit_table_factor` to Visitor

gruuya:
This PR introduces support for temporal querying in [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#for_system_time_as_of) and [MSSQL](https://learn.microsoft.com/en-us/sql/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table?view=sql-server-ver16) for now. It does so by extending the `TableFactor::Table` variant to record and propagate an optional version specification (atm only a timestamp, though we don't validate them).

Related to https://github.com/apache/arrow-datafusion/issues/7292: a visitor method is also derived for `TableFactor` in order to facilitate walking/parsing of table versions used in a given statement, so that adequate `datafusion::datasource::TableProvider` implementations can be loaded prior to planning/execution.


src/ast/query.rs
```diff
@@ -835,6 +843,24 @@ impl fmt::Display for TableAlias {
...
+impl Display for TableVersion {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            TableVersion::Timestamp(timestamp) => {
+                write!(f, " FOR SYSTEM_TIME AS OF '{timestamp}'")?
+            }
+        }
+        Ok(())
+    }
+}
```
gruuya, line 862: This may be a slight problem—some other systems have completely different syntaxes here but we're missing the info on which particular dialect the table version was parsed for.

For instance, Snowflake uses [`AT`](https://docs.snowflake.com/en/sql-reference/constructs/at-before), so if we decided to extend the support to it we wouldn't know how to re-construct the SQL here.

src/ast/query.rs
```diff
@@ -835,6 +843,24 @@ impl fmt::Display for TableAlias {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum TableVersion {
+    Timestamp(String),
```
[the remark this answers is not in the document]
alamb: I looked at MsSQL and it seems to allow several different phrases like `BETWEEN` and `CONTAINED`

https://learn.microsoft.com/en-us/sql/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table?view=sql-server-ver16

I don't think we need to add support for any of this syntax now in this PR, but your choice of `enum` to represent this structure I think is prescient

src/ast/query.rs
```diff
@@ -661,6 +662,9 @@ pub enum TableFactor {
         args: Option<Vec<FunctionArg>>,
         /// MSSQL-specific `WITH (...)` hints such as NOLOCK.
         with_hints: Vec<Expr>,
+        /// Optional version qualifier to facilitate table time-travel, as
+        /// supported by BigQuery and MSSQL.
+        version: Option<TableVersion>,
```
alamb, line 667: given it is so specific maybe this could be called `system_time: Option<SystemTime>` or something? I don't feel strongly

alamb (commented):
Thank you @gruuya  -- this looks very nice. I had some naming suggestions, and to change the type from String to Expr, but otherwise I think this is very nearly ready to go

src/ast/query.rs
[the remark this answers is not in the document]
gruuya: Good idea, will revise it like that.

src/ast/query.rs
gruuya: One thing I wonder is whether the `AsOf` variant should be named explicitly as `ForSystemAsOf`, given that e.g. CockroachDB also has an [`AS OF` clause](https://www.cockroachlabs.com/docs/stable/as-of-system-time) with a somewhat different syntax. Arguably that dialect is not supported now but if someone decides to implement it it would be a misfortunate conflict.

src/ast/query.rs
gruuya: Ideally, I imagine this field would evolve into something capable of capturing other qualifiers beyond time, e.g. something like explicit version numbers as [supported by delta-rs](https://docs.rs/deltalake/latest/deltalake/builder/enum.DeltaVersion.html). To that end, I'd vote to keep it more generic than `system_time`.

gruuya: Perhaps I could rename it to `temporal_qualifier: Option<TemporalQualifier>,`? Feels more SQL-y, and at the same time sufficiently generic such that it could also store explicit version number references (besides timestamps and intervals).

alamb:
It seems like the docs CI failure is real -- specifically https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/5903878906/job/16077037912?pr=951

It is pointing out the lack of a `visit_table_factors` function, in the same way there is a [`visit_statements`](https://docs.rs/sqlparser/latest/sqlparser/ast/fn.visit_statements.html) function. Did you mean to add `visit_table_factors` too?

gruuya:
> Did you mean to add visit_table_factors too?

No, removed that line from the docs now, thanks.

src/ast/query.rs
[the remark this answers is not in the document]
gruuya: Ok, I've renamed `AsOf` to `ForSystemTimeAsOf`, and made it wrap a generic `Expr` instead of a `String`. Arguably this makes the clause more versatile then is supported by the dialects, but that's probably for the better.

alamb (approved):
Thank you @MartinNowak

PR #952, opened 2023-08-20
mssql table name in square brackets

lovasoa:
- Test mssql table name in square brackets
- test compound identifiers

### DIFF
--- a/tests/sqlparser_mssql.rs
+++ b/tests/sqlparser_mssql.rs
@@ -326,9 +326,9 @@ fn parse_delimited_identifiers() {
 
 #[test]
 fn parse_table_name_in_square_brackets() {
-    let select = ms().verified_only_select(r#"SELECT [a column] FROM [a table]"#);
+    let select = ms().verified_only_select(r#"SELECT [a column] FROM [a schema].[a table]"#);
     if let TableFactor::Table { name, .. } = only(select.from).relation {
-        assert_eq!(vec![Ident::with_quote('[', "a table")], name.0);
+        assert_eq!(vec![Ident::with_quote('[', "a schema"), Ident::with_quote('[', "a table")], name.0);
     } else {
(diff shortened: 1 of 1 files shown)


tests/sqlparser_mssql.rs
```diff
@@ -324,6 +324,26 @@ fn parse_delimited_identifiers() {
     //TODO verified_stmt(r#"UPDATE foo SET "bar" = 5"#);
 }
 
+#[test]
+fn parse_table_name_in_square_brackets() {
+    let select = ms().verified_only_select(r#"SELECT [a column] FROM [a schema].[a table]"#);
+    if let TableFactor::Table { name, .. } = only(select.from).relation {
+        assert_eq!(
+            vec![
+                Ident::with_quote('[', "a schema"),
```
alamb, line 333: that is definitely not what I would have expected (I guess I would have expected a single '[a schema]' or '['. 'a schema, ']'

alamb (approved):
Thank you @lovasoa  -- I think this improves the code by making the tests more explicit about how the parser works.

In your opinion is this behavior ok, or would you prefer it to be better (and you are adding a test to document how it works in case it changes)?

tests/sqlparser_mssql.rs
lovasoa: I wanted to implement the feature, so I started by adding a test, and realized it was already there :smile: SQLPage used to parse everything with the generic parser, so mssql identifiers in square brackets didn't work, but they do work in the mssql dialect.

The current behavior seems to match how microsoft sql server parses the queries itself, so it's alright.

alamb: Working on this project has illustrated to me so many peculiarities of SQL dialects. I feel sometimes like I am becoming a connoisseur 🤪

PR #958, opened 2023-08-27
feat: show location info in parse errors

MartinNowak:
- feat!: tokenize with locations by default
- add consistent Display for Location
- show (optional) token location in parse errors
- keep Location line = 0 as noop

This PR does not attempt to add locations to the AST (#757).
Hence it's backwards compatible except for changing the default ParserError messages.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -44,8 +44,8 @@ pub enum ParserError {
 
 // Use `Parser::expected` instead, if possible
 macro_rules! parser_err {
-    ($MSG:expr) => {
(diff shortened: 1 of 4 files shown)


alamb (commented):
Thank you @MartinNowak  -- this looks like a nice change to me. If we can get the CI to pass I think we should merge it.

❤️

MartinNowak:
> If we can get the CI to pass I think we should merge it.

Fixed @alamb

alamb (approved):
Thanks @MartinNowak  -- this looks great to me. Sorry for the delay in review I was away last week

PR #959, opened 2023-08-29
Support mysql `partition` to table selection

chunshao90:
Support mysql partition selection, example:
```
SELECT * FROM employees PARTITION (p0, p2)
```
Refer to https://dev.mysql.com/doc/refman/8.0/en/partitioning-selection.html


chunshao90:
@alamb PTAL

alamb:
Sorry for the delay -- I was away last week. I am looking at this PR now

src/ast/query.rs
```diff
@@ -665,6 +665,8 @@ pub enum TableFactor {
         /// Optional version qualifier to facilitate table time-travel, as
         /// supported by BigQuery and MSSQL.
         version: Option<TableVersion>,
+        /// Partition selection, supported by MySQL.
```
alamb, line 668:
```suggestion
        /// [Partition selection](https://dev.mysql.com/doc/refman/8.0/en/partitioning-selection.html), supported by MySQL.
```

src/parser/mod.rs
```diff
@@ -6210,6 +6210,20 @@ impl<'a> Parser<'a> {
         } else {
             let name = self.parse_object_name()?;
 
+            let partitions = if dialect_of!(self is MySqlDialect)
```
alamb, line 6213: Generally this crate also allows all syntax for the `GenericDialect`, such as

https://github.com/sqlparser-rs/sqlparser-rs/blob/4c3a4ad5a87ef8b4dd24a6e8540243e14fbfa148/src/parser/mod.rs#L756

src/parser/mod.rs
```diff
@@ -6210,6 +6210,20 @@ impl<'a> Parser<'a> {
...
+                && self.parse_keyword(Keyword::PARTITION)
+            {
+                let mut partitions = self.parse_comma_separated(|p| p.parse_tuple(true, false))?;
```
alamb, line 6216: As written I think this only checks the first partition for having more than one tuple element (what about values after index 0?)

I think if you used `parse_expr` here this code would be significantly simpler and there would be no need for error checking

```suggestion
                let mut partitions = self.parse_comma_separated(Parser::parse_expr)?;
```

src/ast/query.rs
```diff
@@ -665,6 +665,8 @@ pub enum TableFactor {
...
+        partitions: Vec<Expr>,
```
alamb, line 669: I wonder why this is an `Expr` ? It seems like each element is a `partition_name` 

https://dev.mysql.com/doc/refman/8.0/en/create-table.html

So perhaps this should be something like 

```suggestion
        partitions: Vec<Ident>,
```

And then call https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_identifier to parse them below?

alamb (commented):
Thank you for the contribution @chunshao90  -- this is looking  pretty close. I left some suggestions for your consideration

src/parser/mod.rs
chunshao90: Returns an error if there are multiple tuples.

src/ast/query.rs
chunshao90: done

alamb (approved):
Looks good to me -- thank you @chunshao90

alamb:
I'll clean up the clippy failures

alamb:
Thanks again @chunshao90 !

PR #960, opened 2023-08-29
feat: allow multiple actions in one `ALTER TABLE` statement

### LINKED ISSUE #675: adding multiple columns in a single `alter table` statement doesn't work
When I try to parse the following valid SQL statement (it runs fine on my MySQL server) via `sqlparser`:

```
alter table sometable add column b integer, add column c integer;
```

Instead of successfully parsing it, I get the following error:

```
Expected end of statement, found: ,
```

It seems like `sqlparser` doesn't understand having multiple additions in a single `alter table` statement. Is it possible to make this work?

ForbesLindesay:
Fixes #675

I primarily focused on ensuring this follows the Postgres syntax: https://www.postgresql.org/docs/15/sql-altertable.html

This should largely work for MySql too though: https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

One possible thing that could be worth considering is that MySql supports a list of "partition_options" after the "alter_options" and while the "alter_options" are comma-separated just like Postgres, the "partition_options" are separated by a space. I'm not sure how best to cleanly represent this in our current AST.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4037,6 +4037,174 @@ impl<'a> Parser<'a> {
         Ok(SqlOption { name, value })
     }
 
+    pub fn parse_alter_table_operation(&mut self) -> Result<AlterTableOperation, ParserError> {
+        let operation = if self.parse_keyword(Keyword::ADD) {
+            if let Some(constraint) = self.parse_optional_table_constraint()? {
+                AlterTableOperation::AddConstraint(constraint)
+            } else {
+                let if_not_exists =
+                    self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
+                if self.parse_keyword(Keyword::PARTITION) {
+                    self.expect_token(&Token::LParen)?;
+                    let partitions = self.parse_comma_separated(Parser::parse_expr)?;
+                    self.expect_token(&Token::RParen)?;
+                    AlterTableOperation::AddPartitions {
+                        if_not_exists,
+                        new_partitions: partitions,
+                    }
+                } else {
+                    let column_keyword = self.parse_keyword(Keyword::COLUMN);
+
+                    let if_not_exists = if dialect_of!(self is PostgreSqlDialect | BigQueryDialect | DuckDbDialect | GenericDialect)
+                    {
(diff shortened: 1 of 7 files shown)


alamb:
> the "partition_options" are separated by a space. I'm not sure how best to cleanly represent this in our current AST

Could they be represented by a `Vec<Partition>` or something similar?

Interestingly https://github.com/sqlparser-rs/sqlparser-rs/pull/959 is also related to mysql partition syntax

tests/sqlparser_common.rs
```diff
@@ -2915,48 +2915,37 @@ fn parse_create_external_table_lowercase() {
...
             assert!(!if_not_exists);
-            assert_eq!("tab", name.to_string());
             assert_eq!("foo", column_def.name.to_string());
             assert_eq!("TEXT", column_def.data_type.to_string());
         }
         _ => unreachable!(),
     };
 
     let rename_table = "ALTER TABLE tab RENAME TO new_tab";
-    match verified_stmt(rename_table) {
```
alamb, line 2938: This is a nice refactoring of the tests -- thank you

tests/sqlparser_postgres.rs
```diff
@@ -588,6 +584,48 @@ fn parse_alter_table_alter_column() {
     }
 }
 
+#[test]
+fn parse_alter_table_add_columns() {
```
alamb, line 588: 👍

src/parser/mod.rs
```diff
@@ -4047,177 +4215,15 @@ impl<'a> Parser<'a> {
...
-                    self.expect_keyword(Keyword::WITH)?;
-                    let table_name = self.parse_object_name()?;
-                    AlterTableOperation::SwapWith { table_name }
-                } else {
-                    return self.expected(
-                        "ADD, RENAME, PARTITION, SWAP or DROP after ALTER TABLE",
-                        self.peek_token(),
-                    );
-                };
+                let operations = self.parse_comma_separated(Parser::parse_alter_table_operation)?;
```
alamb, line 4221: ❤️

alamb (approved):
This looks great -- thank you @ForbesLindesay  🦾 

Sorry for the delay in reviewing -- I was away last week.

PR #963, opened 2023-09-03
`ANY` and `ALL` contains their operators 

SeanTroyUWO:
Changes `Expr::AllOp` and `Expr::AnyOp` to contain the operator which precedes them, rather than be contained by them. This fixes an issue I had as a polars user:  https://github.com/pola-rs/polars/issues/10081.

The rational for this is that despite the order the syntax demands, the conditional operator modifies the behavior of the `ANY` command, and not vice versa.  Similar to `NOT` as with `NOT IN`, the negation modifies the behavior of the `IN`, and the `IN` is the real operation.  In both cases, its more convenient to parse the real operation than the modifier(s).

I did not find any uses the the old `ANY` or `ALL` commands in any of DataFusion, LocustDB, Ballista, GlueSQL or Opteryx. This might still briefly break their builds however. 

If more unit tests are required, please let me know. 

Fixes: https://github.com/sqlparser-rs/sqlparser-rs/issues/936

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -419,10 +419,18 @@ pub enum Expr {
         pattern: Box<Expr>,
         escape_char: Option<char>,
     },
-    /// Any operation e.g. `1 ANY (1)` or `foo > ANY(bar)`, It will be wrapped in the right side of BinaryExpr
-    AnyOp(Box<Expr>),
-    /// ALL operation e.g. `1 ALL (1)` or `foo > ALL(bar)`, It will be wrapped in the right side of BinaryExpr
-    AllOp(Box<Expr>),
+    /// Any operation e.g. `foo > ANY(bar)`, comparison operator is one of [=, >, <, =>, =<, !=]
+    AnyOp{
+        left: Box<Expr>,
+        compare_op: BinaryOperator,
+        right: Box<Expr>
+    },
+    /// ALL operation e.g. `foo > ALL(bar)`, comparison operator is one of [=, >, <, =>, =<, !=]
+    AllOp{
+        left: Box<Expr>,
+        compare_op: BinaryOperator,
+        right: Box<Expr>
+    },
     /// Unary operation e.g. `NOT foo`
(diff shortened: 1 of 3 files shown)


alamb (approved):
Thank you very much for the contribution @SeanTroyUWO -- I apologize for the delay in reviewing. 

This change also appears to be consistent with postgres any mysql (where the ANY/ALL is part of the comparison expression):

https://www.postgresql.org/docs/current/functions-comparisons.html

> 9.24.3. ANY/SOME (array)
> expression operator ANY (array expression)
> expression operator SOME (array expression)

https://dev.mysql.com/doc/refman/8.0/en/any-in-some-subqueries.html

> The ANY keyword, which must follow a comparison operator, ...

Another alternative might be to add it directly to [`Expr::BinaryOp`](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Expr.html#variant.BinaryOp) but I think that would be far more disruptive to downstream users so this looks good to me

SeanTroyUWO:
> Thank you very much for the contribution @SeanTroyUWO -- I apologize for the delay in reviewing.
> 
> This change also appears to be consistent with postgres any mysql (where the ANY/ALL is part of the comparison expression):
> 
> https://www.postgresql.org/docs/current/functions-comparisons.html
> 
> > 9.24.3. ANY/SOME (array)
> > expression operator ANY (array expression)
> > expression operator SOME (array expression)
> 
> https://dev.mysql.com/doc/refman/8.0/en/any-in-some-subqueries.html
> 
> > The ANY keyword, which must follow a comparison operator, ...
> 
> Another alternative might be to add it directly to [`Expr::BinaryOp`](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Expr.html#variant.BinaryOp) but I think that would be far more disruptive to downstream users so this looks good to me

Great, I'll look forward to the next release!

alamb:
Next release is tracked by https://github.com/sqlparser-rs/sqlparser-rs/issues/955

ETA in about 2 weeks

PR #965, opened 2023-09-08
Add missing token loc in parse err msg

ding-young:


alamb (approved):
Thank you very much @ding-young  -- this was a logical conflict introduced yesterday. Very much appreciated. 🙏

PR #966, opened 2023-09-10
Added ability to parse STRUCT and MAP fields as well as nested arrays

scarman-db:
This PR has additions for parsing table columns with fields such as 

```sql
CREATE TABLE t (
  m MAP<STRING, INT>,
  s STRUCT<field1: INT, field2: STRING, field3: INT>,
  a ARRAY<ARRAY<ARRAY<INT>>>
)
```

Additionally, this has the ability to parse back for some SQL dialects that use `[]` instead of the `ARRAY<T>` syntax adding the "SquareBracketedArray` instead of just `Array`

I also have a potential solution to the nested bracket issue, but I have no added it to Maps and structs although it should just work (I have to write tests a well) but I more was not happy with the manner in which I implemented this and I am looking for some suggestions on how to better structure this.


alamb:
Thank you @scarman-db  -- I'll check this out in the next few days

src/ast/data_type.rs
```diff
@@ -198,10 +199,15 @@ pub enum DataType {
     Custom(ObjectName, Vec<String>),
     /// Arrays
     Array(Option<Box<DataType>>),
+    BracketArray(Option<Box<DataType>>),
     /// Enums
     Enum(Vec<String>),
     /// Set
     Set(Vec<String>),
+    /// Map
```
alamb, line 207: Can you please provide a link to what SQL dialect supports this syntax?

src/ast/ddl.rs
```diff
@@ -678,6 +678,25 @@ impl fmt::Display for ColumnOption {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct StructField {
```
alamb, line 684: Can you please add docstrings that explain what this struct is for (with an example SQL snippet)

src/parser/mod.rs
```diff
@@ -257,6 +259,7 @@ pub struct Parser<'a> {
     options: ParserOptions,
     /// ensure the stack does not overflow by limiting recursion depth
     recursion_counter: RecursionCounter,
+    max_depth: usize,
```
alamb, line 262: Please document what this field is foe.

Also, could you please explain why you didn't use the counter in `RecusionCounter`?

alamb:
Marking as draft as this PR is  no longer waiting on feedback

alamb:
Similar PR here https://github.com/sqlparser-rs/sqlparser-rs/pull/1003

scarman-db:
I wanted feedback on the approach, but I looked at the other PR and it seems a bit more well thought out than mine. I'd suggest we can close this PR, but also for the other PR to add a test for the nesting arrays structs.

alamb:
Thank you @scarman-db  -- sorry for the delay in reviewing this PR

PR #493, opened 2022-05-10
Support `UNNEST` as a table factor

sivchari:
This PR adds support `UNNEST` for BigQuery.

```
from_item:
    {
      table_name [ as_alias ] [ FOR SYSTEM_TIME AS OF timestamp_expression ]
      | { join_operation | ( join_operation ) }
      | ( query_expr ) [ as_alias ]
      | field_path
      | unnest_operator
      | cte_name [ as_alias ]
    }

unnest_operator:
    {
      [UNNEST](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unnest)( array_expression )
      | UNNEST( array_path )
      | array_path
    }
    [ as_alias ]
    [ WITH OFFSET [ as_alias ] ]
```

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unnest_operator

PR #968, opened 2023-09-13
Support `UNNEST` as a table factor for PostgreSQL

hexedpackets:
This PR adds support UNNEST for PostgreSQL (originally added in #493, but only for BigQuery and Generic dialects).

https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS


alamb (approved):
Thank you for the contribution @hexedpackets 🙏

PR #969, opened 2023-09-17
parse SQLite pragma statement

marhoily:
Adds support for SQLite PRAGMA statement:

`PRAGMA cache_size = 10`

According to the [docs](https://www.sqlite.org/pragma.html) this is not all the possible syntax variations. If someone reviews my changes, I could add more before merging.


tests/sqlparser_sqlite.rs
```diff
@@ -24,6 +24,51 @@ use sqlparser::ast::*;
 use sqlparser::dialect::{GenericDialect, SQLiteDialect};
 use sqlparser::tokenizer::Token;
 
+#[test]
+fn pragma_no_value() {
```
alamb, line 28: Can you please also add a test for the `[schema-name '.'] pragma-name` variant as well? 

like `PRAGMA foo.cache_size = 5`?

alamb (approved):
Thank you for the contribution @marhoily -- ideally this PR would have one more test, but otherwise it looks great. 

Thanks again

alamb:
I took the liberty of merging this PR up from main and running `cargo fmt` on it to get a clean CI run

PR #971, opened 2023-09-20
Add support for mixed BigQuery table name quoting

### LINKED ISSUE #889: BigQuery ignores identifiers in FROM clause
tl;dr - SELECT * from d.s.t and SELECT * from \`d.s.t\` are equivalent in BigQuery and should return the same response from sqlparser. 

Let's say I have database d, with schema s and table t.

In Snowflake, the following query will work - SELECT * FROM d.s.t 
However, the following query will fail - SELECT * from "d.s.t" as Snowflake will look for a table called d.s.t

However, in BigQuery:
SELECT * from d.s.t and SELECT * from \`d.s.t\` are equivalent and both work. I agree it's not expected and confusing.

I really want to support this in someway in sqlparser: If the Dialect is BigQuery, ignore the identifiers in the FROM clause. 
(I'm getting the wrong result for SELECT * from \`d.s.t\`, the table is with one value instead of 3).

As I'm not a Rust developer, any guidance with this one? Every manipulation i'm trying to do is already after the words are tokenised, so I don't have a way to ignore the identifier. 

Appreciate your guidance here!

Thank you!

iffyio:
As described in the gh issue, BigQuery accepts arbitrary quoting identifiers within table names - such that e.g: `'a.b.c'` == `'a'.'b'.'c'` == `'a.b'.'c'` are all equivalent.

This adds support for such syntax

Fixes #889

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5024,6 +5024,27 @@ impl<'a> Parser<'a> {
                 break;
             }
         }
+
+        // BigQuery accepts any number of quoted identifiers of a table name.
+        // https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_identifiers
+        if dialect_of!(self is BigQueryDialect)
+            && idents.iter().any(|ident| ident.value.contains('.'))
+        {
+            idents = idents
+                .into_iter()
+                .flat_map(|ident| {
+                    ident
+                        .value
+                        .split('.')
+                        .map(|value| Ident {
+                            value: value.into(),
+                            quote_style: ident.quote_style,
+                        })
+                        .collect::<Vec<_>>()
+                })
+                .collect()
+        }
+
         Ok(ObjectName(idents))
     }
 
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -13,6 +13,8 @@
 #[macro_use]
 mod test_utils;
 
+use std::ops::Deref;
+
 use sqlparser::ast::*;
 use sqlparser::dialect::{BigQueryDialect, GenericDialect};
 use test_utils::*;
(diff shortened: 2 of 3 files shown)


src/parser/mod.rs
```diff
@@ -5024,6 +5024,27 @@ impl<'a> Parser<'a> {
...
+        // https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_identifiers
+        if dialect_of!(self is BigQueryDialect)
+            && idents.iter().any(|ident| ident.value.contains('.'))
+        {
+            idents = idents
+                .into_iter()
+                .flat_map(|ident| {
+                    ident
+                        .value
+                        .split('.')
```
alamb, line 5038: does BigQuery support any way to escape `.` (like `'foo\.bar'` I wonder 🤔 )

tests/sqlparser_bigquery.rs
```diff
@@ -84,9 +86,16 @@ fn parse_raw_literal() {
 
 #[test]
 fn parse_table_identifiers() {
-    fn test_table_ident(ident: &str, expected: Vec<Ident>) {
+    fn test_table_ident(ident: &str, canonical: Option<&str>, expected: Vec<Ident>) {
```
alamb, line 89: can you please document what `canonical` represents in this context?

alamb (commented):
Thank you @iffyio

alamb:
I wonder if you could respond to @MartinNowak 's comment https://github.com/sqlparser-rs/sqlparser-rs/issues/889#issuecomment-1563524163 as well? Aka does this change really need to be in sqlparser-rs?

src/parser/mod.rs
iffyio: BigQuery didn't support escaping dots in identifiers either as far as I could tell and tested

iffyio:
> I wonder if you could respond to @MartinNowak 's comment [#889 (comment)](https://github.com/sqlparser-rs/sqlparser-rs/issues/889#issuecomment-1563524163) as well? Aka does this change really need to be in sqlparser-rs?

I'm not entirely sure I follow, the changes made in this MR were based off my understanding of what was being proposed in that comment, unless I misunderstood - maybe @MartinNowak can clarify otherwise

It wouldn't seem reasonable to do this outside the parser - without it, all users of the BigQuery dialect essentially get back wrong object names from the parser in such cases and will need to know to check all identifiers they get back and fix it up with code similar to the MR - effectively re-parsing the object name themself.

tests/sqlparser_bigquery.rs
iffyio: Added!

alamb (approved):
Thank you @iffyio

iffyio:
Thanks! :)

PR #977, opened 2023-09-24
clickhouse: add support for LIMIT BY

lustefaniak:
https://clickhouse.com/docs/en/sql-reference/statements/select/limit-by


src/parser/mod.rs
```diff
@@ -5434,6 +5436,13 @@ impl<'a> Parser<'a> {
                 }
             }
 
+            let limit_by =
+                if dialect_of!(self is ClickHouseDialect) && self.parse_keyword(Keyword::BY) {
```
alamb, line 5440: in general I think we try to have `GenericDialect` support the union of all sql dialects, if possible. So this might be something like

```suggestion
                if dialect_of!(self is ClickHouseDialect | GenericDialect) && self.parse_keyword(Keyword::BY) {
```

tests/sqlparser_clickhouse.rs
```diff
@@ -336,6 +336,13 @@ fn parse_create_table() {
     );
 }
 
+#[test]
+fn parse_limit_by() {
```
alamb, line 340: Can you also please add a test for the multi-column form (e.g. `LIMIT BY asset, class` or similar`)?

alamb (commented):
Thank you (again) @lustefaniak  -- looks very nice.

tests/sqlparser_clickhouse.rs
lustefaniak: Added

alamb (approved):
Thank you @lustefaniak . This looks really nice

PR #978, opened 2023-09-24
Support bigquery `CAST AS x [STRING|DATE] FORMAT` syntax

lustefaniak:
https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_functions#cast_as_date

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -322,6 +322,16 @@ impl fmt::Display for JsonOperator {
     }
 }
 
+/// Options for `CAST` / `TRY_CAST`
+/// BigQuery: <https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax>
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CastFormat {
+    Value(Value),
+    ValueAtTimeZone(Value, Value),
+}
+
 /// An SQL expression of any type.
 ///
 /// The parser does not distinguish between expressions of different types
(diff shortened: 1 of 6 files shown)


src/ast/mod.rs
```diff
@@ -437,19 +437,22 @@ pub enum Expr {
     Cast {
         expr: Box<Expr>,
         data_type: DataType,
+        format: Option<Value>,
```
alamb, line 440: can you please update the documentation here to explain what `format` is (aka what SQL it comes from)?

src/ast/mod.rs
```diff
@@ -437,19 +437,22 @@ pub enum Expr {
...
     },
     /// TRY_CAST an expression to a different data type e.g. `TRY_CAST(foo AS VARCHAR(123))`
     //  this differs from CAST in the choice of how to implement invalid conversions
     TryCast {
         expr: Box<Expr>,
         data_type: DataType,
+        format: Option<Value>,
```
alamb, line 447: FWIW it appears bigquery supports a more sophisticated format clause than just a single value:
https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax

Perhaps we should create a `CastFormat` enum so if we want to extend the supported syntax the API change is less. Something like:

```suggestion
        format: Option<CastFormat>,
```

```rust
/// Options for `CAST` / `TRY_CAST` 
/// BigQuery: https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
enum CastFormat {
   /// Constant string value
  String(Value)
 }
```

alamb (commented):
Thank you @lustefaniak  -- sorry I missed this one in the sea of reiviews. I think it is looking good in my mind 👍

src/ast/mod.rs
alamb: (I don't feel strongly about this -- we can merge it as is as well)

alamb:
I think the CI failures are not related to the changes in this PR -- if you merge up / rebase from main they should be solved. See https://github.com/sqlparser-rs/sqlparser-rs/pull/995

src/ast/mod.rs
lustefaniak: Changed

src/ast/mod.rs
lustefaniak: I think this is really good idea, I added CastFormat and covered `AT TIME ZONE` support too.

alamb (approved):
Looks great -- thank you @lustefaniak

PR #981, opened 2023-09-24
Add test for clickhouse: tokenize `==` as Token::DoubleEq

lustefaniak:
I see that the tokenizer changes were applied for Sqlite, but that PR adds additional tests.


alamb (approved):
Thank you for this @lustefaniak  -- much appreciated

PR #982, opened 2023-09-24
clickhouse: add support for parsing ARRAY JOIN

lustefaniak:
https://clickhouse.com/docs/en/sql-reference/statements/select/array-join

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6040,6 +6040,12 @@ impl<'a> Parser<'a> {
                     relation: self.parse_table_factor()?,
                     join_operator: JoinOperator::OuterApply,
                 }
+            } else if self.parse_keyword(Keyword::ARRAY) {
+                self.expect_keyword(Keyword::JOIN)?;
+                Join {
+                    relation: self.parse_array_join_table_factor()?,
+                    join_operator: JoinOperator::Array,
+                }
             } else {
                 let natural = self.parse_keyword(Keyword::NATURAL);
                 let peek_keyword = if let Token::Word(w) = self.peek_token().token {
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -6338,6 +6345,15 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_array_join_table_factor(&mut self) -> Result<TableFactor, ParserError> {
+        {
+            let expr = self.parse_expr()?;
+
+            let alias = self.parse_optional_table_alias(keywords::RESERVED_FOR_TABLE_ALIAS)?;
```
alamb, line 6352: I am not sure about ARRAY JOIN always being a "field accessor" -- would't it makes sense to parse the argument as [TableFactor](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.TableFactor.html#) and that way it can support being a table / subquery / etc rather than just an expression? That would also handle aliases as well

alamb (commented):
Thank you for the contribution @lustefaniak -- I had a comment about this one

src/parser/mod.rs
lustefaniak: I don't fully understand what TableFactor is or its purpose in the lib 🤔 I will have to give it more read.
I checked Clickhouse's latest docs and there is more syntax I need to support e.g.

```sql
SELECT s, `nest.x`, `nest.y`
FROM nested_test
ARRAY JOIN `nest.x`, `nest.y`;
```

From what I could tell, this behaves much like UNNEST, and there are important limitations to what could be inside such `ARRAY JOIN` hence I'm not so sure if making it generic TableFactor actually makes sense 🤔

alamb: I think of `TableFactor` as basically anything that could be JOINEd (aka show up as a element in the FROM clause)

alamb:
This one appears to be stalled. Marking as Draft until we sort out what is needed

PR #983, opened 2023-09-25
snowflake: support for UNPIVOT and a fix for chained PIVOTs

jmhain:
This adds support for `UNPIVOT`, and also makes a change to the AST for `PIVOT` to represent more cases, such as pivoting a subquery or the result of another `PIVOT` / `UNPIVOT` (SF allows this, BQ does not). I implemented the latter case here but not the former - planning to do that in a follow-up.


src/ast/query.rs
```diff
@@ -713,13 +713,28 @@ pub enum TableFactor {
     /// For example `FROM monthly_sales PIVOT(sum(amount) FOR MONTH IN ('JAN', 'FEB'))`
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/pivot>
     Pivot {
-        #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
-        name: ObjectName,
-        table_alias: Option<TableAlias>,
+        #[cfg_attr(feature = "visitor", visit(with = "visit_table_factor"))]
+        table: Box<TableFactor>,
```
alamb, line 717: I thought about suggesting this field be named `relation` but calling it `table` is consistent with other AST nodes (like `Merge`) so I think this is good

src/ast/query.rs
```diff
@@ -803,32 +818,48 @@ impl fmt::Display for TableFactor {
...
                 )?;
-                for value in pivot_values {
+                for (idx, value) in pivot_values.iter().enumerate() {
+                    if idx > 0 {
+                        write!(f, ", ")?;
+                    }
                     write!(f, "{}", value)?;
-                    if !value.eq(pivot_values.last().unwrap()) {
+                }
+                write!(f, "))")?;
```
alamb, line 840: You can `display_comma_separated` to write this more succinctly (the same comment applies below as well): 

```suggestion
                write!(
                    f,
                    "{} PIVOT({} FOR {} IN ({})",
                    table,
                    aggregate_function,
                    Expr::CompoundIdentifier(value_column.to_vec()),
                    display_comma_separated(pivot_values)
                )?;
```

alamb (approved):
Thank you for the contribution @jmhain -- this looks good to me. I had one suggestion on how to make the code clearer, but we can do that as a follow on PR if you prefer.

src/ast/query.rs
jmhain: Updated both places

alamb (approved):
Thank you @jmhain 🙏

PR #990, opened 2023-09-28
Add JumpWire to users in README

hexedpackets:
Adds a reference to JumpWire as a user of the library.


alamb (approved):
Thanks @hexedpackets !

PR #992, opened 2023-09-30
Support DELETE with ORDER BY and LIMIT (MySQL)

ulrichsg:
For reference: https://dev.mysql.com/doc/refman/5.7/en/delete.html


alamb (approved):
Thank you for the contribution @ulrichsg  👍

PR #995, opened 2023-10-05
Fix for clippy 1.73

alamb:
[https://github.com/rust-lang/rust/blob/master/RELEASES.md was](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1730-2023-10-05) was released today and clippy learned some new tricks


alamb:
Merging this in to get clean CI runs

PR #996, opened 2023-10-06
Support subquery as function arg w/o parens in Snowflake dialect

jmhain:
This adds support for a Snowflake quirk where the enclosing parentheses around a subquery can be omitted if it's the only argument to a function.


alamb (approved):
Thank you @jmhain

PR #997, opened 2023-10-08
Add support for UNION DISTINCT BY NAME syntax

alexander-beedie:
I was just enhancing Polars' SQL support for `UNION` ops and noticed the parser doesn't recognise `UNION DISTINCT BY NAME`; the presence of this (optional) additional keyword is supported by DuckDB (and some other dialects, such as U-SQL[^1]). 

This PR adds a new `DistinctByName` entry to `SetQuantifier` to account for this, and streamlines the related test to avoid some copy/paste (while adding suitable coverage for the new quantifier) 👍 

## Example

* Setup some test data tables/frames:
  ```python
  import polars as pl
  
  df1 = pl.DataFrame({
      "A": [1, 2, 3, 4, 5],
      "B": [5, 4, 3, 2, 1],
  })
  df2 = df1.select("B","A")
  ```

* Execute SQL using Polars (before this PR):
  ```python
  pl.SQLContext( register_globals=True ).execute(
      "SELECT * FROM df1 UNION DISTINCT BY NAME SELECT * FROM df2"
  )
  # sql parser error: 
  #  Expected SELECT, VALUES, or a subquery in the query body, found: BY
  ```
  
* Execute same statement using DuckDB:
  _(will pick up the frames from locals)_
  ```python
  import duckdb
  duckdb.sql(
      "SELECT * FROM df1 UNION DISTINCT BY NAME SELECT * FROM df2"
  ).show()
  # ┌───────┬───────┐
  # │   A   │   B   │
  # │ int64 │ int64 │
  # ├───────┼───────┤
  # │     1 │     5 │
  # │     2 │     4 │
  # │     3 │     3 │
  # │     4 │     2 │
  # │     5 │     1 │
  # └───────┴───────┘
  ```

---

Thanks for all the fantastic work on this library  :)

[^1]: U-SQL "unions" reference: https://learn.microsoft.com/en-us/u-sql/statements-and-expressions/set-rowset/union-and-outer-union-expression


alamb:
I took the liberty of merging up from main and fixing a clippy lint.

PR #998, opened 2023-10-11
Support `IGNORE|RESPECT` NULLs clause in window functions

yuval-illumex:
Adding Support for some Rank function of Snowflake:
[FIRST_VALUE](https://docs.snowflake.com/en/sql-reference/functions/first_value), [LAG](https://docs.snowflake.com/en/sql-reference/functions/lag), [LAST_VALUE](https://docs.snowflake.com/en/sql-reference/functions/last_value),[LEAD](https://docs.snowflake.com/en/sql-reference/functions/lead)


tests/sqlparser_snowflake.rs
```diff
@@ -1064,3 +1064,40 @@ fn test_snowflake_trim() {
...
+#[test]
+fn test_snowflake_rank() {
+    let real_sql = "SELECT col_1, col_2, LAG(col_2) IGNORE NULLS OVER (ORDER BY col_1) FROM t1";
+    assert_eq!(snowflake().verified_stmt(real_sql).to_string(), real_sql);
+
+    let first_sql = "SELECT column1, column2, FIRST_VALUE(column2) OVER (PARTITION BY column1 ORDER BY column2 NULLS LAST) AS column2_first FROM t1";
+    assert_eq!(snowflake().verified_stmt(first_sql).to_string(), first_sql);
+
+    let sql_only_select = "SELECT LAG(col_2,1,0) OVER (PARTITION BY col_3 ORDER BY col_1)";
+    let select = snowflake().verified_only_select(sql_only_select);
```
iffyio, line 1077: It seems this rather extends support for window function options Ignore/Respect NULLs vs snowflake specific, since the options are standard sql (e.g the test `first_sql` and `sql_select_only` should already be parsed without the MR)? If so it might be reasonable to instead extend the `parse_function()` to optionally accept an option where applicable like the following?
```
enum WindowFunctionOption {
	RespectNulls,
	IgnoreNulls
}
```
Since parse_function already contains most of the functionality being added, and prevents window functions from being parsed into potentially different types in the AST which might not be desirable.

yuval-illumex: Thank you @iffyio !
I extracted the null clause to an enum, and moved it to `parse_function`.
I will appreciate any feedback.

Thanks!

iffyio: Oh nice!

One thing that stood out was the mismatch between the field name and the typename:
```rust
nulls_clause: Option<WindowFunctionOption>
```
was curious and took a quick look and it turns out the option is mostly called [null treatment clause](https://www.google.com/search?q=window+function+null+treatment+clause) - maybe we can rename accordingly and so they're consistent?

```
null_treatment: Option<NullTreatment>
```

Other than that I think the changes look good!

yuval-illumex: Thanks @iffyio ! I looked for a better name but didn't found one, you did it! 
Renamed :)

test.sql
```diff
@@ -0,0 +1,8 @@
+SELECT
```
alamb, line 1: What is this file for?

src/ast/mod.rs
```diff
@@ -3650,6 +3667,7 @@ impl fmt::Display for CloseCursor {
 pub struct Function {
     pub name: ObjectName,
     pub args: Vec<FunctionArg>,
+    pub null_treatment: Option<NullTreatment>,
```
alamb, line 3670: Can you please add a docstring explaining what `NullTreatment` is?

src/ast/mod.rs
```diff
@@ -1078,6 +1078,23 @@ impl fmt::Display for WindowFrameUnits {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
```
alamb, line 1081: Please add docstrings to this enum

tests/sqlparser_common.rs
```diff
@@ -2259,18 +2263,34 @@ fn parse_agg_with_order_by() {
             Box::new(MsSqlDialect {}),
             Box::new(AnsiDialect {}),
             Box::new(HiveDialect {}),
+            Box::new(SnowflakeDialect {}),
         ],
         options: None,
     };
 
     for sql in [
```
alamb, line 2271: Why were these tests changed?

This PR adds support for window functions (aka those that have an `OVER` clause) which is different than the `ORDER BY` clause in the aggregate function argument.

I think you should keep the existing tests and add a new one for the IGNORE NULLS / RESPECT NULLs syntax

tests/sqlparser_snowflake.rs
```diff
@@ -1064,3 +1065,25 @@ fn test_snowflake_trim() {
...
+        "SELECT column1, column2, FIRST_VALUE(column2) OVER (PARTITION BY column1 ORDER BY column2 NULLS LAST) AS column2_first FROM t1",
+        "SELECT column1, column2, FIRST_VALUE(column2) OVER (ORDER BY column2 NULLS LAST) AS column2_first FROM t1",
+        "SELECT col_1, col_2, LAG(col_2) OVER (ORDER BY col_1) FROM t1",
+        "SELECT LAG(col_2, 1, 0) OVER (ORDER BY col_1) FROM t1",
+        "SELECT LAG(col_2, 1, 0) OVER (PARTITION BY col_3 ORDER BY col_1)",
+    ] {
+        snowflake().verified_stmt(sql);
+    }
+
+    for sql in [
```
alamb, line 1081: Why are there two lists and not just one 🤔 ?

tests/sqlparser_snowflake.rs
```diff
@@ -1064,3 +1065,25 @@ fn test_snowflake_trim() {
         snowflake().parse_sql_statements(error_sql).unwrap_err()
     );
 }
+
+#[test]
+fn parse_agg_with_order_by() {
```
alamb, line 1070: This is testing window functions syntax, so perhaps a better name would be

```suggestion
fn parse_window_functions() {
```

alamb (commented):
Thanks you for the contribution @yuval-illumex  -- the basic idea is good but I think it needs a few cleanups

alamb:
Thank you for the review @iffyio  -- I appreciate it

tests/sqlparser_common.rs
yuval-illumex: Thanks for the feedback @alamb .
Please look at the syntax of `FIRST_VALUE` (for example) in some of the dialects:
https://docs.snowflake.com/en/sql-reference/functions/first_value
https://learn.microsoft.com/en-us/sql/t-sql/functions/first-value-transact-sql?view=sql-server-ver16
https://www.postgresqltutorial.com/postgresql-window-function/postgresql-first_value-function/

The `Over` Keyword is a must. so the syntax `FIRST_VALUE(x ORDER BY x)` is not valid.
Even if the tests pass, those tests are not aligned with the syntax.
That's why I added two arrays of dialects - for each dialect his supported syntax.

yuval-illumex:
Thank you @alamb for your feedback! Please see my comment on the tests

tests/sqlparser_common.rs
alamb: > The Over Keyword is a must. so the syntax FIRST_VALUE(x ORDER BY x) is not valid.

I think that is true when the FIRST_VALUE function is being used as a window function

Some systems allow `FIRST_VALUE` to be used as normal aggregate functions -- so like

```sql
SELECT 
  FIRST_VALUE(amount ORDER BY time), 
  LAST_VALUE(amount ORDER BY time)
FROM 
  t
GROUP BY
  currency
```

Perhaps @mustafasrepo  can offer some perspective as the author of https://github.com/sqlparser-rs/sqlparser-rs/pull/882

But basically I think we should retain these tests

src/ast/mod.rs
```diff
@@ -1078,6 +1078,24 @@ impl fmt::Display for WindowFrameUnits {
     }
 }
 
+/// Specifies Ignore / Respect NULL wihtin rank functions
```
alamb, line 1081:
```suggestion
/// Specifies Ignore / Respect NULL within window functions.
/// For example 
/// `FIRST_VALUE(column2) IGNORE NULLS OVER (PARTITION BY column1)`
```

alamb (commented):
Thanks @yuval-illumex  -- we are getting close I think

tests/sqlparser_common.rs
mustafasrepo: Actually `FIRST_VALUE` and `LAST_VALUE` can be used as window or aggregate function (similar to `SUM`, `MIN`, etc.). In these tests, they are used as aggregate function. Hence these tests should remain as is.

yuval-illumex: Thanks @alamb  and @mustafasrepo :) Learned something about these functions :)
I maintained the old tests

yuval-illumex:
Thanks @alamb  - it's ready for another review

alamb (approved):
Looks good to me -- thanks again @yuval-illumex  and @mustafasrepo

PR #1000, opened 2023-10-12
Support multiple tables in  `TRUNCATE` command 

mdashti:
The `truncate` command can accept a list of tables (as per the documentation: https://www.postgresql.org/docs/current/sql-truncate.html). This PR fixes the current limitation that only one table can be used in the `TRUNCATE` command.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1193,10 +1193,10 @@ pub enum Statement {
         noscan: bool,
         compute_statistics: bool,
     },
-    /// Truncate (Hive)
+    /// Truncate (Hive and PostgreSQL)
     Truncate {
         #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
-        table_name: ObjectName,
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -1982,12 +1982,20 @@ impl fmt::Display for Statement {
...
             }
             Statement::Truncate {
-                table_name,
+                table_names,
                 partitions,
                 table,
             } => {
                 let table = if *table { "TABLE " } else { "" };
-                write!(f, "TRUNCATE {table}{table_name}")?;
+                write!(
```
alamb, line 1990: Perhaps you can use `display_comma_separated` here instead of making a a new `Vec` via join

alamb (approved):
Thank you @mdashti  - this PR looks good to me, though I have a small suggestion for improvement

alamb:
Converting to draft as CI is failing. Please mark as ready for review when it is

PR #1001, opened 2023-10-12
Support Snowflake - allow number as placeholder (e.g. `:1`)

yuval-illumex:
The following query is valid in snowflake:

`SELECT :1 from table`


src/parser/mod.rs
```diff
@@ -5208,6 +5208,7 @@ impl<'a> Parser<'a> {
         let next_token = self.next_token();
         match next_token.token {
             Token::Word(w) => Ok(w.to_ident()),
+            Token::Number(w, false) => Ok(Ident::new(w)),
```
alamb, line 5211: Doesn't this mean that this will now permit numbers as identifiers everywhere rather than just in placeholders?

I would have expected some of the existing Placeholder logic to be updated rather than the identifier logic.

However, the test clearly shows it is getting the right parsed values 🤔

alamb (commented):
Thank you @yuval-illumex

src/parser/mod.rs
yuval-illumex: @alamb Can you share an example you think won't be valid with this approach?

alamb: For example, with this PR the following sql is parsed as though `1` was a role name, which doesn't seem correc:=t

```sql
alter role 1 with name = "foo";
```

```shell
$ cat /tmp/foo.sql
alter role 1 with name = "foo";
```
```shell
$ cargo run --example cli -- /tmp/foo.sql  --ms
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/examples/cli /tmp/foo.sql --ms`
Parsing from file '/tmp/foo.sql' using MsSqlDialect
2023-10-24T10:45:40.390Z DEBUG [sqlparser::parser] Parsing sql 'alter role 1 with name = "foo";
'...
Round-trip:
'ALTER ROLE 1 RENAME TO "foo"'
Parse results:
[
    AlterRole {
        name: Ident {
            value: "1",
            quote_style: None,
        },
        operation: RenameRole {
            role_name: Ident {
                value: "foo",
                quote_style: Some(
                    '"',
                ),
            },
        },
    },
]
```

on main such a query fails like this:

```shell
2023-10-24T10:46:32.418Z DEBUG [sqlparser::parser] Parsing sql 'alter role 1 with name = "foo";
'...
Error during parsing: ParserError("Expected identifier, found: 1 at Line: 1, Column 12")
```

We should probably add a test for that 🤔

yuval-illumex: @alamb As I agree, I think it makes more sense that an invalid query getting parsed then a valid one (which is being used in my company). 
Unless you can help me with guidance on another way to fix this issue, As I'm not an experienced rust developer :(

yuval-illumex:
@alamb I add failed test and refactored the PR to identify numbers only within placeholder.
I would appreciate your feedback!

src/parser/mod.rs
```diff
@@ -4759,7 +4759,14 @@ impl<'a> Parser<'a> {
             Token::HexStringLiteral(ref s) => Ok(Value::HexStringLiteral(s.to_string())),
             Token::Placeholder(ref s) => Ok(Value::Placeholder(s.to_string())),
             tok @ Token::Colon | tok @ Token::AtSign => {
-                let ident = self.parse_identifier()?;
+                // Not calling self.parse_identifier()? because only in placeholder we want to check numbers as idfentifies
```
alamb, line 4762: ✅  makes sense to me 

```suggestion
                // Not calling self.parse_identifier()? because only in placeholder we want to check numbers as identifies
```

alamb (approved):
Thank you @yuval-illumex  -- this makes sense to me

PR #1002, opened 2023-10-13
`VACUUM` command for PostgreSQL

mdashti:
This PR adds the `VACUUM` command parsing to cover PostgreSQL syntax. The implementation follows the PostgreSQL documentation: https://www.postgresql.org/docs/current/sql-vacuum.html


src/ast/mod.rs
```diff
@@ -1832,6 +1833,11 @@ pub enum Statement {
         name: ObjectName,
         representation: UserDefinedTypeRepresentation,
     },
+
+    Vacuum {
```
alamb, line 1837: Can you please add a documentation string here and a link to the reference documentation? https://www.postgresql.org/docs/current/sql-vacuum.html

src/ast/mod.rs
```diff
@@ -1832,6 +1833,11 @@ pub enum Statement {
...
+        options: Vec<(Keyword, Option<String>)>,
```
alamb, line 1838: The pattern with the rest of the structs in this crate is to do the parsing into fields (so that downstream crates don't have to re-implement the string matching for different keywords)

So in this case this would mean something like the following

```rust
struct VacuumOptions {
  full: Option<bool>,
  freeze: Option<bool>,
...
}
```

src/dialect/postgresql.rs
```diff
@@ -34,6 +34,8 @@ impl Dialect for PostgreSqlDialect {
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::COMMENT) {
             Some(parse_comment(parser))
+        } else if parser.parse_keyword(Keyword::VACUUM) {
```
alamb, line 37: Can you please also add Vacuum to the "GenericDialect" which is the superset of all the other dialects, when there is no conflict. 

Also, I think it would be better to put the parsing code into `parser.rs` rather than here so it is easier to find (though I realize you are following the model of `parse_comment`

alamb (commented):
Thank you for the contribution @mdashti  -- I have a few requests prior to this being ready to merge

alamb:
Marking as draft as this is no longer waiting on comments

PR #1004, opened 2023-10-17
Support `INSERT IGNORE` in `MySql` and `GenericDialect`

emin100:
SQL: INSERT IGNORE INTO tasks (title, priority) VALUES ('Test Some Inserts', 1)
ERROR:  sql parser error: Expected SELECT, VALUES, or a subquery in the query body, found: INTO at Line: 1, Column 15


src/parser/mod.rs
```diff
@@ -6697,6 +6697,8 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let ignore = dialect_of!(self is MySqlDialect) && self.parse_keyword(Keyword::IGNORE);
```
alamb, line 6700: Can you please also add support for GenericDialect?

alamb (commented):
Thank you for the contribution @emin100 -- this PR fails a CI check

I made a PR with a proposed fix that targets this branch -- please give it a look: https://github.com/emin100/sqlparser-rs/pull/1

src/parser/mod.rs
emin100: Generic support added

alamb (approved):
Thank you @emin100

PR #1005, opened 2023-10-18
Support `IN ()` syntax of SQLite

lewiszlw:
image
SQLite3 supports in empty list expression.


lovasoa (approved):

src/parser/mod.rs
```diff
@@ -2390,14 +2394,14 @@ impl<'a> Parser<'a> {
...
-        let ret = self.parse_comma_separated(|p| p.parse_select_item());
+        let ret = self.parse_comma_separated1(|p| p.parse_select_item());
         self.options.trailing_commas = old_value;
 
         ret
     }
 
     /// Parse a comma-separated list of 1+ items accepted by `F`
-    pub fn parse_comma_separated<T, F>(&mut self, mut f: F) -> Result<Vec<T>, ParserError>
+    pub fn parse_comma_separated1<T, F>(&mut self, mut f: F) -> Result<Vec<T>, ParserError>
```
alamb, line 2404: given this is a public function, can you please keep the name the same (I see why it would make sense to name it consistently with `parse_comma_separated0` but in this case I think it would be better not to make such a large scale function name change)

alamb (commented):
Thank you for the contribution @lewiszlw  -- this looks good to me other than the renaming of `parse_comma_separated1`

src/parser/mod.rs
lewiszlw: Ok, changed back.

alamb (approved):
Thank you for the contribution as well as the testing @lewiszlw

alamb:
It appears there are some CI  on this PR

alamb (commented):
Hi @lewiszlw  -- thank you for this PR.

I was going through it and I don't fully understand all the code in `parse_comma_separated0`

Using the following version, the `tests/sqlparser_sqlite.rs` test passes (though some of the unit tests do not)

Can you please add tests in `tests/sqlparser_sqlite.rs` that show what this extra code is doing / what types of queries it is parsing?

```
    /// Parse a comma-separated list of 0+ items accepted by `F`
    pub fn parse_comma_separated0<T, F>(&mut self, f: F) -> Result<Vec<T>, ParserError>
    where
        F: FnMut(&mut Parser<'a>) -> Result<T, ParserError>,
    {
        if matches!(self.peek_token().token, Token::RParen)
        {
            return Ok(vec![])
        }

        self.parse_comma_separated(f)
    }
```

alamb:
my plan with this PR is to use the simplified version of the code I suggested above and merge it tomorrow unless I hear otherwise. Thanks again @lewiszlw  and @lovasoa

lovasoa:
Great ! Can we make a new release after that ?

lewiszlw:
@alamb Sorry for late.
I tested with your code
```
#[test]
fn parse_where_in_empty_list() {

    // case1 succeed
    let sql = "SELECT * FROM t1 WHERE a IN ()";
    let select = sqlite().verified_only_select(sql);
    if let Expr::InList { list, .. } = select.selection.as_ref().unwrap() {
        assert_eq!(list.len(), 0);
    } else {
        unreachable!()
    }

    // case2 fail
    let sql = "SELECT * FROM t1 WHERE a IN (,)";
    let select = sqlite_with_options(ParserOptions::new().with_trailing_commas(true))
        .verified_only_select(sql);
    if let Expr::InList { list, .. } = select.selection.as_ref().unwrap() {
        assert_eq!(list.len(), 0);
    } else {
        unreachable!()
    }

    // case3 succeed
    let sql = "SELECT * FROM t1 WHERE a IN (1)";
    let select = sqlite().verified_only_select(sql);
    if let Expr::InList { list, .. } = select.selection.as_ref().unwrap() {
        assert_eq!(list.len(), 1);
    } else {
        unreachable!()
    }
    
    // case4 succeed
    let sql = "SELECT * FROM t1 WHERE a IN (1,)";
    let select = sqlite_with_options(ParserOptions::new().with_trailing_commas(true))
        .verified_only_select(sql);
    if let Expr::InList { list, .. } = select.selection.as_ref().unwrap() {
        assert_eq!(list.len(), 1);
    } else {
        unreachable!()
    }
}
```

tests/sqlparser_sqlite.rs
```diff
@@ -386,13 +387,36 @@ fn parse_attach_database() {
...
+    let sql = "SELECT * FROM t1 WHERE a IN ()";
+    let select = sqlite().verified_only_select(sql);
+    if let Expr::InList { list, .. } = select.selection.as_ref().unwrap() {
+        assert_eq!(list.len(), 0);
+    } else {
+        unreachable!()
+    }
+
+    sqlite_with_options(ParserOptions::new().with_trailing_commas(true)).one_statement_parses_to(
+        "SELECT * FROM t1 WHERE a IN (,)",
```
alamb, line 401: FWIW this doesn't appear to be valid SQLite syntax:
```
sqlite> create table t as values (1);
sqlite> select * from t where column1 IN (,);
Parse error: near ",": syntax error
  select * from t where column1 IN (,);
                      error here ---^
```

alamb (changes requested):
Thank you again for your contribution @lewiszlw

I am worried that this PR may have unintended side  effects of parsing comma separated values which is critical functionality for sqlparser. 

Thus I would like to request we keep this change simpler and thus less likely to cause unintended consequences.

Here is an alternate proposal: https://github.com/sqlparser-rs/sqlparser-rs/pull/1028

lewiszlw:
Close in favor of https://github.com/sqlparser-rs/sqlparser-rs/pull/1028

PR #1007, opened 2023-10-18
Support `FILTER` in over clause

### LINKED ISSUE #1006: support FILTER clause in window functions with an OVER clause
The way the filter clause is currently implemented seems buggy, and fails to parse expressions like

```sql
select max(x) filter(where y) over () from t
```

which is valid [in SQLite](https://www.sqlite.org/windowfunctions.html) and [postgres](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES)

For instance, the following returns `4` in both postgres and sqlite:

```sql
select max(column1) filter (where column1 % 2 = 0) from (values (1),(2),(3),(4),(5)) t
```

See [the FILTER clause on modern-sql.com](https://modern-sql.com/feature/filter)

lovasoa:
support FILTER clause in window functions

fixes #1006

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -993,9 +993,9 @@ impl Display for WindowType {
 pub struct WindowSpec {
     /// `OVER (PARTITION BY ...)`
     pub partition_by: Vec<Expr>,
-    // OVER (ORDER BY ...)
+    /// `OVER (ORDER BY ...)`
     pub order_by: Vec<OrderByExpr>,
-    // OVER (window frame)
+    /// `OVER (window frame)`
     pub window_frame: Option<WindowFrame>,
 }
 
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -991,8 +991,11 @@ impl Display for WindowType {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct WindowSpec {
+    // OVER (PARTITION BY ...)
     pub partition_by: Vec<Expr>,
+    // OVER (ORDER BY ...)
     pub order_by: Vec<OrderByExpr>,
+    // OVER (window frame)
     pub window_frame: Option<WindowFrame>,
 }
 
@@ -3650,6 +3653,8 @@ impl fmt::Display for CloseCursor {
 pub struct Function {
     pub name: ObjectName,
     pub args: Vec<FunctionArg>,
+    // e.g. `x > 5` in `COUNT(x) FILTER (WHERE x > 5)`
+    pub filter: Option<Box<Expr>>,
     pub over: Option<WindowType>,
     // aggregate functions may specify eg `COUNT(DISTINCT x)`
     pub distinct: bool,
@@ -3697,6 +3702,10 @@ impl fmt::Display for Function {
                 display_comma_separated(&self.args),
                 display_comma_separated(&self.order_by),
             )?;
+
+            if let Some(filter_cond) = &self.filter {
+                write!(f, " FILTER (WHERE {filter_cond})")?;
+            }
 
             if let Some(o) = &self.over {
                 write!(f, " OVER {o}")?;
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -991,7 +991,7 @@ impl Display for WindowType {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct WindowSpec {
-    // OVER (PARTITION BY ...)
+    /// `OVER (PARTITION BY ...)`
     pub partition_by: Vec<Expr>,
     // OVER (ORDER BY ...)
     pub order_by: Vec<OrderByExpr>,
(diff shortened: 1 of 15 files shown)


lovasoa:
@alamb , I'd love to get your feedback on this :slightly_smiling_face: 

As before, I need this for [SQLPage](https://github.com/lovasoa/SQLpage)

src/ast/mod.rs
```diff
@@ -991,8 +991,11 @@ impl Display for WindowType {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct WindowSpec {
+    // OVER (PARTITION BY ...)
```
alamb, line 994: If we use `///` then these will show up in documentation as well:

```suggestion
    /// `OVER (PARTITION BY ...)`
```

src/ast/mod.rs
```diff
@@ -991,8 +991,11 @@ impl Display for WindowType {
...
     pub partition_by: Vec<Expr>,
+    // OVER (ORDER BY ...)
```
alamb, line 996:
```suggestion
    /// `OVER (ORDER BY ...)`
```

src/ast/mod.rs
```diff
@@ -991,8 +991,11 @@ impl Display for WindowType {
...
     pub order_by: Vec<OrderByExpr>,
+    // OVER (window frame)
```
alamb, line 998:
```suggestion
    /// `OVER (window frame)`
```

tests/sqlparser_sqlite.rs
```diff
@@ -290,6 +290,31 @@ fn parse_create_table_with_strict() {
     }
 }
 
+#[test]
+fn parse_window_function_with_filter() {
+    let sql = "SELECT max(x) FILTER (WHERE y) OVER () FROM t";
```
alamb, line 295: I wonder if you  can add another test with for a window function (aka that has an `OVER` clause) but that is not an aggregate. For example:

```sql
SELECT row_number() FILTER (WHERE y) OVER () FROM t
```

I tried it in postgres and found the syntax is parsed but the functionality is not implemented

```
ERROR:  FILTER is not implemented for non-aggregate window functions
LINE 1: SELECT row_number() FILTER (WHERE column1 > 1) OVER () FROM ...
```

src/ast/mod.rs
```diff
@@ -3650,6 +3653,8 @@ impl fmt::Display for CloseCursor {
 pub struct Function {
     pub name: ObjectName,
     pub args: Vec<FunctionArg>,
+    // e.g. `x > 5` in `COUNT(x) FILTER (WHERE x > 5)`
+    pub filter: Option<Box<Expr>>,
```
alamb, line 3657: The current way to represent aggregates with filters (in the select clause) is https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Expr.html#variant.AggregateExpressionWithFilter

The idea of adding the `filter` directly to the Function I think is more elegant and general

What do you think about (in some other PR) removing `AggregateExpressionWithFilter` and just using this more general formulation?

src/dialect/sqlite.rs
```diff
@@ -35,6 +35,10 @@ impl Dialect for SQLiteDialect {
             || ('\u{007f}'..='\u{ffff}').contains(&ch)
     }
 
+    fn supports_filter_during_aggregation(&self) -> bool {
```
alamb, line 38: This PR supports filtering for any function with an `OVER` clause, correct? Thus, this name is no longer quite right I don't think (again maybe this is something we change in a follow on PR) 🤔

alamb (approved):
Thanks @lovasoa -- I have a few suggestions and comments, but I also think we can merge this PR as is

src/ast/mod.rs
lovasoa: Yes, that is a good idea :+1:

lovasoa: But we would have to make sure that supporting things like `(x + y) filter (where z)` was actually accidental

src/dialect/sqlite.rs
lovasoa: Yes, let's look at that in a separate pr. This may need some investigation to make sure we do not remove useful functionality

tests/sqlparser_sqlite.rs
lovasoa: done in https://github.com/sqlparser-rs/sqlparser-rs/pull/1007/commits/dd51436086f0123abe7a31bb24d5515b936a091d

lovasoa:
@alamb : thanks for the review, I addressed your comments. We can see about removing AggregateExpressionWithFilter and supports_filter_during_aggregation, but I'd be careful because it will prevent parsing some statements that parsed before.

src/ast/mod.rs
```diff
@@ -3650,6 +3653,8 @@ impl fmt::Display for CloseCursor {
 pub struct Function {
     pub name: ObjectName,
     pub args: Vec<FunctionArg>,
+    // e.g. `x > 5` in `COUNT(x) FILTER (WHERE x > 5)`
```
lovasoa, line 3656:
```suggestion
    /// e.g. `x > 5` in `COUNT(x) FILTER (WHERE x > 5)`
```

alamb (approved):
Thanks @lovasoa  -- I still thing something is confused here as I don't think SQL supports things like `select foo FILTER(where x > 5) FROM t` which I think this PR does allow. 

However I think this is a reasonable improvement

It would be great to sort out the naming of `Dialect::supports_filter_during_aggregation` as that now also controls `filter during any function`

Issue #1008, opened 2023-10-19
In Snowflake dialect `field/100` tokenizes as `Word(field/100)` instead of `Word(field),Div,Number(100)`


lustefaniak:
Changes in https://github.com/sqlparser-rs/sqlparser-rs/pull/841 broke division support when there are no whitespace characters around(before) `/` for Snowflake.

Right now, I don't know how to fix it properly, and I just disabled those chars from the Snowflake dialect in my fork.

alamb:
Thank you @lustefaniak -- I will check out the proposed PR shortly

PR #1009, opened 2023-10-19
snowflake: Fix handling of `/~%` in the stage name

lustefaniak:
Fixes: #1008

Changed as discussed in the code review. Tokenizer accepts `/`, `%`, `~` normally, and then when parsing stage name, that is accepted as part of identifier.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1004,18 +1004,6 @@ impl<'a> Tokenizer<'a> {
                             }
                         }
                         Some(' ') => Ok(Some(Token::AtSign)),
-                        // Snowflake stage identifier, this should be consumed as multiple dot separated word tokens
-                        Some(_) if dialect_of!(self is SnowflakeDialect) => {
-                            let mut s = "@".to_string();
-                            s.push_str(&peeking_take_while(chars, |ch| {
-                                self.dialect.is_identifier_part(ch)
-                                    || ch == '/'
-                                    || ch == '~'
-                                    || ch == '%'
-                                    || ch == '.'
-                            }));
-                            Ok(Some(Token::make_word(&s, None)))
-                        }
                         Some(sch) if self.dialect.is_identifier_start('@') => {
                             self.tokenize_identifier_or_keyword([ch, *sch], chars)
                         }
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1004,6 +1004,18 @@ impl<'a> Tokenizer<'a> {
                             }
                         }
                         Some(' ') => Ok(Some(Token::AtSign)),
+                        // Snowflake stage identifier, this should be consumed as multiple dot separated word tokens
+                        Some(_) if dialect_of!(self is SnowflakeDialect) => {
+                            let mut s = "@".to_string();
+                            s.push_str(&peeking_take_while(chars, |ch| {
+                                self.dialect.is_identifier_part(ch)
+                                    || ch == '/'
+                                    || ch == '~'
+                                    || ch == '%'
+                                    || ch == '.'
+                            }));
+                            Ok(Some(Token::make_word(&s, None)))
+                        }
                         Some(sch) if self.dialect.is_identifier_start('@') => {
                             self.tokenize_identifier_or_keyword([ch, *sch], chars)
                         }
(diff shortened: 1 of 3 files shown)


tests/sqlparser_snowflake.rs
```diff
@@ -1010,15 +1013,9 @@ fn test_snowflake_stage_object_names() {
...
-        ObjectName(vec![Ident::new("@namespace"), Ident::new("%table_name")]),
-        ObjectName(vec![
-            Ident::new("@namespace"),
-            Ident::new("%table_name/path"),
-        ]),
-        ObjectName(vec![
-            Ident::new("@namespace"),
-            Ident::new("stage_name/path"),
-        ]),
+        ObjectName(vec![Ident::new("@namespace.%table_name")]),
```
alamb, line 1016: why were these tests changed?

alamb (commented):
Thanks @lustefaniak  -- I have a question about why the tests were changed

Perhaps @pawel-big-lebowski can help review this as the author of  https://github.com/sqlparser-rs/sqlparser-rs/pull/841

yuval-illumex:
Just wanted to say thanks to @lustefaniak - I also noticed this issue and wanted to tackle it.

tests/sqlparser_snowflake.rs
lustefaniak: Tbh, I don't know how much big difference that makes, I don't even know if `.` in Snowflake makes a difference or not when checking the test cases. I was even thinking about keeping the test assertions as they were and just ignoring the tests until that gets fixed properly.

lustefaniak: I think that if `.` is a separator (assuming ObjectName was used here), then possibly we could keep `At`,`Mod` etc tokens, and somehow concat them to get SF identifiers, I don't know Snowlake that much to make a decision.

src/tokenizer.rs
```diff
@@ -1004,6 +1004,18 @@ impl<'a> Tokenizer<'a> {
                             }
                         }
                         Some(' ') => Ok(Some(Token::AtSign)),
+                        // Snowflake stage identifier, this should be consumed as multiple dot separated word tokens
+                        Some(_) if dialect_of!(self is SnowflakeDialect) => {
+                            let mut s = "@".to_string();
+                            s.push_str(&peeking_take_while(chars, |ch| {
+                                self.dialect.is_identifier_part(ch)
+                                    || ch == '/'
```
pawel-big-lebowski, line 1012: why is that removed from `is_identifier_part` method?

src/tokenizer.rs
```diff
@@ -1004,6 +1004,18 @@ impl<'a> Tokenizer<'a> {
...
+                                    || ch == '~'
+                                    || ch == '%'
+                                    || ch == '.'
```
pawel-big-lebowski, line 1015: I think this line is causing a problem and requires tests to be changed. 

The way I understand a problem is `/` can be either DIV token or part of identifier and this depends on a previous token (if a previous token before `.` started with `@`). I am not sure if this should be solved within tokenizer.  

Perhaps we should let it become DIV and then parse differently `indentifier div identifier` and check if the first identifier starts with `@`. Looks like a problem to be solved at the parsing layer.

yuval-illumex: @pawel-big-lebowski are there more places that '/' is part of identifier except of 'namespace/path'?
I'm not familiar with another places - maybe better approach will be to handle div in `parse_copy_into`?

pawel-big-lebowski: @yuval-illumex supporting `namespace/path` in `parse_copy_into` sounds good to me.

tests/sqlparser_snowflake.rs
alamb: I am wary of changing tests when we don't understand the reason for why they changed

src/tokenizer.rs
lustefaniak: I redid the parsing so it tokenizes normally and then I collect all `@` `%` `/` as parts of Ident.

pawel-big-lebowski (approved):
Thanks for fixing this. 
Great job.

tests/sqlparser_snowflake.rs
lustefaniak: Updated PR to keep tokenising as before and instead add special handling in stage name parsing.

alamb (approved):
Thank you @lustefaniak for both this PR as well as https://github.com/sqlparser-rs/sqlparser-rs/pull/1025  -- they looks great.

PR #1010, opened 2023-10-19
Support "with" identifiers surrounded by backticks in `GenericDialect`

bitemyapp:
I'm reviving this patch from https://github.com/sqlparser-rs/sqlparser-rs/pull/652

Please let me know what I'd need to do to make this merge-able. I'd like to be able to stop operating a fork of `sqlparser-rs` for parsing our Hive/Athena SQL so any direction you can give is much appreciated.

`parse_describe` validates the behavior for the hive and generic dialects.

### DIFF
--- a/tests/sqlparser_hive.rs
+++ b/tests/sqlparser_hive.rs
@@ -20,7 +20,7 @@ use sqlparser::ast::{
     SelectItem, Statement, TableFactor, UnaryOperator, Value,
 };
 use sqlparser::dialect::{GenericDialect, HiveDialect};
-use sqlparser::parser::ParserError;
+use sqlparser::parser::{ParserError, ParserOptions};
 use sqlparser::test_utils::*;
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/sqlparser_hive.rs
```diff
@@ -32,6 +32,20 @@ fn parse_table_create() {
     hive().verified_stmt(iof);
 }
 
+fn generic(options: Option<ParserOptions>) -> TestedDialects {
+    TestedDialects {
+        dialects: vec![Box::new(GenericDialect {})],
```
alamb, line 37: Can you also please  test `HiveDialect` given this is in the `sqlparser_hive.rs` module?

alamb (approved):
Thank you @bitemyapp  -- I have just a small test request

tests/sqlparser_hive.rs
bitemyapp: I've looked it over and colocated the dialects. I didn't find any test-cases that tested `generic` and not `hive`. Everything was either `hive` alone or `hive` and `generic`, e.g. the `parse_create_function` test. I'm not sure it makes sense to blend them given how they get used in the tests. I think it makes sense to test hive-isms that `generic` should be able to handle in the module that tests hive-isms rather than duplicating everything into the generic test module.

bitemyapp: I don't mind doing a follow-up PR.

Perhaps I could do something like `parse_deeply_nested_expr_hits_recursion_limits` in the common module but parameterized over the dialect, then invoke the helper in the generic & hive test modules? If that sounds acceptable let me know and I will pursue that.

alamb: What I was suggesting was to remove `generic()` and test `parse_describe` with both hive and generic

PR #1012, opened 2023-10-20
fix column `COLLATE` not displayed

lustefaniak:


alamb (approved):
Thank you @lustefaniak  🙏

PR #1018, opened 2023-10-21
Test that `regexp` can be used as an identifier in postgres

lovasoa:
"regexp" is not a keyword in postgres and can be used as an identifier


alamb (approved):
Thank you @lovasoa

PR #1021, opened 2023-10-23
Support for single-quoted identifiers

lovasoa:
This does not support single-quoted table names, but supports the most common case of

```sql
select tablename.'column' from tablename
```

fixes https://github.com/sqlparser-rs/sqlparser-rs/issues/1020


src/parser/mod.rs
```diff
@@ -618,18 +618,29 @@ impl<'a> Parser<'a> {
...
+                                // SQLite has single-quoted identifiers
+                                id_parts.push(Ident::with_quote('\'', s))
+                            }
+                            Token::Mul => {
+                                return Ok(WildcardExpr::QualifiedWildcard(ObjectName(id_parts)));
+                            }
+                            _ => {
+                                return self
+                                    .expected("an identifier or a '*' after '.'", next_token);
+                            }
```
lovasoa, line 643: This we can do later if you want, @alamb , together with support for single-quoted table names.

alamb: I actually tried running the test in this PR without this change and it failed, so I am not quite sure what you are proposing to do later

alamb (approved):
This looks good to me @lovasoa  -- thank you. I had one question about your comment but otherwise I think this is ready to go

alamb:
I took the liberty of merging up from main to resolve a conflict

src/parser/mod.rs
lovasoa: @alamb 

This PR supports `select 't'.*` (which is tested) but not `select 't'.my_column`.

What I was saying is that if you find that this is not clean, we can remove support for `'t'.*` (and the associated part of the test) and re-add it later if/when we add general support for single-quoted table names.

alamb: I think it is ok -- thank you

PR #1023, opened 2023-10-24
Support `date` 'key' when using semi structured data

yuval-illumex:
Add support to the following query:

```
SELECT JSON_DATA:date::date as date
FROM t1

```


alamb (commented):
Thank you @yuval-illumex -- can you please add a test for this feature (so it doesn't get accidentally broken in some future refactor)?

yuval-illumex:
Thanks @alamb , Added test just like the other keywords

alamb (approved):
Thanks @yuval-illumex !

PR #1024, opened 2023-10-25
Snowflake - Support lateral flatten

yuval-illumex:
Snowflake has the possibility to use the keywords [lateral flatten](https://docs.snowflake.com/en/sql-reference/functions/flatten):

`f.value AS "Contact",
  f1.value:type AS "Type",
  FROM persons p,
  lateral flatten(input => p.c, path => 'contact') f,
  lateral flatten(input => f.value:business) f1;`

While it seems `lateral` is only indicating that it's based on the line above, I found it simple enough just to ignore it.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6319,10 +6319,8 @@ impl<'a> Parser<'a> {
     pub fn parse_table_factor(&mut self) -> Result<TableFactor, ParserError> {
         // Allow Support in Snowflake: "FROM lateral flatten(input => f.value:business) f1;"
         // Snowflake also supports LATERAL (query) so we need to idenitify this exact case
-        let flatten = match self.peek_nth_token(1).token {
-            Token::Word(w) if w.keyword == Keyword::FLATTEN => true,
-            _ => false,
-        };
+        let flatten =
+            matches!(self.peek_nth_token(1).token, Token::Word(w) if w.keyword == Keyword::FLATTEN);
 
         if self.parse_keyword(Keyword::LATERAL) && !flatten {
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -6317,7 +6317,12 @@ impl<'a> Parser<'a> {
 
     /// A table name or a parenthesized subquery, followed by optional `[AS] alias`
     pub fn parse_table_factor(&mut self) -> Result<TableFactor, ParserError> {
-        if self.parse_keyword(Keyword::LATERAL) {
+        // Allow Support in Snowflake: "FROM lateral flatten(input => f.value:business) f1;"
+        // Snowflake also supports LATERAL (query) so we need to idenitify this exact case
+        let flatten =
+            matches!(self.peek_nth_token(1).token, Token::Word(w) if w.keyword == Keyword::FLATTEN);
+
+        if self.parse_keyword(Keyword::LATERAL) && !flatten {
```
alamb, line 6325: https://docs.snowflake.com/en/sql-reference/constructs/join-lateral suggests to me that `lateral` can be on any item in the select list (aka any table factor), not simply the `flatten` function

It seems like the `FROM lateral flatten(input => f.value:business)` should be parsed as a 

```rust
TableFactor::Function{ lateral: true, expr: ... }
```

( `TableFactor::Function` doens't have a `lateral` flag yet )

It seems like the comment about lateral must always be followed by a subquery is incorrect, at least for postgres

alamb (commented):
Thank you for the contribution @yuval-illumex

PR #1025, opened 2023-10-26
common: Make sure + - * / % binary operators work the same in all dialects

lustefaniak:
After https://github.com/sqlparser-rs/sqlparser-rs/pull/1009 is merged this will pass.

From what I see in BigQuery docs, `da-sh-es` is invalid identifier when used without quotes.
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#unquoted_identifiers

### DIFF
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -615,8 +615,6 @@ fn parse_table_identifiers() {
         assert!(bigquery().parse_sql_statements(&sql).is_err());
     }
 
-    test_table_ident("da-sh-es", None, vec![Ident::new("da-sh-es")]);
-
     test_table_ident("`spa ce`", None, vec![Ident::with_quote('`', "spa ce")]);
 
     test_table_ident(
--- a/src/dialect/bigquery.rs
+++ b/src/dialect/bigquery.rs
@@ -27,10 +27,6 @@ impl Dialect for BigQueryDialect {
     }
 
     fn is_identifier_part(&self, ch: char) -> bool {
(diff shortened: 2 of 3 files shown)


tests/sqlparser_bigquery.rs
```diff
@@ -615,8 +615,6 @@ fn parse_table_identifiers() {
         assert!(bigquery().parse_sql_statements(&sql).is_err());
     }
 
-    test_table_ident("da-sh-es", None, vec![Ident::new("da-sh-es")]);
```
alamb, line 618: Why was this removed?

alamb (commented):
Thank you very much @lustefaniak  -- I have one question about the changed test, but otherwise looks good to me

I ran the tests locally after merging up from main and they worked great!

tests/sqlparser_bigquery.rs
lustefaniak: It's in the description: 

From what I see in BigQuery docs, da-sh-es is invalid identifier when used without quotes.
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#unquoted_identifiers

lustefaniak:
```
Unquoted identifiers
 - Must begin with a letter or an underscore (_) character.
 - Subsequent characters can be letters, numbers, or underscores (_).
```
So the existing code was wrong, as `da-sh-es` is `da - sh - es`

alamb (approved):
Thank you @lustefaniak

PR #1026, opened 2023-10-26
snowflake: add support for LATERAL FLATTEN and similar

lustefaniak:
I had it done like this, I was sure I already created PR with it 🤔 

Replaces https://github.com/sqlparser-rs/sqlparser-rs/pull/1024

Closes https://github.com/sqlparser-rs/sqlparser-rs/pull/1024

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -695,6 +695,13 @@ pub enum TableFactor {
         expr: Expr,
         alias: Option<TableAlias>,
     },
+    /// `e.g. LATERAL FLATTEN(<args>)[ AS <alias> ]`
+    Function {
+        lateral: bool,
+        name: ObjectName,
+        args: Vec<FunctionArg>,
+        alias: Option<TableAlias>,
+    },
     /// ```sql
     /// SELECT * FROM UNNEST ([10,20,30]) as numbers WITH OFFSET;
     /// +---------+--------+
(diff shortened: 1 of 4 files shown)


yuval-illumex:
@lustefaniak OMG Thanks!!! I was ready for a big refactor but you did it!

src/ast/query.rs
```diff
@@ -695,6 +695,13 @@ pub enum TableFactor {
         expr: Expr,
         alias: Option<TableAlias>,
     },
+    /// `e.g. LATERAL FLATTEN(<args>)[ AS <alias> ]`
+    Function {
```
alamb, line 699: In general it would probably be good to consolidate the `TableFunction` variant with `Function` but I think this is a pragmatic way to add the functionality.

Thank you @lustefaniak

tests/sqlparser_common.rs
```diff
@@ -5976,15 +5976,6 @@ fn lateral_derived() {
     chk(false);
     chk(true);
 
-    let sql = "SELECT * FROM customer LEFT JOIN LATERAL generate_series(1, customer.id)";
```
alamb, line 5979: Can you please update this test to show this is now parsing correctly?

And maybe add a test with some other syntax that doesn't work?

Like
```sql
SELECT * FROM LATERAL UNNEST ([10,20,30]) as numbers WITH OFFSET;
```

alamb (commented):
The code looks nice -- thank you @lustefaniak . A very elegant solution

The only comment I have is in the tests (I think we should retain the error coverage and add a test in common as well)

tests/sqlparser_common.rs
lustefaniak: Added

alamb (approved):
Thanks again @lustefaniak

PR #1027, opened 2023-10-27
snowflake: PIVOT on derived table factors

lustefaniak:


jmhain (approved):
Thanks, I was meaning to do this!

alamb (approved):
Thank you @lustefaniak

PR #1035, opened 2023-11-02
Adds support for PostgreSQL "END"

tobyhede:
Related to https://github.com/sqlparser-rs/sqlparser-rs/issues/1034

Adds support for treating `END` as `COMMIT`. 
END is still used as a statement delimiter, but we only check this case if expecting a statement to end. 

I am not 100% sure this is the right approach, but the tests pass.


alamb (approved):
Thank you for the contribution @tobyhede 🙏

alamb:
Thanks again @tobyhede

PR #1036, opened 2023-11-03
Support `INSERT INTO ... DEFAULT VALUES ...`

CDThomas:
This change adds support for `DEFAULT VALUES` in insert statements.

I'm specifically after [Postgres support](https://www.postgresql.org/docs/current/sql-insert.html), but it looks like `DEFAULT VALUES` is part of the [SQL grammar](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#from-default). It looks to be supported by [SQLite](https://www.sqlite.org/lang_insert.html) and [MSSQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/insert-transact-sql?view=sql-server-ver16#d-inserting-data-into-a-table-with-columns-that-have-default-values) at least as well (I haven't tested with these, though). I took a quick look at the MySQL and BigQuery docs and they don't seem to support `DEFAULT VALUES` (and there are probably others that I haven't checked as well).

This change includes a **breaking change to the AST** that makes the `source` field on `Statement::Insert` optional. This approach is based on the [implementation in Postgres](https://github.com/postgres/postgres/blob/0bc726d95a305ca923b1f284159f40f0d5cf5725/src/backend/parser/gram.y#L12090-L12095) where both `cols` and `selectStmt` (which looks to be the Postgres AST's equivalent to `source`) are empty. I'm not completely sure if this is appropriate for all dialects, so I'm definitely interested in feedback here 🙏


tests/sqlparser_common.rs
```diff
@@ -107,6 +107,71 @@ fn parse_insert_values() {
...
+            assert_eq!(on, None);
+            assert!(returning.is_some());
+            assert_eq!(source, None);
+            assert_eq!(table_name, ObjectName(vec!["test_table".into()]));
+        }
+        _ => unreachable!(),
+    }
+
+    let insert_with_default_values_and_on_conflict =
+        verified_stmt("INSERT INTO test_table DEFAULT VALUES  ON CONFLICT DO NOTHING");
```
CDThomas, line 154: There's some extra whitespace here before `ON CONFLICT` that should be fixed up by https://github.com/sqlparser-rs/sqlparser-rs/pull/1037.

src/ast/mod.rs
```diff
@@ -2241,7 +2241,14 @@ impl fmt::Display for Statement {
                 if !after_columns.is_empty() {
                     write!(f, "({}) ", display_comma_separated(after_columns))?;
                 }
-                write!(f, "{source}")?;
+
+                if let Some(source) = source {
+                    write!(f, "{source}")?;
+                }
+
+                if source.is_none() && columns.is_empty() {
```
lovasoa, line 2249:
```suggestion
                if source.is_none() {
```

I think we should never print `INSERT INTO my_table(my_column);`, which is syntactically invalid.

`INSERT INTO my_table(my_column) DEFAULT VALUES;` should successfully round-trip through sqlparser, since it's a logical error, but not a syntax error.

CDThomas: Thanks for the review @lovasoa.

>INSERT INTO my_table(my_column) DEFAULT VALUES; should successfully round-trip through sqlparser, since it's a logical error, but not a syntax error.

I thought about this a bit and double checked the [SQL grammar](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#from-default) and [implementation in Postgres](https://github.com/postgres/postgres/blob/0bc726d95a305ca923b1f284159f40f0d5cf5725/src/backend/parser/gram.y#L12090-L12095) and both seem to make `INSERT INTO my_table(my_column) DEFAULT VALUES;` invalid at the syntax level. I think it'd make sense for sqlparser to make this scenario a syntax error as well, but I could be missing something.

I also tested a statement out with Postgres in a [DB fiddle](https://www.db-fiddle.com/f/oiiMsDZ5g8727mnEvEx1Px/0) and confirmed that it raises a syntax error:
```sql
-- error: syntax error at or near "DEFAULT"
INSERT INTO test_table (test_column) DEFAULT VALUES;
```

What do you think?

lovasoa: You are right about postgres, but it's not a syntax error in sqlite. 

And `insert into t(x);` is a syntax error everywhere anyway, so I think we should never emit that.

CDThomas:
Just noticed that there's a bug in this implementation.

My additions don't play well with `after_columns` for Hive [here](https://github.com/sqlparser-rs/sqlparser-rs/blob/4cdaa40abe2caa17375245838b63f1295ed1fca0/src/parser/mod.rs#L7105). Looks like `INSERT INTO test_table DEFAULT VALUES (some_column)` doesn't get rejected.

I'll move this back to draft while I fix that up and add some test cases for invalid syntax as well.

CDThomas:
> Just noticed that there's a bug in this implementation.
> 
> My additions don't play well with `after_columns` for Hive [here](https://github.com/sqlparser-rs/sqlparser-rs/blob/4cdaa40abe2caa17375245838b63f1295ed1fca0/src/parser/mod.rs#L7105). Looks like `INSERT INTO test_table DEFAULT VALUES (some_column)` doesn't get rejected.
> 
> I'll move this back to draft while I fix that up and add some test cases for invalid syntax as well.

Should be fixed by https://github.com/sqlparser-rs/sqlparser-rs/pull/1036/commits/4f334bd7370e2805090c08cd1e3a7fca3d5ee6db. This is ready for review again.

src/ast/mod.rs
CDThomas: Ah interesting, I was not aware of the behaviour in SQLite.

I added some coverage for syntax errors with `DEFAULT VALUES` [here](https://github.com/sqlparser-rs/sqlparser-rs/pull/1036/files#diff-ae2dab2c924f4018c1daf0743b8fd912c44c5b63ae7144dd8b951bb3bd254349R186-R214). Do you think that that behaviour looks reasonable?

alamb (approved):
This looks good to me -- thank you @lovasoa for your review. @CDThomas  can you possibly merge up from the `main` branch to get the CI passing on this PR and I'll merge it in?

CDThomas:
Thanks again @alamb and @lovasoa. I've merged in main and CI is green.

alamb (approved):
Thanks again

PR #1037, opened 2023-11-03
Fix extra whitespace printed before `ON CONFLICT`

CDThomas:
I noticed while working on https://github.com/sqlparser-rs/sqlparser-rs/pull/1036 that there's some extra whitespace printed before `ON CONFLICT` clauses in insert statements. This PR should fix it up by removing the extra space in the `Display` impl.


lovasoa (approved):

src/ast/mod.rs
```diff
@@ -3513,7 +3513,7 @@ impl fmt::Display for OnInsert {
                 " ON DUPLICATE KEY UPDATE {}",
                 display_comma_separated(expr)
             ),
-            Self::OnConflict(o) => write!(f, " {o}"),
```
CDThomas, line 3516: The bug is that the whitespace here is doubled up with the whitespace added a few lines below: https://github.com/sqlparser-rs/sqlparser-rs/blob/edeb9a2df0c55bf072dcb66b10661987dc926b2d/src/ast/mod.rs#L3522

alamb (approved):
Thank you @CDThomas

alamb:
Looks like there is a CI failure but after that this PR should be good to go

lovasoa:
yes, I included the small lint fix in my previous pr, which is now merged

CDThomas:
Thank you @alamb and @lovasoa. I just rebased against main and `cargo clippy --all-targets --all-features -- -D warnings` passes for me locally on Rust 1.74.0.

alamb:
Thanks again @CDThomas  and @lovasoa

PR #1038, opened 2023-11-06
add support for MAX as a character length

lovasoa:
See:
https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16
https://learn.microsoft.com/en-us/sql/t-sql/data-types/nchar-and-nvarchar-transact-sql?view=sql-server-ver16

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -525,6 +525,7 @@ pub enum CharacterLength {
         /// Optional unit. If not informed, the ANSI handles it as CHARACTERS implicitly
         unit: Option<CharLengthUnits>,
     },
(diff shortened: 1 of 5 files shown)


lustefaniak:
Also Redshift supports that syntax

lovasoa:
Good, let's kill two birds with one stone :)

@alamb , if you could have a look at it... The changes on this one are minimal.

alamb (approved):
Makes sense to me -- thank you @lovasoa  and @lustefaniak

lovasoa:
thanks @alamb ! That'll be in SQLPage's next version.

PR #1040, opened 2023-11-06
Update sqlparser-derive to use `syn 2.0`

serprex:
https://github.com/dtolnay/syn/releases/tag/2.0.0


tobyhede:
@serprex Some more context on this change would be useful ... I assume it is updating syn?

serprex:
Yes

derive/Cargo.toml
```diff
@@ -18,6 +18,6 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
-syn = "1.0"
+syn = { version = "2.0", features = ["full"] }
```
alamb, line 21: Do we really need "features=full"? Keeping down the dependency list for this crate is important I think given how low level it is

alamb:
Thank you for this contribution @serprex

derive/Cargo.toml
serprex: > full — Data structures for representing the syntax tree of all valid Rust source code, including items and expressions.

Code did not compile without enabling full feature

serprex:
Won't have time this weekend to look into, but will look at how to run tests & fix next week (I only ran `cargo test` in derive crate)

alamb:
Thanks @serprex .  Converting to a draft while review feedback is being addressed. Please mark it ready for review when it is ready for another look

derive/Cargo.toml
```diff
@@ -18,6 +18,6 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
-syn = "1.0"
+syn = { version = "2.0", default-features = false, features = ["printing", "parsing", "derive", "proc-macro"] }
```
alamb, line 21: It seems according to 
https://docs.rs/syn/latest/syn/index.html#optional-features

Listing these 4 will at least avoid

> clone-impls (enabled by default) — Clone impls for all syntax tree types.

alamb (approved):
Looks good to me -- thank you @serprex 

I took the liberty of merging up from main and reducing the feature set used

alamb:
Thanks again @serprex  and @tobyhede

PR #1044, opened 2023-11-09
add {pre,post}_visit_query to Visitor

jmhain:
Title says it all.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -740,7 +740,6 @@ pub enum TableFactor {
     /// For example `FROM monthly_sales PIVOT(sum(amount) FOR MONTH IN ('JAN', 'FEB'))`
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/pivot>
     Pivot {
-        #[cfg_attr(feature = "visitor", visit(with = "visit_table_factor"))]
         table: Box<TableFactor>,
         aggregate_function: Expr, // Function expression
         value_column: Vec<Ident>,
@@ -756,7 +755,6 @@ pub enum TableFactor {
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/unpivot>.
     Unpivot {
-        #[cfg_attr(feature = "visitor", visit(with = "visit_table_factor"))]
         table: Box<TableFactor>,
         value: Ident,
         name: Ident,
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -26,6 +26,7 @@ use crate::ast::*;
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+#[cfg_attr(feature = "visitor", visit(with = "visit_query"))]
 pub struct Query {
     /// WITH (common table expressions, or CTEs)
     pub with: Option<With>,
--- a/derive/README.md
+++ b/derive/README.md
@@ -48,33 +48,86 @@ impl Visit for Bar {
 }
 ```
 
-Additionally certain types may wish to call a corresponding method on visitor before recursing
+Some types may wish to call a corresponding method on the visitor:
 
 ```rust
 #[derive(Visit, VisitMut)]
 #[visit(with = "visit_expr")]
 enum Expr {
-    A(),
-    B(String, #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))] ObjectName, bool),
+    IsNull(Box<Expr>),
+    ..
 }
 ```
 
-Will generate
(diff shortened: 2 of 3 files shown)


derive/src/lib.rs
```diff
@@ -111,19 +111,48 @@ impl Attributes {
     }
 
     /// Returns the pre and post visit token streams
-    fn visit(&self, s: TokenStream) -> (Option<TokenStream>, Option<TokenStream>) {
+    fn visit(&self, s: TokenStream, is_option: bool) -> (Option<TokenStream>, Option<TokenStream>) {
```
alamb, line 114: Can you please also document what `is_option` means in this contex?

alamb (commented):
Thank you for this contribution @jmhain  -- can you please add a test for this new functionality?

jmhain:
@alamb Sorry for not including the tests in the original PR, just wanted to check that this funcitonal

jmhain:
@alamb Sorry for not including tests in the original PR, I just wanted to make sure this functionality was desired first. :) In adding the tests I found a few bugs that resulted from me misunderstanding how the visit macro works. If you include the `#[visit]` attribute on both the type and the field, then the visitor methods actually end up being called twice. My recent PR that added `UNPIVOT` support had the same issue, as did the existing `PIVOT` table factor that I had used as a reference.

I've fixed each of these, added tests that previously reproduced the redundant calls, and added some additional details to the `README.md` for the derive crate to ensure that others don't make the same error.

src/ast/query.rs
```diff
@@ -739,7 +740,6 @@ pub enum TableFactor {
     /// For example `FROM monthly_sales PIVOT(sum(amount) FOR MONTH IN ('JAN', 'FEB'))`
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/pivot>
     Pivot {
-        #[cfg_attr(feature = "visitor", visit(with = "visit_table_factor"))]
```
alamb, line 742: why were these removed?

alamb (commented):
Thank you @jmhain  -- this looks like a nice contribution to me. @tustvold  do you have a few minutes to review this?

derive/src/lib.rs
jmhain: The changes to the derive macro turned out not to be necessary

src/ast/query.rs
jmhain: `TableFactor` is already annotated with this, so it results in two calls to both the `pre_` and `post_visit_table_factor` for the same table factor, e.g. this is the output of one of the new tests I added without this change:

```
 [
     "PRE: STATEMENT: SELECT * FROM monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d) ORDER BY EMPID",
     "PRE: QUERY: SELECT * FROM monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d) ORDER BY EMPID",
     "PRE: TABLE FACTOR: monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d)",
     "PRE: TABLE FACTOR: monthly_sales",
     "PRE: TABLE FACTOR: monthly_sales",
     "PRE: RELATION: monthly_sales",
     "POST: RELATION: monthly_sales",
     "POST: TABLE FACTOR: monthly_sales",
     "POST: TABLE FACTOR: monthly_sales",
     "PRE: EXPR: SUM(a.amount)",
     "PRE: EXPR: a.amount",
     "POST: EXPR: a.amount",
     "POST: EXPR: SUM(a.amount)",
     "POST: TABLE FACTOR: monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d)",
     "PRE: EXPR: EMPID",
     "POST: EXPR: EMPID",
     "POST: QUERY: SELECT * FROM monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d) ORDER BY EMPID",
     "POST: STATEMENT: SELECT * FROM monthly_sales PIVOT(SUM(a.amount) FOR a.MONTH IN ('JAN', 'FEB', 'MAR', 'APR')) AS p (c, d) ORDER BY EMPID",
 ]
```

Note the duplicate calls for `TABLE FACTOR: monthly_sales"`. 

I added a section in the `README.md` for the derive crate that explains this in more detail: https://github.com/jmhain/sqlparser-rs/blob/visit_query/derive/README.md

I can split this into a separate PR if you'd like since it's separate from adding `{pre,post}_visit_query`, I just discovered it now because I almost introduced another instance of this same bug here.

alamb (approved):
Thank you very much @jmhain

Issue #1050, opened 2023-11-19
Support more forms of generated columns for SQLite dialect


takluyver:
SQLite allows some ways of defining a generated column which sqlparser currently rejects.

First, the `GENERATED ALWAYS` keywords are optional:

```rust
Parser::parse_sql(&SQLiteDialect {},
    "CREATE TABLE t1(a INT, b INT AS (a * 2) STORED);")?;
```

```
Error: sql parser error: Expected ',' or ')' after column definition, found: AS at Line: 1, Column 30
```

Second, the clause can have a `VIRTUAL` keyword stuck on the end (instead of `STORED`). This is the default behaviour anyway, just made explicit.

```rust
Parser::parse_sql(&SQLiteDialect {},
    "CREATE TABLE t1(a INT, b INT GENERATED ALWAYS AS (a * 2) VIRTUAL);")?;
```

```
Error: sql parser error: Expected ',' or ')' after column definition, found: VIRTUAL at Line: 1, Column 58
```

~~I think the distinction of stored-or-not is also not exposed in the AST at present.~~

I can have a go at fixing one or both of these, but I wanted to record the issue clearly first.

takluyver:
#1051 fixed `VIRTUAL`. I'll work on making `GENERATED ALWAYS` optional, but probably not until next week.

PR #1051, opened 2023-11-19
Add support for generated virtual columns with expression

takluyver:
This allows the SQLite & MySQL syntax for generated virtual columns, addressing the second part of #1050.

```sql
-- These two statements are equivalent:
CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2) VIRTUAL);
CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2));
```

I've done a partial survey of what different databases support for generated columns:

- SQLite: generate columns from an expression, either VIRTUAL (evaluated on read) or STORED (on write), but must ALWAYS be generated - [docs](https://sqlite.org/gencol.html), [syntax reference](https://sqlite.org/lang_createtable.html)
- MySQL: same options as SQLite - [docs](https://dev.mysql.com/doc/refman/8.0/en/create-table-generated-columns.html)
- DuckDB: only VIRTUAL supported, as of v0.9.2 - [docs](https://duckdb.org/docs/archive/0.9.2/sql/statements/create_table#generated-columns).
- Postgres: columns generated from an expression can only be STORED & ALWAYS (as of v16, the current version), but there are additional options to generate a column from a sequence (AS IDENTITY) - [docs](https://www.postgresql.org/docs/current/ddl-generated-columns.html), [syntax reference](https://www.postgresql.org/docs/current/sql-createtable.html)
- MS-SQL has `GENERATED AS` syntax, but it doesn't seem to overlap with the other DBs above at all - it's something to do with versioning records, with no space for arbitrary expressions - [docs](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16). I haven't tried to cover this in this PR.

It looks like the generated-column code before this PR is based on Postgres.

I've expanded the meaning of `GeneratedAs::ExpStored`, which the docs previously described as Postgres specific, to any column generated from an expressions and stored. `GeneratedAs::Always` with `generation_expr` present now represents a VIRTUAL column in SQLite or MySQL. It might be neater to make stored/virtual separate from always/by-default, but that would be an API break.


tobyhede:
LGTM. 
Is there a test for the `STORED` option? 
And it might be worth adding tests for the MySQL dialect. Makes intention explicit and easier to track regressions.

takluyver:
There's one test covering the same `GENERATED AS expr STORED` syntax in Postgres:

https://github.com/sqlparser-rs/sqlparser-rs/blob/c887c4e54558db3d3940c23300a384b1752cdeaf/tests/sqlparser_postgres.rs#L208-L218

But I can add this for SQLite & MySQL too. :+1: 

If SQLite & MySQL have the same syntax for this, do you prefer me to copy & paste the test into their respective modules, or set up one test to run with both of them? I'll copy-paste for now.

tests/sqlparser_sqlite.rs
```diff
@@ -205,6 +205,17 @@ fn parse_create_sqlite_quote() {
     }
 }
 
+#[test]
+fn parse_create_table_gencol() {
+    let sql_default = "CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2))";
+    let sql_virt = "CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2) VIRTUAL)";
+    sqlite_and_generic().verified_stmt(sql_virt);
```
alamb, line 212: Doesn't one_statement_parsers_to also verify the statement? (as in I think `verified_stmt` is redundant here)

tests/sqlparser_sqlite.rs
```diff
@@ -435,7 +446,6 @@ fn sqlite_with_options(options: ParserOptions) -> TestedDialects {
 
 fn sqlite_and_generic() -> TestedDialects {
     TestedDialects {
-        // we don't have a separate SQLite dialect, so test only the generic dialect for now
```
alamb, line 438: 👍  nice drive by cleanup

tests/sqlparser_mysql.rs
```diff
@@ -507,6 +507,17 @@ fn parse_create_table_comment_character_set() {
     }
 }
 
+#[test]
+fn parse_create_table_gencol() {
+    let sql_default = "CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2))";
+    let sql_virt = "CREATE TABLE t1 (a INT, b INT GENERATED ALWAYS AS (a * 2) VIRTUAL)";
```
alamb, line 513: What do you think about adding additional support to the AST so you can recover the original SQL --- as implemented now this change will not permit people to know when the original query had `VIRTUAL` or not

alamb (commented):
Thank you for this contribution @takluyver  (and the review @tobyhede ). I left some comments -- let me know what you think

tests/sqlparser_sqlite.rs
takluyver: That makes sense, thanks - I'll remove the `verified stmt`.

tests/sqlparser_mysql.rs
takluyver: Virtual is the default, so adding the marker doesn't make a logical difference (at least in the databases I found that support this). Elsewhere in the tests, it looks like you don't preserve differences like this - e.g. `SHOW COLUMNS` vs `SHOW FIELDS` in the MySQL tests. But if you'd like to preserve this one, I'm happy to do that. :slightly_smiling_face: 

If so, I would probably introduce an enum called something like `GenExpressionQualifier` (though that's an ugly name...), which could be None, Virtual or Stored. The Stored option would then be redundant with `GeneratedAs::ExpStored`.

(For my next PR, I'm planning to make the `GENERATED ALWAYS` keywords optional - as they are in MySQL and SQLite. Let me know if you want the presence or absence of those to be preserved too.)

alamb: > Elsewhere in the tests, it looks like you don't preserve differences like this - e.g. SHOW COLUMNS vs SHOW FIELDS in the MySQL tests. But if you'd like to preserve this one, I'm happy to do that. 🙂

Yes, it is true that this crate is not always consistent. However, for newly added code let's try and preserve the ability to round trip the AST (though I see there is no note describing that detail on https://github.com/sqlparser-rs/sqlparser-rs -- I will try and add a note when I have time)

>  Let me know if you want the presence or absence of those to be preserved too.)

yes, please

takluyver: OK, the latest commit (6d69a43) preserves the distinction between `VIRTUAL`, `STORED` and no qualifier. As I mentioned, that does mean there are two fields which both record if `STORED` was used. That's not so tidy, but it preserves compatibility for any code that's checking for `GeneratedAs::ExpStored`.

I've called the new field & enum `generation_expr_mode` & `GeneratedExpressionMode`, which are both quite a mouthful. Let me know if you have a better name for this. :slightly_smiling_face:

src/ast/ddl.rs
```diff
@@ -692,6 +689,16 @@ pub enum GeneratedAs {
     ExpStored,
 }
 
+/// `GeneratedExpressionMode`s are modifiers that follow an expression in a `generated`.
+/// No modifier is typically the same as Virtual.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum GeneratedExpressionMode {
```
alamb, line 697: 👍  looks good to me -- I don't have a better name for this

alamb (approved):
Thank you @takluyver  -- looks good to me

takluyver:
Ooops, I left some debugging bits in the tests. I've removed them now - do you prefer me to squash that into the previous commit and force-push the branch?

alamb:
> Ooops, I left some debugging bits in the tests. I've removed them now - do you prefer me to squash that into the previous commit and force-push the branch?

On merge all the commits will be squashed into a single commit, so I would prefer that you just push new commits to your branch / PR as it makes reviewing the deltas easier

alamb:
Thanks again @takluyver

tests/sqlparser_mysql.rs
alamb: filed https://github.com/sqlparser-rs/sqlparser-rs/pull/1052 to document this property

PR #1058, opened 2023-11-29
Add support for generated columns skipping 'GENERATED ALWAYS' keywords

### LINKED ISSUE #1050: Support more forms of generated columns for SQLite dialect

takluyver:
SQLite, Mysql and DuckDB all let you skip the words `GENERATED ALWAYS`, and use `AS` to introduce a column generated from an expression. See the links in #1051 for details.

I've introduced an extra field in `ColumnOption::Generated` - `generated_kw` is true if the `GENERATED` keyword is present, and false if the SQL skips directly to `AS`. I'm not sure of the naming, but it's the best I've thought of so far. Currently, if `generation_expr` is None, `generated_kw` can only be true (and is assumed to be when recreating the SQL).

Perhaps a neater design would be to have different `ColumnOption` kinds for generated-from-expression vs. generated-from-sequence vs. whatever else. But that would require an API break again.

Closes #1050.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -601,7 +601,7 @@ pub enum ColumnOption {
         generation_expr: Option<Expr>,
         generation_expr_mode: Option<GeneratedExpressionMode>,
         /// false if 'GENERATED ALWAYS' is skipped (option starts with AS)
-        generated_kw: bool,
+        generated_keyword: bool,
     },
 }
 
@@ -643,7 +643,7 @@ impl fmt::Display for ColumnOption {
                 sequence_options,
                 generation_expr,
                 generation_expr_mode,
-                generated_kw,
+                generated_keyword,
             } => {
                 if let Some(expr) = generation_expr {
                     let modifier = match generation_expr_mode {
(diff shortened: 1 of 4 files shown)


src/ast/ddl.rs
```diff
@@ -600,6 +600,8 @@ pub enum ColumnOption {
         sequence_options: Option<Vec<SequenceOptions>>,
         generation_expr: Option<Expr>,
         generation_expr_mode: Option<GeneratedExpressionMode>,
+        /// false if 'GENERATED ALWAYS' is skipped (option starts with AS)
+        generated_kw: bool,
```
tobyhede, line 604: Personal preference, but I would be inclined to expand the name to `generated_keyword` for clarity.

tobyhede (approved):

tobyhede:
LGTM

alamb (approved):
Looks good to me -- thank you @takluyver  and thank you @tobyhede  for the review

takluyver:
Thanks!

PR #1061, opened 2023-12-03
Add support for BigQuery table and view options

iffyio:
Extends the parser with BigQuery support for
- Table/View level options
- Column level options
- Table creation configurations `CLUSTER BY`, `PARTITION BY`


src/ast/mod.rs
```diff
@@ -2713,6 +2752,25 @@ impl fmt::Display for Statement {
                 if let Some(order_by) = order_by {
                     write!(f, " ORDER BY ({})", display_comma_separated(order_by))?;
                 }
+                if let Some(bigquery_config) = big_query_config {
```
tobyhede, line 2755: The names `bigquery_config` and `big_query_config` are different, is that intentional?

src/ast/helpers/stmt_create_table.rs
```diff
@@ -72,6 +72,7 @@ pub struct CreateTableBuilder {
     pub on_commit: Option<OnCommit>,
     pub on_cluster: Option<String>,
     pub order_by: Option<Vec<Ident>>,
+    pub big_query_config: Option<Box<BigQueryCreateTableConfiguration>>,
```
tobyhede, line 75: Are there other dialects that will have similar variations on Create Table?
If yes, might be worth pulling this up to a more generic type - something like `CreateTableConfiguration` with variants for different dialects. 

Might be a premature abstraction.

iffyio: I think that makes sense to do! I've updated to use an enum as a result. There are a couple fields on the struct like the hive_* ones that could use that format so it seems reasonable to introduce yeah

src/ast/mod.rs
iffyio: Ah right that wasn't intentional, updated!

src/ast/ddl.rs
```diff
@@ -527,6 +528,29 @@ impl fmt::Display for ColumnDef {
     }
 }
 
+/// Column definition for a view.
```
alamb, line 531: Could you please add some examples to this doc comment showing what is allowed (and not)?

src/ast/ddl.rs
```diff
@@ -601,6 +625,14 @@ pub enum ColumnOption {
...
         generation_expr_mode: Option<GeneratedExpressionMode>,
     },
+    /// BigQuery specific: Explicit column options in a view [1] or table [2]
+    /// Syntax
+    /// ```sql
+    /// OPTIONS(description="field desc")
+    /// ```
+    /// [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#view_column_option_list
+    /// [2]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#column_option_list
+    SqlOptions(Vec<SqlOption>),
```
alamb, line 635: Given the SQL is `OPTION`s what would you think about keeping the AST similar?

```suggestion
    Options(Vec<SqlOption>),
```

src/ast/mod.rs
```diff
@@ -4220,12 +4279,39 @@ pub struct HiveFormat {
...
+    /// Table options list.
+    /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list>
+    pub options: Option<Vec<SqlOption>>,
+}
+
+/// Represents database specific configuration during table creation.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CreateTableConfiguration {
```
alamb, line 4305: I wonder if you considered inlining these into `Statement::CreateTable`  rather than adding a new structure?

Like

```
/// CREATE TABLE
CreateTable {
    or_replace: bool,
    temporary: bool,
    external: bool,
    /// A partition expression for the table.
    /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#partition_expression>
    partition_by: Option<Expr>,
    /// Table clustering column list.
    /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list>
     cluster_by: Option<Vec<Ident>>,
    /// Table options list.
    /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_option_list>
     options: Option<Vec<SqlOption>>,
...
 ```

I think that might be more consistent with other  dialect specific fields as well as `order_by` and `strict`.

Having an extra struct here also may be confusing as it doesn't seem to mirror the structure of the SQL -- for example,  it seems to imply to me that there is some sort of "CONFIG" keyword / clause that appears in the SQL when https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_statement seems to imply that `PARTITION BY` `CLUSTER BY` and `OPTIONS` are clauses of the `CREATE TABLE` statement itself

**Review by alamb — COMMENTED** [review]
Thank you for the contribution @iffyio  and I apologize for the late reivew (I typically batch the reviews of sqlparser PRs prior to release to maximize efficiency)

I left some suggestions, please let me know what you think

File: src/ast/mod.rs (hunk 5 above)
Comment (iffyio) [reply src/ast/mod.rs:4305]: Ah yes I figured to have them inline initially, agree that makes it less confusing. The only issue was with serializing back to SQL text - since the fields in bigquery's case have a defined order (e.g `PARTITION BY` before `CLUSTER BY`), the inlined fields might end up not being reusable by other dialects if they show up at different parts of the CREATE statement. I'm not so sure if that's worth optimizing for though, so happy to just inline them if that's preferrable

Comment (alamb) [reply src/ast/mod.rs:4305]: I think an inlined version would be better to maintain consistency. If someone needs a different order when serializing for another dialect we can add that as a follow on PR perhaps.


If for some reason we need to keep the separate struct I think that would be acceptable if there were some doc comments explaining the rational (as that is not easy to understand from the code).

Comment (iffyio) [reply src/ast/mod.rs:4305]: That sounds good! Will make updates to this

File: src/ast/ddl.rs (hunk 4 above)
Comment (iffyio) [reply src/ast/ddl.rs]: Done!

File: src/ast/ddl.rs (hunk 3 above)
Comment (iffyio) [reply src/ast/ddl.rs]: Updated the description

File: src/ast/mod.rs (hunk 5 above)
Comment (iffyio) [reply src/ast/mod.rs]: Inlined!

File: tests/sqlparser_bigquery.rs [hunk 6]
```diff
@@ -86,6 +86,168 @@ fn parse_raw_literal() {
...
+                columns
+            );
+            assert_eq!(
+                (
+                    Some(Box::new(Expr::Identifier(Ident::new("_PARTITIONDATE")))),
+                    Some(vec![Ident::new("userid"), Ident::new("age"),]),
+                    Some(vec![
+                        SqlOption {
+                            name: Ident::new("partition_expiration_days"),
+                            value: Expr::Value(number("1")),
```
alamb, line 224: 👍  for testing that the value came through as an `Expr`

alamb (approved):
Thank you @iffyio  -- this is looking great. 

Also thanks again @tobyhede  for your reviews

PR #1063, opened 2023-12-04
add support for CALL statements

lovasoa:
see https://dev.mysql.com/doc/refman/8.0/en/call.html

Adds support for MySQL's CALL statements 
(stored procedure invocation)

This one is very important for [SQLPage](https://github.com/lovasoa/SQLpage) too,
because it allows working around other unsupported syntax
by wrapping logic in stored procedures.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1437,6 +1437,7 @@ pub enum Statement {
         file_format: Option<FileFormat>,
         source: Box<Query>,
     },
+    Call(Function),
     Copy {
         /// The source of 'COPY TO', or the target of 'COPY FROM'
         source: CopySource,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -1715,7 +1716,9 @@ pub enum Statement {
     ///
     /// Note: this is a PostgreSQL-specific statement,
     /// but may also compatible with other SQL.
-    Discard { object_type: DiscardObject },
+    Discard {
+        object_type: DiscardObject,
+    },
```
lovasoa, line 1721: @alamb , the reformatting is made by cargo fmt

tests/sqlparser_mysql.rs
```diff
@@ -1817,6 +1817,14 @@ fn parse_hex_string_introducer() {
...
 }
 
+#[test]
+fn parse_call() {
+    mysql().verified_stmt("CALL my_procedure()");
+    mysql().verified_stmt("CALL my_procedure(1, 'a')");
+    mysql().verified_stmt("CALL my_procedure(1, 'a', @my_var)");
+    mysql().verified_stmt("CALL my_procedure");
+}
+
```
universalmind303, line 1827: It doesn't look like the `CALL` keyword is gated to `mysql` dialect at all. Can we either add some tests to generic, or gate it behind the mysql dialect? 

I'd prefer adding it to generic as other dialects such as duckdb also support `CALL`.

lovasoa: You are right ! I changed the tests.

alamb (approved):
Looks good to me -- thank you @lovasoa

PR #1065, opened 2023-12-14
Replace `type_id()` by trait method to allow wrapping dialects

jjbayer:
Part of the dialect-specific behavior of the parser is defined in the methods of the `Dialect` trait, but other parts of the behavior depend on the `dialect_of!` macro, which in turn checks the `Any::type_id` of the current dialect.

Because if this, it is currently not possible to define variation of a dialect that is treated as the base dialect (e.g. MySQL) by the parser.

Ideally, all dialect-specific behavior would be defined in the implementation of the `Dialect` trait. As a workaround, this PR solves the problem by introducing an overridable trait method that returns the `TypeId`.


src/dialect/mod.rs
```diff
@@ -248,4 +257,98 @@ mod tests {
...
+            fn is_identifier_part(&self, ch: char) -> bool {
+                self.0.is_identifier_part(ch)
+            }
+        }
+
+        let statement = r#"SELECT 'Wayne\'s World'"#;
+        let res1 = Parser::parse_sql(&MySqlDialect {}, statement);
+        let res2 = Parser::parse_sql(&WrappedDialect(MySqlDialect {}), statement);
+        assert!(res1.is_ok());
+        assert_eq!(res1, res2);
```
jjbayer, line 352: Without overriding `dialect()`, this assertion fails with

```
Err(
    TokenizerError(
        "Unterminated string literal at Line: 1, Column 23",
    ),
)
```

alamb (approved):
This makes sense to me -- thank you @jjbayer 

Ideally it would be nicer to have trait functions like `supports_create_view()` that control the parsers behavior rather than checking for specific dialects, as you mention, but that is a much larger change

src/dialect/mod.rs
```diff
@@ -248,4 +257,98 @@ mod tests {
     fn parse_dialect(v: &str) -> Box<dyn Dialect> {
         dialect_from_str(v).unwrap()
     }
+
+    #[test]
+    fn parse_with_wrapped_dialect() {
+        /// Wrapper for a dialect. In a real-world example, this wrapper
```
alamb, line 263: 👨‍🍳 👌  -- very nice example of creating a test illustrating the use case

jjbayer: Thank you!

jjbayer:
@alamb thank you for accepting this PR, much appreciated!

PR #1067, opened 2023-12-15
Add support for DEFERRED, IMMEDIATE, and EXCLUSIVE in SQLite's BEGIN TRANSACTION command

takaebato:
This PR adds support for the following `BEGIN TRANSACTION` syntaxes in SQLite. [docs](https://sqlite.org/lang_transaction.html)
- `BEGIN DEFERRED`
- `BEGIN IMMEDIATE`
- `BEGIN EXCLUSIVE`

Since there was no term in the documentation to refer to the words `DEFERRED`, `IMMEDIATE`, and `EXCLUSIVE`, I decided to use the term `modifier`. The term `mode` is also a candidate, but I avoided it to prevent confusion with [transaction mode](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#transaction-mode) in standard SQL.

~This PR's implementation follows the approach used in [previous PR](https://github.com/sqlparser-rs/sqlparser-rs/pull/281/files) with similar changes.~

I would love  to hear any ideas for better naming or implementations!


takaebato:
I found an [implementation](https://github.com/sqlparser-rs/sqlparser-rs/pull/947/files) that looks better, so I will refactor this PR to follow it.

takaebato:
This commit f2ad35e9d232cd2a4afd4b5715be6135db59416b is somewhat experimental, and I think it could be helpful for thoroughly testing all dialects. 
Additionally, when a new dialect is introduced, updating the ALL_DIALECT_NAMES list ensures it is included in all test cases, lowering the risk of overlooking any updates.

But I think it is a bit complex, so I don't mind at all switching to another approach.
I am looking forward to any feedback!

src/dialect/generic.rs
```diff
@@ -38,4 +38,8 @@ impl Dialect for GenericDialect {
     fn supports_group_by_expr(&self) -> bool {
         true
     }
+
+    fn supports_start_transaction_modifier(&self) -> bool {
```
alamb, line 42: 👍

tests/sqlparser_sqlite.rs
```diff
@@ -431,6 +431,35 @@ fn invalid_empty_list() {
     );
 }
 
+#[test]
+fn parse_start_transaction_with_modifier() {
+    let (supported_dialects, unsupported_dialects) =
+        partition_all_dialects_by_inclusion(vec!["generic", "sqlite"]);
```
alamb, line 437: it isn't clear to me that using a String is preferable here to simply explicitly referring to `SqliteDialect` and `GenericDialect`

Using a string defers any errors to runtime rather than compile time

alamb:
> This commit https://github.com/sqlparser-rs/sqlparser-rs/commit/f2ad35e9d232cd2a4afd4b5715be6135db59416b is somewhat experimental, and I think it could be helpful for thoroughly testing all dialects.

I agree it is fairly complex 

I think the more standard way for testing multiple dialects is to programatically construct `TestedDialects` like

https://github.com/sqlparser-rs/sqlparser-rs/blob/1933f194e76c9b9e228135f4cce1ca0936288dc4/tests/sqlparser_common.rs#L1658-L1668

This has all the available dialects -- perhaps you can derive the incude/exclude lists from that https://github.com/sqlparser-rs/sqlparser-rs/blob/64eccdbba2fbc69fdb76ee451a1c43e30e7bfe3c/src/test_utils.rs#L196-L213 instead as that already exists

alamb (commented):
Thank you for the contribution @takaebato . I have a suggestion on testing, but otherwise I think this PR looks very nice to me 👌

takaebato:
Thanks!
This `all_dialects_but_pg` is close to what was wanted!

tests/sqlparser_sqlite.rs
takaebato: Thanks for the feedback!

(I have not yet been able to learn Rust properly, so I may be missing the point..)
I first tried to pass the arguments like this
```rust
partition_all_dialects_by_inclusion(vec![Box::new(SQLiteDialect {}), Box::new(GenericDialect {})])
```
but I got the following error indicating that trait objects cannot be compared.
```
error[E0277]: can't compare `dyn dialect::Dialect` with `dyn dialect::Dialect`
    --> src/test_utils.rs:230:36
     |
230  |         .filter(|d| !dialect_names.contains(d))
     |                                    ^^^^^^^^ no implementation for `dyn dialect::Dialect == dyn dialect::Dialect`
     |
     = help: the trait `PartialEq` is not implemented for `dyn dialect::Dialect`
     = note: required for `Box<dyn dialect::Dialect>` to implement `PartialEq`
```
I couldn't think of a suitable way to deal with this, so I changed to using String which can be compared.
But yes, this is not a good solution, so you can skip it!

alamb (approved):
Thank you @takaebato

PR #1070, opened 2023-12-21
snowflake: fix rendering of SELECT TOP

jmhain:
In Snowflake, `SELECT TOP` only supports an unparenthesized integer constant as its operand. This syntax was being parsed successfully but the `Display` impl was rendering MSSQL style (i.e. with parentheses), which causes the resultant queries to fail in Snowflake. I fixed this by distinguishing the two different syntactic forms for `TOP` in the AST.


alamb (approved):
Looks good to me -- thank you @jmhain

PR #1072, opened 2023-12-22
Support Mysql `REPLACE` statement and `PRIORITY` clause of  `INSERT`

emin100:
New feature support
- Mysql replace statement was added
- Mysql insert and replace statement priority part were added

Examples:

REPLACE DELAYED INTO tasks (title, priority) VALUES ('Test Some Inserts', 1)

INSERT HIGH_PRIORITY INTO tasks (title, priority) VALUES ('Test Some Inserts', 1)


alamb:
Retrigger CI

src/parser/mod.rs
```diff
@@ -7379,6 +7380,31 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse an REPLACE statement
+    pub fn parse_replace(&mut self) -> Result<Statement, ParserError> {
+        if !dialect_of!(self is MySqlDialect) {
```
alamb, line 7385: Can you please also support Generic dialect, as explained in https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#new-syntax

src/parser/mod.rs
```diff
@@ -7379,6 +7380,31 @@ impl<'a> Parser<'a> {
...
+            return parser_err!("unmatched dialect", self.peek_token().location);
```
alamb, line 7386: If a user gets a "unmatched dialect" error when parsing their SQL I think it will be very confusing and will not tell them how to fix the problem

Can you please change the error to say something like "Unsupported statement REPLACE" or "unexpected keyword REPLACE"?

src/parser/mod.rs
```diff
@@ -7379,6 +7380,31 @@ impl<'a> Parser<'a> {
...
+        }
+
+        let insert = &mut self.parse_insert().unwrap();
+        if let Statement::Insert {
+            replace_into,
+            priority,
+            ..
+        } = insert
+        {
+            if *priority == Some(MysqlInsertPriority::HighPriority) {
```
alamb, line 7396: I don't understand why this is checking for `HIGH_PRIORITY` -- if it is a semantic check https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics please remove the check

If you want to change the parser to only parse `HIGH_PRIORITY` at this location that would be ok too

tests/sqlparser_mysql.rs
```diff
@@ -1035,6 +1036,92 @@ fn parse_ignore_insert() {
     }
 }
 
+#[test]
```
alamb, line 1039: Can you please add test coveage for LOW_PRIORITY and any errors (for example, unsupported `REPLACE`)

alamb (commented):
Thank you for the contribution @emin100  -- I think this is looking pretty close. I left some suggestions for your consideration.

tests/sqlparser_mysql.rs
emin100: Added extra tests

src/parser/mod.rs
emin100: Added generic dialects

src/parser/mod.rs
emin100: Fixed error message

src/parser/mod.rs
emin100: REPLACE INTO statement doesn't support to HighPriority. But insert statement supports. So, it seems ok.

emin100:
@alamb thanks for the suggestions. I changed the pr.

src/parser/mod.rs
alamb: I think this qualifies as a SEMANTIC check rather than a syntactic check. Can you please remove it?

alamb (commented):
Thank you @emin100  -- this is looking quite close. Thank you for sticking with it

src/ast/mod.rs
```diff
@@ -2404,18 +2408,26 @@ impl fmt::Display for Statement {
...
+                replace_into,
+                priority,
             } => {
                 if let Some(action) = or {
                     write!(f, "INSERT OR {action} INTO {table_name} ")?;
                 } else {
                     write!(
                         f,
-                        "INSERT{ignore}{over}{int}{tbl} {table_name} ",
+                        "{start}{prior}{ignore}{over}{int}{tbl} {table_name} ",
```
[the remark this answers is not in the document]
emin100: Ok

src/parser/mod.rs
emin100: Check was removed.

emin100:
@alamb If possible, can you release a new version that includes this change?

alamb (approved):
Thank you @emin100  -- this looks good to me. Thank you for sticking with it

PR #1075, opened 2023-12-24
Support SQLite column definitions with no type

### LINKED ISSUE #743: SQLite dialect doesn't support typeless columns
SQLite supports untyped columns in the schema, like in this example:

```sql
CREATE TABLE log_data (
    log_id INTEGER NOT NULL,
    key TEXT NOT NULL,
    value,
    PRIMARY KEY (log_id, key),
    FOREIGN KEY(log_id) REFERENCES logs(id) ON DELETE CASCADE
);
```

Unfortunately, parsing this `CREATE TABLE` statement with current `sqlparser` (0.27) results in:

```
sql parser error: Expected a data type name, found: ,
```

takluyver:
SQLite allows columns with no type information:

```sql
CREATE TABLE t1 (a, b AS (a * 2), c NOT NULL)
```

As far as I know, other SQL databases all expect a data type for each column. So for now I've allowed this only for the SQLite dialect. Should the generic dialect also support it? The docs suggest that it's 'permissive', but I wasn't sure if this would be stretching the point too far. :slightly_smiling_face: 

Closes #743

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -219,6 +219,8 @@ pub enum DataType {
     /// [hive]: https://docs.cloudera.com/cdw-runtime/cloud/impala-sql-reference/topics/impala-struct.html
     /// [bigquery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
     Struct(Vec<StructField>),
+    /// Specific to SQLite: no type specified (no coercion)
+    Unspecified,
 }
 
 impl fmt::Display for DataType {
@@ -379,6 +381,7 @@ impl fmt::Display for DataType {
                     write!(f, "STRUCT")
                 }
             }
+            DataType::Unspecified => Ok(()),
         }
     }
 }
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -219,7 +219,9 @@ pub enum DataType {
     /// [hive]: https://docs.cloudera.com/cdw-runtime/cloud/impala-sql-reference/topics/impala-struct.html
     /// [bigquery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
     Struct(Vec<StructField>),
-    /// Specific to SQLite: no type specified (no coercion)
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -4219,6 +4223,29 @@ impl<'a> Parser<'a> {
         })
     }
 
+    fn sqlite_untyped_col_helper(&mut self) -> bool {
```
tobyhede, line 4226: A more descriptive name would make the intent a bit clearer here - 'is_sqlite_column_type_unspecified` or something like that.

alamb: I agree with @tobyhede 's suggestion.

src/ast/data_type.rs
```diff
@@ -219,6 +219,8 @@ pub enum DataType {
     /// [hive]: https://docs.cloudera.com/cdw-runtime/cloud/impala-sql-reference/topics/impala-struct.html
     /// [bigquery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
     Struct(Vec<StructField>),
+    /// Specific to SQLite: no type specified (no coercion)
```
alamb, line 222: Can you also please add some of the context (e.g. the example) from the PR's description to this comment? 

Perhaps you can also note it is not supported by Generic dialect as well

alamb (approved):
Thank you @takluyver -- this is looking good to me. I think @tobyhede  has a good style suggestion and it would be really helpful to improve the doc comments before we merged

> Should the generic dialect also support it? The docs suggest that it's 'permissive', but I wasn't sure if this would be stretching the point too far.

I don't have a strong opinion. We can maybe start with SQLite only and then expand coverage for generic dialect if people want it

takluyver:
Thanks both! I've renamed the helper message and tried to document this feature better.

alamb (approved):
LGTM -- thank you very much @takluyver  (and happy new year 🎉 )

takluyver:
Thanks! Happy new year to you :fireworks:

PR #1077, opened 2023-12-30
Add support for ENABLE and DISABLE on ALTER TABLE for pg

tobyhede:
Add support for the following pg syntax

```
# enable
ALTER TABLE tab ENABLE ALWAYS RULE rule_name
ALTER TABLE tab ENABLE ALWAYS TRIGGER trigger_name
ALTER TABLE tab ENABLE REPLICA TRIGGER trigger_name
ALTER TABLE tab ENABLE REPLICA RULE rule_name
ALTER TABLE tab ENABLE ROW LEVEL SECURITY
ALTER TABLE tab ENABLE RULE rule_name
ALTER TABLE tab ENABLE TRIGGER ALL
ALTER TABLE tab ENABLE TRIGGER USER
ALTER TABLE tab ENABLE TRIGGER trigger_name

# disable
ALTER TABLE tab DISABLE ROW LEVEL SECURITY
ALTER TABLE tab DISABLE RULE rule_name
ALTER TABLE tab DISABLE TRIGGER ALL
ALTER TABLE tab DISABLE TRIGGER USER
ALTER TABLE tab DISABLE TRIGGER trigger_name
```


src/tokenizer.rs
```diff
@@ -727,10 +727,7 @@ impl<'a> Tokenizer<'a> {
                     // match binary literal that starts with 0x
                     if s == "0" && chars.peek() == Some(&'x') {
                         chars.next();
-                        let s2 = peeking_take_while(
-                            chars,
-                            |ch| matches!(ch, '0'..='9' | 'A'..='F' | 'a'..='f'),
-                        );
+                        let s2 = peeking_take_while(chars, |ch| ch.is_ascii_hexdigit());
```
tobyhede, line 730: New clippy rule dropped.

src/tokenizer.rs
```diff
@@ -1077,7 +1074,7 @@ impl<'a> Tokenizer<'a> {
                 match chars.peek() {
                     Some('$') => {
                         chars.next();
-                        for (_, c) in value.chars().enumerate() {
+                        for c in value.chars() {
```
tobyhede, line 1077: New clippy rule dropped.

src/dialect/mod.rs
```diff
@@ -349,6 +349,7 @@ mod tests {
             }
         }
 
+        #[allow(clippy::needless_raw_string_hashes)]
```
alamb, line 352: maybe we can fix this in a follow on PR (as in remove the clippy allow and then fix the tests to conform to the new clippy opinions)

alamb (approved):
Looks good to me -- thank you for the contribution @tobyhede

src/dialect/mod.rs
tobyhede: Yeah, is a good idea. I tried to fix, but without the raw the parser errors with `Unterminated string literal`. 
I wasn't sure enough of the underlying intent to make more a dramatic change.

PR #1079, opened 2024-01-01
Support for pg ADD GENERATED in ALTER COLUMN statements

tobyhede:
`ADD GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ]`

Includes the identity/sequence options

```
ALTER TABLE t ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY
ALTER TABLE t ALTER COLUMN id ADD GENERATED BY DEFAULT AS IDENTITY
ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY
ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY ( INCREMENT 1 MINVALUE 1 )
ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY ( )
```

Fixes https://github.com/sqlparser-rs/sqlparser-rs/issues/988


tobyhede:
@alamb Working on the broken test ... not happening locally.

tobyhede:
Actually. PEBKAC. As always 🤦

src/parser/mod.rs
```diff
@@ -4861,11 +4865,37 @@ impl<'a> Parser<'a> {
...
+            } else if self.parse_keywords(&[Keyword::ADD, Keyword::GENERATED]) {
+                let generated_as = if self.parse_keyword(Keyword::ALWAYS) {
+                    Some(GeneratedAs::Always)
+                } else if self.parse_keywords(&[Keyword::BY, Keyword::DEFAULT]) {
+                    Some(GeneratedAs::ByDefault)
+                } else {
+                    None
+                };
+
+                let _ = self.expect_keywords(&[Keyword::AS, Keyword::IDENTITY]);
```
alamb, line 4877: Why is this ignoring the result? shouldn't it at least check and return the error, like:

```suggestion
                self.expect_keywords(&[Keyword::AS, Keyword::IDENTITY])?;
```

src/parser/mod.rs
```diff
@@ -4861,11 +4865,37 @@ impl<'a> Parser<'a> {
...
+
+                let mut sequence_options: Option<Vec<SequenceOptions>> = None;
+
+                if self.peek_token().token == Token::LParen {
+                    self.expect_token(&Token::LParen)?;
+                    sequence_options = self.parse_create_sequence_options().ok();
```
alamb, line 4883: Likewise here, I don't understand why the error is ignored. If there is a reason maybe we can add a comment.

I also wonder if we are sure that if `parse_create_sequence_options` returns an error that it hasn't consumed any tokens

tests/sqlparser_postgres.rs
```diff
@@ -611,6 +611,20 @@ fn parse_alter_table_alter_column() {
...
+        .verified_stmt("ALTER TABLE t ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY");
+    pg_and_generic()
+        .verified_stmt("ALTER TABLE t ALTER COLUMN id ADD GENERATED BY DEFAULT AS IDENTITY");
+    pg_and_generic().verified_stmt("ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY");
+    pg_and_generic().verified_stmt(
+        "ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY ( INCREMENT 1 MINVALUE 1 )",
+    );
+
+    pg_and_generic().verified_stmt("ALTER TABLE t ALTER COLUMN id ADD GENERATED AS IDENTITY ( )");
+}
```
alamb, line 626: Given there is some seemingly non trival error handling, can you please add some tests that hit the errors (e.g. when the two statements I identified above actually return an error?)

alamb (commented):
Thank you for the contribution @tobyhede  -- this looks almost good to me with the exception of some of the error handling, as I have pointed out above

tests/sqlparser_postgres.rs
tobyhede: On it

tobyhede:
@alamb Added the missed errors, and added tests to cover the cases. Sorry for missing such basics.

alamb:
> @alamb Added the missed errors, and added tests to cover the cases. Sorry for missing such basics.

No worries at all -- thank you for your continued contributions!

alamb (approved):

Issue #1069, opened 2023-12-20
Cannot parse table alias or cte alias in Postgres dialect


KennyChenFight:
Sorry for bothering you.
I'm using sqlparser = "0.40.0"
This is my code:
```rust
use sqlparser::dialect::PostgreSqlDialect;
use sqlparser::parser::{Parser, ParserError};

fn main() {
    let dialect = PostgreSqlDialect {};

    let sql1 = r#"
    WITH existing AS (SELECT test_table.id FROM test_tables AS test_table WHERE (a = 12) AND (b = 34)),
     inserted AS (INSERT INTO test_tables AS test_table (id, a, b, c) VALUES (DEFAULT, 56, 78, 90) ON CONFLICT (a, b)
     DO UPDATE SET c = EXCLUDED.c WHERE (test_table.c != EXCLUDED.c)) SELECT c FROM existing"#;

    let sql2 = "INSERT INTO test_tables AS test_table (id, a) VALUES (DEFAULT, 123) ON CONFLICT DO NOTHING RETURNING id";

    let ast1 = Parser::parse_sql(&dialect, sql1);
    match ast1 {
        Ok(_) => {

        }
        Err(e) => {
            println!("fail ast1: {}", e);
        }
    }
    let ast2 = Parser::parse_sql(&dialect, sql2);
    match ast2 {
        Ok(_) => {
        }
        Err(e) => {
            println!("fail ast2: {}", e);
        }
    }
}
```
error output:
```bash
fail ast1: sql parser error: Expected SELECT, VALUES, or a subquery in the query body, found: AS at Line: 3, Column 43
fail ast2: sql parser error: Expected SELECT, VALUES, or a subquery in the query body, found: AS at Line: 1, Column 25
```

Thank you!

alamb:
It would be great to figure out what is going on here and add the relevant support

boydjohnson:
I'm interested in working on this issue. 

So adding these lines of code under https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser/mod.rs#L7459
```
let alias = if dialect_of!(self is PostgreSqlDialect) && self.parse_keyword(Keyword::AS)
{
    Some(self.parse_identifier()?)
} else {
    None
};
```
[...]

PR #1084, opened 2024-01-04
Add support for PostgreSQL Insert table aliases (#1069)

boydjohnson:
This is a fix for #1069

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -14,6 +14,7 @@
 #[cfg(not(feature = "std"))]
 use alloc::{
     boxed::Box,
+    format,
     string::{String, ToString},
     vec::Vec,
 };
@@ -1428,6 +1429,8 @@ pub enum Statement {
         /// TABLE
         #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
         table_name: ObjectName,
+        /// table_name as foo (for PostgreSQL)
+        table_alias: Option<Ident>,
         /// COLUMNS
         columns: Vec<Ident>,
         /// Overwrite (Hive)
@@ -2400,6 +2403,7 @@ impl fmt::Display for Statement {
                 ignore,
                 into,
                 table_name,
+                table_alias,
                 overwrite,
                 partitioned,
                 columns,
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -2411,6 +2414,16 @@ impl fmt::Display for Statement {
                 replace_into,
                 priority,
             } => {
+                let table_name = if let Some(Ident {
+                    value,
+                    quote_style: _,
+                }) = table_alias
+                {
+                    format!("{table_name} AS {value}")
```
alamb, line 2422: It seems like we should format the alias using whatever quote style was originally used

Maybe something like

```suggestion
                let table_name = if let Some(table_alias) = table_alias
                {
                    format!("{table_name} AS {table_alias}")
```

tests/sqlparser_postgres.rs
```diff
@@ -3500,3 +3500,69 @@ fn parse_join_constraint_unnest_alias() {
...
     );
 }
+
+#[test]
+fn test_complex_postgres_insert_with_alias() {
+    let sql1 = "WITH existing AS (SELECT test_table.id FROM test_tables AS test_table WHERE (a = 12) AND (b = 34)), inserted AS (INSERT INTO test_tables AS test_table (id, a, b, c) VALUES (DEFAULT, 56, 78, 90) ON CONFLICT(a, b) DO UPDATE SET c = EXCLUDED.c WHERE (test_table.c <> EXCLUDED.c)) SELECT c FROM existing";
+
+    pg().verified_stmt(sql1);
+
+    let sql2 = "INSERT INTO test_tables AS test_table (id, a) VALUES (DEFAULT, 123)";
```
alamb, line 3510: Could you please:
1. Make the `sql2` a second test (e.g. `test_simple_postgres_insert_with_alias`)
2. Add a test for a quoted identifier, something like  `INSERT INTO test_tables AS "Test_Table" (id, a) VALUES (DEFAULT, 123)`, perhaps

alamb (commented):
Thank you for this contribution @boydjohnson  - this is looking close.

src/ast/mod.rs
boydjohnson: Thanks, I followed your advice. I rebased and force-pushed. I hope that is ok.

alamb (approved):
Thank you @boydjohnson

PR #1091, opened 2024-01-12
Add support for PostgreSQL `^@` starts-with operator

### LINKED ISSUE #1081: Minor: `SingleQuotedByteStringLiteral` actually represents a bit string for PostgreSQL
Looking at the history I think this token was added for BigQuery, where `b'string'` is indeed byte string syntax. On PostgreSQL, however, this actually represents _bit_ string syntax [^1], eg: `b'01010011'`.

https://github.com/sqlparser-rs/sqlparser-rs/blob/a430d1a5a7bb04bbefd0f2fea07bf25c7fbce8b2/src/tokenizer.rs#L63-L64

Since the two syntaxes are identical there isn't anything to be done about the actual parsing (which is fine), but perhaps an additional small docstring note might be useful/helpful given that the token name references bytes?

Regards!

[^1]:  https://www.postgresql.org/docs/current/datatype-bit.html

alexander-beedie:
Support PostgreSQL-specific `^@`[^1] starts-with operator...

```sql
"col" ^@ 'xyz'
```
...as `BinaryOperator::PGStartsWith` (and `Token::CaretAt`) for PostgreSqlDialect and GenericDialect. (Also closes #1081, by adding a minor note about byte/bit strings 👍)

[^1]: PostgreSQL string function/op [docs](https://www.postgresql.org/docs/current/functions-string.html#FUNCTIONS-STRING-OTHER).

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -60,7 +60,8 @@ pub enum Token {
     DoubleQuotedString(String),
     /// Dollar quoted string: i.e: $$string$$ or $tag_name$string$tag_name$
     DollarQuotedString(DollarQuotedString),
-    /// Byte string literal: i.e: b'string' or B'string'
+    /// Byte string literal: i.e: b'string' or B'string' (note that some backends, such as
+    /// PostgreSQL, may treat this syntax as a bit string literal instead, i.e: b'10010101')
     SingleQuotedByteStringLiteral(String),
     /// Byte string literal: i.e: b"string" or B"string"
     DoubleQuotedByteStringLiteral(String),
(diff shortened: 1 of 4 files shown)


src/tokenizer.rs
```diff
@@ -60,7 +60,8 @@ pub enum Token {
     DoubleQuotedString(String),
     /// Dollar quoted string: i.e: $$string$$ or $tag_name$string$tag_name$
     DollarQuotedString(DollarQuotedString),
-    /// Byte string literal: i.e: b'string' or B'string'
+    /// Byte string literal: i.e: b'string' or B'string' (note that some backends, such as
```
alamb, line 63: 👍

tests/sqlparser_postgres.rs
```diff
@@ -1728,6 +1728,7 @@ fn parse_pg_binary_ops() {
         (">>", BinaryOperator::PGBitwiseShiftRight, pg_and_generic()),
         ("<<", BinaryOperator::PGBitwiseShiftLeft, pg_and_generic()),
         ("&&", BinaryOperator::PGOverlap, pg()),
+        ("^@", BinaryOperator::PGStartsWith, pg()),
```
alamb, line 1731: 👍

alamb (approved):
Looks good to me -- thank you @alexander-beedie

PR #1093, opened 2024-01-15
ISSUE-1088: Fix array_agg wildcard behavior

ReppCodes:
The behavior for qualified wildcards in array_agg is broken for postgresql, it throws an error when it should parse. This commit partially reverts 823635d2fcd2642254b14b5737c9758d96a9580b, and adds some test coverage for this behavior.

Notably this moves the WildcardExpr enum back into the main Expr enum, and adjusts usage/impl for it as needed.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -380,7 +380,10 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// CompositeAccess (postgres) eg: SELECT (information_schema._pg_expandarray(array['i','i'])).n
-    CompositeAccess { expr: Box<Expr>, key: Ident },
+    CompositeAccess {
+        expr: Box<Expr>,
+        key: Ident,
+    },
     /// `IS FALSE` operator
     IsFalse(Box<Expr>),
     /// `IS NOT FALSE` operator
@@ -474,7 +477,10 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// Unary operation e.g. `NOT foo`
-    UnaryOp { op: UnaryOperator, expr: Box<Expr> },
+    UnaryOp {
+        op: UnaryOperator,
+        expr: Box<Expr>,
+    },
     /// CONVERT a value to a different data type or character encoding. e.g. `CONVERT(foo USING utf8mb4)`
     Convert {
         /// The expression to convert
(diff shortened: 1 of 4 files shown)


ReppCodes:
I'm not sure why `rustfmt` wraps the `Expr` structs when you add new ones to the bottom, but I felt like fighting with it defeats the purpose of having it so I'm leaving those changes in unless there's an objection.

src/ast/mod.rs
```diff
@@ -380,7 +380,10 @@ pub enum Expr {
         right: Box<Expr>,
     },
     /// CompositeAccess (postgres) eg: SELECT (information_schema._pg_expandarray(array['i','i'])).n
-    CompositeAccess { expr: Box<Expr>, key: Ident },
+    CompositeAccess {
```
alamb, line 383: I wonder why this was reformatted. Maybe you are using a different version of `cargo fmt` or something 🤔

alamb:
Thanks @ReppCodes  -- I'll try and review this carefully over the next day or two

ReppCodes:
> Thanks @ReppCodes -- I'll try and review this carefully over the next day or two

Hi @alamb, just checking back in on this.  

Thank you!

alamb:
cc @panarch

tests/sqlparser_postgres.rs
```diff
@@ -3773,3 +3773,14 @@ fn test_simple_insert_with_quoted_alias() {
...
     )
 }
+
+#[test]
+fn parse_array_agg() {
+    // follows general function with wildcard code path
+    let sql = r#"SELECT GREATEST(sections_tbl.*) AS sections FROM sections_tbl"#;
+    pg().verified_stmt(sql);
+
+    // follows special-case array_agg code path
```
alamb, line 3783: Could you please also add a test with a multi-part identifier? LIke

```sql
SELECT GREATEST(my_schema.sections_tbl.*) AS sections FROM sections_tbl
```

alamb (approved):
👍  looks good to me -- thank you @ReppCodes 

I tested and in fact this is supported in postgres:

```sql
postgres=# SELECT ARRAY_AGG(t.*) AS sections FROM t;
 sections
----------
 {(1)}
(1 row)
```

🤯 

I suggest we add one more test but otherwise this PR is ready to go. 🙏

tests/sqlparser_postgres.rs
ReppCodes: Good idea, I added such tests for both the general path and the `array_agg` path.

alamb (approved):
🚀

alamb:
Thanks again @ReppCodes

PR #1094, opened 2024-01-15
Add APIs to reuse token buffers in `Tokenizer`

0rphon:
This is a super simple PR that adds two new methods to aid in the reuse of token buffers.

- `Tokenizer::tokenize_with_location_into`: operates identically to `Tokenizer::tokenize_with_location` except it allows you to supply your own buffer to use.
- `Parser::tokens`: allows you to retrieve the token buffer from a parser after you're done with it.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -539,21 +539,29 @@ impl<'a> Tokenizer<'a> {
 
     /// Tokenize the statement and produce a vector of tokens with location information
     pub fn tokenize_with_location(&mut self) -> Result<Vec<TokenWithLocation>, TokenizerError> {
+        let mut tokens: Vec<TokenWithLocation> = vec![];
+        self.tokenize_with_location_into(&mut tokens)
+            .map(|_| tokens)
+    }
+
+    /// Tokenize the statement and append tokens with location information into the provided buffer
+    pub fn tokenize_with_location_into(
+        &mut self,
+        buf: &mut Vec<TokenWithLocation>,
+    ) -> Result<(), TokenizerError> {
         let mut state = State {
             peekable: self.query.chars().peekable(),
             line: 1,
             col: 1,
         };
 
-        let mut tokens: Vec<TokenWithLocation> = vec![];
-
         let mut location = state.location();
         while let Some(token) = self.next_token(&mut state)? {
-            tokens.push(TokenWithLocation { token, location });
+            buf.push(TokenWithLocation { token, location });
 
             location = state.location();
         }
-        Ok(tokens)
+        Ok(())
     }
 
     // Tokenize the identifer or keywords in `ch`
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -544,13 +544,13 @@ impl<'a> Tokenizer<'a> {
     /// Tokenize the statement and produce a vector of tokens with location information
     pub fn tokenize_with_location(&mut self) -> Result<Vec<TokenWithLocation>, TokenizerError> {
         let mut tokens: Vec<TokenWithLocation> = vec![];
-        self.tokenize_with_location_into(&mut tokens)
+        self.tokenize_with_location_into_buf(&mut tokens)
             .map(|_| tokens)
     }
-    
+
     /// Tokenize the statement and append tokens with location information into the provided buffer.
     /// If an error is thrown, the buffer will contain all tokens that were successfully parsed before the error.
-    pub fn tokenize_with_location_into(
+    pub fn tokenize_with_location_into_buf(
         &mut self,
         buf: &mut Vec<TokenWithLocation>,
     ) -> Result<(), TokenizerError> {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -547,8 +547,9 @@ impl<'a> Tokenizer<'a> {
         self.tokenize_with_location_into(&mut tokens)
             .map(|_| tokens)
     }
-
-    /// Tokenize the statement and append tokens with location information into the provided buffer
+    
+    /// Tokenize the statement and append tokens with location information into the provided buffer.
+    /// If an error is thrown, the buffer will contain all tokens that were successfully parsed before the error.
     pub fn tokenize_with_location_into(
         &mut self,
         buf: &mut Vec<TokenWithLocation>,
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -31,6 +31,7 @@ use sqlparser::dialect::{
 };
 use sqlparser::keywords::ALL_KEYWORDS;
 use sqlparser::parser::{Parser, ParserError, ParserOptions};
+use sqlparser::tokenizer::Tokenizer;
 use test_utils::{
     all_dialects, alter_table_op, assert_eq_vec, expr_from_projection, join, number, only, table,
     table_alias, TestedDialects,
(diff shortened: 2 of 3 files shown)


alamb (commented):
Thank you for the contribution @0rphon  -- this PR looks pretty close to me

I had some naming sugestions

Also, I think we should add a test of this API in `tests/..` so that a future refactor doesn't accidentally break this API without also having to explicitly change a test

src/tokenizer.rs
```diff
@@ -543,21 +543,30 @@ impl<'a> Tokenizer<'a> {
 
     /// Tokenize the statement and produce a vector of tokens with location information
     pub fn tokenize_with_location(&mut self) -> Result<Vec<TokenWithLocation>, TokenizerError> {
+        let mut tokens: Vec<TokenWithLocation> = vec![];
+        self.tokenize_with_location_into(&mut tokens)
```
trungda, line 547: Just for my education, what do we usually do when we make breaking change in public API like this one? There is a behavior change here right?

alamb: I am not sure why this is a breaking change in the sense of "programs that used to compile against sqlparser will still compile against sqlparser after this change". Perhaps I am missing something but I think this PR just adds a new API

In terms of breaking public API changes, for this crate I normally put a note in the changelog

tests/sqlparser_common.rs
```diff
@@ -8078,3 +8079,16 @@ fn test_release_savepoint() {
...
     one_statement_parses_to("RELEASE test1", "RELEASE SAVEPOINT test1");
 }
+
+#[test]
+fn test_buffer_reuse() {
+    let d = GenericDialect {};
+    let q = "INSERT INTO customer WITH foo AS (SELECT 1) SELECT * FROM foo UNION VALUES (1)";
+    let mut buf = Vec::new();
+    Tokenizer::new(&d, q)
+        .tokenize_with_location_into_buf(&mut buf)
```
alamb, line 8089: 👍

src/tokenizer.rs
```diff
@@ -543,21 +543,30 @@ impl<'a> Tokenizer<'a> {
...
+            .map(|_| tokens)
+    }
+    
+    /// Tokenize the statement and append tokens with location information into the provided buffer.
+    /// If an error is thrown, the buffer will contain all tokens that were successfully parsed before the error.
+    pub fn tokenize_with_location_into(
```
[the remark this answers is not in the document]
0rphon: i changed it to `into_buf` to match the naming notation of standard library like `Read::read_buf` or `BufRead::fill_buf`. let me know if youd rather it be `into_buffer` though!

src/tokenizer.rs
0rphon: This PR *shouldnt* contain any breaking changes to the API. If you looked at the `tokenize_with_location_info_buf` method that's being called here, its functionally equivalent to the old code. I just separated it out so users can supply their own buffer if desired

trungda: Thank you! I misread. I thought that in case of failure, we would still return a partially tokenized vector but actually we are still returning an `Err`.

alamb (approved):
Thank you again for this contribution @0rphon and for the review @trungda

src/tokenizer.rs
0rphon: No worries! The new method `tokenize_with_location_into_buf` will actually return a partial vec of successfully parsed tokens. I specifically needed this behavior for a project im working on that parses a single SQL statement followed by a lot of random unparseable garbage. Then i needed the `into_tokens` API so i could figure out the length of the parsed statement using `Parser::index` and a custom `token_length` function i made. Once i was done with the fork though i realized my changes actually worked perfectly as a way to reuse buffers so i figured id PR it as a little optimization.

trungda: Thank @0rphon I am curious how you figure out the length of the parsed statement as I also need that for a [separate issue](https://github.com/apache/arrow-datafusion/pull/8918#issuecomment-1901243443). I think I'll want byte-length for a precise calculation but it's non-trivial to calculate the byte length for all variants of `Tokens`.

0rphon: I just based it off the formatting you use in `impl Display for Token`. Its definitely a little hacky and probably not accurate in every scenario, hence why i didn't add it to this PR. I was just trying to determine if a text blob contains an SQL statement, so i just needed an approximate len so i could display a rough output of the matched string.

```rust
fn token_len(t: &TokenWithLocation) -> usize {
    match &t.token {
        Token::EOF => 3,
        Token::Word(w) => w.value.len() + w.quote_style.map(|_| 2).unwrap_or_default(),
        Token::Number(n, l) => n.len() + *l as usize,
        Token::Char(c) if c.is_ascii() => 1,
        // todo is this correct?
        Token::Char(_) => 4,
        Token::SingleQuotedString(s) => s.len() + 2,
        Token::DoubleQuotedString(s) => s.len() + 2,
        Token::DollarQuotedString(s) => {
            s.value.len() + s.tag.as_ref().map(|t| t.len() * 2).unwrap_or_default() + 4
        }
        Token::NationalStringLiteral(s) => s.len() + 3,
        Token::EscapedStringLiteral(s) => s.len() + 3,
        Token::HexStringLiteral(s) => s.len() + 3,
        Token::SingleQuotedByteStringLiteral(s) => s.len() + 3,
        Token::DoubleQuotedByteStringLiteral(s) => s.len() + 3,
        Token::RawStringLiteral(s) => s.len() + 3,
        Token::Comma => 1,
        Token::Whitespace(Whitespace::Space) => 1,
        Token::Whitespace(Whitespace::Newline) => 1,
        Token::Whitespace(Whitespace::Tab) => 1,
        Token::Whitespace(Whitespace::SingleLineComment { comment, prefix }) => {
            comment.len() + prefix.len()
        }
        Token::Whitespace(Whitespace::MultiLineComment(s)) => s.len() + 4,
        Token::DoubleEq => 2,
        Token::Spaceship => 3,
        Token::Eq => 1,
        Token::Neq => 2,
        Token::Lt => 1,
        Token::Gt => 1,
        Token::LtEq => 2,
        Token::GtEq => 2,
        Token::Plus => 1,
        Token::Minus => 1,
        Token::Mul => 1,
        Token::Div => 1,
        Token::DuckIntDiv => 2,
        Token::StringConcat => 2,
        Token::Mod => 1,
        Token::LParen => 1,
        Token::RParen => 1,
        Token::Period => 1,
        Token::Colon => 1,
        Token::DoubleColon => 2,
        Token::DuckAssignment => 2,
        Token::SemiColon => 1,
        Token::Backslash => 1,
        Token::LBracket => 1,
        Token::RBracket => 1,
        Token::Ampersand => 1,
        Token::Caret => 1,
        Token::Pipe => 1,
        Token::LBrace => 1,
        Token::RBrace => 1,
        Token::RArrow => 2,
        Token::Sharp => 1,
        Token::ExclamationMark => 1,
        Token::DoubleExclamationMark => 2,
        Token::Tilde => 1,
        Token::TildeAsterisk => 2,
        Token::ExclamationMarkTilde => 2,
        Token::ExclamationMarkTildeAsterisk => 3,
        Token::AtSign => 1,
        Token::ShiftLeft => 2,
        Token::ShiftRight => 2,
        Token::Overlap => 2,
        Token::PGSquareRoot => 2,
        Token::PGCubeRoot => 3,
        Token::Placeholder(s) => s.len(),
        Token::Arrow => 2,
        Token::LongArrow => 3,
        Token::HashArrow => 2,
        Token::HashLongArrow => 3,
        Token::AtArrow => 2,
        Token::ArrowAt => 2,
        Token::HashMinus => 2,
        Token::AtQuestion => 2,
        Token::AtAt => 2,
    }
}
```

trungda: Ah I see. I was thinking of adding the actual byte position of a token in the struct `Location` to be able to achieve what I want. What do you think?

PR #1096, opened 2024-01-16
Add Postgres operators for the LIKE expression variants

### LINKED ISSUE #1095: Add support for Postgres LIKE operators
While at present there is support for the Postgres regex matching operators (`~`, `~*`, `!~`, `!~*`), there is no support for parsing their Postgres LIKE equivalents: https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE. 

These are the same as the regex operators, except that single-tilde's are replaced with double-tilde chars (`~~`, `~~*`, `!~~`, `!~~*`): 

> The operator ~~ is equivalent to LIKE, and \~~* corresponds to ILIKE. There are also !~~ and !~~* operators that represent NOT LIKE and NOT ILIKE, respectively. All of these operators are PostgreSQL-specific. You may see these operator names in EXPLAIN output and similar places, since the parser actually translates LIKE et al. to these operators.

gruuya:
Closes #1095

Extends the `Token` and `BinaryOperator` enums with variants representing `~~`, `~~*`, `!~~`, `!~~*`, in accordance with the present regular expression tokens/operators.

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -131,6 +131,14 @@ pub enum BinaryOperator {
     PGRegexNotMatch,
     /// String does not match regular expression (case insensitively), e.g. `a !~* b` (PostgreSQL-specific)
     PGRegexNotIMatch,
+    /// String matches pattern (case sensitively), e.g. `a ~~ b` (PostgreSQL-specific)
+    PGLikeMatch,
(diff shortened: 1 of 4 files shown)


alamb (approved):
This is a really nice contribution @gruuya  -- thank you. It is well commented and well tested in my opinion ❤️

PR #1098, opened 2024-01-17
Replace `AtomicUsize` with `Cell<usize>` in the recursion counter

### LINKED ISSUE #1097: Replace `AtomicUsize` with `Cell<usize>` in the parser's `RecursionCounter`
It seems the `AtomicUsize` in `RecursionCounter` makes little sense, except to satisfy the borrow checker, since there is no concurrent modification in the code. So why not just replace it with `Cell<usize>`? It provides the same functionality without extra overhead from atomic operations like `fetch_sub` and `fetch_add`.

Before: 
```rust
pub(crate) struct RecursionCounter {
    remaining_depth: Rc<AtomicUsize>,
}
```

After:
```rust
pub(crate) struct RecursionCounter {
    remaining_depth: Rc<Cell<usize>>,
}
```

wzzzzd:
This is a simple PR that eliminates the weird `AtomicUsize`.

Closes #1097

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -61,7 +61,7 @@ macro_rules! return_ok_if_some {
 #[cfg(feature = "std")]
 /// Implementation [`RecursionCounter`] if std is available
 mod recursion {
-    use core::sync::atomic::{AtomicUsize, Ordering};
+    use std::cell::Cell;
     use std::rc::Rc;
 
     use super::ParserError;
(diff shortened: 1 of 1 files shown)


alamb (commented):
Thank you for the contribution @wzzzzd  -- I am not familiar with the relative performance characteristics of `Cell` vs `AtomicUsize` -- I wonder did you see the atomic increments show up in a performance trace?

I think there are likely far more heavy weight operations (like string copying) that dominate most sql parsing operations

wzzzzd:
> Thank you for the contribution @wzzzzd -- I am not familiar with the relative performance characteristics of `Cell` vs `AtomicUsize` -- I wonder did you see the atomic increments show up in a performance trace?
> 
> I think there are likely far more heavy weight operations (like string copying) that dominate most sql parsing operations

I wrote a [microbenchmark](https://github.com/wzzzzd/sqlparser-rs/blob/cell_bench/sqlparser_bench/benches/recursion_counter_bench.rs) to compare the performance of `Cell` vs `AtomicUsize` in a deeply nested query, and the result shows that their difference is negligible (<5%) :(
Nevertheless, I think `Cell` should be preferred from the perspective of code style, since it does not indicate parallelism or concurrency (sql parsing is purely single-threaded).

jmhain:
+1 for this change, I was going to open a PR myself until I found this one. Using an atomic here is weird and unnecessary, since `Cell` exists for exactly this purpose (interior mutability of a value not shared between threads). The `Rc` wrapping it already makes it so that `RecursionCounter` and thus `Parser` are not `Send` or `Sync` so this should have no effect on the crate's api either.

alamb (approved):
Thanks @wzzzzd  and @jmhain

PR #1101, opened 2024-01-18
Allow string values in pragma commands

### LINKED ISSUE #1100: Support sqlite PRAGMA statements with string values
Hi, first of all, thank you for the amazing project.

I'm working on a DB client and need the ability to parse such statements:
```
PRAGMA table_info("sqlite_master")
PRAGMA table_info = "sqlite_master"
```
Which can be parsed and passed to SQLite on other DB clients.

invm:
Small PR to allow string values in pragma commands

```
PRAGMA table_info('sqlite_master')
PRAGMA table_info = "sqlite_master"
```

Closes #1100

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8478,7 +8478,7 @@ impl<'a> Parser<'a> {
             v @ Value::Placeholder(_) => Ok(v),
             _ => {
                 self.prev_token();
-                self.expected("number or string", self.peek_token())
+                self.expected("number or string or ? placeholder", self.peek_token())
             }
         }
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8470,11 +8470,24 @@ impl<'a> Parser<'a> {
         })
     }
 
+    fn parse_pragma_value(&mut self) -> Result<Value, ParserError> {
+        match self.parse_value()? {
+            v @ Value::SingleQuotedString(_) => Ok(v),
+            v @ Value::DoubleQuotedString(_) => Ok(v),
+            v @ Value::Number(_, _) => Ok(v),
+            v @ Value::Placeholder(_) => Ok(v),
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -8470,11 +8470,24 @@ impl<'a> Parser<'a> {
         })
     }
 
+    fn parse_pragma_value(&mut self) -> Result<Value, ParserError> {
+        match self.parse_value()? {
+            v @ Value::SingleQuotedString(_) => Ok(v),
+            v @ Value::DoubleQuotedString(_) => Ok(v),
+            v @ Value::Number(_, _) => Ok(v),
+            v @ Value::Placeholder(_) => Ok(v),
```
alamb, line 8478: Is there test coverage for placeholders? If not, can you pleaes add a test that uses a placeholder so we don't accidentally introduce a regression in the future?

alamb (commented):
Thank you for the contribution @invm  -- I think this code needs a little more test coverage but otherwise looks good to go

src/parser/mod.rs
invm: Totally agree, I don't think it'll introduce a regression as this case was not covered before as well, I've added a test for `table_info = ?`, could not make `table_info($)` work as it does not parse correctly, if you think this case should also be covered I would like a suggestion on how to write it, as `table_info('$')` does not look correct because placeholder usually are not surrounded by quotes.

Let me know it if suffices.

alamb: > Totally agree, I don't think it'll introduce a regression as this case was not covered before as well

yes, I agree -- sorry for my poor wording. What I was trying to say was that without the additional test for `?` someone else in the future might introduce a regression (break the parsing of `?`) without any tests failing.

alamb (approved):

PR #1106, opened 2024-01-20
Use `Option<Expr>` for Min and Max vals in Seq Opts, fix alter col seq display

### LINKED ISSUE #1105: Display for `SequenceOptions` in `AlterColumn` shouldn't check whether the option list is empty, Min and Max shouldn't be listed as `Empty` in Sequence Opt List
Here, the options are defined as `AS IDENTITY [ ( sequence_options ) ]`, which is to say, if parentheses exist, then there's an option list, regardless of whether it is empty.

https://github.com/sqlparser-rs/sqlparser-rs/blob/b0b62887a2bbc464642a17e590b259411bf1c11d/src/ast/ddl.rs#L318-L323

This is correctly implemented here:

https://github.com/sqlparser-rs/sqlparser-rs/blob/b0b62887a2bbc464642a17e590b259411bf1c11d/src/parser/mod.rs#L4949-L4953

Thus, when `sequence_options` is `Some(_)` in

```rust
AlterColumnOperation::AddGenerated {
    generated_as,
    sequence_options
}
```
we expect `Display` to print at least the brackets regardless of the contents. 

However, the implementation for display also checks whether the option list is empty.

https://github.com/sqlparser-rs/sqlparser-rs/blob/b0b62887a2bbc464642a17e590b259411bf1c11d/src/ast/ddl.rs#L355-L368

This plays with an issue in `SequenceOption` parsing, where the `Min` and `Max` fields are given as `empty` regardless of whether they are there:

https://github.com/sqlparser-rs/sqlparser-rs/blob/b0b62887a2bbc464642a17e590b259411bf1c11d/src/parser/mod.rs#L8565-L8583

The tag `empty` is used to remove them in display.

In conclusion, if parentheses exist, then `Min` and `Max` options always make it to the options list therefore Display passes checks and parentheses are printed, then the two fields are not printed because the Display for an empty tag is an empty string. In the case where the parentheses don't exist, the check correctly skips that.

In my opinion, the two fields shouldn't be included in the opt list, and the opt list shouldn't be checked for emptiness.

I will open a PR for this.

PartiallyTyped:
Changes:
1. Display for alter col with sequence options now only relies on presence of Sequence option lists. Control for parentheses is done via Some/None check.
2. Changed Minvalue and Maxvalue in Sequence Option to use Option<expr> instead of ternary some, none, empty. Presence is checked through the option list.

Fixes #1105

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -354,17 +354,13 @@ impl fmt::Display for AlterColumnOperation {
 
                 write!(f, "ADD GENERATED{generated_as} AS IDENTITY",)?;
                 if let Some(options) = sequence_options {
-                    if !options.is_empty() {
-                        write!(f, " (")?;
-                    }
+                    write!(f, " (")?;
 
                     for sequence_option in options {
                         write!(f, "{sequence_option}")?;
                     }
 
-                    if !options.is_empty() {
-                        write!(f, " )")?;
-                    }
+                    write!(f, " )")?;
                 }
                 Ok(())
             }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3734,8 +3734,8 @@ impl fmt::Display for Statement {
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum SequenceOptions {
     IncrementBy(Expr, bool),
-    MinValue(MinMaxValue),
-    MaxValue(MinMaxValue),
+    MinValue(Option<Expr>),
+    MaxValue(Option<Expr>),
     StartWith(Expr, bool),
     Cache(Expr),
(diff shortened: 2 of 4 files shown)


alamb (commented):
Thanks @PartiallyTyped  - likewise this code looks good to me. 

Is there any way to add a test for this behavior? Otherwise it could get broken during a refactor and we wouldn't notice

PartiallyTyped:
@alamb I will add some tests targeting the generated ast to ensure the semantics.

alamb (approved):
Thank you @PartiallyTyped

alamb:
@PartiallyTyped  can you please resolve the conflicts in this PR so I can merge it?

Thank you 🙏

PR #1107, opened 2024-01-22
Add `Qrlew` as a user in README.md

ngrislain:
# Qrlew

## The open-source DP layer for SQL

[Qrlew](https://qrlew.github.io/) (/ˈkɝlu/) is the [open source library](https://github.com/Qrlew) that rewrites SQL queries into privacy-preserving variants using [Differential Privacy (DP)](https://en.wikipedia.org/wiki/Differential_privacy).

Use Qrlew if you want to bring privacy guarantees to your SQL pipelines. It is:
- **SQL-to-SQL**: Qrlew turns SQL queries into differentially-private SQL queries that can be executed at scale on many SQL datastore, in many SQL dialects.
- **Feature-rich**: Qrlew covers the broadest range of SQL queries, including JOIN and nested queries
- **Privacy-optimized**: Qrlew keeps track of tight bounds and ranges throughout each step, minimizing the amount of noise needed to achieve differential privacy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ $ cargo run --features json_example --example cli FILENAME.sql [--dialectname]
 ## Users
 
 This parser is currently being used by the [DataFusion] query engine,
-[LocustDB], [Ballista], [GlueSQL], [Opteryx], [PRQL], and [JumpWire].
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thanks @ngrislain  -- always great to have another user listed!

PR #1109, opened 2024-01-23
Support for unquoted hyphenated identifiers on bigquery

jmhain:
This adds support for unquoted hyphenated identifiers on bigquery, e.g. `select * from foo-bar f join baz-qux b on f.id = b.id`. These are only allowed in contexts where they are unambiguous, such as in a `FROM` or `JOIN` clause. (There may be other cases where they are allowed, I only tested these two.)

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -569,7 +569,7 @@ impl<'a> Parser<'a> {
             FlushType::OptimizerCosts
         } else if self.parse_keywords(&[Keyword::RELAY, Keyword::LOGS]) {
             if self.parse_keywords(&[Keyword::FOR, Keyword::CHANNEL]) {
-                channel = Some(self.parse_object_name().unwrap().to_string());
+                channel = Some(self.parse_object_name(false).unwrap().to_string());
             }
             FlushType::RelayLogs
         } else if self.parse_keywords(&[Keyword::SLOW, Keyword::LOGS]) {
@@ -593,7 +593,7 @@ impl<'a> Parser<'a> {
                         }
                         Keyword::NoKeyword => {
                             self.prev_token();
-                            tables = self.parse_comma_separated(Parser::parse_object_name)?;
+                            tables = self.parse_comma_separated(|p| p.parse_object_name(false))?;
                         }
                         _ => {}
                     },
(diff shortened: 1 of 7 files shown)


src/parser/mod.rs
```diff
@@ -5867,10 +5888,10 @@ impl<'a> Parser<'a> {
 
     /// Parse a possibly qualified, possibly quoted identifier, e.g.
     /// `foo` or `myschema."table"
-    pub fn parse_object_name(&mut self) -> Result<ObjectName, ParserError> {
+    pub fn parse_object_name(&mut self, in_table_clause: bool) -> Result<ObjectName, ParserError> {
```
alamb, line 5891: Can you please add documentation to this function explaining the new paramter and some context of what it is used for?

tests/sqlparser_bigquery.rs
```diff
@@ -878,6 +878,27 @@ fn parse_table_identifiers() {
...
+    test_table_ident_err("foo-`bar`");
+    test_table_ident_err("`foo`-bar");
+    test_table_ident_err("foo-123a");
+    test_table_ident_err("foo - bar");
+    test_table_ident_err("123-bar");
+    test_table_ident_err("bar-");
+}
+
+#[test]
+fn parse_hyphenated_table_identifiers() {
```
alamb, line 897: Can you also include some whole query level error tests as well

like that you can't do something like this (or at least document in tests that it parses to a subtraction?

```
select foo-bar.* from foo-bar
select foo-bar.x
```

alamb (commented):
Thank you for the contribution @jmhain  -- this is wild

> This adds support for unquoted hyphenated identifiers on bigquery, e.g. select * from foo-bar f join baz-qux b on f.id = b.id. These are only allowed in contexts where they are unambiguous, such as in a FROM or JOIN clause. (There may be other cases where they are allowed, I only tested these two.)

Wow the craziness of sql dialects never ceases to amaze me. Big query, why? LOL

jmhain:
@alamb Thanks for the prompt feedback! I pushed a couple commits incorporating your suggestions.

> Wow the craziness of sql dialects never ceases to amaze me. Big query, why? LOL

lol this was my reaction as well

alamb (approved):

alamb:
Thanks again @jmhain

PR #1110, opened 2024-01-24
Fix :start and :end json accesses on snowflake

jmhain:


tests/sqlparser_snowflake.rs
```diff
@@ -229,6 +229,24 @@ fn parse_json_using_colon() {
     );
 
     snowflake().one_statement_parses_to("SELECT a:b::int FROM t", "SELECT CAST(a:b AS INT) FROM t");
+
+    let sql = "SELECT a:start, a:end FROM t";
```
alamb, line 233: I looked for a few minutes and couldn't actually find any mention of this syntax in the documentation. 🤯  but I trust you it exists

alamb (approved):
Thank you @jmhain

tests/sqlparser_snowflake.rs
jmhain: This is just the [regular `:` operator](https://docs.snowflake.com/en/sql-reference/data-types-semistructured#label-accessing-object-by-key) for accessing a field within an `OBJECT` / `VARIANT`. `start` and `end` aren't special here, they're just field names that SF permits but we were failing to parse. Maybe it would have been clearer if I used `SELECT a:start, a:end FROM (select parse_json('{"start":0,"end":1}') a)` for the test.

alamb: 👍  Got it

Issue #1112, opened 2024-01-25
Release version `0.43.0`


alamb:
In my haste to release 0.42.0, I accidentally also released 0.43.0 with no changes 🤦 https://crates.io/crates/sqlparser/0.43.0

alamb:
https://github.com/sqlparser-rs/sqlparser-rs/issues/1114 tracks version 44

PR #1115, opened 2024-01-25
fix release instructions

alamb:


alamb:
This is the source of accidentally releasing `0.43.0` #1112

PR #1118, opened 2024-01-30
BigQuery: support of CREATE VIEW IF NOT EXISTS

7phs:
There is support for `CREATE VIEW IF NOT EXISTS` for general and sqlite dialects (only).

This PR adds a [support it for BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_view_statement)


alamb (approved):
Thank you for the contribution @7phs

PR #1119, opened 2024-01-30
feat: add mysql show status statement

invm:
Hi, 

Wanted to add support for `SHOW [GLOBAL | SESSION] STATUS [LIKE 'pattern' | WHERE expr]`

Taken from: [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/show-status.html)


alamb (approved):
Looks good to me -- thank you for the contribution @invm

PR #1122, opened 2024-02-02
Support `DECLARE` syntax for snowflake and bigquery

iffyio:
The current support for `DECLARE` statements are primarily aimed at postgres-compatible syntax which differs significantly for snowflake and bigquery.
This adds support for both dialects.
Generic is left to still parse the postgres-compatible syntax since that should have the most coverage in terms of support.


alamb (approved):
Thank you so much for this contribution @iffyio. I found this PR very well documented, coded and tested. 🏆 

thanks again and sorry for the delay in review. I have very limited bandwidth for this project sadly

iffyio:
@alamb Not a problem at all, yeah I understood your bandwidth for the project so no hurry on my end! Thanks a lot for the review! :pray:

PR #1062, opened 2023-12-04
add support for JSON_TABLE

lovasoa:
see https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html
and
https://modern-sql.com/blog/2017-06/whats-new-in-sql-2016#json_table

Fixes https://github.com/lovasoa/SQLpage/discussions/146

Closes https://github.com/sqlparser-rs/sqlparser-rs/issues/1019

PR #1123, opened 2024-02-05
Restrain JSON_TABLE table function parsing to MySqlDialect and AnsiDialect

viirya:
`JSON_TABLE` is introduced in SQL:2016 and added into this parser by #1062. But it is not actually implemented by most SQL engines. Having it as general parser support may cause issue like https://github.com/apache/arrow-datafusion/issues/9122. Maybe we should restrain `JSON_TABLE` to `MySqlDialect` for now.

Btw, the PR #1062 added the support also only wrote the test for mysql (`tests/sqlparser_mysql.rs`).

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -23,6 +23,7 @@ use core::fmt::{self, Display};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+use sqlparser::keywords::{ALL_KEYWORDS, ALL_KEYWORDS_INDEX};
 #[cfg(feature = "visitor")]
 use sqlparser_derive::{Visit, VisitMut};
 
@@ -54,6 +55,7 @@ pub use self::value::{
 use crate::ast::helpers::stmt_data_loading::{
     DataLoadingOptions, StageLoadSelectItem, StageParamsObject,
 };
+use crate::keywords::Keyword;
 #[cfg(feature = "visitor")]
 pub use visitor::*;
 
@@ -140,6 +142,21 @@ impl Ident {
             value: value.into(),
             quote_style: Some(quote),
         }
+    }
+
+    /// Returns the defined `Keyword` enum for this identifier if it is a keyword.
+    pub fn find_keyword(&self) -> Option<Keyword> {
+        ALL_KEYWORDS
+            .iter()
+            .enumerate()
+            .find_map(|(idx, &kw)| {
+                if kw.to_string().to_uppercase() == self.value.to_uppercase() {
+                    Some(idx)
+                } else {
+                    None
+                }
+            })
+            .map(|idx| ALL_KEYWORDS_INDEX[idx])
     }
 }
 
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -145,7 +145,7 @@ impl Ident {
     }
 
     /// If this identifier is also a keyword, return the corresponding [`Keyword`].
-    /// 
+    ///
     /// For example even though `AVRO` is a keyword, it can also be used as an
     /// identifier for a column, such as `SELECT avro FROM my_table`.
     pub fn find_keyword(&self) -> Option<Keyword> {
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -144,7 +144,10 @@ impl Ident {
         }
     }
 
-    /// Returns the defined `Keyword` enum for this identifier if it is a keyword.
+    /// If this identifier is also a keyword, return the corresponding [`Keyword`].
+    /// 
+    /// For example even though `AVRO` is a keyword, it can also be used as an
+    /// identifier for a column, such as `SELECT avro FROM my_table`.
     pub fn find_keyword(&self) -> Option<Keyword> {
         ALL_KEYWORDS
             .iter()
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7548,12 +7548,18 @@ impl<'a> Parser<'a> {
             let loc = self.peek_token().location;
             let name = self.parse_object_name(true)?;
 
-            for ident in &name.0 {
-                if ident.quote_style.is_none() && ident.find_keyword().is_some() {
-                    return parser_err!(
-                        "Cannot specify a keyword as identifier for table factor",
-                        loc
-                    );
+            // Prevents using keywords as identifiers for table factor in ANSI mode
+            if dialect_of!(self is AnsiDialect) {
+                for ident in &name.0 {
+                    if ident.quote_style.is_none() && ident.find_keyword().is_some() {
+                        return parser_err!(
+                            format!(
+                                "Cannot specify a keyword as identifier for table factor: {}",
+                                ident.value
+                            ),
+                            loc
+                        );
+                    }
                 }
             }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7506,7 +7506,9 @@ impl<'a> Parser<'a> {
                 with_offset,
                 with_offset_alias,
             })
-        } else if self.parse_keyword(Keyword::JSON_TABLE) {
+        } else if dialect_of!(self is MySqlDialect)
+            && self.parse_keyword(Keyword::JSON_TABLE)
+        {
             self.expect_token(&Token::LParen)?;
             let json_expr = self.parse_expr()?;
             self.expect_token(&Token::Comma)?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2772,7 +2772,7 @@ impl<'a> Parser<'a> {
     }
 
     /// If the current token is the `expected` keyword followed by
-    /// specified tokens,  consume them and returns true.
+    /// specified tokens, consume them and returns true.
     /// Otherwise, no tokens are consumed and returns false.
     pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
         let index = self.index;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2771,6 +2771,27 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// If the current token is the `expected` keyword followed by
+    /// specified tokens, and consume them and returns true.
+    /// Otherwise, no tokens are consumed and returns false.
+    pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        let index = self.index;
+        match self.peek_token().token {
+            Token::Word(w) if expected == w.keyword => {
+                self.next_token();
+                for token in tokens {
+                    if !self.consume_token(token) {
+                        // reset index and return immediately
+                        self.index = index;
+                        return false;
+                    }
+                }
+                true
+            }
+            _ => false,
+        }
+    }
+
     /// If the current and subsequent tokens exactly match the `keywords`
     /// sequence, consume them and returns true. Otherwise, no tokens are
     /// consumed and returns false
@@ -7507,9 +7528,8 @@ impl<'a> Parser<'a> {
                 with_offset_alias,
             })
         } else if dialect_of!(self is MySqlDialect | AnsiDialect)
-            && self.parse_keyword(Keyword::JSON_TABLE)
+            && self.parse_keyword_with_tokens(Keyword::JSON_TABLE, &[Token::LParen])
         {
-            self.expect_token(&Token::LParen)?;
             let json_expr = self.parse_expr()?;
             self.expect_token(&Token::Comma)?;
             let json_path = self.parse_value()?;
@@ -7526,7 +7546,19 @@ impl<'a> Parser<'a> {
                 alias,
             })
         } else {
+            let loc = self.peek_token().location;
             let name = self.parse_object_name(true)?;
+
+            for ident in &name.0 {
+                if ident.quote_style.is_none() {
+                    if ident.find_keyword().is_some() {
+                        return parser_err!(
+                            "Cannot specify a reserved keyword as identifier of table factor",
+                            loc
+                        );
+                    }
+                }
+            }
 
             let partitions: Vec<Ident> = if dialect_of!(self is MySqlDialect | GenericDialect)
                 && self.parse_keyword(Keyword::PARTITION)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7506,7 +7506,9 @@ impl<'a> Parser<'a> {
                 with_offset,
                 with_offset_alias,
             })
-        } else if dialect_of!(self is MySqlDialect | AnsiDialect) && self.parse_keyword(Keyword::JSON_TABLE) {
+        } else if dialect_of!(self is MySqlDialect | AnsiDialect)
+            && self.parse_keyword(Keyword::JSON_TABLE)
+        {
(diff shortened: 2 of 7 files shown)


src/parser/mod.rs
```diff
@@ -7506,7 +7506,9 @@ impl<'a> Parser<'a> {
                 with_offset,
                 with_offset_alias,
             })
-        } else if self.parse_keyword(Keyword::JSON_TABLE) {
+        } else if dialect_of!(self is MySqlDialect)
+            && self.parse_keyword(Keyword::JSON_TABLE)
+        {
```
viirya, line 7511: The only change is here. Other changes are from rustfmt.

Jefffrey (approved):
Maybe add a simple test to generic SQL dialect to ensure this behaviour?

src/parser/mod.rs
```diff
@@ -7506,7 +7506,7 @@ impl<'a> Parser<'a> {
                 with_offset,
                 with_offset_alias,
             })
-        } else if self.parse_keyword(Keyword::JSON_TABLE) {
+        } else if dialect_of!(self is MySqlDialect) && self.parse_keyword(Keyword::JSON_TABLE) {
```
alamb, line 7509: this is the actual change, right? It seems like the changes to derive/src/lib.rs are unrelated

alamb (approved):
Looks good to me -- thank you @viirya 

Is there any chance you can revert the changes to derive/src/lib.rs in this PR?

cc @lovasoa

src/parser/mod.rs
lovasoa: This sounds a little bit adhoc. Shouldn't this be a method on dialect ?

lovasoa: JSON_TABLE seems to be implemented on oracle too, for instance: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html#GUID-3C8E63B5-0B94-4E86-A2D3-3D4831B67C62

lovasoa: And on DB2: https://www.ibm.com/docs/en/i/7.3?topic=data-using-json-table

lovasoa (changes requested):

lovasoa:
`JSON_TABLE` is part of the SQL:2016 standard: https://modern-sql.com/blog/2017-06/whats-new-in-sql-2016
It is implemented in Db2, MariaDB, MySQL and Oracle.

I think it's ok to accept `json_table` as a table name for specific dialects where it has been tested to be accepted unquoted, but the default behavior should probably be to parse it as the `json_table` function, not as a table name.

In simpler cases, we may even be able to distinguish between the unquoted table name and the function call directly during parsing, and make everyone happy.

src/parser/mod.rs
viirya: @alamb Yea, `derive/src/lib.rs` was changed by rustfmt. I restored it manually.

viirya: We don't have dialects for Oracle and DB2.

viirya: > This sounds a little bit adhoc. Shouldn't this be a method on dialect ?

This can be discussed separately, but this syntax follows existing ones.

src/parser/mod.rs
```diff
@@ -7506,7 +7506,7 @@ impl<'a> Parser<'a> {
                 with_offset,
                 with_offset_alias,
             })
-        } else if self.parse_keyword(Keyword::JSON_TABLE) {
+        } else if dialect_of!(self is MySqlDialect | AnsiDialect) && self.parse_keyword(Keyword::JSON_TABLE) {
```
viirya, line 7509: I suppose `AnsiDialect` is for SQL standard, though it doesn't specify which standard it follows. Seems it makes sense to add it in supported dialect.

viirya: It could be the default behavior for dialects which are likely to follow the standard to support the table function.

lovasoa, line 7509: I still don't think the dialect identity should be hardcoded. This should probably be a dialect.supports_json_table. And it should be true in the generic dialect.

viirya: As there are only very few dialects supporting this, I think it makes more sense to not add it into `GenericDialect` currently.

alamb: I agree it would be nicer to do if `self.dialect.supports_json_table()` instead. However, I think we could also do this as a follow on PR as well

viirya: I don't see why it should be true in generic dialect. `GenericDialect` is not strictly SQL standard dialect (`AnsiDialect` is for the purpose). It is for syntax that most dialects use. Currently the table function is only supported by very few dialects.

viirya: > I agree it would be nicer to do if `self.dialect.supports_json_table()` instead. However, I think we could also do this as a follow on PR as well

Agreed. As I mentioned early, this can be discussed separately. Although I wonder if it will result in many similar functions.

alamb: > Although I wonder if it will result in many similar functions.

Yes I think it would. The codebase has a mix of styles currently -- both `dialect_of!` and trait methods

alamb:
I believe CI is failing due to a new rust version being released (and thus new clippy lints). I will fix that separately

alamb (approved):
I think this PR looks good and follows the existing conventions.

I think @viirya  has addressed all comments.  @lovasoa  please let us know if you disagree.

alamb:
I merged up from main to pick up https://github.com/sqlparser-rs/sqlparser-rs/pull/1130

alamb:
I think @lovasoa 's comment on https://github.com/apache/arrow-datafusion/issues/9122#issuecomment-1934215854 is worth considering (should we perhaps instead make the error message more helpful?)

tests/sqlparser_common.rs
```diff
@@ -8407,3 +8407,25 @@ fn test_buffer_reuse() {
...
+
+#[test]
+fn parse_json_table_as_identifier() {
+    let parsed = all_dialects().parse_sql_statements("SELECT * FROM json_table");
+    assert_eq!(
+        ParserError::ParserError(
+            "Cannot specify a reserved keyword as identifier for table factor".to_string()
+        ),
+        parsed.unwrap_err()
+    );
```
viirya, line 8430: @alamb I think this can have a more meaningful message to end users.

tests/sqlparser_common.rs
```diff
@@ -8407,3 +8407,25 @@ fn test_buffer_reuse() {
...
+
+#[test]
+fn parse_json_table_as_identifier() {
+    let parsed = all_dialects().parse_sql_statements("SELECT * FROM json_table");
+    assert_eq!(
+        ParserError::ParserError(
+            "Cannot specify a keyword as identifier for table factor".to_string()
+        ),
+        parsed.unwrap_err()
+    );
```
viirya, line 8430: And, even for the supported dialects, `json_table` cannot be used as table identifier too. Previously it can be.

src/parser/mod.rs
```diff
@@ -2779,7 +2800,6 @@ impl<'a> Parser<'a> {
         let index = self.index;
         for &keyword in keywords {
             if !self.parse_keyword(keyword) {
-                // println!("parse_keywords aborting .. did not find {:?}", keyword);
```
viirya, line 2782: I think this should be removed.

src/parser/mod.rs
```diff
@@ -2771,6 +2771,27 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// If the current token is the `expected` keyword followed by
+    /// specified tokens, and consume them and returns true.
```
alamb, line 2775:
```suggestion
    /// specified tokens,  consume them and returns true.
```

src/parser/mod.rs
```diff
@@ -2771,6 +2771,27 @@ impl<'a> Parser<'a> {
...
+    /// Otherwise, no tokens are consumed and returns false.
+    pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        let index = self.index;
+        match self.peek_token().token {
+            Token::Word(w) if expected == w.keyword => {
+                self.next_token();
```
alamb, line 2781: FWIW I think it would be possible to call `peek_nth_token` here as well to avoid having to reset index. However, this works fine too

https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.peek_nth_token

src/ast/mod.rs
```diff
@@ -141,6 +143,21 @@ impl Ident {
             quote_style: Some(quote),
         }
     }
+
+    /// Returns the defined `Keyword` enum for this identifier if it is a keyword.
```
alamb, line 147:
```suggestion
    /// If this identifier is also a keyword, return the corresponding [`Keyword`].
    /// 
    /// For example even though `AVRO` is a keyword, it can also be used as an
    /// identifier for a column, such as `SELECT avro FROM my_table`.
```

alamb (approved):
Thanks @viirya  -- this looks really nice to me. Thank you for improving the message as well

src/parser/mod.rs
viirya: Good suggestion. Changed to use `peek_nth_token`.

tests/sqlparser_bigquery.rs
```diff
@@ -866,13 +866,13 @@ fn parse_table_identifiers() {
         vec![Ident::with_quote('`', "GROUP"), Ident::new("dataField")],
     );
 
-    // TODO: this should be error
-    // test_table_ident_err("GROUP.dataField");
```
viirya, line 870: This is an error now. So removed the `TODO`.

tests/sqlparser_bigquery.rs
```diff
@@ -866,13 +866,13 @@ fn parse_table_identifiers() {
...
+    test_table_ident_err("GROUP.dataField");
+    test_table_ident_err("abc5.GROUP");
 
     test_table_ident(
-        "abc5.GROUP",
```
viirya, line 873: `GROUP` is a keyword.

alamb (approved):

viirya:
Hmm, there are more tests using keywords as identifiers...

lovasoa:
I don't think banning using keywords as identifiers is a good idea. It will probably break a lot people's applications. Currently, keywords are defined globally in sqlparser, even though the list of supported keywords varies from database from database. Banning everything that is a keyword in one database from use as a keyword globally is probably not a good idea.

I recently had the case with MySQL, which supports `returning` as an identifier: `select 1 as returning` is valid in MySQL. Since `RETURNING` is also a keyword, that would break. 
MS SQL Server also has a lot of keywords that only it uses. `select 1 as open` is valid everywhere, but not in mssql.

viirya:
> I don't think banning using keywords as identifiers is a good idea. It will probably break a lot people's applications. 

I have concerns about this too. But I think this is what you first suggested, no?

lovasoa:
What I was suggesting was just to close https://github.com/apache/arrow-datafusion/issues/9122 as "working as intended" :)

viirya (commented):
Okay. Keyword as identifier is only disallowed with ANSI mode now. Other dialects still can have keywords as identifiers for their custom usage.

I think there are disallowed cases of using keyword as identifier in individual dialects, ie., Each engine has its reserved keyword list.

We probably can consider to enhance dialect behavior regarding this, i.e., to create keyword list per dialect, in the future.

lovasoa:
> Keyword as identifier is only disallowed with ANSI

Does the ansi sql dialect still parse `select 1 as open` correctly, for instance?

viirya:
> Does the ansi sql dialect still parse `select 1 as open` correctly, for instance?

It still parses the sql, although `open` is in keyword list.
Currently it is only disallowed when using it as table identifier.

lovasoa:
Yes, the problem is `select * from open` is valid ansi sql too

lovasoa:
Another example is `select * from view`, which is valid in ansi sql, postgres, sqlite, and mysql and mariadb, but would be refused if this pr were merged. The principle of banning keywords as table name before we have dialect-specific keyword lists **will** make sqlparser start to refuse a lot of valid sql queries it used to accept, even if done just for "ansi sql".

alamb:
So I would like to suggest we take a step back here and review what we are trying to accomplish

What our DataFusion user hit, and I actually hit as well is quite confusing behavior.

Specifically, while it is fine to create a table named `json_table`, like this

DataFusion CLI v35.0.0
❯ create table json_table(x int);
0 rows in set. Query took 0.005 seconds.
```

If you try to query it you get a very non confusing syntax error:

```sql
❯ select * from json_table;  🤔 Invalid statement: sql parser error: Expected (, found: EOF
```

For better or worse I think using `json_table` as a table name is somewhat common.

I think the experience would have been far better if the error had instead looked like:

```sql
❯ select * from json_table;  🤔 Invalid statement: sql parser error: Can not parse json_table clause. Expected (, found: EOF. Hint: to select from a table named json_table, use double quotes "json_table"
```

So my suggestion is:
1. Leave the existing error behavior (as it which is SQL standard, as @lovasoa  points out, however confusing that may be).
1. Improve the error message (which I think is the root cause of the problem)

lovasoa:
There is a third option, that I had suggested above, and that I think would be the best:

 3. whatever the dialect, parse `json_table` as a function in queries if and only if it is followed by `(`.

This way, `select * from json_table` would be accepted, and parsed as an unquoted table name, just as it was before adding support for the json_table function.

alamb:
> I opened an alternative pr to this one here: #1134

Thank you @lovasoa  -- I like #1134 a lot. It would be great to get some other opinions as well

viirya:
> Yes, the problem is, I think `select * from open` is valid ansi sql too

`open` is reserved keyword...

viirya:
> Another example is `select * from view`, which is valid in ansi sql, postgres, sqlite, mysql and mariadb, but would be refused if this pr were merged. The principle of banning keywords as table names before we have dialect-specific keyword lists **will** make sqlparser start to refuse a lot of valid sql queries it used to accept, even if done just for "ansi sql".

`view` is reserved keywords in SQL-92. You simplify the question but the actual situation is more complicated. The fact is reserved keyword list is changed over time and different on different query engines. sqlparser doesn't currently take care about this difference. That is the problem. We don't even state what SQL standard the ansi dialect follows. I agreed that we shouldn't block all keywords as it will break applications. But this is what you suggested at the beginning, right? Let me attach your comment in case you forgot it (https://github.com/apache/arrow-datafusion/issues/9122#issuecomment-1934215854). In the comment, you stated that:

> I think the JSON_TABLE function is part of the SQL standard. So SELECT * FROM json_table is not valid SQL, and datafusion is correct in refusing it.

Doesn't it mean you want to block the SQL syntax which not "valid" SQL?

I know you changed your mind later. 

I just don't like to go forth and back on this PR. So I don't want to follow up this PR. Please move forward with your PR.

lovasoa:
My initial comment was not suggesting to block more keywords. I was saying that the current behavior is reasonable. 

Accepting even more valid syntax is better. 

Blocking syntax that is valid in most databases and currently works in sqlparser is worse.

lovasoa:
Anyway, I think we are all ok with #1134 in the end. 

And your effort was not in vain, it ended up raising important points.

alamb:
Thanks again @viirya and @lovasoa  -- hopefully we can figure out how to improve the communication efficiency going forward -- I agree the solution we arrived at is good, but it took a of back and forth.

viirya:
Thank you @alamb

PR #1125, opened 2024-02-06
Accept table list in truncate

tobyhede:
Parser errors when more than one table in truncate 

```
TRUNCATE users, orders RESTART IDENTITY
```


tobyhede:
Damn it github

PR #1132, opened 2024-02-12
add support for `insert into ... select ... returning ...`

lovasoa:
`insert into test select 1 returning 2` now parses correctly

`select 1 as returning` which is valid only in MySQL and SQL Server and used to parse without error still parses without error (and is now tested)

`select 1 returning`, which is a syntax error in postgres, sqlite and MariaDB, but is valid in MySQL and SQL Server, used to parse without error, and now returns a syntax error.

The `insert into ... select ... returning ...` syntax is quite useful in [SQLPage](https://github.com/lovasoa/SQLpage)


alamb (approved):
Thanks @lovasoa

PR #1134, opened 2024-02-14
accept JSON_TABLE both as an unquoted table name and a table-valued function

lovasoa:
This is an alternative to https://github.com/sqlparser-rs/sqlparser-rs/pull/1123

closes https://github.com/sqlparser-rs/sqlparser-rs/pull/1123
related to https://github.com/apache/arrow-datafusion/issues/9122

see https://github.com/sqlparser-rs/sqlparser-rs/pull/1123#issuecomment-1943607093

This makes sqlparser accept both 

```sql
select * from json_table 
```
And
```sql
select * from json_table(t, '$[*]' COLUMNS (id  INT PATH '$.id'))
```

in all dialects

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -2339,7 +2339,7 @@ fn parse_json_table_is_not_reserved() {
             name: ObjectName(name),
             ..
         } => assert_eq!("JSON_TABLE", name[0].value),
-        _ => unreachable!(),
+        _ => panic!("Expected JSON_TABLE to be parsed as a table name"),
     }
 }
 
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -2332,19 +2332,16 @@ fn test_json() {
 fn parse_json_table_is_not_reserved() {
     // JSON_TABLE is not a reserved keyword in PostgreSQL, even though it is in SQL:2023
     // see: https://en.wikipedia.org/wiki/List_of_SQL_reserved_words
-    match pg().verified_only_select("SELECT * FROM JSON_TABLE") {
-        Select { from, .. } => {
-            assert_eq!(1, from.len());
-            match &from[0].relation {
-                TableFactor::Table {
-                    name: ObjectName(name),
-                    ..
-                } => {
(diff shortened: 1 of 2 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -2328,6 +2328,21 @@ fn test_json() {
     );
 }
 
+#[test]
+fn parse_json_table_is_not_reserved() {
+    // JSON_TABLE is not a reserved keyword in PostgreSQL, even though it is in SQL:2023
+    // see: https://en.wikipedia.org/wiki/List_of_SQL_reserved_words
+    let Select { from, .. } = pg().verified_only_select("SELECT * FROM JSON_TABLE");
```
alamb, line 2335: shall we also test generic ?

alamb (approved):
Thank you @lovasoa  -- I like this PR a lot

alamb:
cc @viirya

tests/sqlparser_postgres.rs
lovasoa: ok, I added generic

viirya:
Okay for me. Although I remember this what you are against to at the beginning...

lovasoa:
@viirya , this PR implements what I had suggested in https://github.com/apache/arrow-datafusion/issues/9122#issuecomment-1936681273

lovasoa:
Yes, my initial comment was about just ignoring the issue. This implements something even better that should make everyone happy.

viirya (commented):
Are you sure `select * from json_table` is valid in the engines which support json_table table function?

It is a reserved keyword in MySQL: https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-J

lovasoa:
Yes, as stated above, this PR makes the parser accept BOTH the table identifier and the function in all dialects. I think this is a reasonable behavior.

lovasoa:
@alamb : Yes, databases implement either the unquoted table reference syntax, or the table valued function syntax. But sqlparser now implements both.

alamb:
Thanks again @lovasoa  and @viirya

lovasoa:
:+1:

PR #1136, opened 2024-02-16
Support `row_alias` and `col_aliases` in `INSERT` statement for mysql and generic dialects

emin100:
Example: INSERT INTO `table` (`date`) VALUES (2024-01-01) AS `alias` (`mek`)


src/ast/mod.rs
```diff
@@ -1536,6 +1536,10 @@ pub enum Statement {
         replace_into: bool,
         /// Only for mysql
         priority: Option<MysqlInsertPriority>,
+        /// Only for mysql
+        as_table: Option<ObjectName>,
```
alamb, line 1540: I think you may need to add an annotation here so it visits the relation name

```suggestion
        #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
        as_table: Option<ObjectName>,
```

src/ast/mod.rs
```diff
@@ -1536,6 +1536,10 @@ pub enum Statement {
         replace_into: bool,
         /// Only for mysql
         priority: Option<MysqlInsertPriority>,
+        /// Only for mysql
```
alamb, line 1539: Could you please add an example to this docstring with an example of the supported syntax?

alamb: Also, it appears that mysql calls these fields `row_alias` and `col_alias`
Screenshot 2024-02-28 at 5 36 08 PM

https://dev.mysql.com/doc/refman/8.0/en/insert.html

What do you think about using the same names?

tests/sqlparser_mysql.rs
```diff
@@ -1327,6 +1327,55 @@ fn parse_priority_insert() {
     }
 }
 
+#[test]
+fn parse_insert_as() {
+    let sql = r"INSERT INTO `table` (`date`) VALUES ('2024-01-01') AS `alias` (`mek`)";
```
alamb, line 1332: I think it would also be worth testing
1.  the case when there is more than one column alias
2. the case when there is only a row_alias and not any column aliases

alamb (commented):
Thank you for the contribution @emin100 . I left some comments. Let me know what you think

alamb:
I apologize for the delay in reviewing

alamb:

src/ast/mod.rs
emin100: It is a good idea. The field names were changed.

tests/sqlparser_mysql.rs
emin100: Tests were fixed.

src/ast/mod.rs
emin100: The annotation was added.

emin100:
@alamb  Thanks for comments.

src/ast/mod.rs
```diff
@@ -1759,6 +1759,11 @@ pub enum Statement {
...
+        #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
+        row_alias: ObjectName,
```
alamb, line 1764: I think the typical pattern is to use `Option<ObjectName>` for optional fields. 

Perhaps the issue here is that we can only use `visit_relation` with a `ObjectName` 🤔 

In that case we could make a new structure like

```
struct InsertAliases { 
  row_alias: ObjectName,
  col_aliases: Option<Vec<Ident>>,
}
```

Or maybe just avoid the `visitor` call 🤔  (and could file a ticket with a bug enhancment)

tests/sqlparser_mysql.rs
```diff
@@ -1330,6 +1330,99 @@ fn parse_priority_insert() {
...
+                    locks: vec![],
+                    for_clause: None,
+                })),
+                source
+            );
+        }
+        _ => unreachable!(),
+    }
+
+    let sql = r"INSERT INTO `table` (`id`, `date`) VALUES (1, '2024-01-01') AS `alias` (`mek_id`, `mek_date`)";
```
alamb, line 1375: Can you also add a test with empty list (which maybe should error)

```
    let sql = r"INSERT INTO `table` (`id`, `date`) VALUES (1, '2024-01-01') AS `alias` ()";
```

alamb (commented):
Thank you @emin100  -- I think this is really close. There is just some strangeness with the row_name not being an `Option` which I think we should sort out

tests/sqlparser_mysql.rs
emin100: New error test was added

src/ast/mod.rs
emin100: The issue was visit_relation with a ObjectName. I think, InsertAliases struct is best solution. I implemented the solution.

src/ast/mod.rs
```diff
@@ -4194,6 +4207,14 @@ pub enum OnInsert {
     OnConflict(OnConflict),
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
```
alamb, line 4212: nice

alamb (approved):
Thank you @emin100  -- sorry it took a few back and forths but I think this PR looks really nice now. Thanks again

PR #1138, opened 2024-02-16
Support for CONNECT BY

jmhain:
This adds support for the [`CONNECT BY`](https://docs.snowflake.com/en/sql-reference/constructs/connect-by) operator, supported by Snowflake, Redshift, and MS SQL.

A bit of awkwardness is the `PRIOR` expressions used to refer to the upper level in the `CONNECT BY` relationships. These expressions are only valid in this context, and `prior` is a perfectly valid identifier elsewhere. Trying to thread that context through the call stack was resulting in a massive diff, so I added some state to the parser to avoid that. Happy to change this is reviewers are not fond of that approach.


src/ast/query.rs
```diff
@@ -92,6 +92,8 @@ impl fmt::Display for Query {
 pub enum SetExpr {
     /// Restricted SELECT .. FROM .. HAVING (no ORDER BY or set operations)
     Select(Box<Select>),
+    /// SELECT .. FROM .. STARTING WITH .. CONNECT BY
+    ConnectBy(ConnectBy),
```
alamb, line 96: Your rationale from the PR description is quite well stated and understandable and I think would be a good addition to the code

```suggestion
    /// 
    /// Note: `ConnectBy` is variant of `SetExpr`, rather than `TableFactor` or a field on `Select` 
    /// because, despite being part of a `SELECT .. FROM` clause, `CONNECT BY` does not compose
    /// with other table factors (e.g. `PIVOT`) in a `FROM`, nor does it permit most options usually available
    ///  in a `SELECT` (e.g. adding a `WHERE` clause in a `SELECT .. CONNECT BY` yields a parse error 
    /// from Snowflake).
    ConnectBy(ConnectBy),
```

src/parser/mod.rs
```diff
@@ -6936,6 +6965,48 @@ impl<'a> Parser<'a> {
             having,
             named_window: named_windows,
             qualify,
+        })))
+    }
+
+    fn with_state<T, F>(&mut self, state: ParserState, mut f: F) -> Result<T, ParserError>
```
alamb, line 6971:
```suggestion
    /// Invoke `f` after first setting the parser's `ParserState` to `state`. 
    ///
    /// Upon return, restores the parser's state to what it started at
    fn with_state<T, F>(&mut self, state: ParserState, mut f: F) -> Result<T, ParserError>
```

tests/sqlparser_snowflake.rs
```diff
@@ -1173,3 +1173,88 @@ fn parse_top() {
...
+    // PRIOR expressions are only valid within a CONNECT BY, and the the token
+    // `prior` is valid as an identifier anywhere else.
+    assert_eq!(
+        snowflake_and_generic()
+            .verified_only_select("SELECT prior FROM some_table")
+            .projection,
+        vec![SelectItem::UnnamedExpr(Expr::Identifier(Ident::new(
+            "prior"
+        )))]
+    );
```
alamb, line 1259: Can you please add some negative cases as well. Perhaps using your rationale in the description:

1. try to compose with other table factors (e.g. PIVOT) in a FROM,
2. Some options usually available in a SELECT (e.g. adding DISTINCT or WHERE clause in a SELECT .. CONNECT BY)

alamb (commented):
Thank you very much for this contribution @jmhain  -- I am truly sorry for the delay in review. 

I think the approach taken by this PR looks good to me. In my opinion, all that is needed prior to merge is:
1. Some additional comments (I left some suggestions)
2. A few negative test cases

Thanks again for such an easy to read and review PR

alamb:

jmhain:
@alamb Many thanks for the review feedback, I greatly appreciate it! And sorry for the delay in getting to this. So I learned that MS SQL and Redshift both also support `CONNECT BY`, and I did some more testing and found that my approach before was really only well suited to Snowflake, so I reworked this significantly. I also added some more tests. Thanks again!

tests/sqlparser_snowflake.rs
jmhain: Most of these cases ended up being supported on other dialects, so they're now allowed.

jmhain:
Just resolved conflicts on this again cc @alamb

alamb (approved):
Thanks again @jmhain

alamb:
Looks like there is a conflict (perhaps with one of your other PRs 😆 ) -- if you can resolve that I'll merge this PR in.

Thanks again

jmhain:
@alamb Done, thank you!

alamb:
🚀

PR #1140, opened 2024-02-18
Fix panic for REPLACE

jjbayer:
Bubble up an error when `parse_insert` fails in `parse_replace`.


alamb (approved):
Thank you very much for the fix and the well commented tests and code @jjbayer

PR #1141, opened 2024-02-18
Add `parse_keyword_with_tokens` for paring  keyword and tokens combination

viirya:
This is a simple follow up to #1134 to add a function `parse_keyword_with_tokens` to deal with the case that the parser expects keyword + token combination, instead of a mix of multiple peek_token, expect_keyword, expect_token calls.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2780,6 +2780,27 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// If the current token is the `expected` keyword followed by
+    /// specified tokens, consume them and returns true.
+    /// Otherwise, no tokens are consumed and returns false.
+    pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        match self.peek_token().token {
+            Token::Word(w) if expected == w.keyword => {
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -2780,6 +2780,27 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// If the current token is the `expected` keyword followed by
+    /// specified tokens, consume them and returns true.
+    /// Otherwise, no tokens are consumed and returns false.
+    pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        match self.peek_token().token {
+            Token::Word(w) if expected == w.keyword => {
+                for (idx, token) in tokens.iter().enumerate() {
+                    if self.peek_nth_token(idx + 1).token != *token {
```
jmhain, line 2790: This is probably a non-issue since `tokens` should always be small, but it seems worth noting that this is unnecessarily quadratic over the length of `tokens` due to the loop in `peek_nth_token`.

viirya: Thanks. Added a note for that.

viirya:
Thank you @alamb @jmhain

PR #1142, opened 2024-02-21
Add ParadeDB to list of known users

philippemnoel:
Hi team,

We're big users of `sqlparser-rs` over at ParadeDB. I was reading the README, saw the list of known users, and figured it might be worth adding ourselves to it. We'll likely make contributions down the line as we're starting to face some limitations.


alamb (approved):
Thanks @philippemnoel

PR #1144, opened 2024-02-22
feat: add support for duckdb style "FROM <tbl>" statements

universalmind303:
allows for parsing of duckdb's `from` queries

closes https://github.com/sqlparser-rs/sqlparser-rs/issues/1143

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -516,6 +516,12 @@ impl<'a> Parser<'a> {
                 Keyword::MERGE => Ok(self.parse_merge()?),
                 // `PRAGMA` is sqlite specific https://www.sqlite.org/pragma.html
                 Keyword::PRAGMA => Ok(self.parse_pragma()?),
+
+                // Duckdb style `FROM <table>` or `FROM tbl SELECT ...`
+                Keyword::FROM if dialect_of!(self is DuckDbDialect | GenericDialect) => {
+                    self.prev_token();
+                    Ok(Statement::Query(Box::new(self.parse_query()?)))
+                }
                 // `INSTALL` is duckdb specific https://duckdb.org/docs/extensions/overview
                 Keyword::INSTALL if dialect_of!(self is DuckDbDialect | GenericDialect) => {
                     Ok(self.parse_install()?)
@@ -6458,10 +6464,8 @@ impl<'a> Parser<'a> {
         } else {
             None
         };
-
         if self.parse_keyword(Keyword::INSERT) {
             let insert = self.parse_insert()?;
-
             Ok(Query {
                 with,
                 body: Box::new(SetExpr::Insert(insert)),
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -6722,6 +6726,8 @@ impl<'a> Parser<'a> {
             SetExpr::Values(self.parse_values(is_mysql)?)
         } else if self.parse_keyword(Keyword::TABLE) {
             SetExpr::Table(Box::new(self.parse_as_table()?))
+        } else if self.parse_keyword(Keyword::FROM) {
```
universalmind303, line 6729: I have no idea why, but adding this branch causes SO in the recursion tests such as `parse_deeply_nested_parens_hits_recursion_limits`

universalmind303: even if the branch is never met, the presence of it is causing SO.

universalmind303: maybe @alamb, @andygrove @nickolay knows why adding this condition causes a SO.

alamb: This branch probably makes the stack frame size increase in debug builds as it adds some new local variables. I think we could fix the test by decreasing the recursion limit

tests/sqlparser_duckdb.rs
```diff
@@ -244,3 +246,188 @@ fn test_duckdb_load_extension() {
...
+        qualify: None,
+        from_before_select: true,
+    };
+    assert_eq!(stmt, expected);
+}
+
+#[test]
+fn test_duckdb_from_statement_with_filter_and_select() {
+    let stmt = duckdb().verified_only_select("FROM t1 SELECT b WHERE a = 1");
+    println!("{:?}", stmt);
```
alamb, line 333: left over?

tests/sqlparser_duckdb.rs
```diff
@@ -244,3 +246,188 @@ fn test_duckdb_load_extension() {
...
+        qualify: None,
+        from_before_select: true,
+    };
+    assert_eq!(stmt, expected);
+}
+
+#[test]
+fn test_duckdb_from_statement_with_filter() {
+    let stmt = duckdb().verified_only_select("FROM t1 WHERE a = 1");
+    println!("{:?}", stmt);
```
alamb, line 289: left over?

src/parser/mod.rs
```diff
@@ -6936,6 +6942,162 @@ impl<'a> Parser<'a> {
             having,
             named_window: named_windows,
             qualify,
+            from_before_select: false,
+        })
+    }
+
+    /// Parse a duckdb style `FROM` statement without a select.
+    /// assuming the initial `FROM` was already consumed
+    pub fn parse_from_select(&mut self) -> Result<Select, ParserError> {
```
alamb, line 6951: this seems largely copied from the normal select parsing -- is there any way to refactor the code to reduce the duplication

alamb (commented):
Thank you for this contribution @universalmind303  -- I left some comments that hopefully will help us get this PR good to go

alamb:

alamb:
FWIW @Nikita-str  has solved the stack overflow issues in https://github.com/sqlparser-rs/sqlparser-rs/pull/1171 if you wanted to revive this PR.

PR #1145, opened 2024-02-22
Support for `(+)` outer join syntax

jmhain:
Snowflake supports a peculiar alternative syntax for outer joins using the `(+)` operator. For example, the following two queries are logically equivalent:

```sql
SELECT t1.c1, t2.c2
FROM t1, t2
WHERE t1.c1 = t2.c2 (+);
```

```sql
SELECT t1.c1, t2.c2
FROM t1 LEFT OUTER JOIN t2
        ON t1.c1 = t2.c2;
```

This adds support to the parser for this syntax. I've only tested on Snowflake, but a web search suggested that MS SQL also supports this syntax, so I permitted it on that dialect as well.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -713,6 +713,21 @@ pub enum Expr {
     /// Qualified wildcard, e.g. `alias.*` or `schema.table.*`.
     /// (Same caveats apply to `QualifiedWildcard` as to `Wildcard`.)
     QualifiedWildcard(ObjectName),
+    /// Some dialects support an older syntax for outer joins where columns are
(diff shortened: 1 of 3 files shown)


alamb:
> Snowflake supports a peculiar alternative syntax for outer joins using the `(+)` operator. For example, the following two queries are logically equivalent:

I recognize that syntax from Oracle as well https://blogs.oracle.com/optimizer/post/outerjoins-in-oracle 👍

alamb (approved):
Looks great to me -- thank you very much @jmhain . The comments and tests are 💯

PR #1148, opened 2024-02-28
ISSUE-1147: Add support for MATERIALIZED CTEs

ReppCodes:
In Postgres, the [NOT] MATERIALIZED keywords can be applied when creating CTEs. Add support for them.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -376,7 +376,36 @@ impl fmt::Display for With {
     }
 }
 
-/// A single CTE (used after `WITH`): `alias [(col1, col2, ...)] AS ( query )`
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CteAsMaterialized {
+    /// The WITH statement does not specify MATERIALIZED behavior
+    Default,
+    /// The WITH statement  specifies AS MATERIALIZED behavior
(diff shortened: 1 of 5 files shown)


src/ast/query.rs
```diff
@@ -376,7 +376,36 @@ impl fmt::Display for With {
     }
 }
 
-/// A single CTE (used after `WITH`): `alias [(col1, col2, ...)] AS ( query )`
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CteAsMaterialized {
+    /// The WITH statement does not specify MATERIALIZED behavior
+    Default,
```
alamb: Most other places in sqlparser-rs use `Option<..>` to represent the case when a clause is not present.

Since I had this branch checked out locally to resolve a merge conflict, I went ahead and made that change as well in 6df536a

alamb (approved):
Thank you @ReppCodes  -- looks good to me

src/ast/query.rs
ReppCodes: Ah for sure!

This route means we should remove the Default enum member though because now it's not useful, and if it does get added at some point it'll print with extra whitespace.

Mind if I add a commit to the PR doing so?

alamb: Sure -- go ahead.  I think I did remove CteAsMaterialized::Default in https://github.com/sqlparser-rs/sqlparser-rs/pull/1148/commits/6df536a782009d29ed37c172abf2be5b17c443ee though

PR #1151, opened 2024-02-29
set_tblproperties

jonathanlehto:
Adding the SET_TBLPROPERTIES command

DataBricks - https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-tblproperties.html#set-tblproperties
Athena - https://docs.aws.amazon.com/athena/latest/ug/alter-table-set-tblproperties.html


jonathanlehto:
Thank you @alamb !

PR #1154, opened 2024-02-29
Support `ALTER TABLE ... SET LOCATION` 

jonathanlehto:
Adding set location to alter table.  It looks like SET is not required explicitly, but that may just be for Athena

Athena SET LOCATION -- https://docs.aws.amazon.com/athena/latest/ug/alter-table-set-location.html


src/ast/mod.rs
```diff
@@ -3000,12 +3001,10 @@ impl fmt::Display for Statement {
...
                 if *external {
-                    write!(
-                        f,
-                        " STORED AS {} LOCATION '{}'",
-                        file_format.as_ref().unwrap(),
-                        location.as_ref().unwrap()
-                    )?;
+                    if file_format.is_some() {
+                        write!(f, " STORED AS {}", file_format.as_ref().unwrap())?;
+                    }
```
alamb, line 3006: I think you can avoid this kind of unwrap by using this syntax:

```suggestion
                    if let Some(file_format) = &file_format {
                        write!(f, " STORED AS {file_format}")?;
                    }
```

As you have done below.

I'll make the change

alamb (approved):
🙏

src/ast/mod.rs
jonathanlehto: Apologies, and, yes, that is better.  Thank you for making that change @alamb !

PR #1157, opened 2024-02-29
Fix bug to allow parenthesis query after `INSERT INTO`

jonathanlehto:
Fixing a bug where the columns after hive check would consume a query

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7135,6 +7135,16 @@ impl<'a> Parser<'a> {
         } else if self.consume_token(&Token::LParen) {
             // CTEs are not allowed here, but the parser currently accepts them
             let subquery = self.parse_query()?;
+
+            let is_generic = dialect_of!(self is GenericDialect);
+            if !is_generic {
+                if let SetExpr::Values(_) = *subquery.body {
+                    return Err(ParserError::ParserError(format!(
+                        "VALUES is not a recognized subquery"
+                    )))
+                }
+            }
+
             self.expect_token(&Token::RParen)?;
             SetExpr::Query(Box::new(subquery))
         } else if self.parse_keyword(Keyword::VALUES) {
@@ -10152,7 +10162,35 @@ mod tests {
     fn test_replace_into_select() {
         let sql = r#"REPLACE INTO t1 (a, b, c) (SELECT * FROM t2)"#;
 
-        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_err());
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
+    }
+
+    #[test]
+    fn test_insert_into_select() {
+        let sql = r#"INSERT INTO t1 (a, b, c) (SELECT * FROM t2)"#;
+
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
+    }
+
+    #[test]
+    fn test_insert_into_values() {
+        let sql = r#"INSERT INTO t1 (a) VALUES(1)"#;
+
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
+    }
+
+    #[test]
+    fn test_insert_into_values_wrapped() {
+        let sql = r#"INSERT INTO t1 (a) (VALUES(1))"#;
+
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
+    }
+
+    #[test]
+    fn test_insert_into_values_wrapped_not_generic() {
+        let sql = r#"INSERT INTO t1 (a) (VALUES(1))"#;
+
+        assert!(Parser::parse_sql(&MySqlDialect {}, sql).is_err());
     }
 
     #[test]
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8411,8 +8411,15 @@ impl<'a> Parser<'a> {
 
                     let partitioned = self.parse_insert_partition()?;
 
-                    // Hive allows you to specify columns after partitions as well if you want.
-                    let after_columns = self.parse_parenthesized_column_list(Optional, false)?;
+                    // make sure we are not about to consume a query
+                    let mut after_columns = vec![];
+                    match self.peek_nth_token(1).token {
+                        Token::Word(w) if w.keyword != Keyword::SELECT => {
+                            // Hive allows you to specify columns after partitions as well if you want.
+                            after_columns = self.parse_parenthesized_column_list(Optional, false)?;
+                        },
+                        _ => {}
+                    };
 
                     let source = Some(Box::new(self.parse_query()?));
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8410,16 +8410,9 @@ impl<'a> Parser<'a> {
             } else {
                 let columns = self.parse_parenthesized_column_list(Optional, is_mysql)?;
 
-                let partitioned = self.parse_insert_partition()?;
-
-                // make sure we are not about to consume a query
-                let mut after_columns = vec![];
-                match self.peek_nth_token(1).token {
-                    Token::Word(w) if w.keyword != Keyword::SELECT => {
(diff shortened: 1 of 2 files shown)


alamb:
Hi @jonathanlehto  -- can you please fix the CI tests? 

I think [one](https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/8098985632/job/22142811676?pr=1157) needs `cargo fmt` and there appears to be another error https://github.com/sqlparser-rs/sqlparser-rs/actions/runs/8098985632/job/22142812958?pr=1157

jonathanlehto:
Apologies, I may not be as familiar with MySQL syntax, but it looks like the test case error is due to asserting that the parser will fail the REPLACE statement.  From comparing the [INSERT](https://dev.mysql.com/doc/refman/8.0/en/insert.html) and [REPLACE](https://dev.mysql.com/doc/refman/8.0/en/replace.html) docs, it looks like that statement should be succeeding (I could very well be wrong).

If that test case should fail, I can get another update in.  I just want to make sure we are on the same page first!

src/parser/mod.rs
```diff
@@ -7943,21 +7943,29 @@ impl<'a> Parser<'a> {
...
 
-                    let partitioned = self.parse_insert_partition()?;
+                let partitioned = self.parse_insert_partition()?;
 
-                    // Hive allows you to specify columns after partitions as well if you want.
-                    let after_columns = self.parse_parenthesized_column_list(Optional, false)?;
+                // make sure we are not about to consume a query
+                let mut after_columns = vec![];
+                match self.peek_nth_token(1).token {
+                    Token::Word(w) if w.keyword != Keyword::SELECT => {
```
alamb, line 7958: Checking for select seems incorrect here

What about a query like 
```sql
INSERT INTO tbla (cola) (VALUES (1))
```

The Hive manual https://cwiki.apache.org/confluence/display/hive/languagemanual+dml doesn't appear to say anything about this syntax

```
Standard syntax:
INSERT OVERWRITE TABLE tablename1 [PARTITION (partcol1=val1, partcol2=val2 ...) [IF NOT EXISTS]] select_statement1 FROM from_statement;
INSERT INTO TABLE tablename1 [PARTITION (partcol1=val1, partcol2=val2 ...)] select_statement1 FROM from_statement;
 
Hive extension (multiple inserts):
FROM from_statement
INSERT OVERWRITE TABLE tablename1 [PARTITION (partcol1=val1, partcol2=val2 ...) [IF NOT EXISTS]] select_statement1
[INSERT OVERWRITE TABLE tablename2 [PARTITION ... [IF NOT EXISTS]] select_statement2]
[INSERT INTO TABLE tablename2 [PARTITION ...] select_statement2] ...;
FROM from_statement
INSERT INTO TABLE tablename1 [PARTITION (partcol1=val1, partcol2=val2 ...)] select_statement1
[INSERT INTO TABLE tablename2 [PARTITION ...] select_statement2]
[INSERT OVERWRITE TABLE tablename2 [PARTITION ... [IF NOT EXISTS]] select_statement2] ...;
 
Hive extension (dynamic partition inserts):
INSERT OVERWRITE TABLE tablename PARTITION (partcol1[=val1], partcol2[=val2] ...) select_statement FROM from_statement;
INSERT INTO TABLE tablename PARTITION (partcol1[=val1], partcol2[=val2] ...) select_statement FROM from_statement;
```

I may be missing something

tests/sqlparser_common.rs
```diff
@@ -106,6 +106,9 @@ fn parse_insert_values() {
     }
 
     verified_stmt("INSERT INTO customer WITH foo AS (SELECT 1) SELECT * FROM foo UNION VALUES (1)");
+
+    // allow parenthesis query after insert into
+    verified_stmt("INSERT INTO tbla (cola) (SELECT cola FROM tblb)");
```
alamb, line 111: What does this syntax mean? Is it the same as 

```sql
INSERT INTO tbla (cola) SELECT cola FROM tblb
```

(no parens?)

alamb (commented):
Thanks for bearing with me @jonathanlehto 

I left some questions about this PR

tests/sqlparser_common.rs
jonathanlehto: I believe so, well at the least, running an explain statement in Athena (Trino) yields no difference between

`INSERT INTO tbla (cola) SELECT cola FROM tblb`
and
`INSERT INTO tbla (cola) (SELECT cola FROM tblb)`

And registers both as valid queries.  From what I recall, I think you can generally wrap any select statements in quotes

src/parser/mod.rs
jonathanlehto: It does look like `INSERT INTO tbla (cola) (VALUES (1))` is valid syntax for Athena / Presto as well.  Maybe it needs to be broadened?

jonathanlehto: From reading those docs, it looks like the PARTITION keyword is required.  I moved the after_columns parsing into that statement.  I don't know why I didn't do that originally, but I think this should resolve the situation.  I added another test case

alamb:
🤔  it seems like the CI is still failing on this PR

src/parser/mod.rs
jonathanlehto: I may have mis-communicated this.  I went ahead and confirmed that the failing test case is incorrect.

Test case:
`
    #[test]
    fn test_replace_into_select() {
        let sql = r#"REPLACE INTO t1 (a, b, c) (SELECT * FROM t2)"#;

        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_err());
    }
`

I created a local mysql table and successfully ran these commands.

`
mysql> INSERT INTO tab (col) (SELECT col FROM tab);
Query OK, 1 row affected (0.02 sec)
Records: 1  Duplicates: 0  Warnings: 0

mysql> REPLACE INTO tab (col) (SELECT col FROM tab);
Query OK, 2 rows affected (0.02 sec)
Records: 2  Duplicates: 0  Warnings: 0
`

However, you can not wrap values in mysql, copied below.  From what I can tell, only Trino / Athena allow values statement to be wrapped.

`
mysql> INSERT INTO tab (col) VALUES (1);
Query OK, 1 row affected (0.02 sec)

mysql> INSERT INTO tab (col) (VALUES (1));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(1))' at line 1
`

Unfortunately, I don't have the time to check every Dialect that is supported, but I know that Athena / Trino will succeed and at least MySQL will fail.

In this case, do we only want to succeed a parenthesis values statement in common, and fail for every other dialect?  There seems to be another bug here

alamb:
It appears the tests are still failing on this PR

jonathanlehto:
I updated the mr to fix the lint and codestyle errors.  I was hoping to get your thoughts on [this](https://github.com/sqlparser-rs/sqlparser-rs/pull/1157#discussion_r1513155120) comment, so, if you have some time, I would love to know what you think about that implementation!

src/parser/mod.rs
```diff
@@ -10150,7 +10163,35 @@ mod tests {
...
+        let sql = r#"INSERT INTO t1 (a, b, c) (SELECT * FROM t2)"#;
+
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
+    }
+
+    #[test]
+    fn test_insert_into_values() {
+        let sql = r#"INSERT INTO t1 (a) VALUES(1)"#;
+
+        assert!(Parser::parse_sql(&GenericDialect {}, sql).is_ok());
```
alamb, line 10180: Can you please put these tests in common with other similar tests, like this:

https://github.com/sqlparser-rs/sqlparser-rs/blob/2eabfb82083f97bb01883cd3c66c0b035343e3ba/tests/sqlparser_common.rs#L495-L501

alamb (commented):
Thanks @jonathanlehto  -- sorry for the delay in reviewing -- the time I can spend on this crate is quite limited and I don't have the bandwidth to study problems in depth typically. 

I took a look over this PR again and left some feedback. Let me know what you think

alamb:

PR #1158, opened 2024-02-29
Clean up nightly clippy lints

alamb:
I noticed the github actions runner was marking these on PRs as the nightly clippy has gotten smarter.

So I ran `cargo clippy +nightly` and fixed the resulting errors

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -5200,7 +5200,13 @@ impl fmt::Display for CopyLegacyCsvOption {
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/ast/mod.rs
```diff
@@ -5200,7 +5200,13 @@ impl fmt::Display for CopyLegacyCsvOption {
     }
 }
 
+/// `MERGE` Statement
 ///
```
alamb, line 5204: Clippy flags empty doc comments now

Issue #1149, opened 2024-02-29
[BUG] parse failed in a joint index of multiple fields in mysql dialect


hubertshelley:
```sql
CREATE TABLE `data` (
  `id` bigint NOT NULL AUTO_INCREMENT,
  `bind_id` varchar(36) COLLATE utf8mb4_general_ci NOT NULL,
  `json` json NOT NULL,
  `source_key` varchar(128) COLLATE utf8mb4_general_ci NOT NULL,
  `is_synced` tinyint(1) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `bind_id_source_key_ids` (`bind_id`,`source_key`) USING BTREE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
```
sql parser error: Expected ',' or ')' after column definition, found: USING at Line: 8, Column 64

PR #1164, opened 2024-03-06
Support MySQL `UNIQUE` table constraint

Nikita-str:
Fix to #1149 and some more.

Allow to parse `[CONSTRAINT [name]] UNIQUE [INDEX | KEY] [index_name] [index_type] (<columns>) [index_option]s ...` as it say in the [specification](https://dev.mysql.com/doc/refman/8.3/en/create-table.html).

Also unallow `CONSTRAINT bar INDEX ind (foo)` for table constraint; as [it was said](https://github.com/Nikita-str/sqlparser-rs/blob/199af79420337d3c03a4853240f49b953804262a/src/ast/ddl.rs#L433) `INDEX ind (foo)` form of table constraint was found only in MySQL, but in MySQL allowed only `INDEX ind (foo)` without `CONSTRAINT bar` before it. 

Some remarks:
* Maybe it have sense to split `TableConstraint::Unique` into `TableConstraint::Unique` & `TableConstraint::Primary`?
* In this PR was created `IndexOption` struct, and maybe it can be unite with some another structure, I unsure.
* In MySQL `CONSTRAINT a UNIQUE b (x)` will set `b` and for `index_name`(in MySQL you can obtain it by `SELECT index_name FROM INFORMATION_SCHEMA.STATISTICS`) and for `constraint_name`(in MySQL you can obtain it by `SELECT constraint_name FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS`) so I don't sure if [names](https://github.com/Nikita-str/sqlparser-rs/blob/199af79420337d3c03a4853240f49b953804262a/src/ast/ddl.rs#L404-L405) really needed to be splitted.


alamb:
> Maybe it have sense to split TableConstraint::Unique into TableConstraint::Unique & TableConstraint::Primary?


Yes, I agree this would make more sense than trying to combine UNIQUE and PRIMARY KEY syntaxes together

Are you willing to make this change? I am really sorry for the delay in reviewing

Nikita-str:
I do the split, but now I'm unsure that it was good idea, because MySql [specification](https://dev.mysql.com/doc/refman/8.3/en/create-table.html) say that there no `[index_name]` in `PRIMARY KEY` variant but actually compiler take `CREATE TABLE table_name (x INT, CONSTRAINT name PRIMARY KEY index_name (x));` as valid query. So `PRIMARY KEY` and `UNIQUE` have less differences than I expected.

Their similarity is visible in tests [during creation](https://github.com/sqlparser-rs/sqlparser-rs/blob/7b2df86b7f99cfbcff21f5269cb0c2cc23e8cf64/tests/sqlparser_mysql.rs#L505) and [during assertion of inner structure](https://github.com/sqlparser-rs/sqlparser-rs/blob/7b2df86b7f99cfbcff21f5269cb0c2cc23e8cf64/tests/sqlparser_mysql.rs#L542).

But still maybe when they are separate it's more expected by an user, idk.

alamb:
> But still maybe when they are separate it's more expected by an user, idk.

I agree -- the sqlparser's crate[ aim is to preserve syntax](https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics) and it is up downstream consumers to decide how to handle the semantics (e.g if they want to treat `UNIQUE` and `PRIMARY` the same(

src/ast/ddl.rs
```diff
@@ -922,6 +1037,36 @@ fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
     ConstraintName(name)
 }
 
+/// If `option` is
```
alamb, line 1045: 👍  I agree this is very nice

alamb (approved):
Thank you very much @Nikita-str  -- I pushed a few small comment improvements and then I'll merge this PR in.

Thanks again and sorry for the delay in reviewing

alamb:
Thanks again @Nikita-str

Issue #999, opened 2023-10-11
Cannot parse named windows with modifiers


jesse-shopify:
sqlparser-rs only supports named windows that are an identifier. This is fine for queries like:

```sql
SELECT sum(salary) OVER w, avg(salary) OVER w
  FROM empsalary
  WINDOW w AS (PARTITION BY depname ORDER BY salary DESC);
```

However, it is insufficient for queries in at least BigQuery and MySQL where modifiers can be added to the named window in the OVER clause.

[MySQL named window example](https://dev.mysql.com/doc/refman/8.0/en/window-functions-named-windows.html)

```sql
SELECT
  DISTINCT year, country,
  FIRST_VALUE(year) OVER (w ORDER BY year ASC) AS first,
  FIRST_VALUE(year) OVER (w ORDER BY year DESC) AS last
FROM sales
WINDOW w AS (PARTITION BY country)
```

[BigQuery named window example](https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls)

```sql
SELECT item, purchases, category, LAST_VALUE(item)
  OVER (
    item_window
    ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING
  ) AS most_popular
FROM Produce
WINDOW item_window AS (
  PARTITION BY category
  ORDER BY purchases)
```

This would build upon prior to work:
- https://github.com/sqlparser-rs/sqlparser-rs/issues/867
- https://github.com/sqlparser-rs/sqlparser-rs/pull/881

alamb:
Thanks for the report @jesse-shopify  -- any contributions to add such support are most welcome

LucaCappelletti94:
Closing as completed in https://github.com/apache/datafusion-sqlparser-rs/pull/1166

PR #1166, opened 2024-03-07
Support named windows in `OVER (window_definition)` clause

Nikita-str:
Issue #999

Parse `window_name` in `OVER ([window_name] [PARTITION BY ..] [ORDER BY ..] ..)`


tests/sqlparser_common.rs
```diff
@@ -4115,6 +4117,62 @@ fn parse_window_functions() {
         }),
         expr_from_projection(&select.projection[0])
     );
+
+    for i in 0..7 {
```
alamb, line 4121: 👍

alamb (approved):
Thank you for the contribution @Nikita-str  -- this makes sense to me

alamb:
> @alamb first of all, should I do something with parse_deeply_nested_parens_hits_recursion_limits test? Because locally it's failed with stack overflow even in main branch. All others test result is ok.

I think we may need to update the default limit to have a lower limit (it is right on the edge of stack limit in the debug build, and so any change to the stack size leads to a failure)

Maybe we can do it in a separate PR -- any chance you are willing to do so? If not I will try to do so but it may be several days

Nikita-str:
@alamb do you suggest to change `const DEFAULT_REMAINING_DEPTH: usize = 50;` in `src\parser\mod.rs` to `40` for example?

Maybe it's better to wrap this test code into new thread with bigger stack size and then wait until the thread finished?:
```rust
    const KB: usize = 1024;
    const MB: usize = 1024 * KB;
    std::thread::Builder::new().stack_size(8 * MB).spawn(move || { 
        // prev code
    }).unwrap().join().unwrap();
```

alamb:
> Maybe it's better to wrap this test code into new thread with bigger stack size and then wait until the thread finished?:

I agree that seems like a less problematic approach

alamb:
> @alamb do you suggest to change `const DEFAULT_REMAINING_DEPTH: usize = 50;` in `src\parser\mod.rs` to `40` for example?

Yes, this is basically my suggestion

The rationale is that doing so will prevent stack overflows / panics for debug builds.

Another thing we can do is to see if we can reduce the stack frame size for some of the intermediate function calls (reduce stack frame means make fewer local variables)

Nikita-str:
> Maybe we can do it in a separate PR -- any chance you are willing to do so? 
Ok, I will do it, at least by wrapping in new thread.
If I solve it by wrapping then I would to wrap all tests that assert `ParserError::RecursionLimitExceeded` to prevent future the same errors with stack overflow at least in already existed tests, ok?

alamb:
You are not the first to hit this problem -- I think @universalmind303  saw it in https://github.com/sqlparser-rs/sqlparser-rs/pull/1144

Thus i think we need to do some real work -- I will try and find time later this week to help

alamb:
I merged up from main and now this test is passing 🎉

alamb:
Thanks again @Nikita-str -- this is a very nice contribution

PR #1170, opened 2024-03-10
Add identifier quote style to `Dialect` trait

backkem:
This relates to apache/arrow-datafusion#9517 where we're adding an SQL `unparser` to turn DataFusion Expr/Plans back into SQL.

I extended the method to allow passing the identifier value. This may be helpful in case we want to add heuristics to only add quotes if truly needed by the dialect.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -108,6 +108,10 @@ pub trait Dialect: Debug + Any {
     fn is_delimited_identifier_start(&self, ch: char) -> bool {
         ch == '"' || ch == '`'
     }
+    /// Return the character used to quote identifiers.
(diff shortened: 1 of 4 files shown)


backkem:
cc @alamb

alamb:
Looks good to me -- thank you @backkem  

I think before we could merge this PR we need some tests ( mostly so if it got broken by accident in the future we would know because a tet would fail)

backkem:
I added a basic test. It feels somewhat artificial but I didn't want to build additional 'unparser' logic in the test case.

alamb (approved):
> I added a basic test. It feels somewhat artificial but I didn't want to build additional 'unparser' logic in the test case.


I agree it feels somewhat forced, but I think that is also OK in this case. We can improve things later if necessary.

Thank you @backkem

PR #1171, opened 2024-03-12
solve `stack overflow` on `RecursionLimitExceeded` during debug building

Nikita-str:
The main idea is move \<usage of creation large structures on stack after which they moved into `Box`> into \<separate fns>. So for each execution branch would be reserved less stack memory.

Current changes focused on case of parsing seq of repeated `Token::LParen`(`(...(`) because the problem appeared on `parse_deeply_nested_parens_hits_recursion_limits` test(`the test`).

So in [first commit](https://github.com/Nikita-Str/sqlparser-rs/commit/1362e4642c3c0cfd17fcb7ed2db3c7245f46165e) instead of
```rust
let large_struct = self.parse_large_struct()?;
let boxed_large_struct = Box::new(large_struct);
// without compiler optimization on stack reserves
// memory for `LargeStruct`
```
used
```rust
fn parse_boxed_large_struct(&mut self) -> Result<Box<LargeStruct>, ..> {
   self.parse_large_struct().map(Box::new)
}
..
let boxed_large_struct = self.parse_boxed_large_struct()?;
// on stack reserves only memory for pointer
```

In [second commit](https://github.com/Nikita-Str/sqlparser-rs/commit/cfc15256eb192351cd40ece60787a451788f3f1c) used `Result::map` as wrapper fn to avoid placing object on stack.

In [third commit](https://github.com/Nikita-Str/sqlparser-rs/commit/63baa83c2a4b146ec6d04e355815feb9f87867f4) move out some cases of parsing in separate fns to split used stack by execution flow. It's done slightly ugly, maybe I shouldn't move it into [local mod](https://github.com/Nikita-Str/sqlparser-rs/commit/63baa83c2a4b146ec6d04e355815feb9f87867f4#diff-fc6a8d66b8cb6bd48a119a70e489cbcef82e7f551e7cf08e8e972ef4e774ef49R6864-R6901).

---
**The next words is compiler/OS related**

It seems like at least my compiler in debug build in next case reserve separate space on stack for `Result<Ok, Err>` and for `Ok`
```rust
let x = return_result()?; // seems like here reserved space for `Result<Ok, Err>` and for `Ok`
```
because after first commit `the test` worked with `DEFAULT_REMAINING_DEPTH = 60;` and before this commit it panics with `DEFAULT_REMAINING_DEPTH = 50;` and without this assumption such things shouldn't happen(in `the test` trace changed from `call:parse_query_body` to `call:parse_boxed_query_body THEN call:parse_query_body`). `sizeof(SetExpr) = 0x3C0` and stack allocation in `parse_query` decreased by `0xB00` (almost 3 times).

Locally after all 3 commits the stack memory reservation changed from `0x6020` to `0x1C80` for `parse_query` & from `0x1F30` to `0x1780` for `parse_query_body` (but in `(...(`-case there is additional stack with size ~ `sizeof(SetExpr)` & `sizeof(Query)`). The real decrease in stack memory reserving to one recursion step in `(...(`-case is `~2.1` times on my local machine.


Nikita-str:
In short, on local machine `stack_overflow` in `parse_deeply_nested_parens_hits_recursion_limits` test fixed. It panics in debug build only after `const DEFAULT_REMAINING_DEPTH: usize = 125;`.

alamb (approved):
Thank you so much @Nikita-str  -- think this change makes sense and your code is very nicely written. I am very much sorry for the delay in reviewing. 

It think the new `parse_boxed_...` methods should not be pub, at least until someone explicitly needs them, to keep the API surface area of this crate smaller. 

I also think the inner module definitions were a little strange -- so I pushed some changes to to follow your other pattern of returning `Box`ed bodies.

alamb:
I also validated that with the changes in this PR the tests from https://github.com/sqlparser-rs/sqlparser-rs/pull/1166 also pass 🚀

PR #1173, opened 2024-03-13
Parse `SUBSTRING` `FROM` syntax in all dialects, reflect change in the AST

### LINKED ISSUE #1168: SUBSTRING function calls are always formatted as ` SUBSTRING(... FROM ... FOR ...)`
Parsing the following valid SQLite statement:

```sql
select substring('xxx', 1, 2)
```

then dumping the parsed result as sql results in the following syntax, which is invalid in SQLite:

```sql
select SUBSTRING('xxx' FROM 1 FOR 2)
```

lovasoa:
always accept both substring function syntaxes in all dialects,
and ensure SUBSTRING expression are always formatted with the same syntax as the one they were parsed with.

This both simplifies the code (making it easier to maintain as databases evolve) AND creates a higher-fidelity syntax tree for dialects that support both syntaxes

fixes  #1168

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -570,7 +570,7 @@ pub enum Expr {
 
         /// false if the expression is represented using the `SUBSTRING(expr [FROM start] [FOR len])` syntax
         /// true if the expression is represented using the `SUBSTRING(expr, start, len)` syntax
-        /// This flags is used for formatting.
+        /// This flag is used for formatting.
         special: bool,
     },
     /// ```sql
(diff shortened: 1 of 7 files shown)


alamb (approved):
I agree @lovasoa  -- this design is consistent with the rest of sqlparser-rs and the design of "parse the syntax and leave semantics to the user"

Since it is now possible to distinguish between the two syntactic forms via AST I believe users can implement whatever behavior they desire.

THank you very much

alamb (approved):

alamb:
Thanks again @lovasoa

PR #1177, opened 2024-03-14
Add support for quoted string backslash escaping

iffyio:
This adds support for parsing string literals on
dialects that treat backslash character as an escape
character. As an example, the following previously failed
to parse by dialects like BigQuery where the syntax is valid.
```sql
SELECT 'a\'b';
```

Moves the SQL `like` and `similar_to` tests from individual
dialects to common since the tests were identical.


src/tokenizer.rs
```diff
@@ -1235,6 +1245,11 @@ impl<'a> Tokenizer<'a> {
                 '\\' => {
                     // consume
                     chars.next();
+
+                    if allow_escape {
```
alamb, line 1249: This behavior seems different than how MySqlDialect behaves

Specifically, with MySQL the escape characters are transformed into their literal values (`e.g. `'a\"b'` would be parsed to `a"b` while this PR would parse it to `a\"b`

What do you think about making this consistent?

alamb (commented):
Thank you for this PR @iffyio -- I am sorry it took me so long to review it properly. I am a little concerned about the difference between how this PR works and how it works for MySqlDialect

Is there any way we can unify the behavior?

src/tokenizer.rs
iffyio: Ah yes I'll take a closer look at this to keep the same behavior for Mysql

iffyio: @alamb Looking into updating this and turns out I didn't follow the comment entirely and was unable to infer the inconsistency for mysql - could you clarify the problem once more? It seems Github/markdown unfortunately reformatted the expected and desired output in your example so that they became identical :sweat_smile:

alamb: As in I think the tokenizer for mysql (assuming `self.unescape` is true) actually does the unescaping -- so a string with an escape character (`"\x20"`) would actually be tokenized as a space (`" "`)

iffyio: That makes sense! I've updated this have the new logic respect the unescape similar to mysql - let me know if that's what you had in mind!

alamb (approved):
Looks great to me -- thank you @iffyio  🙏

PR #1178, opened 2024-03-14
BigQuery: support unquoted hyphen in table/view declaration

iffyio:
Adds support for proper object name parsing in bigquery to within a `CREATE TABLE` or `CREATE VIEW` statement. Similar to the parsing logic for when the table/view is later referenced.

The lacking part was that bigquery allows unquoted hyphens in the first part of an object name - so that e.g.
`CREATE TABLE my-project.mydataset.mytable`
wasn't being parsed correctly.


alamb (approved):
Thank you @iffyio

PR #1179, opened 2024-03-14
Add support for arbitrary map access expr

iffyio:
Unlike array access, the current logic for
accessing maps had custom handling and didn't
support cases like negative numbers or function
calls.
This fixes the issue by allowing the key access to be parsed as an arbitrary expression.

Also includes support on BigQuery where both dot and
bracket notations are allowed in the same expression
https://cloud.google.com/bigquery/docs/nested-repeated#query_nested_and_repeated_columns

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2473,23 +2473,43 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_map_access(&mut self, expr: Expr) -> Result<Expr, ParserError> {
-        let key = self.parse_map_key()?;
-        let tok = self.consume_token(&Token::RBracket);
-        debug!("Tok: {}", tok);
-        let mut key_parts: Vec<Expr> = vec![key];
-        while self.consume_token(&Token::LBracket) {
-            let key = self.parse_map_key()?;
-            let tok = self.consume_token(&Token::RBracket);
-            debug!("Tok: {}", tok);
-            key_parts.push(key);
+        let key = self.parse_expr()?;
+        self.expect_token(&Token::RBracket)?;
+
+        let mut keys = vec![MapAccessKey {
+            key,
+            syntax: MapAccessSyntax::Bracket,
+        }];
+        loop {
+            let key = match self.peek_token().token {
+                Token::LBracket => {
+                    self.next_token(); // consume `[`
+                    let key = self.parse_expr()?;
+                    self.expect_token(&Token::RBracket)?;
+                    MapAccessKey {
+                        key,
+                        syntax: MapAccessSyntax::Bracket,
+                    }
+                }
+                // Access on BigQuery nested and repeated expressions can
+                // mix notations in the same expression.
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -2473,23 +2473,43 @@ impl<'a> Parser<'a> {
...
+                    self.expect_token(&Token::RBracket)?;
+                    MapAccessKey {
+                        key,
+                        syntax: MapAccessSyntax::Bracket,
+                    }
+                }
+                // Access on BigQuery nested and repeated expressions can
+                // mix notations in the same expression.
+                // https://cloud.google.com/bigquery/docs/nested-repeated#query_nested_and_repeated_columns
+                Token::Period if dialect_of!(self is BigQueryDialect) => {
```
alamb, line 2497: What do you think about adding GenericDialect here too?

alamb (approved):
Thank you so much @iffyio  -- this looks really nice. I had a small suggestion but it may not make sense.

Also, in general thank you for your work on this (and the other related PRs) and the reviews. Is there any chance you are interested in helping be a maintainer of this crate? As is probably obvious I do not have a huge amount of time and I think it is suffering because of that.

src/parser/mod.rs
iffyio: Yeah I suspect it would be valid to include Generic here without conflicting (from an entry point pov I remember with the current setup Generic doesn't fall into this function today because together [with Postgres](https://github.com/validio-io/sqlparser-rs/blob/d67fd34a8544e4dfc2e03feb1e8858f4a517ddd0/src/parser/mod.rs#L2404-L2408) it parses the expression as an Array instead - but that could change in the future)

iffyio:
@alamb Thanks a lot for the reviews :pray:

Regarding maintainership, sure I'm happy to help out in this regard and can allocate more time to reviews and tracking issues etc. Do let me know how to effectively help out here

alamb:
> Regarding maintainership, sure I'm happy to help out in this regard and can allocate more time to reviews and tracking issues etc. Do let me know how to effectively help out here

Thank you @iffyio  -- that is great news. 

Basically the work is reviewing other PRs and getting them ready for merge (which means doing what you already do well -- make sure they are tested, documented, and don't cause overly painful changes for downstream crates)

There is more documentation on https://github.com/sqlparser-rs/sqlparser-rs/discussions/940 

Let me know if you want to learn more or have a video call to discuss the possibility

iffyio:
That sounds good to me! Will look to help out further in that case :+1:

alamb:
> That sounds good to me! Will look to help out further in that case 👍

BTW if you see a PR that is ready to go in your opinion please just mention (@alamb) me and I'll take a look. Otherwise I likely won't see it as I don't monitor all alerts from this repo

iffyio:
> BTW if you see a PR that is ready to go in your opinion please just mention (@alamb) me and I'll take a look. Otherwise I likely won't see it as I don't monitor all alerts from this repo

Ah good to know, will do!

PR #1181, opened 2024-03-15
support clickhouse parameterized views

### LINKED ISSUE #1167: ClickHouse parameterized view
I'd like to parse ClickHouse SQL which contains parameters for their [parameterized views](https://clickhouse.com/docs/en/sql-reference/statements/create/view#parameterized-view). The SQL looks like

`CREATE VIEW view AS SELECT * FROM TABLE WHERE Column1={column1:datatype1}`

Note the parameter is in the { } brackets. I don't think this could be implemented in the `Dialect` as curly braces are tokenized without deferring to the dialect. It would probably have to be handled by the tokenizer as a special case like numbers are for MySQL & Hive.

Lordworms:
Closes #1167

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -792,6 +792,17 @@ pub enum Expr {
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/where#joins-in-the-where-clause>.
     OuterJoin(Box<Expr>),
+
+    /// ClickHouse supported parameterized view
+    /// for example
+    /// ```sql
+    /// CREATE VIEW view AS SELECT * FROM TABLE WHERE Column1={column1:datatype1} and Column2={column2:datatype2}
+    /// ```
+    /// see doc <https://clickhouse.com/docs/en/sql-reference/statements/create/view#parameterized-view>
+    ParameterizedViewArg {
+        column: Ident,
+        data_type: DataType,
+    },
 }
 
 impl fmt::Display for CastFormat {
@@ -1250,6 +1261,9 @@ impl fmt::Display for Expr {
             }
             Expr::OuterJoin(expr) => {
                 write!(f, "{expr} (+)")
+            }
+            Expr::ParameterizedViewArg { column, data_type } => {
+                write!(f, "{{{}:{}}}", column, data_type)
             }
         }
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -936,8 +936,8 @@ impl<'a> Parser<'a> {
         let data_type = self.parse_data_type()?;
         self.expect_token(&Token::RBrace)?;
         Ok(Expr::ParameterizedViewArg {
-            column: (name),
-            data_type: (data_type),
+            column: name,
+            data_type,
         })
     }
     /// Parse an expression prefix
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -928,14 +928,38 @@ impl<'a> Parser<'a> {
 
         Ok(Statement::ReleaseSavepoint { name })
     }
-
+    /// Parse clickhouse view parameter style `column1:data_type`
+    pub fn parse_column_with_data_type(&mut self) -> Result<Expr, ParserError> {
+        self.expect_token(&Token::LBrace)?;
+        let name = self.parse_identifier(false)?;
+        self.expect_token(&Token::Colon)?;
+        let data_type = self.parse_data_type()?;
+        self.expect_token(&Token::RBrace)?;
+        Ok(Expr::ParameterizedViewArg {
+            column: (name),
+            data_type: (data_type),
+        })
+    }
     /// Parse an expression prefix
     pub fn parse_prefix(&mut self) -> Result<Expr, ParserError> {
         // allow the dialect to override prefix parsing
         if let Some(prefix) = self.dialect.parse_prefix(self) {
             return prefix;
         }
-
+        // ClickHouse support using {column:Datatype} format in its creation of parametrised views
+        match self.peek_token().token {
+            // check if last token is WHERE and if it is do it, else do the following logic
+            Token::LBrace if dialect_of!(self is ClickHouseDialect | GenericDialect) => {
+                self.prev_token();
+                // prev token is EQ
+                if self.consume_token(&Token::Eq) {
+                    return self.parse_column_with_data_type();
+                } else {
+                    self.next_token();
+                }
+            }
+            _ => {}
+        }
         // PostgreSQL allows any string literal to be preceded by a type name, indicating that the
         // string literal represents a literal of that type. Some examples:
         //
@@ -2542,6 +2566,12 @@ impl<'a> Parser<'a> {
             }
             self.parse_map_access(expr)
         } else if Token::Colon == tok {
+            if dialect_of!(self is ClickHouseDialect) && self.index() > 0 {
+                return Ok(Expr::ParameterizedViewArg {
+                    column: Ident::new(expr.to_string()),
+                    data_type: self.parse_data_type()?,
+                });
+            }
             Ok(Expr::JsonAccess {
                 left: Box::new(expr),
                 operator: JsonOperator::Colon,
--- a/tests/sqlparser_clickhouse.rs
+++ b/tests/sqlparser_clickhouse.rs
@@ -400,3 +400,8 @@ fn clickhouse_and_generic() -> TestedDialects {
         options: None,
     }
 }
+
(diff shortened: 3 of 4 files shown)


src/parser/mod.rs
```diff
@@ -853,14 +853,22 @@ impl<'a> Parser<'a> {
 
         Ok(Statement::ReleaseSavepoint { name })
     }
-
+    pub fn parse_column_with_data_type(&mut self) -> Result<Expr, ParserError> {
+        self.expect_token(&Token::LBrace)?;
+        let column_with_data_type = self.parse_expr()?;
+        self.expect_token(&Token::RBrace)?;
+        Ok(column_with_data_type)
+    }
```
iffyio, line 861:
```suggestion

    pub fn parse_column_with_data_type(&mut self) -> Result<Expr, ParserError> {
        self.expect_token(&Token::LBrace)?;
        let column_with_data_type = self.parse_expr()?;
        self.expect_token(&Token::RBrace)?;
        Ok(column_with_data_type)
    }

```

src/ast/mod.rs
```diff
@@ -733,6 +733,10 @@ pub enum Expr {
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/where#joins-in-the-where-clause>.
     OuterJoin(Box<Expr>),
+    ColumnWithDataType {
```
iffyio, line 736: Could we include a description for this new variant (ideally including a link to the docs)?

src/ast/mod.rs
```diff
@@ -733,6 +733,10 @@ pub enum Expr {
...
+        columns: Box<Expr>,
+        data_type: DataType,
+    },
```
iffyio, line 739: I'm wondering if naming wise, 'ColumnWithDataType' sounds a bit too generic since it applies to a lot of unrelated scenarios. From the docs, it sounds incorrect as well to have the `columns: Box<Expr>`?
-  The actual column in `WHERE a = {b:INT}` is `a`, while `b` would be more of a function parameter
-  `b` given its a function parameter likely needs to be an `Ident` instead of an `Expr`

thinking roughly something along this line could be more descriptive
```rust
ParameterizedViewArg{
    name: Ident,
    data_type: DataType,
}
```

src/parser/mod.rs
```diff
@@ -853,14 +853,22 @@ impl<'a> Parser<'a> {
...
     pub fn parse_prefix(&mut self) -> Result<Expr, ParserError> {
         // allow the dialect to override prefix parsing
         if let Some(prefix) = self.dialect.parse_prefix(self) {
             return prefix;
         }
-
+        // ClickHouse support using {column:Datatype} format in its creation of parametrised views
+        if dialect_of!(self is ClickHouseDialect) && self.peek_token().token == Token::LBrace {
+            return self.parse_column_with_data_type();
+        }
```
iffyio, line 871: Currently the parse logic for this feature is spread across this and the parse_infix function where one triggers the other indirectly, slightly harder to follow as a result - one suggestion to simplify this would be:
- The current function already includes a match onthe token type, so we can introduce a clause
   ```rust
   Token::LBRACE if dialect_of!(self is Clickhouse) => {
       self.prev_token();
       self.parse_column_with_data_type()
   }
    ```
- The current `parse_column_with_data_type` can parse the components directly - especially given that it does not seem correct for it to parse an inner expression. It becomes
   ```rust
   // consume LBRACE
   let name = self.parse_identifier()?;
   self.expect(Token::Colon)?;
   let data_type = self.parse_data_type()?;
   // consume RBRACE
   ```

src/ast/mod.rs
Lordworms: Sure, I'll change it

src/ast/mod.rs
Lordworms: Sure!

src/parser/mod.rs
Lordworms: make sense to me.

src/ast/mod.rs
```diff
@@ -733,6 +733,17 @@ pub enum Expr {
...
     OuterJoin(Box<Expr>),
+
+    /// ClickHouse supported parameterized view
+    /// for example
+    /// ```sql
+    /// CREATE VIEW view AS SELECT * FROM TABLE WHERE Column1={column1:datatype1} and Column2={column2:datatype2}
+    /// ```
+    /// see doc <https://clickhouse.com/docs/en/sql-reference/statements/create/view#parameterized-view>
+    ParameterizedViewArg {
+        columns: Ident,
```
alamb, line 744: Since there is only a single column here, I think this would make more sense to be named

```suggestion
        column: Ident,
```

src/parser/mod.rs
```diff
@@ -853,14 +853,30 @@ impl<'a> Parser<'a> {
 
         Ok(Statement::ReleaseSavepoint { name })
     }
-
+    pub fn parse_column_with_data_type(&mut self) -> Result<Expr, ParserError> {
```
alamb, line 856:
```suggestion
    /// Parse clickhouse view parameter style `column1:data_type`
    pub fn parse_column_with_data_type(&mut self) -> Result<Expr, ParserError> {
```

src/parser/mod.rs
```diff
@@ -853,14 +853,30 @@ impl<'a> Parser<'a> {
...
     /// Parse an expression prefix
     pub fn parse_prefix(&mut self) -> Result<Expr, ParserError> {
         // allow the dialect to override prefix parsing
         if let Some(prefix) = self.dialect.parse_prefix(self) {
             return prefix;
         }
-
+        // ClickHouse support using {column:Datatype} format in its creation of parametrised views
+        match self.peek_token().token {
+            Token::LBrace if dialect_of!(self is ClickHouseDialect) => {
```
alamb, line 875: Can you please also support this syntax in `GenericDialect` as well as explained in https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#new-syntax ?

(also add test for it below)

tests/sqlparser_clickhouse.rs
```diff
@@ -394,3 +394,8 @@ fn clickhouse_and_generic() -> TestedDialects {
         options: None,
     }
 }
+
+#[test]
+fn parse_create_parametrised_views() {
+    clickhouse().verified_stmt("CREATE VIEW view AS SELECT * FROM A WHERE Column1 = {column1:datatype1} AND Column2 = {column2:datatype2}");
```
alamb, line 400: Can you also please test:
1. In generic dialect
2. A single column definition
3. Error case (like `where Column1 = {column1}`)
4. Error case (like `where`) (no column definition)

alamb (commented):
Thank you @Lordworms  -- this is a really nice contribution . I think it just needs a little polish and would be ready to go

fyi @flipchart

src/ast/mod.rs
Lordworms: Sure

tests/sqlparser_clickhouse.rs
Lordworms: Sure, I'll add these tommorrow

src/parser/mod.rs
Lordworms: sure

alamb (approved):
Thank you @Lordworms

alamb:
I took the liberty of fixing the compile error and pushing a commit to fix it

alamb:
Hi @Lordworms  -- it seems like there is a real test failure on this PR. Can you investigate?

Lordworms:
> Hi @Lordworms -- it seems like there is a real test failure on this PR. Can you investigate?

It seems to have some conflict with the DuckDB literal. I'll fix this.

Lordworms:
@alamb  I have resolved it and would you please take a look at it when you are available? Thanks!

tests/sqlparser_common.rs
```diff
@@ -8685,3 +8685,28 @@ fn parse_map_access_expr() {
...
     }
 }
+fn general_dialect() -> TestedDialects {
+    TestedDialects {
+        dialects: vec![Box::new(GenericDialect {})],
+        options: None,
+    }
+}
+#[test]
+fn parse_create_parametrised_views() {
```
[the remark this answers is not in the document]
Lordworms: Sure

alamb:
I am not sure what the current status of this PR is -- it has a bunch of conflicts and I am not quite sure if it is ready for another review. Sorry for the delay. Marking it as a draft until we sort that out so it isn't on my review queue

PR #1182, opened 2024-03-15
Fix parsing of equality binop in function argument

jmhain:
https://github.com/sqlparser-rs/sqlparser-rs/issues/1057 added support for using the `=` operator for named function arguments. This syntax conflicts with passing an `=` binop expr as a function argument, and so this caused a regression on dialects that don't support this style of named argument. There were two problems in particular:

1. `foo(a = b)` would parse as a named argument rather than a comparison of `a` and `b`.
2. `foo(1=1)` resulted in a parsing failure, which is problematic as many people have the habit of writing `1=1` instead of `TRUE`.

So rather than accepting this everywhere, we should only accept it on dialects that have this form of named argument. The original PR only mentions DuckDB so that's the only dialect I enabled it for. If there are any others, please let me know and I'll add them.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -68,7 +68,7 @@ impl TestedDialects {
                 }
                 Some((dialect, parsed))
             })
-            .unwrap()
+            .expect("tested dialects cannot be empty")
             .1
     }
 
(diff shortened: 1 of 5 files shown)


alamb (approved):
Thanks @jmhain  -- this looks good to me.

src/test_utils.rs
```diff
@@ -218,14 +209,30 @@ where
...
-            .into_iter()
-            .filter(|d| !except(d.as_ref()))
-            .collect(),
+        dialects: all_dialects,
         options: None,
     }
 }
 
+/// Returns all dialects matching the given predicate.
+pub fn all_dialects_where<F>(predicate: F) -> TestedDialects
```
alamb, line 218: I appreciate this improvement of the code while you were in here. Thank you

PR #1187, opened 2024-03-18
Fix parse `COPY INTO` stage names without parens for SnowFlake

mobuchowski:
`COPY INTO` should not assume that parens surround Snowflake's stage names.


alamb (approved):
Thank you very much @mobuchowski 

Sorry for the delay in reviewing

Issue #1186, opened 2024-03-18
Why custom Dialect doesn't work properly


groobyming:
I have extended a Dialect named MyDialect, and implemented the is_identifier_start method. is_identifier_start allows the identifier to start with a number, but why does my test code report an error? The prompt says that it cannot start with a number, and if I run it with the native HiveDialect, the code No problem again

**My Code：**
```
#[derive(Debug)]
pub struct MyDialect;

    impl Dialect for MyDialect {
    fn is_delimited_identifier_start(&self, ch: char) -> bool {
        (ch == '"') || (ch == '`')
    }

    fn is_identifier_start(&self, ch: char) -> bool {
        ch.is_ascii_lowercase() || ch.is_ascii_uppercase() || ch.is_ascii_digit() || ch == '$'
    }

    fn is_identifier_part(&self, ch: char) -> bool {
        ch.is_ascii_lowercase()
            || ch.is_ascii_uppercase()
            || ch.is_ascii_digit()
            || ch == '_'
            || ch == '$'
            || ch == '{'
            || ch == '}'
    }

    fn supports_filter_during_aggregation(&self) -> bool {
        true
    }
}

#[cfg(test)]
mod tests {
    use sqlparser::parser::Parser;

    #[test]
    pub fn test_ast() {
        let sql = "SELECT * from 1_a";
        let expected_sql = "SELECT * from 1_a"
            .replace("\n", "");
        let dialect = crate::core::ast::MyDialect {};
        let ast = Parser::parse_sql(&dialect, sql).unwrap();
        assert_eq!(ast[0].to_string(), expected_sql);
    }
}
```

**HIveDialect Code（from sqlparser-rs）**
use crate::dialect::Dialect;

/// A [`Dialect`] for [Hive](https://hive.apache.org/).
```
#[derive(Debug)]
pub struct HiveDialect {}

impl Dialect for HiveDialect {
    fn is_delimited_identifier_start(&self, ch: char) -> bool {
        (ch == '"') || (ch == '`')
    }

    fn is_identifier_start(&self, ch: char) -> bool {
        ch.is_ascii_lowercase() || ch.is_ascii_uppercase() || ch.is_ascii_digit() || ch == '$'
    }

    fn is_identifier_part(&self, ch: char) -> bool {
        ch.is_ascii_lowercase()
            || ch.is_ascii_uppercase()
            || ch.is_ascii_digit()
            || ch == '_'
            || ch == '$'
            || ch == '{'
[...]

PR #1188, opened 2024-03-20
Extended dialect triat to support numeric prefixed identifiers

groobyming:
Relates to  #1186. This PR includes changes to resolve the issue and enhance project functionality. Review appreciated.


src/tokenizer.rs
```diff
@@ -800,7 +800,9 @@ impl<'a> Tokenizer<'a> {
 
                     // mysql dialect supports identifiers that start with a numeric prefix,
                     // as long as they aren't an exponent number.
-                    if dialect_of!(self is MySqlDialect | HiveDialect) && exponent_part.is_empty() {
+                    if (dialect_of!(self is MySqlDialect | HiveDialect) && exponent_part.is_empty())
+                        || self.dialect.supports_numeric_prefix()
```
alamb, line 804: I wonder why the check for `self.dialect.supports_numeric_prefix()` doesn't also check `exponent_part.is_empty`? 

That would mean that tokens like `1e4` are going to be parsed as identifiers

What if you wrote this check like

```rust
  if self.dialect.supports_numeric_prefix() && exponent_part.is_empty()) {
```

And then changed `MySqlDialect` and `HiveDialect` to return true for `supports_numeric_prefix`?

src/dialect/mod.rs
```diff
@@ -143,6 +143,10 @@ pub trait Dialect: Debug + Any {
     fn supports_start_transaction_modifier(&self) -> bool {
         false
     }
+    /// Returns true if the dialect supports identifiers that start with a numeric prefix
```
alamb, line 146: Could you please add an example of such an identifier? (e.g. `123Column`)?

alamb (commented):
Thank you for this contribution @groobyming  -- I have a suggestion that I think would make the code cleaner and perhaps also work more like how you want

I apologize for the delay in reviewing

alamb:

src/tokenizer.rs
groobyming: Great suggestion! I overlooked this scenario, and I'll fix this logic issue

src/dialect/mod.rs
groobyming: Our scenario will use numbers as business ids. The rule for table names is the business id plus an underscore followed by the specific table name. 
eg：59901_user_login

alamb (approved):
Thanks again @groobyming  -- I took the liberty of moving the check to be entirely trait based and merging up to resolve conflicts.

PR #1191, opened 2024-03-23
Support more `DateTimeField` variants

iffyio:
- Adds `DATETIME` variant to `DatetimeField`
- Adds support for snowflake abbreviations via `Custom` variant. https://docs.snowflake.com/en/sql-reference/functions-date-time#supported-date-and-time-parts
- Adds support for BigQuery weekday `WEEK(MONDAY)` syntax https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#extract

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -12,6 +12,13 @@
 
 #[cfg(not(feature = "std"))]
 use alloc::string::String;
+
+#[cfg(not(feature = "std"))]
+use alloc::format;
+
+#[cfg(not(feature = "std"))]
+use alloc::string::ToString;
+
 use core::fmt;
 
 #[cfg(feature = "bigdecimal")]
@@ -20,6 +27,7 @@ use bigdecimal::BigDecimal;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+use crate::ast::Ident;
 #[cfg(feature = "visitor")]
 use sqlparser_derive::{Visit, VisitMut};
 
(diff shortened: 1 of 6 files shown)


src/ast/value.rs
```diff
@@ -148,47 +164,67 @@ pub enum DateTimeField {
...
-            DateTimeField::Timezone => "TIMEZONE",
-            DateTimeField::TimezoneAbbr => "TIMEZONE_ABBR",
-            DateTimeField::TimezoneHour => "TIMEZONE_HOUR",
-            DateTimeField::TimezoneMinute => "TIMEZONE_MINUTE",
-            DateTimeField::TimezoneRegion => "TIMEZONE_REGION",
-            DateTimeField::NoDateTime => "NODATETIME",
-        })
+        f.write_str(
+            match self {
+                DateTimeField::Year => "YEAR".to_string(),
```
alamb, line 180: This change to call `to_string()` will now cause a new string to be allocated for each type.

I think it would be better / more effiicent to use `write!` directly -- like

```suggestion
              match self {
                DateTimeField::Year => f.write_str("YEAR"),
                DateTimeField::Month => f.write_str("MONTH"),
                ...
```

src/ast/value.rs
```diff
@@ -12,6 +12,13 @@
 
 #[cfg(not(feature = "std"))]
 use alloc::string::String;
+
+#[cfg(not(feature = "std"))]
```
alamb, line 16: If you use the suggestion below I think you will not need these imports either

alamb (commented):
Thank you @iffyio (and I very much apologize for the delay in reviewing). The only thing I think this PR needs is to avoid creating intermediate strings in the `Display` fmt

alamb (commented):
In order to keep the code flowing (and the fact it took me so long to review this PR) I fixed the string copy issue and pushed a fix to this PR.

Thanks again @iffyio

alamb (approved):

src/ast/value.rs
iffyio: Ah yes that makes a lot more sense, thanks for the follow up fix too!

alamb: (for anyone following along it was https://github.com/sqlparser-rs/sqlparser-rs/pull/1209)

PR #1192, opened 2024-03-24
BigQuery support inline comment with hash syntax

iffyio:
Adds support for single line comments using `#` syntax. GenericDialect isn't included because the syntax conflicts with Postgres' JSON access syntax `#>`

https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#single_line_comments


alamb (approved):
Thanks @iffyio  -- 🙏

PR #1193, opened 2024-03-24
Fix dollar quoted string tokenizer

ZacJW:
Currently the tokenizer is written in a way that assumes that once it's seen the starting delimiter (the first `$...$`) the next $ it sees must be the start of the ending delimiter, but this needn't be the case. This change checks to see if what might be the end delimiter actually matches, and if it doesn't considers it part of the string and keeps searching for the end.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1139,7 +1139,7 @@ impl<'a> Tokenizer<'a> {
                                     return self.tokenizer_error(
                                         chars.location(),
                                         "Unterminated dollar-quoted, expected $",
-                                    )
+                                    );
                                 }
                             }
                             if chars.peek() == Some(&'$') {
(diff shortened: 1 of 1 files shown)


alamb (commented):
Thanks for this contribution @ZacJW 

Can you please add some tests illustrating the behavior of this change (and so we don't accidentally break it in a future refactor).

Please mark it as ready for review when it is ready for another look

Thanks again and I am sorry for the delay

ZacJW:
I've added four unit tests:
- Dollar quoted with tag (`$tag$`)
- Dollar quoted without tag (`$$`)
- Unterminated dollar quoted with tag (tokenization is expected to fail)
- Unterminated dollar quoted without tag (tokenization is expected to fail)

In writing the last one I noticed that the error message is different. Before my change there were four different error messages about the string being unterminated. Was this deliberate so an error could be tracked back to the site that created it?

alamb (approved):
Thank you again for this contribution @ZacJW  -- the test cases really help and the code looks 👌

PR #1194, opened 2024-03-26
Add support for DuckDB struct literal syntax

### LINKED ISSUE #1129: support more dialects of `struct` creation
Currently the only struct syntax that is support is bigquery's dialect. 

It'd be nice to support duckdb's struct literal syntax https://duckdb.org/docs/sql/data_types/struct. 

see also
https://github.com/GlareDB/glaredb/issues/2112

gstvg:
Fixes #1129

Existing Struct expression supports the BigQuery semantics, which allow for optionally typed fields, that if typed, also have an optional name.
DuckDB has incompatible semantics: only named fields without any type specification.

Should I create a separate expression, perharps `NamedFieldsStruct`, or change the `Struct` to have an inner enum `Struct(enum { MaybeTyped(...), Named(...) })` ?

Thanks!

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -347,6 +347,23 @@ impl fmt::Display for StructField {
     }
 }
 
+/// A dictionary field within a dictionary.
+///
+/// [duckdb]: https://duckdb.org/docs/sql/data_types/struct#creating-structs
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct DictionaryField {
+    pub key: Ident,
+    pub value: Box<Expr>,
+}
+
+impl fmt::Display for DictionaryField {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}: {}", self.key, self.value)
+    }
+}
+
 /// Options for `CAST` / `TRY_CAST`
 /// BigQuery: <https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax>
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -669,21 +686,12 @@ pub enum Expr {
     /// ```sql
     /// STRUCT<[field_name] field_type, ...>( expr1 [, ... ])
     /// ```
-    /// `DuckDB` specific `Struct` literal expression [2]
-    /// Syntax:
-    /// ```sql
-    /// syntax: {'field_name': expr1[, ... ]}
-    /// ```
     /// [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
-    /// [2]: https://duckdb.org/docs/sql/data_types/struct#creating-structs
     Struct {
         /// Struct values.
         values: Vec<Expr>,
         /// Struct field definitions.
         fields: Vec<StructField>,
-        /// true if uses duckdb array notation syntax(no change in semantics)
-        /// this field is used for formatting
-        array_notation: bool,
     },
     /// `BigQuery` specific: An named expression in a typeless struct [1]
     ///
@@ -696,6 +704,13 @@ pub enum Expr {
         expr: Box<Expr>,
         name: Ident,
     },
+    /// `DuckDB` specific `Struct` literal expression [1]
+    /// Syntax:
+    /// ```sql
+    /// syntax: {'field_name': expr1[, ... ]}
+    /// ```
+    /// [1]: https://duckdb.org/docs/sql/data_types/struct#creating-structs
+    Dictionary(Vec<DictionaryField>),
     /// An array index expression e.g. `(ARRAY[1, 2])[1]` or `(current_schemas(FALSE))[1]`
     ArrayIndex {
         obj: Box<Expr>,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -669,12 +669,21 @@ pub enum Expr {
...
     /// [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
+    /// [2]: https://duckdb.org/docs/sql/data_types/struct#creating-structs
     Struct {
         /// Struct values.
         values: Vec<Expr>,
         /// Struct field definitions.
         fields: Vec<StructField>,
+        /// true if uses duckdb array notation syntax(no change in semantics)
+        /// this field is used for formatting
+        array_notation: bool,
```
jayzhan211, line 686: I think it is almost different than the existing struct,  I would prefer we introduce a new Expr.
Note that `Map` in Duckdb has the similar syntax, so I don't think we should name it struct-specific.

I would call it `Expr::Dict` por `Expr::Dictionary`
Screenshot 2024-03-27 at 08 02 07

jayzhan211:
You can ask the question in issue. The PR draft doesn't send the notification.

alamb:
> Should I create a separate expression, perharps NamedFieldsStruct, or change the Struct to have an inner enum Struct(enum { MaybeTyped(...), Named(...) }) ?

I think given the difference in syntax, having a new struct like `NamedFieldsStruct` would be better / less confusing / easier to handle

Thank you again for working on this @gstvg  and @jayzhan211

gstvg:
thanks @jayzhan211 and @alamb
I created a new Expr named Dictionary, which currently only supports struct literals.

I tried adding supporting for DuckDB `MAP` too, but since map keys can be any expression, it conflicted with Snowflake-only JsonAccess with colon `json_field:key`, which is currently parsed for all dialects. Restricting it to Snowflake solves the problem. Should i push it here PR too?

jayzhan211:
> thanks @jayzhan211 and @alamb I created a new Expr named Dictionary, which currently only supports struct literals.
> 
> I tried adding supporting for DuckDB `MAP` too, but since map keys can be any expression, it conflicted with Snowflake-only JsonAccess with colon `json_field:key`, which is currently parsed for all dialects. Restricting it to Snowflake solves the problem. Should i push it here PR too?

I don't understand what you mean by Snowflake-only JsonAccess with colon `json_field:key`, can you elaborate it more.
It seems the difference between struct and map in Duckdb is that the `key_type` can be different for map in each row. They are both json-like objects.

gstvg:
Sure. I would like to solve all this at once, to be honest.

Snowflake JSON access uses `:` to query data

image
[https://docs.snowflake.com/en/user-guide/json-basics-tutorial-query](url)

DuckDB maps key can be any type, not only strings, but may have different set of keys for each row

image
[https://duckdb.org/docs/sql/data_types/map.html](url)

To support any type of key, I used `Parser::parse_expr` instead of `parse_identifier` 

Currently, because of the following arms in [parse_infix](https://github.com/sqlparser-rs/sqlparser-rs/blob/44727891713114b72546facf21a335606380845a/src/parser/mod.rs#L2409) and [get_next_precedence](https://github.com/sqlparser-rs/sqlparser-rs/blob/44727891713114b72546facf21a335606380845a/src/parser/mod.rs#L2646), when `parse_expr` is called to get the key expr, it parses the whole `key: value` as a JsonAccess

If those two arms are moved to the Snowflake dialect, or conditioned to a `dialect_of!(Snowflake)`, parse_expr will return the key only, not the whole key value, but the `json_column:json_key` would be restricted to the Snowflake dialect

I didn't find any other way to do that

I've pushed it to another [branch](https://github.com/gstvg/sqlparser-rs/tree/duckdb_struct_map) if you have interest


> > which is currently parsed for all dialects

> I think we have not yet supported any.

Sorry, I mean the `json_column:json_key` is parsed for all dialects, not restricted to the Generic and/or Snowflake

gstvg:
Others dialects that uses this syntax:

[Snowflake](https://docs.snowflake.com/en/sql-reference/data-types-semistructured#object-constants)
image
Syntax limited to Object constants, so `parse_value` or `parse_identifier` can be called instead of `parse_expr`


[ClickHouse](https://clickhouse.com/docs/en/sql-reference/data-types/map)
image
Has the same conflicts as DuckDB map, which also means it has the same solution 😀

jayzhan211:
> How about considering `Expr::JsonAcess` as your `Expr::DictionaryField`, it seems to me they are equivalent but the json operator is `colon`. Similar to parsing `[e1, e2...]` but `{e1, e2...}` instead.
> 
> `Expr::Dictionary(Vec<Expr>)`, where we can get Expr::JsonAccess (left colon right)

I met the issue while trying out my idea #1197

alamb (approved):
I think this looks good to me: the use of `parse_expr` and restricting it to DuckDB makes sense to me

If/When people need support for this type in Snowflake and the other dialects, I think we can handle them separately

alamb:
I also pushed d9d16ce to fix the doc test failues

alamb:
@jayzhan211  does this change look good for you?

jayzhan211:
I think either struct and map and other dictionaries like syntax `{"key": "value"}` is possible to share the same parsing rule.
It is fine to do it specifically for now, but I think we can share the similarity for other dialects or syntax later on.

alamb:
Thanks @jayzhan211  and @gstvg !

gstvg:
> How about considering `Expr::JsonAcess` as your `Expr::DictionaryField`, it seems to me they are equivalent but the json operator is `colon`. Similar to parsing `[e1, e2...]` but `{e1, e2...}` instead.
> 
> `Expr::Dictionary(Vec<Expr>)`, where we can get Expr::JsonAccess (left colon right)

Good idea. But making Dictionary a `Vec<Expr>` wouldn't be unergonomic for users? Extending it, what about something like this:

```
    fn parse_duckdb_dictionary_field(&mut self) -> Result<DictionaryField, ParserError> {
        if let Expr::JsonAccess { left, operator: JsonOperator::Colon, right } = self.parse_expr()? {
            Ok(DictionaryField {
                key: left,
                value: right,
            })
        } else {
            self.expected(":", self.peek_token())
        }
    }
```

jayzhan211:
> > How about considering `Expr::JsonAcess` as your `Expr::DictionaryField`, it seems to me they are equivalent but the json operator is `colon`. Similar to parsing `[e1, e2...]` but `{e1, e2...}` instead.
> > `Expr::Dictionary(Vec<Expr>)`, where we can get Expr::JsonAccess (left colon right)
> 
> Good idea. But making Dictionary a `Vec<Expr>` wouldn't be unergonomic for users? Extending it, what about something like this:
> 
> ```
>     fn parse_duckdb_dictionary_field(&mut self) -> Result<DictionaryField, ParserError> {
>         if let Expr::JsonAccess { left, operator: JsonOperator::Colon, right } = self.parse_expr()? {
>             Ok(DictionaryField {
>                 key: left,
>                 value: right,
>             })
>         } else {
>             self.expected(":", self.peek_token())
>         }
>     }
> ```

It looks good to me.

PR #1195, opened 2024-03-26
Add support for DuckDB functions named arguments with assignment operator

gstvg:
[Some nested DuckDB functions use assignment operator](https://duckdb.org/docs/sql/functions/nested.html)

```
struct_insert(struct, name := any, ...)
struct_pack(name := any, ...)
union_value(tag := any)
```

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4556,6 +4556,8 @@ pub enum FunctionArgOperator {
     Equals,
     /// function(arg1 => value1)
     RightArrow,
+    /// function(arg1 := value1)
+    Assignment,
 }
 
 impl fmt::Display for FunctionArgOperator {
@@ -4563,6 +4565,7 @@ impl fmt::Display for FunctionArgOperator {
         match self {
             FunctionArgOperator::Equals => f.write_str("="),
             FunctionArgOperator::RightArrow => f.write_str("=>"),
+            FunctionArgOperator::Assignment => f.write_str(":="),
         }
     }
 }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4557,7 +4557,7 @@ pub enum FunctionArgOperator {
     /// function(arg1 => value1)
     RightArrow,
     /// function(arg1 := value1)
-    Assignment,
+    DuckAssignment,
 }
 
 impl fmt::Display for FunctionArgOperator {
@@ -4565,7 +4565,7 @@ impl fmt::Display for FunctionArgOperator {
         match self {
             FunctionArgOperator::Equals => f.write_str("="),
             FunctionArgOperator::RightArrow => f.write_str("=>"),
-            FunctionArgOperator::Assignment => f.write_str(":="),
+            FunctionArgOperator::DuckAssignment => f.write_str(":="),
         }
     }
 }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4557,7 +4557,7 @@ pub enum FunctionArgOperator {
     /// function(arg1 => value1)
     RightArrow,
     /// function(arg1 := value1)
-    DuckAssignment,
+    Assignment,
 }
 
 impl fmt::Display for FunctionArgOperator {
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -4556,13 +4556,16 @@ pub enum FunctionArgOperator {
     Equals,
     /// function(arg1 => value1)
     RightArrow,
+    /// function(arg1 := value1)
+    Assignment,
```
universalmind303, line 4560: can we rename this to `DuckAssignment` just to be consistent with the tokenizer naming. 

```suggestion
    DuckAssignment,
```

gstvg: Of course, it's done!

universalmind303 (approved):

src/ast/mod.rs
```diff
@@ -4556,13 +4556,16 @@ pub enum FunctionArgOperator {
     Equals,
     /// function(arg1 => value1)
     RightArrow,
+    /// function(arg1 := value1)
+    DuckAssignment,
```
jayzhan211, line 4560: I suggest a general name like Assignment

gstvg: I renamed both Token and FunctionArg to Assignment , since [snowflake uses it too](https://github.com/gstvg/sqlparser-rs/blob/e8ad6af36b7019cfee181101fb47463a944d64eb/src/parser/mod.rs#L4143)

What do you think @universalmind303 ?

If is best to split the Token rename to another PR I can do it, just let me know, thanks!

jmhain:
Other dialects might use `:=` to mean something different, for example on mysql, it appears to be an expression that can be used anywhere in a query [to assign to a user-defined variable](https://dev.mysql.com/doc/refman/8.3/en/assignment-operators.html), e.g. in `SELECT sqrt(@var1 := 1), @var1`. So this should be put behind a dialect method similar to what I did here to fix a related regression on Snowflake: https://github.com/sqlparser-rs/sqlparser-rs/pull/1182

gstvg:
> Other dialects might use `:=` to mean something different, for example on mysql, it appears to be an expression that can be used anywhere in a query [to assign to a user-defined variable](https://dev.mysql.com/doc/refman/8.3/en/assignment-operators.html), e.g. in `SELECT sqrt(@var1 := 1), @var1`. So this should be put behind a dialect method similar to what I did here to fix a related regression on Snowflake: #1182

Good catch @jmhain, thanks!
Is possible to use `dialect_of!(DuckDBDialect)` instead of adding a new Dialect method?

jmhain:
@gstvg Yeah that sounds fine to me!

src/parser/mod.rs
```diff
@@ -8602,6 +8602,18 @@ impl<'a> Parser<'a> {
                 arg,
                 operator: FunctionArgOperator::Equals,
             })
+        } else if dialect_of!(self is DuckDbDialect) && self.peek_nth_token(1) == Token::Assignment
+        {
+            let name = self.parse_identifier(false)?;
+
+            self.expect_token(&Token::Assignment)?;
+            let arg = self.parse_wildcard_expr()?.into();
```
jayzhan211, line 8610: is wildcard included?

gstvg: Probably not, thanks. To be honest, I just copied from the code few lines above which does the same for `=>` and `=`. Changing them to `parse_expr` doesn't break any test. Don't know if they need it too.
Anyway, i changed mine to `parse_expr`, thanks again 🙏

src/parser/mod.rs
```diff
@@ -8602,6 +8602,18 @@ impl<'a> Parser<'a> {
                 arg,
                 operator: FunctionArgOperator::Equals,
             })
+        } else if dialect_of!(self is DuckDbDialect) && self.peek_nth_token(1) == Token::Assignment
```
alamb: Normally we also support Generic dialect as well for syntaxes. I took the liberty of adding this in another commit

alamb (approved):
Thank you for the contribution @gstvg  -- sorry I am behind and it took so long to review this

PR #1200, opened 2024-04-01
fix: support UTF-8 char as escape_char

TennyZhuang:


alamb (commented):
Thank you for the contribution @TennyZhuang 

Can you please:
1. Remove the itertools dependency (we try to keep the required dependencies very low in this crate as it is quite low level)
2. Add a test showing what behavior this is fixing (as well as to ensure we don't break this feature accidentally in the future)

alamb:

PR #1202, opened 2024-04-06
Support  `PARALLEL ... and for `..ON NULL INPUT ...` to `CREATE FUNCTION`

dimfeld:
Adds two sets of keywords for CREATE FUNCTION, as described at https://www.postgresql.org/docs/current/sql-createfunction.html 

## Parallel-save

- PARALLEL SAFE
- PARALLEL UNSAFE
- PARALLEL RESTRICTED

## Called on null input

- CALLED ON NULL INPUT
- RETURNS NULL ON NULL INPUT
- STRICT


alamb (approved):
Looks good to me -- thank you for the contribution @dimfeld

alamb:
Sorry for the delayed review. I am behind

Issue #1211, opened 2024-04-11
Error on some `GRANT`s when trailing commas are enabled


MohamedAbdeen21:
# How to reproduce

```rs
use sqlparser::{
    dialect::GenericDialect,
    parser::{Parser, ParserError, ParserOptions},
};

fn main() -> Result<(), ParserError> {
    let dialect = GenericDialect {};
    let options = ParserOptions::new().with_trailing_commas(true);

    let result = Parser::new(&dialect)
        .with_options(options)
        .try_with_sql("GRANT USAGE, SELECT ON SEQUENCE p TO u")?
        .parse_statements();

    println!("{:?}", result);

    Ok(())
}
```

# Output
```rs
Err(ParserError("Expected ON, found: SELECT at Line: 1, Column 14"))
```

# Expected Output

Should parse successfully.

```rs
Ok([Grant { privileges: Actions([Usage, Select { columns: None }]), objects: Sequenc
es([ObjectName([Ident { value: "p", quote_style: None }])]), grantees: [Ident { valu
e: "u", quote_style: None }], with_grant_option: false, granted_by: None }])
```

# Cause

`parse_comma_separated` guesses that the comma is a trailing comma since it's followed by a reserved keyword (SELECT).

PR #1212, opened 2024-04-11
Enhancing Trailing Comma Option

MohamedAbdeen21:
# Fixes

This change addresses #1211 and [arrow-datafusion#9949](https://github.com/apache/arrow-datafusion/issues/9949).

# Changes included

## Added
1. Add `supports_trailing_commas()` and `supports_projection_trailing_commas()` to the `Dialect` trait, enabling `supports_projection_trailing_commas()` for `BigQuery` and `SnowFlake`, and `supports_trailing_commas()` for `DuckDb` [link](https://duckdb.org/2022/05/04/friendlier-sql.html#trailing-commas)
2. Use new `Dialect` trait methods to set `trailing_comma` parser option when creating new parsers using `Parser::new(dialect)`, instead of defaulting to `false` for all dialects and waiting for the user to set it manually through `ParserOptions`.
3. Error on trailing commas in projection list when dialect doesn't support it (Note that this doesn't break backwards compatibility)  *.

## Bug fixes
4. Handling the case mentioned in #1211, by temporarily disabling `trailing_comma` option when parsing `GRANT` statements. *

## Breaking changes
5. Trailing commas were allowed in all create statements, regardless of the dialect. This should not be the case since this behavior is dialect specific. `trailing_comma` is now considered when parsing create statements.

\* I can use some help regarding changes 3 and 4:
3. The keyword is hard-coded, and only works for projection lists.
4. This may cause issues if the user expects the `GRANT` statement to allow trailing commas.

Please, if you have other alternative approaches or suggestions, be sure to share them.

# Are these changes tested?

Yes


src/dialect/mod.rs
```diff
@@ -157,6 +157,14 @@ pub trait Dialect: Debug + Any {
         // return None to fall back to the default behavior
         None
     }
+    /// Does the dialect support trailing commas around the query?
+    fn supports_trailing_commas(&self) -> bool {
```
alamb, line 161: `supports_trailing_commas` suggests to me that it is supporting trailing commas everywhere, and `supports_projection_trailing_commas` suggests just for projection

If `supports_trailing_commas`  means just for queries, can you please name it to something more specific?

tests/sqlparser_common.rs
```diff
@@ -8419,6 +8421,40 @@ fn parse_trailing_comma() {
...
+        options: None,
+    };
+
+    assert!(trailing_commas
+        .parse_sql_statements("SELECT name, age, FROM employees;")
+        .is_err());
+}
+
+#[test]
+fn parse_porjection_trailing_comma() {
```
alamb, line 8437:
```suggestion
fn parse_projection_trailing_comma() {
```

alamb (commented):
Thanks @MohamedAbdeen21 

The idea sounds good and I left a few small comments

I think the biggest concern I have is now there seems to two  ways ways of setting trailing comma support -- globally via `ParserOptions` and then per dialect

I think if we want to make trailing comma support part of the dialect, we should remove the parser options and just use the dialect trait everywhere

cc @ankrgyl  who I think added trailing comma support initially

src/dialect/mod.rs
MohamedAbdeen21: This option should try to allow trailing commas everywhere. I'm not really following what you mean by "just for queries"

MohamedAbdeen21:
> I think the biggest concern I have is now there seems to two ways ways of setting trailing comma support -- globally via ParserOptions and then per dialect

I think that's completely ok, given that setting it manually overrides the dialect settings. This can give the users flexibility when choosing a dialect for their projects.

> I think if we want to make trailing comma support part of the dialect, we should remove the parser options and just use the dialect trait everywhere

While I don't like how we have an entire `ParserOptions` struct that only has 2 attributes, I can see how customizing behaviors can be a desirable feature. Also, this will break backwards compatibility since `ParserOptions` is a pub struct.

src/dialect/mod.rs
MohamedAbdeen21: I noticed something while going over the tests, Create table by default allows trailing commas, because it doesn't use the `parse_comma_separated` function, which was added later.

tests/sqlparser_common.rs
```diff
@@ -3354,8 +3354,13 @@ fn parse_create_table_clone() {
 
 #[test]
 fn parse_create_table_trailing_comma() {
-    let sql = "CREATE TABLE foo (bar int,)";
-    all_dialects().one_statement_parses_to(sql, "CREATE TABLE foo (bar INT)");
+    let dialect = TestedDialects {
```
alamb, line 3357: I am worried about this change -- I think it means that the trailing comma will no longer be supported by all dialects

src/parser/mod.rs
```diff
@@ -292,7 +292,7 @@ impl<'a> Parser<'a> {
             index: 0,
             dialect,
             recursion_counter: RecursionCounter::new(DEFAULT_REMAINING_DEPTH),
-            options: ParserOptions::default(),
+            options: ParserOptions::new().with_trailing_commas(dialect.supports_trailing_commas()),
```
alamb, line 295: this seems like a good change

src/parser/mod.rs
```diff
@@ -8530,6 +8535,9 @@ impl<'a> Parser<'a> {
                 with_privileges_keyword: self.parse_keyword(Keyword::PRIVILEGES),
             }
         } else {
+            let old_value = self.options.trailing_commas;
```
alamb, line 8538: I think this is a change we should discuss in a separate PR as I think it will cause some statements that used to parse to stop working (even if though it is more consistent)

alamb (commented):
Thanks again for this PR @MohamedAbdeen21  -- I am sorry it took so long to provide a review. I am quite worried about the change in existing behavior for parsing trailing commas. Can you please fix those and I'll give it another look?

src/parser/mod.rs
MohamedAbdeen21: Although this was the original target for the PR, I lost track of the issue and ended up trying to address multiple points at once. I'm fine with delegating this to a separate PR.

tests/sqlparser_common.rs
MohamedAbdeen21: Note that all dialects allowed trailing commas only in create statements because create statements are parsed by a special function and not by the `parse_comma_separated` that actually considers the `trailing_commas` option. So that was a bug not a feature; however, as mentioned before, that behavior can be restored using the parsingOptions struct.

PS: I pointed that out in our previous conversation and recommended a follow-up to migrate the create statement to use the general `parse_comma_separated` function.

src/parser/mod.rs
MohamedAbdeen21: I don't see how that will break any existing statements. The original behavior was that if trailing commas were enabled and a DCL used 'SELECT', the parsing would fail.

This change temporarily ignores the trailing commas option when parsing DCL.

This is definitely not an ideal solution and should have its own PR (and therefore the corresponding issue should not be closed, even tho this was supposed to be the main target issue), but this has to be pushed with the new dialect trait method to pass the tests.

MohamedAbdeen21:
Thanks for the review @alamb, the only breaking change here is the create statements when trailing_commas are not enabled in the parserOptions, which I think was a bug mistaken for a feature.

The other change with the DCL doesn't affect the current behavior; if anything, statements that used to fail will now parse.

I'm sorry that the PR is addressing multiple points at once, with some of the changes being temporary solutions that require follow-ups. I'll try to make future PRs more focused to ease the review process.

lovasoa (approved):

alamb (approved):
Thank you @MohamedAbdeen21  and @lovasoa

MohamedAbdeen21:
Thanks for the review! @alamb @lovasoa

alamb:
Sorry for the delay -- I am spread quite thin these days -- though we are discussing a few solutions to improve the situation https://github.com/sqlparser-rs/sqlparser-rs/issues/1294

PR #1217, opened 2024-04-14
Support BigQuery `MERGE` syntax

iffyio:
This extends the parser for `MERGE` statements with support for BigQuery specific constructs.
- `WHEN NOT MATCHED BY TARGET`
- `WHEN NOT MATCHED BY SOURCE`
- Inserts via the `ROW` keyword

https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement


src/ast/mod.rs
```diff
@@ -5591,72 +5594,175 @@ impl fmt::Display for CopyLegacyCsvOption {
...
-        predicate: Option<Expr>,
-        columns: Vec<Ident>,
-        values: Values,
-    },
+pub enum MergeInsertExpr {
+    /// The expression is defined from an explicit `VALUES` clause
+    ///
+    /// Example:
+    /// ```sql
+    /// INSERT (product, quantity) VALUES(product, quantity)
```
jmhain, line 5652: According to SF docs, the column list for this is optional:

```plaintext
notMatchedClause ::=
   WHEN NOT MATCHED [ AND <case_predicate> ] THEN INSERT [ ( <col_name> [ , ... ] ) ] VALUES ( <expr> [ , ... ] )
```

So it might make sense to represent this as:

```rust
enum MergeInsertKind {
    Values(Values),
    Row,
}

enum MergeInsertExpr {
    columns: Vec<Ident>, // empty means column list absent
    kind: MergeInsertKind,
}
```

src/parser/mod.rs
```diff
@@ -9528,61 +9541,81 @@ impl<'a> Parser<'a> {
...
+                            MergeInsertExpr::Values { columns, values }
+                        }
+                    };
+                    MergeAction::Insert(insert_expr)
+                }
+                Some(_) => {
+                    return Err(ParserError::ParserError(
+                        "expected UPDATE, DELETE or INSERT in merge clause".to_string(),
+                    ));
+                }
```
jmhain, line 9607:
```suggestion
                Some(_) => unreachable!(),
```

src/ast/mod.rs
iffyio: Updated to use this repr!

src/parser/mod.rs
iffyio: Updated, I've merged the `Some(_) | None` clause, it still returns an error, I figure that's preferred here vs panicking in case a mistake slips in

jmhain:
This looks good to me! cc @alamb

alamb (approved):
Thanks @iffyio  and @jmhain  for the review 🙏

PR #1218, opened 2024-04-14
Accept WINDOW clause after QUALIFY when parsing

iffyio:
The positioning of QUALIFY and WINDOW clauses differ between dialects.
BigQuery requires that WINDOW clauses come after QUALIFY which fails on the parser as a result.
This updates the parser to accept the clauses in either order.
~It's also possible to flag this behavior on the dialect level so that we don't parse one for the other (didn't opt for that in this case because it felt like a minor detail but probably make more sense), let me know if that's preferred!~ Actually found some dialects like databricks can accept either order so probably makes sense to have the parser accept in any order


alamb (approved):
Looks good to me -- thanks @iffyio

alamb:
This branch has conflicts so I couldn't merge it directly. Instead i made https://github.com/sqlparser-rs/sqlparser-rs/pull/1248 with the same commit

PR #1219, opened 2024-04-14
Support for MSSQL `CONVERT` styles

iffyio:
Adds support to accept extra parameters on the CONVERT function

https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16#style

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -543,6 +543,10 @@ pub enum Expr {
         charset: Option<ObjectName>,
         /// whether the target comes before the expr (MSSQL syntax)
         target_before_value: bool,
+        /// How to translate the expression.
+        ///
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -543,6 +543,10 @@ pub enum Expr {
         charset: Option<ObjectName>,
         /// whether the target comes before the expr (MSSQL syntax)
         target_before_value: bool,
+        /// How to translate the expression.
+        ///
+        /// [MSSQL]: https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16#style
```
alamb, line 548: https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16#style is 🤯

alamb (approved):
Looks good to me -- thank you @iffyio

PR #1221, opened 2024-04-14
Preserve double colon casts (and simplify cast representations)

jmhain:
It's been a slight annoyance to me that `::` casts are converted to `CAST()`-style casts when rewriting SQL using this crate, so this PR makes it so we preserve the cast syntax used in the AST. We also had three AST nodes for `CAST`, `TRY_CAST`, and `SAFE_CAST` which didn't seem necessary, so I reduced this to just `Expr::Cast` with a `kind` field that determines the particular syntax / mode of cast used.

This change seems to be well covered by existing tests, but let me know if you have any suggestions for additional tests I could add here.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -408,6 +408,26 @@ impl fmt::Display for MapAccessKey {
     }
 }
 
+/// The syntax used for in a cast expression.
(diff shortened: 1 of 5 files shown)


iffyio:
LGTM! cc @alamb

alamb (approved):

alamb:
Thanks @jmhain  and @iffyio  🚀

PR #1222, opened 2024-04-15
Support Snowflake `MATCH_RECOGNIZE` syntax

jmhain:


src/ast/query.rs
```diff
@@ -823,6 +823,238 @@ pub enum TableFactor {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} AS {}", self.symbol, self.definition)
+    }
+}
+
+/// A symbol in a `MATCH_RECOGNIZE` pattern.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum Symbol {
```
iffyio, line 966: One thing I'm wondering not sure if its actually preferrable - for Symbol, Pattern and Quantifier enums, would it be reasonable to prefix them with e.g `MatchRecognizeSymbol` etc? Not sure if it matters in practice today, just felt reading it that given those names are quite generic, there's potential that it causes ambiguities down the line. I think they can also be left as is if you feel its clearer this way and we can always adjust later on if needed

src/parser/mod.rs
```diff
@@ -8357,10 +8358,241 @@ impl<'a> Parser<'a> {
                 }
             }
 
+            if self.dialect.supports_match_recognize()
```
iffyio, line 8361: should this use the `dialect_of` check instead or is the trait method required for this feature?

tests/sqlparser_common.rs
```diff
@@ -8685,3 +8685,263 @@ fn parse_map_access_expr() {
...
+    };
+    fn check(options: &str, expect: TableFactor) {
+        let select = all_dialects_where(|d| d.supports_match_recognize()).verified_only_select(
+            &format!("SELECT * FROM my_table MATCH_RECOGNIZE({options})"),
+        );
+        assert_eq!(&select.from[0].relation, &expect);
+    }
+    check(
+        concat!(
+            "PARTITION BY company ",
```
iffyio, line 8707: can we add some simpler roundtrip-only tests that parses statements without the optional parts of match recognize like `partition by`?

iffyio (commented):
thanks @jmhain! I think this looks good, left a couple minor comments/questions

src/ast/query.rs
jmhain: Yeah these names are a bit generic, thanks for pointing that out. Pushed a commit to clean this up.

src/parser/mod.rs
jmhain: I prefer to use the trait method here because it can be reused in the test with `all_dialects_where(|d| d.supports_match_recognize()`. This way if it's added for another dialect in the future (e.g. [Trino](https://trino.io/docs/current/sql/match-recognize.html)), those tests will run without having to remember to add the new dialect to them.

iffyio: Ah I see, not sure if we have a guideline for this but feels to me like the dialect trait could get bloated quickly otherwise. Not sure though, deferring to @alamb if you have thoughts (and the PR looks good to me for a review too!)

jmhain: Yeah that's a good point. We could do something like this:

```
struct FeatureSupport {
    match_recognize: bool,
    .. options
}

trait Dialect {
    fn features() -> &'static FeatureSupport;

    // methods requiring custom code
}
```

If `DialectFeatureSupport` gets too big, we can make it hierarchical.

alamb: I don't have a great story for this question -- it seems to me that the crate is currently inconsistent where sometimes it uses a trait method and sometimes uses the hard coded `IsDialect` .. 

I think the `FeatureSupport` idea is neat, and if we didn't already have 2 patterns it would probably be the way to go. However, given we already have 2 methods, I think adding a third would be less than ideal 

Ideally we would have a single way and I think the best thing for users of the crate is methods on Dialect (mostly so they can define their own custom dialects, which I have seen some of them do). However, I fully acknowledge that the trait will get very large for sure.

However, a wholesale refactoring of the crate is probably not feasible at this time. 

If we can get some consensus between the three of us maybe I can add something to the documentation on the preferred method (dialectof or trait method)

tests/sqlparser_common.rs
```diff
@@ -8685,3 +8685,296 @@ fn parse_map_access_expr() {
...
+        Repetition(
+            Box::new(Group(Box::new(Concat(vec![
+                Exclude(Named(Ident::new("S3"))),
+                Symbol(Named(Ident::new("S4"))),
+            ])))),
+            OneOrMore,
+        ),
+    );
+
+    // the grand finale (example taken from snowflake docs)
```
alamb, line 8946: 🤯

alamb (approved):
Thanks @jmhain  and @iffyio

alamb:
@jmhain  -- this PR looks like it has some conflicts -- would it be possible to resolve them?

jmhain:
@alamb Yep! Done

alamb:
Thanks @jmhain

src/parser/mod.rs
iffyio: The FeatureSupport proposal seems like it balances the tradeoffs closely but yeah, agree that its likely not ideal to introduce another mechanism for this. At least the Dialect trait methods route sounds very reasonable to me as a preferred method in that case since as mentioned it helps users with custom dialect setup and it makes it easier to add test coverage for dialects.

Issue #1204, opened 2024-04-07
Always wrap statement in a named struct


tisonkun:
In [GreptimeDB](https://github.com/GreptimeTeam/greptimedb/blob/v0.7.1/src/sql/src/statements/delete.rs), we write code like:

```rust
#[derive(Debug, Clone, PartialEq, Eq, Visit, VisitMut)]
pub struct Delete {
    // Can only be sqlparser::ast::Statement::Delete variant
    pub inner: Statement,
}
```

It's because in sqlparser-rs, these statement variants are unnamed struct:

```rust
    Delete {
        /// Multi tables delete are supported in mysql
        tables: Vec<ObjectName>,
        /// FROM
        from: Vec<TableWithJoins>,
        /// USING (Snowflake, Postgres, MySQL)
        using: Option<Vec<TableWithJoins>>,
        /// WHERE
        selection: Option<Expr>,
        /// RETURNING
        returning: Option<Vec<SelectItem>>,
    },
```

I wonder if we can always wrap statement in a named struct so that downstream software can reuse the AST and impls (like for Display) more smoothly.

I don't know if it's a breaking change or we're generally OK with this.

alamb:
Here is a similar discussion in the past: https://github.com/sqlparser-rs/sqlparser-rs/issues/311

I agree it would be nice to do

It would be a large code change both for this crate and downstream

tisonkun:
@alamb Thanks for your information! I'm glad to see it's discussed and previous efforts exist, so I don't have to spend the same time exploring those options.

I can see we have one consensus and two options now:

* Consensus: It would be nice to wrap statements in a named struct, in general.
* Options: Since the main concern is code change size and breaking changes for this crate and downstream, then:
  1. Do it at once, as previous efforts tried. Depend on the maintenance status, few ownership can drive such a decision (OT: this is a shared challenge [for the Rust teams also](https://smallcultfollowing.com/babysteps/blog/2024/04/05/ownership-in-rust/)).
[...]

PR #1224, opened 2024-04-15
Encapsulate `Insert` and `Delete` into specific structs

tisonkun:
cc @alamb 

This refers to #1204  where we will pull out some of the higher level enum variants into their own `Struct`s for easier use

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -37,6 +37,7 @@ pub use self::ddl::{
     ReferentialAction, TableConstraint, UserDefinedTypeCompositeAttributeDef,
     UserDefinedTypeRepresentation, ViewColumnDef,
 };
+pub use self::dml::{Delete, Insert};
 pub use self::operator::{BinaryOperator, UnaryOperator};
 pub use self::query::{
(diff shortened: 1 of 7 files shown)


src/ast/mod.rs
```diff
@@ -2912,24 +2866,25 @@ impl fmt::Display for Statement {
...
-                source,
-                table,
-                on,
-                returning,
-                replace_into,
-                priority,
-                insert_alias,
-            } => {
+            Statement::Insert(insert) => {
+                let Insert {
```
alamb, line 2870: I don't think it matters, but just FYI you can do the same thing in the single let statement

```suggestion
            Statement::Insert(Insert {
                ..
             }) => {
```

alamb (approved):
Thank you @tisonkun  -- while this will require downstream changes, I think they are mechanical and will be straightforward for users and this will make it easier to manage these statements going forward.

src/ast/mod.rs
tisonkun: Thanks for your comment! I'll remember this and update this snippet when I have some time. You're correct. I just follow the intuitive "refactor" :D

tisonkun:
Thanks for your review and merge!

PR #1226, opened 2024-04-17
fix: have wildcard replace work in duckdb and snowflake syntax

HiranmayaGundu:
Snowflake and DuckDB also support `SELECT * REPLACE() from table`, so enabling the functionality for both dialects. 
I added test cases so that the examples given in the respective documentation passed:
duck db: https://duckdb.org/docs/sql/query_syntax/select.html
snowflake: https://docs.snowflake.com/en/sql-reference/sql/select

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9059,7 +9059,7 @@ impl<'a> Parser<'a> {
             None
         };
 
-        let opt_replace = if dialect_of!(self is GenericDialect | BigQueryDialect | ClickHouseDialect)
+        let opt_replace = if dialect_of!(self is GenericDialect | BigQueryDialect | ClickHouseDialect |  DuckDbDialect | SnowflakeDialect)
         {
             self.parse_optional_select_item_replace()?
         } else {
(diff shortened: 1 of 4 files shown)


tests/sqlparser_duckdb.rs
```diff
@@ -508,3 +508,31 @@ fn test_duckdb_named_argument_function_with_assignment_operator() {
         expr_from_projection(only(&select.projection))
     );
 }
+
+#[test]
+fn test_select_wildcard_with_replace() {
```
iffyio, line 513: Would it make sense to move the test to `sqlparser_common` so that all dialects can share? (moving the ones in bigquery and clickhouse) - since they all test the same functionality, it would let us not duplicate the same test for each dialect as we do currently in this PR

HiranmayaGundu: Thanks for the review @iffyio! I moved the tests to `sqlparser_common`

iffyio (approved):
LGTM!
@alamb

PR #1232, opened 2024-04-24
fix: Add support for `MAX` for `NVARCHAR`

bombsimon:
`NVARCHAR` works similar to `VARCHAR` except it uses unicode for the character set. This means that it supports the `MAX` keyword for some dialects just like `VARCHAR`.

This will change the wrapped type for `Nvarchar` to a `CharacterLength` instead of `u64`.

---
The support for this was already mentioned in a comment in the code but wasn't supported so I was hoping this would be a reasonable change.

https://github.com/sqlparser-rs/sqlparser-rs/blob/ce85084debefd69d2d43a1bdde73e25f2cb0c9d6/src/ast/data_type.rs#L537-L538

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -40,7 +40,7 @@ pub enum DataType {
     /// Variable-length character type e.g. VARCHAR(10)
     Varchar(Option<CharacterLength>),
     /// Variable-length character type e.g. NVARCHAR(10)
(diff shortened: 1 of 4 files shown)


iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @bombsimon  and @iffyio

PR #1237, opened 2024-04-27
BigQuery: Support window clause using named window

iffyio:
BigQuery allows window clause declaration using
either a window spec or a reference to another
named window.
This adds support for the latter.

https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls#ref_named_window


jmhain:
lgtm! cc @alamb

alamb (approved):
Looks good to me -- thanks @jmhain and @iffyio  

I really appreciate that you are helping with reviews (as well as the direct contributions) ❤️

PR #1239, opened 2024-04-27
Support BigQuery window function null treatment

iffyio:
Syntax differs for BigQuery on positioning of the
`IGNORE|RESPECT NULL` clause within a window function. This extends the parser to cover that syntax.

https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#first_value

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -13,11 +13,16 @@
 //! SQL Abstract Syntax Tree (AST) types
 #[cfg(not(feature = "std"))]
 use alloc::{
+    borrow::Cow,
     boxed::Box,
     format,
     string::{String, ToString},
     vec::Vec,
 };
+
+#[cfg(feature = "std")]
+use std::borrow::Cow;
+
 use core::fmt::{self, Display};
 
 #[cfg(feature = "serde")]
@@ -1403,6 +1408,35 @@ impl fmt::Display for NullTreatment {
             NullTreatment::IgnoreNulls => "IGNORE NULLS",
             NullTreatment::RespectNulls => "RESPECT NULLS",
         })
+    }
+}
+
+/// Specifies Ignore / Respect NULL within window functions.
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum NullTreatmentType {
+    /// The declaration is part of the function's arguments.
+    ///
+    /// ```sql
+    /// FIRST_VALUE(x IGNORE NULLS) OVER ()
+    /// ```
+    FunctionArg(NullTreatment),
+    /// The declaration occurs after the function call.
+    ///
+    /// ```sql
+    /// FIRST_VALUE(x) IGNORE NULLS OVER ()
+    /// ```
+    AfterFunction(NullTreatment),
+}
+
+impl Display for NullTreatmentType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let null_treatment = match self {
+            NullTreatmentType::FunctionArg(n) => n,
+            NullTreatmentType::AfterFunction(n) => n,
+        };
+        write!(f, "{null_treatment}")
     }
 }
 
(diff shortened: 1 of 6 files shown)


alamb (commented):
My reading of the referenced [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#first_value) don't seem to permit syntax like `FIRST_VALUE(x) IGNORE NULLS OVER`

Are we sure this is supported by BigQuery?

```
FIRST_VALUE (value_expression [{RESPECT | IGNORE} NULLS])
OVER over_clause

over_clause:
  { named_window | ( [ window_specification ] ) }

window_specification:
  [ named_window ]
  [ PARTITION BY partition_expression [, ...] ]
  ORDER BY expression [ { ASC | DESC }  ] [, ...]
  [ window_frame_clause ]
```

iffyio:
> Are we sure this is supported by BigQuery?

@alamb yeah so bigquery doesn't support `FIRST_VALUE(x) IGNORE NULLS OVER`, that one we have support for in the parser today - e.g [snowflake](https://docs.snowflake.com/en/sql-reference/functions/first_value#syntax) uses that variant. This one adds support for the variant linked in the doc, so  `FIRST_VALUE(x IGNORE NULLS) OVER` with similar example. Does that clarify the difference in behavior?

src/ast/mod.rs
```diff
@@ -1404,6 +1404,36 @@ impl fmt::Display for NullTreatment {
...
+    /// The declaration is part of the function's arguments.
+    ///
+    /// ```sql
+    /// FIRST_VALUE(x IGNORE NULLS) OVER ()
+    /// ```
+    FunctionArg(NullTreatment),
+    /// The declaration occurs after the function call.
+    ///
+    /// ```sql
+    /// FIRST_VALUE(x IGNORE NULLS) OVER ()
```
alamb, line 1421: This looks like a copy paste error?

```suggestion
```

src/ast/mod.rs
```diff
@@ -4785,15 +4815,18 @@ pub struct Function {
...
-    // Snowflake/MSSQL supports different options for null treatment in rank functions
-    pub null_treatment: Option<NullTreatment>,
+    /// Specifies Ignore / Respect NULL within window functions.
+    ///
+    /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#first_value)
+    /// [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/first_value)
+    pub null_treatment: Option<NullTreatmentType>,
     pub over: Option<WindowType>,
-    // aggregate functions may specify eg `COUNT(DISTINCT x)`
+    /// aggregate functions may specify eg `COUNT(DISTINCT x)`
```
alamb, line 4824: 💯  for the driveby cleanup

src/ast/mod.rs
```diff
@@ -4828,19 +4861,25 @@ impl fmt::Display for Function {
...
                 self.name,
                 if self.distinct { "DISTINCT " } else { "" },
                 display_comma_separated(&self.args),
                 display_comma_separated(&self.order_by),
+                match self.null_treatment {
+                    Some(NullTreatmentType::FunctionArg(null_treatment)) => {
+                        format!(" {null_treatment}")
+                    }
+                    _ => "".to_string(),
+                }
```
alamb, line 4874: I think you can avoid this seeming allocation in `"".to_string()` by using [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html):

```suggestion
                match self.null_treatment {
                    Some(NullTreatmentType::FunctionArg(null_treatment)) => {
                        Cow::Owned(format!(" {null_treatment}"))
                    }
                    _ => Cow::Borrowed("")
                }
```

alamb (approved):
> > Are we sure this is supported by BigQuery?
> 
> @alamb yeah so bigquery doesn't support `FIRST_VALUE(x) IGNORE NULLS OVER`, that one we have support for in the parser today - e.g [snowflake](https://docs.snowflake.com/en/sql-reference/functions/first_value#syntax) uses that variant. This one adds support for the variant linked in the doc, so `FIRST_VALUE(x IGNORE NULLS) OVER` with similar example. Does that clarify the difference in behavior?

Ah, I see now. BigQuery / Snowflake, why!!!! LOL

Overall this looks good to me -- I have two small suggestions but we could do them as a follow on PR too if you prefer.

Thank you @iffyio

src/ast/mod.rs
iffyio: Done!

src/ast/mod.rs
iffyio: Fixed!

alamb:
Thanks @iffyio

jmhain:
Sorry for posting this comment post-merge, but I was looking at fixing up our support for aggs that use the `WITHIN GROUP` syntax, and I realized that `Function` is going to get a bit unwieldy with additional options over time. I'd like to actually split off a separate `FunctionArgList` type to encapsulate a lot of these, and that would suggest another way of representing the inside-vs-outside parens issue this PR addressed:

```
// options inside the parens
struct FunctionArgList {
    distinct: boolean,
    args: Vec<FunctionArg>,
    limit: ... // for array_agg
    on_overflow: ... // for list_agg
    null_treatment: Option<NullTreatment>,
}

struct Function {
    args: FunctionArgList,

    // options outside the parens
    null_treatment: Option<NullTreatment>,
    within_group: ...
    over: ...
    filter: ...
}
```

@alamb @iffyio wdyt? If we do this, we should do it before the next release to avoid API churn...

iffyio:
That sounds reasonable to me!

PR #1241, opened 2024-04-28
Support Struct datatype parsing for GenericDialect

### LINKED ISSUE #1240: Support Struct datatype parsing for GenericDialect
Current BigQueryDialect is capable of parsing struct data type such as
```
create table t (s struct<n int, s varchar>);
```
According to [README](https://github.com/sqlparser-rs/sqlparser-rs/tree/0b5722afbfb60c3e3a354bc7c7dc02cb7803b807?tab=readme-ov-file#new-syntax), this feature should also support GenericDialect, however in current code, before parsing this syntax, we [only allow BigQueryDialect to have this syntax](https://github.com/sqlparser-rs/sqlparser-rs/blob/2f437db2a68724e4ae709df22f53999d24804ac7/src/parser/mod.rs#L5280)

duongcongtoai:
Fix #1240

### DIFF
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -395,19 +395,25 @@ fn parse_nested_data_types() {
 fn parse_invalid_brackets() {
     let sql = "SELECT STRUCT<INT64>>(NULL)";
     assert_eq!(
-        bigquery_and_generic().parse_sql_statements(sql).unwrap_err(),
+        bigquery_and_generic()
+            .parse_sql_statements(sql)
+            .unwrap_err(),
         ParserError::ParserError("unmatched > in STRUCT literal".to_string())
     );
 
     let sql = "SELECT STRUCT<STRUCT<INT64>>>(NULL)";
     assert_eq!(
-        bigquery_and_generic().parse_sql_statements(sql).unwrap_err(),
+        bigquery_and_generic()
+            .parse_sql_statements(sql)
+            .unwrap_err(),
         ParserError::ParserError("Expected (, found: >".to_string())
     );
 
     let sql = "CREATE TABLE table (x STRUCT<STRUCT<INT64>>>)";
     assert_eq!(
-        bigquery_and_generic().parse_sql_statements(sql).unwrap_err(),
+        bigquery_and_generic()
+            .parse_sql_statements(sql)
+            .unwrap_err(),
         ParserError::ParserError(
             "Expected ',' or ')' after column definition, found: >".to_string()
         )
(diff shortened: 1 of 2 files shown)


tests/sqlparser_bigquery.rs
```diff
@@ -350,7 +350,7 @@ fn parse_create_table_with_options() {
 #[test]
 fn parse_nested_data_types() {
     let sql = "CREATE TABLE table (x STRUCT<a ARRAY<INT64>, b BYTES(42)>, y ARRAY<STRUCT<INT64>>)";
-    match bigquery().one_statement_parses_to(sql, sql) {
+    match bigquery_and_generic().one_statement_parses_to(sql, sql) {
```
duongcongtoai, line 353: also test for generic dialect

tests/sqlparser_bigquery.rs
```diff
@@ -632,12 +638,12 @@ fn parse_typed_struct_syntax() {
         expr_from_projection(&select.projection[1])
     );
 
-    let sql = r#"SELECT STRUCT<DATE>("2011-05-05"), STRUCT<DATETIME>(DATETIME '1999-01-01 01:23:34.45'), STRUCT<FLOAT64>(5.0), STRUCT<INT64>(1)"#;
-    let select = bigquery().verified_only_select(sql);
+    let sql = r#"SELECT STRUCT<DATE>('2011-05-05'), STRUCT<DATETIME>(DATETIME '1999-01-01 01:23:34.45'), STRUCT<FLOAT64>(5.0), STRUCT<INT64>(1)"#;
```
duongcongtoai, line 641: change " to ', so it is understandable by generic dialect also

alamb: Rather than updating the existing test, can you please keep the existing test (that uses `"`) and add new coverage for bigquery_and_generic dialect (with the `'`)?

This is to avoid losing existing test coverage

alamb (commented):
Thank you for this contirbution @duongcongtoai  -- the code in this PR looks great to me. I think we need to avoid changing the existing tests, but then this PR wll be good to go

tests/sqlparser_bigquery.rs
duongcongtoai: i added extra test, was thinking of using a scenario struct to reuse the query logic and just replacing " with ' in the middle before we invoke, but i guess a bit boilerplating is easier to read

alamb: Yeah, thank you -- I know it is a lot of repetition but it is easier to understand I think. Thank you 🙏

alamb (approved):
Thank you @duongcongtoai -- looks great to me

PR #1242, opened 2024-04-29
Support ?-based jsonb operators in Postgres

ReppCodes:
We already support most postgres jsonb operators, but were missing the ?-based ones. This commit adds support for them in both the tokenizer and the AST enums.  This is simplified in the tokenizer with a dialect-specific carve-out, since Postgres thankfully does not also use ? for anonymous prepared statement parameters.

This resolves https://github.com/sqlparser-rs/sqlparser-rs/issues/1236

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -2401,6 +2401,51 @@ fn test_json() {
         },
         select.selection.unwrap(),
     );
+
+    let sql = r#"SELECT info FROM orders WHERE info ? 'b'"#;
+    let select = pg().verified_only_select(sql);
+    assert_eq!(
+        Expr::BinaryOp {
+            left: Box::new(Expr::Identifier(Ident::new("info"))),
+            op: BinaryOperator::Question,
+            right: Box::new(Expr::Value(Value::SingleQuotedString("b".to_string()))),
+        },
+        select.selection.unwrap(),
+    );
+
+    let sql = r#"SELECT info FROM orders WHERE info ?& ARRAY['b', 'c']"#;
+    let select = pg().verified_only_select(sql);
+    assert_eq!(
+        Expr::BinaryOp {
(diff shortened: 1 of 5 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -2358,6 +2358,51 @@ fn test_json() {
         },
         select.selection.unwrap(),
     );
+
+    let sql = r#"SELECT info FROM orders WHERE info ? 'b'"#;
```
alamb, line 2362: Can we please add coverage for all new operators that were added? 

Perhaps just verifying that they roundtrip ( no need to reverify the parsed structure for all of them), perhaps using `one_statement_parses_to` or something similar

alamb (commented):
Thank you @ReppCodes  - this looks great. I think we need a few more tests but otherwise 👨‍🍳 👌 

Note @jmhain  made some significant changes to how json access is represented in https://github.com/sqlparser-rs/sqlparser-rs/pull/1215 

I think we should merge that PR first and then this PR

tests/sqlparser_postgres.rs
ReppCodes: @alamb The tests at lines 2373 and at 2390 should cover the other of the three new operators added, I believe? Sorry, not sure what I'm missing here.

ReppCodes: Also, I'm cleaning up the conflicts after getting that refactor merged in.  Should be up shortly.

ReppCodes: And done.  Assuming the three added tests hit the operators as needed, should be good to go.  Just let me know!

Thanks!

alamb: Yup --- looks good to me. Thank you (I think it was unclear to me that only three new operators got added initially). Sorry for my confusion

alamb (approved):
Thank you @ReppCodes  -- looks like a very nice contribution to me 🙏

PR #1247, opened 2024-05-01
Consolidate representation of function calls, remove `AggregateExpressionWithFilter`, `ArraySubquery`, `ListAgg` and `ArrayAgg`

jmhain:
This addresses a handful of issues and redundancies related to our representation of function calls.

* I split off a new `FunctionArguments` enum with three options:
    * A regular argument list with a list of additional clauses
    * No argument list - for representing things like `CURRENT_TIMESTAMP`. This is better than the `special` bit we had because the latter allows for many invalid AST states.
    * A subquery. This is used to losslessly represent Snowflake's support for unparenthesized subqueries as function arguments, and allowed removal of the special `ArraySubquery` expression variant.
* Reimplemented the fix from https://github.com/sqlparser-rs/sqlparser-rs/pull/1239 to take advantage of the new structure.
* I added `WITHIN GROUP` support to `Function`, which makes functions like  `percentile_cont(...) within group (...)` on Postgres work. 
* I removed the special `ListAgg` and `ArrayAgg` expression variants and subsumed them into `Function`
* Removed the redundant `AggregateExpressionWithFilter` expression variant. This was only needed before because of `ArrayAgg` / `ListAgg` being separate expression nodes.
* On BigQuery, `ARRAY_CONCAT_AGG` and `STRING_AGG` support the same set of options as `ARRAY_AGG`, and we can parse those now.
* The `ALL` in `SELECT COUNT(ALL column)` is now preserved in roundtrips.

This lays the groundwork for adding additional special function argument clauses like bigquery's `ANY_VALUE(foo HAVING MIN value)`.


iffyio (approved):
LGTM! Very nice with the new representation, thanks @jmhain!

cc @alamb

universalmind303 (approved):

tests/sqlparser_bigquery.rs
```diff
@@ -1824,18 +1824,7 @@ fn parse_map_access_expr() {
...
-                        number("2"),
-                    )))],
-                    filter: None,
-                    null_treatment: None,
-                    over: None,
-                    distinct: false,
-                    special: false,
-                    order_by: vec![],
-                }),
+                call("safe_offset", [Expr::Value(number("2"))]),
```
alamb, line 1827: ❤️

src/ast/mod.rs
```diff
@@ -4829,22 +4779,169 @@ impl fmt::Display for CloseCursor {
...
+        if !self.clauses.is_empty() {
+            write!(f, " {}", display_separated(&self.clauses, " "))?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
```
alamb, line 4890: These comments are really lovely. Thank you @jmhain

alamb (approved):
👏  -- thank you @jmhain  -- this is really nice. While it will be a non trivial change for some downstream crates I think it will be for the better. 

Also, thank you for the reviews @iffyio  and @universalmind303 

🚀

src/ast/mod.rs
```diff
@@ -4829,22 +4779,169 @@ impl fmt::Display for CloseCursor {
...
+    /// ```
     ///
-    /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#first_value)
     /// [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/first_value)
-    pub null_treatment: Option<NullTreatmentType>,
+    pub null_treatment: Option<NullTreatment>,
+    /// The `OVER` clause, indicating a window function call.
     pub over: Option<WindowType>,
-    /// aggregate functions may specify eg `COUNT(DISTINCT x)`
-    pub distinct: bool,
```
tisonkun, line 4842: How do I get this property now then?

tisonkun: I can see `DuplicateTreatment` now. But it can cause downstream to restructure the handling code to adjust these structural changes.

Trying it out ...

alamb: Let us know if there is something wrong that can't be handled -- I can make another release with a fix if necessary

tisonkun: I'm trying to upgrade sqlparser to 0.46.0 this morning but found that except my breaking changes on Insert and Delete struct, it includes others like this one and the JsonAccess refactor. So I stop the work until I can have time to understand the changes :P

jmhain: Let me know if there's any way I can help...

tisonkun: @jmhain Thanks for offering the help. I just created https://github.com/apache/datafusion/pull/10392 that you can pull locally and run cargo build to see the compilation failure and see if you have the idea to fix it up.

Welcome to leave a comment on that PR or create a PR target to that PR. I'm actively responding.

jmhain: This is what I've got so far that I'm reasonably sure is correct without more context on DataFusion: https://github.com/apache/datafusion/commit/261f346140d0d6a265ef0f9c81ea1fcd007d5928

Out of time for now but I'll try to take a second look later tonight at the remaining JSON access changes needed

PR #1249, opened 2024-05-03
fix: allow snowflake to accept values without parens

HiranmayaGundu:
Snowflake and Databricks both allow: 
```
"SELECT * FROM VALUES (1, 2), (2,3) AS tbl (id, val)";
```
this adds support for that


alamb (approved):
LGTM -- thank you @HiranmayaGundu

PR #1252, opened 2024-05-04
Support multiple `SET` variables

iffyio:
Both BigQuery and Snowflake support assigning multiple variables via `SET`

```
SET (a, b) = (1, 2)
```
This adds support for such syntax

https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#set 

https://docs.snowflake.com/en/sql-reference/sql/set#syntax

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2275,7 +2275,8 @@ pub enum Statement {
         role_name: Option<Ident>,
     },
     /// ```sql
-    /// SET <variable>
+    /// SET <variable> = expression;
+    /// SET (variable[, ...]) = (expression[, ...]);
     /// ```
     ///
     /// Note: this is not a standard SQL statement, but it is supported by at
@@ -2284,7 +2285,7 @@ pub enum Statement {
     SetVariable {
         local: bool,
         hivevar: bool,
-        variable: ObjectName,
+        variables: OneOrManyWithParens<ObjectName>,
         value: Vec<Expr>,
     },
     /// ```sql
@@ -3823,7 +3824,7 @@ impl fmt::Display for Statement {
             }
             Statement::SetVariable {
                 local,
-                variable,
+                variables,
                 hivevar,
                 value,
             } => {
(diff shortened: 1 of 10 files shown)


src/ast/mod.rs
```diff
@@ -1749,6 +1749,40 @@ pub enum FromTable {
     WithoutKeyword(Vec<TableWithJoins>),
 }
 
+/// A variable in a `SET` statement.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum SetVariables {
```
jmhain, line 1756: This syntactic pattern of `{ thing | (thing1, thing2, ...) }` is somewhat common in SQL so I was thinking we could actually make this generic like I did in my PR over here: https://github.com/sqlparser-rs/sqlparser-rs/pull/1257/files#diff-06ccb0b0ac8767ba02e52e9776bd9acfc212c87f8e943b06d093233c46b9e15aR830

An advantage of having a single type for this is that we could later add an `IntoIter` impl so consumers of these types don't need to deal with the multiple variants here.

jmhain (approved):
lgtm! cc @alamb

src/ast/mod.rs
iffyio: Nice yeah I think that makes a lot of sense!

alamb: is the idea to do this in a follow on PR(s)?

alamb (approved):

alamb:
Screenshot 2024-05-06 at 3 39 57 PM

Apparently this branch has conflicts that must be resolved before I can merge

src/ast/mod.rs
jmhain: Yeah, I can open a PR for it after this one merges

iffyio: Ah right the other PR's merged I can replace the type while rebasing :+1:

iffyio: Rebased and replaced the type!

PR #1254, opened 2024-05-04
Allow keyword as field names in BigQuery struct syntax

iffyio:
Updates the logic for inferring named vs anonymouse field syntax in a BigQuery struct. The previous logic incorrectly assumed that seeing a keyword meant no field name was specified.
Updates the logic to instead look further ahead at both tokens in order to correctly infer if both name and type are present.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2155,13 +2155,12 @@ impl<'a> Parser<'a> {
     fn parse_big_query_struct_field_def(
         &mut self,
     ) -> Result<(StructField, MatchedTrailingBracket), ParserError> {
-        let is_anonymous_field = if let Token::Word(w) = self.peek_token().token {
-            ALL_KEYWORDS
-                .binary_search(&w.value.to_uppercase().as_str())
-                .is_ok()
-        } else {
-            false
-        };
+        // Look beyond the next item to infer whether both field name
+        // and type are specified.
+        let is_anonymous_field = !matches!(
+            (self.peek_nth_token(0).token, self.peek_nth_token(1).token),
+            (Token::Word(_), Token::Word(_))
+        );
 
         let field_name = if is_anonymous_field {
             None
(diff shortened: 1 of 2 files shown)


tests/sqlparser_bigquery.rs
```diff
@@ -792,6 +792,31 @@ fn parse_typed_struct_syntax_bigquery() {
         },
         expr_from_projection(&select.projection[1])
     );
+
+    // Keywords may be used as field names.
+    let sql = r#"SELECT STRUCT<key INT64, value INT64, join INT64>(42)"#;
```
jmhain, line 797: It doesn't look like BQ will actually accept any keyword as a field name here. Running this query I get

```
Syntax error: Unexpected keyword JOIN at [2:39]
```

I think it's fine for us to succeed in parsing this anyway but it'd be nice for the test to use a query that works in BQ.

On that note, renaming `join` also complains about the missing fields in the constructor:

```
STRUCT type has 3 fields but constructor call has 1 fields at [2:8]
```

src/parser/mod.rs
```diff
@@ -2155,13 +2155,12 @@ impl<'a> Parser<'a> {
...
-            ALL_KEYWORDS
-                .binary_search(&w.value.to_uppercase().as_str())
-                .is_ok()
-        } else {
-            false
-        };
+        // Look beyond the next item to infer whether both field name
+        // and type are specified.
+        let is_anonymous_field = !matches!(
+            (self.peek_nth_token(0).token, self.peek_nth_token(1).token),
```
jmhain, line 2161: This gave me an idea: https://github.com/sqlparser-rs/sqlparser-rs/pull/1255

tests/sqlparser_bigquery.rs
iffyio: I updated the test to be less confusing/misleading now. Yeah join specifically doesn't seem to be allowed by bigquery, the keyword mainly refers to the parser's list of keywords vs bigquery's keywords (in the example fields `key` and `value` incorrectly fail to parse otherwise).

jmhain (approved):
lgtm! cc @alamb

alamb (approved):
🙏

PR #1255, opened 2024-05-04
Idea: add const generic peek_tokens method to parser

jmhain:
This provides a nicer API for peeking multiple tokens ahead.

Also, while this is probably not super impactful in the grand scheme of things since you're generally not peeking more than a few tokens at a time, this technically decreases the complexity of doing so from O(n^2) to O(n).

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2895,6 +2895,47 @@ impl<'a> Parser<'a> {
         self.peek_nth_token(0)
     }
 
+    /// Returns the `N` next non-whitespace tokens that have not yet been
+    /// processed.
+    ///
+    /// Example:
+    /// ```rust
+    /// # use sqlparser::dialect::GenericDialect;
+    /// # use sqlparser::parser::Parser;
+    /// # use sqlparser::keywords::Keyword;
+    /// # use sqlparser::tokenizer::{Token, Word};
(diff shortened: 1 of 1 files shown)


iffyio:
LGTM! cc @alamb

iffyio (approved):

src/parser/mod.rs
```diff
@@ -2895,6 +2893,52 @@ impl<'a> Parser<'a> {
...
+    ///         Token::Word(Word { keyword: Keyword::BY, .. }),
+    ///     ]
+    /// ));
+    /// ```
+    pub fn peek_tokens<const N: usize>(&self) -> [Token; N] {
+        self.peek_tokens_with_location()
+            .map(|with_loc| with_loc.token)
+    }
+
+    pub fn peek_tokens_with_location<const N: usize>(&self) -> [TokenWithLocation; N] {
```
alamb, line 2923: Perhaps can we add a docstring to this method too (even if it says "see [`Self::peek_tokens`] for details")?

alamb (approved):
This is pretty neat -- thank you @jmhain 

I think one more docstring and it will be ready to go

Thanks a lot.

cc @iffyio

src/parser/mod.rs
jmhain: Done

PR #1257, opened 2024-05-04
Databricks: support for lambda functions

jmhain:


src/dialect/mod.rs
```diff
@@ -209,6 +209,14 @@ pub trait Dialect: Debug + Any {
     fn supports_dictionary_syntax(&self) -> bool {
         false
     }
+    /// Returns true if the dialect supports lambda functions, for example:
+    ///
+    /// ```sql
+    /// SELECT transform(array(1, 2, 3), x -> x + 1); -- returns [2,3,4]
+    /// ```
+    fn supports_lambda_functions(&self) -> bool {
```
jmhain, line 217: One thing I was unsure of here is whether to enable this for the generic dialect. Our policy is to enable everything for generic, but this syntax conflicts with the `->` operator from dialects like Postgres. So supporting this would be a breaking change of somewhat dubious benefit, since we both gain and lose a feature in generic. Thoughts?

tests/sqlparser_databricks.rs
```diff
@@ -28,3 +29,98 @@ fn test_databricks_identifiers() {
...
+                    params: OneOrManyWithParens::One(Ident::new("x")),
+                    body: Box::new(Expr::IsNull(Box::new(Expr::Identifier(Ident::new("x")))))
+                })
+            ]
+        ),
+    );
+
+    let res = databricks().parse_sql_statements("SELECT EXISTS (");
+    assert_eq!(
+        // TODO: improve this error message...
```
iffyio, line 59: was the todo meant for this PR or as a general improvement/followup with error reporting?

iffyio (approved):
LGTM! cc @alamb

tests/sqlparser_databricks.rs
jmhain: the latter

tests/sqlparser_databricks.rs
```diff
@@ -28,3 +29,98 @@ fn test_databricks_identifiers() {
...
+                        ],
+                        else_result: Some(Box::new(Expr::Value(number("1"))))
+                    })
+                })
+            ]
+        )),
+        databricks().verified_only_select(sql).projection[0]
+    );
+
+    databricks().verified_expr(
```
alamb, line 122: pretty neat!

src/dialect/mod.rs
alamb: I think this makes sense to not support via GenericDialect when there is a conflict

🤔  that might be a good thing to add to the policy  https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#new-syntax

alamb (approved):
lGTM -- thank you @jmhain

alamb:
Thanks again for the review @iffyio

PR #1260, opened 2024-05-05
feat/adding scalar-variable expression

devanbenz:
- Adds `ScalarVariable` expression type
- `ScalarVariable` has two fields one for the corresponding SQL datatype and one of the name of the variable
- Adds test for fmt::Display 
- Requirement for https://github.com/apache/datafusion/issues/9726

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -798,6 +798,16 @@ pub enum Expr {
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/where#joins-in-the-where-clause>.
     OuterJoin(Box<Expr>),
+<<<<<<< Updated upstream
+=======
+    /// A reference to the prior level in a CONNECT BY clause.
+    Prior(Box<Expr>),
+    /// Scalar variable creation e.g. `[@]foo INT`
+    ScalarVariable {
+        data_type: DataType,
+        name: String,
+    },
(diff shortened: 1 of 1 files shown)


src/ast/mod.rs
```diff
@@ -793,6 +793,11 @@ pub enum Expr {
     OuterJoin(Box<Expr>),
     /// A reference to the prior level in a CONNECT BY clause.
     Prior(Box<Expr>),
+    /// Scalar variable creation e.g. `[@]foo INT`
```
iffyio, line 796: Would it be possible to add a link to some documentation that uses this expression? I think that would be helpful in this case - e.g I'm unfamiliar with the syntax and unable to tell how it's used as a result

src/ast/mod.rs
```diff
@@ -793,6 +793,11 @@ pub enum Expr {
...
+    ScalarVariable {
```
iffyio, line 797: hmm, it looks like we're missing implementation in the PR for how the parser actually parses/constructs the ScalarVariable expression or is that intended?

devanbenz:
@iffyio it is likely that this PR will not be needed, there is another route I am taking with the Scalar Variable in datafusion 👍 will leave it up just in case though.

alamb:
marking as draft as we sort out what to do here

PR #1262, opened 2024-05-06
Support triple quoted strings

iffyio:
Add tokenizer support for triple quoted strings for BigQuery.

https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_literals

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -42,6 +42,12 @@ pub enum Value {
     SingleQuotedString(String),
     // $<tag_name>$string value$<tag_name>$ (postgres syntax)
     DollarQuotedString(DollarQuotedString),
+    /// Triple single quoted strings: Example '''abc'''
+    /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#quoted_literals)
+    TripleSingleQuotedString(String),
+    /// Triple double quoted strings: Example """abc"""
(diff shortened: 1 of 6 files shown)


src/ast/value.rs
```diff
@@ -73,14 +94,25 @@ impl fmt::Display for Value {
             Value::Number(v, l) => write!(f, "{}{long}", v, long = if *l { "L" } else { "" }),
             Value::DoubleQuotedString(v) => write!(f, "\"{}\"", escape_double_quote_string(v)),
             Value::SingleQuotedString(v) => write!(f, "'{}'", escape_single_quote_string(v)),
+            Value::TripleSingleQuotedString(v) => {
+                write!(f, "'''{}'''", escape_single_quote_string(v))
```
jmhain, line 98: Do we need an `escape_triple_single_quoted_string()`? Per their docs it looks like a single `'` is allowed in one of these unescaped.

iffyio: Yeah good catch! I think that some special handling is needed here when displaying the string, will take a closer look

iffyio: Updated, the unescaped quotes are now left as is

jmhain:
lgtm! cc @alamb

alamb (commented):
wow -- BigQuery should get some sort of recognition of inventing the most new SQL syntax I have recently seen https://twitter.com/andrewlamb1111/status/1789243972788846746

alamb (approved):

alamb:
Thanks @iffyio  and @jmhain

Issue #1266, opened 2024-05-09
The DoubleColon cast always skips the AT TIME ZONE part, leading to strange behavior.


dmitrybugakov:
We have [discovered](https://github.com/apache/datafusion/issues/10310) that during DoubleColon cast, the parser consistently skips the AT TIME ZONE part. It functions correctly only when nested.

Consider the following select statement:
`"SELECT c FROM t WHERE c >= '2019-03-27T22:00:00.000Z'::timestamp AT TIME ZONE 'Europe/Brussels'"`

This results in the following AST, where we attempt to cast the condition under the WHERE clause instead of only the right part:

`[Statement(Query(Query { with: None, body: Select(Select { distinct: None, top: None, projection: [UnnamedExpr(Identifier(Ident { value: "c", quote_style: None }))], into: None, from: [TableWithJoins { relation: Table { name: ObjectName([Ident { value: "t", quote_style: None }]), alias: None, args: None, with_hints: [], version: None, partitions: [] }, joins: [] }], lateral_views: [], selection: Some(AtTimeZone { timestamp: BinaryOp { left: Identifier(Ident { value: "c", quote_style: None }), op: GtEq, right: Cast { expr: Value(SingleQuotedString("2019-03-27T22:00:00.000Z")), data_type: Timestamp(None, None), format: None } }, time_zone: "Europe/Brussels" }), group_by: Expressions([]), cluster_by: [], distribute_by: [], sort_by: [], having: None, named_window: [], qualify: None, value_table_mode: None }), order_by: [], limit: None, limit_by: [], offset: None, fetch: None, locks: [], for_clause: None }))]`

PR #1267, opened 2024-05-09
Fix DoubleColon cast skipping AT TIME ZONE #1266

dmitrybugakov:
Closes: #1266

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1528,14 +1528,18 @@ impl<'a> Parser<'a> {
     pub fn parse_optional_cast_format(&mut self) -> Result<Option<CastFormat>, ParserError> {
         if self.parse_keyword(Keyword::FORMAT) {
             let value = self.parse_value()?;
-            if self.parse_keywords(&[Keyword::AT, Keyword::TIME, Keyword::ZONE]) {
-                Ok(Some(CastFormat::ValueAtTimeZone(
-                    value,
-                    self.parse_value()?,
-                )))
-            } else {
-                Ok(Some(CastFormat::Value(value)))
+            match self.parse_optional_time_zone()? {
+                Some(tz) => Ok(Some(CastFormat::ValueAtTimeZone(value, tz))),
+                None => Ok(Some(CastFormat::Value(value))),
             }
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub fn parse_optional_time_zone(&mut self) -> Result<Option<Value>, ParserError> {
+        if self.parse_keywords(&[Keyword::AT, Keyword::TIME, Keyword::ZONE]) {
+            self.parse_value().map(Some)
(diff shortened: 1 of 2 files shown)


src/ast/mod.rs
```diff
@@ -424,7 +424,7 @@ pub enum CastKind {
     /// See <https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#safe_casting>.
     SafeCast,
     /// `<expr> :: <datatype>`
-    DoubleColon,
+    DoubleColonCast,
```
jmhain, line 427: Please rename this back to `DoubleColon`. The others have the word `Cast` because you actually use the word `CAST` in the syntax. This is also a breaking change without enough benefit to justify.

lovasoa: agreed

lovasoa (changes requested):
Can we also remove the formatting changes that are not related to the PR ?

lovasoa:
Is  CastFormat::ValueAtTimeZone still needed if we parse the cast as a separate "at time zone" ast node ?

dmitrybugakov:
> Is CastFormat::ValueAtTimeZone still needed if we parse the cast as a separate "at time zone" ast node ?

I'm currently working on the infix part where there is no formatting involved. However, in the prefix part, we are using the BigQuery style where the format is a part of the expression itself. 

https://github.com/sqlparser-rs/sqlparser-rs/blob/e3692f4681e958e1ed2110494636032e5d53b710/src/ast/mod.rs#L313

jmhain:
This isn't something this PR necessarily needs to address, but I think we should actually get rid `Expr::AtTimeZone` and add `BinaryOperator::AtTimeZone`. Reasons;

* `AT TIME ZONE` _is_ a binary operator like any other we have. The RHS can actually be any expression that resolves to a timezone, e.g. `<expr> AT TIME ZONE ('America/' || 'Los_Angeles')` works, as do column references, none of which successfully parse right now.
* Once we support expressions on the RHS, a whole bunch of other precedence issues come into play. For example, does `<expr> at time zone <tz> || 'foo'` concat to the timezone or to the entire  `at time zone` expr? (iirc it's the latter and pg will implicitly cast the ts to a string). We should be handling these ambiguities using the existing Pratt parsing logic we have.

lovasoa:
agreed. but we should merge this pr first. It fixes a clear bug and does not require a refactoring

jmhain (approved):
lgtm, cc @alamb 

I filed https://github.com/sqlparser-rs/sqlparser-rs/issues/1270 for addressing my comment above.

lovasoa (approved):

alamb:
Thanks @dmitrybugakov @jmhain and @lovasoa  🙏

alamb (approved):

PR #1272, opened 2024-05-16
Support expression in AT TIME ZONE and fix precedence

### LINKED ISSUE #1270: Support column references and other expressions in RHS of AT TIME ZONE operator
We currently require the RHS of `AT TIME ZONE` operator to be a string literal, but in postgres, it's allowed to be an expression. For example, an expression such as `<timestamp> AT TIME ZONE ('America/' || some_column)` is accepted. This change would make `Expr::AtTimeZone` isomorphic to `Expr::BinaryOp`, so when implementing this I suggest we eliminate `Expr::AtTimeZone` and add `BinaryOperator::AtTimeZone`.

We'll need to be careful to get the precedence right. In `<timestamp> AT TIME ZONE 'America/Los Angeles'::text` the cast is on the time zone string literal, but in `<timestamp> AT TIME ZONE 'America/Los Angeles' || 'foo'` it's concatenating the result of the entirely `AT TIME ZONE` binop (which is implicitly casted into text).

jmhain:
Postgres supports an expression for the timezone input to the `AT TIME ZONE` operator. Through some experimentation, I discovered that the precedence is actually not that of the "(any other operator)" entry in the [documentation](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE), so I corrected this as well. The latter makes the workaround from https://github.com/sqlparser-rs/sqlparser-rs/pull/1267/files no longer necessary - the test added in that PR passes without it.

Resolves #1270.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -584,7 +584,7 @@ pub enum Expr {
     /// AT a timestamp to a different timezone e.g. `FROM_UNIXTIME(0) AT TIME ZONE 'UTC-06:00'`
     AtTimeZone {
         timestamp: Box<Expr>,
-        time_zone: String,
+        time_zone: Box<Expr>,
     },
     /// Extract a field from a timestamp e.g. `EXTRACT(MONTH FROM foo)`
     ///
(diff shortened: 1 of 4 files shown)


iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @jmhain  and @iffyio

PR #1278, opened 2024-05-19
Support: Databricks and generic: fix for `values` as table name

jmhain:
In the previous release, we added support for `VALUES` tables in a `FROM` clause without a containing subquery, as supported by Snowflake and Databricks. However, this caused a regression in DataFusion which considers `values` a valid table name. I did some testing and Databricks permits this as well (Snowflake does not).


alamb (approved):
Thank you @jmhain

PR #1289, opened 2024-05-27
ClickHouse: support of create table query with primary key and parametrised table engine

7phs:
This PR implements extended support of query `CREATE TABLE` in ClickHouse that contains the following additional parameters and differentiation:

1. [Primary key](https://clickhouse.com/docs/en/sql-reference/statements/create/table#primary-key). It is located before `ORDER BY` in  a `CREATE TABLE` query;
2. [Table Engine with parameters](https://clickhouse.com/docs/en/engines/table-engines)
3. `ORDER BY` can include expression and not only indication.


src/parser/mod.rs
```diff
@@ -5012,17 +5018,23 @@ impl<'a> Parser<'a> {
...
+            self.parse_optional_click_house_create_table_config()?
+        } else {
+            Default::default()
+        };
+
         let order_by = if self.parse_keywords(&[Keyword::ORDER, Keyword::BY]) {
             if self.consume_token(&Token::LParen) {
                 let columns = if self.peek_token() != Token::RParen {
-                    self.parse_comma_separated(|p| p.parse_identifier(false))?
+                    self.parse_comma_separated(|p| p.parse_expr())?
```
iffyio, line 5030: can we change the type of order_by to something like [`OneOrManyWithParens<Expr>`](https://github.com/sqlparser-rs/sqlparser-rs/blob/a5b0ec183b2ca2f54627a7f14f6a0daa0eb4e2d2/src/ast/mod.rs#L824-L835)? i.e. if using parenthesis here we use the Many variant, otherwise the One variant.
I think that removes the issue with the `ORDER BY tuple(i)` vs `ORDER BY (tuple(i))` mismatch as a result?

src/parser/mod.rs
```diff
@@ -5012,17 +5018,23 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let click_house_config = if dialect_of!(self is ClickHouseDialect) {
```
iffyio, line 5021:
```suggestion
        let click_house_config = if dialect_of!(self is ClickHouseDialect | GenericDialect) {
```
if the behavior isn't conflicting with existing behavior on the generic dialect we can have the dialect accept it as well, we tend to do that when it allows https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#new-syntax

tests/sqlparser_clickhouse.rs
```diff
@@ -220,6 +220,88 @@ fn parse_create_table() {
     );
 }
 
+#[test]
+fn parse_create_table_with_primary_key() {
+    match clickhouse().one_statement_parses_to(
```
iffyio, line 225: per the generic dialect comment earlier, if the dialect allows, then we can change this to `clickhouse_and_generic()`

tests/sqlparser_clickhouse.rs
```diff
@@ -220,6 +220,88 @@ fn parse_create_table() {
...
+        concat!(
+            r#"CREATE TABLE db.table (`i` Int, `k` Int)"#,
+            " ENGINE=SharedMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')",
+            " PRIMARY KEY tuple(i)",
+            " ORDER BY tuple(i)",
```
iffyio, line 230: We can probably add a test case that the parser rejects e.g. `ORDER BY tuple(i), tuple(k)`

src/parser/mod.rs
```diff
@@ -10540,6 +10567,16 @@ impl<'a> Parser<'a> {
...
 
+    fn parse_optional_idents(&mut self) -> Result<Option<Vec<Ident>>, ParserError> {
+        if self.consume_token(&Token::LParen) {
+            let partitions = self.parse_comma_separated(|p| p.parse_identifier(false))?;
+            self.expect_token(&Token::RParen)?;
+            Ok(Some(partitions))
+        } else {
+            Ok(None)
+        }
+    }
```
iffyio, line 10578: instead of duplicating this function we have the caller do
```rust
let engine_parameters = if self.peek_token() == LParen {
   Some(self.parse_partitions())
} else {
    None
}
```
then makes sense to rename `parse_partitions` to something more generic now like `parse_parenthesized_identifiers` or similar
```

File: src/parser/mod.rs [hunk 6]
```diff
@@ -5153,6 +5167,19 @@ impl<'a> Parser<'a> {
...
+    fn parse_optional_click_house_create_table_config(
+        &mut self,
+    ) -> Result<ClickHouseTableConfiguration, ParserError> {
+        let mut primary_key = None;
+        if self.parse_keywords(&[Keyword::PRIMARY, Keyword::KEY]) {
+            primary_key = Some(Box::new(self.parse_expr()?));
+        };
+
+        Ok(ClickHouseTableConfiguration { primary_key })
+    }
```
iffyio, line 5181: would it make sense to inline this similar to how engine is handled? (thinking since currently the function is only looking for primary key which would probably not be unique to clickhouse)

src/ast/mod.rs
```diff
@@ -1992,6 +1992,7 @@ pub enum Statement {
         like: Option<ObjectName>,
         clone: Option<ObjectName>,
         engine: Option<String>,
+        engine_parameters: Option<Vec<Ident>>,
```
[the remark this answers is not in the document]
7phs: It is align with my initial idea.

I tried to make PR more compatible without changing a type of existing field.

Updated code with `struct TableEngine`.

src/parser/mod.rs

tests/sqlparser_clickhouse.rs
7phs: Added verification of this error into this unit test.

src/parser/mod.rs
7phs: Renamed `parse_partitions` -> `parse_parenthesized_identifiers ` and updated code.

src/parser/mod.rs
7phs: Rid of this function and executes parsing of `PRIMARY KEY` in the function `parse_create_table` as other keywords.

tests/sqlparser_clickhouse.rs
7phs: Updated.

src/parser/mod.rs
7phs: Updated `order_by` as an `Option<OneOrManyWithParens<Expr>>`

tests/sqlparser_clickhouse.rs
```diff
@@ -211,15 +211,102 @@ fn parse_delimited_identifiers() {
...
-        r#"CREATE TABLE "x" ("a" "int") ENGINE=MergeTree ORDER BY ("x") AS SELECT * FROM "t" WHERE true"#,
+        r#"CREATE TABLE "x" ("a" "int") ENGINE=MergeTree ORDER BY "x" AS SELECT * FROM "t" WHERE true"#,
     );
 }
 
+#[test]
+fn parse_create_table_with_primary_key() {
+    match clickhouse_and_generic().one_statement_parses_to(
+        concat!(
+            r#"CREATE TABLE db.table (`i` Int, `k` Int)"#,
```
iffyio, line 224:
```suggestion
            r#"CREATE TABLE db.table (`i` INT, `k` INT)"#,
```
since we're not testing the data type parsing here, we can use the same casing, and it let's us use `clickhouse_and_generic().verified_stmt()` instead?

iffyio (approved):
LGTM! cc @alamb

tests/sqlparser_clickhouse.rs
7phs: Makes sense.
I updated this unit-test.

7phs:
Thanks @iffyio for review!

alamb (approved):
Love it -- thank you @7phs  and @iffyio

alamb:
Screenshot 2024-06-07 at 7 10 43 AM

It appears this PR has some conflicts that need to be addressed, but then we can merge it in.

Thanks again

7phs:
Thanks @alamb for review and that you letting me know about conflicts.

I rebased my PR with `main` and resolved conflicts.

alamb:
Thanks @7phs !

Issue #1283, opened 2024-05-22
Regression parsing array slice expressions (`foo[1:2]`) in `0.46`


alamb:
As reported by @tisonkun  in https://github.com/sqlparser-rs/sqlparser-rs/pull/1215#issuecomment-2119108808

This query ran in version `0.45.0`
```sql
select make_array(1, 2, 3)[1:2], make_array(1.0, 2.0, 3.0)[2:3], make_array('h', 'e', 'l', 'l', 'o')[2:4];
```

The query fails in version `0.46.0`

```sql
External error: query failed: DataFusion error: SQL error: ParserError("Expected variant object key name, found: 2")
```

It appears to have been introduced in https://github.com/sqlparser-rs/sqlparser-rs/pull/1215

@jmhain 's analysis is https://github.com/sqlparser-rs/sqlparser-rs/pull/1215#issuecomment-2125141230

alamb:
This was fixed in 0.48: https://github.com/sqlparser-rs/sqlparser-rs/pull/1290

PR #1290, opened 2024-05-27
Support for Postgres array slice syntax

jmhain:
This is related to #1283, though I don't know what SQL dialect DataFusion uses so I'm not sure whether this alone resolves it. cc @alamb @tisonkun 

Prior to 0.46, we were able to parse an expression such as `select make_array(1, 2, 3)[1:2]`, but this was only because we were incorrectly interpreting the `1:2` as a JSON access (Snowflake syntax) of a property `2` on the value `1`. (Snowflake doesn't actually allow the field name to start with a number.)

On PostgreSQL, this syntax represents an array slice access. An important detail here is that these two syntaxes are mutually exclusive, since the array slice syntax can accept arbitrary expressions for either bound. For example, `array[foo:bar]` on Snowflake means: "access the property of `array` whose name is the value of the `bar` field in `foo`", while on Postgres it means "return the subarray starting at the index from column `foo` until the index from the column `bar`".

As part of this change, I also renamed `ArrayIndex` to `Subscript`. My intention is to remove the `MapAccess` variant and subsume that functionality into `Subscript` and `CompositeAccess` in a follow-up, as it's entirely redundant.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -820,18 +820,19 @@ pub enum Subscript {
     /// {2,3,4,5}
     /// ```
     ///
-    /// Stride notation is also supported
+    /// The lower and/or upper bound can be omitted to slice from the start or
+    /// end of the array respectively.
+    ///
+    /// See <https://www.postgresql.org/docs/current/arrays.html#ARRAYS-ACCESSING>.
+    ///
+    /// Also supports an optional "stride" as the last element (this is not
+    /// supported by postgres), e.g.
     ///
     /// ```plaintext
     /// => select (array[1,2,3,4,5,6])[1:6:2];
     /// -----------
     /// {1,3,5}
     /// ```
-    ///
-    /// The lower and/or upper bound can be omitted to slice from the start or
-    /// end of the array respectively.
-    ///
-    /// See <https://www.postgresql.org/docs/current/arrays.html#ARRAYS-ACCESSING>.
     Slice {
         lower_bound: Option<Expr>,
         upper_bound: Option<Expr>,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -820,6 +820,14 @@ pub enum Subscript {
     /// {2,3,4,5}
     /// ```
     ///
+    /// Stride notation is also supported
+    ///
+    /// ```plaintext
+    /// => select (array[1,2,3,4,5,6])[1:6:2];
+    /// -----------
+    /// {1,3,5}
+    /// ```
+    ///
     /// The lower and/or upper bound can be omitted to slice from the start or
     /// end of the array respectively.
     ///
@@ -827,6 +835,7 @@ pub enum Subscript {
     Slice {
         lower_bound: Option<Expr>,
         upper_bound: Option<Expr>,
+        stride: Option<Expr>,
     },
 }
 
@@ -837,6 +846,7 @@ impl fmt::Display for Subscript {
             Subscript::Slice {
                 lower_bound,
                 upper_bound,
+                stride,
             } => {
                 if let Some(lower) = lower_bound {
                     write!(f, "{lower}")?;
@@ -844,6 +854,10 @@ impl fmt::Display for Subscript {
                 write!(f, ":")?;
                 if let Some(upper) = upper_bound {
                     write!(f, "{upper}")?;
+                }
+                if let Some(stride) = stride {
+                    write!(f, ":")?;
+                    write!(f, "{stride}")?;
                 }
                 Ok(())
             }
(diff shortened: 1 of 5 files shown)


tisonkun:
@jmhain Cool! I'm going to test this branch on the datafusion patch.

Does this PR change user-facing APIs or it fixes the parsing logic internally?

tisonkun:
It's ready at https://github.com/apache/datafusion/pull/10392. Seems the AST changes a non-trivial to adapt. I suppose we should treat the datafusion case as a regression test for this PR :D

tisonkun:
This patch cannot parse `select make_array(1, 2, 3)[1:2:2]`, which is previously:

```
ArrayIndex {
    obj: "...",
    indexes: [
        JsonAccess {
            left: JsonAccess {
                left: Value(Number("1", false)),
                operator: Colon,
                right: Value(Number("2", false)),
            },
            operator: Colon,
            right: Value(Number("2", false)),
        },
    ],
},
```

tisonkun:
And we may need to pack `Subscipt` as a variant in Expr, so that we can translate it into a logical plan like ListSlices. The planner method generally accepts an Expr.

src/ast/mod.rs
```diff
@@ -745,10 +745,10 @@ pub enum Expr {
...
     Dictionary(Vec<DictionaryField>),
-    /// An array index expression e.g. `(ARRAY[1, 2])[1]` or `(current_schemas(FALSE))[1]`
-    ArrayIndex {
-        obj: Box<Expr>,
-        indexes: Vec<Expr>,
+    /// An access of nested data using subscript syntax, for example `array[2]`.
+    Subscript {
+        expr: Box<Expr>,
+        subscript: Box<Subscript>,
     },
```
tisonkun, line 752: For the last comment, datafusion has a method:

```rust
    fn plan_indices(
        &self,
        expr: SQLExpr,
        schema: &DFSchema,
        planner_context: &mut PlannerContext,
    ) -> Result<GetFieldAccess> {
```

where takes only the `subscript` part. And:

```rust
    fn plan_indexed(
        &self,
        expr: Expr,
        mut keys: Vec<SQLExpr>,
        schema: &DFSchema,
        planner_context: &mut PlannerContext,
    ) -> Result<Expr> {
```

where the `keys` is the `subscript` part (`vec![subscript]`). The logic is shared with MapAccess so I don't think it's easy to refactor.

tisonkun:
Another previously pattern: `select make_array(1, 2, 3)[1:2][2]` now panic with: `called `Result::unwrap()` on an `Err` value: ParserError("Expected variant object key name, found: 2 at Line: 1, Column 30")`.

src/ast/mod.rs
tisonkun: We may still have:

```rust
    ArrayIndex {
        obj: Box<Expr>,
        indexes: Vec<Expr>,
    },
```

which a new variant:

```
Subscript { ... }
```

that becomes the `indexes` part.

alamb: I have an idea how to fix this -- working on it

alamb:
Sorry -- I accidentally pushed the commit 4df6dc8 directly to this branch. We can revert it if we don't like it it. 

I am going to see if I can make datafusion work with this

jmhain:
@alamb No worries, thank you for taking care of this! Quick question: is this stride syntax DataFusion-specific? If so, it'd be good for us to add some comments as such because currently the code makes it seem like this is a Postgres feature.

alamb:
This seems to have been enough to unblock our upgrade here: https://github.com/apache/datafusion/pull/10392

> @alamb No worries, thank you for taking care of this! Quick question: is this stride syntax DataFusion-specific? If so, it'd be good for us to add some comments as such because currently the code makes it seem like this is a Postgres feature.

I was going to say "I hope not as DataFusion tries not to invest new syntax".... But I think it might be 🤦  -- I took a quick google around and couldn't find any other databases that permit the stride syntax. I'll add a note to this PR

tisonkun:
@alamb do you test this kind of SQL `select make_array(1, 2, 3)[1:2][2]`?

I don't find cases to guard regressions on slices combo.

jmhain:
@tisonkun There's [this](https://github.com/sqlparser-rs/sqlparser-rs/pull/1290/files#diff-8f97a2d7b555175934fcd50fd0f2718df01104ae1ac576e4cf5bab8e1a9a76d4R2054) double slice round-trip test and iirc there are a few existing tests with multiple subscripts.

@alamb Yeah this is all good with me. It is an API break though, so I don't think we can get away with a version number like 0.46.1.

alamb:
> @tisonkun There's [this](https://github.com/sqlparser-rs/sqlparser-rs/pull/1290/files#diff-8f97a2d7b555175934fcd50fd0f2718df01104ae1ac576e4cf5bab8e1a9a76d4R2054) double slice round-trip test and iirc there are a few existing tests with multiple subscripts.

I also added an explicit test in def916c

> @alamb Yeah this is all good with me. It is an API break though, so I don't think we can get away with a version number like 0.46.1.

Agreed -- I'll just make a normal release maybe

alamb (approved):
Thanks again @jmhain  and @tisonkun

PR #1292, opened 2024-05-28
ClickHouse: create view with fields and data types

7phs:
The CREATE VIEW syntax for ClickHouse is a bit different from many warehouses in the sense that you can specify the type of the columns when creating the view instead of them being deferred.

This PR extends parsing `CREATE VIEW` with this way of declaring view in ClickHouse.

Related issue: https://github.com/sqlparser-rs/sqlparser-rs/issues/1277

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7122,7 +7122,17 @@ impl<'a> Parser<'a> {
         } else {
             None
         };
-        Ok(ViewColumnDef { name, options })
+        let data_type = if dialect_of!(self is ClickHouseDialect) {
+            let (data_type, _) = self.parse_data_type_helper()?;
+            Some(data_type)
+        } else {
+            None
+        };
+        Ok(ViewColumnDef {
+            name,
+            data_type,
+            options,
+        })
     }
 
     /// Parse a parenthesized comma-separated list of unqualified, possibly quoted identifiers
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7123,8 +7123,7 @@ impl<'a> Parser<'a> {
             None
         };
         let data_type = if dialect_of!(self is ClickHouseDialect) {
-            let (data_type, _) = self.parse_data_type_helper()?;
-            Some(data_type)
+            Some(self.parse_data_type()?)
         } else {
             None
         };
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -7122,7 +7122,17 @@ impl<'a> Parser<'a> {
         } else {
             None
         };
-        Ok(ViewColumnDef { name, options })
+        let data_type = if dialect_of!(self is ClickHouseDialect) {
+            let (data_type, _) = self.parse_data_type_helper()?;
```
iffyio, line 7126:
```suggestion
            let (data_type, _) = self.parse_data_type()?;
```
should this be parse_data_type instead?

tests/sqlparser_clickhouse.rs
```diff
@@ -220,6 +220,40 @@ fn parse_create_table() {
     );
 }
 
+#[test]
+fn parse_create_view_with_fields_data_types() {
+    match clickhouse().one_statement_parses_to(
+        r#"CREATE VIEW v (i "int", f String) AS SELECT * FROM t"#,
+        r#"CREATE VIEW v (i "int", f STRING) AS SELECT * FROM t"#,
```
iffyio, line 227: Are we testing something specific about the behavior with the different casing? Thinking if not we can probably use verified_stmt with one canonical representation of the text?

tests/sqlparser_clickhouse.rs
```diff
@@ -220,6 +220,40 @@ fn parse_create_table() {
     );
 }
 
+#[test]
+fn parse_create_view_with_fields_data_types() {
+    match clickhouse().one_statement_parses_to(
```
iffyio, line 225: Can we add a test case that clickhouse fails parsing if the datatype is missing?

7phs: Makes sense.
Added a test with invalid query `CREATE VIEW` without data types.

tests/sqlparser_clickhouse.rs
7phs: Updated with `verified_stmt`.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @7phs  and @iffyio  for the review 🙏

PR #1299, opened 2024-06-03
Adding support for PostgreSQL's CustomOperator(String)

amrutadotorg:
This code adds the following PGroonga operators to the BinaryOperator enum:

PGroongaFullTextSearch for the &@ operator
PGroongaEasyQuerySearch for the &@~ operator
PGroongaSimilarSearch for the &@* operator
PGroongaAdvancedSearch for the `&`` operator
PGroongaArrayKeywordSearch for the &@| operator
PGroongaArrayQuerySearch for the &@~| operator

Additionally, the fmt::Display implementation has been updated to handle these new operators appropriately.

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -240,6 +240,18 @@ pub enum BinaryOperator {
     /// See [CREATE OPERATOR](https://www.postgresql.org/docs/current/sql-createoperator.html)
     /// for more information.
     PGCustomBinaryOperator(Vec<String>),
+    /// PGroonga Full text search by a keyword operator, e.g. `a &@ b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
+    PGroongaFullTextSearch,
+    /// PGroonga Full text search by easy to use query language operator, e.g. `a &@~ b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
+    PGroongaEasyQuerySearch,
+    /// PGroonga Similar search operator, e.g. `a &@* b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
+    PGroongaSimilarSearch,
+    /// PGroonga Advanced search by ECMAScript like query language operator, e.g. `a &` b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
+    PGroongaAdvancedSearch,
+    /// PGroonga Full text search by an array of keywords operator, e.g. `a &@| b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
+    PGroongaArrayKeywordSearch,
+    /// PGroonga Full text search by an array of queries in easy to use query language operator, e.g. `a &@~| b` (PGroonga-specific. See <https://pgroonga.github.io/reference/>)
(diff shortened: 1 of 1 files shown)


lovasoa (changes requested):
I don't think this works. You didn't write any parsing code for these new operators.

And I think a better approach would be to have a `CustomOperator(String)`, because anyone can define any operator, and it would be great if we were able to parse them independently of the particular case of pgroonga.

I think we need to work on this in the tokenizer.

lovasoa:
The documentation for allowed characters in custom operators is here: https://www.postgresql.org/docs/current/sql-createoperator.html

amrutadotorg:
what would you of having something like this in the _tokenizer.rs_ ?

```
pub enum Token {
    EOF,
    Word(Word),
    Number(String, bool),
    Char(char),
    SingleQuotedString(String),
    DoubleQuotedString(String),
    CustomOperator(String),
}
```

```
impl<'a> Tokenizer<'a> {
    fn next_token(&mut self) -> Token {
        // Existing logic...

        // Detect and return a `CustomOperator` token if a custom operator is found.
        let start_pos = self.chars.clone();  // Save the starting position
        let mut operator = String::new();
        while let Some(&ch) = self.chars.peek() {
            // Check if the character is valid for a custom operator
            if "+-*/<>~=!@#%^&|`?".contains(ch) {
                operator.push(ch);
                self.chars.next();
            } else {
                break;
            }
        }

        if operator.len() > 1 && (operator.ends_with('+') || operator.ends_with('-')) {
            if !operator.contains('~') && !operator.contains('!') && !operator.contains('@') &&
               !operator.contains('#') && !operator.contains('%') && !operator.contains('^') &&
               !operator.contains('&') && !operator.contains('|') && !operator.contains('`') && 
               !operator.contains('?') {
                // Invalid operator according to the rules <https://www.postgresql.org/docs/current/sql-createoperator.html>
                self.chars = start_pos; // Rewind to the start position
                return Token::Char(self.chars.next().unwrap());  // Fallback to a single char token
            }
        }

        if !operator.is_empty() {
            return Token::CustomOperator(operator);
        }

        // Fallback to other token parsing logic
        // ...
    }
}
```

lovasoa:
No, you would have to integrate with the existing operator tokenization logic not to break it. 

It's a little bit technical, so if what you are interested in is just getting full text search to work for you, the easiest is probably for you to declare your own function that wraps pgroonga's custom operator, and can that function instead of the operator in your code to work around the current operator parsing bug.

lovasoa:
@alamb , currently, the tokenizer is dialect-independent. 

However, the same expression `a&@b` would be tokenized as `identifier(a) operator(&@) identifier(b)` in postgres and `identifier(a) operator(&) identifier(@b)` in Microsoft SQL Server. 

Do you think it would be okay to add partial support for custom operators, ignoring the ones that end in @ and are not followed by a whitespace? Or would you require full support for differentiated tokenization before merging support for custom operators?

lovasoa (changes requested):
I think this PR can be closed in favor of https://github.com/sqlparser-rs/sqlparser-rs/pull/1302

PR #1302, opened 2024-06-06
add support for custom operators in postgres

lovasoa:
Fixes https://github.com/sqlparser-rs/sqlparser-rs/issues/1298
Fixes https://github.com/lovasoa/SQLpage/issues/372
Closes https://github.com/sqlparser-rs/sqlparser-rs/pull/1299

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -966,15 +966,12 @@ impl<'a> Tokenizer<'a> {
                         Some('>') => {
                             chars.next();
                             match chars.peek() {
-                                Some('>') => {
-                                    chars.next();
-                                    Ok(Some(Token::LongArrow))
-                                }
-                                _ => Ok(Some(Token::Arrow)),
+                                Some('>') => self.consume_for_binop(chars, "->>", Token::LongArrow),
+                                _ => self.start_binop(chars, "->", Token::Arrow),
                             }
                         }
                         // a regular '-' operator
-                        _ => Ok(Some(Token::Minus)),
+                        _ => self.start_binop(chars, "-", Token::Minus),
                     }
                 }
                 '/' => {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1008,28 +1008,24 @@ impl<'a> Tokenizer<'a> {
                         Some(sch) if self.dialect.is_identifier_start('%') => {
                             self.tokenize_identifier_or_keyword([ch, *sch], chars)
                         }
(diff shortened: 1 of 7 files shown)


lovasoa:
@alamb , could you have a look ? The PR is quite simple, it extends the tokenizer to be able to parse [postgresql custom operators](https://www.postgresql.org/docs/current/sql-createoperator.html).

I (and @amrutadotorg) need that for SQLPage.

src/tokenizer.rs
```diff
@@ -1206,6 +1212,41 @@ impl<'a> Tokenizer<'a> {
...
+            }
+            s.push(ch);
+            is_custom_operator = true;
+            chars.next();
+        }
+        if is_custom_operator {
+            Ok(Some(Token::CustomBinaryOperator(s)))
+        } else {
+            Ok(Some(default))
+        }
```
jmhain, line 1247: We could avoid the string allocation for the common case of a non-custom binop, something like:

```suggestion
        let mut custom = None;
        while let Some(&ch) = chars.peek() {
            if !self.dialect.is_custom_operator_part(ch) {
                break;
            }

            custom.get_or_insert_with(|| prefix.to_string()).push(ch);
            chars.next();
        }

        Ok(Some(custom.map(Token::CustomBinaryOperator).unwrap_or(default)))
```

jmhain (approved):
nice, this looks great! cc @alamb

src/tokenizer.rs
lovasoa: Very True ! Let's do it

lovasoa: done

tests/sqlparser_postgres.rs
```diff
@@ -1770,18 +1793,73 @@ fn parse_pg_binary_ops() {
...
+    }
+}
+
+#[test]
+fn parse_pg_custom_binary_ops() {
+    // Postgres supports declaring custom binary operators, using any character in the following set:
+    //  + - * / < > = ~ ! @ # % ^ & | ` ?
+
+    // Here, we test the ones used by common extensions
+    let operators = [
```
alamb, line 1806: 😮  -- very cool -- thanks @lovasoa

alamb (approved):
Thank you @lovasoa  and @jmhain

lovasoa:
Thank you for merging ! This will open a world of new possibilities in SQLPage, using new postgres extensions.

PR #1291, opened 2024-05-27
Encapsulate `CreateTable`, `CreateIndex` into specific structs

philipcristiano:
cc @alamb, @tisonkun 

Continue moving structs as part of #1204 

The `CreateTableBuilder` could maybe be combined into the `CreateTable`. For the purposes of minimizing breaking changes I did not attempt to merge them.

PR #1307, opened 2024-06-11
CreateIndex: Move Display fmt to struct

philipcristiano:
Move the Display trait `fmt` implementation to the CreateIndex struct to to allow for display!

The struct was created in a0f511cb21efd9e992fb4c6ef273ef41802d308c with the CreateTable `fmt` being moved, but the CreateIndex was not.

Follows up from #1291


jmhain (approved):
lgtm, cc @alamb

alamb (approved):
nice -- thank you @philipcristiano  and @jmhain

PR #1309, opened 2024-06-12
Support use of `BY NAME` quantifier across all set ops

alexander-beedie:
It turns out that that use of the "BY NAME" quantifier was being artificially constrained to `UNION` set ops - however, there _are_ dialects that allow it for _all_ set ops (which actually makes a lot of sense).

I'm about to implement this additional "BY NAME" set op support for the Polars SQL interface (which will need this patch), but  the expanded usage _is_ already found in the wild; see Microsoft's U-SQL dialect (as used in Azure Data Lake Analytics) for an example:

* `EXCEPT [BY NAME]`: https://learn.microsoft.com/en-gb/archive/usql/statements-and-expressions/set-rowset/except-expression
* `INTERSECT [BY NAME]`: https://learn.microsoft.com/en-gb/archive/usql/statements-and-expressions/set-rowset/intersect-expression

**Also:**
* Spotted a missing backtick in the docs - added.
* Fixed some minor typos in comments/docstrings.
* Added a note to the README that `polars` also uses the parser 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,12 +114,11 @@ $ cargo run --features json_example --example cli FILENAME.sql [--dialectname]
 
 ## Users
 
-This parser is currently being used by the [DataFusion] query engine,
-[LocustDB], [Ballista], [GlueSQL], [Opteryx], [PRQL], [Qrlew], [JumpWire], and [ParadeDB].
+This parser is currently being used by the [DataFusion] query engine, [LocustDB],
+[Ballista], [GlueSQL], [Opteryx], [Polars], [PRQL], [Qrlew], [JumpWire], and [ParadeDB].
 
 If your project is using sqlparser-rs feel free to make a PR to add it
 to this list.
-
 
 ## Design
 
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -8138,7 +8138,7 @@ impl<'a> Parser<'a> {
 
     pub fn parse_set_quantifier(&mut self, op: &Option<SetOperator>) -> SetQuantifier {
         match op {
-            Some(SetOperator::Union) => {
+            Some(SetOperator::Except) | Some(SetOperator::Intersect | SetOperator::Union) => {
```
jmhain, line 8141:
```suggestion
            Some(SetOperator::Except | SetOperator::Intersect | SetOperator::Union) => {
```

jmhain (approved):
lgtm, many thanks for the typo cleanups! cc @alamb

README.md
```diff
@@ -210,6 +209,7 @@ licensed as above, without any additional terms or conditions.
 [Ballista]: https://github.com/apache/arrow-ballista
 [GlueSQL]: https://github.com/gluesql/gluesql
 [Opteryx]: https://github.com/mabel-dev/opteryx
+[Polars]: https://pola.rs/
```
alamb, line 212: 🎉

alamb (approved):
Thank you @jmhain  and @alexander-beedie

PR #1310, opened 2024-06-13
Fix typo in `sqlparser-derive` README

leoyvens:
This is just from reading it, but I think that bit is wrong.


lovasoa (approved):

alamb (approved):
Thank you @leoyvens  and @lovasoa

PR #1317, opened 2024-06-17
add support for update statements that contain tuple assignments

lovasoa:
This allows the correct parsing of

```sql
UPDATE x SET (a, b) = (1, 2)
```

while also maintaining the parsing of

```sql
UPDATE x SET a.b = 1
```

This includes a change to the ast::Assignment struct

Fixes https://github.com/sqlparser-rs/sqlparser-rs/issues/1311
Fixes https://github.com/lovasoa/SQLpage/issues/403

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4594,13 +4594,35 @@ impl fmt::Display for GrantObjects {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct Assignment {
-    pub id: Vec<Ident>,
+    pub target: AssignmentTarget,
     pub value: Expr,
 }
 
 impl fmt::Display for Assignment {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} = {}", display_separated(&self.id, "."), self.value)
+        write!(f, "{} = {}", self.target, self.value)
+    }
+}
+
+/// Left-hand side of an assignment in an UPDATE statement,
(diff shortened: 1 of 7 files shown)


src/ast/mod.rs
```diff
@@ -4594,13 +4594,35 @@ impl fmt::Display for GrantObjects {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct Assignment {
-    pub id: Vec<Ident>,
+    pub target: AssignmentTarget,
```
alamb, line 4597: I wonder if there is a reason to distinguish between a single column update and a multi column update?

I think it is common to update more than one column at a time in an UPDATE statement, and there is no reason to treat single column and multi-column updates differently

As in perhaps this could simply be

```suggestion
    pub target: Vec<ObjectName>,
```

alamb (approved):
Thank you for the contribution @lovasoa  -- I left a suggestion, but I also think this method works too so we could merge the PR without changes too

Let me know what you think

src/ast/mod.rs
lovasoa: @alamb : We need a way to distinguish between `UPDATE t SET c = expr` and `UPDATE t SET (c) = expr`. The first one is the traditional UPDATE syntax (which supports multiple assignments, but that's not the question here), and the second one is the tuple update syntax, which expects `expr` to be a subselect or a row expression. The two are semantically different, so we need a way to represent them differently.

alamb: Makes sense to me -- thank you for the clarification

PR #1318, opened 2024-06-17
Re-enable trailing commas in DCL

### LINKED ISSUE #1211: Error on some `GRANT`s when trailing commas are enabled

MohamedAbdeen21:
This PR should close #1211 and is a follow-up to #1212 which ignores trailing commas option when parsing DCLs.

Tests included.

cc @lovasoa

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3278,6 +3278,33 @@ impl<'a> Parser<'a> {
         ret
     }
 
+    pub fn parse_actions_list(
+        &mut self,
+    ) -> Result<Vec<(Keyword, Option<Vec<Ident>>)>, ParserError> {
+        let mut values = vec![];
+        loop {
+            values.push(self.parse_grant_permission()?);
+            if !self.consume_token(&Token::Comma) {
+                break;
+            } else if self.options.trailing_commas {
+                match self.peek_token().token {
+                    Token::Word(kw)
+                        if keywords::RESERVED_FOR_GRANT_PERMISSIONS.contains(&kw.keyword) =>
+                    {
+                        break;
+                    }
+                    Token::RParen
+                    | Token::SemiColon
+                    | Token::EOF
+                    | Token::RBracket
+                    | Token::RBrace => break,
+                    _ => continue,
+                }
+            }
+        }
+        Ok(values)
+    }
+
     /// Parse a comma-separated list of 1+ items accepted by `F`
     pub fn parse_comma_separated<T, F>(&mut self, mut f: F) -> Result<Vec<T>, ParserError>
     where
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -3278,6 +3278,31 @@ impl<'a> Parser<'a> {
         ret
     }
 
+    pub fn parse_actions_list(
+        &mut self,
+    ) -> Result<Vec<(Keyword, Option<Vec<Ident>>)>, ParserError> {
```
MohamedAbdeen21, line 3283: Getting a clippy warning about `very complex type`, can I just ignore it? or do we prefer aliasing the type?

lovasoa: The type is indeed hard to read. I think the warning shouldn't be ignored.

Shouldn't the `Vec<Ident>` be an `ObjectName` ?

src/parser/mod.rs
```diff
@@ -3278,6 +3278,32 @@ impl<'a> Parser<'a> {
         ret
     }
 
+    #[allow(clippy::type_complexity)]
```
lovasoa, line 3281: better to actually fix the problem

lovasoa (changes requested):

src/parser/mod.rs
MohamedAbdeen21: pushed

lovasoa (approved):

alamb:
Thanks @lovasoa  and @MohamedAbdeen21  -- I am sorry I didn't see this PR earlier.

alamb (approved):
Looks great to me -- thank you

PR #1319, opened 2024-06-19
Improve error messages with additional colons

### LINKED ISSUE #1259: Add missing colon to sql parser error message after "Column" and "Expected"
sql parser error messages are missing a colon after "Column" and "Excepted"

How it is:
`sql parser error: Expected (, found: 2345 at Line: 2, Column 8`

How it should be IMO:
`sql parser error: Expected: (, found: 2345 at Line: 2, Column: 8`

LorrensP-2158466:
This should close #1259.

Added colons at the end of `Expected` and `Column` in ParserErrors.
Tests are also updated and they passed.

I do think this creates another "issue":
Currently, a lot of errors read nice without the colon after `Expected`:
"Expected an expression, found: ... "
and others would not read nice:
"Expected (, found: ..."

But with this change, we would flip these experiences:
"Expected: an expression, found: ..." and "Expected (, found: ..."

I do think we should keep the colons, but change the error messages, this is a bit much more work, but if asked i can do it.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3116,7 +3116,7 @@ impl<'a> Parser<'a> {
     /// Report `found` was encountered instead of `expected`
     pub fn expected<T>(&self, expected: &str, found: TokenWithLocation) -> Result<T, ParserError> {
         parser_err!(
-            format!("Expected {expected}, found: {found}"),
+            format!("Expected: {expected}, found: {found}"),
             found.location
         )
     }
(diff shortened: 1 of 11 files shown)


alamb (commented):
Thank you @LorrensP-2158466  -- this looks good to me

Any concerns @jmhain @iffyio or @lovasoa ?

alamb (approved):

iffyio (approved):
Looks good to me! Thanks @LorrensP-2158466!

lovasoa (approved):

alamb:
Thanks everyone!

PR #1320, opened 2024-06-20
Allow semi-colon at the end of UNCACHE statement

### LINKED ISSUE #1244: UNCACHE fails to parse with a semicolon
`parse_sql` normally doesn't care whether a single statement is terminated by a semicolon or end of string, but when parsing `UNCACHE` it does.

This deviation seems like a bug.

Discovered on sqlparser v0.43.1, tested on v0.45.0:

```rust
use sqlparser::{
    dialect::GenericDialect,
    parser::{Parser, ParserError},
};

fn main() {

    // With other SQL, semicolon being there or not does not seem to matter:
    {
        let no_semicolon = Parser::parse_sql(&dialect, "select 1");
        assert!(no_semicolon.is_ok());
    }
    {
        let with_semicolon = Parser::parse_sql(&dialect, "select 1;");
        assert!(with_semicolon.is_ok());
    }

    // With the UNCACHE statement, it does seem to matter.
    // Parsing only works when the semicolon is omitted:
    {
        let no_semicolon = Parser::parse_sql(&dialect, "uncache table foo");
        println!("no semicolon: {no_semicolon:?}");
        assert!(no_semicolon.is_ok());
    }
    {
        // BUG
        let with_semicolon = Parser::parse_sql(&dialect, "uncache table foo;");
        println!("with semicolon: {with_semicolon:?}");
        assert_eq!(
            with_semicolon,
            Err(ParserError::ParserError(
                "Expected an `EOF`, found: ; at Line: 1, Column 18".to_string()
            ))
        );
    }
}
```

LorrensP-2158466:
Closes #1244.

The function that parses UNCACHE statements now also accepts a semi-colon as the end of the statement.

Had to modify one test because the error message `expected ...` now also contains the semi-colon.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3611,21 +3611,13 @@ impl<'a> Parser<'a> {
 
     /// Parse a UNCACHE TABLE statement
     pub fn parse_uncache_table(&mut self) -> Result<Statement, ParserError> {
-        let has_table = self.parse_keyword(Keyword::TABLE);
-        if has_table {
-            let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
-            let table_name = self.parse_object_name(false)?;
-            if self.peek_token().token == Token::EOF {
-                Ok(Statement::UNCache {
-                    table_name,
-                    if_exists,
-                })
-            } else {
-                self.expected("an `EOF`", self.peek_token())
-            }
-        } else {
-            self.expected("a `TABLE` keyword", self.peek_token())
-        }
+        self.expect_keyword(Keyword::TABLE)?;
+        let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
+        let table_name = self.parse_object_name(false)?;
+        Ok(Statement::UNCache {
+            table_name,
+            if_exists,
+        })
     }
 
     /// SQLite-specific `CREATE VIRTUAL TABLE`
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -8451,19 +8451,31 @@ fn parse_uncache_table() {
 
     let res = parse_sql_statements("UNCACHE TABLE 'table_name' foo");
     assert_eq!(
+<<<<<<< HEAD
         ParserError::ParserError("Expected: an `EOF`, found: foo".to_string()),
+=======
+        ParserError::ParserError("Expected: end of statement, found: foo".to_string()),
+>>>>>>> c2fb339 (change tests)
         res.unwrap_err()
     );
 
     let res = parse_sql_statements("UNCACHE 'table_name' foo");
     assert_eq!(
+<<<<<<< HEAD
         ParserError::ParserError("Expected: a `TABLE` keyword, found: 'table_name'".to_string()),
+=======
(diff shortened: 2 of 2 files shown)


src/parser/mod.rs
```diff
@@ -3615,13 +3615,15 @@ impl<'a> Parser<'a> {
...
                     table_name,
                     if_exists,
-                })
-            } else {
-                self.expected("an `EOF`", self.peek_token())
+                }),
+                token_loc => self.expected("`EOF` or `;`", token_loc),
             }
         } else {
             self.expected("a `TABLE` keyword", self.peek_token())
```
iffyio, line 3629: (not necessarily related to this PR) looks like we can avoid the has_table condition altogether by using `self.expect_keyword` instead of `self.parse_keyword` to parse the `TABLE` keyword? if so could we make that change to simplify the function in the same go (assuming its a straightforward change)

src/parser/mod.rs
```diff
@@ -3615,13 +3615,15 @@ impl<'a> Parser<'a> {
         if has_table {
             let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
             let table_name = self.parse_object_name(false)?;
-            if self.peek_token().token == Token::EOF {
-                Ok(Statement::UNCache {
+            match self.peek_token() {
+                TokenWithLocation {
+                    token: Token::EOF | Token::SemiColon,
+                    ..
```
iffyio, line 3621: hmm I'm wondering if we don't do an explicitly check for a terminal token after all? looking at the other stmt parse functions seem that's how they handle it, by only parsing the components they need and exiting

LorrensP-2158466: Yeah that's true, but for some reason this parse function only accepts EOF as and of statement. I'll see if we can remove this check.

src/parser/mod.rs
LorrensP-2158466: Oh nice catch! I can fix this in this PR if this is straightforward, otherwise I can open up a new issue.

LorrensP-2158466:
I did what you recomended.
The `self.expect_keyword` works: i just had to change the string in the tests.
And letting the `self.parse_statements` function handle end of statement instead of this function also works, i just had to change the error strings in the tests.

Thanks for the help!

alamb:
Looks like this PR now has a conflict

LorrensP-2158466:
I did a dumb thing, it's fixed now i think.

iffyio (approved):
LGMT! cc @alamb

src/parser/mod.rs
```diff
@@ -3611,21 +3611,13 @@ impl<'a> Parser<'a> {
...
-                    if_exists,
-                })
-            } else {
-                self.expected("an `EOF`", self.peek_token())
-            }
-        } else {
-            self.expected("a `TABLE` keyword", self.peek_token())
-        }
+        self.expect_keyword(Keyword::TABLE)?;
+        let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
```
alamb, line 3615: ❤️

alamb (approved):
Thank you @LorrensP-2158466 -- it is great to see a bug fixed by simplifying the code

Thank you @iffyio  for nthe review

PR #1321, opened 2024-06-23
Fix Snowflake `SELECT *` wildcard `REPLACE ... RENAME` order

alexander-beedie:
Affects parsing of Snowflake `SELECT * <opts>` wildcard options that contain _both_ `RENAME` and `REPLACE` options.

As per the Snowflake docs[^1]:
```
If you specify REPLACE and RENAME:
You must specify REPLACE first
```
Made a minor fix and added a unit test to validate it and provide the missing coverage.

[^1]: https://docs.snowflake.com/en/sql-reference/sql/select#parameters


iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @alexander-beedie  and @iffyio

PR #1322, opened 2024-06-24
Support for DuckDB Union datatype

gstvg:
e.g. `CREATE TABLE tbl1 (u UNION(num INTEGER, str VARCHAR));`

[https://duckdb.org/docs/sql/data_types/union.html](https://duckdb.org/docs/sql/data_types/union.html)

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2246,6 +2246,32 @@ impl<'a> Parser<'a> {
         ))
     }
 
+    /// DuckDB specific: Parse a Union type definition as a sequence of field-value pairs.
+    ///
+    /// Syntax:
+    ///
+    /// ```sql
+    /// UNION(field_name field_type[,...])
+    /// ```
+    ///
+    /// [1]: https://duckdb.org/docs/sql/data_types/union.html
+    fn parse_union_type_def(&mut self) -> Result<Vec<UnionField>, ParserError> {
+        self.expect_keyword(Keyword::UNION)?;
+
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -2246,6 +2246,32 @@ impl<'a> Parser<'a> {
...
+    /// UNION(field_name field_type[,...])
+    /// ```
+    ///
+    /// [1]: https://duckdb.org/docs/sql/data_types/union.html
+    fn parse_union_type_def(&mut self) -> Result<Vec<UnionField>, ParserError> {
+        self.expect_keyword(Keyword::UNION)?;
+
+        self.expect_token(&Token::LParen)?;
+
+        let fields = self.parse_comma_separated0(|p| {
```
gstvg, line 2263: Zero field unions are invalid on DuckDB and don't even parse. Should we error here or let downstream users handle it?

[https://shell.duckdb.org/#queries=v0,create-table-tbl1-(u-union()-)~](https://shell.duckdb.org/#queries=v0,create-table-tbl1-(u-union()-)~)

jayzhan211: I think it is fine to return error here.

gstvg: Done

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Likewise -- looks good to me -- thank you @iffyio  and @gstvg

PR #1323, opened 2024-06-25
Support `GROUP BY WITH MODIFIER` for ClickHouse

git-hulk:
For the syntax, please refer to:
https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#rollup-modifier


src/ast/query.rs
```diff
@@ -1865,27 +1873,57 @@ impl fmt::Display for SelectInto {
...
+            GroupByExpr::All(modifiers) => {
+                write!(f, "GROUP BY ALL")?;
+                write!(f, " {}", display_separated(modifiers, " "))?;
+                Ok(())
+            }
+            GroupByExpr::Expressions(col_names, modifiers) => {
                 let col_names = display_comma_separated(col_names);
-                write!(f, "GROUP BY ({col_names})")
+                write!(f, "GROUP BY ({col_names})")?;
+                write!(f, " {}", display_separated(modifiers, " "))?;
```
iffyio, line 1925: In both clauses do we need to check that modifiers is not empty before displaying it? (Looks otherwise like we may get a trailing space in the output if there are no modifiers)

src/ast/query.rs
```diff
@@ -1865,27 +1873,57 @@ impl fmt::Display for SelectInto {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum WithModifier {
```
iffyio, line 1879:
```suggestion
pub enum GroupByWithModifier {
```
thinking we can mention group by in the name to make it unambiguous?

src/ast/query.rs
```diff
@@ -299,11 +299,19 @@ impl fmt::Display for Select {
...
+                }
+            }
+            GroupByExpr::Expressions(exprs, modifiers) => {
                 if !exprs.is_empty() {
                     write!(f, " GROUP BY {}", display_comma_separated(exprs))?;
                 }
+                if !modifiers.is_empty() {
+                    write!(f, " {}", display_separated(modifiers, " "))?;
+                }
             }
```
iffyio, line 315: this seems like its identical to the display method on GroupByExpr, if so would it make sense to simplify a bit by calling display instead?

src/ast/query.rs
iffyio, line 1879: Can we add a description and link to the clickhouse docs here? currently its only specified in the `ALL` variant which seems uneven in terms of placement.
Also it would help in this case with a syntax example in the comment like `GROUP BY year WITH ROLLUP` so its clear to folks at a glance what it does without consulting the docs

tests/sqlparser_clickhouse.rs
```diff
@@ -626,6 +626,24 @@ fn parse_create_materialized_view() {
     clickhouse_and_generic().verified_stmt(sql);
 }
 
+#[test]
+fn parse_group_by_with_modifier() {
+    match clickhouse_and_generic().verified_stmt("SELECT * FROM t GROUP BY x WITH ROLLUP WITH CUBE")
```
iffyio, line 631: For the tests, following cases I think we can cover

- We enumerate all the WITH modifiers we introduced so that they covered by some test
- We add `GROUP BY ALL WITH ...` examples
For above would probably make sense to use a table of scenarios to enumerate e.g pseudocode

```
let clauses = ["x", "a, b", "ALL"]
let modifiers = ["WITH ROLLUP", "WITH CUBE", "WITH ROLLUP WITH CUBE" ... ]
for clause in clauses
    for modifier in modifiers
         clickhouse_and_generic().verified_stmt!("SELECT * FROM t GROUP BY [clause} {modifier})
         ...
```

- Negative tests e.g `GROUP BY x WITH ROLLUP WITH`, `GROUP BY x WITH WITH ROLLUP`, `GROUP BY WITH ROLLUP`

src/parser/mod.rs
```diff
@@ -8287,13 +8287,36 @@ impl<'a> Parser<'a> {
...
+            let expressions = if self.parse_keyword(Keyword::ALL) {
+                None
             } else {
-                GroupByExpr::Expressions(self.parse_comma_separated(Parser::parse_group_by_expr)?)
+                Some(self.parse_comma_separated(Parser::parse_group_by_expr)?)
+            };
+
+            let mut modifiers = vec![];
+            loop {
+                if dialect_of!(self is ClickHouseDialect | GenericDialect)
```
iffyio, line 8298: The dialect check can live outside the loop? since it never changes, and we would let us avoid entering the loop for non-clickhouse dialects

src/parser/mod.rs
```diff
@@ -8287,13 +8287,36 @@ impl<'a> Parser<'a> {
...
+                    && self.parse_keyword(Keyword::WITH)
+                {
```
iffyio, line 8300: We can probably simplify the loop body with something like the following?
```rust
if !self.parse_keyword(Keyword::WITH) {
   break;
}
let modifier match self.expect_one_of_keywords(ROLLUP, CUBE, TOTALS) {
  ROLLUP => WithModifier::Rollup
  ..
};
modifiers.push(modifier);
```

git-hulk:
@iffyio Thanks for your great review and improvements. All of them are resolved, pls retake a look.

src/ast/query.rs
```diff
@@ -1865,27 +1858,68 @@ impl fmt::Display for SelectInto {
...
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum GroupByWithModifier {
+    /// ClickHouse supports GROUP BY WITH modifiers(includes ROLLUP|CUBE|TOTALS).
+    /// e.g. GROUP BY year WITH ROLLUP WITH TOTALS
+    ///
+    /// [ClickHouse]: <https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#rollup-modifier>
```
iffyio, line 1868:
```suggestion
/// ClickHouse supports GROUP BY WITH modifiers(includes ROLLUP|CUBE|TOTALS).
/// e.g. GROUP BY year WITH ROLLUP WITH TOTALS
///
/// [ClickHouse]: <https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#rollup-modifier>
#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
pub enum GroupByWithModifier {
```
since the doc applies to the enum (rather than the Rollup variant specifically

src/ast/query.rs
```diff
@@ -298,14 +298,7 @@ impl fmt::Display for Select {
...
         }
-        match &self.group_by {
-            GroupByExpr::All => write!(f, " GROUP BY ALL")?,
-            GroupByExpr::Expressions(exprs) => {
-                if !exprs.is_empty() {
-                    write!(f, " GROUP BY {}", display_comma_separated(exprs))?;
-                }
-            }
-        }
+        write!(f, "{}", self.group_by)?;
```
iffyio, line 301:
```suggestion
        write!(f, " {}", self.group_by)?;
```
Can we change the group_by display method to not include a leading space? its easier to have the caller insert the leading space, that way the display method is agnostic of where its being placed (and can be reused in different contexts as a result), and in this case each individual variant wont have to include its own leading space

src/ast/query.rs
```diff
@@ -1865,27 +1858,68 @@ impl fmt::Display for SelectInto {
...
+                write!(f, " GROUP BY ALL")?;
+                if !modifiers.is_empty() {
+                    write!(f, " {}", display_separated(modifiers, " "))?;
+                }
+                Ok(())
+            }
+            GroupByExpr::Expressions(col_names, modifiers) => {
+                if col_names.is_empty() {
+                    return Ok(());
+                }
```
iffyio, line 1916: would it make sense to skip this check instead? if I'm understanding correctly, it seems to change the current behavior - if the list of expressions is empty, the current behavior looks like it will display `GROUP BY`, but this new behavior won't display anything. Not sure if we can end up with an empty list today but I imagine if we suceeded in parsing an empty list then we would want the display to include the `GROUP BY` that it parsed so the current behavior seems reasonable in any case?

src/parser/mod.rs
```diff
@@ -8287,13 +8287,37 @@ impl<'a> Parser<'a> {
...
+                    let keyword = self.expect_one_of_keywords(&[
+                        Keyword::ROLLUP,
+                        Keyword::CUBE,
+                        Keyword::TOTALS,
+                    ])?;
+                    modifiers.push(match keyword {
+                        Keyword::ROLLUP => GroupByWithModifier::Rollup,
+                        Keyword::CUBE => GroupByWithModifier::Cube,
+                        Keyword::TOTALS => GroupByWithModifier::Totals,
+                        _ => unreachable!(),
```
iffyio, line 8311:
```suggestion
                        _ => return parser_err!("BUG: expected to match GroupBy modifier keyword", self.peek_token().location),
```
I think its reasonable to use unreachable in this scenario but we can return an arbitrary error to be safer in case there's a mistake in the future so that we fail graceful on the unlucky SQL vs crashing the system.

iffyio (commented):
@git-hulk left some minor comments, the PR looks good overall thanks!

src/ast/query.rs
git-hulk: @iffyio The old behavior also checks if the expressions is empty in Select's display: https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/query.rs#L304. So I think GroupBy should be omitted if the expressions is empty and move the check to GroupBy itself.

git-hulk:
@iffyio Thanks for your review, the left comments are resolved. Please take a review while you get time.

src/ast/query.rs
iffyio: Ah got it, thanks for clarifying! Yeah give that was the previous select behavior omits it then makes sense to preserve it

iffyio (approved):
LGTM! cc @alamb

tests/sqlparser_clickhouse.rs
```diff
@@ -626,6 +626,61 @@ fn parse_create_materialized_view() {
...
+                        );
+                    }
+                }
+                _ => unreachable!(),
+            }
+        }
+    }
+
+    // invalid cases
+    let invalid_cases = [
```
alamb, line 671: 💯  for negative cases

alamb (approved):
Thank you @git-hulk  and @iffyio

tests/sqlparser_clickhouse.rs
git-hulk: This should be credited to the help of @iffyio, thank you!

PR #1324, opened 2024-06-25
Support `DROP PROCEDURE` statement

### LINKED ISSUE #1282: Add support for DROP PROCEDURE
https://github.com/sqlparser-rs/sqlparser-rs/blob/9d15f7e9a92a4ef36fcc6d3166fd0221a22a20e2/src/parser/mod.rs#L4217
drop procedure stmt is allowed in many sql client impl.
(https://dev.mysql.com/doc/refman/8.0/en/drop-procedure.html,
https://learn.microsoft.com/en-us/sql/t-sql/statements/drop-procedure-transact-sql?view=sql-server-ver16,
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/DROP-PROCEDURE.html,
...)

LorrensP-2158466:
Potentially closes #1282 

This introduces a new statement: `DropProcedure`. It has the same fields as `DropFunction` because it also has the same syntax.

The other way i could have implemented this was using the `Drop` statement and introducing a new `ObjectType` called `ObjectType::Procedure`, but because `DROP PROCEDURE ...` is the same (just different word) as `DROP FUNCTION...`, I thought this was weird to do.

I also added a test case in the same way as `DropFunction`.

Let me know what you guys think of this!

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2210,7 +2210,7 @@ pub enum Statement {
     DropFunction {
         if_exists: bool,
         /// One or more function to drop
-        func_desc: Vec<DropFuncDesc>,
+        func_desc: Vec<DropFunctionDesc>,
         /// `CASCADE` or `RESTRICT`
         option: Option<ReferentialAction>,
     },
@@ -2220,7 +2220,7 @@ pub enum Statement {
     DropProcedure {
         if_exists: bool,
         /// One or more function to drop
-        proc_desc: Vec<DropFuncDesc>,
+        proc_desc: Vec<DropFunctionDesc>,
         /// `CASCADE` or `RESTRICT`
         option: Option<ReferentialAction>,
     },
@@ -5910,12 +5910,12 @@ impl fmt::Display for DropFunctionOption {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub struct DropFuncDesc {
+pub struct DropFunctionDesc {
     pub name: ObjectName,
     pub args: Option<Vec<OperateFunctionArg>>,
 }
 
-impl fmt::Display for DropFuncDesc {
+impl fmt::Display for DropFunctionDesc {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self.name)?;
         if let Some(args) = &self.args {
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2193,7 +2193,17 @@ pub enum Statement {
     DropFunction {
         if_exists: bool,
         /// One or more function to drop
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -5867,12 +5893,12 @@ impl fmt::Display for DropFunctionOption {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub struct DropFunctionDesc {
+pub struct DropFuncDesc {
```
iffyio, line 5896: was it a specific reason to rename `DropFunctionDesc` to `DropFuncDesc`? if not we can probably undo the change given its a breaking change and the benefit is arguable, (and we have other similarly named objects like `DropFunctionOption`, `parse_drop_function_desc`)

src/parser/mod.rs
```diff
@@ -4529,7 +4531,26 @@ impl<'a> Parser<'a> {
...
+    /// DROP PROCEDURE [ IF EXISTS ] name [ ( [ [ argmode ] [ argname ] argtype [, ...] ] ) ] [, ...]
+    /// [ CASCADE | RESTRICT ]
+    /// ```
+    fn parse_drop_procedure(&mut self) -> Result<Statement, ParserError> {
+        let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
+        let proc_desc = self.parse_comma_separated(Parser::parse_drop_function_desc)?;
+        let option = match self.parse_one_of_keywords(&[Keyword::CASCADE, Keyword::RESTRICT]) {
+            Some(Keyword::CASCADE) => Some(ReferentialAction::Cascade),
+            Some(Keyword::RESTRICT) => Some(ReferentialAction::Restrict),
+            _ => None,
```
iffyio, line 4544:
```suggestion
            Some(invalid) => return parser_err!(...)
            None => None,
```
Should this rather return an error if the action is unexpected?
seems otherwise e.g. `DROP PROCEDURE testproc SET NULL` will be accepted by the parser but it will silently discard the `SET NULL` part (we can also include a test case demonstrating the behavior)

src/ast/mod.rs
LorrensP-2158466: Ah yeah you're right. I playing around with some new names since DropProcedure uses it. I'll change it back.

alamb: done in bb0dad8

src/parser/mod.rs
alamb: I think they way that parse_one_of_keywords works, it will only return CASCADE or RESTRICT here (never Some(..)). 

I added a test in e782bf5 and changed the code to try and make this clearer

alamb (approved):

alamb:
Thank you @LorrensP-2158466  and @iffyio

LorrensP-2158466:
Thanks for doing those final changes! It's very busy at the moment, sorry.

alamb:
> Thanks for doing those final changes! It's very busy at the moment, sorry.

No problems! The best software in my opinion is a good team effort. 🚀

PR #1327, opened 2024-06-30
Support `SETTINGS` pairs for ClickHouse dialect

git-hulk:
SETTINGS in query is supported by ClickHouse, for example:

```
SELECT * FROM t SETTINGS max_threads = 1, max_block_size = 10000
```

For more information, please refer to:

https://clickhouse.com/docs/en/sql-reference/statements/select#settings-in-select-query

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7933,13 +7933,11 @@ impl<'a> Parser<'a> {
             let settings = if dialect_of!(self is ClickHouseDialect|GenericDialect)
                 && self.parse_keyword(Keyword::SETTINGS)
             {
-                let mut key_values: Vec<Setting> = vec![];
-                self.parse_comma_separated(|p| {
+                let key_values = self.parse_comma_separated(|p| {
                     let key = p.parse_identifier(false)?;
                     p.expect_token(&Token::Eq)?;
                     let value = p.parse_value()?;
-                    key_values.push(Setting { key, value });
-                    Ok(())
+                    Ok(Setting { key, value })
                 })?;
                 Some(key_values)
             } else {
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -7928,6 +7930,24 @@ impl<'a> Parser<'a> {
...
+                loop {
+                    let key = self.parse_identifier(false)?;
+                    self.expect_token(&Token::Eq)?;
+                    let value = self.parse_value()?;
+                    key_values.push(Setting { key, value });
+                    if !self.consume_token(&Token::Comma) {
+                        break;
+                    }
+                }
+                Some(key_values)
```
iffyio, line 7946: I think we might be able to reuse the [parse_comma_separated](https://github.com/sqlparser-rs/sqlparser-rs/blob/cdba898cfe6e0aed073655f0d3a56e38eed4eb75/src/parser/mod.rs#L3338) functionality here, could you take a look at that function to see if it works for this case?

git-hulk: @iffyio Good point, it's done now.

src/parser/mod.rs
```diff
@@ -7928,6 +7930,22 @@ impl<'a> Parser<'a> {
...
+            {
+                let mut key_values: Vec<Setting> = vec![];
+                self.parse_comma_separated(|p| {
+                    let key = p.parse_identifier(false)?;
+                    p.expect_token(&Token::Eq)?;
+                    let value = p.parse_value()?;
+                    key_values.push(Setting { key, value });
+                    Ok(())
+                })?;
+                Some(key_values)
```
iffyio, line 7944:
```suggestion
                let key_values = self.parse_comma_separated(|p| {
                    let key = p.parse_identifier(false)?;
                    p.expect_token(&Token::Eq)?;
                    let value = p.parse_value()?;
                    Ok(Setting { key, value })
                })?;
                Some(key_values)
```
Ah so I think no need to manually build up the list since parse_comma_separated already returns the vec, something like above example?

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @git-hulk  and @iffyio  for the review

PR #1330, opened 2024-07-03
[ClickHouse] Add support for WITH FILL to OrderByExpr

nickpresta:
ClickHouse supports the `ORDER BY ... WITH FILL` modifier:

https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier

`WITH FILL` itself supports a simple "from", "to", and "step" parameters, and a more sophisticated `INTERPOLATE` option.


tests/sqlparser_common.rs
```diff
@@ -2103,18 +2108,114 @@ fn parse_select_order_by_nulls_order() {
...
                 nulls_first: Some(false),
+                with_fill: None,
             },
         ],
         select.order_by
     );
     assert_eq!(Some(Expr::Value(number("2"))), select.limit);
 }
 
+#[test]
```
iffyio, line 2125: given this is clickhouse specific should these tests rather live in `sqlparser_clickhouse.rs`?

tests/sqlparser_common.rs
```diff
@@ -2103,18 +2108,114 @@ fn parse_select_order_by_nulls_order() {
...
+fn parse_select_order_by_with_fill() {
```
iffyio, line 2126: Can we add/extend test cases demonstrating behavior for the following cases?

- multiple interpolate clauses e.g `order by a with fill from 10 interpolate (a as b) interpolate (c as d)`
- interpolate inbetween with clauses e.g. `order by a with fill interpolate (a as b), b with fill interpolate (c as d)`
- interpolate without body e.g. `order by a with fill interpolate`
- interpolate with empty body e.g. `order by a with fill interpolate ()`
- interpolate without with fill e.g. `order by a interpolate (a as b)`
- incomplete with fills e.g. `order by a with fill from to 20`, `order by a with fill to step 2` etc

src/parser/mod.rs
```diff
@@ -10408,13 +10408,75 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let with_fill = if self.parse_keywords(&[Keyword::WITH, Keyword::FILL]) {
+            Some(self.parse_with_fill()?)
+        } else {
+            None
+        };
```
iffyio, line 10415: We can guard this behavior with the `dialect_of!(self is ClickhouseDialect | GenericDialect)` so that it only applies to clickhouse and the generic dialects

src/ast/query.rs
```diff
@@ -1642,6 +1645,62 @@ impl fmt::Display for OrderByExpr {
...
+
+/// ClickHouse `WITH FILL` modifier for `ORDER BY` clause.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct WithFill {
+    pub from: Option<Expr>,
+    pub to: Option<Expr>,
+    pub step: Option<Expr>,
+    pub interpolate: Vec<Interpolation>,
```
iffyio, line 1663: I wonder if this part is correct? from their docs they say the following
> INTERPOLATE can be applied to columns not participating in ORDER BY WITH FILL

similarly looking at the grammar in the doc, interpolate seems to be a single clause applied to the order by expression rather than to the individual with fills?

src/ast/query.rs
```diff
@@ -1642,6 +1645,62 @@ impl fmt::Display for OrderByExpr {
...
             None => (),
         }
+        if let Some(ref with_fill) = self.with_fill {
+            write!(f, " {}", with_fill)?
+        }
+        Ok(())
+    }
+}
+
+/// ClickHouse `WITH FILL` modifier for `ORDER BY` clause.
```
iffyio, line 1655: Can we include the link to the doc here as well?

src/ast/query.rs
nickpresta: Using the example from the [docs](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier), it isn't valid to say `ORDER BY n INTERPOLATE (inter AS inter + 1)`:

```
SELECT n, source, inter FROM (
   SELECT toFloat32(number % 10) AS n, 'original' AS source, number as inter
   FROM numbers(10) WHERE number % 3 = 1
) ORDER BY n INTERPOLATE (inter AS inter + 1);

Syntax error: failed at position 163 ('INTERPOLATE') (line 4, col 14):

SELECT n, source, inter FROM (
   SELECT toFloat32(number % 10) AS n, 'original' AS source, number as inter
   FROM numbers(10) WHERE number % 3 = 1
) ORDER BY n INTERPOLATE (inter AS inter + 1);

Expected one of: token, Dot, OR, AND, IS NOT DISTINCT FROM, IS NULL, IS NOT NULL, BETWEEN, NOT BETWEEN, LIKE, ILIKE, NOT LIKE, NOT ILIKE, REGEXP, IN, NOT IN, GLOBAL IN, GLOBAL NOT IN, MOD, DIV, alias, AS, DESCENDING, DESC, ASCENDING, ASC, NULLS, COLLATE, WITH FILL, Comma, LIMIT, OFFSET, FETCH, SETTINGS, UNION, EXCEPT, INTERSECT, INTO OUTFILE, FORMAT, end of query
```

The `INTERPOLATE` must be associated with the `WITH FILL` modifier:

```
SELECT
    n,
    source,
    inter
FROM
(
    SELECT
        toFloat32(number % 10) AS n,
        'original' AS source,
        number AS inter
    FROM numbers(10)
    WHERE (number % 3) = 1
)
ORDER BY n ASC WITH FILL
INTERPOLATE ( inter AS inter + 1 )

Query id: f8146c67-565f-4e19-8fe8-5dcfd9531add

   ┌─n─┬─source───┬─inter─┐
1. │ 1 │ original │     1 │
2. │ 2 │          │     2 │
3. │ 3 │          │     3 │
4. │ 4 │ original │     4 │
5. │ 5 │          │     5 │
6. │ 6 │          │     6 │
7. │ 7 │ original │     7 │
   └───┴──────────┴───────┘

7 rows in set. Elapsed: 0.001 sec.
```

The `FROM`, `TO`, and `STEP` options can be left out, but there needs to be a `WITH FILL`


Example playground available here: https://fiddle.clickhouse.com/96df3a22-5fa1-4c0e-8be6-36d963373425

tests/sqlparser_common.rs
nickpresta: > multiple interpolate clauses e.g `order by a with fill from 10 interpolate (a as b) interpolate (c as d)`

This isn't supported syntax. It would be `multiple interpolate clauses e.g order by a with fill from 10 interpolate (a as b, c as d)` ([playground](https://fiddle.clickhouse.com/07af6393-3204-4826-be29-467ab2bcbab4)).

Do you want a test to show this syntax fails?

> interpolate inbetween with clauses e.g. `order by a with fill interpolate (a as b), b with fill interpolate (c as d)`

This syntax isn't supported (the multiple `INTERPOLATE`).

Do you want a test to show this syntax fails?

> interpolate without body e.g. `order by a with fill interpolate`

Done!

> interpolate with empty body e.g. `order by a with fill interpolate ()`

Done!

> interpolate without with fill e.g. `order by a interpolate (a as b)`

As explained https://github.com/sqlparser-rs/sqlparser-rs/pull/1330#discussion_r1666736373, this isn't supported.

Do you want a test for it?

> incomplete with fills e.g. `order by a with fill from to 20, order by a with fill to step 2 etc`

Not entirely sure what this test would look like - each of the options (e.g. `from`, `to`, `step`) require an expression.

src/ast/query.rs
nickpresta: I've updated the structure for `INTERPOLATE` in [7022e42](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/7022e4227fe00ae4914688ad91e355effd8a159b) to match.

Thanks for catching this

nickpresta:
One question I had around the naming - do we like `InterpolationArg` and `Interpolation` or would we prefer this to be `InterpolateArg` and `Interpolate` to match the keyword and field name?

The values describe and "interpolation" which is why I named it that way to begin with, but I don't feel strongly either way.

tests/sqlparser_common.rs
iffyio: Ah yeah so for the invalid ones the idea is that they become tests that show the parser fails, heres one for example https://github.com/sqlparser-rs/sqlparser-rs/blob/7022e4227fe00ae4914688ad91e355effd8a159b/tests/sqlparser_clickhouse.rs#L485-L491

src/ast/query.rs
```diff
@@ -1627,6 +1627,14 @@ pub struct OrderByExpr {
...
     /// Optional `NULLS FIRST` or `NULLS LAST`
     pub nulls_first: Option<bool>,
+    /// Optional: `WITH FILL`
+    /// Supported by [ClickHouse syntax]: <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier>
+    pub with_fill: Option<WithFill>,
+    /// Optional: `INTERPOLATE`
+    /// Supported by [ClickHouse syntax]
+    ///
+    /// [ClickHouse syntax]: <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier>
+    pub interpolate: Option<InterpolationArg>,
```
iffyio, line 1637:
```suggestion
    pub interpolate: Option<Interpolation>,
```
Yeah regarding the naming, I think it would be reasonable in this case to have name the struct `Interpolate` as well so we get `pub interpolate: Interpolate`. We could have something like

```
struct Interpolate {
    exprs: Option<Vec<ExprWithAlias>>
}
```
thinking e.g.
```
INTERPOLATE => Interpolate { exprs: None}
```
```
INTERPOLATE () => Interpolate { exprs: Some(vec![]) }
```
[ExprWithAlias can be reused from here](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/query.rs#L831-L843)

src/parser/mod.rs
```diff
@@ -10408,13 +10408,86 @@ impl<'a> Parser<'a> {
...
+
+        let interpolate = if dialect_of!(self is ClickHouseDialect | GenericDialect)
+            && self.parse_keyword(Keyword::INTERPOLATE)
+        {
+            if self.consume_token(&Token::LParen) {
+                if self.peek_token().token == Token::RParen {
```
iffyio, line 10423:
```suggestion
                if self.consume_token(&Token::RParen) {
```
this could do similar to lparen?

iffyio, line 10423: Oh actually to simplify the nested if/else, the interpolation parsing can use, `parse_comma_separated0` function instead, that one allows an empy list of arguments, seems like that would let us avoid the extra nested parens checks here?

src/parser/mod.rs
```diff
@@ -10408,13 +10408,86 @@ impl<'a> Parser<'a> {
...
+    }
+
+    // Parse a INTERPOLATE expression (ClickHouse dialect)
+    pub fn parse_interpolation(&mut self) -> Result<Interpolation, ParserError> {
+        let column = self.parse_expr()?;
+        let formula = if self.parse_keyword(Keyword::AS) {
+            Some(self.parse_expr()?)
+        } else {
+            None
+        };
```
iffyio, line 10487: by reusing ExprWithAlias we'd likely be able to reuse the parsing function for that as well here https://github.com/sqlparser-rs/sqlparser-rs/blob/7022e4227fe00ae4914688ad91e355effd8a159b/src/parser/mod.rs#L9527-L9536

src/ast/query.rs
nickpresta: Just to be explicit since our comments overlapped a bit:


> Ah I see, I think that makes sense! So it seems to only be a representation issue - interpolate can't be specified without `With Fill` but it still part of the order_by clause rather than a single `OrderByExpr` right or I might be misunderstanding still?
> 
> my understanding is that:
> 
> * its not valid to have multiple `INTERPOLATE` in an order by expression? e.g. this is invalid
> 
> ```sql
> ORDER BY a WITH FILL INTERPOLATE, b WITH FILL INTERPOLATE ...
> ```

That is correct. There can only be a single `INTERPOLATE` at the end of an `ORDER BY`, and multiple `WITH FILL`s can be added. Having at least a single `WITH FILL` is required for `INTERPOLATE` to work at all.

> * Then if I pass in the following to the parser:
> 
> ```sql
> ORDER BY a WITH FILL, b WITH FILL, c WITH FILL INTERPOLATE
> ```
> 
> * I get back a list of order by expressions
> 
> ```
> vec![
> 	OrderByExpr{a, WithFill{interpolate}},
> 	OrderByExpr{b, WithFill{interpolate}},
> 	OrderByExpr{c, WithFill{interpolate}},
> ]
> ```
> 
> where potentially only one of those entries in the list can actually contain _the_ interpolate.
> 
> does this describe the current behavior?


I added a test called `parse_select_order_by_with_fill_interpolate` which has:

```sql
ORDER BY
    fname ASC NULLS FIRST WITH FILL FROM 10 TO 20 STEP 2
    lname DESC NULLS LAST WITH FILL FROM 30 TO 40 STEP 3
    INTERPOLATE (col1 AS col1 + 1)
```

and parsing that will give back:

```rust
vec![
    OrderByExpr {
        expr: Expr::Identifier(Ident::new("fname")),
        asc: Some(true),
        nulls_first: Some(true),
        with_fill: Some(WithFill {
            from: Some(Expr::Value(number("10"))),
            to: Some(Expr::Value(number("20"))),
            step: Some(Expr::Value(number("2"))),
        }),
        interpolate: None,
    },
    OrderByExpr {
        expr: Expr::Identifier(Ident::new("lname")),
        asc: Some(false),
        nulls_first: Some(false),
        with_fill: Some(WithFill {
            from: Some(Expr::Value(number("30"))),
            to: Some(Expr::Value(number("40"))),
            step: Some(Expr::Value(number("3"))),
        }),
        interpolate: Some(InterpolationArg::Columns(vec![Interpolation {
            column: Expr::Identifier(Ident::new("col1")),
            formula: Some(Expr::BinaryOp {
                left: Box::new(Expr::Identifier(Ident::new("col1"))),
                op: BinaryOperator::Plus,
                right: Box::new(Expr::Value(number("1"))),
            }),
        }]))
    },
]
```

src/parser/mod.rs
nickpresta: Will swap to use `parse_comma_separated0` so we don't need this at all.

src/ast/query.rs
nickpresta: I'll update the naming to `Interpolate` instead of `Interpolation`.

I don't think this can use `ExprWithAlias` since the `AS x` part is actually an expression, not an alias. For [example](https://fiddle.clickhouse.com/846d0c38-136e-4743-8256-0d2fe6297123), this is valid:

```sql
INTERPOLATE(thing AS toDecimal64(log2(thing) * 2 + 1, 3))
```

So I think we need this to be an `Expr`. After doing more digging, it looks like the initial column can only be an `Expr::Identifier(Ident)` so updating the `Interpolate.column` to be `Ident` makes the most sense (to me).

nickpresta: I pushed [47ba442](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/47ba442bdf6e39a5f7ae3626ca6f34ffb8986424) that swaps `column` to an `Ident` and updates the naming, etc.

src/parser/mod.rs
nickpresta: I don't think we can use `ExprWithAlias` based on https://github.com/sqlparser-rs/sqlparser-rs/pull/1330#discussion_r1667012551

tests/sqlparser_common.rs
nickpresta: Sounds good. I'll add a few of those now.

nickpresta: Added tests for the failure case in [ebcd31e](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/ebcd31e6b1ffa7fd282f179f573215036a04b57e):

- ✅ multiple interpolate clauses e.g `order by a with fill from 10 interpolate (a as b) interpolate (c as d)`
- ✅ interpolate inbetween with clauses e.g. `order by a with fill interpolate (a as b), b with fill interpolate (c as d)`
- ✅ interpolate without with fill e.g. `order by a interpolate (a as b)`
- ✅ incomplete with fills e.g. `order by a with fill from to 20, order by a with fill to step 2 etc`

nickpresta:
I added some tests for validation that invalid syntax is rejected. This required me to add a validation function specifically for `GenericDialect` and `ClickHouseDialect` to check that we have `WITH FILL` when we have an `INTERPOLATE` and ensuring that `INTERPOLATE` is in the last `OrderByExpr`

The only other way I could think to tackle this type of validation would be to change the `Query.order_by` to not be a `Vec<OrderByExpr>` but something like:

```rust
struct OrderByClause {
  order_bys: Vec<OrderByExpr>,
  interpolate: Option<InterpolateArg>,
}
```

but that seems much more disruptive to everyone who uses `Query.order_by` vs keeping this localized to just the the ClickHouse parsing.

src/ast/query.rs
```diff
@@ -1642,6 +1650,80 @@ impl fmt::Display for OrderByExpr {
...
+/// ClickHouse `INTERPOLATE` clause for use in `ORDER BY` clause when using `WITH FILL` modifier.
+/// Supported by [ClickHouse syntax]
+///
+/// [ClickHouse syntax]: <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier>
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct Interpolate {
+    pub column: Ident,
+    pub formula: Option<Expr>,
```
iffyio, line 1710:
```suggestion
    pub expr: Option<Expr>,
```
wondering can this be called expr or is formula the actual term for it? (if that's the term then sounds good as is)

src/ast/query.rs
```diff
@@ -1642,6 +1650,80 @@ impl fmt::Display for OrderByExpr {
...
+}
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum InterpolateArg {
+    NoBody,
+    Columns(Vec<Interpolate>),
+}
```
iffyio, line 1719: instead of having two variant enum can we use Option to represent this within the interpolate struct? e.g.
```rust
struct Interpolate {
    expr: Option<Vec<InterpolateExpr>>
}
struct InterpolateExpr {
      column,
      formular
}
```
    

```

File: src/parser/mod.rs [hunk 11]
```diff
@@ -7888,7 +7888,9 @@ impl<'a> Parser<'a> {
             let body = self.parse_boxed_query_body(0)?;
 
             let order_by = if self.parse_keywords(&[Keyword::ORDER, Keyword::BY]) {
-                self.parse_comma_separated(Parser::parse_order_by_expr)?
+                let order_by_exprs = self.parse_comma_separated(Parser::parse_order_by_expr)?;
+                self.validate_order_by_exprs_for_interpolate_and_with_fill(&order_by_exprs)?;
```
iffyio, line 7892: yeah so the parser generally doesn't do semantic validations explicitly except for maybe trivial cases, in this case it seems correct to have the invalid state non-representable in the AST, it does become a bit invasive of a change but that should be fine since it would leave us with a more accurate representation of the syntax

src/ast/query.rs
iffyio: Ah that makes sense!

src/ast/query.rs
nickpresta: Will rename - the ClickHouse docs call it `exprN` (N number of expressions) and that seems consistent with other structures 👍

src/ast/query.rs
nickpresta: Updated in [ec263d2](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/ec263d2b82f596b2f57b187ace865a707c345c92)

nickpresta:
@alamb Thanks for that information. I believe I've addressed the latest feedback from @iffyio (rename + refactor structure). 

Let me know if there's anything else I need to do (e.g. changelog updates).

src/ast/query.rs
```diff
@@ -1671,6 +1679,81 @@ impl fmt::Display for OrderByExpr {
...
             None => (),
         }
+        if let Some(ref with_fill) = self.with_fill {
+            write!(f, " {}", with_fill)?
+        }
+        if let Some(ref interpolate) = self.interpolate {
+            match &interpolate.expr {
+                Some(exprs) => write!(
+                    f,
+                    " INTERPOLATE ({})",
```
git-hulk, line 1689: Is it better to use `display_comma_separated` here?

nickpresta: Yep. Good call. I forgot this existed. Thank you. Fixed in [3e8b141](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/3e8b141b7a19f8c16c535c62c17e61a3400e2dd1)

git-hulk (approved):
LGTM

src/ast/query.rs
```diff
@@ -62,8 +62,18 @@ impl fmt::Display for Query {
...
         write!(f, "{}", self.body)?;
-        if !self.order_by.is_empty() {
-            write!(f, " ORDER BY {}", display_comma_separated(&self.order_by))?;
+        if !self.order_by.exprs.is_empty() {
+            write!(
+                f,
+                " ORDER BY {}",
+                display_comma_separated(&self.order_by.exprs)
+            )?;
+            if let Some(ref interpolate) = self.order_by.interpolate {
```
nickpresta, line 71: I was conflicted about this - we're only going to write out `INTERPOLATE` if there are expressions, which makes sense (e.g. you can't have a free-standing `INTERPOLATE (...)` since it requires an `ORDER BY` but I suppose this _could_ be confusing since you can create:

```rust
OrderBy {
  exprs: vec![],
  interpolate: Some(Interpolate { ... }),
}
```

but you won't see `INTERPOLATE` written.

I don't know why anyone would do this as it's never valid so it seems OK since it means we will still have a valid query, instead of the alternative, which will always produce an invalid query.

iffyio (commented):
Left a couple comments, otherwise I think this looks good!

src/ast/query.rs
nickpresta: With this change, it would be possible to craft:

```rust
Some(OrderBy {
    exprs: vec![],
    interpolate: Interpolate { ... }
})
```

which would display:

```sql
ORDER BY INTERPOLATE
```

which gets interpreted as `INTERPOLATE` being a field name.

I'm not entirely sure that's preferable - it feels like every dialect is going to require _some_ identifier and then anything else added on. I'm not sure how much we can future-proof this for dialects that would accept an `ORDER BY` without identifiers but other keywords after. Maybe this falls under "user error" since it's possible to craft a bunch of scenarios that aren't actually valid and we entrust developers to understand how to use each AST node correctly.

---

I've pushed an update with this change in [44deb69](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/44deb69db2ddb4c9448e9417f7e98fcd3056aa90), but can always revert it if we think it's problematic.

src/parser/mod.rs
```diff
@@ -10511,13 +10517,79 @@ impl<'a> Parser<'a> {
...
+            None
+        };
+
+        Ok(WithFill { from, to, step })
+    }
+
+    // Parse a set of comma seperated INTERPOLATE expressions (ClickHouse dialect)
+    // that follow the INTERPOLATE keyword in an ORDER BY clause with the WITH FILL modifier
+    pub fn parse_interpolations(&mut self) -> Result<Option<Interpolate>, ParserError> {
+        if dialect_of!(self is ClickHouseDialect | GenericDialect)
```
iffyio, line 10563: We can move this dialect check outside the function and to the caller, and have the function only care about the parsing logic.

one minor, we can also avoid nesting the function body by returning early
```rust
if !self.parse_keyword(Keyword::INTERPOLATE) {
     return Ok(None)
}

if self.consume_token(...)
Ok(Some(...))
```

src/ast/query.rs
iffyio: I imainge that's fine as long as the parser itself doesn't wrongly accept SQL that builds up an invalid struct. in that, a user on the other hand can manually build up that struct however they see fit.

iffyio (approved):
One minor comment re pulling out the dialect check, LGTM! cc @alamb

src/parser/mod.rs
nickpresta: Done in [000e08d](https://github.com/sqlparser-rs/sqlparser-rs/pull/1330/commits/000e08d3f0849b93b70f056614b51539f3adebfe). Thanks!

nickpresta:
Hey @alamb - is this something that can be merged into `main`? I understand this change won't make it into a version until `0.49.0` is released (I hope 🤞), but I could at least target a commit on `main` instead of this branch.

Thank you!

alamb:
Sure thing -- sorry @nickpresta. I am sorry for the delay. I pushed a commit to try and resolve a test failure and hope to merge this in once the CI completes

alamb (approved):

nickpresta:
No problem at all! I appreciate everyone's effort to get this merged in.

Thank you @alamb, @iffyio, and @git-hulk for your time and all the feedback.

PR #1337, opened 2024-07-09
Add support of table function WITH ORDINALITY modifier for Postgre Parser

### LINKED ISSUE #1336: add support for `WITH ORDINALITY`
The following is valid sql, and works in postgres, but cannot be parsed with sqlparser:

```sql
select * from my_srf() with ordinality;
```

https://www.postgresql.org/docs/current/functions-srf.html

Initially reported in https://github.com/lovasoa/SQLpage/discussions/491#discussioncomment-9986790

git-hulk:
This closes #1336


git-hulk:
@lovasoa Can help to review when you get time, thank you!

src/parser/mod.rs
```diff
@@ -9216,6 +9216,18 @@ impl<'a> Parser<'a> {
                 None
             };
 
+            let mut with_ordinality = false;
+            if dialect_of!(self is PostgreSqlDialect|GenericDialect)
```
lovasoa, line 9220: There is a lot of postgres-derived dialects. Hardcoding supported dialects like that limits the usability. I would either add a method in the dialect trait, or simply always parse the construct, no matter the dialect

lovasoa (changes requested):
WITH ORDINALITY is also supported after UNNEST, which I think needs special handling in sqlparser

tests/sqlparser_postgres.rs
```diff
@@ -4183,3 +4186,40 @@ fn parse_create_table_with_options() {
         _ => unreachable!(),
     }
 }
+
+#[test]
+fn test_table_function_with_ordinality() {
+    let sql = "SELECT * FROM generate_series(1, 10) WITH ORDINALITY AS t";
```
lovasoa, line 4192: I think you can use verify_only_select here.

git-hulk:
> WITH ORDINALITY is also supported after UNNEST, which I think needs special handling in sqlparser

Yes, good catch. Done in commit: https://github.com/sqlparser-rs/sqlparser-rs/pull/1337/commits/a3cb42a3536141ef4e9c2a334a63c8aa39b8fabe.

git-hulk:
cc @alamb

PR #1338, opened 2024-07-10
Enable PARTITION BY feature for PostgreSQL while parsing the create table statement

### LINKED ISSUE #1281: Partition by not supported - postgre
Have this query:
```
CREATE TABLE measurement (
    city_id         int not null,
    logdate         date not null,
    peaktemp        int,
    unitsales       int
) PARTITION BY RANGE (logdate);
```
Got this error:
`sql parser error: Expected end of statement, found: PARTITION at Line: 7, Column 7
`
Doc:
https://www.postgresql.org/docs/current/ddl-partitioning.html

git-hulk:
This closes #1281.

PARTITION BY has been supported but wasn't enabled for the PostgreSQL dialect, this PR only enables it and adds a test case to make sure it works well.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -31,7 +31,7 @@ use recursion::RecursionCounter;
 use IsLateral::*;
 use IsOptional::*;
 
-use crate::ast::helpers::stmt_create_table::{BigQueryTableConfiguration, CreateTableBuilder};
+use crate::ast::helpers::stmt_create_table::{CreateTableBuilder, CreateTableConfiguration};
 use crate::ast::*;
 use crate::dialect::*;
 use crate::keywords::{Keyword, ALL_KEYWORDS};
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -5416,7 +5416,8 @@ impl<'a> Parser<'a> {
             None
         };
 
-        let big_query_config = if dialect_of!(self is BigQueryDialect | GenericDialect) {
+        let big_query_config = if dialect_of!(self is BigQueryDialect | PostgreSqlDialect | GenericDialect)
```
iffyio, line 5419: the `parse_optional_big_query_create_table_config()` function becomes a bit misleading with this change since its no longer bigquery specific (and would mean that we'd be supporting bigquery features for postgres which isn't necessariy desirable).
We can probably still reuse that function but make it dialect agnostic, e.g if we rename it to `parse_optional_create_table_config` and move the dialect_of checks into the function so that partitionby is guarded on both bigquery and postgres, while the rest are guarded on bigquery?

git-hulk: @iffyio Good point. To be honest, I didn't notice this naming issue while adding the new dialect. I will rename and move the dialect_of guarded inside the function.

git-hulk: @iffyio Done in the last commit, please take a look again when you get time ❤️

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @git-hulk and @iffyio

Issue #1274, opened 2024-05-17
Support postgresql on commit for ctas


fdzuJ:
Add support for postgresql `[ ON COMMIT { PRESERVE ROWS | DELETE ROWS | DROP } ]`. Currently panics:

```
use sqlparser::dialect::PostgreSqlDialect;
use sqlparser::parser::Parser;
 
fn main() {
    let sql = "create local temporary table test on commit preserve rows as select 1 as a;";
 
    let dialect = PostgreSqlDialect {};
 
    let ast = Parser::parse_sql(&dialect, sql).unwrap();
 
    for stmt in ast.iter() {
        println!("{}", stmt.to_string())
    }
}
```

`ParserError("Expected end of statement, found: as at Line: 1, Column 59")`

git-hulk:
Hi @fdzuJ 

I cannot find where mentioned the syntax that supports the alias after ON COMMIT clause, seems it only supports [TABLESPACE tablespace_name] according to Postgre documentation: https://www.postgresql.org/docs/current/sql-createtable.html.

It will be helpful if you can provide the reference and I will be happy to support this feature if it did miss in sqlparse-rs.

fdzuJ:
Hi @git-hulk!
Here is documentation for this syntax.

https://www.postgresql.org/docs/current/sql-createtableas.html

git-hulk:
@fdzuJ Thank you! I will submit a PR to resolve this.

git-hulk:
@fdzuJ Would you mind having a try at the latest version, it should be resolved now. cc @alamb

LucaCappelletti94:
Closing as solved in https://github.com/apache/datafusion-sqlparser-rs/pull/1339

PR #1339, opened 2024-07-12
Fix AS query clause should be after the create table options

git-hulk:
According to MySQL[1] and PostgreSQL[2] create table syntax, the AS query should be sit after the table options. But it is parsed before table options include CHARSET and COLLATE, as well as ON COMMIT clause. This PR resolves this issue by moving the position of AS query, and the issue #1274 will be aslo resolved after this PR.

MySQL:

```SQL
CREATE [TEMPORARY] TABLE [IF NOT EXISTS] tbl_name
    [(create_definition,...)]
    [table_options]
    [partition_options]
    [IGNORE | REPLACE]
    [AS] query_expression
```

PostgreSQL:

```SQL
CREATE [ [ GLOBAL | LOCAL ] { TEMPORARY | TEMP } | UNLOGGED ] TABLE [ IF NOT EXISTS ] table_name
    [ (column_name [, ...] ) ]
    [ USING method ]
    [ WITH ( storage_parameter [= value] [, ... ] ) | WITHOUT OIDS ]
    [ ON COMMIT { PRESERVE ROWS | DELETE ROWS | DROP } ]
    [ TABLESPACE tablespace_name ]
    AS query
    [ WITH [ NO ] DATA ]
```

[1] https://dev.mysql.com/doc/refman/8.0/en/create-table.html
[2] https://www.postgresql.org/docs/current/sql-createtableas.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5422,13 +5422,6 @@ impl<'a> Parser<'a> {
             Default::default()
         };
 
-        // Parse optional `AS ( query )`
-        let query = if self.parse_keyword(Keyword::AS) {
-            Some(self.parse_boxed_query()?)
-        } else {
-            None
-        };
-
         let default_charset = if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARSET]) {
             self.expect_token(&Token::Eq)?;
             let next_token = self.next_token();
@@ -5469,6 +5462,13 @@ impl<'a> Parser<'a> {
             };
 
         let strict = self.parse_keyword(Keyword::STRICT);
+
+        // Parse optional `AS ( query )`
+        let query = if self.parse_keyword(Keyword::AS) {
+            Some(self.parse_boxed_query()?)
+        } else {
+            None
+        };
 
         let comment = if self.parse_keyword(Keyword::COMMENT) {
             let _ = self.consume_token(&Token::Eq);
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5463,13 +5463,6 @@ impl<'a> Parser<'a> {
 
         let strict = self.parse_keyword(Keyword::STRICT);
 
-        // Parse optional `AS ( query )`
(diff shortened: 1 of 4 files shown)


alamb (approved):
Makes sense to me -- thank you @git-hulk 

cc @iffyio

src/parser/mod.rs
```diff
@@ -5470,6 +5463,13 @@ impl<'a> Parser<'a> {
 
         let strict = self.parse_keyword(Keyword::STRICT);
 
+        // Parse optional `AS ( query )`
+        let query = if self.parse_keyword(Keyword::AS) {
+            Some(self.parse_boxed_query()?)
+        } else {
+            None
+        };
+
```
iffyio, line 5472: Should the block come after the `COMMENT` is parsed? (not sure re which dialects use the comment syntax  and I'm assuming the `AS` is usually the last word in the create table).

quick look at the display function for create table seemed like we print the comment much earlier in the statement which might be weird (unrelated to this change I imagine though, just noticed while looking there)

git-hulk: @iffyio Yes, good call. The comment syntax was introduced in PR https://github.com/sqlparser-rs/sqlparser-rs/pull/1059 for MySQL. And after looking into the MySQL syntax, it's included in the table option. So you're right that it should parsed after the AS query.

```MySQL
table_option: {
    AUTOEXTEND_SIZE [=] value
  | AUTO_INCREMENT [=] value
  | AVG_ROW_LENGTH [=] value
  | [DEFAULT] CHARACTER SET [=] charset_name
  | CHECKSUM [=] {0 | 1}
  | [DEFAULT] COLLATE [=] collation_name
  | COMMENT [=] 'string']
 ...
```

refer to: https://dev.mysql.com/doc/refman/8.4/en/create-table.html

iffyio (approved):
LGTM!

alamb:
Looks good to me -- thank you @git-hulk and @iffyio

PR #1341, opened 2024-07-14
Snowflake: support position with normal function call syntax

jmhain:
On Snowflake, the special `IN` syntax is optional: https://docs.snowflake.com/en/sql-reference/functions/position. This PR updates the parser to accept this.


tests/sqlparser_snowflake.rs
```diff
@@ -2256,3 +2256,9 @@ fn asof_joins() {
           "ORDER BY s.observed",
     ));
 }
+
+#[test]
+fn test() {
```
iffyio, line 2261:
```suggestion
fn test_parse_position() {
```
or something similar?

tests/sqlparser_snowflake.rs
```diff
@@ -2256,3 +2256,9 @@ fn asof_joins() {
...
+    snowflake().verified_query("SELECT position('an', 'banana', 1)");
+    snowflake().verified_query("SELECT n, h, POSITION(n IN h) FROM pos");
```
iffyio, line 2263: We can probably include an assertion demonstrating that the position parse to different expression types?

Also would it make more sense to have this in common given the code isn't guarded by snowflake dialect?

tests/sqlparser_common.rs
```diff
@@ -8202,17 +8202,10 @@ fn parse_position() {
 
 #[test]
 fn parse_position_negative() {
-    let sql = "SELECT POSITION(foo) from bar";
```
git-hulk, line 8205: Should we add this case to valid test as well?

alamb (commented):
I think the comments from @iffyio  and @git-hulk  are good. Once those get done I think this PR is good to go.

Thanks everyone

tests/sqlparser_common.rs
jmhain: It's syntactically valid, but I don't think any DBMS accepts it this so I'm prefer not to add a test for it.

tests/sqlparser_snowflake.rs
jmhain: lol, thanks, forgot to go back and update it :sweat_smile:

tests/sqlparser_snowflake.rs
jmhain: Done

jmhain:
@iffyio @alamb Addressed the comments. Sorry for the delay in getting back to this!

iffyio (approved):
LGTM! cc @alamb

git-hulk (approved):

alamb (approved):

alamb:
Thank you @jmhain  @iffyio and @git-hulk

PR #1342, opened 2024-07-16
Add support of parsing ON CLUSTER in ALTER TABLE for ClickHouse

git-hulk:
ClickHouse dialect supports `ON CLUSTER` clause for ALTER TABLE:

```sql
ALTER TABLE table_name ON CLUSTER cluster_name ADD COLUMN c UInt32
```

Please refer to: https://clickhouse.com/docs/en/sql-reference/statements/alter

### DIFF
--- a/src/ast/dml.rs
+++ b/src/ast/dml.rs
@@ -126,7 +126,7 @@ pub struct CreateTable {
     pub on_commit: Option<OnCommit>,
     /// ClickHouse "ON CLUSTER" clause:
     /// <https://clickhouse.com/docs/en/sql-reference/distributed-ddl/>
-    pub on_cluster: Option<String>,
+    pub on_cluster: Option<Ident>,
     /// ClickHouse "PRIMARY KEY " clause.
     /// <https://clickhouse.com/docs/en/sql-reference/statements/create/table/>
     pub primary_key: Option<Box<Expr>>,
--- a/src/ast/dml.rs
+++ b/src/ast/dml.rs
@@ -206,11 +206,7 @@ impl Display for CreateTable {
             name = self.name,
         )?;
         if let Some(on_cluster) = &self.on_cluster {
-            write!(
-                f,
-                " ON CLUSTER {}",
-                on_cluster.replace('{', "'{").replace('}', "}'")
-            )?;
+            write!(f, " ON CLUSTER {}", on_cluster)?;
         }
         if !self.columns.is_empty() || !self.constraints.is_empty() {
             write!(f, " ({}", display_comma_separated(&self.columns))?;
--- a/tests/sqlparser_clickhouse.rs
+++ b/tests/sqlparser_clickhouse.rs
@@ -25,6 +25,7 @@ use sqlparser::ast::Value::Number;
 use sqlparser::ast::*;
 use sqlparser::dialect::ClickHouseDialect;
 use sqlparser::dialect::GenericDialect;
+use sqlparser::parser::ParserError::ParserError;
 
 #[test]
 fn parse_map_access_expr() {
@@ -1089,6 +1090,27 @@ fn parse_create_table_on_commit_and_as_query() {
         }
         _ => unreachable!(),
     }
+}
+
+#[test]
+fn test_alter_table_with_on_cluster() {
+    let sql = "ALTER TABLE t ON CLUSTER 'cluster' ADD CONSTRAINT bar PRIMARY KEY (baz)";
+    match clickhouse_and_generic().verified_stmt(sql) {
+        Statement::AlterTable {
+            name, on_cluster, ..
+        } => {
+            assert_eq!(name.to_string(), "t");
+            assert_eq!(on_cluster, Some("'cluster'".to_string()));
+        }
+        _ => unreachable!(),
+    }
+
+    let res = clickhouse_and_generic()
+        .parse_sql_statements("ALTER TABLE t ON CLUSTER 123 ADD CONSTRAINT bar PRIMARY KEY (baz)");
+    assert_eq!(
+        res.unwrap_err(),
+        ParserError("Expected: identifier or cluster literal, found: 123".to_string())
+    )
 }
 
 fn clickhouse() -> TestedDialects {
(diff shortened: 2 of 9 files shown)


src/ast/dml.rs
```diff
@@ -206,11 +206,7 @@ impl Display for CreateTable {
             name = self.name,
         )?;
         if let Some(on_cluster) = &self.on_cluster {
-            write!(
-                f,
-                " ON CLUSTER {}",
-                on_cluster.replace('{', "'{").replace('}', "}'")
```
alamb, line 212: Why did you make this change? I am not familar with the syntax but it seems like a change in behavior. Maybe we could ping the original author to figure out the rationale for adding the extra `'`

alamb (commented):
Thank you @git-hulk

src/ast/dml.rs
git-hulk: @alamb Sorry for forgetting to highlight this change while creating PR. From my understanding, it's used to add a single quote for the cluster name if it's a single quote string. And I moved this behavior into the function: `parse_optional_on_cluster`.

This place was introduced in PR #1291, to see if @philipcristiano can help to review and correct me if I was wrong. Thank you!

philipcristiano: #1291 moved the code. I'm not familiar with the origination of it or why Clickhouse would have the `on_cluster` quoted. 

On a cursory glance it looks as if SQL generation would be missing quotes. This PR appears to testing parsing `ON CLUSTER` with and without quotes but maybe there wasn't a test for outputting SQL that would detect any change in functionality here.

git-hulk: @philipcristiano Thanks for your reply. It's already tested in [parse_create_table_on_cluster](https://github.com/sqlparser-rs/sqlparser-rs/blob/028ada8350d3b2ada4aa67f5e828b318565590f2/tests/sqlparser_common.rs#L3505), so I didn't add in this PR?

git-hulk: `ON CLUSTER` for the create table was first introduced in PR #527. @andyrichardson would you mind having a look at this PR?

tests/sqlparser_clickhouse.rs
```diff
@@ -826,6 +827,27 @@ fn parse_create_table_on_commit_and_as_query() {
     }
 }
 
+#[test]
+fn test_alter_table_with_on_cluster() {
```
iffyio, line 831: we can move the test similar to the create stmt over to common? since the current code applies to all dialects?

tests/sqlparser_common.rs
```diff
@@ -3495,7 +3495,7 @@ fn parse_create_table_on_cluster() {
     let sql = "CREATE TABLE t ON CLUSTER '{cluster}' (a INT, b INT)";
     match generic.verified_stmt(sql) {
         Statement::CreateTable(CreateTable { on_cluster, .. }) => {
-            assert_eq!(on_cluster.unwrap(), "{cluster}".to_string());
+            assert_eq!(on_cluster.unwrap(), "'{cluster}'".to_string());
```
iffyio, line 3498: hmm as far as I can tell the new behavior looks more reasonable to me, curious though looking at the previous behavior, shouldn't this have been wrapped around the single quotes already? (wondering if I'm missing something, because this line seems to suggest that this `replace` being done in the display wasnt working? `on_cluster.replace('{', "'{").replace('}', "}'")`

tests/sqlparser_clickhouse.rs
git-hulk: Good call, will move it soon.

tests/sqlparser_common.rs
git-hulk: @iffyio For the original implementation[1], it changed the single quote string into String, so it use the replace way for the special char like '{' which must be inside the the single quote string.

[1] https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser/mod.rs#L5390

git-hulk:
The CI lint error would be fixed after PR #1352 or #1353 was merged

lovasoa:
For the CI errors, they are not caused by the changes in this pr. https://github.com/sqlparser-rs/sqlparser-rs/pull/1357

tests/sqlparser_common.rs
iffyio: Ah right that makes sense!

src/ast/dml.rs
iffyio: seems the old impl was using the replace functionality to manually inject single quotes when when it was time to display - not familiar with clickhouse so unclear why it used curly braces specifically to anchor the replacement, but in any case I think that behavior would technically be incorrect in some cases like e.g. if clickhouse identifiers allowed curly braces.

With the new behavior, we seem to workaround the issue by instead injecting the single quote during parsing. I wonder if, for correctness and to avoid similar confusion later on, we can instead move away from the current String representation of `on_cluster`, since that the value can be more than a string (e.g. it doesn't look like we'd handle quoted identifiers properly today)? Easiest there would probably be to represent it as an Expr which would be over-representing the type but would let us handle more variants if needed later on, if a dedicated type is deemed to not be worth it in this case.

iffyio: Ah right since ident already includes the single quoted strings then we probably dont need the enum right or? thinking should be enough with `pub on_cluster: Option<Ident>`?

git-hulk: Yes, `Option<Ident>` is enough for now. I'm wondering if we should add an enum to extend more values and create a unified display for this enum. But I'm good with `Option<Ident>` to avoid over-design for the future.

iffyio: I imagine we can probably go with Ident and switch to an enum in the future if we need to (if that happens the context would be clearer as to what the enum would look like)

git-hulk: Sure, will change to `Option<Ident>`.

git-hulk: @iffyio Done, pls have a look at the last commit when you get time. Thank you!

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you (again) @git-hulk and @iffyio

Issue #1332, opened 2024-07-08
Release sqlparser-rs version `0.49.0`


alamb:
Follow on to https://github.com/sqlparser-rs/sqlparser-rs/issues/1296

Filing a ticket here to track the work of a release (mostly so others can follow along)

Target Release Date: Approximately 2024-08-15

Release instructions: https://github.com/sqlparser-rs/sqlparser-rs/blob/main/docs/releasing.md

Steps:
- [x] Make a PR to update `CHANGELOG.md`: https://github.com/sqlparser-rs/sqlparser-rs/pull/1350
- [x] run `cargo release`
- [x] verify the new release is on crates.io: https://crates.io/crates/sqlparser/0.49.0
- [x] Make ticket for next release: https://github.com/sqlparser-rs/sqlparser-rs/issues/1349

alamb:
I plan to make this release early (this week) to get https://github.com/sqlparser-rs/sqlparser-rs/pull/1346 from @alexander-beedie

alamb:
working on this one

alamb:
`0.49.0` is on crates.io: https://crates.io/crates/sqlparser/0.49.0

PR #1346, opened 2024-07-18
Fix quoted identifier regression edge-case with "from" in SELECT

alexander-beedie:
#1212 introduced a regression in the recently released 0.48, disallowing the (valid) quoted identifier `"from"` in SELECT clauses, eg:

```sql
SELECT dt, rate, "to", "from" FROM fx_rates
```
Improved the check by additionally confirming that the _potentially_ invalid `from` is not quoted, and added coverage.

---

FYI: this was caught by our (Polars) SQL unit tests while I was looking to upgrade us to 0.48 to take advantage of a couple of small PRs I added recently. Not sure if you usually make minor point releases to address regressions? If not, we can wait until 0.49 (or temporarily point to the fix on a branch), so no rush if that's the case 👍

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10309,7 +10309,7 @@ impl<'a> Parser<'a> {
             Expr::Wildcard => Ok(SelectItem::Wildcard(
                 self.parse_wildcard_additional_options()?,
             )),
-            Expr::Identifier(v) if v.value.to_lowercase() == "from" => {
+            Expr::Identifier(v) if v.value.to_lowercase() == "from" && v.quote_style.is_none() => {
                 parser_err!(
                     format!("Expected an expression, found: {}", v),
                     self.peek_token().location
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -8959,7 +8959,7 @@ fn parse_non_latin_identifiers() {
 
 #[test]
 fn parse_trailing_comma() {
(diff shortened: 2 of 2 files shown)


alamb (approved):
Thank you very much @alexander-beedie 

cc @MohamedAbdeen21 as the author of https://github.com/sqlparser-rs/sqlparser-rs/pull/1212

@alexander-beedie I don't think we are setup to make a `0.48.1` patch release at this time, but I think I could make a small `0.49.0` release.

alexander-beedie:
> @alexander-beedie I don't think we are setup to make a `0.48.1` patch release at this time, but I think I could make a small `0.49.0` release.

That would be great; whatever best fits your release model is fine by me 🙂

MohamedAbdeen21:
Nice catch. Thanks for fixing it

alamb:
Thanks again -- I'll try and release the next version sometime this week: https://github.com/sqlparser-rs/sqlparser-rs/issues/1332

alexander-beedie:
> Thanks again -- I'll try and release the next version sometime this week: #1332

Fantastic; will finish updating on our side once it's out.

alamb:
This fix has been published as part of `0.49.0` which is on crates.io: https://crates.io/crates/sqlparser/0.49.0

alexander-beedie:
> This fix has been published as part of `0.49.0` which is on crates.io: https://crates.io/crates/sqlparser/0.49.0

Many thanks!

PR #1347, opened 2024-07-19
Allow to use `()` as the GROUP BY nothing

### LINKED ISSUE #1092: GROUP BY () not supported
Hello. First of all, thanks for this amazing project!
I've ran into an issue that explicit `GROUP BY ()` is not supported.

For example, when we write:
`SELECT SUM (value) from sales`,
it is supposed to be the same as:
`SELECT SUM (value) from sales group by ()`,
but the latter is not parseable by the parser. It produces:
`ParserError("Expected an expression:, found: )")`.

I suppose that the issue is that it expects a row expression there, but () is just a special syntax for grouping by nothing.
You technically can specify `GROUP BY (), value`, which would just group by `value`, but if only () is present it treats all rows like a single group.

How would you suggest fixing this?

git-hulk:
Currently, PostgreSQL supports using `()` as one of the group by elements that represent the meaning of nothing. Please refer to the GROUP BY Clause section in [PostgreSQL](https://www.postgresql.org/docs/16/sql-select.html).

This PR will close #1092.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1474,6 +1474,11 @@ impl<'a> Parser<'a> {
                 let result = self.parse_comma_separated(|p| p.parse_tuple(true, true))?;
                 self.expect_token(&Token::RParen)?;
                 Ok(Expr::Rollup(result))
+            } else if self.consume_tokens(&[Token::LParen, Token::RParen]) {
+                // PostgreSQL allow to use empty tuple as a group by expression,
+                // e.g. `GROUP BY (), name`. Please refer to GROUP BY Clause section in
+                // [PostgreSQL](https://www.postgresql.org/docs/16/sql-select.html)
+                Ok(Expr::Tuple(vec![]))
             } else {
                 self.parse_expr()
             }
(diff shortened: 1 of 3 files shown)


git-hulk:
cc @iffyio

src/parser/mod.rs
```diff
@@ -1474,6 +1474,11 @@ impl<'a> Parser<'a> {
                 let result = self.parse_comma_separated(|p| p.parse_tuple(true, true))?;
                 self.expect_token(&Token::RParen)?;
                 Ok(Expr::Rollup(result))
+            } else if self.consume_tokens(&[Token::LParen, Token::RParen]) {
```
iffyio, line 1477: Thinking we can either limit the syntax to the postgres/generic dialect here if we want it postgres specific, or we can move the test case to common.rs so that the test covers all dialect?

git-hulk: @iffyio Good point. The group by nothing is not only supported in Postgre, so I moved it to the common and covered all dialects that support the group by expression.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
🚀

alamb:
Thank you @git-hulk and @iffyio

PR #1351, opened 2024-07-24
Allow to use the TABLE keyword in DESC|DESCRIBE|EXPLAIN TABLE statement

### LINKED ISSUE #1160: snowflake describe table not supported
Snowflake actually uses the keyword TABLE after DESC[RIBE] and it doesn't seems to be supported at the moment

See https://docs.snowflake.com/en/sql-reference/sql/desc-table

git-hulk:
This closes #1160.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7972,10 +7972,13 @@ impl<'a> Parser<'a> {
                     _ => {}
                 }
 
+                // only allow to use TABLE keyword for DESC|DESCRIBE statement
+                let has_table_keyword = self.parse_keyword(Keyword::TABLE);
                 let table_name = self.parse_object_name(false)?;
                 Ok(Statement::ExplainTable {
                     describe_alias,
                     hive_format,
+                    has_table_keyword,
                     table_name,
                 })
             }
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -7972,10 +7972,13 @@ impl<'a> Parser<'a> {
                     _ => {}
                 }
 
+                // only allow to use TABLE keyword for DESC|DESCRIBE statement
+                let has_table_keyword = self.parse_keyword(Keyword::TABLE);
```
git-hulk, line 7976: I didn't add the dialect guard since the `TABLE` keyword is allowed in multiple dialects(Snowflake and ClicHouse), and it should be good to let the downstream determine if it's a valid syntax.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @git-hulk  and @iffyio

alamb:
@samuelcolvin  found a seeming regression due to this change here:  https://github.com/sqlparser-rs/sqlparser-rs/issues/1385

src/parser/mod.rs
alamb: See https://github.com/sqlparser-rs/sqlparser-rs/issues/1385

PR #1352, opened 2024-07-25
Adding support for parsing CREATE TRIGGER statements

LucaCappelletti94:
This pull request starts from what was left off from [this other closed pull request from two years ago](https://github.com/sqlparser-rs/sqlparser-rs/pull/756). @zidaye had already made a considerable effort, I have primarily updated the PR so that it works with the code base two years later and extended the testing for its methods.

Looking forward to see it merged so I may be able to add support [for TRIGGER in GlueSQL](https://github.com/gluesql/gluesql/issues/1520)


alamb:
fyi @lovasoa

git-hulk:
@LucaCappelletti94 Would you mind fixing the fmt issue and test failure found in CI?

LucaCappelletti94:
Yeah, will get started shortly.

LucaCappelletti94:
@alamb, @git-hulk I need another GitHub Action run to be approved please.

lovasoa (approved):

src/parser/mod.rs
```diff
@@ -4139,6 +4160,163 @@ impl<'a> Parser<'a> {
...
+
+            let mut referencing = vec![];
+            if self.parse_keyword(Keyword::REFERENCING) {
+                while let Some(refer) = self.parse_trigger_referencing()? {
+                    referencing.push(refer);
+                }
+            }
+
+            let for_each = if self.parse_keywords(&[Keyword::FOR]) {
+                let _ = self.parse_keyword(Keyword::EACH);
```
git-hulk, line 4183: It'll change `FOR` to `FOR EACH` while formatting SQL, perhaps we should record if `EACH` keyword existing?

src/test_utils.rs
```diff
@@ -111,10 +111,10 @@ impl TestedDialects {
     /// that:
     ///
     /// 1. parsing `sql` results in the same [`Statement`] as parsing
-    /// `canonical`.
```
git-hulk, line 114: Was this file formatted using `cargo fmt`?

src/ast/mod.rs
```diff
@@ -3344,6 +3418,37 @@ impl fmt::Display for Statement {
...
+                condition,
+                exec_body,
+            } => {
+                write!(
+                    f,
+                    "CREATE {or_replace}TRIGGER {name} {period}",
+                    or_replace = if *or_replace { "OR REPLACE " } else { "" },
+                )?;
+                if !event.is_empty() {
+                    write!(f, " {}", display_separated(event, "or"))?;
```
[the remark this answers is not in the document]
LucaCappelletti94: Hi @git-hulk, I am not sure which refactoring you would want me to make here. Could you clarify?

src/test_utils.rs
LucaCappelletti94: One of the fmt warnings was about `canonical` not being adequately spaced, so I changed that.

src/parser/mod.rs
```diff
@@ -4139,6 +4160,163 @@ impl<'a> Parser<'a> {
...
+
+    pub fn parse_trigger_period(&mut self) -> Result<TriggerPeriod, ParserError> {
+        match self.parse_one_of_keywords(&[Keyword::BEFORE, Keyword::AFTER, Keyword::INSTEAD]) {
+            Some(Keyword::BEFORE) => Ok(TriggerPeriod::Before),
+            Some(Keyword::AFTER) => Ok(TriggerPeriod::After),
+            Some(Keyword::INSTEAD) => {
+                if self.parse_keyword(Keyword::OF) {
+                    Ok(TriggerPeriod::InsteadOf)
+                } else {
+                    self.expected("an `OF` after `INSTEAD` ", self.peek_token())
```
git-hulk, line 4232:
```suggestion
                    self.expected("an `OF` after `INSTEAD`", self.peek_token())
```

src/ast/mod.rs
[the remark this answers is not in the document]
git-hulk: @LucaCappelletti94 Sorry for the vague comment. What I mean is to change:

```
write!(f, " {}", display_separated(event, "or"))?;
```

to

```
write!(f, " {}", display_separated(event, "OR"))?;
```

src/test_utils.rs
git-hulk: Sure, thank you. I just want to confirm whether those are expected changes in this file.

src/ast/mod.rs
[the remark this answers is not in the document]
LucaCappelletti94: Ah okay, sure. Maybe I should introduce a constant for this string?

src/parser/mod.rs
LucaCappelletti94: Ok, doing.

src/ast/mod.rs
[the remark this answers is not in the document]
git-hulk: Just replace "or" with "OR" is also good for me.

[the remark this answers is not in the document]
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: Done, and included both "FOR" and "FOR EACH" in the test suite.

git-hulk (approved):

git-hulk:
cc @alamb

src/parser/mod.rs
```diff
@@ -4139,6 +4160,167 @@ impl<'a> Parser<'a> {
...
+                let args = self.parse_comma_separated(Parser::parse_function_arg)?;
+                self.expect_token(&Token::RParen)?;
+                args
+            }
+        } else {
+            vec![]
+        };
+
+        Ok(FunctionDesc { name, args })
+    }
```
iffyio, line 4322: this looks similar to [parse_drop_function_desc](https://github.com/sqlparser-rs/sqlparser-rs/blob/d5a6af12fd8fae89d9e861dc143dafd5378a67ab/src/parser/mod.rs#L4846C8-L4846C32) can we reuse the logic for both if so?
I wonder if it would make sense to reuse the `DropFunctionDesc` as well (we might then be able to justify renaming the type to a generic `FunctionDesc`)

src/parser/mod.rs
```diff
@@ -4139,6 +4160,167 @@ impl<'a> Parser<'a> {
...
+                (None, false)
+            };
+
+            let condition = if self.parse_keyword(Keyword::WHEN) {
+                Some(self.parse_expr()?)
+            } else {
+                None
+            };
+
+            let exec_body = if self.parse_keyword(Keyword::EXECUTE) {
```
iffyio, line 4204: similar to the one_of, we also have a [self.expect_keyword](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser/mod.rs#L3317)

(can we do a pass through the changes to replace the `if parse .. else error` pattern to reuse the functionality?)

src/parser/mod.rs
```diff
@@ -4139,6 +4160,167 @@ impl<'a> Parser<'a> {
...
+            if self.parse_keyword(Keyword::REFERENCING) {
+                while let Some(refer) = self.parse_trigger_referencing()? {
+                    referencing.push(refer);
+                }
+            }
+
+            let (for_each, include_each) = if self.parse_keywords(&[Keyword::FOR]) {
+                let include_each = self.parse_keyword(Keyword::EACH);
+                (
+                    match self.parse_one_of_keywords(&[Keyword::ROW, Keyword::STATEMENT]) {
```
iffyio, line 4185: we can use [self.expect_one_of_keywords](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser/mod.rs#L3303)

tests/sqlparser_postgres.rs
```diff
@@ -4441,3 +4441,218 @@ fn test_table_unnest_with_ordinality() {
         _ => panic!("Expecting TableFactor::UNNEST with ordinality"),
     }
 }
+
+#[test]
+fn parse_create_trigger() {
```
iffyio, line 4446: Can we include coverage for following?
- multiple events? (e.g. `INSERT OR UPDATE ...`)
- `EXECUTE PROCEDURE`
- execute with one or more function arguments
- (if the parser supports it), timing like `DEFERRABLE`, `INITIALLY IMMEDIATE` etc'
- referencing without alias (`REFERENCING NEW TABLE FOR ...`)

Would be nice if we could also include negative tests, like invalid combinations or incomplete statements that we expect the parser to reject

src/ast/mod.rs
```diff
@@ -53,6 +53,12 @@ pub use self::query::{
     TableAlias, TableFactor, TableVersion, TableWithJoins, Top, TopQuantity, ValueTableMode,
     Values, WildcardAdditionalOptions, With, WithFill,
 };
+
+pub use self::trigger::{
+    ExecBodyType, FunctionDesc, TriggerEvent, TriggerExecBody, TriggerObject, TriggerPeriod,
```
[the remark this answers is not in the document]
LucaCappelletti94: Proposed names?

tests/sqlparser_postgres.rs
LucaCappelletti94: Hi, I am not too familiar with these cases, could you kindly provide a full sql example of each one so I can make sure those are covered?

src/ast/trigger.rs
```diff
@@ -0,0 +1,173 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! SQL Abstract Syntax Tree (AST) for triggers.
+use super::*;
+
+/// Function describe in DROP FUNCTION.
```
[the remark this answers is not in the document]
LucaCappelletti94: By the looks of it, it seems a copy paste error. On it.

src/ast/mod.rs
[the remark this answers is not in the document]
LucaCappelletti94: Renamed with version with "TRIGGER" as prefix.

src/parser/mod.rs
```diff
@@ -4139,6 +4160,167 @@ impl<'a> Parser<'a> {
...
+            refer_type,
+            is_as,
+            transition_relation_name,
+        }))
+    }
+
+    pub fn parse_trigger_exec_body(&mut self) -> Result<TriggerExecBody, ParserError> {
+        let exec_type = match self.parse_one_of_keywords(&[Keyword::FUNCTION, Keyword::PROCEDURE]) {
+            Some(Keyword::FUNCTION) => ExecBodyType::Function,
+            Some(Keyword::PROCEDURE) => ExecBodyType::Proceduer,
```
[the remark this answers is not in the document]
LucaCappelletti94: Integrated proposed typo fix but refused commit because of collision with ExecBodyType being renamed.

src/parser/mod.rs
LucaCappelletti94: Yeah the two things are basically twins, I will proceed to keep only the generic `FunctionDesc`.

LucaCappelletti94: Done - uniformed the two and updated testing accordingly.

LucaCappelletti94: Maybe it should also be uniformed for `CreateFunction`, but I believe such refactoring would be outside the scope of this PR.

src/parser/mod.rs
LucaCappelletti94: Thank you for pointing out the `expect_keyboard` method, I have refactored the code accordingly now in every position where I found a similar pattern.

src/parser/mod.rs
LucaCappelletti94: Done!

tests/sqlparser_postgres.rs
LucaCappelletti94: Now handled multiple events and added several cases in test suite.

LucaCappelletti94: Also procedure added now.

LucaCappelletti94: And added testing for multiple function arguments.

LucaCappelletti94: And added also DEFERRABLE / INITIALLY IMMEDIATE.

LucaCappelletti94: Added test suite for referencing

LucaCappelletti94: Added some negative testing

src/parser/mod.rs
```diff
@@ -4139,6 +4160,199 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse statements of the DropTrigger type such as:
+    ///
+    /// ```sql
+    /// DROP TRIGGER [ IF EXISTS ] name ON table_name [ CASCADE | RESTRICT ]
+    /// ```
+    pub fn parse_drop_trigger(&mut self) -> Result<Statement, ParserError> {
+        if dialect_of!(self is PostgreSqlDialect) {
```
iffyio, line 4169:
```suggestion
        if !dialect_of!(self is PostgreSqlDialect | GenericDialect) {
              return self.expected(...)
        }
        ...
```
We can return early instead, so that the body of the function can be unnested, also we can include [genericdialect](https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#new-syntax) support as well

(same for create trigger)

tests/sqlparser_postgres.rs
```diff
@@ -4441,3 +4441,614 @@ fn test_table_unnest_with_ordinality() {
...
+    for referencing in [
```
iffyio, line 4447: I'm a bit worried the multiple nestings make this test somewhat harder to follow, and as a result make changes to, going forward. To simplify we could have one or few asserts that looks at the body of the statement.
Then for tests on the diffferent variations of the various components of the statement, we can rely on shorter roundtrip-only assertions - here's one example test that does similar https://github.com/sqlparser-rs/sqlparser-rs/blob/e81a58cd88a55db26f16c035aac04ed4e10c1c59/tests/sqlparser_common.rs#L7288-L7337

tests/sqlparser_postgres.rs
```diff
@@ -4441,3 +4441,614 @@ fn test_table_unnest_with_ordinality() {
...
+                                        Ident::new("balance"),
+                                        Ident::new("name"),
+                                    ]),
+                                    TriggerEvent::Insert,
+                                    TriggerEvent::Delete,
+                                ],
+                                "UPDATE OF balance, name OR INSERT OR DELETE",
+                            ),
+                        ] {
+                            for when in [
```
iffyio, line 4625:
```suggestion
                            for period in [
```
should this be called period instead of when?

src/parser/mod.rs
```diff
@@ -4139,6 +4160,199 @@ impl<'a> Parser<'a> {
...
+
+            let mut referencing = vec![];
+            if self.parse_keyword(Keyword::REFERENCING) {
+                while let Some(refer) = self.parse_trigger_referencing()? {
+                    referencing.push(refer);
+                }
+            }
+
+            let (trigger_object, include_each) =
+                self.expect_keyword(Keyword::FOR).and_then(|_| {
```
iffyio, line 4234:
```suggestion
                self.expect_keyword(Keyword::FOR)?;
```
We can exit with an error immediately instead of nesting the logic within closures, this pattern is currently done in a few places in the PR, can we do a pass through to unnest them?

src/ast/ddl.rs
```diff
@@ -1154,16 +1154,50 @@ fn display_option_spaced<T: fmt::Display>(option: &Option<T>) -> impl fmt::Displ
 /// `<constraint_characteristics> = [ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ] [ ENFORCED | NOT ENFORCED ]`
 ///
 /// Used in UNIQUE and foreign key constraints. The individual settings may occur in any order.
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[derive(Debug, Copy, Clone, Default, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct ConstraintCharacteristics {
+    /// `[ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]`
+    pub deferrable: DeferrableCharacteristics,
```
iffyio, line 1162: hmm not sure I follow, why did this change?

src/ast/mod.rs
```diff
@@ -2589,6 +2596,91 @@ pub enum Statement {
...
+        /// every row affected by the trigger event, or just once per SQL statement.
+        trigger_object: TriggerObject,
+        /// Whether to include the `EACH` term of the `FOR EACH`, as it is optional syntax.
+        include_each: bool,
+        ///  Triggering conditions
+        condition: Option<Expr>,
+        /// Execute logic block
+        exec_body: TriggerExecBody,
+        /// The characteristic of the trigger, which include whether the trigger is `DEFERRABLE`, `INITIALLY DEFERRED`, or `INITIALLY IMMEDIATE`,
+        characteristics: Option<DeferrableCharacteristics>,
```
iffyio, line 2669: Ah I see, would it make sense to use `ConstraintCharacteristics` directly instead?
 that would also let us use the parser `parse_constraint_characteristics` for it likely, which would be simpler. It would technically contain a superset of what's allowed for triggers in terms of representation but I think that would be okay

tests/sqlparser_postgres.rs
LucaCappelletti94: Hi - I am not sure I understood. The number of things to test is combinatorial, so, naturally, there will be lots of nested loops for all of the combinations, right? I believe it is much better to have nested for loops than extremely repetitive tests because as it is right now one can edit a single thing and be done with it.

I have not understood how the example you provided handles combinatoriality.

src/ast/mod.rs
LucaCappelletti94: `ConstraintCharacteristics` has the attribute `enforced`, which would not make sense for a `TRIGGER`. For this reason, I have separated the two structs so that the code redundancy is minimal while not having incorrect attributes in the `CreateTrigger` struct. Using `parse_constraint_characteristics` would allow for parsing of incorrect `CREATE TRIGGER` statements that include `ENFORCED` statements, and would require us to check that such a term is not included, making the overall code more complex.

src/ast/ddl.rs
LucaCappelletti94: I am marking this conversation as resolved since this topic discussion has moved to the other conversation.

src/parser/mod.rs
LucaCappelletti94: By using the `and_then`, the logic relative to the presence or absence of a keyword is delimited in scope and I believe it makes the logic easier to follow. If I were to turn that in "the rows that follow the expect keyword", I believe it would not be as clear.

Could you please expand on the motivation for this proposed refactoring other than "we can write it also this other way"?

src/parser/mod.rs
LucaCappelletti94: Applied the proposed change to both `create_trigger` and `drop_trigger`. Do let me know whether the change in commit ba5d47399067b6e4513befc661b48080844c9165 matches your expectations, so we may mark this conversation as resolved.

src/parser/mod.rs
iffyio: Ah, yeah so to clarify with an example, semantically both following styles are roughly equivalent:

```rust
let a = doA()?;
let b = doB()?;
combine(a,b)
```
vs

```rust
doA().and_then(|a| doB().map(|b| (a,b))).and_then(|a,b| combine(a,b))
```

I think its arguable that `and_then` style is easier to follow. I can understand that its a subjective thing however.

My initial comment mentioned nesting the code which I think is one side-effect/downside of the `and_then` style because each layer of nesting in code generally tends to obfuscate the code even if slightly. (again, this is also arguable/subjective)

In any case, I did a quick search in this parser.mod file in this PR for occurrences of `and_then` and there are only 5 matches, all of which are being introduced in this PR so I think its safe to say that the codebase prefers the former style?

LucaCappelletti94: Ok, I will adjust the code to the codebase style. Do let me know if you identify other analogous style divergences.

LucaCappelletti94: Is it okay if I leave the `and_then` associated to `expect_one_of_keywords`?

LucaCappelletti94: Now there remains only the following two `and_then`: [`parse_trigger_period`](https://github.com/LucaCappelletti94/sqlparser-rs/blob/c043b583c091dc5d1b8bf0d4cff53cc399ae20d5/src/parser/mod.rs#L4273-L4285) and [`parse_trigger_event`](https://github.com/LucaCappelletti94/sqlparser-rs/blob/c043b583c091dc5d1b8bf0d4cff53cc399ae20d5/src/parser/mod.rs#L4287-L4312). Do let me know whether you would want also these removed.

tests/sqlparser_postgres.rs
iffyio: Ah so it was test scenarios in this style specifically I was trying to highlight in the link
https://github.com/sqlparser-rs/sqlparser-rs/blob/e81a58cd88a55db26f16c035aac04ed4e10c1c59/tests/sqlparser_common.rs#L7325-L7330

maybe to use a clearer example, the test right before the test in question https://github.com/sqlparser-rs/sqlparser-rs/blob/c3a96e0be51b7057e6f597fc56eab1876fe6a1c4/tests/sqlparser_postgres.rs#L4350-L4387

so e.g. just calling

```sql
pg_and_generic().verified_expr("CURRENT_TIMESTAMP AT TIME ZONE tz");
```
as a test case can be similarly effective in demonstrating the parser behavior, given that we already have other test case(s) with more detailed asserts on the actual returned AST body for the statement/expr type being tested.

Re the combinatorial aspects of the syntax, a for loop could be reasonable when there are 1-2 variables, but this case feels extreme (has more cons than pros) to me so that it makes sense to me to consider the alternative style - my thinking is also that, when reading the code, if one needs to see that a specific syntax is supported, its clear from the list of raw sql what's supported vs not. Similarly, if one needs to include a new test case for a combination/new-variant, they only need to add one-line of code. In both cases, it would be a hassle to have to figure out the layers of nesting and mechanism this specific test uses in order to find what they're looking for, or to find if/where to inject their new test case.

Above is subjective ofc, I speak for only myself that the current func is harder for me to tell what's tested vs what's not - and the function sticks out when compared to others in the codebase (including those with more complicated syntax).

src/parser/mod.rs
```diff
@@ -4139,6 +4160,194 @@ impl<'a> Parser<'a> {
...
+        .then(|| {
+            Ok::<_, ParserError>(
+                match self.expect_one_of_keywords(&[Keyword::IMMEDIATE, Keyword::DEFERRED])? {
+                    Keyword::IMMEDIATE => DeferrableInitial::Immediate,
+                    Keyword::DEFERRED => DeferrableInitial::Deferred,
+                    _ => unreachable!(),
+                },
+            )
+        })
+        .transpose()?;
```
iffyio, line 4226: we can also flatten this? a regular if else would suffice i imagine

src/parser/mod.rs
```diff
@@ -4139,6 +4160,194 @@ impl<'a> Parser<'a> {
...
+        Ok(Some(TriggerReferencing {
+            refer_type,
+            is_as,
+            transition_relation_name,
+        }))
+    }
+
+    pub fn parse_trigger_exec_body(&mut self) -> Result<TriggerExecBody, ParserError> {
+        let exec_type = self
+            .expect_one_of_keywords(&[Keyword::FUNCTION, Keyword::PROCEDURE])
```
iffyio, line 4338: I think this would be another candidate to flatten?

src/parser/mod.rs
iffyio: Yes please we can update those as well thanks! Also highlighted a couple others in the latest comments I think we can update

src/ast/mod.rs
iffyio: hmm, the issues I saw were that:
- with the changes to `ConstraintCharacteristics`, the representation of that struct becomes somewhat haphazard, previously it was a bit clear how the different fields belong together, now the extra nesting of `deferrable` is arbitrary which is confusing (looking at the struct its not at all clear why deferrable is nested within it, and there's no reason for it to be.)
- re checking `ENFORCED` downstream, I think that's covered by this part of the [project's goals](https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics) - in that a downstream project will likely usually need to do some semantic analysis - keep in mind that in this specific scenario, if e.g. another dialect has different requirements/additional characteristics that need to live in the new  deferrable struct, you'd still end up having to  check in your project after all.

The changes feels to me like they've pushed the complexity into this library rather than handle outside (which would likely be simpler to) - with the changes, we seem to have introduced an arbitrary layer of abstraction/indirection, and introduced duplicate [parsing logic](https://github.com/sqlparser-rs/sqlparser-rs/blob/c043b583c091dc5d1b8bf0d4cff53cc399ae20d5/src/parser/mod.rs#L4208-L4226) for that layer

src/parser/mod.rs

src/parser/mod.rs

src/parser/mod.rs

src/ast/mod.rs
LucaCappelletti94: Restored to `ConstraintCharacteristics` in commit 09581e74360d56eec40b7b34cf1767e35b1af702

tests/sqlparser_postgres.rs
LucaCappelletti94: Ok, what about keeping both the combinatorial loop and adding more examples of parsable strings such as the ones you propose? This way there is both the complete testing of all possible branches and combinations, plus the examples that a user may CTRL+F.

src/ast/ddl.rs
```diff
@@ -1210,22 +1201,24 @@ impl ConstraintCharacteristics {
...
-            (None, None, Some(enforced)) => write!(f, "{enforced}"),
-            (None, Some(initial), None) => write!(f, "{initial}"),
-            (None, Some(initial), Some(enforced)) => write!(f, "{initial} {enforced}"),
-            (Some(deferrable), None, None) => write!(f, "{deferrable}"),
-            (Some(deferrable), None, Some(enforced)) => write!(f, "{deferrable} {enforced}"),
-            (Some(deferrable), Some(initial), None) => write!(f, "{deferrable} {initial}"),
-            (Some(deferrable), Some(initial), Some(enforced)) => {
-                write!(f, "{deferrable} {initial} {enforced}")
-            }
+        let mut parts = vec![];
```
iffyio, line 1204: are the changes here now solely a rewrite (if they're no longer part of the PR we can revert them)?

tests/sqlparser_postgres.rs
iffyio: The combinatorial loop is primarily the source of complexity in my mind and would ideally be flattened in order to simplify it.

src/ast/ddl.rs
LucaCappelletti94: Done in commit [058cf7f](https://github.com/sqlparser-rs/sqlparser-rs/pull/1352/commits/058cf7faee2f74193804592446826e7b5f564ea3)

src/ast/data_type.rs
```diff
@@ -319,6 +319,10 @@ pub enum DataType {
     /// [`SQLiteDialect`](crate::dialect::SQLiteDialect), from statements such
     /// as `CREATE TABLE t1 (a)`.
     Unspecified,
+    /// Trigger data type, returned by functions associated with triggers
+    ///
+    /// [postgresql]: https://www.postgresql.org/docs/current/plpgsql-trigger.html
+    Trigger,
```
iffyio, line 325: for the datatype can we add a test case (noticed I couldn't find test code exercising it)?

tests/sqlparser_postgres.rs
```diff
@@ -4453,6 +4763,312 @@ fn test_escaped_string_literal() {
...
+            default_ddl_collation: None,
+            with_aggregation_policy: None,
+            with_row_access_policy: None,
+            with_tags: None,
+        }
+    );
+
+    // Check the second statement
+
+    dbg!(&create_function);
```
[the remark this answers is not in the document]
LucaCappelletti94: Thank you for pointing this one out, I must have mistakenly committed it.

src/ast/data_type.rs
LucaCappelletti94: There is one here: https://github.com/LucaCappelletti94/sqlparser-rs/blob/7ab8fdb05ba937b0e14c376866b5ac280e5b59c6/tests/sqlparser_postgres.rs#L4866

iffyio: Ah right, didn't notice it was the datatype within a function, sounds good!

tests/sqlparser_postgres.rs
LucaCappelletti94: I have flattened and tried to clean as much the loop, and I have now included more cases like the CONSTRAINT keyword and the `FROM {referenced_table_name}` case. This covers all possible combinations of the CREATE TRIGGER, and the combinatoriality led me to discover a few cases I had not covered in the parser adequately. I hope this is now more readable, do let me know.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ sqlparser_derive = { version = "0.2.0", path = "derive", optional = true }
 simple_logger = "5.0"
 matches = "0.1"
 pretty_assertions = "1"
+itertools = "0.13"
```
lovasoa, line 43: I'm not a huge fan of adding a new dependency...

LucaCappelletti94: In the meantime, I will rewrite it without the use of itertools. Do let me know which variant you prefer.

src/parser/mod.rs
```diff
@@ -2026,6 +2026,25 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse a keyword-separated list of 1+ items accepted by `F`
+    pub fn parse_keyword_separated<T, F>(
```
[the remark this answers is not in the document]
LucaCappelletti94: I have moved the function `parse_keyword_separated ` to where you suggested.

tests/sqlparser_postgres.rs
iffyio: Hmm, I think the situation didn't improved with the latest changes unfortunately, if anything the custom iterating obfuscates further compared to a familiar for-loop. By flattening I meant as in my previous comments/examples on this thread - is it something preventing us from using the same style as the linked examples for this syntax (I imagine there are plenty examples with that style in the codebase to follow)?

Cargo.toml
lovasoa: I think your version is okay. I would have written nested for loops, though.

LucaCappelletti94: Ok! I find the flattened for-loops a bit easier to read, as otherwise, it gets to a very high level of indentation. Do let me know whether you have any more notes on the PR.

tests/sqlparser_postgres.rs
LucaCappelletti94: Those approaches simply do not cover all possible combinations that are relevant to these objects. Yesterday, extending the combinations further, I found corner cases that I would have not thought of otherwise.

Writing out extensively the tests with the approach you have suggested, **as I understand it,** would require writing an extremely large amount of tests, which would ultimately be less readable than the current simple for loop.

tests/test_utils/extend_tuple.rs
```diff
@@ -0,0 +1,177 @@
...
+            to_extend.5,
+            to_extend.6,
+            to_extend.7,
+            to_extend.8,
+            to_extend.9,
+            to_extend.10,
+            self,
+        )
+    }
+}
```
lovasoa, line 177: This is, in my opinion, worse than a bunch of nested for loops... But I wonldn't prevent merging over that. I think this PR is ok.

lovasoa (approved):

tests/test_utils/extend_tuple.rs
LucaCappelletti94: Yeah, these are quite ugly - I can wrap them in macros to remove at least the redundancy.

lovasoa: I don't feel like that would be worth it. A few nested for loops are better than : 
 - adding a new dependency
 - creating a new trait that will be used just in one test case
 - creating a new trait AND a macro to implement that trait

We are talking about a single test. No need to overcomplicate things.

Even the principle of testing all possible combinations of syntaxes sounds a bit overkill to me.

I don't think we should delay merging that PR for longer. I would have preferred simpler tests, but the current state is okay. We can merge this now, I think.

@alamb

LucaCappelletti94: Ok!

LucaCappelletti94:
I am not sure why the linter screams about that trait, it is definitely being used.

lovasoa:
@LucaCappelletti94 , it doesn't seem to be used in tests/test_utils/mod.rs where it's imported

tests/test_utils/extend_tuple.rs
iffyio: I'm not in favor of merging this (I think its too much machinery/code for a test case). The previous version with the combinatorial for loops I'd rather not either but I don't feel as strongly against merging that one at least so I'm defo OK with doing that if everyone's fine with it. (doesn't mean we can't merge this either ofc, merely my preference)

LucaCappelletti94: I have replaced the complete CreateTrigger test suite with some selected test cases in commit [459dae1](https://github.com/sqlparser-rs/sqlparser-rs/pull/1352/commits/459dae1415590f301e9ab272f426bb8ff5140d4e), as I believe the request was. I am not sure whether this was what was requested.

lovasoa (approved):
great, good for me !

alamb (approved):
Thank you @LucaCappelletti94 @git-hulk @lovasoa and @iffyio 

This is one of the most well documented PRs I have seen in a long time. 

I have one more question about some dead looking code, but I also think we could remove this as a follow on PR

tests/sqlparser_postgres.rs
```diff
@@ -4442,6 +4442,184 @@ fn test_table_unnest_with_ordinality() {
     }
 }
 
+fn possible_trigger_referencing_variants() -> Vec<Vec<TriggerReferencing>> {
```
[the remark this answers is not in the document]
LucaCappelletti94: Dead code removed, I am not sure why clippy didn't say anything.

alamb:
I took the liberty of merging up from main and pushing a small test fix to this branch -- I also think that by doing so the CI should run automatically on this PR from now on

alamb:
I plan to wait a day or two so @iffyio  has time to review if they would like prior to merge

iffyio (approved):
Thanks @LucaCappelletti94!

alamb:
Thanks again everyone -- this was some nice teamwork

PR #1358, opened 2024-07-29
Parse SETTINGS clause for ClickHouse table-valued functions

Jesse-Bakker:
Allows parsing the [`executable`](https://clickhouse.com/docs/en/engines/table-functions/executable) table-valued function, for example

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -899,6 +899,14 @@ impl fmt::Display for ExprWithAlias {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct TableFunctionArgs {
+    pub args: Vec<FunctionArg>,
+    pub settings: Option<Vec<Setting>>,
+}
+
 /// A table name or a parenthesized subquery with an optional alias
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -916,7 +924,7 @@ pub enum TableFactor {
         /// This field's value is `Some(v)`, where `v` is a (possibly empty)
         /// vector of arguments, in the case of a table-valued function call,
         /// whereas it's `None` in the case of a regular table name.
-        args: Option<Vec<FunctionArg>>,
+        args: Option<TableFunctionArgs>,
         /// MSSQL-specific `WITH (...)` hints such as NOLOCK.
         with_hints: Vec<Expr>,
         /// Optional version qualifier to facilitate table time-travel, as
@@ -1314,7 +1322,12 @@ impl fmt::Display for TableFactor {
                     write!(f, "PARTITION ({})", display_comma_separated(partitions))?;
                 }
                 if let Some(args) = args {
-                    write!(f, "({})", display_comma_separated(args))?;
+                    write!(f, "(")?;
+                    write!(f, "{}", display_comma_separated(&args.args))?;
+                    if let Some(ref settings) = args.settings {
+                        write!(f, ", SETTINGS {}", display_comma_separated(settings))?;
+                    }
+                    write!(f, ")")?;
                 }
                 if *with_ordinality {
                     write!(f, " WITH ORDINALITY")?;
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -899,11 +899,16 @@ impl fmt::Display for ExprWithAlias {
     }
 }
 
+/// Arguments to a table-valued function
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct TableFunctionArgs {
     pub args: Vec<FunctionArg>,
+    /// ClickHouse-specific SETTINGS clause.
+    /// For example,
+    /// `SELECT * FROM executable('generate_random.py', TabSeparated, 'id UInt32, random String', SETTINGS send_chunk_header = false, pool_size = 16)`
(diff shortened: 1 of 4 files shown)


tests/sqlparser_clickhouse.rs
```diff
@@ -1091,6 +1091,38 @@ fn parse_create_table_on_commit_and_as_query() {
     }
 }
 
+#[test]
+fn parse_select_table_function_settings() {
+    let sql = r#"SELECT * FROM table_function(arg, SETTINGS setting = 3)"#;
```
iffyio, line 1096: Can we extend the test case to cover more than one setting?

src/ast/query.rs
```diff
@@ -916,7 +924,7 @@ pub enum TableFactor {
         /// This field's value is `Some(v)`, where `v` is a (possibly empty)
         /// vector of arguments, in the case of a table-valued function call,
         /// whereas it's `None` in the case of a regular table name.
-        args: Option<Vec<FunctionArg>>,
+        args: Option<TableFunctionArgs>,
```
iffyio, line 927: instead of introducing a new type, can we add a new variant to [FunctionArg](https://github.com/sqlparser-rs/sqlparser-rs/blob/c48f157995d5c57a970aa1c91f3045f5b653d3b1/src/ast/mod.rs#L4739-L4746) and update the parser to conditionally accept it for [clickhouse/generic](https://github.com/sqlparser-rs/sqlparser-rs/blob/c48f157995d5c57a970aa1c91f3045f5b653d3b1/src/parser/mod.rs#L10290)?
I think that would let us reuse existing types and be a less invasive change.

e.g.
```rust
pub enum FunctionArg {
    Named {
        name: Ident,
        arg: FunctionArgExpr,
        operator: FunctionArgOperator,
    },
    Unnamed(FunctionArgExpr),
    Settings {
        name: Ident, arg: Expr
    }
}
```

src/ast/query.rs
```diff
@@ -899,6 +899,14 @@ impl fmt::Display for ExprWithAlias {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct TableFunctionArgs {
+    pub args: Vec<FunctionArg>,
+    pub settings: Option<Vec<Setting>>,
```
iffyio, line 907: Can we include a description of the syntax and link to the docs? that would help folks in navigating the code later on

src/ast/query.rs
Jesse-Bakker: The reason I went with a new type is twofold:
 - The `SETTINGS` clause is a single clause, at the end of the args list
 - It is only allowed on table-valued functions

Another option is to add a `settings: Vec<Setting>` field to `TableFactor::Table`. That loses the information that it's part of the argument list, but retains the single-clause and table-valued function only aspect and is more compatible.
wdyt?

iffyio: Ah I see, yeah the current representation sounds reasonable given settings is a single clause

tests/sqlparser_clickhouse.rs
Jesse-Bakker: I added a setting. I didn't expand on the test too much, because the settings parsing is already more exhaustively tested by the tests for `SETTINGS` in queries

tests/sqlparser_clickhouse.rs
```diff
@@ -1091,6 +1091,44 @@ fn parse_create_table_on_commit_and_as_query() {
     }
 }
 
+#[test]
+fn parse_select_table_function_settings() {
```
git-hulk, line 1095: Could you please add the negative case for parsing SETTINGS?

src/parser/mod.rs
```diff
@@ -3426,6 +3426,27 @@ impl<'a> Parser<'a> {
         Ok(values)
     }
 
+    fn parse_comma_separated_end(&mut self) -> Option<Token> {
```
git-hulk, line 3429: Seems the returned token is never used, so it should be good to use `bool` instead of `Option`? @iffyio How do you think?

src/parser/mod.rs
```diff
@@ -10317,6 +10329,30 @@ impl<'a> Parser<'a> {
...
+            }
+            args.push(self.parse_function_args()?);
+            if self.parse_comma_separated_end().is_some() {
+                break None;
+            }
+        };
+        self.expect_token(&Token::RParen)?;
+        Ok(TableFunctionArgs { args, settings })
+    }
+
```
git-hulk, line 10355: Is it more intuitive with the following implementation?
```suggestion
    fn parse_table_function_args(&mut self) -> Result<TableFunctionArgs, ParserError> {
        if self.consume_token(&Token::RParen) {
            return Ok(TableFunctionArgs {
                args: vec![],
                settings: None,
            });
        }

        let mut args = vec![];
        let settings = loop {
            if let Some(settings) = self.parse_settings()? {
                break Some(settings);
            }
            args.push(self.parse_function_args()?);
            if self.parse_comma_separated_end().is_some() {
                break None;
            }
        };
        self.expect_token(&Token::RParen)?;
        Ok(TableFunctionArgs { args, settings })
    }

```

src/parser/mod.rs
Jesse-Bakker: Upon re-reading the implementation, the return value is confusing (it returns `Some(Token::Comma)` if there is no comma and the method name is confusing as well. Maybe it should just be `fn is_comma_separated_end(&self) -> bool` and the call-site is `if !self.consume_token(&Token::Comma) || self.is_comma_separated_end() { break; }`

git-hulk: Yes, but I'm wondering if it'll be strange to return `ControlFlow` since the function mainly checks if it reached the end. I prefer changing to `is_parse_comma_separated_end` and returning a boolean value.

iffyio: Yeah calling the function `is_parse_comma_separated_end` and having it return a boolean sound reasonable to me!

git-hulk (approved):

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @Jesse-Bakker  and @git-hulk and @iffyio

alamb:
🚀

PR #1360, opened 2024-08-01
Support `Dialect` level precedence, update Postgres `Dialect` to match Postgres

samuelcolvin:
~~This is ugly but working, I wanted to get feedback before proceeding.~~ I think this is now in a good state, definitely ready to review.

This solves https://github.com/sqlparser-rs/sqlparser-rs/issues/814#issuecomment-2196624432, e.g. it gives the right AST for `SELECT foo->'bar'='spam'` etc.

But before I go further, I want to confirm a few things?

### Can we update precedence globally?

The claim at 

https://github.com/sqlparser-rs/sqlparser-rs/blob/cc13841a370190df00cfc623593453054ac187e9/src/parser/mod.rs#L2972

Isn't really true at all, `sqlparser-rs` doesn't use the lexical precedence from PostgreSQL 7, even if it did, that's extremely old.

Could we just update those values to reference [a more recent version](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE)?

I assume not since that order is presumably a not-too-bad approximation for lots of databases, and changing it would be a big breaking change

### Can we even update `PostgresSQL` dialect

Presumably people are relying on the current logic, will you even accept a PR to update `PostgreSqlDialect` even if it's wrong now?

Or should a create a `RecentPostgreSqlDialect` as a new dialect?

### What about operators that don't work in some dialects?

Postgres doesn't actually support `XOR`, but there are tests asserting that it works, do you want to keep supporting it with an invented precedence, or cause an error?

### Can I move precedence numbers to the `Dialect` trait?

The `Precedence` enum shown here is pretty ugly, would you mind me moving the precedence stuff, including the default implementation of `get_next_precedence` to the `Dialect` trait?

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -44,10 +44,10 @@ pub use self::postgresql::PostgreSqlDialect;
 pub use self::redshift::RedshiftSqlDialect;
 pub use self::snowflake::SnowflakeDialect;
 pub use self::sqlite::SQLiteDialect;
-pub use crate::keywords;
 use crate::ast::{Expr, Statement};
-use crate::parser::{Parser, ParserError};
+pub use crate::keywords;
 use crate::keywords::Keyword;
+use crate::parser::{Parser, ParserError};
 use crate::tokenizer::Token;
 
 #[cfg(not(feature = "std"))]
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -44,10 +44,13 @@ pub use self::redshift::RedshiftSqlDialect;
 pub use self::snowflake::SnowflakeDialect;
 pub use self::sqlite::SQLiteDialect;
 pub use crate::keywords;
-use crate::parser::{Parser, ParserError, Precedence};
+use crate::parser::{Parser, ParserError};
 
+use crate::keywords::Keyword;
+use crate::tokenizer::Token;
 #[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
+use log::debug;
 
 /// Convenience check if a [`Parser`] uses a certain dialect.
 ///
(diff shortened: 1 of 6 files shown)


samuelcolvin:
The above questions stand, but I think this is in a state where it could at least be reviewed. @alamb could you kick off tests.

I'm going to use it in Logfire via a git ref now.

src/dialect/mod.rs
```diff
@@ -300,13 +304,172 @@ pub trait Dialect: Debug + Any {
         // return None to fall back to the default behavior
         None
     }
+
+    /// Get the precedence of the next token. This "full" method means all precedence logic and remain
```
samuelcolvin, line 308: This is moved to `Dialect` so dialects can change the precedence without all the numeric constants being shared between modules.

it's implemented as a separate function from `get_next_precedence`:
1. To avoid breaking the API
2. To allow dialects to customise precedence but optionally fall back to the default behaviour, as demonstrated `Snowflake`.

src/dialect/mod.rs
```diff
@@ -300,13 +304,172 @@ pub trait Dialect: Debug + Any {
...
+
     /// Dialect-specific statement parser override
     fn parse_statement(&self, _parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         // return None to fall back to the default behavior
         None
     }
+
+    /// The following precedence values are used directly by `Parse` or in dialects,
+    /// so have to be made public by the dialect.
+    fn prec_double_colon(&self) -> u8 {
```
samuelcolvin, line 418: these are the precedence values that either `Parser` needs to know about, or are used by implementations of the trait, e.g. `prec_double_colon` is used by `Snowflake`.

src/dialect/mod.rs
```diff
@@ -300,13 +304,172 @@ pub trait Dialect: Debug + Any {
...
+    }
+
+    fn prec_unknown(&self) -> u8 {
+        UNKNOWN_PREC
+    }
 }
 
+// Define the lexical Precedence of operators.
+//
+// Uses (APPROXIMATELY) <https://www.postgresql.org/docs/7.0/operators.htm#AEN2026> as a reference
```
samuelcolvin, line 449: This statement really isn't true, hence I added `APPROXIMATELY`.

We could rewrite to "was originally inspired by" or something?

src/dialect/postgresql.rs
```diff
@@ -20,6 +21,23 @@ use crate::tokenizer::Token;
...
 
+const DOUBLE_COLON_PREC: u8 = 140;
+const BRACKET_PREC: u8 = 130;
+const COLLATE_PREC: u8 = 120;
+const AT_TZ_PREC: u8 = 110;
+const CARET_PREC: u8 = 100;
+const MUL_DIV_MOD_OP_PREC: u8 = 90;
+const PLUS_MINUS_PREC: u8 = 80;
+// there's no XOR operator in PostgreSQL, but support it here to avoid breaking tests
+const XOR_PREC: u8 = 75;
```
samuelcolvin, line 32: I'm not sure what to do about this, if we remove `XOR` logic from Postgres, one tests fails, but I guess people might well be using it.

alamb:
I started the tests @samuelcolvin 

I haven't had a chance to review this PR yet, but the basic idea of chaning to follow postgres more closely, but giving people the ability to use the old behavior if they want (via a custom Dialect trait), makes sense to me.

Maybe @iffyio  @jmhain  or @lovasoa  have an opinion they would like to share

samuelcolvin:
Currently it's the other way around: default behaviour is unchanged, `PostgreSQL` dialect us changed to match modern postgres, happy to change.

alamb:
Thanks for the ping @samuelcolvin  -- reviewing now

src/dialect/mod.rs
alamb: I think this is fine

src/dialect/postgresql.rs
alamb: yeah I think we should just leave it in unless there is a compelling reason to take it out

alamb (approved):
Thank you @samuelcolvin  -- I think this looks great to me. I have some ideas on how to improve the documentation, but will make those in a follow on PR

I also updated the PR description to try and highlight the movement of precedence to the `Dialect`

tests/sqlparser_postgres.rs
```diff
@@ -4476,3 +4476,115 @@ fn test_unicode_string_literal() {
...
+                    left: Box::new(Expr::BinaryOp {
+                        left: Box::new(Expr::Identifier(Ident {
+                            value: "foo".to_string(),
+                            quote_style: None,
+                        })),
+                        op: arrow_operator,
+                        right: Box::new(Expr::Value(Value::SingleQuotedString("bar".to_string()))),
+                    }),
+                    op: BinaryOperator::Eq,
+                    right: Box::new(Expr::Value(Value::SingleQuotedString("spam".to_string()))),
```
lovasoa, line 4498: You could avoid the verbosity by using `verified_expr` instead of `verified_stmt` here

src/dialect/postgresql.rs
```diff
@@ -67,6 +85,102 @@ impl Dialect for PostgreSqlDialect {
...
+            | Token::StringConcat
+            | Token::Sharp
+            | Token::ShiftRight
+            | Token::ShiftLeft
+            | Token::Pipe
+            | Token::Ampersand
+            | Token::CustomBinaryOperator(_) => PG_OTHER_PREC,
+            _ => self.prec_unknown(),
+        };
+        Some(Ok(precedence))
```
lovasoa, line 181: I fear a little bit for the added maintenance burden here. What do you think @alamb ?

There are still many SQL tokens to add to sqlparser, and there is a chance people will add them only in the generic dialect, leaving postgres silently broken... Isn't there ?

lovasoa (changes requested):
sorry for the late review

src/dialect/postgresql.rs
alamb: I do think dialect divergence is a possibly and likely made more so by this PR 

In fact, in some ways the point of this PR is to start to diverge the dialects with more accurate precedence rules 🤔 

The only way I know to  guard against this divergence  is with test coverage for all new features (which we are pretty good about in general). 

Do you have any other thoughts about how to improve the situation @lovasoa ?

tests/sqlparser_postgres.rs
alamb: It is a great idea -- I made a PR with this change: https://github.com/sqlparser-rs/sqlparser-rs/pull/1367

src/dialect/postgresql.rs
lovasoa: I think we could change postgres' `get_next_precedence` to just call the generic dialect's `get_next_precedence` and "fixing up" the result. This way, we remove the duplication, and when we change the generic dialect, the postgres version reflects the changes automatically.

lovasoa: Something like 

```rust
fn get_next_precedence(&self, parser: &Parser) -> Option<Result<u8, ParserError>> {
    if some_particular_case { return some_particular_precedence }
    GenericDialect.get_next_precedence(parser)
}
```

samuelcolvin: @lovasoa I agree it's unfortunate, there's so much duplication in the the postgres dialect.

However your `GenericDialect.get_next_precedence(parser)` solution don't help, we can already return `None` from `get_next_precedence` to fallback to the default behaviour.

---

The reason the Postgres dialect is completely reimplementing the logic is that it has to change the precedence values for virtually every token to match Postgres.

We can either:
* implement ~15 more methods on the trait to get every single precedence value
* or, add another abstraction on top of `Token` and replace all the `prec_*` methods with a single `get_prec(&self, precedence_level: PrecedenceLevel)`

These would avoid the need to duplicate the logic in `Dialect::get_next_precedence_default`, but would introduce (albeit less) duplication elsewhere.

samuelcolvin: Implemented in #1379.

src/dialect/postgresql.rs
```diff
@@ -67,6 +85,102 @@ impl Dialect for PostgreSqlDialect {
         )
     }
 
+    fn get_next_precedence(&self, parser: &Parser) -> Option<Result<u8, ParserError>> {
```
lovasoa, line 88: @alamb @samuelcolvin : wouldn't it be better to have a `Result<Option<u8>>` instead of an `Option<Result<u8>>`

samuelcolvin: that would:
* differ from the signature of `Iterator::next()`
* be a breaking change to the API

alamb: I also think the `transpose()` method can be used to convert `Option<Result<..>>` to `Result<Option<..>>` 

https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose

I don't have any particular preference for one over the other

PR #1361, opened 2024-08-02
Support parsing empty map literal syntax for DuckDB and Genric

goldmedal:
# Description
DuckDB supports creating an empty map like
```
D select map {};
┌────────────────────────────────────────────────┐
│ main.map(main.list_value(), main.list_value()) │
│             map(integer, integer)              │
├────────────────────────────────────────────────┤
│ {}                                             │
└────────────────────────────────────────────────┘
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3473,8 +3473,8 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a comma-separated list of 0+ items accepted by `F`
-    /// - [trailing_commas]: support trailing_commas or not
-    /// - [end_token]: expected end token for the closure (e.g. [Token::RParen], [Token::RBrace] ...)
+    /// * `trailing_commas` - support trailing_commas or not
+    /// * `end_token` - expected end token for the closure (e.g. [Token::RParen], [Token::RBrace] ...)
     pub fn parse_comma_separated0<T, F>(
         &mut self,
         f: F,
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2939,7 +2939,7 @@ impl<'a> Parser<'a> {
             Expr::InList {
                 expr: Box::new(expr),
                 list: if self.dialect.supports_in_empty_list() {
-                    self.parse_comma_separated0(Parser::parse_expr)?
+                    self.parse_comma_separated0(Parser::parse_expr, self.options.trailing_commas, Token::RParen)?
                 } else {
                     self.parse_comma_separated(Parser::parse_expr)?
                 },
@@ -3481,17 +3481,18 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a comma-separated list of 0+ items accepted by `F`
-    pub fn parse_comma_separated0<T, F>(&mut self, f: F) -> Result<Vec<T>, ParserError>
+    /// - [trailing_commas]: support trailing_commas or not
+    /// - [end_token]: expected end token for the closure (e.g. [Token::RParen], [Token::RBrace] ...)
+    pub fn parse_comma_separated0<T, F>(&mut self, f: F, trailing_commas: bool, end_token: Token) -> Result<Vec<T>, ParserError>
     where
         F: FnMut(&mut Parser<'a>) -> Result<T, ParserError>,
     {
-        // ()
-        if matches!(self.peek_token().token, Token::RParen) {
+        if self.peek_token().token == end_token {
             return Ok(vec![]);
         }
-        // (,)
-        if self.options.trailing_commas
-            && matches!(self.peek_tokens(), [Token::Comma, Token::RParen])
+
+        if trailing_commas
+            && self.peek_tokens() == [Token::Comma, end_token]
         {
             let _ = self.consume_token(&Token::Comma);
             return Ok(vec![]);
@@ -4061,7 +4062,7 @@ impl<'a> Parser<'a> {
                 })
             };
         self.expect_token(&Token::LParen)?;
-        let args = self.parse_comma_separated0(parse_function_param)?;
+        let args = self.parse_comma_separated0(parse_function_param, self.options.trailing_commas, Token::RParen)?;
         self.expect_token(&Token::RParen)?;
 
         let return_type = if self.parse_keyword(Keyword::RETURNS) {
@@ -10715,7 +10716,7 @@ impl<'a> Parser<'a> {
         }
 
         if self.consume_token(&Token::LParen) {
-            let interpolations = self.parse_comma_separated0(|p| p.parse_interpolation())?;
+            let interpolations = self.parse_comma_separated0(|p| p.parse_interpolation(), self.options.trailing_commas, Token::RParen)?;
             self.expect_token(&Token::RParen)?;
             // INTERPOLATE () and INTERPOLATE ( ... ) variants
             return Ok(Some(Interpolate {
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

src/parser/mod.rs
```diff
@@ -2352,12 +2352,14 @@ impl<'a> Parser<'a> {
...
     fn parse_duckdb_map_literal(&mut self) -> Result<Expr, ParserError> {
         self.expect_token(&Token::LBrace)?;
-
-        let fields = self.parse_comma_separated(Self::parse_duckdb_map_field)?;
-
-        self.expect_token(&Token::RBrace)?;
-
-        Ok(Expr::Map(Map { entries: fields }))
+        if self.peek_token().token == Token::RBrace {
+            let _ = self.next_token(); // consume }
```
dharanad, line 2356: Nit: ambiguous comments

goldmedal: I followed the same style as parsing the array literal. Should I enhance them?
https://github.com/sqlparser-rs/sqlparser-rs/blob/d49acc67b13e1d68f2e6a25546161a68e165da4f/src/parser/mod.rs#L1853

dharanad: My bad, i interpreted wrongly. This look good.

dharanad (approved):

src/parser/mod.rs
```diff
@@ -2352,12 +2352,14 @@ impl<'a> Parser<'a> {
     /// [map]: https://duckdb.org/docs/sql/data_types/map.html#creating-maps
     fn parse_duckdb_map_literal(&mut self) -> Result<Expr, ParserError> {
         self.expect_token(&Token::LBrace)?;
-
-        let fields = self.parse_comma_separated(Self::parse_duckdb_map_field)?;
-
-        self.expect_token(&Token::RBrace)?;
-
-        Ok(Expr::Map(Map { entries: fields }))
+        if self.peek_token().token == Token::RBrace {
```
iffyio, line 2355: is it possible to switch to `let fields = self.parse_comma_separated0(Self::parse_duckdb_map_field)?;` instead? if so that would let us skip the if/else here

goldmedal: I don't think we can do it. It seems that `parse_comma_separated0` is for parentheses, but MAP uses braces.
https://github.com/sqlparser-rs/sqlparser-rs/blob/d49acc67b13e1d68f2e6a25546161a68e165da4f/src/parser/mod.rs#L3487

iffyio: Ah i see, I'm thinking we could still reuse the logic (thinking its currently a bit odd that `parse_comma_separated0` is hardcoded to parenthesis while `parse_comma_separated` is token agnostic as it should be). Seems there's ony a couple of usages of `parse_comma_separated0`, we could change it to something like `parse_comma_separated0(|| {}, Token::RParen)`?

goldmedal: I addressed this in 0763f52 and 5d81654. I share the same logic for `brace`, `parenthesis`, and `bracket`. I think it looks better. Thanks.

src/parser/mod.rs
```diff
@@ -3479,18 +3473,22 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a comma-separated list of 0+ items accepted by `F`
-    pub fn parse_comma_separated0<T, F>(&mut self, f: F) -> Result<Vec<T>, ParserError>
+    /// * `trailing_commas` - support trailing_commas or not
+    /// * `end_token` - expected end token for the closure (e.g. [Token::RParen], [Token::RBrace] ...)
+    pub fn parse_comma_separated0<T, F>(
+        &mut self,
+        f: F,
+        trailing_commas: bool,
```
goldmedal, line 3481: I also extract this argument because this option isn't shared by different end_token in a dialect. For example, DuckDB supports the trailing_commas for the parenthesis.
```
select * from t where a in (1,2,);
```
However, it doesn't support the trailing_commas for the braces.
```
D select {,};
Parser Error: syntax error at or near ","
LINE 1: select {,};
```

src/parser/mod.rs
```diff
@@ -1849,17 +1849,9 @@ impl<'a> Parser<'a> {
...
-            Ok(Expr::Array(Array {
-                elem: vec![],
-                named,
-            }))
-        } else {
-            let exprs = self.parse_comma_separated(Parser::parse_expr)?;
-            self.expect_token(&Token::RBracket)?;
-            Ok(Expr::Array(Array { elem: exprs, named }))
-        }
+        let exprs = self.parse_comma_separated0(Parser::parse_expr, false, Token::RBracket)?;
```
iffyio, line 1852: do we know if this change is going to be correct for all dialects? Otherwise we can probably keep the behavior to match the parser configuration

src/parser/mod.rs
iffyio: hmm I'm not too sure if pulling out the trailing_comma option is worth it/desirable - the current behavior isn't specified on the dialect level, but rather on the parser itself, to be overridden by the user, and I wonder if it'll get confusing to have to figure out where each dialect supports or doesnt support trailing commas. Would it be okay to keep the behavior as it was before? (i.e. always guarding internally on `self.options.trailing_comma`)

goldmedal: > hmm I'm not too sure if pulling out the trailing_comma option is worth it/desirable - the current behavior isn't specified on the dialect level, but rather on the parser itself, to be overridden by the user, and I wonder if it'll get confusing to have to figure out where each dialect supports or doesnt support trailing commas. Would it be okay to keep the behavior as it was before? (i.e. always guarding internally on `self.options.trailing_comma`)

I see. I tried more cases for DuckDB, I found the trailing_comma behaviors are shared.
```sql
D select * from t where a in (1,2,);
┌────────┐
│   a    │
│ int32  │
├────────┤
│ 0 rows │
└────────┘
D select {'a':1,};
┌──────────────────────────┐
│ main.struct_pack(a := 1) │
│    struct(a integer)     │
├──────────────────────────┤
│ {'a': 1}                 │
└──────────────────────────┘
D select [1,2,];
┌───────────────────────┐
│ main.list_value(1, 2) │
│        int32[]        │
├───────────────────────┤
│ [1, 2]                │
└───────────────────────┘
```
I think we can use `self.options.trailing_comma` internally.

src/parser/mod.rs
goldmedal: I'm not sure but all the tests are passed. The logic is the same as what I did for the map literal (actually, I followed it to implement the empty map). Should I roll back it?

iffyio (approved):
LGTM! cc @alamb

src/parser/mod.rs
```diff
@@ -3479,18 +3468,20 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a comma-separated list of 0+ items accepted by `F`
-    pub fn parse_comma_separated0<T, F>(&mut self, f: F) -> Result<Vec<T>, ParserError>
+    /// * `end_token` - expected end token for the closure (e.g. [Token::RParen], [Token::RBrace] ...)
+    pub fn parse_comma_separated0<T, F>(
+        &mut self,
+        f: F,
+        end_token: Token,
```
alamb, line 3475: ❤️  that is a nice generalization

alamb (approved):
Thanks @goldmedal  and @iffyio

alamb:
(and @git-hulk  and @dharanad 🤯 )

PR #1363, opened 2024-08-05
Update version of GitHub Actions

git-hulk:
This PR updates the version of the following actions:

- hecrj/setup-rust-action from v1 to v2
- actions/checkout from v2 to v4

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,31 +8,31 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Set up Rust
-        uses: hecrj/setup-rust-action@v1
+        uses: hecrj/setup-rust-action@v2
         with:
           components: rustfmt
           # Note that `nightly` is required for `license_template_path`, as
           # it's an unstable feature.
           rust-version: nightly
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


iffyio (approved):
LGTM! cc @alamb

.github/workflows/rust.yml
```diff
@@ -8,31 +8,31 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Set up Rust
-        uses: hecrj/setup-rust-action@v1
+        uses: hecrj/setup-rust-action@v2
```
alamb, line 11: I am a little worried about this action as it doesn't look very actively maintained 

https://github.com/hecrj/setup-rust-action/issues

However, this is probably ok -- maybe we can find time to setup rust ourselves (using `rustup install`)

Here is how we do it in DataFusion: https://github.com/apache/datafusion/blob/main/.github/actions/setup-builder/action.yaml

alamb (approved):
Thanks @iffyio  and @git-hulk

.github/workflows/rust.yml
git-hulk: @alamb Thanks for your review and comment. I will be happy to do this when I get time.

PR #1366, opened 2024-08-06
Improve comments on `Dialect`

alamb:
While reviewing @samuelcolvin 's PR on https://github.com/sqlparser-rs/sqlparser-rs/pull/1360 I noticed several areas of the `Dialect` documentation that could be improved. 

This PR updates the documentation for DIalect

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -66,7 +66,8 @@ macro_rules! dialect_of {
 /// Encapsulates the differences between SQL implementations.
 ///
 /// # SQL Dialects
-/// SQL implementations deviatiate from one another, either due to
+///
+/// SQL implementations deviate from one another, either due to
 /// custom extensions or various historical reasons. This trait
 /// encapsulates the parsing differences between dialects.
 ///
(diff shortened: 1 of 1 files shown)


src/dialect/mod.rs
```diff
@@ -114,16 +115,20 @@ pub trait Dialect: Debug + Any {
     fn is_delimited_identifier_start(&self, ch: char) -> bool {
         ch == '"' || ch == '`'
     }
+
```
alamb, line 118: the use of whitespace between functions was inconsistent -- sometimes there was a space and sometimes not. Thus I made it all consistent with my personal preference (put a space between functions0

lovasoa (approved):

src/dialect/mod.rs
samuelcolvin: 👍 I really wish `rustfmt` would fix this to be consistent.

src/dialect/mod.rs
```diff
@@ -299,24 +330,33 @@ pub trait Dialect: Debug + Any {
...
+    /// If `None` is returned, falls back to the default behavior.
     fn get_next_precedence(&self, _parser: &Parser) -> Option<Result<u8, ParserError>> {
         // return None to fall back to the default behavior
         None
     }
 
-    /// Get the precedence of the next token. This "full" method means all precedence logic and remain
-    /// in the dialect. while still allowing overriding the `get_next_precedence` method with the option to
-    /// fallback to the default behavior.
+    /// Get the precedence of the next token, looking at the full token stream.
```
samuelcolvin, line 344: Thanks for fixing this, sorry my comment was so broken.

I'm not sure "full token stream" is really the right distinction, my name was probably poor.

If we didn't might breaking the API, we could probably call the above `get_next_precedence_opt` - the only real difference is that it returns an `Option` so can fallback to the default behaviour.

Or we could call this method `get_next_precedence_default`.

alamb: We haven't yet released this PR so we can change the name without any API breakages yet 

Any chance you can make a PR with a  proposed name change?

alamb: @samuelcolvin  made a PR here: https://github.com/sqlparser-rs/sqlparser-rs/pull/1378

PR #1368, opened 2024-08-07
minor: Implement common traits for OneOrManyWithParens

gstvg:
This allows accessing all inner values without matching

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -994,7 +994,26 @@ impl fmt::Display for LambdaFunction {
 
 /// Encapsulates the common pattern in SQL where either one unparenthesized item
 /// such as an identifier or expression is permitted, or multiple of the same
-/// item in a parenthesized list.
+/// item in a parenthesized list. For accessing items regardless of the form,
+/// `OneOrManyWithParens` implements `Deref<Target = [T]>` and `IntoIterator`,
+/// so you can call slice methods on it and iterate over items
+/// # Examples
+/// Acessing as a slice:
+/// ```
+/// # use sqlparser::ast::OneOrManyWithParens;
+/// let one = OneOrManyWithParens::One("a");
+///
+/// assert_eq!(one[0], "a");
+/// assert_eq!(one.len(), 1);
+/// ```
+/// Iterating:
+/// ```
+/// # use sqlparser::ast::OneOrManyWithParens;
+/// let one = OneOrManyWithParens::One("a");
+/// let many = OneOrManyWithParens::Many(vec!["a", "b"]);
+///
+/// assert_eq!(one.into_iter().chain(many).collect::<Vec<_>>(), vec!["a", "a", "b"] );
+/// ```
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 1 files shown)


jmhain (commented):
lgtm, cc @alamb

jmhain (approved):

alamb (commented):
Thanks @gstvg  -- can you please add a test showing what this contribution allows? That way we won't accidentally break it in the future

alamb:

gstvg:
> Thanks @gstvg -- can you please add a test showing what this contribution allows? That way we won't accidentally break it in the future

Sure, it makes sense, so much so that while I was writing it, I realized it wasn't enough for my intended usage 😅
I ended up implementing 3 more traits
Thanks @jmhain and @alamb

src/ast/mod.rs
```diff
@@ -1004,6 +1005,51 @@ pub enum OneOrManyWithParens<T> {
...
+    #[cfg(feature = "std")]
+    type IntoIter = core::iter::Chain<core::option::IntoIter<T>, std::vec::IntoIter<T>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let (one, many) = match self {
+            OneOrManyWithParens::One(one) => (Some(one), Vec::new()),
+            OneOrManyWithParens::Many(many) => (None, many),
+        };
+
+        one.into_iter().chain(many)
```
gstvg, line 1049: This is a bit strange, it could be 
1) `vec![one]` and just return `Vec::into_iter`, but it allocates, or
2) create and implement a new iterator, but it's much more work, and we won't get for free additional iterator traits impls.

Just let me know if other approach is preferable

alamb: > create and implement a new iterator, but it's much more work, and we won't get for free additional iterator traits impls.

this is what i would have expected. 

Not sure what you mean "won't get for free additional iterator traits impl"

Another potential approach is returning a `Box`ed iterator

src/ast/mod.rs
```diff
@@ -1004,6 +1005,51 @@ pub enum OneOrManyWithParens<T> {
     Many(Vec<T>),
 }
 
+impl<T> Deref for OneOrManyWithParens<T> {
```
alamb, line 1008: This is looking good -- I think if we are going to add this code we should also add examples to the documentation f `OneOrManyWithParens` otherwise it will likely be hard to find

So like 
```rust 
/// # Example: accessing as a slice:
/// ...
```

```rust 
/// # Example: iterating
/// ...
```

alamb (commented):
Thanks @gstvg  -- this is looking close

src/ast/mod.rs
gstvg: Done!

src/ast/mod.rs
gstvg: It wasn't too much work after all :grimacing: 

> Not sure what you mean "won't get for free additional iterator traits impl"

It's actually just `DoubleEndedIterator`, I thought there were more

alamb (approved):
This looks good to me -- thank you @gstvg  and I apologize for the delay in reviewing

alamb:
Thanks again @gstvg

gstvg:
Thanks @jmhain and @alamb :pray: 

> This looks good to me -- thank you @gstvg and I apologize for the delay in reviewing

It's okay, don't worry. If I ever need something to be reviewed with priority, I will ping you or another maintainer

PR #1372, opened 2024-08-09
Support DuckDB struct syntax and support list of struct syntax

jayzhan211:
Close https://github.com/sqlparser-rs/sqlparser-rs/issues/1245

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -519,7 +519,7 @@ impl fmt::Display for DataType {
                         StructBracketKind::Parentheses => {
                             write!(f, "STRUCT({})", display_comma_separated(fields))
                         }
-                        StructBracketKind::AngleBrakets => {
+                        StructBracketKind::AngleBrackets => {
                             write!(f, "STRUCT<{}>", display_comma_separated(fields))
                         }
                     }
@@ -625,7 +625,7 @@ fn format_clickhouse_datetime_precision_and_timezone(
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum StructBracketKind {
     Parentheses,
-    AngleBrakets,
+    AngleBrackets,
 }
 
 /// Timestamp and Time data types information about TimeZone formatting.
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -302,7 +302,7 @@ pub enum DataType {
     ///
     /// [hive]: https://docs.cloudera.com/cdw-runtime/cloud/impala-sql-reference/topics/impala-struct.html
     /// [bigquery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type
-    Struct(Vec<StructField>),
+    Struct(Vec<StructField>, StructBracketKind),
     /// Union
     ///
     /// [duckdb]: https://duckdb.org/docs/sql/data_types/union.html
@@ -513,9 +513,16 @@ impl fmt::Display for DataType {
                 }
                 write!(f, ")")
             }
-            DataType::Struct(fields) => {
+            DataType::Struct(fields, bracket) => {
                 if !fields.is_empty() {
-                    write!(f, "STRUCT<{}>", display_comma_separated(fields))
+                    match bracket {
+                        StructBracketKind::Parentheses => {
+                            write!(f, "STRUCT({})", display_comma_separated(fields))
+                        }
+                        StructBracketKind::AngleBrakets => {
+                            write!(f, "STRUCT<{}>", display_comma_separated(fields))
+                        }
+                    }
                 } else {
                     write!(f, "STRUCT")
                 }
@@ -611,6 +618,14 @@ fn format_clickhouse_datetime_precision_and_timezone(
     write!(f, ")")?;
 
     Ok(())
+}
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum StructBracketKind {
+    Parentheses,
+    AngleBrakets,
 }
 
 /// Timestamp and Time data types information about TimeZone formatting.
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -620,11 +620,14 @@ fn format_clickhouse_datetime_precision_and_timezone(
     Ok(())
 }
 
+/// Type of brackets used for `STRUCT` literals.
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum StructBracketKind {
+    /// Example: `STRUCT(a INT, b STRING)`
     Parentheses,
+    /// Example: `STRUCT<a INT, b STRING>`
     AngleBrackets,
 }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7258,7 +7258,7 @@ impl<'a> Parser<'a> {
                     trailing_bracket = _trailing_bracket;
                     Ok(DataType::Struct(
                         field_defs,
-                        StructBracketKind::AngleBrakets,
+                        StructBracketKind::AngleBrackets,
                     ))
                 }
                 Keyword::UNION if dialect_of!(self is DuckDbDialect | GenericDialect) => {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1074,7 +1074,7 @@ impl<'a> Parser<'a> {
                 Keyword::MATCH if dialect_of!(self is MySqlDialect | GenericDialect) => {
                     self.parse_match_against()
                 }
-                Keyword::STRUCT if dialect_of!(self is BigQueryDialect | GenericDialect ) => {
+                Keyword::STRUCT if dialect_of!(self is BigQueryDialect | GenericDialect) => {
                     self.prev_token();
                     self.parse_bigquery_struct_literal()
                 }
@@ -2231,7 +2231,11 @@ impl<'a> Parser<'a> {
             if token == Token::Lt {
                 self.expect_closing_angle_bracket(trailing_bracket)?
             } else {
-                self.expect_closing_right_paren(trailing_bracket)?
+                if !trailing_bracket.0 {
+                    self.expect_token(&Token::RParen)?;
+                };
+
+                false.into()
             },
         ))
     }
@@ -2441,25 +2445,6 @@ impl<'a> Parser<'a> {
                     true.into()
                 }
                 _ => return self.expected(">", self.peek_token()),
-            }
-        } else {
-            false.into()
-        };
-
-        Ok(trailing_bracket)
-    }
-
-    fn expect_closing_right_paren(
-        &mut self,
-        trailing_bracket: MatchedTrailingBracket,
-    ) -> Result<MatchedTrailingBracket, ParserError> {
-        let trailing_bracket = if !trailing_bracket.0 {
-            match self.peek_token().token {
-                Token::RParen => {
-                    self.next_token();
-                    false.into()
-                }
-                _ => return self.expected(")", self.peek_token()),
             }
         } else {
             false.into()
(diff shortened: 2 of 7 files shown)


src/parser/mod.rs
```diff
@@ -1074,10 +1074,14 @@ impl<'a> Parser<'a> {
                 Keyword::MATCH if dialect_of!(self is MySqlDialect | GenericDialect) => {
                     self.parse_match_against()
                 }
-                Keyword::STRUCT if dialect_of!(self is BigQueryDialect | GenericDialect) => {
+                Keyword::STRUCT if dialect_of!(self is BigQueryDialect | GenericDialect ) => {
```
git-hulk, line 1077: Can you revert this line?

src/parser/mod.rs
```diff
@@ -1074,10 +1074,14 @@ impl<'a> Parser<'a> {
...
                     self.prev_token();
                     self.parse_bigquery_struct_literal()
                 }
+                Keyword::STRUCT if dialect_of!(self is DuckDbDialect ) => {
```
git-hulk, line 1081:
```suggestion
                Keyword::STRUCT if dialect_of!(self is DuckDbDialect) => {
```

src/parser/mod.rs
```diff
@@ -2444,6 +2470,25 @@ impl<'a> Parser<'a> {
...
+        trailing_bracket: MatchedTrailingBracket,
+    ) -> Result<MatchedTrailingBracket, ParserError> {
+        let trailing_bracket = if !trailing_bracket.0 {
+            match self.peek_token().token {
+                Token::RParen => {
+                    self.next_token();
+                    false.into()
+                }
+                _ => return self.expected(")", self.peek_token()),
+            }
```
git-hulk, line 2484:
```suggestion
            self.expect_token(&Token::RParen)?;
            false.into()
```

src/parser/mod.rs
```diff
@@ -7229,10 +7274,18 @@ impl<'a> Parser<'a> {
                         ))))
                     }
                 }
+                Keyword::STRUCT if dialect_of!(self is DuckDbDialect) => {
+                    self.prev_token();
+                    // let field_defs=
```
git-hulk, line 7279:
```suggestion
```

tests/sqlparser_duckdb.rs
```diff
@@ -32,6 +32,27 @@ fn duckdb_and_generic() -> TestedDialects {
...
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, i INTEGER)[])"#;
+    let select = duckdb().parse_sql_statements(sql).unwrap().pop().unwrap();
+    assert_eq!(select.to_string(), canonical);
+
+    // nested struct
+    let canonical = r#"CREATE TABLE t1 (s STRUCT<v VARCHAR, s STRUCT<a1 INTEGER, a2 VARCHAR>>[])"#;
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, s STRUCT(a1 INTEGER, a2 VARCHAR))[])"#;
+    let select = duckdb().parse_sql_statements(sql).unwrap().pop().unwrap();
+    assert_eq!(select.to_string(), canonical);
+}
```
git-hulk, line 54: Is it possible to add some negative cases?

git-hulk:
cc @iffyio

git-hulk (approved):
LGTM, great.

jayzhan211:
Thanks for your review @git-hulk

src/parser/mod.rs
```diff
@@ -2227,7 +2228,15 @@ impl<'a> Parser<'a> {
 
         Ok((
             field_defs,
-            self.expect_closing_angle_bracket(trailing_bracket)?,
+            if token == Token::Lt {
```
iffyio, line 2231: There's a slight mismatch I'm thinking might not be worth having - the open delimeter token is configurable now but here the function still needs to figure out for itself what the value is for the closing token.
The function is a bit complicated on its own, specifically due to the `<` delimeter. then since duckdb syntax doesn't have that I'm thinking it makes sense to use a dedicated function that is simpler.

Something roughly like this I think should be all duckdb need, would it make sense to add this instead and have duckdb (and any future dialect sharing the same syntax) use it?
```rust
   fn parse_duck_db_struct_type_def<F>(
        &mut self,
    ) -> Result<Vec<StructField>, ParserError>
    {
        let start_token = self.peek_token();
        self.expect_keyword(Keyword::STRUCT)?;

        self.expect_token(Token::LParen)?;

        let field_defs = self.parse_comma_separated(Self::parse_struct_field_def)?;

        self.expect_token(Token::RParen)?;

        Ok(field_defs)
    }
```

jayzhan211: I have separate implementation at the beginning, but since they have quite many similarity so I merge them into once function.

git-hulk (approved):

tests/sqlparser_bigquery.rs
```diff
@@ -21,6 +21,16 @@ use sqlparser::dialect::{BigQueryDialect, GenericDialect};
...
 
+#[test]
+fn test_struct() {
+    // nested struct
+    let canonical = r#"CREATE TABLE t1 (s STRUCT<v VARCHAR, s STRUCT<a1 INTEGER, a2 VARCHAR>>[])"#;
+    let sql = r#"CREATE TABLE t1 (s STRUCT<v VARCHAR, s STRUCT<a1 INTEGER, a2 VARCHAR>>[])"#;
+    let select = bigquery().parse_sql_statements(sql).unwrap().pop().unwrap();
+    // TODO: '>>' is incorrect parsed in bigquery syntax
+    assert_ne!(select.to_string(), canonical);
+}
```
iffyio, line 32: not sure I follow if/why this test is needed, since bigquery doesn't support `[]` syntax for arrays?

src/parser/mod.rs
```diff
@@ -2231,6 +2231,18 @@ impl<'a> Parser<'a> {
         ))
     }
 
+    fn parse_duckdb_struct_type_def(&mut self) -> Result<Vec<StructField>, ParserError> {
+        self.expect_keyword(Keyword::STRUCT)?;
+        self.expect_token(&Token::LParen)?;
+        let field_defs = self
+            .parse_comma_separated(Self::parse_struct_field_def)?
+            .into_iter()
+            .map(|(f, _)| f)
```
iffyio, line 2240: Oh this looks like a bug for us to silently drop the returned flag since there might be an extraneous `>`, seems like we would instead do something like this
```rust
let field_defs = self.parse_comma_separated(|self| {
    Ok(StructField {
        field_name: Some(self.parse_identifier(false)?), // I assume field name isn't optional for duckdb?
        field_type: self.parse_data_type(),
    })

File: src/parser/mod.rs [hunk 9]
```diff
@@ -2231,6 +2231,18 @@ impl<'a> Parser<'a> {
         ))
     }
 
+    fn parse_duckdb_struct_type_def(&mut self) -> Result<Vec<StructField>, ParserError> {
```
iffyio, line 2234: Can we include a comment linking to the duckdb docs where the struct syntax is defined?

tests/sqlparser_duckdb.rs
```diff
@@ -32,6 +32,39 @@ fn duckdb_and_generic() -> TestedDialects {
     }
 }
 
+#[test]
+fn test_struct() {
+    // basic struct
+    let canonical = r#"CREATE TABLE t1 (s STRUCT<v VARCHAR, i INTEGER>)"#;
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, i INTEGER))"#;
```
iffyio, line 39: Oh I think this is incorrect, no need for the input and output to not match, in this case the output is significantly different syntax from what went into the parser. We can probably add a `comma_delimeter: bool`  or similar flag to the StructField struct so that it knows how to display accordingly?

jayzhan211: I'm not clear about this. Do you mean add flag so we have different canonical name for `struct<>` and `struct()`?

alamb:
I am hoping to [make a sqlparser release](https://github.com/sqlparser-rs/sqlparser-rs/issues/1349) soon (like tomorrow) -- do we think this PR is close to ready (aka should I hold the release for this PR?)

tests/sqlparser_duckdb.rs
iffyio: Ah yeah that's what I meant - my bad realised I wrote `comma_delimeter` but meant to write `parens_delimeter` instead. Yeah so like if that flag is true display uses `struct (...)` otherwise `struct<...>`

jayzhan211:
> fn parse_duckdb_struct_type_

@alamb You can go ahead first

jayzhan211:
> I am hoping to [make a sqlparser release](https://github.com/sqlparser-rs/sqlparser-rs/issues/1349) soon (like tomorrow) -- do we think this PR is close to ready (aka should I hold the release for this PR?)

It is close. We can release together with this PR

src/parser/mod.rs
```diff
@@ -2231,6 +2231,23 @@ impl<'a> Parser<'a> {
         ))
     }
 
+    /// Duckdb Struct Data Type <https://duckdb.org/docs/sql/data_types/struct.html#retrieving-from-structs>
+    fn parse_duckdb_struct_type_def(&mut self) -> Result<Vec<StructField>, ParserError> {
+        self.expect_keyword(Keyword::STRUCT)?;
+        self.expect_token(&Token::LParen)?;
+        let struct_body = self.parse_comma_separated(|parser| {
+            let field_name = parser.parse_identifier(false).ok();
```
iffyio, line 2239: it looks like we're silently dropping an error here, can we propagate the error as `let field_name = parser.parse_identifier(false)?;`?

tests/sqlparser_duckdb.rs
```diff
@@ -32,6 +32,36 @@ fn duckdb_and_generic() -> TestedDialects {
...
+    assert_eq!(statement.to_string(), sql);
+
+    // nested struct
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, s STRUCT(a1 INTEGER, a2 VARCHAR))[])"#;
+    let statement = duckdb().verified_stmt(sql);
+    assert_eq!(statement.to_string(), sql);
+
+    // failing test
+    let sql_list = vec![
+        r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, i INTEGER)))"#,
```
iffyio, line 54: we can include a test case that fails if no fieldname is provided? since that's not optional for this dialect

jayzhan211: added

tests/sqlparser_duckdb.rs
```diff
@@ -32,6 +32,38 @@ fn duckdb_and_generic() -> TestedDialects {
...
+
+    // struct array
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, i INTEGER)[])"#;
+    let statement = duckdb().verified_stmt(sql);
+    assert_eq!(statement.to_string(), sql);
+
+    // nested struct
+    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, s STRUCT(a1 INTEGER, a2 VARCHAR))[])"#;
+    let statement = duckdb().verified_stmt(sql);
+    assert_eq!(statement.to_string(), sql);
```
iffyio, line 50:
```suggestion
    let statement = duckdb().verified_stmt(sql);

    // struct array
    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, i INTEGER)[])"#;
    let statement = duckdb().verified_stmt(sql);

    // nested struct
    let sql = r#"CREATE TABLE t1 (s STRUCT(v VARCHAR, s STRUCT(a1 INTEGER, a2 VARCHAR))[])"#;
    let statement = duckdb().verified_stmt(sql);
```
we can remove the explicit string asserts, they're covered by [verified_stmt](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/test_utils.rs#L128-L130)

iffyio (approved):
One minor cleanup comment, otherwise LGTM! cc @alamb

alamb:
I am working on updating the tests and adding some small comments. I'll be done shortly

src/ast/data_type.rs
```diff
@@ -618,6 +625,17 @@ fn format_clickhouse_datetime_precision_and_timezone(
     Ok(())
 }
 
+/// Type of brackets used for `STRUCT` literals.
```
alamb, line 628: I also added some comments on this struct

alamb:
Thanks again @jayzhan211 and @iffyio and @git-hulk

alamb (approved):

jayzhan211:
Great! Thanks @iffyio @alamb and @git-hulk

PR #1374, opened 2024-08-10
supporting snowflake extract syntax

### LINKED ISSUE #1373: Support Snowflake-specific EXTRACT syntax
Per the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/functions/extract), `EXTRACT` functions can be written one of two ways:

```sql
SELECT
  EXTRACT(hour from foo)
FROM bar
```

```sql
SELECT 
  EXTRACT('hour', foo)
FROM bar
```
This will result in a parse error:
```console
ParseError(ParserError("Expected: date/time field, found: 'hour' at Line: 14, Column: 13"))
```

The `parse_extract_expr()` method on the `Parser` explicitly expects a `FROM` keyword:
https://github.com/sqlparser-rs/sqlparser-rs/blob/1e209d87415a5adfedccac8cee3e2860122e4acb/src/parser/mod.rs#L1682-L1692

I am happy to contribute and submit a PR if someone can point me in the direction of how to override this function _just_ for Snowflake.

seve-martinez:
Closes #1373

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -480,7 +480,7 @@ pub enum CastKind {
 /// `EXTRACT` syntax variants.
 ///
 /// In Snowflake dialect, the `EXTRACT` expression can support either the `from` syntax
-/// or the comma syntax. 
+/// or the comma syntax.
 ///
 /// See <https://docs.snowflake.com/en/sql-reference/functions/extract>
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -490,7 +490,7 @@ pub enum ExtractSyntax {
     /// `EXTRACT( <date_or_time_part> FROM <date_or_time_expr> )`
     From,
     /// `EXTRACT( <date_or_time_part> , <date_or_timestamp_expr> )`
-    Comma
+    Comma,
 }
 
 /// An SQL expression of any type.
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -1682,12 +1682,21 @@ impl<'a> Parser<'a> {
     pub fn parse_extract_expr(&mut self) -> Result<Expr, ParserError> {
         self.expect_token(&Token::LParen)?;
         let field = self.parse_date_time_field()?;
-        self.expect_keyword(Keyword::FROM)?;
+        
+        let syntax = if self.parse_keyword(Keyword::FROM) {
+            ExtractSyntax::From
+        } else if self.consume_token(&Token::Comma) {
```
iffyio, line 1688: The test case covers only snowflake/generic while functionality here is added for all dialects - if the behavior is intended for snowflake, we can add a dialect_of check here? (Otherwise we can move the test to common.rs to cover all dialects)

src/ast/mod.rs
```diff
@@ -477,6 +477,15 @@ pub enum CastKind {
     DoubleColon,
 }
 
+/// `EXTRACT` syntax types.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum ExtractSyntax {
+    From,
+    Comma
```
iffyio, line 486: Can we add the example syntax showing the difference between the two? (Also for the comma, lets add a link to the snowflake docs that mention the syntax)

seve-martinez:
@iffyio thank you for the review. I've updated the PR per your suggestions.

Snowflake also offers the option to pass a string to the `EXTRACT` method like so:
```sql
EXTRACT('hour', foo)
```

Do you have any recommendation on how to handle that case here?

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @seve-martinez  and @iffyio  for the review. Looks great to me

alamb:
I took the liberty of running `cargo fmt` and pushing a commit to this branch to get the CI passing

PR #1377, opened 2024-08-12
adding support for scale in CEIL and FLOOR functions

### LINKED ISSUE #1376: `CEIL` and `FLOOR` do not support comma syntax
The current implementation only supports `CEIL`/`FLOOR` calls when including `TO <DateTimeField>`. However this syntax appears to be rare and so far I've only found it supported in [Kinesis](https://docs.aws.amazon.com/kinesisanalytics/latest/sqlref/sql-reference-ceil.html), which isn't supported by `sqlparser` at all.

The majority of dialects support a single expression, or an expression and a scale separated by a comma:
```sql
SELECT CEIL(321.56, 1)

-- prints 321.6

SELECT CEIL(321.56)
-- prints 322
```

[Snowflake](https://docs.snowflake.com/en/sql-reference/functions/ceil)
[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#ceil)
[Hive](https://www.tutorialspoint.com/hive/hive_built_in_functions.htm#:~:text=2.0-,ceil()%20function,-hive%3E%20SELECT)
[ClickHouse](https://clickhouse.com/docs/en/sql-reference/functions/rounding-functions#ceiling)
[DuckDb](https://duckdb.org/docs/sql/functions/numeric.html#ceilx)
[MsSql](https://www.w3schools.com/sql/func_sqlserver_ceiling.asp)
[MySql](https://www.w3schools.com/sql/func_mysql_ceil.asp)
[Postgres](https://www.postgresqltutorial.com/postgresql-math-functions/postgresql-ceil/)
[Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_CEILING_FLOOR.html)
[SQLite](https://www.sqlite.org/lang_mathfunc.html#:~:text=arctangent%20of%20X.-,ceil(X)%0Aceiling(X),-Return%20the%20first)

The existing impl:
https://github.com/sqlparser-rs/sqlparser-rs/blob/1e209d87415a5adfedccac8cee3e2860122e4acb/src/parser/mod.rs#L1694-L1716

Because the `CEIL`/`FLOOR` [expressions](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L652-L662) expect a `DateTimeField`, I'm not sure it makes sense to just pass a numeric literal to this and force into a `Custom(Ident)` type.

Happy to open a PR for this, but wanted to understand the best way to handle `field` attribute here.

seve-martinez:
Closes #1376.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -493,6 +493,17 @@ pub enum ExtractSyntax {
     Comma,
 }
 
+/// The syntax used in a CEIL or FLOOR expression.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CeilFloorKind {
+    /// `CEIL( <expr> TO <DateTimeField>)`
+    DateTimeField(DateTimeField),
+    /// `CEIL( <expr> [, <scale>])`
+    Scale(Value),
+}
+
 /// An SQL expression of any type.
 ///
 /// The parser does not distinguish between expressions of different types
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -477,6 +477,17 @@ pub enum CastKind {
     DoubleColon,
 }
 
+/// The syntax used in a CEIL or FLOOR expression.
```
alamb, line 480: Whatcan you add the context about coming from kinesis and a link to the docs as well in these doc comments?

alamb (commented):
Thank you for this contribution  @seve-martinez

iffyio (approved):
LGTM!

alamb:
Thanks again @seve-martinez  and @iffyio

PR #1379, opened 2024-08-14
Simple custom lexical precedence in PostgreSQL dialect

samuelcolvin:
As suggest mentioned by @lovasoa in https://github.com/sqlparser-rs/sqlparser-rs/pull/1360#discussion_r1705403984 the previous approach had unnecessary duplication, this should reduce the duplication of logic while still allowing complete freedom for how dialects decide precedence.

This doesn't change the logic for `PostgreSqlDialect` except that `Token::DuckIntDiv` will now return `MulDivModOp` precedence, not `UNKOWN`.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -354,13 +354,18 @@ pub trait Dialect: Debug + Any {
         if let Some(precedence) = self.get_next_precedence(parser) {
             return precedence;
         }
+        macro_rules! p {
+            ($precedence:ident) => {
+                self.prec_value(Precedence::$precedence)
+            };
+        }
 
         let token = parser.peek_token();
         debug!("get_next_precedence_full() {:?}", token);
         match token.token {
-            Token::Word(w) if w.keyword == Keyword::OR => Ok(OR_PREC),
-            Token::Word(w) if w.keyword == Keyword::AND => Ok(AND_PREC),
-            Token::Word(w) if w.keyword == Keyword::XOR => Ok(XOR_PREC),
+            Token::Word(w) if w.keyword == Keyword::OR => Ok(p!(Or)),
+            Token::Word(w) if w.keyword == Keyword::AND => Ok(p!(And)),
+            Token::Word(w) if w.keyword == Keyword::XOR => Ok(p!(Xor)),
 
             Token::Word(w) if w.keyword == Keyword::AT => {
                 match (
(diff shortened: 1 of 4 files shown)


lovasoa (approved):
:tada:

src/dialect/mod.rs
```diff
@@ -457,88 +461,57 @@ pub trait Dialect: Debug + Any {
...
-
-    /// Return the precedence of the unary `NOT` operator.
-    ///
-    /// For example `NOT (a OR b)`
-    ///
-    /// Default is 15.
-    fn prec_unary_not(&self) -> u8 {
-        UNARY_NOT_PREC
+    /// Uses (APPROXIMATELY) <https://www.postgresql.org/docs/7.0/operators.htm#AEN2026> as a reference
+    fn prec_value(&self, prec: Precedence) -> u8 {
```
alamb, line 467: I think this is a good idea. 

My only potential concern with this approach is that it could be potentially slower, as it now has many more function calls, however I don't actually think we have a sqlparser benchmark so at this point so trying to make such a change would likely be a pre-mature optimization. 

If anyone needs additional performance, we can do that as follow on PRs

src/dialect/postgresql.rs
```diff
@@ -197,24 +134,25 @@ impl Dialect for PostgreSqlDialect {
...
-        BETWEEN_LIKE_PREC
-    }
-
-    fn prec_like(&self) -> u8 {
-        BETWEEN_LIKE_PREC
-    }
-
-    fn prec_unary_not(&self) -> u8 {
-        NOT_PREC
+    fn prec_value(&self, prec: Precedence) -> u8 {
```
alamb, line 137: ❤️

alamb (approved):
Thank you @samuelcolvin  -- this looks like an improvement to me. Thank you for the follow up

Since we haven't released the changes in https://github.com/sqlparser-rs/sqlparser-rs/pull/1360 to crates.io yet, this isn't an API change

I'll wait a day before merging this in case  @lovasoa  would like to review or comment on this PR as well

samuelcolvin:
thanks a lot.

samuelcolvin:
@alamb if we care about performance, we should stop using dynamic dispatch and make the parser generic around the dialect, with that we could make lots of these methods `const` (or drop the `Precedence` enum and just have const values on the trait) and probably improve performance significantly in general.

That would of course be a big change to the public API, but worthwhile IMHO.

alamb:
> @alamb if we care about performance, we should stop using dynamic dispatch and make the parser generic around the dialect, with that we could make lots of these methods `const` (or drop the `Precedence` enum and just have const values on the trait) and probably improve performance significantly in general.
> 
> That would of course be a big change to the public API, but worthwhile IMHO.

I agree with both points

alamb:
> In my use case (SQLPage), SQL parsing overhead is negligible. Have you ever been bottlenecked by sql parsing?

No. There is more discussion here as well: https://github.com/sqlparser-rs/sqlparser-rs/issues/1381#issuecomment-2291126604

PR #1387, opened 2024-08-16
feat: support different USE statement syntaxes

kacpermuda:
I'd like to add support for different syntaxes of the `USE` statement, used to set a default DATABASE/SCHEMA for set of queries.

Some implementations are calling their top level structure `CATALOG` (e.g. [Databricks](https://docs.databricks.com/en/data-governance/unity-catalog/index.html#object-hierarchy-in-the-metastore), [Trino](https://trino.io/docs/current/overview/concepts.html#catalog)), but most of them stick to `DATABASE` so I assumed that we should stick to naming: `top-level container = DATABASE`, `lower-level container = SCHEMA`. For the full identifier we end up with: `DATABASE.SCHEMA.TABLE`.

In some implementations `DATABASE` and `SCHEMA` are aliases, but usually these implementations have a two-layer structure (`DATABASE.TABLE`, without `SCHEMA`) - in that case we should keep `DATABASE` as `DATABASE` and leave `SCHEMA` empty.

There are different syntaxes of USE statement across dialects, but I implemented it like:
(use syntax => resolved db and schema names)
1. `USE x;` -> `DATABASE=x; SCHEMA=null;` (Databricks is adjusted to `DATABASE=null; SCHEMA=x;`)
2. `USE x.y;` -> `DATABASE=x; SCHEMA=y;`
3. `USE DATABASE x;` -> `DATABASE=x; SCHEMA=null;` (Databricks is adjusted to `DATABASE=null; SCHEMA=x;` In Databricks `DATABASE` is an alias for `SCHEMA` and they both mean lower level structure while the top level structure is called `CATALOG`)
4. `USE SCHEMA y;` -> `DATABASE=null; SCHEMA=y;`
5. `USE SCHEMA x.y;` -> `DATABASE=x; SCHEMA=y;` (Snowflake specific syntax, SCHEMA keyword is optional and should not influence the parsing of fully qualified name)
6. `USE CATALOG x;` -> `DATABASE=x; SCHEMA=null;` (Databricks-specific keyword: `CATALOG`)
7. `USE DEFAULT;` -> `DATABASE=null; SCHEMA=null;`  (Hive specific, means going back to default database set - we don't know what it is just by looking at the query.)

Docs about USE statement in different dialects:
- [MySQL 8.4 USE](https://dev.mysql.com/doc/refman/8.4/en/use.html)
- [MySQL 5.7 USE](https://dev.mysql.com/doc/refman/5.7/en/use.html)
- [Snowflake USE SCHEMA](https://docs.snowflake.com/en/sql-reference/sql/use-schema)
- [Snowflake USE DATABASE](https://docs.snowflake.com/en/sql-reference/sql/use-database)
- [SQL Server USE](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/use-transact-sql?view=sql-server-ver16)
- [ClickHouse USE](https://clickhouse.com/docs/en/sql-reference/statements/use)
- [DuckDB USE](https://duckdb.org/docs/sql/statements/use.html)
- [Hive USE](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-UseDatabase)
- [Databricks USE CATALOG](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-use-catalog.html)
- [Databricks USE SCHEMA](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-use-schema.html)
- [Databricks USE DATABASE](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-usedb.html)

Some comments:
- Bigquery, PostgreSQL, Redshift, SQLite do not allow USE statements;

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9225,63 +9225,31 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_use(&mut self) -> Result<Statement, ParserError> {
-        // What should be treated as keyword in given dialect
-        let allowed_keywords = if dialect_of!(self is HiveDialect) {
-            vec![Keyword::DEFAULT]
+        // Determine which keywords are recognized by the current dialect
+        let parsed_keyword = if dialect_of!(self is HiveDialect) {
+            // HiveDialect accepts USE DEFAULT; statement without any db specified
+            if self.parse_keyword(Keyword::DEFAULT) {
+                return Ok(Statement::Use(Use::Default));
+            }
+            None // HiveDialect doesn't expect any other specific keyword after `USE`
         } else if dialect_of!(self is DatabricksDialect) {
-            vec![Keyword::CATALOG, Keyword::DATABASE, Keyword::SCHEMA]
+            self.parse_one_of_keywords(&[Keyword::CATALOG, Keyword::DATABASE, Keyword::SCHEMA])
         } else if dialect_of!(self is SnowflakeDialect) {
-            vec![Keyword::DATABASE, Keyword::SCHEMA]
+            self.parse_one_of_keywords(&[Keyword::DATABASE, Keyword::SCHEMA, Keyword::WAREHOUSE])
         } else {
-            vec![]
-        };
-        let parsed_keyword = self.parse_one_of_keywords(&allowed_keywords);
-
-        // Hive dialect accepts USE DEFAULT; statement without any db specified
-        if dialect_of!(self is HiveDialect) && parsed_keyword == Some(Keyword::DEFAULT) {
-            return Ok(Statement::Use {
-                db_name: None,
-                schema_name: None,
-                keyword: Some("DEFAULT".to_string()),
-            });
-        }
-
-        // Parse the object name, which might be a single identifier or fully qualified name (e.g., x.y)
-        let parts = self.parse_object_name(false)?.0;
-        let (db_name, schema_name) = match parts.len() {
-            1 => {
-                // Single identifier found
-                if dialect_of!(self is DatabricksDialect) {
-                    if parsed_keyword == Some(Keyword::CATALOG) {
-                        // Databricks: CATALOG keyword provided, treat as database name
-                        (Some(parts[0].clone()), None)
-                    } else {
-                        // Databricks: DATABASE, SCHEMA or no keyword provided, treat as schema name
-                        (None, Some(parts[0].clone()))
-                    }
-                } else if dialect_of!(self is SnowflakeDialect)
-                    && parsed_keyword == Some(Keyword::SCHEMA)
-                {
-                    // Snowflake: SCHEMA keyword provided, treat as schema name
-                    (None, Some(parts[0].clone()))
-                } else {
-                    // Other dialects: treat as database name by default
-                    (Some(parts[0].clone()), None)
-                }
-            }
-            2 => (Some(parts[0].clone()), Some(parts[1].clone())),
-            _ => {
-                return Err(ParserError::ParserError(
-                    "Invalid format in the USE statement".to_string(),
-                ))
-            }
+            None // No specific keywords for other dialects, including GenericDialect
         };
 
-        Ok(Statement::Use {
-            db_name,
-            schema_name,
-            keyword: parsed_keyword.map(|kw| format!("{:?}", kw)),
-        })
+        let obj_name = self.parse_object_name(false)?;
+        let result = match parsed_keyword {
+            Some(Keyword::CATALOG) => Use::Catalog(obj_name),
+            Some(Keyword::DATABASE) => Use::Database(obj_name),
+            Some(Keyword::SCHEMA) => Use::Schema(obj_name),
+            Some(Keyword::WAREHOUSE) => Use::Warehouse(obj_name),
+            _ => Use::Object(obj_name),
+        };
+
+        Ok(Statement::Use(result))
     }
 
     pub fn parse_table_and_joins(&mut self) -> Result<TableWithJoins, ParserError> {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9225,8 +9225,63 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_use(&mut self) -> Result<Statement, ParserError> {
-        let db_name = self.parse_identifier(false)?;
-        Ok(Statement::Use { db_name })
+        // What should be treated as keyword in given dialect
+        let allowed_keywords = if dialect_of!(self is HiveDialect) {
+            vec![Keyword::DEFAULT]
+        } else if dialect_of!(self is DatabricksDialect) {
+            vec![Keyword::CATALOG, Keyword::DATABASE, Keyword::SCHEMA]
+        } else if dialect_of!(self is SnowflakeDialect) {
+            vec![Keyword::DATABASE, Keyword::SCHEMA]
+        } else {
+            vec![]
+        };
+        let parsed_keyword = self.parse_one_of_keywords(&allowed_keywords);
+
+        // Hive dialect accepts USE DEFAULT; statement without any db specified
+        if dialect_of!(self is HiveDialect) && parsed_keyword == Some(Keyword::DEFAULT) {
+            return Ok(Statement::Use {
+                db_name: None,
+                schema_name: None,
+                keyword: Some("DEFAULT".to_string()),
+            });
+        }
+
+        // Parse the object name, which might be a single identifier or fully qualified name (e.g., x.y)
+        let parts = self.parse_object_name(false)?.0;
+        let (db_name, schema_name) = match parts.len() {
+            1 => {
+                // Single identifier found
+                if dialect_of!(self is DatabricksDialect) {
+                    if parsed_keyword == Some(Keyword::CATALOG) {
+                        // Databricks: CATALOG keyword provided, treat as database name
+                        (Some(parts[0].clone()), None)
+                    } else {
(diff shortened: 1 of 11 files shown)


src/parser/mod.rs
```diff
@@ -9225,8 +9225,63 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_use(&mut self) -> Result<Statement, ParserError> {
-        let db_name = self.parse_identifier(false)?;
-        Ok(Statement::Use { db_name })
+        // What should be treated as keyword in given dialect
+        let allowed_keywords = if dialect_of!(self is HiveDialect) {
```
git-hulk, line 9229: It'd be better to include `GenericDialect` as well. cc @iffyio

kacpermuda: AFAIU it does not have any specific keywords, so right now it would fall into `else` clause but you'd like to have a separate `if` for it with empty keywords?

git-hulk: @kacpermuda oh, sorry for not stating this clearly. What I mean is to include `GenericDialect` inside `dialect_of!`, so that the new syntaxes are valid to the generic dialect. For example;

`let allowed_keywords = if dialect_of!(self is HiveDialect)` can be changed to `let allowed_keywords = if dialect_of!(self is HiveDialect | GenericDialect)`

kacpermuda: Okay, I see. Maybe i should create a separate if for the `GenericDialect` that will include all the keywords possible in all other dialects? Otherwise when using `GenericDialect` we will fall into whatever `if` is first. Not sure how this `GenericDialect` works 😄

git-hulk: @kacpermuda My bad, I forgot the if condition only contains dialect. It makes sense to add another one for GenericDialect. Or just keep the current implementation and to see if @iffyio has any comments.

kacpermuda: Sure, let's wait and see then.

iffyio: Ah since support varies quite a bit, we could have GenericDialect preserve the old behavior supporting just `USE abc`? thinking since that seems to be the most common variant

src/parser/mod.rs
```diff
@@ -9225,8 +9225,63 @@ impl<'a> Parser<'a> {
...
+            vec![Keyword::DEFAULT]
+        } else if dialect_of!(self is DatabricksDialect) {
+            vec![Keyword::CATALOG, Keyword::DATABASE, Keyword::SCHEMA]
+        } else if dialect_of!(self is SnowflakeDialect) {
+            vec![Keyword::DATABASE, Keyword::SCHEMA]
+        } else {
+            vec![]
+        };
+        let parsed_keyword = self.parse_one_of_keywords(&allowed_keywords);
```
iffyio, line 9238: Representation wise would it make sense to do something like this instead?

```rust
enum Use {
  Catalog(ObjectName), // e.g. `USE CATALOG foo.bar`
  Schema(ObjectName), // e.g. `USE SCHEMA foo.bar`
  Database(ObjectName), // e.g. `USE DATABASE foo.bar`
  Default, // e.g. `USE DEFAULT`
  Object(ObjectName), // e.g. `USE foo.bar`
}
Statement::Use(Use)
```

I'm thinking that could be clearer what's expected, and likely simplify the impl since we would be able to avoid  the manual parsing of object names

src/parser/mod.rs
kacpermuda: Yes, i think with current behaviour that's exactly what's happening - the `GenericDialect` falls into else clause with no keywords, thus supporting only `USE mydb;`. Is that okay for you or should i still create another `else if` for `GenericDialect` just so it's explicit ?

tests/sqlparser_databricks.rs
```diff
@@ -189,3 +189,47 @@ fn test_values_clause() {
...
+    assert_eq!(
+        databricks().verified_stmt("USE my_schema"),
+        Statement::Use {
+            db_name: None,
+            schema_name: Some(Ident::new("my_schema")),
+            keyword: None
+        }
+    );
+    assert_eq!(
+        databricks().verified_stmt("USE CATALOG my_catalog"),
```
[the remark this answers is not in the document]
kacpermuda: Correct, I've added a test case for that. Thanks for the catch 😄

src/parser/mod.rs
kacpermuda: Do i correctly understand what you propose: you think we could modify the `src/ast/mod.rs:Statement.Use` to accept a single argument, this new `Use` Enum with different values for each syntax used? 

Initially, I considered using an enum instead of a plain String for the keyword, or even taking it a step further with an enum for each syntax, like you propose. However, I might be mistaken, but from the user’s perspective, when later implementing a Visitor for this Use statement, all I care about is easy access to the current database and schema set and the specific syntax used doesn’t really matter. Correct?

This is also my first time writing anything in Rust, so if there are best practices that support your solution, please let me know. I'm not very familiar with them yet, and I'd be happy to make adjustments.

iffyio: Ah yeah the proposal is indeed to modify `Statement::Use` and so that it accepts the enum. Exactly, the dialect/syntax used won't matter after the result comes out of the parser so the visitor would only look at the enum variant and handle each variant accordingly

src/parser/mod.rs
iffyio: Sounds good! We can keep it as is in that case yeah

src/parser/mod.rs
kacpermuda: I think this kind of switch from `Statement::Use(db_name)` to `Statement::Use(UseEnum)` would also be breaking change for those who already implemented their Visitors right? No sure what's the deprecation policy, but this is something to consider as well when doing such change. Given everything we've discussed, do you think this would be still a good approach to add that Enum? I'm open to your thoughts.

iffyio: Yeah I think we can still add it. It would be a breaking change, given the stage of the project we do make those when needed unfortunately. Ideally/usually they're mechanical changes to adopt but we try to keep them small in scope in any case to make it easier for downstream crates to adapt to the changes that go into each release.
Specifically for the `Statement` enum there's ongoing work [here](https://github.com/sqlparser-rs/sqlparser-rs/issues/1204) to move away from unnamed structs as variants so that this would likely be an eventual change to the `Statement::Use` variant

kacpermuda: I've started implementing it and I have one more question, and two examples to show what i mean. Hopefully it's clear:

First, using [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-use-schema.html) example, users can provide query like:
`USE foo;` -> It sets SCHEMA to `foo`, but most of dialects would set DATABASE to `foo` with that query.
`USE SCHEMA foo;` -> It also sets SCHEMA to `foo`

After we switch to the Enum we would have:
`USE foo;` -> `UseEnum(Object(foo))`
`USE SCHEMA foo;` -> `UseEnum(Schema(foo))`

When implementing Visitor, the User would have to know that Object(foo) in Databricks sets the SCHEMA and f.e. in MySQL sets the DATABASE.

Then, using [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/use-schema) example, users can provide
`USE SCHEMA foo.bar;` -> It will set DATABASE to `foo` and SCHEMA to `bar`, but f.e. Databricks allow a single identifier after SCHEMA keyword (e.g. only `foo` not `foo.bar`). 

After we switch to the Enum you proposed, we would have:
`USE SCHEMA foo.bar;` -> `UseEnum(Schema(foo.bar))`

When implementing Visitor, the User can't assume that Schema contains a single identifier (of course w still type it as (Schema(ObjectName)) so there is no confusion), but he has to do more parsing of Schema(foo.bar) to still account for different possibilities depending on the Dialect.

We could also possibly translate this query to Object() and not Schema() `USE SCHEMA foo.bar;` -> `UseEnum(Object(foo.bar))`, to show that we are dealing with full identifier, but the we lose the information about the keyword being used.

To sum up, i'm not saying it's wrong, just wanted to make sure that this is the desired division of responsibility - that the result returned by the parser still requires parsing effort from the user and knowing what is allowed / possible in different dialects. Or maybe I'm missing something here? Maybe my assumption writing current approach was wrong, that i wanted to deliver parsed information already and keep all the dialect-specific parsing inside 😄

kacpermuda: Regardless of my doubts I implemented the desired approach with Enum. Let me know if it looks better now @iffyio

iffyio: The changes look good to me thanks! Re the databricks example, yeah I think that would be expected behavior - goal of the parser is to try to faithfully represent the SQL syntax - semantics is deferred to the consumer of the output AST. So that if the input SQL is `USE foo;`, then the expected parser output indeed becomes `Use(Object(foo))`, an output `Use(Schema(foo))` I think could be surprising parser behavior.

> After we switch to the Enum you proposed, we would have:
> `USE SCHEMA foo.bar; -> UseEnum(Schema(foo.bar))`

Similarly, that looks like expected parser output to me.

Not exactly the same scenario but potentially related section from the project's goals re syntax vs semantics: https://github.com/sqlparser-rs/sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @kacpermuda  and @iffyio 

This change makes sense to me

alamb:
🚀

PR #1389, opened 2024-08-19
Support create index with clause

lewiszlw:
See postgresql doc: https://www.postgresql.org/docs/current/sql-createindex.html


src/ast/dml.rs
```diff
@@ -45,6 +45,7 @@ pub struct CreateIndex {
     pub if_not_exists: bool,
     pub include: Vec<Ident>,
     pub nulls_distinct: Option<bool>,
+    pub with: Vec<Expr>,
```
git-hulk, line 48: It'd be great if can add a link to the docs here.

lewiszlw: Added.

git-hulk (approved):

tests/sqlparser_common.rs
```diff
@@ -7516,6 +7517,51 @@ fn test_create_index_with_using_function() {
...
             assert!(if_not_exists);
             assert!(include.is_empty());
+            assert!(with.is_empty());
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn test_create_index_with_with_function() {
```
iffyio, line 7527:
```suggestion
fn test_create_index_with_with_clause() {
```

src/parser/mod.rs
```diff
@@ -5174,6 +5174,15 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let with = if self.parse_keyword(Keyword::WITH) {
```
iffyio, line 5177: I'm wondering would it make more sense to have this limited to Postgres and Generic dialects? I'm guessing this functionality isn't common (took a quick look and found mysql has but the syntax is different so the logic won't be reusable if we later add support)

lewiszlw: Looks like mysql doesn't have `WITH` clause, it's `WITH PARSER`. https://dev.mysql.com/doc/refman/8.4/en/create-index.html
But I'm ok if this needs to be limited to Postgres and Generic dialects.

iffyio: Yeah I think we can limit it to the pg and generic dialects - mysql support will probably be the same in that we'd land in this code block since we need to first consume a WITH in both cases

lewiszlw: Done.

tests/sqlparser_postgres.rs
```diff
@@ -2261,6 +2262,51 @@ fn parse_create_index() {
...
             assert_eq_vec(&["col1", "col2"], &columns);
             assert!(include.is_empty());
+            assert!(with.is_empty());
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn test_create_index_with_with_clause() {
```
iffyio, line 2272: since the functionality uses a method on the dialect, we can rewrite the test in [this style](https://github.com/sqlparser-rs/sqlparser-rs/blob/222b7d127acb265472cc9bf2e81ef2453a53a9ce/tests/sqlparser_common.rs#L2765) in common.rs so that future dialects that gain support get test coverage automatically

lewiszlw: Cool

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @lewiszlw  @iffyio and @git-hulk 

I pushed a commit to this PR to fix the CI doc failure too

alamb:
Thanks again !

PR #1394, opened 2024-08-22
allow `DateTimeField::Custom` with `EXTRACT` in Postgres

samuelcolvin:
I'm really not a fan of this `dialect_of!` stuff, it feels like I'm back writing some kind of odd Python. (I've changed this)

Follows #1191.

This will allow us to match postgres syntax:

```sql
postgres@localhost:postgres> select extract(seconds from '2 seconds'::interval);
╒══════════╕
│ extract  │
╞══════════╡
│ 2.000000 │
╘══════════╛
SELECT 1
Time: 0.007s
postgres@localhost:postgres> select extract('seconds' from '2 seconds'::interval);
╒══════════╕
│ extract  │
╞══════════╡
│ 2.000000 │
╘══════════╛
SELECT 1
Time: 0.007s
postgres@localhost:postgres> select extract(minutes from '2 seconds'::interval);
╒═════════╕
│ extract │
╞═════════╡
│ 0       │
╘═════════╛
SELECT 1
Time: 0.007s
postgres@localhost:postgres> select extract(ms from '2 seconds'::interval);
╒══════════╕
│ extract  │
╞══════════╡
│ 2000.000 │
╘══════════╛
SELECT 1
Time: 0.006s
postgres@localhost:postgres> select extract('ms' from '2 seconds'::interval);
╒══════════╕
│ extract  │
╞══════════╡
│ 2000.000 │
╘══════════╛
SELECT 1
Time: 0.006s
```

I guess all the unit names and abbreviations allowed by `INTERVAL` which I added in https://github.com/apache/arrow-rs/pull/6211 are allowed, we should be able to use `IntervalUnit` in datafusion to process the custom value.


samuelcolvin:
I think this is good to go.

tests/sqlparser_postgres.rs
```diff
@@ -4998,3 +4998,51 @@ fn arrow_cast_precedence() {
         }
     )
 }
+
```
iffyio, line 5001: could we add the tests in the generic style [like so](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/tests/sqlparser_common.rs#L2765) so that it covers all and future dialects that support this syntax?

samuelcolvin: done.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @samuelcolvin  and @iffyio

PR #1396, opened 2024-08-22
fix interval precedence for interval expressions

samuelcolvin:
I'm not sure this is exactly the right solution, but it's definitely closer to correct for PostgreSQL.

### DIFF
--- a/examples/cli.rs
+++ b/examples/cli.rs
@@ -31,7 +31,7 @@ Usage:
 $ cargo run --example cli FILENAME.sql [--dialectname]
 
 To print the parse results as JSON:
-$ cargo run --feature json_example --example cli FILENAME.sql [--dialectname]
+$ cargo run --features json_example --example cli FILENAME.sql [--dialectname]
 
 "#,
     );
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -482,6 +482,7 @@ pub trait Dialect: Debug + Any {
             Precedence::UnaryNot => 15,
             Precedence::And => 10,
             Precedence::Or => 5,
+            Precedence::Interval => 4,
         }
     }
 
@@ -522,6 +523,7 @@ pub enum Precedence {
     UnaryNot,
     And,
     Or,
+    Interval,
 }
 
 impl dyn Dialect {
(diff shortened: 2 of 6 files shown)


samuelcolvin (commented):
Note, datafusion has [lots of hacky logic](https://github.com/apache/datafusion/blob/87c1c173c8adb781d02e9907af297734f8a981ed/datafusion/sql/src/expr/value.rs#L210-L319) dedicated to working around this issue.

We could probably remove all that hacky logic if we fixed this properly in sqlparser-rs.

samuelcolvin:
Okay, I've done some more research, and basically there are three groups of databases with pretty distinct ways of parsing interval expressions:

## 1. PostgreSQL-like

Expect `INTERVAL` expressions to be in the form

```sql
SELECT INTERVAL '1 second'
```

These bind `interval` to the literal tightly.

There's pretty good support for parsing these interval expressions in arrow-rs (added in https://github.com/apache/arrow-rs/pull/6211).

Databases: Postgres, DuckDB, Redshift, Snowflake

## 2. MySQL-like

Expect `INTERVAL` expressions to be in the form

```sql
SELECT INTERVAL 1 SECOND
```

they also support `INTERFVAL 1 + 1 SECOND`, so I presume the logic is "keep parsing literals until you meet one of the following units, there's a good list of these units [here](https://dev.mysql.com/doc/refman/9.0/en/expressions.html).

Databases: MySQL, BigQuery,  Databricks, Hive, ClickHouse

### Sqlite and MsSql

Don't seem to support interval syntax at all.

---

Critically, from what I can tell, type 1 and type 2 don't support the other syntax at all.

Currently `sqlparser-rs` seems to partially support both syntaxes, with a preference for the MySQL style. Datafusion then tries to [hack around](https://github.com/apache/datafusion/blob/87c1c173c8adb781d02e9907af297734f8a981ed/datafusion/sql/src/expr/value.rs#L210-L319) the problems with this.

I would therefore suggest we change the logic to have two separate functions for the two approaches.

@alamb if you're happy with that I can try to implement this.

samuelcolvin:
Okay, this is wrong, since syntax like

```sql
SELECT INTERVAL '1' DAY;
```

Will be parsed incorrectly. I'm also somewhat uncomfortable that tests aren't currently failing.

Having looked through this for about an hour, I think the current implementation is pretty messy, but the solution definitely isn't as simple as I had hoped.

I think the only solution is to rip lots of the current hacks out, and reimplement the interval parsing from scratch in a saner way.

samuelcolvin:
replaced by #1398.

PR #1398, opened 2024-08-23
Fix `INTERVAL` parsing to support expressions and units via dialect

samuelcolvin:
Replaces #1396.

# Background

I've done some more research, and basically there are three groups of databases with pretty distinct ways of parsing interval expressions:

## 1. PostgreSQL-like (here `require_interval_qualifier => false`)

Expect `INTERVAL` expressions to be in the form

```sql
SELECT INTERVAL '1 second'
```

These bind `interval` to the literal tightly.

There's pretty good support for parsing these interval expressions in arrow-rs (added in https://github.com/apache/arrow-rs/pull/6211).

These dialects also support syntax like `INTERVAL 1 SECOND`, but notably do not support syntax of the form `INTERVAL 1 + 1 SECOND`.

Databases: Postgres, DuckDB, Redshift, Snowflake

## 2. MySQL-like aka (here `require_interval_qualifier => true`)

Expect `INTERVAL` expressions to be in the form

```sql
SELECT INTERVAL 1 SECOND
```

they also support `INTERVAL 1 + 1 SECOND`, so I presume the logic is "keep parsing literals until you meet one of the following units", there's a good list of these units [here](https://dev.mysql.com/doc/refman/9.0/en/expressions.html). **Note these aren't all included in sqlparser-rs, that would be a good follow-up PR.**

These dialects do not support syntax like `INTERVAL '1 second'` — e.g. they require the unit/qualifier to be provided

Databases: MySQL, BigQuery,  Databricks, Hive, ClickHouse

### Sqlite and MsSql

Don't seem to support interval syntax at all.

# The issue

Currently `sqlparser-rs` seems to partially support both syntaxes, with a preference for the MySQL style. 

This meant that `INTERVAL '1 second' > x` was wrongly being interpreted as `INTERVAL ('1 second' > x)`.

Datafusion then tries to [hack around](https://github.com/apache/datafusion/blob/87c1c173c8adb781d02e9907af297734f8a981ed/datafusion/sql/src/expr/value.rs#L210-L319) the problem but it wasn't a water tight solution.

## Change Proposed

This PR adds `fn require_interval_qualifier(&self) -> bool` to dialects and uses that to decide how to parse interval expressions. `require_interval_qualifier => false` also means expressions within intervals (e.g. `INTERVAL 1 + 1 DAY`) are forbidden.

The `GenericDialect` is set to `require_interval_qualifier => false` since intervals without units are vastly more common than intervals with expressions.

The changes required are actually smaller than I feared.

Once this is used in datafusion, it should allow us to remove the [`interval` hack](https://github.com/apache/datafusion/blob/87c1c173c8adb781d02e9907af297734f8a981ed/datafusion/sql/src/expr/value.rs#L210-L319) completely.


src/parser/mod.rs
```diff
@@ -56,15 +56,6 @@ macro_rules! parser_err {
     };
 }
 
-// Returns a successful result if the optional expression is some
```
samuelcolvin, line 59: drive-by change, this seemed really ugly, I couldn't resist fixing it.

src/parser/mod.rs
```diff
@@ -896,35 +887,6 @@ impl<'a> Parser<'a> {
         Ok(expr)
     }
 
-    pub fn parse_interval_expr(&mut self) -> Result<Expr, ParserError> {
```
samuelcolvin, line 899: These are "moved" to after `parse_interval`, so the ordering makes more sense, but really they're completely rewritten.

tests/sqlparser_bigquery.rs
```diff
@@ -830,16 +830,14 @@ fn parse_typed_struct_syntax_bigquery() {
         expr_from_projection(&select.projection[3])
     );
 
-    let sql = r#"SELECT STRUCT<INTERVAL>(INTERVAL '1-2 3 4:5:6.789999'), STRUCT<JSON>(JSON '{"class" : {"students" : [{"name" : "Jane"}]}}')"#;
+    let sql = r#"SELECT STRUCT<INTERVAL>(INTERVAL '1' DAY), STRUCT<JSON>(JSON '{"class" : {"students" : [{"name" : "Jane"}]}}')"#;
```
samuelcolvin, line 833: `INTERVAL '1-2 3 4:5:6.789999'` isn't a valid bigquery interval, I checked.

tests/sqlparser_common.rs
```diff
@@ -4464,9 +4464,9 @@ fn parse_window_functions() {
                avg(bar) OVER (ORDER BY a \
                RANGE BETWEEN 1 PRECEDING AND 1 FOLLOWING), \
                sum(bar) OVER (ORDER BY a \
-               RANGE BETWEEN INTERVAL '1' DAY PRECEDING AND INTERVAL '1 MONTH' FOLLOWING), \
+               RANGE BETWEEN INTERVAL '1' DAY PRECEDING AND INTERVAL '1' MONTH FOLLOWING), \
```
samuelcolvin, line 4467: again, `interval '1 MONTH'` isn't valid with MySQL type databases.

samuelcolvin (commented):
@alamb please could you take a look at this.

samuelcolvin:
conflicts fixed. This is ready to review.

samuelcolvin:
This is tested in datafusion in https://github.com/apache/datafusion/pull/12222.

samuelcolvin:
Okay, having thought about this more, I'm not convinced the workaround I have now on the generic dialect of

```rs
    fn allow_interval_expressions(&self) -> bool {
        true
    }

    fn require_interval_units(&self) -> bool {
        false
    }
```

Makes sense.

It reduces the effect of the change, but at the expense of some incorrect behaviour, e.g.: `SELECT interval '1 day' > interval '1 hour'` is interpreted as `SELECT interval ('1 day' > interval '1 hour')`. This is the original behaviour I came to fix on Postgres.

I think `GenericDialect` should choose one flavor and stick to it, either:
* MySQL flavor — expressions are allowed, units are required
* or, PostgreSQL flavor — expressions are not allowed, units are not required

I don't mind which (we're using `PostgresDialect`), but I will say that the number of changes required in datafusion tests will be much smaller if we go with Postgres flavor, and my instinct is that "no units" (e.g. `INTERVAL '1 day'`) is much more commonly used than expressions like `INTERVAL '1' + '1' day`.

I've spent long enough on this with no feedback, so I'll wait for feedback before proceeding. 

But this is fairly urgent for us since interval precedence is pretty broken in datafusion right now. cc @alamb @andygrove @git-hulk

git-hulk:
This fix looks correct to me, but it might affect a lot for the downstream. To see if @iffyio @alamb feels good about this change?

src/parser/mod.rs
```diff
@@ -2160,6 +2103,43 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    pub fn next_token_is_unit(&mut self) -> bool {
```
alamb, line 2106: Since this is a `pub fn` I think we should add some documentation or make it non `pub`

tests/sqlparser_bigquery.rs
```diff
@@ -830,16 +829,14 @@ fn parse_typed_struct_syntax_bigquery() {
         expr_from_projection(&select.projection[3])
     );
 
-    let sql = r#"SELECT STRUCT<INTERVAL>(INTERVAL '1-2 3 4:5:6.789999'), STRUCT<JSON>(JSON '{"class" : {"students" : [{"name" : "Jane"}]}}')"#;
+    let sql = r#"SELECT STRUCT<INTERVAL>(INTERVAL '2' HOUR), STRUCT<JSON>(JSON '{"class" : {"students" : [{"name" : "Jane"}]}}')"#;
```
alamb, line 832: Could you please 
1. leave the existing test as is to show what the effect of the changes are on this query? (I think it would error?)
2. Add a new test for this new `'2' HOUR` variant

alamb: Looking at the BigQuery syntax in https://cloud.google.com/bigquery/docs/reference/standard-sql/interval_functions I actually think the new test is more useful / correct -- all the intervals appear to have a unit after them, such as:

```sql
...
  UNNEST([INTERVAL '1-2 3 4:5:6.789999' YEAR TO SECOND,
          INTERVAL '0-13 370 48:61:61' YEAR TO SECOND]) AS i
...
```

alamb (approved):
Thank you @samuelcolvin  -- and I very much apologize for the delay in reviewing this PR.

I think this PR is well documented and well tested and the overall changes make sense to me. The biggest thing I think that we should do (I think alluded to by @git-hulk) is ensure the change for end users of this crate are well understood:
1. Are there queries that would have parsed previously that now will not? 
2. If "yes", how could users get the old behavior if they wanted?

I have updated the PR title to be a little more specific -- can you double check this?

cc @iffyio @lovasoa and @jmhain

lovasoa:
Great PR ! Maybe a nitpick about the name: `allow_interval_expressions` is probably a little bit confusing. I would have named it the other way around: `allow_expressions_in_interval`...

tests/sqlparser_bigquery.rs
samuelcolvin: yup, BigQuery is MySQL flavoured  when it comes to intervals, so intervals without units are not permitted.

src/parser/mod.rs
samuelcolvin: done.

samuelcolvin:
Okay I think this is ready @alamb @lovasoa.

> The biggest thing I think that we should do (I think alluded to by @git-hulk) is ensure the change for end users of this crate are well understood:

I believe this PR will cause minimal changes to end users. This will cause the following changes:
* With the generic dialect expressions within intervals (e.g. `INTERVAL 1 + 1 DAYS` will not be disallowed)
* Same with Postgres, snowflake, duckdb, redshift — From testing Postgres and duckdb, this matches their actual behaviour
* With MySQL, BigQuery, databricks, hive — units are now required (e.g. `INTERVAL 1 day` works, but `interval '1 day'`) does not — From testing MySQL and BigQuery, this matches their actual behaviour

samuelcolvin:
(I've updated my comment above to correct a very confusing "not" -> "now")

lovasoa (approved):

alamb:
I'll plan to merge this PR tomorrow Sep 5 unless anyone else would like time to review or otherwise comment

alamb:
🚀

PR #1401, opened 2024-08-27
With Order Support for Memory Tables

mertak-synnada:
Add support for WITH ORDER syntax for Memory Tables


src/ast/dml.rs
```diff
@@ -124,6 +124,7 @@ pub struct CreateTable {
     pub default_charset: Option<String>,
     pub collation: Option<String>,
     pub on_commit: Option<OnCommit>,
+    pub with_order: Vec<Vec<OrderByExpr>>,
```
iffyio, line 127: Could we include here a link to the docs where this syntax comes from? (e.g. similar to the on_cluster below)

iffyio, line 127: Can we include test cases demonstrating the behavior introduced by the changes?

src/parser/mod.rs
```diff
@@ -5589,7 +5589,23 @@ impl<'a> Parser<'a> {
         let clustered_by = self.parse_optional_clustered_by()?;
         let hive_formats = self.parse_hive_formats()?;
         // PostgreSQL supports `WITH ( options )`, before `AS`
-        let with_options = self.parse_options(Keyword::WITH)?;
+        let mut with_options: Vec<SqlOption> = vec![];
+        let mut order_exprs: Vec<Vec<OrderByExpr>> = vec![];
+        if self.parse_keyword(Keyword::WITH) {
+            if self.parse_keyword(Keyword::ORDER) {
```
iffyio, line 5595:
```suggestion
        if self.parse_keywords(&[Keyword::WITH, Keyword::ORDER]) {
```
Also I think we can keep the postgres path separate since this feature doesn't apply to that dialect? - so that this part remains the same
```rust
        let with_options = self.parse_options(Keyword::WITH)?;
```
We can make the fuctionality conditional with e.g. `if self.dialect.supports_with_order()` or similar method

src/parser/mod.rs
```diff
@@ -5589,7 +5589,23 @@ impl<'a> Parser<'a> {
...
+                self.expect_token(&Token::LParen)?;
+                loop {
+                    order_exprs.push(vec![self.parse_order_by_expr()?]);
+                    if !self.consume_token(&Token::Comma) {
+                        self.expect_token(&Token::RParen)?;
+                        break;
+                    }
+                }
```
iffyio, line 5603: Can the `parse_comma_separated()` method be reused for this functionality?

src/parser/mod.rs
```diff
@@ -12364,6 +12385,43 @@ mod tests {
         );
     }
 
+    #[test]
```
iffyio, line 12388: hmm I think this would probably not be the ideal place for the test to live - if the feature isn't bound to any dialect, or if it uses a dialect method to guard the feature, we can add the tests to common e.g. https://github.com/sqlparser-rs/sqlparser-rs/blob/main/tests/sqlparser_common.rs#L4405

src/ast/dml.rs
alamb: FWIW I did some research and could not find any existing databases that use this syntax.

ClickHouse does seem to have a way to specify order as part of a `CREATE TABLE` statement: https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree

@mertak-synnada rather than introducing special DataFusion only syntax support, what do you think about extending DataFusion to use the existing ClickHouse syntax?

```sql
CREATE TABLE ....
ORDER BY <...>
```

We might have to change the `GenericDialect` or add some feature to the `Dialect` trait to permit other dialects to parse such syntax, but it would be nice to align to some other existing syntax rather than creating something new

ozankabak: I see two possible ways forward:
1. We can change DF to support `ORDER BY` after `CREATE` and deprecate `WITH ORDER`. This way we would have Clickhouse-like syntax for both memory and external tables, and `WITH ORDER` syntax only for external tables. The latter asymmetry is a little weird but we can explain it away by deprecating `WITH ORDER`.
2. We can support both `ORDER BY` and `WITH ORDER` for both memory and external tables in DF (they would be synonyms). We'd need to add `WITH ORDER` for ordinary `CREATE`s here though.

I am OK with both, with a slight preference to 2 because DF has had `WITH ORDER` for a while now. What do you think?

alamb: I think in this case we should distinguish between DataFusion's needs and sqlparser-rs's needs

If we are going to introduce `WITH ORDER` to sqlparser-rs I think it should follow the existing pattern of being connected to a specific dialect (in this case `DatafusionDialect`). 

Thus I would personally vote for option 1 (change DF)

> 1. We can change DF to support ORDER BY after CREATE and deprecate WITH ORDER. This way we would have Clickhouse-like syntax for both memory and external tables, and WITH ORDER syntax only for external tables. The latter asymmetry is a little weird but we can explain it away by deprecating WITH ORDER.

alamb:
Marking as draft as i think this is still under discussion and not in need of more review. Please mark it as review if I got that wrong or when it is ready

src/ast/dml.rs
ozankabak: Makes sense - let's do it

PR #1402, opened 2024-08-27
Fix identifier starts with `$` should be regarded as a placeholder in SQLite

git-hulk:
Currently, `$var` would be parsed as an identifier instead of a placeholder in SQLite dialect, which is unexpected from the SQLite's documentation: https://www.sqlite.org/lang_expr.html#varparam.

This relates to the issue comment: https://github.com/sqlparser-rs/sqlparser-rs/issues/291#issuecomment-2309797948


iffyio (approved):
LGTM! cc @alamb

alamb (approved):
❤️ 

Thanks @git-hulk  and @iffyio

PR #1403, opened 2024-08-28
Add support for postgres TRUNCATE syntax

tobyhede:
Correctly parses the following pg options:
```
ONLY
RESTART IDENTITY | CONTINUE IDENTITY 
CASCADE | RESTRICT 
```

And supports providing a list of table names 
```
TRUNCATE table_a, table_b, table_c
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -682,7 +682,11 @@ impl<'a> Parser<'a> {
     pub fn parse_truncate(&mut self) -> Result<Statement, ParserError> {
         let table = self.parse_keyword(Keyword::TABLE);
         let only = self.parse_keyword(Keyword::ONLY);
-        let table_name = self.parse_object_name(false)?;
+
+        let table_names = self.parse_comma_separated(|p| p.parse_object_name(false))?;
+
+        // Unwrap is safe - the preceding parse  fails if there is not at least one table name
+        let table_name = table_names.first().unwrap().clone();
 
         let mut partitions = None;
         if self.parse_keyword(Keyword::PARTITION) {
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -681,17 +681,44 @@ impl<'a> Parser<'a> {
...
+        let table_name = table_names.first().unwrap().clone();
+
         let mut partitions = None;
         if self.parse_keyword(Keyword::PARTITION) {
             self.expect_token(&Token::LParen)?;
             partitions = Some(self.parse_comma_separated(Parser::parse_expr)?);
             self.expect_token(&Token::RParen)?;
         }
+
+        let identity = if self.parse_keywords(&[Keyword::RESTART, Keyword::IDENTITY]) {
```
git-hulk, line 698: I think we can add the dialect guard if expected to support for PostgreSQL only.

src/parser/mod.rs
```diff
@@ -681,17 +681,44 @@ impl<'a> Parser<'a> {
...
+            Some(TruncateIdentityOption::Restart)
+        } else if self.parse_keywords(&[Keyword::CONTINUE, Keyword::IDENTITY]) {
+            Some(TruncateIdentityOption::Continue)
+        } else {
+            None
+        };
+
+        let cascade = if self.parse_keyword(Keyword::CASCADE) {
```
git-hulk, line 706: Same as the above

src/parser/mod.rs
tobyhede: On it

tobyhede:
Fixing tests. Pulling changes from our fork and apparently the commits are not as cleanly separated as I thought

PR #1406, opened 2024-08-29
Suppor postgres `TRUNCATE` syntax

tobyhede:
Correctly parses the following pg options:

```
ONLY
RESTART IDENTITY | CONTINUE IDENTITY 
CASCADE | RESTRICT 
```

And supports providing a list of table names
```
TRUNCATE table_a, table_b, table_c
```

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2013,11 +2013,12 @@ pub enum Statement {
     Truncate {
         #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
         table_name: ObjectName,
+        table_names: Vec<ObjectName>,
         partitions: Option<Vec<Expr>>,
         /// TABLE - optional keyword;
         table: bool,
         /// Postgres-specific option
-        /// TRUNCATE [ TABLE ] [ ONLY ] name
+        /// [ TRUNCATE TABLE ONLY ]
         only: bool,
         /// Postgres-specific option
         /// [ RESTART IDENTITY | CONTINUE IDENTITY ]
@@ -3140,7 +3141,8 @@ impl fmt::Display for Statement {
                 Ok(())
             }
             Statement::Truncate {
-                table_name,
+                table_name: _,
+                table_names,
                 partitions,
                 table,
                 only,
@@ -3149,7 +3151,15 @@ impl fmt::Display for Statement {
             } => {
                 let table = if *table { "TABLE " } else { "" };
                 let only = if *only { "ONLY " } else { "" };
-                write!(f, "TRUNCATE {table}{only}{table_name}")?;
+
+                let table_names = table_names
+                    .iter()
+                    .map(|table_name| table_name.to_string()) // replace `to_string()` with the appropriate method if necessary
+                    .collect::<Vec<String>>()
+                    .join(", ");
+
+                write!(f, "TRUNCATE {table}{only}{table_names}")?;
+
                 if let Some(identity) = identity {
                     match identity {
                         TruncateIdentityOption::Restart => write!(f, " RESTART IDENTITY")?,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -2013,9 +2013,19 @@ pub enum Statement {
     Truncate {
         #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
         table_name: ObjectName,
+        table_names: Vec<ObjectName>,
```
tobyhede, line 2016: Design note: not totally happy with this, but `visit_relation` requires an `ObjectName`, and I didn't want to get caught attempting to refactor that code. `table_name` is populated during parse.

src/parser/mod.rs
```diff
@@ -681,17 +681,49 @@ impl<'a> Parser<'a> {
 
     pub fn parse_truncate(&mut self) -> Result<Statement, ParserError> {
         let table = self.parse_keyword(Keyword::TABLE);
-        let table_name = self.parse_object_name(false)?;
+        let only = self.parse_keyword(Keyword::ONLY);
+
+        let table_names = self.parse_comma_separated(|p| p.parse_object_name(false))?;
+
+        // Unwrap is safe - the preceding parse  fails if there is not at least one table name
+        let table_name = table_names.first().unwrap().clone();
```
iffyio, line 689:
```suggestion
        let Some(table_name) = table_names.first().clone() else {
               parser_err!(...)
        };
```
Can we use something like this format in any case? I think that would be safer if the semantics of the preceeding parse fails, or the code gets restructured etc in the future.

src/ast/mod.rs
iffyio: hmm the workaround to have both `table_name` and `table_names` doesn't seem desirable. (extending `visit_relation` specifically I suspect might not be desirable either).
Could be that if a truncate can reference multiple tables then maybe its no longer eligible for the visitor functionality and we can remove the derive instead (replacing `table_name` with `table_names`) - just thinking out loud based on how similar statements like `UPDATE` seem to be treated in this regard today which seemed to suggest there's no support currently for such cases? But yeah, I'm not familiar with the use cases to tell what makes sense, cc @alamb @lovasoa if you have thoughts?

src/ast/mod.rs
```diff
@@ -3131,12 +3141,38 @@ impl fmt::Display for Statement {
...
-                write!(f, "TRUNCATE {table}{table_name}")?;
+                let only = if *only { "ONLY " } else { "" };
+
+                let table_names = table_names
+                    .iter()
+                    .map(|table_name| table_name.to_string())
+                    .collect::<Vec<String>>()
+                    .join(", ");
+
+                write!(f, "TRUNCATE {table}{only}{table_names}")?;
```
iffyio, line 3161:
```suggestion
                write!(f, "TRUNCATE {table}{only}{table_names}", table_names = display_comma_separated(...))?;
```
does the helper function apply to this case, it might simplify the join if so?

src/ast/mod.rs
alamb: Could we work around the need to call `visit_relation` on a single object by adding another layer of indirection? Something like

```rust
/// Target of a `TRUNCATE TABLE` command
/// 
/// Note this is its own struct because `visit_relation` requires an `ObjectName` (not a `Vec<ObjectName>`)
struct TruncateTableTarget {
  /// name of the table being truncated
  name: ObjectName,
}

pub enum Statement { 
  ...
  Truncate {
    table_names: Vec<TruncateTableTarget>,
    ....
  }
}
```
```

alamb (approved):
Thanks @tobyhede  -- looks good to me.

I'll plan to leave this open for a few days to give @iffyio  and others a chance to review as well if they want

PR #1410, opened 2024-09-05
fix: Fix stack overflow in parse_subexpr

eejbyfeldt:
This patch moves the guard from parse_expr to parse_subexpr. This helps
prevent us from overflowing the stack in more cases. Also added a test case for where the current code has a stackoverflow. The test cases is a expression contains a repeated +(unary plus) leading to a stackoverflow.

Found by running the fuzzer.


alamb (approved):
Looks good to me. Thank you @eejbyfeldt

alamb:
FYI @iffyio

PR #1411, opened 2024-09-05
fix: Guard against stack overflow in parse_table_and_joins

eejbyfeldt:
It possible to hit a stack overflow due to `parse_table_and_joins`
calling `parse_table_factor` which then calls `parse_table_and_joins`.

This fixes the issue by adding a recurssion guard in
`parse_table_and_joins`.

Found by running the fuzzer.


eejbyfeldt:
New test case still overflows in debug builds. Probably needs similar changes as done in #1171

PR #1413, opened 2024-09-06
Fix Hive table comment should be after table column definitions

### LINKED ISSUE #1395: HiveDialect doesn't support create table with PARTITIONED BY and CLUSTERED BY clause
When I used the Hive dialect to parse the following Hive SQL, I encountered the following error.
```
    CREATE TABLE user_info_bucketed(user_id BIGINT, firstname STRING, lastname STRING)
    COMMENT 'A bucketed copy of user_info'
    PARTITIONED BY(ds STRING)
    CLUSTERED BY(user_id) INTO 256 BUCKETS;
```
with error
```
Err(ParserError("Expected: end of statement, found: PARTITIONED at Line: 4, Column: 5"))
```
Here is the code I tested.
```rust
use sqlparser::dialect::HiveDialect;

fn main() {
    let query = r#"
    CREATE TABLE user_info_bucketed(user_id BIGINT, firstname STRING, lastname STRING)
    COMMENT 'A bucketed copy of user_info'
    PARTITIONED BY(ds STRING)
    CLUSTERED BY(user_id) INTO 256 BUCKETS;"#;
    let dialect = HiveDialect {};
    let ast = Parser::parse_sql(&dialect, query);
    println!("ast={:?}", ast);
}
```

image

git-hulk:
For the documentation, please refer to:
https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-CreateTable

This closes #1395

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5735,18 +5735,14 @@ impl<'a> Parser<'a> {
         let strict = self.parse_keyword(Keyword::STRICT);
 
         // Excludes Hive dialect here since it has been handled after table column definitions.
-        if !dialect_of!(self is HiveDialect) {
-            comment = if self.parse_keyword(Keyword::COMMENT) {
-                let _ = self.consume_token(&Token::Eq);
-                let next_token = self.next_token();
-                match next_token.token {
-                    Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
-                    _ => self.expected("comment", next_token)?,
-                }
-            } else {
-                None
-            };
-        }
+        if !dialect_of!(self is HiveDialect) && self.parse_keyword(Keyword::COMMENT) {
+            let _ = self.consume_token(&Token::Eq);
+            let next_token = self.next_token();
+            comment = match next_token.token {
+                Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
+                _ => self.expected("comment", next_token)?,
+            }
+        };
 
         // Parse optional `AS ( query )`
         let query = if self.parse_keyword(Keyword::AS) {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5613,6 +5613,17 @@ impl<'a> Parser<'a> {
 
         // parse optional column list (schema)
         let (columns, constraints) = self.parse_columns()?;
+        let mut comment = if dialect_of!(self is HiveDialect)
+            && self.parse_keyword(Keyword::COMMENT)
+        {
+            let next_token = self.next_token();
+            match next_token.token {
+                Token::SingleQuotedString(str) => Some(CommentDef::AfterColumnDefsWithoutEq(str)),
+                _ => self.expected("comment", next_token)?,
+            }
+        } else {
+            None
+        };
 
         // SQLite supports `WITHOUT ROWID` at the end of `CREATE TABLE`
         let without_rowid = self.parse_keywords(&[Keyword::WITHOUT, Keyword::ROWID]);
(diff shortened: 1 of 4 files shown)


git-hulk:
cc @iffyio

src/parser/mod.rs
```diff
@@ -5723,16 +5734,19 @@ impl<'a> Parser<'a> {
...
-            let next_token = self.next_token();
-            match next_token.token {
-                Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
-                _ => self.expected("comment", next_token)?,
-            }
-        } else {
-            None
-        };
+        // For Hive, the comment is after the table column definitions
+        if !dialect_of!(self is HiveDialect) {
```
iffyio, line 5738: is this correct to switch this to being guarded on the Hive dialect (other dialects didn't support it previously)?

src/parser/mod.rs
```diff
@@ -5723,16 +5734,19 @@ impl<'a> Parser<'a> {
...
+            comment = if self.parse_keyword(Keyword::COMMENT) {
+                let _ = self.consume_token(&Token::Eq);
+                let next_token = self.next_token();
+                match next_token.token {
+                    Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
```
iffyio, line 5743: > For Hive, the comment is after the table column definitions

Not sure I followed this part, the comment seems to suggest this should also use `CommentDef::AfterColumnDefsWithoutEq`?

src/parser/mod.rs
git-hulk: I see, the comment line is a bit misleading. I use `!dialect_of!(self is HiveDialect)` here to disallow `Hive` to parse the comment here?

src/parser/mod.rs
git-hulk: It should be correct to use `CommentDef::WithoutEq` since it will parse comment line for other dialects except here.

src/parser/mod.rs
```diff
@@ -5723,16 +5734,19 @@ impl<'a> Parser<'a> {
...
-            let _ = self.consume_token(&Token::Eq);
-            let next_token = self.next_token();
-            match next_token.token {
-                Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
-                _ => self.expected("comment", next_token)?,
-            }
-        } else {
-            None
-        };
+        // For Hive, the comment is after the table column definitions
```
git-hulk, line 5737:
```suggestion
        // Excludes Hive dialect here since it has been handled after table column definitions.
```

src/parser/mod.rs
```diff
@@ -5723,16 +5734,19 @@ impl<'a> Parser<'a> {
...
-            match next_token.token {
-                Token::SingleQuotedString(str) => Some(CommentDef::WithoutEq(str)),
-                _ => self.expected("comment", next_token)?,
-            }
-        } else {
-            None
-        };
+        // Excludes Hive dialect here since it has been handled after table column definitions.
+        if !dialect_of!(self is HiveDialect) {
+            comment = if self.parse_keyword(Keyword::COMMENT) {
```
iffyio, line 5739: we can probably merge this with the dialect check to be `if !dialect_of() && self.parse_keyword()`? I think that lets us skip the `else { None }` block

src/parser/mod.rs
iffyio: Oh right! 😅 my bad seems I'm used to seeing if dialect_of and my brain skipped the !

src/parser/mod.rs
iffyio: Yeah that sounds good! this was due to my misreading the dialect check

iffyio (approved):
LGTM! cc @alamb

src/parser/mod.rs
git-hulk: Good point, has been changed according to the suggestion.

alamb (approved):
Thanks @git-hulk and @iffyio

PR #1416, opened 2024-09-08
minor: Add `dabricks` dialect to `dialect_from_str`

milenkovicm:
Databricks dialect is missing in `dialect_from_str` function

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -587,6 +587,7 @@ pub fn dialect_from_str(dialect_name: impl AsRef<str>) -> Option<Box<dyn Dialect
         "bigquery" => Some(Box::new(BigQueryDialect)),
         "ansi" => Some(Box::new(AnsiDialect {})),
         "duckdb" => Some(Box::new(DuckDbDialect {})),
+        "databricks" => Some(Box::new(DatabricksDialect {})),
         _ => None,
     }
 }


Dandandan:
> Databricks dialect is missing in `dialect_from_str` function

Is there anything that you plan implementing for Databricks dialect? I think databricks mostly should be able to use ansidialect.

milenkovicm:
I'm not sure I'll be able to implement anything in short term, I've tried to use it from datafusion to realise it's not there.

the edge case I've found is `CREATE TABLE USING` https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html

please ignore PR if it should not be expose

alamb (approved):
I think this is a reasonable change and improvement. Thank you @milenkovicm  and @Dandandan 

Maybe it would be great to add some tests to avoid new dialects being missed, but we could do that as a follow on PR

milenkovicm:
also, as a follow up, once sqlparser has been upgraded, error messages have to handle new dialect as well

- https://github.com/apache/datafusion/blob/6e790cf1f363652c3f2ed64286da75a1d5e88cb9/datafusion-cli/src/exec.rs#L218
- https://github.com/apache/datafusion/blob/6e790cf1f363652c3f2ed64286da75a1d5e88cb9/datafusion/core/src/execution/session_state.rs#L491

maybe generic approach should be taken like in https://github.com/apache/datafusion/blob/6e790cf1f363652c3f2ed64286da75a1d5e88cb9/datafusion-cli/src/helper.rs#L75

alamb:
> maybe generic approach should be taken like in https://github.com/apache/datafusion/blob/6e790cf1f363652c3f2ed64286da75a1d5e88cb9/datafusion-cli/src/helper.rs#L75

I generic approach that lists all the dialects programatically makes a lot of sense to me.

alamb:
I merged this PR up from main and added a small test

PR #1419, opened 2024-09-09
Fix parsing of negative values

agscpp:
Currently, when passing negative numbers, the sql parser returns an error. While the PostgreSQL syntax allows you to specify negative values ​​when creating a sequence.

**Example:**
```
CREATE SEQUENCE seq START -100 INCREMENT -10 MINVALUE -1000 MAXVALUE 1;
```

**Error text:**
```
ParserError("Expected: a value, found: -")
```

**Link to PostgreSQL documentation**:
https://www.postgresql.org/docs/current/sql-createsequence.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7274,7 +7274,7 @@ impl<'a> Parser<'a> {
                 let placeholder = tok.to_string() + &ident.value;
                 Ok(Value::Placeholder(placeholder))
             }
-            tok @ Token::Minus => {
+            tok @ Token::Minus | tok @ Token::Plus => {
                 let next_token = self.next_token();
                 match next_token.token {
                     Token::Number(n, l) => Ok(Value::Number(
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7274,22 +7274,6 @@ impl<'a> Parser<'a> {
                 let placeholder = tok.to_string() + &ident.value;
                 Ok(Value::Placeholder(placeholder))
             }
-            tok @ Token::Minus | tok @ Token::Plus => {
-                let next_token = self.next_token();
-                match next_token.token {
-                    Token::Number(n, l) => {
-                        if tok == Token::Minus {
-                            Ok(Value::Number(
-                                Self::parse(tok.to_string() + &n, location)?,
-                                l,
-                            ))
-                        } else {
-                            Ok(Value::Number(Self::parse(n, location)?, l))
-                        }
-                    }
-                    _ => self.expected("number", next_token),
-                }
-            }
             unexpected => self.expected(
                 "a value",
                 TokenWithLocation {
@@ -7300,6 +7284,7 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse an unsigned numeric literal
     pub fn parse_number_value(&mut self) -> Result<Value, ParserError> {
         match self.parse_value()? {
             v @ Value::Number(_, _) => Ok(v),
@@ -7307,6 +7292,26 @@ impl<'a> Parser<'a> {
             _ => {
                 self.prev_token();
                 self.expected("literal number", self.peek_token())
+            }
+        }
+    }
+
+    /// Parse a numeric literal as an expression. Returns a [`Expr::UnaryOp`] if the number is signed,
+    /// otherwise returns a [`Expr::Value`]
+    pub fn parse_number_value_with_sign(&mut self) -> Result<Expr, ParserError> {
+        let next_token = self.next_token();
+        match next_token.token {
+            Token::Plus => Ok(Expr::UnaryOp {
+                op: UnaryOperator::Plus,
+                expr: Box::new(Expr::Value(self.parse_number_value()?)),
+            }),
+            Token::Minus => Ok(Expr::UnaryOp {
+                op: UnaryOperator::Minus,
+                expr: Box::new(Expr::Value(self.parse_number_value()?)),
+            }),
+            _ => {
+                self.prev_token();
+                Ok(Expr::Value(self.parse_number_value()?))
             }
         }
     }
@@ -11625,29 +11630,29 @@ impl<'a> Parser<'a> {
         if self.parse_keywords(&[Keyword::INCREMENT]) {
             if self.parse_keywords(&[Keyword::BY]) {
                 sequence_options.push(SequenceOptions::IncrementBy(
-                    Expr::Value(self.parse_number_value()?),
+                    self.parse_number_value_with_sign()?,
                     true,
                 ));
             } else {
                 sequence_options.push(SequenceOptions::IncrementBy(
-                    Expr::Value(self.parse_number_value()?),
+                    self.parse_number_value_with_sign()?,
                     false,
                 ));
             }
         }
         //[ MINVALUE minvalue | NO MINVALUE ]
         if self.parse_keyword(Keyword::MINVALUE) {
-            sequence_options.push(SequenceOptions::MinValue(Some(Expr::Value(
-                self.parse_number_value()?,
-            ))));
+            sequence_options.push(SequenceOptions::MinValue(Some(
+                self.parse_number_value_with_sign()?,
+            )));
         } else if self.parse_keywords(&[Keyword::NO, Keyword::MINVALUE]) {
             sequence_options.push(SequenceOptions::MinValue(None));
         }
         //[ MAXVALUE maxvalue | NO MAXVALUE ]
         if self.parse_keywords(&[Keyword::MAXVALUE]) {
-            sequence_options.push(SequenceOptions::MaxValue(Some(Expr::Value(
-                self.parse_number_value()?,
-            ))));
+            sequence_options.push(SequenceOptions::MaxValue(Some(
+                self.parse_number_value_with_sign()?,
+            )));
         } else if self.parse_keywords(&[Keyword::NO, Keyword::MAXVALUE]) {
             sequence_options.push(SequenceOptions::MaxValue(None));
         }
@@ -11656,21 +11661,19 @@ impl<'a> Parser<'a> {
         if self.parse_keywords(&[Keyword::START]) {
             if self.parse_keywords(&[Keyword::WITH]) {
                 sequence_options.push(SequenceOptions::StartWith(
-                    Expr::Value(self.parse_number_value()?),
+                    self.parse_number_value_with_sign()?,
                     true,
                 ));
             } else {
                 sequence_options.push(SequenceOptions::StartWith(
-                    Expr::Value(self.parse_number_value()?),
+                    self.parse_number_value_with_sign()?,
                     false,
                 ));
             }
         }
         //[ CACHE cache ]
         if self.parse_keywords(&[Keyword::CACHE]) {
-            sequence_options.push(SequenceOptions::Cache(Expr::Value(
-                self.parse_number_value()?,
-            )));
+            sequence_options.push(SequenceOptions::Cache(self.parse_number_value_with_sign()?));
         }
         // [ [ NO ] CYCLE ]
         if self.parse_keywords(&[Keyword::NO, Keyword::CYCLE]) {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7277,10 +7277,16 @@ impl<'a> Parser<'a> {
             tok @ Token::Minus | tok @ Token::Plus => {
                 let next_token = self.next_token();
                 match next_token.token {
-                    Token::Number(n, l) => Ok(Value::Number(
-                        Self::parse::<String>(tok.to_string() + &n, location)?,
-                        l,
-                    )),
+                    Token::Number(n, l) => {
+                        if tok == Token::Minus {
+                            Ok(Value::Number(
+                                Self::parse(tok.to_string() + &n, location)?,
+                                l,
+                            ))
+                        } else {
+                            Ok(Value::Number(Self::parse(n, location)?, l))
+                        }
+                    }
                     _ => self.expected("number", next_token),
                 }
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -7274,6 +7274,16 @@ impl<'a> Parser<'a> {
                 let placeholder = tok.to_string() + &ident.value;
                 Ok(Value::Placeholder(placeholder))
             }
+            tok @ Token::Minus => {
```
git-hulk, line 7277: Should handle `+` as well? And for the sign number value, [another place](https://github.com/sqlparser-rs/sqlparser-rs/blob/4875dadbf5e691dd1e8ef179145829d8ca3ef4d8/src/parser/mod.rs#L1160) represents it with `UnaryOp`, guess should keep consistent? cc @iffyio

tests/sqlparser_postgres.rs
```diff
@@ -277,6 +277,16 @@ fn parse_create_sequence() {
         "CREATE TEMPORARY SEQUENCE IF NOT EXISTS name3 INCREMENT 1 NO MINVALUE MAXVALUE 20 OWNED BY NONE",
     );
 
+    let sql7 = "CREATE SEQUENCE name4
```
git-hulk, line 280: This change didn't guard with the PostgreSQL dialect guard, so should move test case to common.

src/parser/mod.rs
```diff
@@ -7274,6 +7274,16 @@ impl<'a> Parser<'a> {
                 let placeholder = tok.to_string() + &ident.value;
                 Ok(Value::Placeholder(placeholder))
             }
+            tok @ Token::Minus | tok @ Token::Plus => {
```
alamb, line 7277: Normally I would expect that the `-` is included as part of the token itself  otherwise negative numbers wouldn't be able to be parsed at all anywhere

I wonder if the issue is that the of token being handled by create sequence needs to be expanded 🤔

alamb:
I started the CI tests and there appear to be some problems with this PR

src/parser/mod.rs
git-hulk: @alamb Not sure if I understand your point correctly. `token::Minus` might be a sign symbol for a number or a binary operation, and this cannot be determined when parsing it into a token.

tests/sqlparser_postgres.rs
git-hulk: I think it should be good to keep it inside PostgreSQL unless more dialects support this syntax.

src/parser/mod.rs
git-hulk: Oh, what I mean is if we should return the negative number as the UnaryOp expression like the previous implementation.

tests/sqlparser_common.rs
```diff
@@ -2819,6 +2819,29 @@ fn parse_window_function_null_treatment_arg() {
     );
 }
 
+#[test]
+fn parse_signed_value() {
+    let sql1 = "CREATE SEQUENCE name1
```
alamb, line 2824: I double checked that this syntax is indeed valid in postgres (🤯 )

```sql
postgres=# CREATE SEQUENCE name1
    AS BIGINT
    INCREMENT   -15
    MINVALUE - 2000  MAXVALUE -50
    START WITH -   60;
CREATE SEQUENCE

postgres=# SELECT *
FROM information_schema.sequences
ORDER BY sequence_name;
 sequence_catalog | sequence_schema | sequence_name | data_type | numeric_precision | numeric_precision_radix | numeric_scale | start_value | minimum_value | maximum_value | increment | cycle_option
------------------+-----------------+---------------+-----------+-------------------+-------------------------+---------------+-------------+---------------+---------------+-----------+--------------
 postgres         | public          | name1         | bigint    |                64 |                       2 |             0 | -60         | -2000         | -50           | -15       | NO
(1 row)
```

src/parser/mod.rs
alamb: My primary concern is that this change seems to have much larger potential scope than simply supporting negative numbers for `CREATE SEQUENCE` -- it seems like it would change parsing of all values. 

However, I am clearly confused -- I added a test that shows  that `-10` is parsed as `-(10)` (aka a unary minus). See  https://github.com/sqlparser-rs/sqlparser-rs/pull/1421. That test still passes on this PR, so it must not have as much of an impact as I thought

tests/sqlparser_common.rs
```diff
@@ -2819,6 +2819,29 @@ fn parse_window_function_null_treatment_arg() {
...
+    AS BIGINT
+    INCREMENT   +10
+    MINVALUE + 30  MAXVALUE +5000
+    START WITH +   45";
+    one_statement_parses_to(
+        sql2,
+        "CREATE SEQUENCE name2 AS BIGINT INCREMENT 10 MINVALUE 30 MAXVALUE 5000 START WITH 45",
+    );
+}
+
```
alamb, line 2844: Can you also please add a test for your original reproducer?

```sql
CREATE SEQUENCE seq START -100 INCREMENT -10 MINVALUE -1000 MAXVALUE 1;
```

I tried adding it locally and it doesn't seem to pass for me

```

CREATE SEQUENCE seq START -100 INCREMENT -10 MINVALUE -1000 MAXVALUE 1;: ParserError("Expected: end of statement, found: INCREMENT")
thread 'parse_create_sequence' panicked at src/test_utils.rs:119:61:
CREATE SEQUENCE seq START -100 INCREMENT -10 MINVALUE -1000 MAXVALUE 1;: ParserError("Expected: end of statement, found: INCREMENT")
stack backtrace:
```

I tested like this:

```diff
andrewlamb@Andrews-MacBook-Pro-2:~/Software/sqlparser-rs$ git diff 99b98ca86cf0591a89a644e06742758a70aa04ea
diff --git a/tests/sqlparser_postgres.rs b/tests/sqlparser_postgres.rs
index 1ebb5d5..89aee71 100644
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -277,6 +277,14 @@ fn parse_create_sequence() {
         "CREATE TEMPORARY SEQUENCE IF NOT EXISTS name3 INCREMENT 1 NO MINVALUE MAXVALUE 20 OWNED BY NONE",
     );

+    // negative increments
+    let sql7 = "CREATE SEQUENCE seq START -100 INCREMENT -10 MINVALUE -1000 MAXVALUE 1;";
+    pg().one_statement_parses_to(
+        sql7,
+        "CREATE TEMPORARY SEQUENCE IF NOT EXISTS name3 INCREMENT 1 NO MINVALUE MAXVALUE 20 OWNED BY NONE",
+    );
+
+
     assert!(matches!(
         pg().parse_sql_statements("CREATE SEQUENCE foo INCREMENT 1 NO MINVALUE NO"),
         Err(ParserError::ParserError(_))
```

Then ran with 
```shell
cargo test --test sqlparser_postgres
```

alamb (commented):
Thank you @agscpp -- this is a very nice contribution.

I think this PR needs a test with the originally reported SQL error (and some examples without spaces between minus signs) but is otherwise good to go from my perspective.

I would appreciate someone else taking a look too (cc @iffyio @jmhain ) as I can't stop worrying this will have unintended consequences

src/parser/mod.rs
git-hulk: > My primary concern is that this change seems to have much larger potential scope than simply supporting negative numbers for CREATE SEQUENCE -- it seems like it would change parsing of all values.

It did affect the parsing of all values.

> However, I am clearly confused -- I added a test that shows that -10 is parsed as -(10) (aka a unary minus). See https://github.com/sqlparser-rs/sqlparser-rs/pull/1421. That test still passes on this PR, so it must not have as much of an impact as I thought

Yes, I also mentioned this in the comment https://github.com/sqlparser-rs/sqlparser-rs/pull/1419#discussion_r1751560914, the negative number will be parsed into an unary operation. So I'm wondering if we should keep it consistent by parsing it as an unary operation in PR as well.

tests/sqlparser_common.rs
alamb: Thanks -- could you please add at least one test with this syntax (no space between `-` and the numbers)? I think this is important coverage as goes down a different parsing path

src/parser/mod.rs
iffyio: It seems when parsing expressions, the signed number literals are handled by the precedence logic when looking to parse a sub expr [here](https://github.com/sqlparser-rs/sqlparser-rs/blob/cb0c511b05f4acc5867dce6e8b00ec38884cd058/src/parser/mod.rs#L1158-L1172) is where we usually land. From what I could tell, we don't end up hitting the new code when parsing expressions due to calling `parse_prefix` beforehand - though I can't tell how well the change works with dialect overrides to the precedence logic so there might be some potential for surprising behavior. Also can't say for sure if there are other considerations other than when parsing subexpressions

The doc for [Token::Number](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/tokenizer.rs#L54-L55) at least explicitly says it contains unsigned number and since that's what `Value::Number` maps to, I think it might be reasonable to continue to always represent signed numbers as unary Exprs vs sometimes baked into the Value.

For this PR, issue seems that the clauses in [CREATE SEQUENCE](https://github.com/sqlparser-rs/sqlparser-rs/blob/cb0c511b05f4acc5867dce6e8b00ec38884cd058/src/parser/mod.rs#L11658) specifically are looking to parse number literals but the [parse_number_value](https://github.com/sqlparser-rs/sqlparser-rs/blob/cb0c511b05f4acc5867dce6e8b00ec38884cd058/src/parser/mod.rs#L7333) helper currently looks to parse exactly a `Value::Number`, maybe they can instead be updated to call `parse_expr`, expecially since the expected types on the enum are already [Expr](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L4712-L4715).

There are few similar calls to `parse_number_value()` I suspect they could have similar issue of rejecting signed numbers, so if we end up keeping the current behavior, a follow up might be to make it explicit what `parse_number_value` returns/rejects.

src/parser/mod.rs
```diff
@@ -7295,6 +7295,29 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_number_value_with_sign(&mut self) -> Result<Expr, ParserError> {
```
iffyio, line 7298:
```suggestion
    pub fn parse_number(&mut self) -> Result<Expr, ParserError> {
```
Also thinking especially since this is made a public function, that we should write a doc comment about what it does? and potentially mention that the return type for signed is a Unary Expr

src/parser/mod.rs
```diff
@@ -7295,6 +7295,29 @@ impl<'a> Parser<'a> {
...
+                        op: UnaryOperator::Plus,
+                        expr: Box::new(Expr::Value(self.parse_number_value()?)),
+                    })
+                } else {
+                    Ok(Expr::UnaryOp {
+                        op: UnaryOperator::Minus,
+                        expr: Box::new(Expr::Value(self.parse_number_value()?)),
+                    })
+                }
+            }
```
iffyio, line 7313:
```suggestion
            Token::Minus => {
                Ok(Expr::UnaryOp {
                    op: UnaryOperator::Minus,
                    expr: Box::new(Expr::Value(self.parse_number_value()?)),
                })
            }
            Token::Plus => {
                Ok(Expr::UnaryOp {
                    op: UnaryOperator::Plus,
                    expr: Box::new(Expr::Value(self.parse_number_value()?)),
                })
            }
```
thinking we can avoid the double work of match+comparison?

src/parser/mod.rs
```diff
@@ -7295,6 +7296,26 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a numeric literal and return Expr::UnaryOp if the number is signed,
+    /// otherwise return Expr::Value
```
iffyio, line 7300:
```suggestion
    /// Parse a numeric literal as an expression. Returns a [`Expr::UnaryOp`] if the number is signed,
    /// otherwise returns a [`Expr::Value`]
```

src/parser/mod.rs
iffyio: Can we update the method name as well? the current `parse_number_value_with_sign` somewhat suggests that only signed numbers will be parsed which might be misleading

iffyio (approved):
Left a couple minor comments, otherwise LGTM!

src/parser/mod.rs
```diff
@@ -11609,29 +11630,29 @@ impl<'a> Parser<'a> {
         if self.parse_keywords(&[Keyword::INCREMENT]) {
             if self.parse_keywords(&[Keyword::BY]) {
                 sequence_options.push(SequenceOptions::IncrementBy(
-                    Expr::Value(self.parse_number_value()?),
+                    self.parse_number_value_with_sign()?,
```
alamb, line 11633: ❤️

tests/sqlparser_common.rs
alamb: > By the way, do tests really need to be moved to a common module? Tests for sequences are currently written only for PostgreSQL

That seems like it might be a good idea to make them more discoverable

alamb (approved):
Thank you @agscpp  as well as @iffyio and @git-hulk  for the review

I think there are a few more suggestions about comments from @iffyio  but otherwise I think this PR is good to go

src/parser/mod.rs
iffyio: Ah so I was thinking as in the original suggestion - if it would make sense to just call this `parse_number()`?

iffyio (approved):
LGTM!

git-hulk (approved):

alamb:
🚀  -- thanks everyone

PR #1427, opened 2024-09-13
feat: adding Display implementation to DELETE and INSERT

### LINKED ISSUE #1420: Add `Display` implementation for other `DML` operations
Much like `Select` and `CreateTable`, the ability to programmatically construct queries from AST parts is incredibly useful.

However, `Insert`, `Delete`, and `Alter` do not implement Display, making it difficult to produce these statements as strings.

seve-martinez:
Closes #1420 .


src/ast/dml.rs
```diff
@@ -487,6 +487,90 @@ pub struct Insert {
     pub insert_alias: Option<InsertAliases>,
 }
 
+impl Display for Insert {
```
iffyio, line 490: I'm thinking we want to reuse the display implementation in the corresponding insert/delete statement display as well? Otherwise the new code is largely a duplicate of what we do there, and we'd need identical test cases which might not be worth it

seve-martinez: Oh wow I had no idea these existed 😅 

Asking to learn: How would I reuse what's already here in the other statements? Or are you just saying that instead of this implementation, it'd be better for me to use construct a `Statement::Insert` if I needed to do some programmatic construction

iffyio: Ah so I was rather thinking, taking insert as an example, that if we have the display implementation here, then we should be able to replace [this code](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L3402-L3484) with something roughly like
```rust
Statement::Insert(insert) => write!(f, "{insert}")
```
since conceptually the representations should be the same?

iffyio:
@seve-martinez could you take a look at the ci faiiure? I think you might need to fetch the latest changes from main for the codestyle step

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @iffyio  and @seve-martinez

alamb:
Looks like we have a CI failure to deal with now

alamb:
As I am preparing a release today I pushed a commit to get this CI passing

alamb:
Thanks again @seve-martinez and @iffyio

PR #1428, opened 2024-09-14
Allow to use ON CLUSTER cluster_name in TRUNCATE syntax

git-hulk:
See https://clickhouse.com/docs/en/sql-reference/statements/truncate#truncate-table

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -708,6 +708,8 @@ impl<'a> Parser<'a> {
             };
         };
 
+        let on_cluster = self.parse_optional_on_cluster()?;
+
         Ok(Statement::Truncate {
             table_names,
             partitions,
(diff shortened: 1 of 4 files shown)


git-hulk:
cc @iffyio

src/parser/mod.rs
```diff
@@ -708,13 +708,16 @@ impl<'a> Parser<'a> {
             };
         };
 
+        let on_cluster = self.parse_optional_on_cluster()?;
```
git-hulk, line 711: ON CLUSTER is a ClickHouse-specific syntax, but we don't add the dialect guard for the sake of keeping consistent with other places.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank yoU @iffyio  and @git-hulk

Issue #1430, opened 2024-09-14
stop the use of `dialect_of` and referring to specific dialects in parser


samuelcolvin:
TL;DR; — I want to stop `Parser` from referring to specific dialects, and instead consult the dialect about specific behaviour.

In particular, use of `dialect_of` and custom dialect behaviour in `Parser` should be removed

Abstract reasons to do this:
* it's very esoteric in Rust, I've never seen another library use this approach, It's like reading (bad) python, in Rust
* it's poor separation of concerns — the Parser should concentrate on parsing SQL, and defer to the dialect about how to behave in some scenarios, rather than holding information about how mysql differs from Snowflake etc.

Concrete reasons to do this:
* I want to use "from  first" syntax (#1400) with otherwise postgres syntax — currently dialects can't opt into certain behaviour because the parser is checking for a specific dialect, not for a given behaviour

I've already discussed this with @alamb, I'll try to do this in multiple PRs, to make it easier to review.

iffyio:
Linking for context related discussion https://github.com/sqlparser-rs/sqlparser-rs/discussions/1340

PR #1431, opened 2024-09-14
move primitive dialect configuration into a struct

samuelcolvin:
Step towards #1430.

I personally think this should make it much easier to add more flags to dialects. But I can see it might not be to everyone's taste.

So I want confirmation it's a good idea before prcoeeding.

This should:
* make it less verbose to add more options to dialects
* make it easier to define dialects that inherit from other dialects.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -69,7 +69,7 @@ macro_rules! dialect_of {
 /// * It must be a const setting, not value that's dependent on context - that has to be a function
 /// * have a default that matches `Default::default()`
 #[derive(Debug, Default, Clone)]
-pub struct DialectSettings {
+pub struct DialectFlags {
     /// Determine if the dialect supports escaping characters via '\' in string literals.
     ///
     /// Some dialects like BigQuery and Snowflake support this while others like
@@ -253,8 +253,8 @@ pub trait Dialect: Debug + Any {
         self.type_id()
     }
 
-    fn settings(&self) -> DialectSettings {
-        DialectSettings::default()
+    fn flags(&self) -> DialectFlags {
+        DialectFlags::default()
     }
 
     /// Determine if a character starts a quoted identifier. The default
@@ -295,7 +295,7 @@ pub trait Dialect: Debug + Any {
 
     /// Does the dialect support trailing commas in the projection list?
     fn supports_projection_trailing_commas(&self) -> bool {
-        self.settings().supports_trailing_commas
+        self.flags().supports_trailing_commas
     }
 
     /// Dialect-specific infix parser override
@@ -610,15 +610,17 @@ mod tests {
         struct WrappedDialect(MySqlDialect);
 
         impl Dialect for WrappedDialect {
-            fn settings(&self) -> DialectSettings {
-                let s = self.0.settings();
-                DialectSettings {
+            fn flags(&self) -> DialectFlags {
+                let s = self.0.flags();
+                DialectFlags {
                     supports_filter_during_aggregation: s.supports_filter_during_aggregation,
-                    supports_within_after_array_aggregation: s.supports_within_after_array_aggregation,
+                    supports_within_after_array_aggregation: s
+                        .supports_within_after_array_aggregation,
                     supports_group_by_expr: s.supports_group_by_expr,
                     supports_in_empty_list: s.supports_in_empty_list,
                     convert_type_before_value: s.convert_type_before_value,
-                    supports_string_literal_backslash_escape: s.supports_string_literal_backslash_escape,
+                    supports_string_literal_backslash_escape: s
+                        .supports_string_literal_backslash_escape,
                     ..Default::default()
                 }
             }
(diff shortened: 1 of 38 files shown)


src/dialect/mod.rs
```diff
@@ -675,7 +610,20 @@ mod tests {
         struct WrappedDialect(MySqlDialect);
 
         impl Dialect for WrappedDialect {
-            fn dialect(&self) -> std::any::TypeId {
+            fn settings(&self) -> DialectSettings {
+                let s = self.0.settings();
+                DialectSettings {
```
samuelcolvin, line 615: this is slightly ugly because `WrappedDialect` only implemented a few of the methods on the dialect.

But this actually demonstrates one of the advtanges of the approach in this PR:

If this test had been written with the new behaviour, this method would become

```rs
fn settings(&self) -> DialectSettings {
  let mut s = self.0.settings();
  s.custom_thing = true;
  s
}
```

And it would stay up-to-date even as new setting were added. Unlike the current behaviour.

I can update this test if you like?

src/dialect/mod.rs
```diff
@@ -691,64 +639,40 @@ mod tests {
...
             }
 
-            fn supports_string_literal_backslash_escape(&self) -> bool {
-                self.0.supports_string_literal_backslash_escape()
-            }
-
             fn is_proper_identifier_inside_quotes(
                 &self,
-                chars: std::iter::Peekable<std::str::Chars<'_>>,
+                chars: Peekable<Chars<'_>>,
```
samuelcolvin, line 644: it would be great if clippy or rustfmt could warn about these unnecessary paths, rustrover warns, hence the drive by change.

samuelcolvin (commented):
overall this PR is significantly negative LOC — showing that this approach is at least less verbose.

src/dialect/mod.rs
```diff
@@ -107,6 +253,10 @@ pub trait Dialect: Debug + Any {
         self.type_id()
     }
 
+    fn flags(&self) -> DialectFlags {
+        DialectFlags::default()
+    }
```
samuelcolvin, line 258: I don't like creating a new struct on every call (and there are a lot).

I'd rather store the struct on each dialect, then the signature here can before

```rs
fn flags(&self) -> &DialectFlags {
  &self.settings
}
```

The problem then is that every struct needs a `fn new()` or impl `Default`, and the signature to create all dialects changes everywhere, but if we're happy with that breaking change, the whole thing would become much cleaner. WDYT?

samuelcolvin: done.

src/dialect/mod.rs
```diff
@@ -63,6 +63,152 @@ macro_rules! dialect_of {
     };
 }
 
+/// Constant settings for a dialect.
+///
+/// For configuration to be defined here, rather than on the rate:
```
iffyio, line 68: > on the rate

was this a typo?

src/dialect/mod.rs
```diff
@@ -63,6 +63,152 @@ macro_rules! dialect_of {
...
+/// * It must be a const setting, not value that's dependent on context - that has to be a function
+/// * have a default that matches `Default::default()`
+#[derive(Debug, Default, Clone)]
+pub struct DialectFlags {
```
iffyio, line 72: Would it make sense to call it e.g. `DialectSettings`, `DialectConstants` or similar if we envision this potentially containing more than boolean flags?

iffyio (commented):
@samuelcolvin Thanks for tackling this! The changes look very reasonable to me

cc @alamb also @jmhain I think you had a similar idea for this, in case you have any thoughts

src/dialect/mod.rs
samuelcolvin: Happy to change a name.

alamb (commented):
Thank you @samuelcolvin and @iffyio 

In my opinion:
1. Adding `impl Default` for the dialects is nice as it is sort of a usability papercut (it would also nice to have a `new()` function, to follow convention)
2. In terms of `DialectFlags` I personally think the improvement is not worth the API churn, though I don't really know how many people have their own Dialect trait implementations 🤔

samuelcolvin:
> I personally think the improvement is not worth the API churn

Okay, I'll leave this open for a few days in case anyone else is particularly keen for it, otherwise  I'll close this and work on #1430 by adding lots of boolean functions.

alamb:
> Okay, I'll leave this open for a few days in case anyone else is particularly keen for it, otherwise I'll close this and work on https://github.com/sqlparser-rs/sqlparser-rs/issues/1430 by adding lots of boolean functions.

Maybe we should contemplate creating a parser that parameterized on the dialect  (`Parser<Dialect>` -- I know we have discussed this before and it likely isn't a huge deal)

jmhain:
> Maybe we should contemplate creating a parser that parameterized on the dialect (Parser<Dialect> -- I know we have discussed this before and it likely isn't a huge deal)

This would cause a lot of code bloat for use cases that require parsing many different dialects. What's the benefit of doing it?

samuelcolvin:
What do you mean by "a lot of code bloat"?

Surely you can create one parser for each dialect? The memory overhead would be tiny.

Avoiding dynamic dispatch would lead to better performance, smaller binaries and (most importantly) more canonical code in the library.

alamb:
> What do you mean by "a lot of code bloat"?
> 
> Surely you can create one parser for each dialect? The memory overhead would be tiny.
> 
> Avoiding dynamic dispatch would lead to better performance, smaller binaries and (most importantly) more canonical code in the library.

I think the idea is if we made a parser you could end up with multiple copies of the Parser code for each one you instantiated -- like  `Parser<GenericDialect>` and `Parser<BigQueryDialect>` 

I think one workaround would be to make your own `DynamicDialect` that dispatched dynamically and have `Parser<DynamicDialect>`

However I think I may have lead us offtopic here. The dynamic dispatch has always kind of annoyed me but I have never had any actual justification to change it. Sorry for starting this discusision

alamb:
Converting to a draft as I don't think this PR is waiting on review anymore

samuelcolvin:
Closing as I don't think we're going this way anymore.

PR #1432, opened 2024-09-18
MS SQL Server: add support for IDENTITY column option

7phs:
This PR implements support [IDENTITY](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property?view=sql-server-ver16) column option of T-SQL (MS SQL Server). Example:

```
CREATE TABLE [dbo].[table_name]
(
  [columnA] int IDENTITY(1,1) NOT NULL,
  [columnB] nvarchar(255) NULL
)
```

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -1050,6 +1050,20 @@ impl fmt::Display for ColumnOptionDef {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct IdentityProperty {
+    pub seed: Expr,
+    pub increment: Expr,
+}
+
+impl fmt::Display for IdentityProperty {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}, {}", self.seed, self.increment)
+    }
+}
+
 /// `ColumnOption`s are modifiers that follow a column definition in a `CREATE
 /// TABLE` statement.
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -1126,7 +1140,7 @@ pub enum ColumnOption {
     /// IDENTITY [ (seed , increment) ]
     /// ```
     /// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
-    Identity(Option<SqlOption>),
+    Identity(Option<IdentityProperty>),
 }
 
 impl fmt::Display for ColumnOption {
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -1120,6 +1120,13 @@ pub enum ColumnOption {
     /// [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#view_column_option_list
     /// [2]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#column_option_list
     Options(Vec<SqlOption>),
+    /// MS SQL Server specific: Creates an identity column in a table.
+    /// Syntax
+    /// ```sql
+    /// IDENTITY [ (seed , increment) ]
+    /// ```
+    /// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+    Identity(Option<SqlOption>),
 }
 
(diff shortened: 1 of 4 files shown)


src/ast/ddl.rs
```diff
@@ -1120,6 +1120,13 @@ pub enum ColumnOption {
     /// [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#view_column_option_list
     /// [2]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#column_option_list
     Options(Vec<SqlOption>),
+    /// MS SQL Server specific: Creates an identity column in a table.
+    /// Syntax
+    /// ```sql
+    /// IDENTITY [ (seed , increment) ]
+    /// ```
+    /// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+    Identity(Option<SqlOption>),
```
bombsimon, line 1129: Is this correct to have a `SqlOption` in the identity? Thinking this would allow a lot of invalid identity values where the only possible value from the docs seems to be a tuple so `Identity(Option<(Value, Value)>)`? That way we don't have to add `Identity` both here as a `ColumnOption` and as a new `SqlOption`

7phs: Thanks for the suggestion!

You’re correct that `Identity(Option<(Value, Value)>)` would potentially prevent invalid identity values. However, implementing `Identity(Option<(Value, Value)>)` as proposed presents some challenges. Specifically, the traits `ast::visitor::VisitMut` and `ast::visitor::Visit` are not implemented for `(ast::value::Value, ast::value::Value)`, which makes it difficult to use in this form.

An alternative approach could be to introduce a dedicated struct like:

```rust
pub struct IdentityColumn {
    pub seed: Value,
    pub increment: Value,
}
```
And then use it as `Identity(Option< IdentityColumn>)`.

Let me know what you think!

bombsimon: I think that sounds like a better idea compared to adding the variant to `SqlOption` but maybe @iffyio have some input before you go that route?

iffyio: Yeah having it as a dedicated type sounds reasonable (maybe call it `IdentityProperty` instead of column since the former is mentioned in the docs - I'm guessing the column is more of the effect)?

src/ast/mod.rs
```diff
@@ -5818,6 +5818,13 @@ pub enum SqlOption {
         range_direction: Option<PartitionRangeDirection>,
         for_values: Vec<Expr>,
     },
+    /// MS SQL Server specific: Optional parameters of identity column
+    /// E.g.
+    ///
+    ///   IDENTITY(1, 2)
+    ///
+    /// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+    Identity { seed: Value, increment: Value },
```
iffyio, line 5827: One thing regarding the types, if the seed or increment can be a negative number we would want to represent it as an expression because the Value type isn't able to represent those - see [parse_number()](https://github.com/sqlparser-rs/sqlparser-rs/blob/04a53e57538524a9cda8ca7304312113849aec52/src/parser/mod.rs#L7454)

src/parser/mod.rs
```diff
@@ -6043,6 +6043,20 @@ impl<'a> Parser<'a> {
             && dialect_of!(self is MySqlDialect | SQLiteDialect | DuckDbDialect | GenericDialect)
         {
             self.parse_optional_column_option_as()
+        } else if self.parse_keyword(Keyword::IDENTITY)
+            && dialect_of!(self is MsSqlDialect | GenericDialect)
+        {
+            let parameters = if self.expect_token(&Token::LParen).is_ok() {
```
iffyio, line 6049: We can replace the `expect_token().is_ok()` with `consume_token()`?

tests/sqlparser_mssql.rs
```diff
@@ -908,6 +908,114 @@ fn parse_create_table_with_invalid_options() {
...
+                    name: None,
+                    option: ColumnOption::NotNull,
+                },
+            ],
+        ),
+    ];
+
+    for (sql, column_options) in with_column_options {
+        assert_eq!(
+            ms().verified_stmt(sql),
```
iffyio, line 953:
```suggestion
            ms_and_generic().verified_stmt(sql),
```

src/ast/ddl.rs
7phs: Thank you for suggestions! I updated it with a dedicated type `IdentityProperty`

src/parser/mod.rs

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @7phs  and @iffyio

Issue #161, opened 2020-04-17
Add Span to Tokens, AST nodes


mullr:
I'd like to use this library for some refactoring and analysis tools. It would be very helpful to have source line/column information available on the AST nodes, something along the lines of `Span` from `syn`. (https://docs.rs/syn/1.0.17/syn/spanned/trait.Spanned.html)

I see that line/column is already tracked in the tokenizer. I'm thinking something along the lines of:

```rust
struct LineColumn {
    line: usize,
    column: usize,
}

struct Span {
  start: LineColumn,
  end: LineColumn,
}

struct TokenOccurence {
  token: Token,
  span: Span
}

struct Query {
  span: Span,
  pub ctes: Vec<Cte>,
 ...
}
```

The main objection I can think of is that it some bulk to `Token`s, and to each AST node. This could be alleviated, if desired, by hiding them behind a feature flag. I looked at `syn` to see how they do it; the approach there is very similar. Afict, their feature flagging doesn't actually affect the node representation / token representation. They just carry it all the time. 

Since the implementation here doesn't appear to be hyper-optimized, and since it's good enough for syn, it seems like this approach should be okay. 

If I submitted a PR along these lines, would you consider accepting it?

If I submitted a PR along these lines, could it possibly be accepted?

nickolay:
> If I submitted a PR along these lines, would you consider accepting it?

I hope @andygrove and @benesch would comment.

As I'm also interested in analysis, I'd love something like this implemented. 

Do you think this can be implemented without adding too much boilerplate? (In the tests, in particular, they're rather verbose as it is.)


[...]

PR #1435, opened 2024-09-19
Implement `Spanned` to retrieve source locations on AST nodes

Nyrox:
This PR adds a new trait `Spanned` to retrieve source spans on AST nodes, see #161 , by recursively traversing the AST and combining spans. This approach is in contrast to the one taken in #839 and #790 by trying to minimise the amount of breaking changes for downstream users, by avoiding wrapping everything in `WithSpan<T>`.

## Main Changes

- Spanned trait and Span type added
- TokenWithLocation now stores Span, instead of Location
- Ident now stores a source span (omitted from Partial/-EQ for compatibility and tests)
- Some AST nodes store source tokens where not intrusive
- Parser::parse_keyword_token added

The general philosophy of the PR is to be "good enough" without breaking things. As a result certain expressions will have broken or incorrect spans. I imagine these can be cleaned up in future PRs (which might require breaking changes).

f.e. many expressions do not include keywords in their span i.e.
```sql
<expr> IS NOT NULL
|....|
```  
  
will have it's source span simply reported as `<expr>::span` and there is many such cases, some of which are easier to fix than others. For expressions we can't generate spans for, I use `Span::EMPTY` as a sort of sentinel value to indicate missing information.

With this approach the only downstream changes a user should have to do to upgrade, should be adding additional fields when matching on AST nodes.

## Future Work

- Store spans for `ast::value::Value`. This seems like a breaking change to me, which would require a `WithSpan<T>` like type
- Store keyword TokenWithLocation for expressions that currently don't have them
- Implement spans for the rest of the AST, namely Statements. I focused on getting Queries done.


Nyrox:
@alamb Hey, we've started using this functionality internally with pretty great success so far. It's still a draft for now, because of the missing todo!'s, but I would appreciate feedback on the overall design and if you think this can get merged in the foreseeable future once issues are addressed. 😄

alamb:
Thanks @Nyrox  -- I'll try and take a look over the next few days

cc @lovasoa @iffyio and @jmhain

src/parser/mod.rs
```diff
@@ -3188,19 +3196,24 @@ impl<'a> Parser<'a> {
...
-            found.location
+            found.span.start
         )
     }
 
     /// If the current token is the `expected` keyword, consume it and returns
     /// true. Otherwise, no tokens are consumed and returns false.
     #[must_use]
-    pub fn parse_keyword(&mut self, expected: Keyword) -> bool {
+    pub fn parse_keyword(&mut self, expected: Keyword) -> Option<TokenWithLocation> {
```
lovasoa, line 3206: Wouldn't it be easier (and less breaking) to keep the existing `parse_keyword` function that returns a bool and add a new `get_keyword` or `parse_keyword_token` or `parse_keyword_with_loc` that returns a token with location ?

Nyrox: That's fine with me, I just figured no reason to have two functions that do the same thing, but one just omits information. And going forward I expect more and more places in the parser to want the Token version (error handling as well as more source spans). I don't expect any downstream users to be using internal parser methods, but I might be wrong. 😅 Hyrum's Law and all

yuyang-ok:
How is this `PR` going? is it going to merge or what?

Nyrox:
Alright, I have now gotten rid of all the todo!s and warnings and un-drafted the PR. There is a lot of missing implementations of spans and I have documented those (in hindsight maybe a derive macro would have been the way to go here, someone better at writing those than me is free to take a shot at it 😅 ).

src/parser/mod.rs
Nyrox: I reverted the change and added `parse_keyword_token` as via your suggestion. I also changed `parse_keyword` to be implemented as `parse_keyword_token(keyword).is_some()` to not duplicate code. 😃

iffyio (commented):
Thanks for tackling this @Nyrox! took a quick look and left some comments inline, I'll make some time to do another pass

src/ast/mod.rs
```diff
@@ -156,6 +179,30 @@ impl Ident {
...
+            quote_style: None,
+            span,
+        }
+    }
+
+    pub fn with_quote_and_span<S>(quote: char, span: Span, value: S) -> Self
+    where
+        S: Into<String>,
+    {
+        assert!(quote == '\'' || quote == '"' || quote == '`' || quote == '[');
```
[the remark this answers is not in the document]
Nyrox: Removing it sounds good to me, I just wanted to have the same behaviour as `Ident::with_quote` which still has the check, so it should prob. be removed in both places then

src/ast/spans.rs
```diff
@@ -0,0 +1,1768 @@
...
+    pub fn test_union() {
+        let query = crate::parser::Parser::new(&GenericDialect)
+            .try_with_sql(
+                "SELECT a FROM postgres.public.source UNION SELECT a FROM postgres.public.source",
+            )
+            .unwrap()
+            .parse_query()
+            .unwrap();
+
+        query.span();
```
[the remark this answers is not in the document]
Nyrox: The intent here was to assert it doesn't panic 😅 I have replaced the old tests with slightly better ones

src/ast/spans.rs
```diff
@@ -0,0 +1,1768 @@
...
+            .try_with_sql(
+                "SELECT id, name FROM users LEFT JOIN companies ON users.company_id = companies.id",
+            )
+            .unwrap()
+            .parse_query()
+            .unwrap();
+
+        assert_eq!(
+            query.span(),
+            Span::new((1, 1).into(), (1, 109 - 28 + 1).into())
```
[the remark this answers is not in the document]
Nyrox: Smth. like that. I replaced it with smarter tests

src/ast/spans.rs
```diff
@@ -0,0 +1,1798 @@
...
+    fn span(&self) -> Span {
+        self.expr.span().union(&self.alias.span)
+    }
+}
+
+impl Spanned for OrderByExpr {
+    fn span(&self) -> Span {
+        self.expr
+            .span()
+            .union_opt(&self.with_fill.as_ref().map(|f| f.span()).clone())
```
iffyio, line 1532:
```suggestion
            .union_opt(&self.with_fill.as_ref().map(|f| f.span()))
```

src/ast/spans.rs
```diff
@@ -0,0 +1,1798 @@
...
+                stride,
+            } => union_spans(
+                [
+                    lower_bound.as_ref().map(|i| i.span()),
+                    upper_bound.as_ref().map(|i| i.span()),
+                    stride.as_ref().map(|i| i.span()),
+                ]
+                .iter()
+                .flatten()
+                .cloned(),
```
iffyio, line 1205:
```suggestion
                .into_iter()
                .flatten()
```

iffyio (commented):
Thanks @Nyrox! The changes look reasonable to me overall given the discussion in the GH issue. Left some comments, one mostly wondering around the equality behavior now that the token location is embedded within the AST

tests/sqlparser_common.rs
```diff
@@ -9805,6 +9853,7 @@ fn parse_map_access_expr() {
 #[test]
 fn parse_connect_by() {
     let expect_query = Select {
+        select_token: TokenWithLocation::wrap(Token::make_keyword("SELECT")),
```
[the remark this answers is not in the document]
Nyrox: Yeah I agree it can be a bit tricky. For this particular case it doesn't matter I think, because TokenWithLocation's custom Eq implementation ignoring spans. So as long as selects actually have _a_ select token they will compare.
What actually is the use case of having Eq on ast nodes? I guess it's so you can throw them into HashMaps?

src/ast/mod.rs
```diff
@@ -132,8 +137,25 @@ pub struct Ident {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.value == other.value && self.quote_style == other.quote_style
+    }
+}
+
+impl core::hash::Hash for Ident {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.value.hash(state);
+        self.quote_style.hash(state);
+    }
```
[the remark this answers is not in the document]
Nyrox: 👍 good idea, done

tests/sqlparser_common.rs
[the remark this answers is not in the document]
iffyio: Ah, no my mistake I missed that the span is being ignored in the comparison here as well. the actual token isn't currently ignored however? thinking that potentially makes it awkward needing to figure out what a token on a property should be, and not sure if there are guarantees provided by the parser on its value? (I'm assuming having the token included in the node is useful)

Not only Eq, but PartialEq as well, basically thinking as with today, any two subexpressions should remain equal regardless of what part of the tree they show up in (for example required when taking fingerprints of sql and comparing them)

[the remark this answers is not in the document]
Nyrox: Hmm yeah. I mean "Token" in most of these cases is going to be Keyword, so there is some certainty to them being predictable. I.e. if two queries have different tokens, I would very much expect them to actually be different queries.

But I also get if it opens too many cans of worms, so we could just omit it from the trait implementations again, it's just a bit annoying then that we would have to start writing custom implementations for every type that is storing a Token.

The Token is useful, because you need to store it (or a at least it's Span) so you can include it for reported spans. 
I.e. `select a from b`, if you don't store the token for 'select' can't include that as part of it's span :/

src/ast/spans.rs
```diff
@@ -0,0 +1,1798 @@
...
+
+impl Spanned for Function {
+    fn span(&self) -> Span {
+        union_spans(
+            self.name
+                .0
+                .iter()
+                .map(|i| i.span)
+                .chain(iter::once(self.args.span())),
+        )
```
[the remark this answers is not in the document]
Nyrox: Good point, I'll go over the ones that are not doing this yet and fix them up 👍

src/ast/mod.rs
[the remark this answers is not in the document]
Nyrox: What's the verdict here? Remove or not? :P I am not sure it's in the scope of this PR tbh.

src/ast/spans.rs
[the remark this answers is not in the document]
Nyrox: Done. Thanks for the suggestion, I'll definitely stash this technique in my memory. 😄

src/ast/mod.rs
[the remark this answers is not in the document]
iffyio: Ah I didn't realise the behavior exists already, it sounds reasonable to follow suit. The current behavior feels unusual to me though, and I agree that it'd be out of scope for this PR to change it. I can take a closer look at it and we can have the behavior here match the outcome

tests/sqlparser_common.rs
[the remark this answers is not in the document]
iffyio: Yeah I agree it can become unweildy to try to workaround this by needing custom impls for each type, also as you mention its probably a reasonable assumption that the queries are different if they have different tokens (given that they come from the same parser)

src/ast/spans.rs
```diff
@@ -40,28 +41,85 @@ pub trait Spanned {
...
+        let Query {
+            with,
+            body,
+            order_by,
+            limit,
+            limit_by,
+            offset,
+            fetch,
+            locks: _,         // todo
+            for_clause: _,    // todo, mysql specific
```
lovasoa, line 53: @Nyrox : `for_clause` is specific to microsoft sql server (mssql), not mysql :)

Nyrox: oops 😅 , fixed

Nyrox:
Bump 😅

lustefaniak:
Hi @Nyrox!
Thanks for this PR and pushing locations support further! 

We had an internal discussion to try to upstream the location changes we use in SYNQ. I found some time to finish rebase today and then noticed your PR. (https://github.com/apache/datafusion-sqlparser-rs/pull/1480)

I like your suggested changes, as they are very noninvasive for the downstream consumers. If this could be included in the main, that would be absolutely amazing. Having location inside every Ident is probably best. In my PR, I wrapped all Idents with location, but our fork uses them only when we need them. Having it always available inside of `Indent` sounds much better. 

It would also be good to design and include wrapping of nodes. It is essential, especially for enum cases where adding location to each of the enum cases would be too verbose. Also in some places those "smaller" locations just don't make much sense. In my PR, I'm using https://github.com/apache/datafusion-sqlparser-rs/pull/1480/commits/fc8e44afd66c3a5100a6546d8edba36c65870315#diff-fc6a8d66b8cb6bd48a119a70e489cbcef82e7f551e7cf08e8e972ef4e774ef49R12259-R12308 to align captured span to a non-whitespace portion of the query based on token indexes.

Nyrox:
Hi @lustefaniak, thanks I remember seeing your fork when initially researching this 😅 
I agree something like a `WithSpan<T>` type will probably be needed to improve spans, especially in the case of enums as you said. I think for now the focus, as by the discussions with @alamb was mostly on just reducing the breaking surface as much as we can and get _something_ in. Maybe in a future PR we can start thinking about wrapping all the enums, but then at least we have a gradual roll out 😃

Nyrox:
Hey all, think we've gotten quite far here - is there anything actively blocking this from getting approved? Are there any actions I can take to help you guys or do you just need the time to get to this item? 😊

iffyio:
@Nyrox sorry for the delay in getting to this

I don't have any major comments regarding the impl and it seems to align with previous discussions so I'm suspecting this is close.

One aspect I'm not super clear on is how accurate/comprehensive we'll look to be regarding this feature going forward (primarily as it pertains to testing affected code paths), i.e. whether as a best effort, or similar to the other parser functionality (which expectedly requires more effort from PR authors when adding/updating features in the codebase) in order to properly power use cases like @lustefaniak mentions in #1480, or somewhere in-between.

In any case, I would wait for @alamb for when time allows to take a look and provide any thoughts/feedback on the overall approach!

alamb:
Thanks @Nyrox and @iffyio -- i will try and find time this weke to review this propsoal

mkarbo:
@alamb that's really appreciated 🙌

src/ast/mod.rs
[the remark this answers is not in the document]
Nyrox: Can I close this comment, if this is gonna be done in a future PR or should we scram it in here if a decision is made? 😄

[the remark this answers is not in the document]
iffyio: Yeah let's ignore is for this PR

alamb:
Sorry for my silence / inability to find time to review this PR in depth. I am struggling with too many thing

TLDR I I think we should:
1. Get the tests passing on this PR and merge it in (if not all AST nodes have spans initially, we can file tickets to add the information over the next few releases)
2. Add some documentation explaining what is happening and how downstream consumers need to change (I think basically by ignoring the newly added fields

@iffyio  and I spoke about this a bit -- basically in my opinion adding source locations to sqlparser-rs is one of the last major remaining features it is missing.

However, it also has the potential to be very disruptive to downstream consumers.

It seems to me the requirements for this change are:
1. Any changes required by users of this crate should be mechanical
3. We should clearly document what changes are needed on upgrade (ideally with examples)

An example of a mechanical change might be adding
```rust
match node {  
  ast::Query { 
    field1,
    field2, 
    location: _,  // add a new line to ignored location
}
```

An example of a non mechanical change might be wrapping all AST nodes in some other struct so the structure of the `match` no longer works

I looked at this PR and I think it basically satisfies 1 for me so I think we should proceed. Would it be possible to write up something, perhaps in the `README` that explains  that we are adding new spans to the code and how we will handle them?

Nyrox:
Okay, I wrote some documentation and fixed up the tests so CI should pass now (I hope :P )
Let me know what you think of the docs, I wasn't quite sure what to write so I hope it's decent.

@iffyio I ended up running into the problem we discussed further up about comparing queries. It slipped by me that Tokens store the raw string in addition to the Keyword enum, causing `select 5` and `SELECT 5` to compare as unequal. I have added a new wrapper type `IgnoreField` to work around this issue and make sure the tokens actually do not impact semantics. I am not exactly sure if this is _the best_ fix, but I think if we want queries to compare as semantically equal rather than syntactically then this is a decent solution for now. lmk what you think

src/ast/query.rs
```diff
@@ -276,6 +280,8 @@ impl fmt::Display for Table {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct Select {
+    /// SELECT
+    pub select_token: IgnoreField<TokenWithLocation>,
```
iffyio, line 284: For the comment we can probably be more descriptive to mention that this points to the position of the SELECT keyword? (similar for others probably if we want to make it more obvious which part of the statement is being referred to)

src/ast/helpers/ignore_field.rs
```diff
@@ -0,0 +1,69 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use core::cmp::{Eq, Ord, Ordering, PartialEq, PartialOrd};
+use core::fmt::{self, Debug, Formatter};
+use core::hash::{Hash, Hasher};
+
+/// A wrapper type that ignores the field when comparing or hashing.
+pub struct IgnoreField<T>(pub T);
```
iffyio, line 23: Instead of using a generic type would it make sense to represent via a regular wrapper type (thinking since the plan is only to wrap the tokens we attach to the nodes)? e.g.
```rust
pub struct AttachedToken(TokenWithLocation);

impl From<Token> for AttachedToken {}
impl From<TokenWithLocation> for AttachedToken {}
impl Hash for AttachedToken {}
// maybe even a default/noop impl to make it easier in scenarios like test cases where the info is irrelevant
// etc.

pub struct With {
    pub with_token: AttachedToken,
    pub recursive: bool,
    // ...
}
```

src/ast/query.rs
```diff
@@ -556,6 +563,8 @@ pub struct Cte {
     pub query: Box<Query>,
     pub from: Option<Ident>,
     pub materialized: Option<CteAsMaterialized>,
+    // needed for accurate span reporting
```
iffyio, line 566: can we update the comment to rather say what's being pointed to (thinking all attached tokens are implicitly needed for span reporting so that no need to treat this case specially)?

src/ast/mod.rs
```diff
@@ -140,8 +145,39 @@ pub struct Ident {
...
+    /// The span of the identifier in the original SQL string.
+    pub span: Span,
+}
+
+impl PartialEq for Ident {
+    fn eq(&self, other: &Self) -> bool {
+        let Ident {
+            value,
+            quote_style,
+            // backwards compat
```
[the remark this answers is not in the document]
Nyrox: yup 👍

src/ast/query.rs
Nyrox: yup 👍

src/ast/query.rs
Nyrox: yup 👍

src/ast/helpers/ignore_field.rs
Nyrox: doneso 👍

src/ast/mod.rs
```diff
@@ -902,10 +967,10 @@ pub enum Expr {
         /// `<search modifier>`
         opt_search_modifier: Option<SearchModifier>,
     },
-    Wildcard,
+    Wildcard(TokenWithLocation),
```
iffyio, line 970:
```suggestion
    Wildcard(AttachedToken),
```

src/ast/mod.rs
```diff
@@ -902,10 +967,10 @@ pub enum Expr {
...
     /// Qualified wildcard, e.g. `alias.*` or `schema.table.*`.
     /// (Same caveats apply to `QualifiedWildcard` as to `Wildcard`.)
-    QualifiedWildcard(ObjectName),
+    QualifiedWildcard(ObjectName, TokenWithLocation),
```
iffyio, line 973:
```suggestion
    QualifiedWildcard(ObjectName, AttachedToken),
```

src/ast/helpers/attached_token.rs
```diff
@@ -0,0 +1,68 @@
...
+use core::cmp::{Eq, Ord, Ordering, PartialEq, PartialOrd};
+use core::fmt::{self, Debug, Formatter};
+use core::hash::{Hash, Hasher};
+
+use crate::tokenizer::TokenWithLocation;
+
+/// A wrapper type for attaching tokens to AST nodes that should be ignored in comparisons and hashing.
+/// This should be used when a token is not relevant for semantics, but is still needed for
+/// accurate source location tracking.
+#[derive(Clone)]
```
iffyio, line 27: Ah I think we might  be missing the serde and visit derive on this struct given its attached to structs in the ast?

tests/sqlparser_postgres.rs
```diff
@@ -2507,6 +2520,7 @@ fn parse_array_subquery_expr() {
                     op: SetOperator::Union,
                     set_quantifier: SetQuantifier::None,
                     left: Box::new(SetExpr::Select(Box::new(Select {
+                        select_token: TokenWithLocation::wrap(Token::make_keyword("SELECT")).into(),
```
iffyio, line 2523: To make places like tests that don't care about the token leaner, could we add a helper function like `AttachedToken::empty()` that returns a dummy token?

src/tokenizer.rs
```diff
@@ -431,36 +433,143 @@ pub struct Location {
...
+impl PartialOrd<TokenWithLocation> for TokenWithLocation {
+    fn partial_cmp(&self, other: &TokenWithLocation) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for TokenWithLocation {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
+        let TokenWithLocation { token, span: _ } = self;
+        token.cmp(&other.token)
```
iffyio, line 572: Ah do we still need these custom impls now that we have the wrapper type or we can go back to the derives for TokenWithLocation now?

Nyrox: Yeah I think it's fine. Idk. if rust-analyzer can accurately do find references on trait implementations, but it did not find any, so I think it's only used upstream where we should have the wrapper now

tests/sqlparser_postgres.rs
Nyrox: I was sceptical about this, but it actually works pretty well 😅 
Not sure what Token to pick as a dummy token, so I went with `Token::EOF`

Nyrox:
Thanks for the great review. I was definitely a bit sloppy yesterday (Monday 🤮).
I merged main again if you want to give that a quick review as well. :)

tests/sqlparser_postgres.rs
iffyio: Yeah EOF seems like natural/reasonable thanks!

iffyio (approved):
LGTM! Thanks for tackling this @Nyrox!
cc @alamb

alamb:
I started looking over this PR -- and it is looking great. I took the liberty of merging up from main and fixing clippy/fmt.

I am going to test out the implications of this change by making a PR to DataFusion and test out the documentation.

alamb (approved):
Thank you so much Thank you so much @Nyrox and @iffyio  and @lovasoa  -- this is epic work.

Also, thank you to @lustefaniak @yuyang-ok for your comments

Many people have tried this feature but non have prevailed. 👏 . If we are ever colocated I totally owe you an in person meet up / 🍻   with a beverage of your choice

In terms of next steps:
1. I will file a ticket with the current state of the project / spans which can hopefully let us spread out the work for adding span information to the rest of parse tree over time. 
2. I think it would be good to consider changing the offsets in Location to be `u32` rather than `usize` which would reduce the memory requirements significantly I thin.
3.  I have a few ideas to improve the documentation, but I will propose some follow on PRs to do that. 

BTW I tried updating DataFusion to use this change here https://github.com/apache/datafusion/pull/13546 and it went quite smoothly

Let's leave this PR open for a few more days to get any more feedback and then plan to merge it in 🚀

alamb:
I plan to merge this PR tomorrow unless there are any other comments

Dandandan:
Did we run some benchmarks (e.g. cargo bench)?

alamb:
> Did we run some benchmarks (e.g. cargo bench)?

No, I did not (it is not clear to me we have such a thing). Let me look

alamb:
```shell
cd sqlparser_bench
cargo bench
git remote add Nyrox https://github.com/Nyrox/sqlparser-rs.git
git fetch Nyrox
git checkout Nyrox/main
```


Here is the benchmark result (it appears to be about 10%-15% slower according to the benchmark):

```
sqlparser-rs parsing benchmark/sqlparser::select
                        time:   [3.1255 µs 3.1265 µs 3.1276 µs]
                        change: [+15.235% +17.353% +19.853%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe
sqlparser-rs parsing benchmark/sqlparser::with_select
                        time:   [21.672 µs 21.683 µs 21.694 µs]
                        change: [+11.446% +11.549% +11.653%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
```

alamb:
> Thanks, looks great. I think 15% degradation is fully worth it (and might be gained back if someone looks at optimizing sqlparser-rs) :)

Yeah, I was looking at the flamegraph and there are a bunch of obvious things to improve performance (like changing `next_token` not to copy each token)

alamb:
I filed a ticket to discuss improving performance: 
- https://github.com/apache/datafusion-sqlparser-rs/issues/1557

Nyrox:
I also noticed that there is a bunch of calls to `<Location as Display>::fmt` which seems quite strange to me. Not sure what the test is doing, but I don't think the parser should be calling that on a non-error path? :p

alamb:
> I also noticed that there is a bunch of calls to <Location as Display>::fmt which seems quite strange to me. Not sure what the test is doing, but I don't think the parser should be calling that on a non-error path? :p

I agree -- it is strange that Parser::expected woud show up so much. One place I see it used is generating errors 🤔 

https://github.com/apache/datafusion-sqlparser-rs/blob/2e90e105a74bf9f50f2bad6c22992759ddb06880/src/parser/mod.rs#L3424-L3429

Screenshot 2024-11-26 at 9 51 11 AM

alamb:
Amusingly when I ran with the fixed benchmarks the result is basically the same (15% slower)

```
sqlparser-rs parsing benchmark/sqlparser::select
                        time:   [6.5723 µs 6.5856 µs 6.6010 µs]
                        change: [+14.669% +15.020% +15.364%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  11 (11.00%) high mild
  1 (1.00%) high severe
sqlparser-rs parsing benchmark/sqlparser::with_select
                        time:   [32.234 µs 32.253 µs 32.277 µs]
                        change: [+14.270% +14.402% +14.538%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe
```

[fixed-flamegraph](https://github.com/user-attachments/assets/7ceae3a0-dec2-462d-9a00-0b49afd9d2bc)

fixed-flamegraph

alamb:
I am more convinced than ever that we could make a huge performance improvement by doing something like:
-  https://github.com/apache/datafusion-sqlparser-rs/issues/1558

alamb:
🚀  -- thanks again @Nyrox @iffyio @lovasoa @lustefaniak @mkarbo @Dandandan  and @yuyang-ok  for helping push this along. It is pretty amazing we'll finally get new things

cc @ankrgyl  it's finally happening

ankrgyl:
Amazing!!!

alamb:
BTW if anyone has time to help review another PR, this one adds a bunch of documentation and examples for this feature:
- https://github.com/apache/datafusion-sqlparser-rs/pull/1549

alamb:
BTW here is a PR from @davisp that recovers all the performance lost adding tokens (and then some) ❤️ 
- https://github.com/apache/datafusion-sqlparser-rs/pull/1587

PR #1441, opened 2024-09-24
Snowflake: support of views column comment

7phs:
[Snowflake supports](https://docs.snowflake.com/en/sql-reference/sql/create-view) adding comments to fields within a view:

```sql
CREATE OR REPLACE VIEW v (a COMMENT 'Comment of field') AS SELECT a FROM table1
```

This MR introduces support for this feature.

Additionally, I noticed that the view's comment was not in the correct order—this issue surfaced while testing with a large query from a Snowflake production environment. The comment for the view should appear after the column list. I've included this minor fix in this MR as well. Please let me know if you think it would be better to address this in a separate MR.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3831,18 +3831,18 @@ impl fmt::Display for Statement {
                         .map(|to| format!(" TO {to}"))
                         .unwrap_or_default()
                 )?;
+                if !columns.is_empty() {
+                    write!(f, " ({})", display_comma_separated(columns))?;
+                }
+                if matches!(options, CreateTableOptions::With(_)) {
+                    write!(f, " {options}")?;
+                }
                 if let Some(comment) = comment {
                     write!(
                         f,
                         " COMMENT = '{}'",
                         value::escape_single_quote_string(comment)
                     )?;
-                }
-                if matches!(options, CreateTableOptions::With(_)) {
-                    write!(f, " {options}")?;
-                }
-                if !columns.is_empty() {
-                    write!(f, " ({})", display_comma_separated(columns))?;
                 }
                 if !cluster_by.is_empty() {
                     write!(f, " CLUSTER BY ({})", display_comma_separated(cluster_by))?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8193,16 +8193,10 @@ impl<'a> Parser<'a> {
     fn parse_view_column(&mut self) -> Result<ViewColumnDef, ParserError> {
         let name = self.parse_identifier(false)?;
         let mut options = vec![];
-        if dialect_of!(self is BigQueryDialect | GenericDialect)
-            && self.parse_keyword(Keyword::OPTIONS)
-        {
-            self.prev_token();
-            if let Some(option) = self.parse_optional_column_option()? {
-                options.push(option);
-            }
-        };
-        if dialect_of!(self is SnowflakeDialect | GenericDialect)
-            && self.parse_keyword(Keyword::COMMENT)
+        if (dialect_of!(self is BigQueryDialect | GenericDialect)
+            && self.parse_keyword(Keyword::OPTIONS))
+            || (dialect_of!(self is SnowflakeDialect | GenericDialect)
+                && self.parse_keyword(Keyword::COMMENT))
         {
             self.prev_token();
             if let Some(option) = self.parse_optional_column_option()? {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8197,26 +8197,25 @@ impl<'a> Parser<'a> {
             && self.parse_keyword(Keyword::OPTIONS)
         {
             self.prev_token();
-            let option = ColumnOption::Options(self.parse_options(Keyword::OPTIONS)?);
-            options.push(option);
+            if let Some(option) = self.parse_optional_column_option()? {
+                options.push(option);
+            }
         };
         if dialect_of!(self is SnowflakeDialect | GenericDialect)
             && self.parse_keyword(Keyword::COMMENT)
         {
-            let next_token = self.next_token();
-            let option = match next_token.token {
-                Token::SingleQuotedString(str) => ColumnOption::Comment(str),
-                _ => self.expected("string literal", next_token)?,
-            };
-            options.push(option);
-        };
-        let data_type = if dialect_of!(self is ClickHouseDialect) {
-            Some(self.parse_data_type()?)
-        } else {
-            None
+            self.prev_token();
+            if let Some(option) = self.parse_optional_column_option()? {
+                options.push(option);
+            }
         };
         let options = if !options.is_empty() {
             Some(options)
+        } else {
+            None
+        };
+        let data_type = if dialect_of!(self is ClickHouseDialect) {
+            Some(self.parse_data_type()?)
         } else {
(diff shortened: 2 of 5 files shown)


src/parser/mod.rs
```diff
@@ -8192,19 +8192,34 @@ impl<'a> Parser<'a> {
...
+        };
+        if dialect_of!(self is SnowflakeDialect | GenericDialect)
+            && self.parse_keyword(Keyword::COMMENT)
+        {
+            let next_token = self.next_token();
+            let option = match next_token.token {
+                Token::SingleQuotedString(str) => ColumnOption::Comment(str),
+                _ => self.expected("string literal", next_token)?,
+            };
+            options.push(option);
```
iffyio, line 8211: can this reuse the `parse_optional_column_option` function?

tests/sqlparser_snowflake.rs
```diff
@@ -2400,3 +2400,29 @@ fn parse_use() {
         );
     }
 }
+
+#[test]
+fn parse_view_column_descriptions() {
+    let sql = "CREATE OR REPLACE VIEW v (a COMMENT 'Comment of field') COMMENT = 'Comment of view' AS SELECT a FROM table1 AS o";
+
```
iffyio, line 2407: Can we also add a roundtrip test with multiple columns (where some have comments and other don't in the same statement), then another that only has column comment but not table-level comment?

src/ast/mod.rs
```diff
@@ -3831,19 +3831,19 @@ impl fmt::Display for Statement {
                         .map(|to| format!(" TO {to}"))
                         .unwrap_or_default()
                 )?;
+                if !columns.is_empty() {
```
iffyio, line 3834: could you explain what this change does? it looks like its swapping the order in which things show up but then I figure that should've led to some test failing, so wondering if I'm misunderstanding the effect of the change

7phs: `COMMENT` of view should be after a column list for example in

1.  [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-view) 
2. [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_view_statement)
3. [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-view.html)

I found that an order is invalid with `Display` while testing current changes with a query from production.

This fix makes order of displaying `COMMENT` options aligning sql specification.

tests/sqlparser_snowflake.rs
7phs: I updated the unit-test with multiple columns case and removed a view comment.

src/parser/mod.rs
7phs: Updated for `OPTIONS` and `COMMENT`.

src/ast/mod.rs
7phs: I added a dedicated unit-test for Snowflake to cover this case.

iffyio: I see, It doesn't seem like BigQuery supports COMMENT so should be unaffected. For this part, what's the intention behind the change?
```rust
if matches!(options, CreateTableOptions::With(_)) {
                    write!(f, " {options}")?;
}
```

src/parser/mod.rs
```diff
@@ -8192,11 +8192,25 @@ impl<'a> Parser<'a> {
...
+        if dialect_of!(self is BigQueryDialect | GenericDialect)
             && self.parse_keyword(Keyword::OPTIONS)
         {
             self.prev_token();
-            Some(self.parse_options(Keyword::OPTIONS)?)
+            if let Some(option) = self.parse_optional_column_option()? {
+                options.push(option);
+            }
+        };
+        if dialect_of!(self is SnowflakeDialect | GenericDialect)
```
iffyio, line 8204:
```suggestion
        } else if dialect_of!(self is SnowflakeDialect | GenericDialect)
```
nit: maybe an else if to show that both conditions do the same for different dialects

tests/sqlparser_snowflake.rs
```diff
@@ -2400,3 +2400,36 @@ fn parse_use() {
         );
     }
 }
+
+#[test]
+fn view_comment_option_should_be_after_column_list() {
```
iffyio, line 2405: Can we add a case like this mixed?
```sql
CREATE VIEW v (a COMMENT 'a comment', b, c COMMENT 'c comment`) COMMENT = 'Comment' AS SELECT a FROM t
```

src/ast/mod.rs
7phs: An order of entities if CREATE VIEW in `Snowflake` and `Databricks` cases DDL

`CREATE VIEW` -> `COLUMN LIST` -> [`(WITH) OPTIONS`] -> [`COMMENT`]

This change puts all available entities in order aligned DDL specifications. An order in main is reversed for column list, [with] options and comment.

src/parser/mod.rs
7phs: Updated with 
```
if (dialect_of!(self is BigQueryDialect | GenericDialect)
            && self.parse_keyword(Keyword::OPTIONS))
            || (dialect_of!(self is SnowflakeDialect | GenericDialect)
                && self.parse_keyword(Keyword::COMMENT))
        {
```
by `clippy` message.

tests/sqlparser_snowflake.rs
7phs: I added this case into a test.

src/parser/mod.rs
```diff
@@ -8192,11 +8192,19 @@ impl<'a> Parser<'a> {
...
-            Some(self.parse_options(Keyword::OPTIONS)?)
+            if let Some(option) = self.parse_optional_column_option()? {
+                options.push(option);
+            }
+        };
+        let options = if !options.is_empty() {
+            Some(options)
         } else {
             None
         };
```
iffyio, line 8210: would it make more sense to rewrite to something like this?
```rust
if (dialect_of!(....)) {
  self.prev_token();
  if let Some(option) = self.parse_optional_column_option()? {
    Some(vec![option])
  } else {None}
} else {
  None
};
```
thinking it helps with the following:
- we avoid the mut option
- the `is_empty` check as well, that one doesn't seem necessary, because the if condition already checks that we have an option

src/ast/mod.rs
iffyio: Ah, so the change of `CreateTableOptions::With(_)` I think is lacking test coverage in the MR, which made it hard to see why it was needed. Can we add tests for that as well?

src/parser/mod.rs

src/ast/mod.rs
iffyio: @7phs I think just one follow up needed on this one to add a test demonstrating the WITH behavior? i.e `CREATE OR REPLACE VIEW v (a COMMENT 'a comment', b, c COMMENT 'c comment') WITH (foo=bar) COMMENT = 'Comment' AS SELECT a FROM t`

7phs: @iffyio I added this case (as one more case) into the existing unit test.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @7phs and @iffyio

PR #1442, opened 2024-09-25
Add SQLite "ON CONFLICT" column option in CREATE TABLE statements

nucccc:
Thank you for all the work done on the parser, I hope this can be of any utility.

### Original issue

I was parsing `CREATE TABLE` statements when I discovered that SQLite allows for a non-standard `ON CONFLICT` option related to columns:

https://www.sqlite.org/lang_conflict.html

Considering an example from https://database.guide/how-on-conflict-works-in-sqlite/ the following sql statement could not be parsed:

```sql
CREATE TABLE Products( 
    ProductId INTEGER PRIMARY KEY, 
    ProductName NOT NULL ON CONFLICT IGNORE, 
    Price
);
```

resulting in the following error

```
ParserError("Expected: ',' or ')' after column definition ...
```

### Description of the syntax

Basically there is a column option that allows to specify how to handle conflicts in case of specific columns. The syntax seemed to me very clear in the plot at https://www.sqlite.org/syntax/conflict-clause.html, in which th keywords `ON CONFLICT` are followed by one of the keywords `[ROLLBACK, ABORT, FAIL, IGNORE, REPLACE]`

### My implementation

I noticed that the error originated in function `parse_columns` at

https://github.com/sqlparser-rs/sqlparser-rs/blob/affe8b549884a351ead4f35aa8bdf4cae8c93e4b/src/parser/mod.rs#L5874-L5879

when the parser did not recognize the option, still didn't find a comma.

So I added a column option parsing in function `parse_optional_column_option`:

https://github.com/nucccc/sqlparser-rs/blob/6154f66dc3bad88c87c40750ee882a23fb894a56/src/parser/mod.rs#L6086-L6103

in which if the dialect is SQLite (only SQL syntax I know that allows for this option) and the keywords `ON` and `CONFLICT` follow, it then checks if any of `[ROLLBACK, ABORT, FAIL, IGNORE, REPLACE]` is then present. In case none of them are found, a parser error is returned.

I added a dedicated column option `OnConflict(Keyword)`. I thought to make it accept a keyword, since the option is characterized by one of the 5 possible subsequent keywords.

I don't know if an `Expr` would have been a more appropriate choice instead of the `Keyword` enum. Most of the other column options accept an `Expr`, no one was accepting a `Keyword` from what I noticed.

I admit my limited knowledge of this codebase (maybe the `Keyword` concept does not need to exist at the ddl level). In case there is a better way to represent this column option I would kindly accept any suggestion.

I did my best in implementing an appropriate unit test.

### Caveats

From what I read on https://www.sqlite.org/lang_conflict.html "The ON CONFLICT clause applies to UNIQUE, NOT NULL, CHECK, and PRIMARY KEY constraints." In my implementation the presence of these constraints on the column is not taken into account, but it seemed to me (from the repo's readme) that this library concerns syntax, not semantics. So I thought maybe it's an upper level problem to verify that the `ON CONFLICT` clause is applied on a column with one of the above mentioned constraints.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6096,10 +6096,7 @@ impl<'a> Parser<'a> {
             ]);
             match on_confl {
                 Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
-                _ => self.expected(
-                    "one of ROLLBACK, ABORT, FAIL, IGNORE or REPLACE",
-                    self.peek_token(),
-                ),
+                _ => unreachable!()
             }
         } else {
             Ok(None)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6083,6 +6083,24 @@ impl<'a> Parser<'a> {
                 None
             };
             Ok(Some(ColumnOption::Identity(property)))
+        } else if dialect_of!(self is SQLiteDialect)
+            && self.parse_keywords(&[Keyword::ON, Keyword::CONFLICT])
+        {
+            // Support ON CONFLICT for SQLite
+            let on_confl = self.parse_one_of_keywords(&[
+                Keyword::ROLLBACK,
+                Keyword::ABORT,
+                Keyword::FAIL,
+                Keyword::IGNORE,
+                Keyword::REPLACE,
+            ]);
+            match on_confl {
+                Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
+                _ => self.expected(
+                    "one of ROLLBACK, ABORT, FAIL, IGNORE or REPLACE",
+                    self.peek_token(),
+                ),
+            }
         } else {
             Ok(None)
         }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6083,7 +6083,7 @@ impl<'a> Parser<'a> {
                 None
             };
             Ok(Some(ColumnOption::Identity(property)))
-        } else if dialect_of!(self is SQLiteDialect)
+        } else if dialect_of!(self is SQLiteDialect | GenericDialect)
             && self.parse_keywords(&[Keyword::ON, Keyword::CONFLICT])
         {
             // Support ON CONFLICT for SQLite
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6087,16 +6087,16 @@ impl<'a> Parser<'a> {
             && self.parse_keywords(&[Keyword::ON, Keyword::CONFLICT])
         {
             // Support ON CONFLICT for SQLite
-            let on_confl = self.parse_one_of_keywords(&[
+            let on_conflict = self.parse_one_of_keywords(&[
                 Keyword::ROLLBACK,
                 Keyword::ABORT,
                 Keyword::FAIL,
                 Keyword::IGNORE,
                 Keyword::REPLACE,
             ]);
-            match on_confl {
+            match on_conflict {
                 Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
-                _ => unreachable!()
+                _ => unreachable!(),
             }
         } else {
             Ok(None)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6087,17 +6087,15 @@ impl<'a> Parser<'a> {
             && self.parse_keywords(&[Keyword::ON, Keyword::CONFLICT])
         {
             // Support ON CONFLICT for SQLite
-            let on_conflict = self.parse_one_of_keywords(&[
-                Keyword::ROLLBACK,
-                Keyword::ABORT,
-                Keyword::FAIL,
-                Keyword::IGNORE,
-                Keyword::REPLACE,
-            ]);
-            match on_conflict {
-                Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
-                _ => unreachable!(),
-            }
+            Ok(Some(ColumnOption::OnConflict(
+                self.expect_one_of_keywords(&[
+                    Keyword::ROLLBACK,
+                    Keyword::ABORT,
+                    Keyword::FAIL,
+                    Keyword::IGNORE,
+                    Keyword::REPLACE,
+                ])?,
+            )))
         } else {
             Ok(None)
         }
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -6083,6 +6083,24 @@ impl<'a> Parser<'a> {
...
+        {
+            // Support ON CONFLICT for SQLite
+            let on_confl = self.parse_one_of_keywords(&[
+                Keyword::ROLLBACK,
+                Keyword::ABORT,
+                Keyword::FAIL,
+                Keyword::IGNORE,
+                Keyword::REPLACE,
+            ]);
+            match on_confl {
```
git-hulk, line 6097: Would it be better to rename `on_confl` to the full form `on_conflict`?

src/parser/mod.rs
```diff
@@ -6083,6 +6083,24 @@ impl<'a> Parser<'a> {
...
+                Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
+                _ => self.expected(
+                    "one of ROLLBACK, ABORT, FAIL, IGNORE or REPLACE",
+                    self.peek_token(),
+                ),
```
git-hulk, line 6102: It should be good to use unreadable here since `parse_one_of_keywords` has guarded the keyword.
```suggestion
                _ => unreachable!()
```

tests/sqlparser_sqlite.rs
```diff
@@ -281,6 +281,104 @@ fn parse_create_table_gencol() {
     sqlite_and_generic().verified_stmt("CREATE TABLE t1 (a INT, b INT AS (a * 2) STORED)");
 }
 
+#[test]
+fn parse_create_table_on_conflict_col() {
+    let sql_rollback = "CREATE TABLE t1 (a INT, b INT ON CONFLICT ROLLBACK)";
+    match sqlite().verified_stmt(sql_rollback) {
```
git-hulk, line 287: Will it be better to change it into a for-loop to simplify the test case.

```rust
 for keyword in ["ROLLBACK", "ABORT", "FAIL", "IGNORE", "REPLACE"] {
    let sql = "CREATE TABLE t1 (a INT, b INT ON CONFLICT {keyword})"
    ...
 }
```

nucccc: Completely agree, I modified the code accordingly

src/parser/mod.rs
nucccc: Completely agree, I modified the code accordingly

src/parser/mod.rs
nucccc: I saw that unreachable is often used through the rest of the parser, I modified the code accordingly.

git-hulk (approved):
Great! cc @iffyio

src/parser/mod.rs
```diff
@@ -6083,6 +6083,21 @@ impl<'a> Parser<'a> {
...
+            let on_conflict = self.parse_one_of_keywords(&[
+                Keyword::ROLLBACK,
+                Keyword::ABORT,
+                Keyword::FAIL,
+                Keyword::IGNORE,
+                Keyword::REPLACE,
+            ]);
+            match on_conflict {
+                Some(keyword) => Ok(Some(ColumnOption::OnConflict(keyword))),
+                _ => unreachable!(),
```
iffyio, line 6099: Ah so this behavior doesn't seem desirable, if we get wrong sql the parser shouldnt panic? I think rather the code would be roughly e.g.
```rust
Ok(Some(ColumnOption::OnConflict(self.expect_one_of_keywords(&[...])?)))
```

src/parser/mod.rs
```diff
@@ -6083,6 +6083,21 @@ impl<'a> Parser<'a> {
                 None
             };
             Ok(Some(ColumnOption::Identity(property)))
+        } else if dialect_of!(self is SQLiteDialect)
```
iffyio, line 6086:
```suggestion
        } else if dialect_of!(self is SQLiteDialect | GenericDialect)
```
we can extend the generic dialect if it doesn't conflict with existing behavior?

tests/sqlparser_sqlite.rs
```diff
@@ -281,6 +282,38 @@ fn parse_create_table_gencol() {
...
+fn parse_create_table_on_conflict_col() {
+    for keyword in [
+        Keyword::ROLLBACK,
+        Keyword::ABORT,
+        Keyword::FAIL,
+        Keyword::IGNORE,
+        Keyword::REPLACE,
+    ] {
+        let sql = format!("CREATE TABLE t1 (a INT, b INT ON CONFLICT {:?})", keyword);
+        match sqlite().verified_stmt(&sql) {
```
iffyio, line 295:
```suggestion
        match sqlite_and_generic().verified_stmt(&sql) {
```
just a heads up, in case the GenericDialect suggestion is applied, this bit changes as such

nucccc: I applied the commit, but without having the GenericDialect allowing the "ON CONFLICT" syntax this leads to an error in the tests, since parser righteously returns a ParserError needing of a comma:

```rust
ParserError("Expected: ',' or ')' after column definition, found: ON")
```
In case the GenericDialect was to allow such column option than a test case with
```
match sqlite_and_generic().verified_stmt(&sql) {
```
would pass, but at the moment it doesn't.

So this depends on extending the GenericDialect or not.

src/parser/mod.rs
nucccc: Extending the GenericDialect doesn't conflict with existing behaviour.

But for example this syntax doesn't work with a Postgres database (and I suspect many others).

But if you think it is proper to extend the GenericDialect (I have no opinion on this matter) I shall do it.

src/parser/mod.rs
nucccc: I fully agree with the suggestion, I adapted the code accordingly

nucccc:
I applied all the suggestions provided.

In doing this, I extended the GenericDialect as proposed to account for the `ON CONFLICT` column option.

I also modified the code docs to avoid the `cargo doc` error on the bare url.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @iffyio  and @nucccc

PR #1443, opened 2024-09-25
Support `DROP DATABASE`

### LINKED ISSUE #454: `Database` `ObjectType`
Now that `CREATE DATABASE` is supported, it would be great if `ObjectType::Database` were available for `DROP DATABASE`.

linhr:
This fixes #454.


tests/sqlparser_postgres.rs
```diff
@@ -893,6 +893,20 @@ fn parse_drop_schema_if_exists() {
     }
 }
 
+#[test]
+fn parse_drop_database_if_exists() {
```
iffyio, line 897: can this test case also live in common.rs similar to the other drop test case?

linhr: Done!

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @linhr  and @iffyio  🙏

alamb:
🚀

PR #1444, opened 2024-09-25
make `parse_expr_with_alias` public

### LINKED ISSUE #1439: parse_subexpr doesn't handle alias
It seems like [parse_subexpr](https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_subexpr) doesn't handle alias.

Here is some example:
```sql
1 + 3 * col0 as new_col0
```
It returns...
```
BinaryOp { left: Value(Number("1", false)), op: Plus, right: BinaryOp { left: Value(Number("3", false)), op: Multiply, right: Identifier(Ident { value: "col0", quote_style: None }) } }
```

I would like to work on this, but I have almost no experience writing parser, would probably need some help.

Eason0729:
Some very small change, as the title said.

close #1439

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10290,7 +10290,7 @@ impl<'a> Parser<'a> {
         Ok(ExprWithAlias { expr, alias })
     }
 
-    fn parse_expr_with_alias(&mut self) -> Result<ExprWithAlias, ParserError> {
+    pub fn parse_expr_with_alias(&mut self) -> Result<ExprWithAlias, ParserError> {
         let expr = self.parse_expr()?;
         let alias = if self.parse_keyword(Keyword::AS) {
             Some(self.parse_identifier(false)?)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10289,6 +10289,19 @@ impl<'a> Parser<'a> {
 
         Ok(ExprWithAlias { expr, alias })
     }
+    /// return new expression with alias
+    ///
+    /// Example
+    /// ```
+    /// # use sqlparser::parser::{Parser, ParserError};
+    /// # use sqlparser::dialect::GenericDialect;
+    /// # fn main() ->Result<(), ParserError> {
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -10290,7 +10290,7 @@ impl<'a> Parser<'a> {
         Ok(ExprWithAlias { expr, alias })
     }
 
-    fn parse_expr_with_alias(&mut self) -> Result<ExprWithAlias, ParserError> {
+    pub fn parse_expr_with_alias(&mut self) -> Result<ExprWithAlias, ParserError> {
```
alamb, line 10293: Could you please also add some documentation to this function (perhaps with an example or two) now that it will become part of the public API?

alamb (commented):
Thanks @Eason0729  -- makes sense to me

src/parser/mod.rs
Eason0729: Sorry for missing that piece of document, I have updated it.

src/parser/mod.rs
```diff
@@ -10289,8 +10289,21 @@ impl<'a> Parser<'a> {
 
         Ok(ExprWithAlias { expr, alias })
     }
+    /// return new expression with alias
```
iffyio, line 10292:
```suggestion
    /// Parses an expression with an optional alias

    /// Examples:

    /// ```sql
    /// SUM(price) AS total_price
    /// ```

    /// ```sql
    /// SUM(price)
    /// ```
```

Maybe something roughly like this? Mentioning that the alias is optional and then I'm thinking it might not be worth spelling out how to use the parser here (since the parser doc already covers it) given it would be more work to keep in sync

alamb (approved):
I took the liberty of applying @iffyio 's suggestion and I think the docs are now looking very nice

Thank you @Eason0729 and @iffyio 

Screenshot 2024-09-29 at 5 57 03 AM

PR #1446, opened 2024-09-28
Implements ALTER POLICY syntax for PostgreSQL

git-hulk:
```sql
ALTER POLICY name ON table_name RENAME TO new_name

ALTER POLICY name ON table_name
    [ TO { role_name | PUBLIC | CURRENT_ROLE | CURRENT_USER | SESSION_USER } [, ...] ]
    [ USING ( using_expression ) ]
    [ WITH CHECK ( check_expression ) ]
```

For the documentation, please refer:

https://www.postgresql.org/docs/current/sql-alterpolicy.html


git-hulk:
cc @iffyio

src/parser/alter.rs
```diff
@@ -36,6 +39,66 @@ impl<'a> Parser<'a> {
...
+        self.expect_keyword(Keyword::ON)?;
+        let table_name = self.parse_object_name(false)?;
+
+        let to = if self.parse_keyword(Keyword::TO) {
+            Some(self.parse_comma_separated(|p| p.parse_owner())?)
+        } else {
+            None
+        };
+
+        if self.parse_keyword(Keyword::RENAME) {
```
iffyio, line 64: from the doc it looks like the `RENAME` vs others are  are mutually exclusive, so that if we parse a `TO` it probably won't be correct to potentially also parse a rename afterwards as the current code suggests?
e.g `ALTER POLICY a on t TO owner RENAME TO new_name` seems like its accepted by the parser?

git-hulk: @iffyio Yes, you're right. I falsely parse TO before REANME, will fix it soon. Thank you for your correction.

git-hulk: @iffyio I have fixed this and also added test cases to prevent this happen.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @git-hulk -- there is a documentation tweak, but I'll make it and get this PR merged.

Thanks @iffyio  for the review

alamb:
I merged up from main to resolve a conflict

tests/sqlparser_common.rs
```diff
@@ -11091,42 +11091,82 @@ fn test_create_policy() {
 }
 
 #[test]
-fn test_drop_policy() {
```
alamb, line 11094: I messed up the merge here.  will fix

git-hulk:
@alamb Thank you!

alamb:
🚀

PR #1447, opened 2024-09-28
Expand handling of `LIMIT 1, 2` handling to include sqlite

joshuawarner32:
Sqlite itself parses this syntax, so this parser should parse it when operating in sqlite mode.

Tangential fixups:
* I also noticed that ClickhouseDialect was missing from `all_dialects()`, so I rectified that. This required excluding ClickhouseDialect from one existing test, but it works everywhere else.
* Added `TestedDialects::new(...)` function to more cleanly be able to specify exactly the dialects this LIMIT comma syntax works for, in tests
* Added a test for `LIMIT 1, 2`
* (very minor) fix capitalization of `is` in doc comment for `dialect_of!` macro

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -60,7 +60,7 @@ use alloc::boxed::Box;
 
 /// Convenience check if a [`Parser`] uses a certain dialect.
 ///
-/// `dialect_of!(parser Is SQLiteDialect |  GenericDialect)` evaluates
+/// `dialect_of!(parser is SQLiteDialect |  GenericDialect)` evaluates
 /// to `true` if `parser.dialect` is one of the [`Dialect`]s specified.
 macro_rules! dialect_of {
     ( $parsed_dialect: ident is $($dialect_type: ty)|+ ) => {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -316,6 +316,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Does the dialect support parsing `LIMIT 1, 2` as `LIMIT 2 OFFSET 1`?
+    fn supports_limit_comma(&self) -> bool {
+        false
+    }
+
     /// Does the dialect support trailing commas in the projection list?
     fn supports_projection_trailing_commas(&self) -> bool {
         self.supports_trailing_commas()
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8715,7 +8715,7 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if dialect_of!(self is GenericDialect | MySqlDialect | ClickHouseDialect)
+                if dialect_of!(self is GenericDialect | MySqlDialect | ClickHouseDialect | SQLiteDialect)
                     && limit.is_some()
                     && offset.is_none()
                     && self.consume_token(&Token::Comma)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8715,7 +8715,7 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if dialect_of!(self is GenericDialect | MySqlDialect | ClickHouseDialect | SQLiteDialect)
+                if self.dialect.supports_limit_comma()
                     && limit.is_some()
                     && offset.is_none()
                     && self.consume_token(&Token::Comma)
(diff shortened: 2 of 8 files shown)


src/parser/mod.rs
```diff
@@ -8715,7 +8715,7 @@ impl<'a> Parser<'a> {
                     offset = Some(self.parse_offset()?)
                 }
 
-                if dialect_of!(self is GenericDialect | MySqlDialect | ClickHouseDialect)
+                if dialect_of!(self is GenericDialect | MySqlDialect | ClickHouseDialect | SQLiteDialect)
```
alamb, line 8718: Something else we have been discussing is moving from this `dialect_of` macro to a method on the `Dialect trait` instead (the current code uses both patterns)

So it would look something like
```rust
if self.dialect.supports_limit_comma() ...
```

The benefits are that user defined Dialects can control the behavior and that there is a better place for documenting the behavior

Maybe something to think about for a future PR

src/test_utils.rs
```diff
@@ -211,22 +211,27 @@ impl TestedDialects {
...
-        Box::new(SnowflakeDialect {}) as Box<dyn Dialect>,
-        Box::new(HiveDialect {}) as Box<dyn Dialect>,
-        Box::new(RedshiftSqlDialect {}) as Box<dyn Dialect>,
-        Box::new(MySqlDialect {}) as Box<dyn Dialect>,
-        Box::new(BigQueryDialect {}) as Box<dyn Dialect>,
-        Box::new(SQLiteDialect {}) as Box<dyn Dialect>,
-        Box::new(DuckDbDialect {}) as Box<dyn Dialect>,
-        Box::new(DatabricksDialect {}) as Box<dyn Dialect>,
-    ];
+    specific_dialects(vec![
```
alamb, line 214: 👍

src/test_utils.rs
```diff
@@ -211,22 +211,27 @@ impl TestedDialects {
...
+        Box::new(BigQueryDialect {}),
+        Box::new(SQLiteDialect {}),
+        Box::new(DuckDbDialect {}),
+        Box::new(DatabricksDialect {}),
+        Box::new(ClickHouseDialect {}),
+    ])
+}
+
+/// Returns a specific set of dialects.
+pub fn specific_dialects(dialects: Vec<Box<dyn Dialect>>) -> TestedDialects {
```
alamb, line 232: This could also potentially be a constructor on TestedDialects

So like
```rust 
impl TestedDialects {
  pub fn new(dialects: Vec<Box<dyn Dialect>>) -> Self {
    Self { 
     dialects, 
     options: None,
    }
  }
}
```

Perhaps

tests/sqlparser_common.rs
```diff
@@ -6795,7 +6795,8 @@ fn parse_create_view_with_options() {
 #[test]
 fn parse_create_view_with_columns() {
     let sql = "CREATE VIEW v (has, cols) AS SELECT 1, 2";
-    match verified_stmt(sql) {
+    // TODO: why does this fail for ClickHouseDialect?
```
alamb, line 6798: Could you perhaps file a ticket to track this isse?

tests/sqlparser_common.rs
```diff
@@ -8587,17 +8588,26 @@ fn verified_expr(query: &str) -> Expr {
 
 #[test]
 fn parse_offset_and_limit() {
-    let sql = "SELECT foo FROM bar LIMIT 2 OFFSET 2";
+    let sql = "SELECT foo FROM bar LIMIT 1 OFFSET 2";
```
alamb, line 8591: 👍  that is a good change

src/dialect/mod.rs
```diff
@@ -60,7 +60,7 @@ use alloc::boxed::Box;
 
 /// Convenience check if a [`Parser`] uses a certain dialect.
 ///
-/// `dialect_of!(parser Is SQLiteDialect |  GenericDialect)` evaluates
+/// `dialect_of!(parser is SQLiteDialect |  GenericDialect)` evaluates
```
alamb, line 63: ❤️

alamb (approved):
Thank you @joshuawarner32  -- this looks like a very nice improvement. I left some comments but I don't think any is needed to merge this PR

FYI @iffyio

src/parser/mod.rs
joshuawarner32: Makes sense to me! Super easy, so will include this in the next commit to this PR.

tests/sqlparser_common.rs
joshuawarner32: Will do!

iffyio (approved):
LGTM!

alamb:
Thanks again @joshuawarner32 and @iffyio

PR #1448, opened 2024-09-29
Add a note discouraging new use of `dialect_of` macro

alamb:
@samuelcolvin  has brought up the idea of standardizing on one pattern for different Dialect parsing behavior

I don't have the ambition to try and change the existing code, but we can encourage new contributions to follow the preferred behavior with some comments. 

cc @iffyio @tobyhede and @lovasoa


jmhain (approved):

iffyio (approved):

lovasoa (approved):
good idea!

alamb:
Thanks for the review!

PR #1454, opened 2024-10-02
Snowflake: support for extended column options in `CREATE TABLE`

7phs:
Snowflake [supports](https://docs.snowflake.com/en/sql-reference/sql/create-table) several extended options for column definitions:
```
-- Column definition
    <col_name> <col_type>
      [ inlineConstraint ]
      [ NOT NULL ]
      [ COLLATE '<collation_specification>' ]
      [
        {
          DEFAULT <expr>
          | { AUTOINCREMENT | IDENTITY }
            [
              {
                ( <start_num> , <step_num> )
                | START <num> INCREMENT <num>
              }
            ]
            [ { ORDER | NOORDER } ]
        }
      ]
      [ [ WITH ] MASKING POLICY <policy_name> [ USING ( <col_name> , <cond_col1> , ... ) ] ]
      [ [ WITH ] PROJECTION POLICY <policy_name> ]
      [ [ WITH ] TAG ( <tag_name> = '<tag_value>' [ , <tag_name> = '<tag_value>' , ... ] ) ]
      [ COMMENT '<string_literal>' ]
```

This merge request implements support for these additional column options.


src/ast/ddl.rs
```diff
@@ -1098,17 +1098,116 @@ impl fmt::Display for ColumnOptionDef {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
```
iffyio, line 1101: For the introduced options, can we add a short description on what they are and a link to the snowflake docs that mention where they come from? It'll make it easier to navigate later on when folks land on the struct/enum without prior context

src/parser/mod.rs
```diff
@@ -6193,24 +6186,160 @@ impl<'a> Parser<'a> {
...
+            let tags = self.parse_comma_separated(Self::parse_tag)?;
+            self.expect_token(&Token::RParen)?;
+
+            Ok(Some(ColumnOption::Tags(tags)))
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn parse_snowflake_autoincrement_or_identity_option_column(
```
iffyio, line 6255: A similar comment on these ones, thinking it'll help with a comment showing the syntax/example of what's being parsed, ideally with a link to the doc?

src/ast/ddl.rs
```diff
@@ -1098,17 +1098,116 @@ impl fmt::Display for ColumnOptionDef {
...
     pub increment: Expr,
 }
 
-impl fmt::Display for IdentityProperty {
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum IdentityOrder {
+    Order,
+    Noorder,
```
iffyio, line 1138:
```suggestion
    NoOrder,
```

src/ast/ddl.rs
```diff
@@ -1291,11 +1406,13 @@ impl fmt::Display for ColumnOption {
...
-                    write!(f, "({parameters})")?;
-                }
-                Ok(())
+                write!(f, "{parameters}")
+            }
+            Policy(parameters) => {
+                write!(f, "{parameters}")
+            }
+            Tags(tags) => {
+                write!(f, "WITH TAG ({})", display_comma_separated(tags))
```
iffyio, line 1415: hmm I wonder if its correct to include the `WITH` here, maybe the `Tags` enum variant should rather be called `WithTags` in that case or something similar. But looking at the syntax, the WITH looks optional? If so we should include test coverage for both scenarios, then I'm thinking the variant can look like
```rust
struct TagsColumnOption {
    tags: Vec<Tag>,
    with: bool
}
enum  ColumnOption {
  ...
  Tags(TagsColumnOption)
}

File: src/ast/ddl.rs [hunk 5]
```diff
@@ -1182,13 +1281,29 @@ pub enum ColumnOption {
...
+    /// Creates an identity or an autoincrement column in a table.
     /// Syntax
     /// ```sql
-    /// IDENTITY [ (seed , increment) ]
+    /// { IDENTITY | AUTOINCREMENT } [ (seed , increment) | START num INCREMENT num ] [ ORDER | NOORDER ]
     /// ```
     /// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
-    Identity(Option<IdentityProperty>),
+    /// [Snowflake]: https://docs.snowflake.com/en/sql-reference/sql/create-table
+    Identity(Identity),
```
iffyio, line 1291: which variant does `Identity(None)` translate to in the new representation?

src/parser/mod.rs
```diff
@@ -6193,24 +6186,160 @@ impl<'a> Parser<'a> {
             && dialect_of!(self is MySqlDialect | SQLiteDialect | DuckDbDialect | GenericDialect)
         {
             self.parse_optional_column_option_as()
+        } else if self.parse_keyword(Keyword::AUTOINCREMENT)
+            && dialect_of!(self is SnowflakeDialect | SQLiteDialect |  GenericDialect)
+        {
+            if dialect_of!(self is SnowflakeDialect) {
```
iffyio, line 6192: One general comment, we recently started to discourage the use of `dialect_of` macro in the code base. See [here](https://github.com/apache/datafusion-sqlparser-rs/blob/1e0460a/src/dialect/mod.rs#L63-L68). Can we convert the usages here to use the dialect method as described in the linked doc where possible?

To keep the scope limited to the PR's concerns I'd suggest we only look to convert the new introductions, like the current line. In the places where we're adding snowflake as an extra variant to the dialect_of check like here
```
if dialect_of!(self is SnowflakeDialect | SQLiteDialect | GenericDialect)
```
we can keep as is, since trying to convert such usages would be out of scope for this PR I think

iffyio: Oh actually, another suggestion I'm thinking makes more sense. the current code has a lot of snowflake specific guards now which isn't ideal, I'm thinking now it's worth doing similar to what we have for when parsing statements like create table, where we shell out to the dialect to handle any custom syntax. That would limit the changes to this file, and have the new logic live within the snowflake dialect.
See the [dialect.parse_statement()](https://github.com/apache/datafusion-sqlparser-rs/blob/2f076ed84195e2d9798f549ef18d95f7ae8fb49c/src/parser/mod.rs#L465-L468) for what it looks like.
Essentially to add a similar `dialect.parse_column_option()` and call the function here beforehand is what I'm thinking?

src/parser/mod.rs
```diff
@@ -6193,24 +6186,160 @@ impl<'a> Parser<'a> {
...
+                    order: None,
+                },
+            ))))
+        } else if ((self.parse_keyword(Keyword::WITH)
+            && self
+                .parse_one_of_keywords(&[Keyword::MASKING, Keyword::PROJECTION])
+                .is_some())
+            || self
+                .parse_one_of_keywords(&[Keyword::MASKING, Keyword::PROJECTION])
+                .is_some())
```
iffyio, line 6234: I think for these similar to the tag options, we'll need to track whether the `WITH` keyword was specified? If so I think one way to simplify the logic would be e.g.
```rust
let with = self.parse_keyword(WITH);
let policy = if self.parse_one_of_keywords(masking) { Policy { with, self.parse_policy() } }
```

tests/sqlparser_snowflake.rs
```diff
@@ -525,6 +525,307 @@ fn test_snowflake_single_line_tokenize() {
...
+                        }]
+                    },
+                    ColumnDef {
+                        name: "d".into(),
+                        data_type: DataType::Int(None),
+                        collation: None,
+                        options: vec![ColumnOptionDef {
+                            name: None,
+                            option: ColumnOption::Identity(Identity::Identity(IdentityProperty {
+                                #[cfg(not(feature = "bigdecimal"))]
```
iffyio, line 616: Ah if the feature flags are for the value number there's a helper we can use e.g. `number("1")` to produce a number value

tests/sqlparser_snowflake.rs
```diff
@@ -525,6 +525,307 @@ fn test_snowflake_single_line_tokenize() {
...
+                    },
+                ]
+            );
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn test_snowflake_create_table_with_collated_column() {
```
iffyio, line 651: for the tests, in addition to the single option ones like these can we add coverage for using multiple options (and a mixture of WITH and without the keyword where applicable)?

src/ast/ddl.rs
7phs: It translates to `ColumnOption::Identity(Identity::Identity(IdentityProperty{parameters: None, order: None}))`

src/ast/ddl.rs
7phs: I added a reasonable documentation for structs with links to a Snowflake doc.

src/parser/mod.rs
7phs: I added a reasonable documentation for structs with links to a Snowflake doc and moved this function into `snowflake` dialect implementation.

src/ast/ddl.rs
7phs: Updated.

src/ast/ddl.rs
7phs: Current implementation of similar options of table has no this flag - it is a reason that I skipped it.

I added `with` flag to existing structs and updated tags columns option definition.

src/parser/mod.rs
7phs: I moved all snowflake specific code of column options parsing into snowflake dialect implementation.

src/parser/mod.rs
7phs: I used this suggestion related to `with` parsing in common column option parsing function and in snowflake specific one.

tests/sqlparser_snowflake.rs
7phs: Applied for tests of `snowflake` and `ms sql`

tests/sqlparser_snowflake.rs
7phs: I updated all existing tests with two variants of tested SQL - without and with `WITH`.
Additionally I added one more test with several column options in SQL to make sure that all these options are parsed together.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+/// Syntax
+/// ```sql
+/// [ WITH ] TAG ( <tag_name> = '<tag_value>' [ , <tag_name> = '<tag_value>' , ... ] )
+/// ```
+/// [Snowflake]: https://docs.snowflake.com/en/sql-reference/sql/create-table
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct TagsColumnOption {
+    pub with: bool,
```
iffyio, line 1254: Can we add a comment mentioning what the `with` parameter is for? since its a custom and public parameter

src/parser/mod.rs
```diff
@@ -6251,10 +6265,30 @@ impl<'a> Parser<'a> {
                     Keyword::REPLACE,
                 ])?,
             )))
+        } else if self.parse_keywords(&[Keyword::TAG])
```
iffyio, line 6268: To avoid prematurely consuming the with and then later potentially rewinding to prev token, we can probably do something like

```rust
else if (self.parse_keyword([TAG) || self.parse_keywords([WITH, TAG])) && dialect_of()
```
Also can we switch the dialect_of for a dialect method similar to the others? Related question: is the tags support here so that we can also cover generic dialect? (if so and we didn't have support for the generic dialect previously, then I'm thinking we can just keep it limited to snowflake and move the code over to the dialect)

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
     }
 }
 
+/// Identity is a column option for defining an identity or autoincrement column in a creating table statement.
```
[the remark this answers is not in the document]
7phs: Updated

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+/// { IDENTITY | AUTOINCREMENT } [ (seed , increment) | START num INCREMENT num ] [ ORDER | NOORDER ]
+/// ```
+/// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+/// [Snowflake]: https://docs.snowflake.com/en/sql-reference/sql/create-table
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum Identity {
+    Autoincrement(IdentityProperty),
+    Identity(IdentityProperty),
```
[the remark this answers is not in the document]
7phs: Added a doc for all options.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+/// Syntax
+/// ```sql
+/// { IDENTITY | AUTOINCREMENT } [ (seed , increment) | START num INCREMENT num ] [ ORDER | NOORDER ]
+/// ```
+/// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+/// [Snowflake]: https://docs.snowflake.com/en/sql-reference/sql/create-table
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum Identity {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
-impl fmt::Display for IdentityProperty {
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum IdentityOrder {
+    Order,
+    NoOrder,
+}
+
+impl fmt::Display for Identity {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+            write!(f, "{parameters}")?;
+        }
+        if let Some(order) = &property.order {
+            write!(f, "{order}")?;
+        }
+        Ok(())
+    }
+}
+
+impl fmt::Display for IdentityFormat {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+/// Syntax
+/// ```sql
+/// (seed , increment)
+/// ```
+/// [MS SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property
+/// [Snowflake]: https://docs.snowflake.com/en/sql-reference/sql/create-table
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum IdentityFormat {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+                    f,
+                    " START {} INCREMENT {}",
+                    parameters.seed, parameters.increment
+                )
+            }
+        }
+    }
+}
+
+impl fmt::Display for IdentityOrder {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
```diff
@@ -1096,17 +1096,172 @@ impl fmt::Display for ColumnOptionDef {
...
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ColumnPolicyProperty {
+    pub with: bool,
+    pub policy_name: Ident,
+    pub using_columns: Option<Vec<Ident>>,
+}
+
+impl fmt::Display for ColumnPolicy {
```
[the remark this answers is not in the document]
7phs: Updated.

src/ast/ddl.rs
7phs: Added a doc comment for this option and for the column policy option.

src/parser/mod.rs
7phs: A parsing of `TAG` column options moved to a snowflake dialect implementation.

src/dialect/snowflake.rs
```diff
@@ -149,6 +153,44 @@ impl Dialect for SnowflakeDialect {
...
+                parse_column_policy_property(parser, with)
+                    .map(|p| Some(ColumnOption::Policy(ColumnPolicy::ProjectionPolicy(p)))),
+            )
+        } else if parser.parse_keywords(&[Keyword::TAG]) {
+            Some(parse_column_tags(parser, with).map(|p| Some(ColumnOption::Tags(p))))
+        } else {
+            // needs to revert initial state of parser if dialect finds any matching
+            if with {
+                parser.prev_token();
+            }
```
iffyio, line 189: for this part can we use `parser.maybe_parse`? I think that would handle this automatically

src/parser/mod.rs
```diff
@@ -6255,6 +6266,15 @@ impl<'a> Parser<'a> {
             Ok(None)
         }
     }
+
+    pub fn parse_tag(&mut self) -> Result<Tag, ParserError> {
```
iffyio, line 6270:
```suggestion
    pub(crate) fn parse_tag(&mut self) -> Result<Tag, ParserError> {
```

src/dialect/snowflake.rs
7phs: Applied.

src/parser/mod.rs

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @7phs  and @iffyio

PR #1456, opened 2024-10-04
Add support for the LIKE ANY and ILIKE ANY pattern-matching condition

yoavcloud:
Snowflake supports pattern matching over a list of patterns with the use of the case-sensitive LIKE ANY and case-insensitive ILIKE ANY functions. Added support for the ANY option in the Expr::Like and Expr::ILike variants.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -613,6 +613,7 @@ pub enum Expr {
     /// `[NOT] LIKE <pattern> [ESCAPE <escape_character>]`
     Like {
         negated: bool,
+        any: bool,
         expr: Box<Expr>,
         pattern: Box<Expr>,
         escape_char: Option<String>,
@@ -620,6 +621,7 @@ pub enum Expr {
     /// `ILIKE` (case-insensitive `LIKE`)
     ILike {
         negated: bool,
+        any: bool,
         expr: Box<Expr>,
         pattern: Box<Expr>,
         escape_char: Option<String>,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -613,13 +613,15 @@ pub enum Expr {
...
         negated: bool,
+        any: bool,
         expr: Box<Expr>,
         pattern: Box<Expr>,
         escape_char: Option<String>,
     },
     /// `ILIKE` (case-insensitive `LIKE`)
     ILike {
         negated: bool,
+        any: bool,
```
iffyio, line 624: Could we add a link to the snowflake doc where the syntax is mentioned? It would help with context for someone that comes across it in the codebase

src/parser/mod.rs
```diff
@@ -2734,6 +2734,10 @@ impl<'a> Parser<'a> {
                     let regexp = self.parse_keyword(Keyword::REGEXP);
                     let rlike = self.parse_keyword(Keyword::RLIKE);
                     if regexp || rlike {
+                        if self.parse_keyword(Keyword::ANY) {
+                            self.prev_token();
+                            return self.expected("pattern after RLIKE or REGEXP", self.peek_token());
+                        }
```
iffyio, line 2740: Ah not sure I followed this part, we don't seem to take any action for when parsing ANY is successful?

yoavcloud: The only point of this condition was to add a meaningful error message when parsing incorrect syntax, for example: `RLIKE ANY`. See the last check of the unit test. Can be removed, I debated this one.

yoavcloud: Removed in the latest commit.

src/ast/mod.rs
yoavcloud: Addressed in the latest commit

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @yoavcloud and @iffyio

PR #1458, opened 2024-10-05
Add support of `EXPLAIN QUERY PLAN` syntax for SQLite dialect

### LINKED ISSUE #1455: `EXPLAIN QUERY PLAN` fails to parse (sqlite)
## problem
queries with [explain query plan](https://sqlite.org/eqp.html) fail when using the sqlite dialect

## example
query:
```sql
select * from "sensor_data"
```

error:
```sh
sql parser error: Expected: end of statement, found: plan at Line: 1, Column: 15
```

git-hulk:
This closes #1455 and introduces one keyword: `PLAN`.

For the syntax documentation, please refer to:

https://sqlite.org/lang_explain.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3316,18 +3316,17 @@ impl fmt::Display for Statement {
 
                 if *query_plan {
                     write!(f, "QUERY PLAN ")?;
-                } else {
-                    if *analyze {
-                        write!(f, "ANALYZE ")?;
-                    }
+                }
+                if *analyze {
+                    write!(f, "ANALYZE ")?;
+                }
 
-                    if *verbose {
-                        write!(f, "VERBOSE ")?;
-                    }
+                if *verbose {
+                    write!(f, "VERBOSE ")?;
+                }
 
-                    if let Some(format) = format {
-                        write!(f, "FORMAT {format} ")?;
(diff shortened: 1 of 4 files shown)


git-hulk:
cc @iffyio

src/ast/mod.rs
```diff
@@ -3302,22 +3307,27 @@ impl fmt::Display for Statement {
...
 
-                if *analyze {
-                    write!(f, "ANALYZE ")?;
-                }
+                if *query_plan {
+                    write!(f, "QUERY PLAN ")?;
+                } else {
+                    if *analyze {
+                        write!(f, "ANALYZE ")?;
+                    }
```
iffyio, line 3322: can we skip the else condition if it works just as well to do?
```rust
if query_plan {
   write!()
}
if analyze {
  write!()
}
```
thinking that would keep this part simpler, even though the parser only accepts one or the other for the same statement

git-hulk: @iffyio Yes, that makes sense. It's better to keep the fmt simpler.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thanks @git-hulk  and @iffyio

PR #1460, opened 2024-10-07
Add PostgreSQL specfic "CREATE TYPE t AS ENUM (...)" support.

caldwell:
See: https://www.postgresql.org/docs/current/sql-createtype.html

I implemented this as a separate `Statement::CreateTypeAsEnum { name, labels }` so it wouldn't be as invasive to the existing `Statement::CreateType` (which I would imagine helps forwards/backwards compatibility for the crate's users).

I used the `PostgreSqlDialect`'s comment as a blueprint but that means that the generic dialect doesn't get this feature (unless I'm missing something). I didn't see `parse_statement()` implemented in `GenericDialect` though I suppose it could be (and call out to other dialects' `parse_statement()`s). I didn't want to set that precedent without asking.

### DIFF
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -28,7 +28,7 @@
 // limitations under the License.
 use log::debug;
 
-use crate::ast::{CommentObject, Statement};
+use crate::ast::{CommentObject, ObjectName, Statement, UserDefinedTypeRepresentation};
 use crate::dialect::{Dialect, Precedence};
 use crate::keywords::Keyword;
 use crate::parser::{Parser, ParserError};
@@ -138,6 +138,9 @@ impl Dialect for PostgreSqlDialect {
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::COMMENT) {
             Some(parse_comment(parser))
+        } else if parser.parse_keyword(Keyword::CREATE) {
+            parser.prev_token(); // unconsume the CREATE in case we don't end up parsing anything
+            parse_create(parser)
         } else {
             None
         }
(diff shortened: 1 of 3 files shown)


alamb (approved):
Thank you for this contribution @caldwell 


> implemented in GenericDialect though I suppose it could be (and call out to other dialects' parse_statement()s). I didn't want to set that precedent without asking.

I agree -- I think unless there is a good reason to start migrating the code to a different pattern, following the existing patterns is probably the best

cc @iffyio

src/ast/mod.rs
```diff
@@ -3181,6 +3181,14 @@ pub enum Statement {
         representation: UserDefinedTypeRepresentation,
     },
     /// ```sql
+    /// CREATE TYPE <name> AS ENUM
+    /// ```
+    /// Note: this is a PostgreSQL-specific statement. See <https://www.postgresql.org/docs/current/sql-createtype.html>
+    CreateTypeAsEnum {
+        name: ObjectName,
+        labels: Vec<Ident>,
+    },
```
iffyio, line 3190: I'm thinking it could make sense to include it as part of the existing `CreateType`, especially since its already an enum there wouldn't be as much of disadvantage in terms of backwards compatibility? then going forward we'd be able to extend it in a similar manner with future extensions to the statement. something like:
```rust
pub enum UserDefinedTypeRepresentation {
    Composite {
        attributes: Vec<UserDefinedTypeCompositeAttributeDef>,
    },
    // new variant
    Enum(CreateTypeEnum)
}
```

src/dialect/postgresql.rs
```diff
@@ -138,6 +138,14 @@ impl Dialect for PostgreSqlDialect {
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::COMMENT) {
             Some(parse_comment(parser))
+        } else if parser.parse_keyword(Keyword::CREATE) {
+            match parse_create(parser) {
+                Some(result) => Some(result),
+                None => {
+                    parser.prev_token(); // unconsume the CREATE if we didn't end up parsing anything
+                    None
+                }
```
iffyio, line 147: it might be possible to simplify this if we call `prev_token()` before calling `parse_create`? thinking since the latter already returns an Option representing the same result

alamb:
@caldwell  have you had a chance to consider @iffyio 's suggestions above?

caldwell:
Apologies, not yet. I'll try to look at it today.

src/dialect/postgresql.rs
caldwell: Yes, that can make it much simpler. I also tried using peek to avoid doing "consume, unconsume, consume" but it looks like there's no nice `peek_keyword()` type function and so it ended up looking like:
```rust
if matches!(parser.peek_token().token, Token::Word(Word { keyword: Keyword::CREATE, .. })) {
    parse_create(parser)
}
```
That's a bit of a mouthful and I ended up just going with your suggestion:
```rust
if parser.parse_keyword(Keyword::CREATE) {
    parser.prev_token(); // unconsume the CREATE in case we don't end up parsing anything
    parse_create(parser)
}
```

src/ast/mod.rs
caldwell: Oh! I somehow missed that `UserDefinedTypeRepresentation` was an enum. I thought it was a struct which would be more invasive to change, hence the new `CreateTypeAsEnum`. But yeah, as an enum that's a much better place for it.

I ended up with:
```rust
pub enum UserDefinedTypeRepresentation {
    Composite {
        attributes: Vec<UserDefinedTypeCompositeAttributeDef>,
    },
    /// Note: this is PostgreSQL-specific. See <https://www.postgresql.org/docs/current/sql-createtype.html>
    Enum { labels: Vec<Ident> },
}
```
…so that `Enum` matches `Composite`. If you think it should be a real struct instead of the anonymous enum struct I can change it.

caldwell:
I amended my commit per @iffyio's suggestions. Thanks for taking the time to look at it and apologies again for taking too long to respond.

alamb:
> I amended my commit per @iffyio's suggestions. Thanks for taking the time to look at it and apologies again for taking too long to respond.

 no worries -- thank you@

caldwell:
I pushed a fix for the formatting test failure.

alamb (approved):
Looks good to me -- thanks @caldwell

iffyio (approved):
LGTM!

alamb:
🚀

PR #1461, opened 2024-10-07
Add "DROP TYPE" support.

caldwell:
See: https://www.postgresql.org/docs/current/sql-droptype.html
     https://learn.microsoft.com/en-us/sql/t-sql/statements/drop-type-transact-sql?view=sql-server-ver16
     https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/DROP-TYPE.html

According to the postgresql docs (above), their implementation is from ansi with only the `IF EXISTS` being an extension. mssql supports the `IF EXISTS` extension as well though it looks like Oracle doesn't.

I implemented this as a new `ObjectType::Type` which lets the existing `Statement::Drop` handle it generically.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -5713,6 +5713,7 @@ pub enum ObjectType {
     Role,
     Sequence,
     Stage,
+    Type,
 }
 
 impl fmt::Display for ObjectType {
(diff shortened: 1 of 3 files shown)


alamb (approved):
Makes sense to me . Thank you @caldwell 

cc @iffyio

iffyio (approved):
LGTM!

alamb:
🚀

PR #1462, opened 2024-10-07
Add support for ASC and DESC in CREATE TABLE column constraints for SQLite.

caldwell:
See: https://sqlite.org/lang_createtable.html

I didn't directly copy the `AUTOINCREMENT` code right above the diff because I happened to read #1430 the other day. Instead I created a new function in Dialect and enabled it in `SQLiteDialect` and `GenericDialect`—I hope this is ok.

I'm also a little disappointed that I couldn't deduplicate the ASC and DESC handling in src/parser/mod.rs. Maybe if/when [`if let` chains](https://github.com/rust-lang/rust/issues/53667) are stabilized they could be consolidated cleanly…

### DIFF
--- a/src/dialect/generic.rs
+++ b/src/dialect/generic.rs
@@ -103,4 +103,8 @@ impl Dialect for GenericDialect {
     fn supports_limit_comma(&self) -> bool {
         true
     }
+
+    fn supports_asc_desc_in_column_definition(&self) -> bool {
+        true
+    }
 }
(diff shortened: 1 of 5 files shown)


alamb (approved):
Looks good (again) to me @caldwell 

Thank you for breaking these PRs up into smaller ones for review

cc @iffyio

iffyio (approved):
LGTM!

PR #1463, opened 2024-10-09
chore: Add asf.yaml

Xuanwo:
This PR adds `asf.yaml` which mainly copied from https://github.com/apache/datafusion/blob/main/.asf.yaml

After this change, we can avoid sending too many traffic to dev@d.a.o.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,43 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,43 @@
...
+# This file controls the settings of this repository
+#
+# See more details at
+# https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features
+
+notifications:
+  commits: commits@datafusion.apache.org
+  issues: github@datafusion.apache.org
+  pullrequests: github@datafusion.apache.org
+  jira_options: link label worklog
```
tisonkun, line 27: I don't think this repo associated with a JIRA project so we may not need this line.

Xuanwo, line 27:
```suggestion
```

Xuanwo: Thanks, removed.

alamb (approved):
Thank you @Xuanwo and @tisonkun

.asf.yaml
```diff
@@ -0,0 +1,42 @@
...
+  enabled_merge_buttons:
+    squash: true
+    merge: false
+    rebase: false
+  features:
+    issues: true
+  protected_branches:
+    main:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
```
alamb, line 42: I would appreciate not putting branch protection on until we have enough committers who focus on sqlparser -- I am working on this

```suggestion
```

Xuanwo: Makes sense.

alamb:
Thank you @Xuanwo  and @tisonkun -- very much appreciated

PR #1464, opened 2024-10-09
fix for maybe_parse preventing parser from erroring on recursion limit

tomershaniii:
This PR fixes a bug in the recursion checks, the bug is caused by maybe_parse ignoring the returned error, which (in the case of recursion depth check) prevents the Error from propagating back to the caller and may lead to unexpected parsing results.

The fix involves specific handling of the `RecursionLimitExceeded` error within  `maybe_parse` and returning the error up the call stack.


iffyio (commented):
Thanks for this fix @tomershaniii! Would it be possible to include a test case that passes with this fix? If so we could add one [here](https://github.com/apache/datafusion-sqlparser-rs/blob/d39601c6fdf58b77482258d9b22b8c8672b0bb9c/tests/sqlparser_common.rs#L9140-L9167) for example

tomershaniii:
@iffyio took some yack shaving but it's finally here :-) check out `test_parentheses_overflow`
Some context for the other changes:

- the parse_boxed variants were doing the boxing when exiting the recursion, which resulted in unreasonable low number of stack frames before overflowing, in the PRs i removed the `parse_boxed` functions and added native boxing to the respective `parse_` functions

src/ast/mod.rs
```diff
@@ -3191,7 +3191,7 @@ pub enum Statement {
         /// Table confs
         options: Vec<SqlOption>,
         /// Cache table as a Query
-        query: Option<Query>,
+        query: Option<Box<Query>>,
```
alamb, line 3194: this is a nice change

src/ast/query.rs
```diff
@@ -1103,7 +1103,7 @@ pub enum PivotValueSource {
     /// Pivot on all values returned by a subquery.
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/pivot#pivot-on-column-values-using-a-subquery-with-dynamic-pivot>.
-    Subquery(Query),
+    Subquery(Box<Query>),
```
alamb, line 1106: 👍

src/parser/mod.rs
```diff
@@ -3510,16 +3510,19 @@ impl<'a> Parser<'a> {
...
         let index = self.index;
-        if let Ok(t) = f(self) {
-            Some(t)
-        } else {
-            self.index = index;
-            None
+        match f(self) {
+            Ok(t) => Ok(Some(t)),
+            // Unwind stack if limit exceeded
+            Err(ParserError::RecursionLimitExceeded) => Err(ParserError::RecursionLimitExceeded),
```
alamb, line 3521: 👍

alamb (approved):
Looks great to me -- than you @tomershaniii

alamb:
Looks like a CI check is failing on this PR

iffyio (approved):
LGTM! thanks @tomershaniii

tomershaniii:
CI issues fixed,  please re-run

tomershaniii:
Build broken by another commit, please re-run CI

alamb:
🤔 CI is still failing unfortunately

tomershaniii:
yet another rebase fix @alamb plz...

alamb:
🚀

alamb:
Thanks for sticking with this @tomershaniii  and for the review @iffyio

PR #1466, opened 2024-10-10
MySQL dialect: Add support for hash comments

hansott:
See https://dev.mysql.com/doc/refman/8.4/en/comments.html


iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Thank you @hansott and @iffyio

PR #1467, opened 2024-10-11
Add support for parsing MsSql alias with equals

yoavcloud:
This PR addresses the MsSql syntax for select item alias using the `=` operator, for example:
```
SELECT col_alias = col FROM tbl
```

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -562,7 +562,13 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// For example: SELECT col_alias = col FROM tbl
+    /// Returns true if this dialect supports treating the equals operator `=` within a [`SelectItem`]
+    /// as an alias assignment operator, rather than a boolean expression.
+    /// For example: the following statements are equivalent for such a dialect:
+    /// ```sql
+    ///  SELECT col_alias = col FROM tbl;
+    ///  SELECT col_alias AS col FROM tbl;
+    /// ```
     fn supports_eq_alias_assigment(&self) -> bool {
         false
     }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -562,7 +562,7 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if this dialect supports treating the equals operator `=` within a [`SelectItem`]
+    /// Returns true if this dialect supports treating the equals operator `=` within a `SelectItem`
     /// as an alias assignment operator, rather than a boolean expression.
     /// For example: the following statements are equivalent for such a dialect:
     /// ```sql
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -561,6 +561,11 @@ pub trait Dialect: Debug + Any {
     fn supports_asc_desc_in_column_definition(&self) -> bool {
         false
     }
+
+    /// For example: SELECT col_alias = col FROM tbl
+    fn supports_eq_alias_assigment(&self) -> bool {
+        false
+    }
 }
 
 /// This represents the operators for which precedence must be defined
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -11182,29 +11182,11 @@ impl<'a> Parser<'a> {
                 )
             }
             expr => {
-                // Parse a [`SelectItem`] based on an [MsSql] syntax that uses the equal sign
-                // to denote an alias, for example: SELECT col_alias = col FROM tbl
-                // [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations
-                let expr = if self.dialect.supports_eq_alias_assigment() {
-                    if let Expr::BinaryOp {
-                        ref left,
-                        op: BinaryOperator::Eq,
-                        ref right,
-                    } = expr
-                    {
-                        if let Expr::Identifier(alias) = left.as_ref() {
-                            return Ok(SelectItem::ExprWithAlias {
-                                expr: *right.clone(),
-                                alias: alias.clone(),
-                            });
-                        }
+                if self.dialect.supports_eq_alias_assigment() {
+                    if let Some(select_item) = Self::maybe_unpack_alias_assignment(&expr) {
+                        return Ok(select_item);
                     }
-                    expr
-                } else {
-                    expr
-                };
-
-                // Parse the common AS keyword for aliasing a column
+                }
                 self.parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)
                     .map(|alias| match alias {
                         Some(alias) => SelectItem::ExprWithAlias { expr, alias },
@@ -12219,6 +12201,28 @@ impl<'a> Parser<'a> {
             return true;
         }
         false
+    }
+
+    /// Parse a [`SelectItem`] based on an [MsSql] syntax that uses the equal sign
+    /// to denote an alias, for example: SELECT col_alias = col FROM tbl
+    /// [MsSql]: <https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations>
+    fn maybe_unpack_alias_assignment(expr: &Expr) -> Option<SelectItem> {
+        if let Expr::BinaryOp {
+            left,
+            op: BinaryOperator::Eq,
+            right,
+            ..
+        } = expr
+        {
+            if let Expr::Identifier(ref alias) = **left {
+                return Some(SelectItem::ExprWithAlias {
+                    expr: *right.clone(),
+                    alias: alias.clone(),
+                });
+            }
+        }
+
+        None
     }
 }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -11181,13 +11181,40 @@ impl<'a> Parser<'a> {
                     self.peek_token().location
                 )
             }
-            expr => self
-                .parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)
-                .map(|alias| match alias {
-                    Some(alias) => SelectItem::ExprWithAlias { expr, alias },
-                    None => SelectItem::UnnamedExpr(expr),
-                }),
+            expr => {
+                if dialect_of!(self is MsSqlDialect) {
+                    if let Some(select_item) = self.parse_mssql_alias_with_equal(&expr) {
+                        return Ok(select_item);
+                    }
+                }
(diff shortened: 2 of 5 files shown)


src/parser/mod.rs
```diff
@@ -11173,15 +11173,42 @@ impl<'a> Parser<'a> {
...
                 )
             }
-            expr => self
-                .parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)
-                .map(|alias| match alias {
-                    Some(alias) => SelectItem::ExprWithAlias { expr, alias },
-                    None => SelectItem::UnnamedExpr(expr),
-                }),
+            expr => {
+                if dialect_of!(self is MsSqlDialect) {
```
iffyio, line 11177: We recently started to discourage the use of [dialect_of](https://github.com/apache/datafusion-sqlparser-rs/blob/383226c3e81d9e5ed64f6f188815dfc9d7fea317/src/dialect/mod.rs#L63-L72) can we update this to use a method on the Dialect trait as described in the link?
We could add a method like `supports_eq_alias_assigment` or similar

src/parser/mod.rs
```diff
@@ -11173,15 +11173,42 @@ impl<'a> Parser<'a> {
...
+                        None => SelectItem::UnnamedExpr(expr),
+                    })
+            }
         }
     }
 
+    /// Parse a [`SelectItem`] based on an MsSql syntax that uses the equal sign
+    /// to denote an alias, for example: SELECT col_alias = col FROM tbl
+    /// <https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations>    
+    fn parse_mssql_alias_with_equal(&mut self, expr: &Expr) -> Option<SelectItem> {
```
iffyio, line 11194:
```suggestion
    fn maybe_unpack_alias_assignment(expr: &Expr) -> Option<SelectItem> {
```
Thinking since this is only a helper function that it doesn't get confused for a parser/parsing method

src/parser/mod.rs
```diff
@@ -11173,15 +11173,42 @@ impl<'a> Parser<'a> {
...
+                        Some(alias) => SelectItem::ExprWithAlias { expr, alias },
+                        None => SelectItem::UnnamedExpr(expr),
+                    })
+            }
         }
     }
 
+    /// Parse a [`SelectItem`] based on an MsSql syntax that uses the equal sign
+    /// to denote an alias, for example: SELECT col_alias = col FROM tbl
+    /// <https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations>
```
iffyio, line 11193:
```suggestion
    /// Parse a [`SelectItem`] based on an [MsSql] syntax that uses the equal sign
    /// to denote an alias, for example: SELECT col_alias = col FROM tbl
    /// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations    
```

src/parser/mod.rs
```diff
@@ -11173,15 +11173,42 @@ impl<'a> Parser<'a> {
...
+                if let Expr::Identifier(ref alias) = **left {
+                    return Some(SelectItem::ExprWithAlias {
+                        expr: *right.clone(),
+                        alias: alias.clone(),
+                    });
+                }
+            }
+        }
+
+        None
```
iffyio, line 11209:
```suggestion
        if let Expr::BinaryOp {
            left: Expr::Identifier(alias), op: BinaryOperator::Eq, right,
        } = expr {
            SelectItem::ExprWithAlias {
                expr: right,
                alias: Expr::Identifier(alias),
            }

        } else {
           expr
        }
```
Could be simplified to the following? I think no need to pass in a reference and return an option, we can move the value and return the same value if untouched?
Also we can skip the `..` operator when deconstructing in this case, so that if the field set changes, we would be notified to ensure the logic remains correct.

yoavcloud: But this would make the function return two different types, as SelectItem is not an Expr variant, no?

iffyio: Ah yeah that's correct, actually I'm thinking in that case we can inline the function since its small enough?

yoavcloud:
@iffyio I tried inlining the function, it's not as short as I wish it would, due to pattern matching the boxed fields. But maybe I'm missing something... Would love your feedback.

src/dialect/mod.rs
```diff
@@ -561,6 +561,11 @@ pub trait Dialect: Debug + Any {
     fn supports_asc_desc_in_column_definition(&self) -> bool {
         false
     }
+
+    /// For example: SELECT col_alias = col FROM tbl
```
iffyio, line 565:
```suggestion
    /// Returns true if this dialect supports treating the equals operator `=` within a [`SelectItem`]
    /// as an alias assignment operator, rather than a boolean expression.
    /// For example: the following statements are equivalent for such a dialect:
    /// ```sql
    ///  SELECT col_alias = col FROM tbl;
    ///  SELECT col_alias AS col FROM tbl;
    /// ```
```
Something like this to illustrate what the flag implies Im thinking?

src/parser/mod.rs
```diff
@@ -11173,12 +11173,36 @@ impl<'a> Parser<'a> {
...
-            expr => self
-                .parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)
-                .map(|alias| match alias {
-                    Some(alias) => SelectItem::ExprWithAlias { expr, alias },
-                    None => SelectItem::UnnamedExpr(expr),
-                }),
+            expr => {
+                // Parse a [`SelectItem`] based on an [MsSql] syntax that uses the equal sign
+                // to denote an alias, for example: SELECT col_alias = col FROM tbl
+                // [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/queries/select-examples-transact-sql?view=sql-server-ver16#b-use-select-with-column-headings-and-calculations
```
iffyio, line 11179:
```suggestion
```
We can add the mssql link as part of the trait impl for the dialect instead? Then here I think we would be able to skip the comment altogether since the `self.dialect.supports_eq_alias_assignment()` along with the documentation on the trait definition for the method would suffice. It'll let us keep the doc for the feature centralized

src/parser/mod.rs
```diff
@@ -11173,12 +11173,36 @@ impl<'a> Parser<'a> {
...
+                } else {
+                    expr
+                };
+
+                // Parse the common AS keyword for aliasing a column
+                self.parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)
+                    .map(|alias| match alias {
+                        Some(alias) => SelectItem::ExprWithAlias { expr, alias },
+                        None => SelectItem::UnnamedExpr(expr),
+                    })
```
iffyio, line 11204:
```suggestion
let (expr, alias) = match expr {
    Expr::BinaryOp {
        left: Expr::Identifier(alias),
        BinaryOperator::Eq,
        right,
    } if self.dialect.supports_eq_alias_assigment() => {
        (expr, Some(alias))
    }
    expr => {
        // Parse the common AS keyword for aliasing a column
        self.parse_optional_alias(keywords::RESERVED_FOR_COLUMN_ALIAS)?
    }
};
Ok(alias
    .map(|alias| SelectItem::ExprWithAlias{ expr, alias })
    .unwrap_or_else(|| SelectItem::UnnamedExpr(expr)))
```
The current inline looks reasonable, I think its mostly rustfmt that making it look a like it has bit more code than it does, but I think we could simplify the logic a bit as above to help with that and readability/intent as well? hopefully?

tests/sqlparser_mssql.rs
```diff
@@ -1024,6 +1024,17 @@ fn parse_create_table_with_identity_column() {
     }
 }
 
+#[test]
```
iffyio, line 1027: Ah sorry I forgot to mention this part earlier, the test looks reasonable but now with the method on the dialect, can we move the test to common.rs and write it in a [similar style to this](https://github.com/apache/datafusion-sqlparser-rs/blob/cb7eeed2aa7b4fdeb402e0df6e56a8b662633efb/tests/sqlparser_common.rs#L2781)? that would allow any future dialects with the same support to be covered automatically by the test.

Also for this feature, can we include an assertion that for the dialects that don't support the flag, verifying that they indeed end up with an unnamed select item?

src/parser/mod.rs
yoavcloud: I don't think you can match to `Expr::Identifier` in L11182 because `left` is a boxed field. This further complicates the logic of the function because you now need another `if let` condition to check the value of `left`. The complexity ends up similar to what I suggested.

I suggest going back to the helper function design, it cleanly encapsulates this rather edgy case of MsSql without complicating the main flow of parsing the common syntax for aliasing a column.

yoavcloud:
@iffyio Fixed all comments but went back to a helper function design as the elegance of the `parse_select_item` function is compromised by inlining this logic.

src/parser/mod.rs
iffyio: this works as its own match clause for example:
```rust
            Expr::BinaryOp {
                left,
                op: BinaryOperator::Eq,
                right,
            } if self.dialect.supports_eq_alias_assigment() && matches!(left.as_ref(), Expr::Identifier(_)) => {
                let Expr::Identifier(alias) = *left else {
                    return parser_err!(
                        "BUG: expected identifier expression as alias",
                        self.peek_token().location
                    );
                };
                Ok(SelectItem::ExprWithAlias {
                    expr: *right,
                    alias,
                })
            }
```
I think the least desirable option involves the expression cloning, i.e. the `right.clone()` I think we want to avoid in any case

yoavcloud:
@iffyio I adopted your change to inline the function in the latest commit

tests/sqlparser_common.rs
```diff
@@ -11432,3 +11432,15 @@ fn test_any_some_all_comparison() {
     verified_stmt("SELECT c1 FROM tbl WHERE c1 <> SOME(SELECT c2 FROM tbl)");
     verified_stmt("SELECT 1 = ANY(WITH x AS (SELECT 1) SELECT * FROM x)");
 }
+
+#[test]
+fn test_alias_equal_expr() {
+    let dialects = all_dialects_where(|d| d.supports_eq_alias_assigment());
```
iffyio, line 11438: @yoavcloud just one comment on the test [from here](https://github.com/apache/datafusion-sqlparser-rs/pull/1467#discussion_r1798900193) if we can include a similar case for the other dialects verifying that the `some_alias = some_column` gets parsed as an actual expression vs an alias?

Beyond that I think this should be good to go, thanks!

yoavcloud: Good idea, done!

PR #1470, opened 2024-10-11
feat: handle different placing of type names

emilsivervik:
Hey, 

First of all I just want to thank for all the awesome work on this library!

I've came across something that didn't work as expected a couple of months ago and created a fix for myself but now took the time to bring it here also. I'm sure there are some things left to do so I would appreciate some rubber-🦆 

The issue is when trying to parse create a create statement

```
// Working
CREATE TABLE tb (id INT, KEY USING HASH (id))
```
```
// Not working
CREATE TABLE tb (id INT, KEY (id) USING HASH)
```

There was also an issue using order on the keys, such as this
```
CREATE TABLE tb (id INT, KEY USING HASH (id ASC))
```
which I'm not entirely sure how you'd like to handle, from my side I just did what was the appropriate solution (not using it) but I'd imagine from your end that would not be ideal solution.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3476,12 +3476,13 @@ impl<'a> Parser<'a> {
         }
         Ok(values)
     }
-    /// Parse a comma-separated list of 1+ items accepted by `F`
+    /// Parse a comma-separated list of 1+ items accepted by `F` with possibility to skip [`Keyword`] and consume [`Keyword`].
+    /// Extended behavior to [`Self::parse_comma_separated`]
     pub fn parse_comma_separated_with_skip_and_consume<T, F>(
         &mut self,
-        mut f: F,
         skip: &[Keyword],
         consume: &[Keyword],
+        mut f: F,
     ) -> Result<Vec<T>, ParserError>
     where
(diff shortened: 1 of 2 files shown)


iffyio:
Hi @emilsivervik Sure! Could you take a look at the [CLUSTERED INDEX impl](https://github.com/apache/datafusion-sqlparser-rs/blob/6cfa42c3bd0a7f1d084af001b99845e05ea3b5da/src/parser/mod.rs#L6781-L6786) which has similar syntax and likely solution for parsing and representing the `(id ASC)` type scenarios?
I'm thinking we'd probably be able to reuse that type, for example if we extend the `IndexType` parsing and representation to include:

```rust
pub enum IndexType {
    BTree(Option<Vec<ClusteredIndex>>),
    Hash(Option<Vec<ClusteredIndex>>),
}
```
And then something similar for the KEY support. Let me know if you have any questions.

Also, from the tests I'm assuming this is a mysql syntax? could you include a link to the docs maybe in the PR description where this syntax is described?

iffyio:
Marking this as draft in the meantime as we're no longer waiting on a review.
@emilsivervik please feel free to undraft and ping when you want us to take another look!

Issue #1471, opened 2024-10-12
hive doesn't support `!a` expression


wugeer:
Hive supports `!` as a synonym for `not`, but the current project's Hive dialect does not support the `select !a` syntax.

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27362046#LanguageManualUDF-LogicalOperators

![image](https://github.com/user-attachments/assets/61b39751-d0ee-4a9c-96a8-3d9f9d1a4d7c)

The reproduction steps are as follows:

```bash
echo "select not (false), !a" > ok.sql
cargo run --features json_example --example cli ok.sql  --hive
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/cli ok.sql --hive`
Parsing from file 'ok.sql' using HiveDialect
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] Parsing sql 'select not (false), !a
'...
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] parsing expr
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] parsing expr
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] parsing expr
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] prefix: Value(Boolean(false))
2024-10-12T07:25:23.412Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithLocation { token: RParen, location: Location { line: 1, column: 18 } }
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] next precedence: 0
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] prefix: Nested(Value(Boolean(false)))
2024-10-12T07:25:23.412Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithLocation { token: Comma, location: Location { line: 1, column: 19 } }
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] next precedence: 0
2024-10-12T07:25:23.412Z DEBUG [sqlparser::parser] prefix: UnaryOp { op: Not, expr: Nested(Value(Boolean(false))) }
2024-10-12T07:25:23.413Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithLocation { token: Comma, location: Location { line: 1, column: 19 } }
[...]

PR #1472, opened 2024-10-12
hive: support for special not expression `!a` and raise error for `a!` factorial operator

wugeer:
Closes: https://github.com/apache/datafusion-sqlparser-rs/issues/1471
Closes: https://github.com/apache/datafusion-sqlparser-rs/issues/1473

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -51,6 +51,8 @@ pub enum UnaryOperator {
     PGPrefixFactorial,
     /// Absolute value, e.g. `@ -9` (PostgreSQL-specific)
     PGAbs,
+    /// Unary logical not operator: e.g. `! false` (Hive-specific)
+    BangNot,
 }
 
 impl fmt::Display for UnaryOperator {
@@ -65,6 +67,7 @@ impl fmt::Display for UnaryOperator {
             UnaryOperator::PGPostfixFactorial => "!",
             UnaryOperator::PGPrefixFactorial => "!!",
             UnaryOperator::PGAbs => "@",
+            UnaryOperator::BangNot => "!",
         })
     }
 }
--- a/src/dialect/hive.rs
+++ b/src/dialect/hive.rs
@@ -51,4 +51,8 @@ impl Dialect for HiveDialect {
     fn require_interval_qualifier(&self) -> bool {
         true
     }
+
+    fn supports_bang_not_operator(&self) -> bool {
+        true
+    }
 }
--- a/src/dialect/hive.rs
+++ b/src/dialect/hive.rs
@@ -52,6 +52,7 @@ impl Dialect for HiveDialect {
         true
     }
 
+    /// See Hive <https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27362061#Tutorial-BuiltInOperators>
     fn supports_bang_not_operator(&self) -> bool {
         true
     }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -596,7 +596,7 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if the dialect supports `!a` expressions
+    /// Returns true if the dialect supports `!a` syntax for boolean `NOT` expressions.
     fn supports_bang_not_operator(&self) -> bool {
         false
     }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -575,6 +575,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports `a!` expressions
+    fn supports_factorial_operator(&self) -> bool {
+        false
+    }
+
     /// Returns true if this dialect supports treating the equals operator `=` within a `SelectItem`
     /// as an alias assignment operator, rather than a boolean expression.
     /// For example: the following statements are equivalent for such a dialect:
@@ -588,6 +593,11 @@ pub trait Dialect: Debug + Any {
 
     /// Returns true if this dialect supports the `TRY_CONVERT` function
     fn supports_try_convert(&self) -> bool {
+        false
+    }
+
+    /// Returns true if the dialect supports `!a` expressions
+    fn supports_bang_not_operator(&self) -> bool {
         false
     }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2054,13 +2054,6 @@ impl<'a> Parser<'a> {
         }
     }
 
-    pub fn parse_bang_not(&mut self) -> Result<Expr, ParserError> {
-        Ok(Expr::UnaryOp {
-            op: UnaryOperator::BangNot,
-            expr: Box::new(self.parse_subexpr(self.dialect.prec_value(Precedence::UnaryNot))?),
-        })
-    }
-
     /// Parses fulltext expressions [`sqlparser::ast::Expr::MatchAgainst`]
     ///
     /// # Errors
@@ -2832,40 +2825,32 @@ impl<'a> Parser<'a> {
                 data_type: self.parse_data_type()?,
                 format: None,
             })
-        } else if Token::ExclamationMark == tok {
-            if self.dialect.supports_factorial_operator() {
-                match expr {
-                    Expr::Value(_) | Expr::Identifier(_) | Expr::Nested(_) | Expr::BinaryOp{..} => Ok(Expr::UnaryOp {
-                        op: UnaryOperator::PGPostfixFactorial,
-                        expr: Box::new(expr),
-                    }),
-                    _ => {
-                        self.expected(
-                            "Value or Identifier or  Nested or BinaryOp struct before factorial operator(!)", self.peek_token())
-                    },
-                }
-            } else if self.dialect.supports_bang_not_operator() {
-                let token = self.next_token();
-                match token.token {
-                    Token::Word(_) | Token::Number(..) => Ok(Expr::UnaryOp {
+        } else if Token::ExclamationMark == tok && self.dialect.supports_factorial_operator() {
+            Ok(Expr::UnaryOp {
+                op: UnaryOperator::PGPostfixFactorial,
+                expr: Box::new(expr),
+            })
+        } else if Token::ExclamationMark == tok && self.dialect.supports_bang_not_operator() {
+            let token = self.next_token();
+            match token.token {
+                Token::Word(_) | Token::Number(..)
+                    if !matches!(
+                        expr,
+                        Expr::Value(_)
+                            | Expr::Identifier(_)
+                            | Expr::Nested(_)
+                            | Expr::BinaryOp { .. }
+                    ) =>
+                {
+                    Ok(Expr::UnaryOp {
                         op: UnaryOperator::BangNot,
                         expr: Box::new(expr),
-                    }),
-                    _ => {
-                        parser_err!(
-                            "current dialect support bang not operator, but with wrong synx",
-                            tok.location
-                        )
-                    }
+                    })
                 }
-            } else {
-                parser_err!(
-                    format!(
-                        "current dialect: {:?} does not support factorial operator or bang not operator",
-                        self.dialect
-                    ),
-                    self.peek_token().location
-                )
+                _ => parser_err!(
+                    "current dialect support bang not operator, but with wrong syntax",
+                    tok.location
+                ),
             }
         } else if Token::LBracket == tok {
             if dialect_of!(self is PostgreSqlDialect | DuckDbDialect | GenericDialect) {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2830,28 +2830,6 @@ impl<'a> Parser<'a> {
                 op: UnaryOperator::PGPostfixFactorial,
                 expr: Box::new(expr),
             })
-        } else if Token::ExclamationMark == tok && self.dialect.supports_bang_not_operator() {
-            let token = self.next_token();
-            match token.token {
-                Token::Word(_) | Token::Number(..)
-                    if !matches!(
-                        expr,
-                        Expr::Value(_)
-                            | Expr::Identifier(_)
-                            | Expr::Nested(_)
-                            | Expr::BinaryOp { .. }
-                    ) =>
-                {
-                    Ok(Expr::UnaryOp {
-                        op: UnaryOperator::BangNot,
-                        expr: Box::new(expr),
-                    })
-                }
-                _ => parser_err!(
-                    "current dialect support bang not operator, but with wrong syntax",
-                    tok.location
-                ),
-            }
         } else if Token::LBracket == tok {
             if dialect_of!(self is PostgreSqlDialect | DuckDbDialect | GenericDialect) {
                 self.parse_subscript(expr)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1194,6 +1194,14 @@ impl<'a> Parser<'a> {
                     ),
                 })
             }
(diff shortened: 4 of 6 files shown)


src/parser/mod.rs
```diff
@@ -2796,9 +2811,27 @@ impl<'a> Parser<'a> {
...
+                        );
+                    }
+                }
+                _ => {}
+            };
+            let op = if !self.dialect.supports_factorial_operator() {
+                UnaryOperator::SpecialNot
+            } else {
+                UnaryOperator::PGPostfixFactorial
+            };
```
iffyio, line 2832: I think we can simplify this match block a couple ways:

- We can remove the `match expr` block above that returns the `parser_err`
- We can add a `self.dialect.supports_bang_not_operator()` this would be true for hive

then for this logic we do
```rust
let op = if self.dialect.supports_factorial_operator() {
    UnaryOperator::PGPostfixFactorial
} else if self.dialect.supports_bang_not_operator() {
    UnaryOperator::SpecialNot
} else {
    self.expected("...)
}
```

src/ast/operator.rs
```diff
@@ -51,6 +51,8 @@ pub enum UnaryOperator {
     PGPrefixFactorial,
     /// Absolute value, e.g. `@ -9` (PostgreSQL-specific)
     PGAbs,
+    /// Special Not, e.g. `! false` (Hive-specific)
+    SpecialNot,
```
iffyio, line 55:
```suggestion
    /// Unary logical not operator: e.g. `! false` (Hive-specific)
    BangNot,
```
Figured we can say Bang instead of Special since its conventional to call bang operator

src/parser/mod.rs
```diff
@@ -1174,6 +1175,14 @@ impl<'a> Parser<'a> {
                     ),
                 })
             }
+            Token::ExclamationMark if !self.dialect.supports_factorial_operator() => {
```
iffyio, line 1178:
```suggestion
            Token::ExclamationMark if self.dialect.supports_bang_not_operator() => {
```
See comment a bit lower for rationale to have a dedicated api for this behavior

tests/sqlparser_hive.rs
```diff
@@ -537,6 +537,33 @@ fn parse_use() {
     );
 }
 
+#[test]
+fn parse_hive_unary_ops() {
+    let hive_unary_ops = &[("!", UnaryOperator::SpecialNot)];
+
+    for (str_op, op) in hive_unary_ops {
```
iffyio, line 544: the single element for loop seems superflous, we can write the SQL query directly?

tests/sqlparser_hive.rs
```diff
@@ -537,6 +537,33 @@ fn parse_use() {
     );
 }
 
+#[test]
+fn parse_hive_unary_ops() {
```
iffyio, line 541: one general comment regarding the tests, since we're relying on the dialect api to guard the functionality, can we move the tests to common.rs and write them in this style?
https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_common.rs#L2781

Also we should do the same for the factorial tests since the current PR seems to be introducing two APIs to the dialect trait

src/dialect/mod.rs
```diff
@@ -561,6 +561,11 @@ pub trait Dialect: Debug + Any {
     fn supports_asc_desc_in_column_definition(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports `a!` expressions
+    fn supports_factorial_operator(&self) -> bool {
+        false
+    }
```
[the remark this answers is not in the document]
iffyio: Actually I think we can ignore this comment and leave as is, the other dialects will currently parse factorial but its not necessarily desirable (as far as i can tell, among the supported dialects pg is the only one and the syntax is deprecated)

tests/sqlparser_hive.rs
wugeer: Good idea! I will try it. :)

tests/sqlparser_hive.rs
wugeer: Yes, I will change it. :)

src/parser/mod.rs
wugeer: Good idea, I will try it. :)

wugeer:
@iffyio Thank you for your  suggestions—I learned a lot from them， I've resubmitted an updated version of the code. When you have a moment, could you please review it? Thanks!:)

src/parser/mod.rs
```diff
@@ -1074,6 +1074,7 @@ impl<'a> Parser<'a> {
                     }))
                 }
                 Keyword::NOT => self.parse_not(),
+
```
iffyio, line 1077:
```suggestion
```

src/ast/operator.rs
iffyio: This comment is unadressed? (in that the PR still uses 'special not')

src/parser/mod.rs
iffyio: Similarly, this comment looks unadressed? The new feature ideally has its own `dialect` method. Otherwise it can get complicated to try to infer when a feature should take precedence

src/parser/mod.rs
iffyio: I think this isn't fully implemented yet? we still have the match expr which we'd want to avoid

tests/sqlparser_hive.rs
iffyio: The test comments are also unadressed?

src/ast/operator.rs
wugeer: @iffyio Sorry, my oversight caused the code not to be merged.

src/parser/mod.rs
wugeer: The code has been re-pushed.

wugeer:
@iffyio Sorry, my oversight caused the code not to be merged. The code has been re-pushed.

src/dialect/mod.rs
```diff
@@ -590,6 +595,11 @@ pub trait Dialect: Debug + Any {
     fn supports_try_convert(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports `!a` expressions
```
iffyio, line 599:
```suggestion
    /// Returns true if the dialect supports `!a` syntax for boolean `NOT` expressions.
```

src/dialect/hive.rs
```diff
@@ -51,4 +51,8 @@ impl Dialect for HiveDialect {
     fn require_interval_qualifier(&self) -> bool {
         true
     }
+
+    fn supports_bang_not_operator(&self) -> bool {
```
iffyio, line 55: Do we have a link to the syntax? If so it would be good to include that here as part of the documentation

src/parser/mod.rs
```diff
@@ -2800,11 +2814,40 @@ impl<'a> Parser<'a> {
...
-            })
+            if self.dialect.supports_factorial_operator() {
+                match expr {
+                    Expr::Value(_) | Expr::Identifier(_) | Expr::Nested(_) | Expr::BinaryOp{..} => Ok(Expr::UnaryOp {
+                        op: UnaryOperator::PGPostfixFactorial,
+                        expr: Box::new(expr),
+                    }),
+                    _ => {
+                        self.expected(
+                            "Value or Identifier or  Nested or BinaryOp struct before factorial operator(!)", self.peek_token())
```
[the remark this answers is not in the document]
wugeer: Because both `Hive` and `PostgreSQL` dialects support the exclamation mark, we check the token before the exclamation mark to determine if it's a factorial operator, while also avoiding the incorrect identification of Hive `BangNot` syntax as PostgreSQL  `PostfixFactorial`.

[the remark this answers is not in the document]
wugeer: Yes, you're right. After testing, returning `Ok(Expr::UnaryOp ...)` directly here doesn't affect the result.

src/parser/mod.rs
```diff
@@ -2800,11 +2814,40 @@ impl<'a> Parser<'a> {
...
+                    },
+                }
+            } else if self.dialect.supports_bang_not_operator() {
```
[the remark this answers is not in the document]
wugeer: Because both `Hive` and `PostgreSQL` dialects support the exclamation mark, we check the token after the exclamation mark to determine if it's a `BangNot` operator, while also avoiding the incorrect identification of PostgreSQL `PostfixFactorial` as Hive `BangNot` operator.

tests/sqlparser_common.rs
```diff
@@ -11460,3 +11460,49 @@ fn test_try_convert() {
...
+    .into_iter()
+    .enumerate()
+    {
+        assert_eq!(
+            SelectItem::UnnamedExpr(Expr::UnaryOp { op: op, expr }),
+            projection[i]
+        )
+    }
+
+    let sql = "SELECT a!";
```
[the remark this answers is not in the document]
wugeer: Good catch! I will fix this.

src/dialect/hive.rs
wugeer: Fixed!

wugeer (commented):
Have Done! @iffyio

src/parser/mod.rs
```diff
@@ -2800,11 +2814,40 @@ impl<'a> Parser<'a> {
                 format: None,
             })
         } else if Token::ExclamationMark == tok {
-            // PostgreSQL factorial operation
-            Ok(Expr::UnaryOp {
-                op: UnaryOperator::PGPostfixFactorial,
-                expr: Box::new(expr),
-            })
+            if self.dialect.supports_factorial_operator() {
```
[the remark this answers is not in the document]
wugeer: Fixed!

tests/sqlparser_common.rs
```diff
@@ -11460,3 +11460,54 @@ fn test_try_convert() {
...
     dialects.verified_expr("TRY_CONVERT('foo', VARCHAR(MAX))");
 }
+
+#[test]
+fn parse_bang_not() {
+    let dialects = all_dialects_where(|d| d.supports_bang_not_operator());
+    let sql = "SELECT !a, !(b > 3)";
+    let Select { projection, .. } = dialects.verified_only_select(sql);
+
+    for (i, (op, expr)) in [
```
iffyio, line 11470: the `op` variable seems redundant given its always set to `UnaryOperator::BangNot`? so that the loop can probably be simplified to `for (i, expr) in   [...]`

src/parser/mod.rs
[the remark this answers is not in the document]
iffyio: Ah so both dialects support that syntax but not to mean the same thing (hence the dialect flags), the hive syntax is only a prefix operator so that I expect we wouldn't need any code at this stage (`parse_infix()`) for it (if we do then we can demonstrate that with a test case that fails to parse a valid expression without the block, but I didn't think there should be such a case)?
So that the block `} else if Token::ExclamationMark == tok && self.dialect.supports_bang_not_operator() {` doesn't seem to have any effect and ideally should be removed if so.

tests/sqlparser_common.rs
```diff
@@ -11460,3 +11460,54 @@ fn test_try_convert() {
         all_dialects_where(|d| d.supports_try_convert() && !d.convert_type_before_value());
     dialects.verified_expr("TRY_CONVERT('foo', VARCHAR(MAX))");
 }
+
+#[test]
+fn parse_bang_not() {
```
iffyio, line 11465: since this MR also introduces the `supports_factorial_operator` dialect method, can we add a similar test case for that one as well?

tests/sqlparser_common.rs
wugeer: Good idea. Have Done!

src/parser/mod.rs
```diff
@@ -2799,12 +2806,33 @@ impl<'a> Parser<'a> {
...
+                {
+                    Ok(Expr::UnaryOp {
+                        op: UnaryOperator::BangNot,
+                        expr: Box::new(expr),
+                    })
+                }
+                _ => parser_err!(
+                    "current dialect support bang not operator, but with wrong syntax",
+                    tok.location
+                ),
```
iffyio, line 2834:
```suggestion
```
@wugeer I think one comment, rationale I [described here](https://github.com/apache/datafusion-sqlparser-rs/pull/1472#discussion_r1824158563) could you take a look? (realised it might have been missed since that thread was resolved)

I suspect all that's need there would be to remove this block

iffyio (commented):
@wugeer just one comment I think was missed, otherwise this looks close, thanks!

src/parser/mod.rs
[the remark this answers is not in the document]
wugeer: @iffyio Sorry, I missed this reply.
My intention with this code block was to provide additional context checks and more meaningful error messages. Without it, the error messages would be less precise, so I’m hesitant about whether to remove it or not.
without it
```bash
 cargo run --features json_example --example cli err.sql --hive
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/examples/cli err.sql --hive`
Parsing from file 'err.sql' using HiveDialect
2024-10-31T23:24:09.204Z DEBUG [sqlparser::parser] Parsing sql 'select a!, (b+c)!, C+D !
'...
2024-10-31T23:24:09.204Z DEBUG [sqlparser::parser] parsing expr
2024-10-31T23:24:09.204Z DEBUG [sqlparser::parser] prefix: Identifier(Ident { value: "a", quote_style: None })
2024-10-31T23:24:09.204Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithLocation { token: ExclamationMark, location: Location { line: 1, column: 9 } }
2024-10-31T23:24:09.204Z DEBUG [sqlparser::parser] next precedence: 50
Error during parsing: ParserError("No infix parser for token ExclamationMark at Line: 1, Column: 9")
```
currently
```bash
cargo run --features json_example --example cli err.sql --hive
   Compiling sqlparser v0.51.0 (/home/xiehw/code/person/rust/sqlparser-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.04s
     Running `target/debug/examples/cli err.sql --hive`
Parsing from file 'err.sql' using HiveDialect
2024-10-31T23:24:17.700Z DEBUG [sqlparser::parser] Parsing sql 'select a!, (b+c)!, C+D !
'...
2024-10-31T23:24:17.700Z DEBUG [sqlparser::parser] parsing expr
2024-10-31T23:24:17.700Z DEBUG [sqlparser::parser] prefix: Identifier(Ident { value: "a", quote_style: None })
2024-10-31T23:24:17.700Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithLocation { token: ExclamationMark, location: Location { line: 1, column: 9 } }
2024-10-31T23:24:17.700Z DEBUG [sqlparser::parser] next precedence: 50
Error during parsing: ParserError("current dialect support bang not operator, but with wrong syntax at Line: 1, Column: 9")
```

src/parser/mod.rs
wugeer: @iffyio Thank you for the reminder and the review. Have Done! :)

iffyio: Oh @wugeer do we need to push local changes to the MR?, this part of the code still looks unchanged?

wugeer: @iffyio  Let’s remove  this complex context validation logic in this part for now to keep the original logic simple. I've pushed the updated code. Thank you for review! :)

iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

wugeer:
@alamb Could you spare some time to review this PR? Thanks! :)

alamb (approved):
Looks good to me -- thank you @wugeer  and @iffyio

PR #1474, opened 2024-10-12
Fix #1469 (SET ROLE regression)

### LINKED ISSUE #1469: regression: `SET role = 'foobar'` was broken by the latest release
Using a sql keyword as a variable name used to work in previous versions and was broken by the latest release.
This has caused issues downstream like https://github.com/sqlpage/SQLPage/discussions/644

The problem appears with:

```sql
SET role = 'foobar';
```

which is valid SQL (at least in postgres), and which now fails parsing with `ParserError("Expected: identifier, found: =")`

lovasoa:
this fixes a regression that was introduced with the new SetRole statement

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9416,29 +9416,24 @@ impl<'a> Parser<'a> {
         }
     }
 
-    fn parse_set_role(&mut self, modifier: Option<Keyword>) -> Option<Statement> {
+    /// Parse a `SET ROLE` statement. Expects SET to be consumed already.
+    fn parse_set_role(&mut self, modifier: Option<Keyword>) -> Result<Statement, ParserError> {
+        self.expect_keyword(Keyword::ROLE)?;
         let context_modifier = match modifier {
             Some(Keyword::LOCAL) => ContextModifier::Local,
             Some(Keyword::SESSION) => ContextModifier::Session,
             _ => ContextModifier::None,
         };
 
-        if self.parse_keyword(Keyword::NONE) {
-            Some(Statement::SetRole {
-                context_modifier,
-                role_name: None,
-            })
-        } else if matches!(
-            self.peek_token().token,
-            Token::Word(_) | Token::DoubleQuotedString(_) | Token::SingleQuotedString(_)
-        ) {
-            Some(Statement::SetRole {
-                context_modifier,
-                role_name: Some(self.parse_identifier(false).ok()?),
-            })
-        } else {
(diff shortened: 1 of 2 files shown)


lovasoa:
Hello ! 
  @alamb , this fixes a regression that had a big impact on SQLPage, where it is common to name a variable `role`
  Initially reported in https://github.com/sqlpage/SQLPage/discussions/644#discussioncomment-10914889

src/parser/mod.rs
```diff
@@ -9416,27 +9416,42 @@ impl<'a> Parser<'a> {
...
+
+        if self.parse_keyword(Keyword::NONE) {
+            Some(Statement::SetRole {
+                context_modifier,
+                role_name: None,
+            })
+        } else if matches!(
+            self.peek_token().token,
+            Token::Word(_) | Token::DoubleQuotedString(_) | Token::SingleQuotedString(_)
+        ) {
```
iffyio, line 9434: Not sure I followed the intention with this change, it seems to match the behavior of `parse_identifier()`?

src/parser/mod.rs
```diff
@@ -9416,27 +9416,42 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_set_role(&mut self, modifier: Option<Keyword>) -> Option<Statement> {
```
iffyio, line 9419: thinking maybe this function should return a Result instead as usual, the current rewind behavior doesn't seem correct otherwise?

If this part throws an error, it would have already consumed some tokens from the stream.
```rust
                role_name: Some(self.parse_identifier(false).ok()?),
```
So that the assumption being made at the caller here becomes incorrect
```rust
                self.prev_token(); // replace the "role" token which was actually a variable name
```
(Can we add a test case demonstrating this expected behavior in any case?)

Thinking if `parse_set_role` is a normal parse function, then the caller can instead do `self.maybe_parse(parse_set_role)` which will do the automatic rewind if it turns out not to be a legit `SET ROLE` statement in any scenario?

lovasoa:
I changed the code to use maybe_parse, which is indeed much cleaner and resolves both your points. Thanks @iffyio !

src/parser/mod.rs
lovasoa: parse_identifier cannot return an error here (because of the checks above). But you are right, maybe_parse is both more robust and more readable. I changed the code.

iffyio (approved):
LGTM! cc @alamb

lovasoa:
Thanks for merging, @alamb !

PR #1476, opened 2024-10-15
ClickHouse: support `alter table update ...`

framlog:
It's to support `ALTER TABLE [<database>.]<table> UPDATE <column> = <expression> WHERE <filter_expr>`.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -36,6 +36,8 @@ use crate::ast::{
 use crate::keywords::Keyword;
 use crate::tokenizer::Token;
 
+use super::Assignment;
+
 /// An `ALTER TABLE` (`Statement::AlterTable`) operation
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -142,6 +144,17 @@ pub enum AlterTableOperation {
     UnfreezePartition {
         partition: Partition,
         with_name: Option<Ident>,
+    },
+    /// `UPDATE <column> = <expression> [, ...] [IN PARTITION partition_id] WHERE <filter_expr>`
+    /// Note: this is a ClickHouse-specific operation, please refer to
+    /// [ClickHouse](https://clickhouse.com/docs/en/sql-reference/statements/alter/update)
+    Update {
+        /// Column assignments
+        assignments: Vec<Assignment>,
+        /// PARTITION
+        partition_id: Option<Ident>,
+        /// WHERE
+        selection: Option<Expr>,
     },
     /// `DROP PRIMARY KEY`
     ///
@@ -543,6 +556,20 @@ impl fmt::Display for AlterTableOperation {
                 write!(f, "UNFREEZE {partition}")?;
                 if let Some(name) = with_name {
                     write!(f, " WITH NAME {name}")?;
+                }
+                Ok(())
+            }
+            AlterTableOperation::Update {
+                assignments,
+                partition_id,
+                selection,
+            } => {
+                write!(f, "UPDATE {}", display_comma_separated(assignments))?;
+                if let Some(partition_id) = partition_id {
+                    write!(f, " IN PARTITION {}", partition_id)?;
+                }
+                if let Some(selection) = selection {
+                    write!(f, " WHERE {}", selection)?;
                 }
                 Ok(())
             }
--- a/src/dialect/clickhouse.rs
+++ b/src/dialect/clickhouse.rs
@@ -31,6 +31,11 @@ impl Dialect for ClickHouseDialect {
         self.is_identifier_start(ch) || ch.is_ascii_digit()
     }
 
+    // See https://clickhouse.com/docs/en/sql-reference/statements/alter/update
+    fn supports_alter_table_update(&self) -> bool {
+        true
+    }
+
     fn supports_string_literal_backslash_escape(&self) -> bool {
         true
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1081,7 +1081,7 @@ impl<'a> Parser<'a> {
                     self.parse_bigquery_struct_literal()
                 }
                 Keyword::PRIOR if matches!(self.state, ParserState::ConnectBy) => {
-                    let expr = self.parse_subexpr(self.dialect.prec_value(Precedence::PlusMinus))?;
+                    let expr = self.parse_subexpr(self.dialect.prec_value(Precedence::Pipe))?;
                     Ok(Expr::Prior(Box::new(expr)))
                 }
                 Keyword::MAP if self.peek_token() == Token::LBrace && self.dialect.support_map_literal_syntax() => {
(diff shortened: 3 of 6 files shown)


src/parser/mod.rs
```diff
@@ -7156,6 +7156,26 @@ impl<'a> Parser<'a> {
                 partition,
                 with_name,
             }
+        } else if dialect_of!(self is ClickHouseDialect|GenericDialect)
+            && self.parse_keyword(Keyword::UPDATE)
+        {
+            let assignments = self.parse_comma_separated(Parser::parse_assignment)?;
+            let partition_id =
+                if self.parse_keyword(Keyword::IN) && self.parse_keyword(Keyword::PARTITION) {
```
git-hulk, line 7164:
```suggestion
                if self.parse_keywords(&[Keyword::IN, Keyword::PARTITION]) {
```

tests/sqlparser_clickhouse.rs
```diff
@@ -1422,6 +1422,33 @@ fn parse_create_table_on_commit_and_as_query() {
     }
 }
 
+#[test]
+fn parse_update_data() {
+    let sql = "ALTER TABLE t UPDATE col1 = 1, col2 = col3 + col4 WHERE cod4 = 1";
```
git-hulk, line 1427: Could you please add some negative test cases?

framlog: Updated. But I'm not sure what I've added is negative test cases you mentioned.

git-hulk (approved):
LGTM, cc @iffyio

src/ast/ddl.rs
```diff
@@ -143,6 +145,17 @@ pub enum AlterTableOperation {
         partition: Partition,
         with_name: Option<Ident>,
     },
+    /// `UPDATE <column> = <expression> [, ...] [IN PARTITION partition_id] WHERE <filter_expr>` 
+    /// Note: this is a ClickHouse-specific operation, please refer to
+    /// [ClickHouse](https://clickhouse.com/docs/en/sql-reference/statements/alter/update)
+    UpdateData {
```
iffyio, line 151:
```suggestion
    Update {
```
Thinking we can call it update without the data suffix which doesn't seem to add significant context?

src/parser/mod.rs
```diff
@@ -7156,6 +7156,25 @@ impl<'a> Parser<'a> {
                 partition,
                 with_name,
             }
+        } else if dialect_of!(self is ClickHouseDialect|GenericDialect)
```
iffyio, line 7159: Could you take a look at adding a method to the [dialect trait](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/mod.rs#L63-L72) for this functionality? Also note that testing wise as a result the test in moving to [common.rs in this style](https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_common.rs#L2781)

tests/sqlparser_clickhouse.rs
```diff
@@ -1422,6 +1422,38 @@ fn parse_create_table_on_commit_and_as_query() {
...
```
iffyio, line 1427: Can we add a test case with `IN PARTITION`?

src/ast/ddl.rs
framlog: ok. Renamed.

src/parser/mod.rs
framlog: I changed it and moved the case to the new file.

tests/sqlparser_clickhouse.rs
framlog: Added.

tests/sqlparser_common.rs
```diff
@@ -11432,3 +11432,24 @@ fn test_any_some_all_comparison() {
     verified_stmt("SELECT c1 FROM tbl WHERE c1 <> SOME(SELECT c2 FROM tbl)");
     verified_stmt("SELECT 1 = ANY(WITH x AS (SELECT 1) SELECT * FROM x)");
 }
+
+#[test]
+fn test_parse_alter_table_update() {
+    let dialects = all_dialects_where(|d| d.supports_alter_table_update());
```
iffyio, line 11438: Since the change introduces a new node in the AST, one style of test I think could be useful is to assert the returned AST that it matches what we expect. i.e a test  in this style for [alter_policy](https://github.com/apache/datafusion-sqlparser-rs/blob/17080f1994689009a319cbb8b08314e7c70c228b/tests/sqlparser_common.rs#L11340-L11358)
We can take one scenario like `ALTER TABLE t UPDATE col1 = 1, col2 = 2 IN PARTITION abc WHERE col3 = 1` and verify that the `assignments`, `partition` and `selection` are where we expect them to be in the node

src/dialect/clickhouse.rs
```diff
@@ -31,6 +31,10 @@ impl Dialect for ClickHouseDialect {
         self.is_identifier_start(ch) || ch.is_ascii_digit()
     }
 
+    fn supports_alter_table_update(&self) -> bool {
```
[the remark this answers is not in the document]
framlog: sure

src/parser/mod.rs
```diff
@@ -1081,7 +1081,7 @@ impl<'a> Parser<'a> {
                     self.parse_bigquery_struct_literal()
                 }
                 Keyword::PRIOR if matches!(self.state, ParserState::ConnectBy) => {
-                    let expr = self.parse_subexpr(self.dialect.prec_value(Precedence::PlusMinus))?;
+                    let expr = self.parse_subexpr(self.dialect.prec_value(Precedence::Pipe))?;
```
iffyio, line 1084: Oh was there a reason for this change?

src/parser/mod.rs
```diff
@@ -7156,6 +7156,35 @@ impl<'a> Parser<'a> {
                 partition,
                 with_name,
             }
+        } else if self.dialect.supports_alter_table_update() && self.parse_keyword(Keyword::UPDATE)
+        {
+            let mut assignments = vec![];
```
iffyio, line 7161: One minor request, could we move this new block to its own function to keep the branch of the current function smaller/more-focused?

src/parser/mod.rs
```diff
@@ -7156,6 +7156,35 @@ impl<'a> Parser<'a> {
...
+            loop {
+                let target = self.parse_assignment_target()?;
+                self.expect_token(&Token::Eq)?;
+                // NOTE: Maybe it's better to save the index before parse_subexpr to do a real look
```
iffyio, line 7165: Ah so regarding the earlier comment, I think what we could do would be to make `self.parse_assignment` reusable:

```rust
pub fn parse_assignment(&mut self) -> Result<Assignment, ParserError> {
	self.parse_assignment_with_prec(self.dialect.prec_unknown())
}

fn parse_assignment_with_precedence(self, precedence: u8) -> Result<Assignment, ParserError> {
	let target = self.parse_assignment_target()?;
    self.expect_token(&Token::Eq)?;
	let value = self.parse_subexpr(precedence)?;
    Ok(Assignment { target, value })

}
```
That would make it possible to use `parse_comma_separated` as well, since the issue seems to be with the `IN PARTITION` precedence when parsing the assigned expr

iffyio:
Setting this to draft since its not waiting for review in the meantime, please feel free to undraft when ready!

PR #1480, opened 2024-10-22
locations: Add support for Spans with exact location of AST nodes

lustefaniak:
@jakeswenson initially suggested this PR in https://github.com/apache/datafusion-sqlparser-rs/pull/839. We are using it in SYNQ to drive our code-column lineage experience. I rebased it on the latest main. 

After I rebased it to suggest this change, I noticed an ongoing effort in https://github.com/apache/datafusion-sqlparser-rs/pull/1435.

If you consider the design suggested here for inclusion, I plan to add tests. We have an extensive internal test suite living in a downstream project using a fork of this library.

In our case, the accuracy of code locations is critical since we use it to drive the whole navigation experience. I like the `WithSpan<T>` approach better, as I know what element and where has accurate locations. In our original codebase not all `Indent`s are wrapped in `WithSpan` but since this rebase effort was not a small thing I decided to replace all `Ident` with `WithSpan<Ident>`. Also to reduce future conflicts/PRs to add more locations.

TODO:
- [ ] Add tests ensuring accurate location capture.
- [ ] Remove `empty_span()` from parser. We didn't have some new nodes/quirks on our older fork.
- [ ] Add additional `WithSpan<T>` where it makes sense, e.g. around `Expr`

### DIFF
--- a/src/ast/dcl.rs
+++ b/src/ast/dcl.rs
@@ -29,7 +29,7 @@ use serde::{Deserialize, Serialize};
 use sqlparser_derive::{Visit, VisitMut};
 
 use super::{Expr, Ident, Password};
-use crate::ast::{display_separated, ObjectName};
+use crate::ast::{display_separated, ObjectName, WithSpan};
 
 /// An option in `ROLE` statement.
 ///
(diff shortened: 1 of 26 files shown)


alamb:
Now that we have merged
- https://github.com/apache/datafusion-sqlparser-rs/pull/1435

I wonder if there is some way to reuse / get help testing spans in sqlarser-rs. Thoughts welcome on
- https://github.com/apache/datafusion-sqlparser-rs/issues/1563

PR #1482, opened 2024-10-26
Clickhouse SQL generation for datatypes.

grigoryan-sergey:
The SQL generated by the parser does not work inside Clickhouse. The problem is in datatypes, which are case sensitive in Clickhouse. I suggest generating a Clickhouse version for datatypes. This will still work for other databases where datatypes are not case sensitive (like postgres) and will also work for Clickhouse. Interesting enough some datatypes are also not case sensitive in Clickhouse, but the ones which are case sensitive, are converted in the PR.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -400,7 +400,7 @@ impl fmt::Display for DataType {
                 format_type_with_optional_length(f, "INT4", zerofill, false)
             }
             DataType::Int8(zerofill) => {
-                format_type_with_optional_length(f, "INT8", zerofill, false)
+                format_type_with_optional_length(f, "Int8", zerofill, false)
             }
             DataType::Int16 => {
                 write!(f, "Int16")
@@ -409,7 +409,7 @@ impl fmt::Display for DataType {
                 write!(f, "Int32")
             }
             DataType::Int64 => {
-                write!(f, "INT64")
+                write!(f, "Int64")
             }
             DataType::Int128 => {
                 write!(f, "Int128")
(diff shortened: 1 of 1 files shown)


iffyio (commented):
Hi @grigoryan-sergey hmm with the current changes I expect a lot of tests to fail since we''ve also changed types that affect other dialects. I wonder if there's a way to support both styles :thinking: (I guess first would be if Clickhouse is the only supported dialect with a different behavior)

grigoryan-sergey:
Hi @iffyio. What do you mean by both styles? Camel case as in Clickhouse and uppercase notations? Currently in the code both of this styles are kind of mixed. I agree with you, it would be best to support both styles. In this PR I suggested to change only those which affect Clickhouse SQL generation. 
Basically the styles were mixed before and are also mixed after the PR :). The problem with supporting two styles, is the complication. We need to introduce dialects not for just parsing but also for generating the SQL string. I am not that good at the knowledge of the codebase, so I cannot tell will it be complicated or not to implement. But my current proposal fixes the problem for the database which has case sensitive types (Clickhouse) and does not brake compatability (cases were mixed before and also after).

iffyio:
Ah yeah by both styles I mean that, today when displaying sql, the parser _always_ uses upper case e.g. `FLOAT64`. While the changes in this PR updates that behavior such that for some types it now uses camel case `Float64` - which I expect will break a lot of tests (and worse may be undesirable for other dialects). The changes aren't specific to clickhouse, Float64 for example is used by BigQuery.

I agree that it will be quite invasive to make this behavior configurable. On the other hand, I fear that using camel case for some types vs upper case for others will be quite haphazard (especially if Clickhouse isn't alone with this requirement).

Currently I'm not sure what a good way to approach this issue is actually :thinking:

pravic:
@iffyio This looks like a job for the `Dialect` trait. Or for a new trait, e.g. `Formatter`.

PR #1483, opened 2024-10-27
Fix build

yoavcloud:
It appears the build has been broken for several days now: https://github.com/apache/datafusion-sqlparser-rs/actions/runs/11447418489

### DIFF
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -230,15 +230,18 @@ pub fn parse_comment(parser: &mut Parser) -> Result<Statement, ParserError> {
 }
 
 pub fn parse_create(parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
-    let name = parser.maybe_parse(|parser| -> Result<ObjectName, ParserError> {
+    if let Ok(name) = parser.maybe_parse(|parser| -> Result<ObjectName, ParserError> {
         parser.expect_keyword(Keyword::CREATE)?;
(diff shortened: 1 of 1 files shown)


wugeer:
@yoavcloud Hello, maybe you can link the issue https://github.com/apache/datafusion-sqlparser-rs/issues/1484 to this PR. :)

src/dialect/postgresql.rs
```diff
@@ -230,15 +230,18 @@ pub fn parse_comment(parser: &mut Parser) -> Result<Statement, ParserError> {
 }
 
 pub fn parse_create(parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
-    let name = parser.maybe_parse(|parser| -> Result<ObjectName, ParserError> {
+    if let Ok(name) = parser.maybe_parse(|parser| -> Result<ObjectName, ParserError> {
```
iffyio, line 233: Ah should this rather return the error? since the error would be a recursion error, so that the return type fot `parse_create` becomes `Result<Option<Result<statement`

yoavcloud: True, updated in the next commit

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

alamb:
Thank you @yoavcloud  -- we are very close to adding more review capacity in this repo. Stay tuned. 

Thank you as always @iffyio

alamb (approved):

PR #1487, opened 2024-10-30
Add support for SHOW DATABASES/SCHEMAS/TABLES/VIEWS in Hive

yoavcloud:
This PR adds support for the above statements in the Hive dialect, making the previously MySQL-only structs a bit more generic.


src/ast/mod.rs
```diff
@@ -4376,7 +4410,31 @@ impl fmt::Display for Statement {
                     full = if *full { "FULL " } else { "" },
                 )?;
                 if let Some(db_name) = db_name {
-                    write!(f, " FROM {db_name}")?;
+                    let keyword = match &db_name_keyword {
+                        Some(Keyword::FROM) => "FROM",
+                        Some(Keyword::IN) => "IN",
+                        _ => "", // unexpected
```
git-hulk, line 4416: Can use `unreachable!()` to panic if it's an unexpected condition.

src/ast/mod.rs
```diff
@@ -4376,7 +4410,31 @@ impl fmt::Display for Statement {
...
+                db_name_keyword, 
+                db_name, 
+                filter 
+            } => {
+                write!(f, "SHOW VIEWS")?;
+                if let Some(db_name) = db_name {
+                    let keyword = match &db_name_keyword {
+                        Some(Keyword::FROM) => "FROM",
+                        Some(Keyword::IN) => "IN",
+                        _ => "", // unexpected
```
git-hulk, line 4435: Same as the above comment.

tests/sqlparser_hive.rs
```diff
@@ -534,6 +534,20 @@ fn parse_use() {
     );
 }
 
+#[test]
+fn test_show() {
+    hive().verified_stmt("SHOW DATABASES");
```
git-hulk, line 539: It would be better to use `hive_and_generic` since it also works for generic dialects.

src/ast/mod.rs
yoavcloud: Fixed

src/ast/mod.rs
yoavcloud: Fixed

tests/sqlparser_hive.rs
yoavcloud: Fixed

tests/sqlparser_mysql.rs
```diff
@@ -374,13 +380,15 @@ fn parse_show_tables() {
...
+            db_name_keyword: None,
             db_name: None,
             filter: Some(ShowStatementFilter::Where(
                 mysql_and_generic().verified_expr("1 = 2")
             )),
         }
     );
-    mysql_and_generic().one_statement_parses_to("SHOW TABLES IN mydb", "SHOW TABLES FROM mydb");
+    mysql_and_generic().verified_stmt("SHOW TABLES IN mydb");
+    mysql_and_generic().verified_stmt( "SHOW TABLES FROM mydb");
```
git-hulk, line 391:
```suggestion
    mysql_and_generic().verified_stmt("SHOW TABLES FROM mydb");
```

git-hulk:
cc @iffyio

tests/sqlparser_mysql.rs
yoavcloud: Fixed

git-hulk (approved):
LGTM, cc @iffyio

tests/sqlparser_hive.rs
```diff
@@ -534,6 +534,20 @@ fn parse_use() {
...
 
+#[test]
+fn test_show() {
+    hive_and_generic().verified_stmt("SHOW DATABASES");
+    hive_and_generic().verified_stmt("SHOW DATABASES LIKE '%abc'");
+    hive_and_generic().verified_stmt("SHOW SCHEMAS");
+    hive_and_generic().verified_stmt("SHOW SCHEMAS LIKE '%abc'");
+    hive_and_generic().verified_stmt("SHOW TABLES");
+    hive_and_generic().verified_stmt("SHOW TABLES IN db1");
+    hive_and_generic().verified_stmt("SHOW TABLES IN db1 'abc'");
```
iffyio, line 545: for tables/views in addition to `IN` can we also include testcases for `FROM`?

tests/sqlparser_hive.rs
```diff
@@ -534,6 +534,20 @@ fn parse_use() {
     );
 }
 
+#[test]
```
iffyio, line 537: thinking we can move this test over to common? the current code isn't specific to hive and the feature/syntax seems popular enough that the parser can always accept it as it does today

src/ast/mod.rs
```diff
@@ -4376,7 +4410,31 @@ impl fmt::Display for Statement {
...
                 )?;
                 if let Some(db_name) = db_name {
-                    write!(f, " FROM {db_name}")?;
+                    let keyword = match &db_name_keyword {
+                        Some(Keyword::FROM) => "FROM",
+                        Some(Keyword::IN) => "IN",
+                        _ => unreachable!(),
+                    };
+                    write!(f, " {} {db_name}", keyword)?;
+                }
```
iffyio, line 4419: I think representation wise we can represent the IN/FROM explicitly? e.g.

```rust
enum ShowClause {
   IN,
   FROM
}
Statement::ShowViews{ clause: Option<ShowClause>, db_name: Option<Ident> }
```
The unreachable panic I don't think fits well here since there's no invariant guarding that panic (the parser codepath is quite far away from this and could possibly change subtly to trigger it), I think it'd be reasonable as in the previous version to display an incorrect sql (which would be less consequential given we have a bug vs than crash the user's service). Though in this case likely we can rely on the compiler to make such a state unavoidable via the enum route.

Also would it make more sense to skip the nested `if let` since the db_name and db_name_keyword are set independently. as in

```rust
if let Some(db_name_keyword) = db_name_keyword {
    write(f, " {db_name_keyword}")?;
}
if let Some(db_name) = db_name {
    write(f, " {db_name})?;
}
```

src/ast/mod.rs
```diff
@@ -2782,12 +2783,30 @@ pub enum Statement {
         filter: Option<ShowStatementFilter>,
     },
     /// ```sql
+    /// SHOW DATABASES [LIKE 'pattern']
+    /// ```
+    ShowDatabases { pattern: Option<String> },
```
iffyio, line 2788: Would it make sense to also use `ShowStatementFilter` to represent this (if there's a chance that the syntax can be extended)?

tests/sqlparser_hive.rs
yoavcloud: Sure, according to the hive docs, it's only for views though.

tests/sqlparser_hive.rs
yoavcloud: I understand the preference now is for the parser to err on the side of accepting queries even if the dialect doesn't actually support them right? If so, then yes, we can move the tests to common. The SHOW syntax is somewhat popular but it's not a standard.

iffyio: Yeah we have a few scenarios where the parser accepts input that isn't supported by the dialect. I figured here it could be easier to do so just as well, the alternative would be to add explicit methods to the dialect trait i.e `dialect.supports_show_databases()` `dialect.supports_show_schemas()` etc which I think would also be fine if that's preferrable from an impl pov

src/ast/mod.rs
yoavcloud: Great idea

tests/sqlparser_hive.rs
yoavcloud: Kept it at the generic level. I have some more SHOW statements for Snowflake to add, so we'll see if these challenge this decision.

iffyio (approved):
LGTM! cc @alamb

yoavcloud:
Sorry, did a last change and forgot cargo fmt... Changes committed now, if you can run the CI again that would be great @alamb

alamb:
🚀

yoavcloud:
@iffyio @alamb why isn't this merged?

alamb:
> @iffyio @alamb why isn't this merged?

Sorry -- am behind.

alamb (approved):

yoavcloud:
Thank you!

PR #1488, opened 2024-10-30
Fix complex blocks warning when running clippy

git-hulk:
```shell
warning: in a `match` scrutinee, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
   --> src/dialect/postgresql.rs:233:74
    |
233 |       match parser.maybe_parse(|parser| -> Result<ObjectName, ParserError> {
    |  __________________________________________________________________________^
234 | |         parser.expect_keyword(Keyword::CREATE)?;
235 | |         parser.expect_keyword(Keyword::TYPE)?;
236 | |         let name = parser.parse_object_name(false)?;
...   |
239 | |         Ok(name)
240 | |     }) {
    | |_____^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#blocks_in_conditions
    = note: `#[warn(clippy::blocks_in_conditions)]` on by default
```

### DIFF
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -230,14 +230,16 @@ pub fn parse_comment(parser: &mut Parser) -> Result<Statement, ParserError> {
 }
 
 pub fn parse_create(parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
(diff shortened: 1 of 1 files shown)


iffyio (approved):
LGTM! cc @alamb

Issue #1492, opened 2024-10-30
add support for `FOR ORDINALITY` in `JSON_TABLE`


lovasoa:
Initially reported in https://github.com/sqlpage/SQLPage/issues/658

This is a follow-up on https://github.com/apache/datafusion-sqlparser-rs/issues/1019

I already implemented most of `json_table`, but we are still missing support for  `FOR ORDINALITY` columns.

Reference: https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html#function_json-table

Exemple:

```sql
SELECT jt.*
FROM JSON_TABLE(
  '["Alice", "Bob", "Charlie"]',
  '$[*]' COLUMNS(
    row_num FOR ORDINALITY,
    name VARCHAR(50) PATH '$'
  )
) AS jt;
```

[try online](https://sqliteonline.com/#sqltext=%23url-sqlite%3Ddb-mariadb%0D%0A%23tab-name%3DMariaDB%0D%0ASELECT%20jt.*%0AFROM%20JSON_TABLE(%0A%20%20'%5B%22Alice%22%2C%20%22Bob%22%2C%20%22Charlie%22%5D'%2C%0A%20%20'%24%5B*%5D'%20COLUMNS(%0A%20%20%20%20row_num%20FOR%20ORDINALITY%2C%0A%20%20%20%20name%20VARCHAR(50)%20PATH%20'%24'%0A%20%20)%0A)%20AS%20jt%3B)

PR #1493, opened 2024-10-30
add support for `FOR ORDINALITY` and `NESTED` in JSON_TABLE

lovasoa:
fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1492

see the specs here: https://dev.mysql.com/doc/refman/8.0/en/json-table-functions.html#function_json-table

@alamb : this is a fix for the following bug in SQLPage: https://github.com/sqlpage/SQLPage/issues/658

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2281,14 +2281,65 @@ impl fmt::Display for ForJson {
 /// FROM JSON_TABLE(
 ///     '["a", "b"]',
 ///     '$[*]' COLUMNS (
-///         value VARCHAR(20) PATH '$'
+///         name FOR ORDINALITY,
+///         value VARCHAR(20) PATH '$',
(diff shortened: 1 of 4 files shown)


src/ast/query.rs
```diff
@@ -2281,14 +2281,65 @@ impl fmt::Display for ForJson {
 /// FROM JSON_TABLE(
```
iffyio, line 2281: Could we include the link to the mysql doc here?

iffyio (approved):
LGTM! cc @alamb

alamb:
Thanks @lovasoa and @iffyio

lovasoa:
Thank you for merging, @alamb :blush: 

Could you maybe also take a look at https://github.com/apache/datafusion-sqlparser-rs/pull/1490 ?

Issue #1450, opened 2024-10-01
`select from @local_var` fails to parse


joshua-auchincloss:
Noting the engine fails to parse valid local variables, despite being previously declared. E.g query below to highlight this (engine = mssql)

```sql
declare @a int
set @a = 2
select @a * 4
```

Raises end of statement error (`found @a at Line 3, column xxx`)

wugeer:
What's more interesting is that after I swapped the order of the SQL statements, the result was successfully output.
![image](https://github.com/user-attachments/assets/891aa9ec-e500-432b-838d-2ac3ef01fc3e)
![image](https://github.com/user-attachments/assets/451a1ec0-f69f-42a5-af60-80ac072564be)
![image](https://github.com/user-attachments/assets/19e35ffd-30de-4a0a-898f-de450d18734f)
![image](https://github.com/user-attachments/assets/ac352823-29e0-49f7-abd9-140fbc82c0e4)

PR #1497, opened 2024-11-01
Fix the parsing error in MSSQL for multiple statements that include `DECLARE` statements

wugeer:
This PR resolves an error in MSSQL that occurs when parsing multiple SQL statements containing `DECLARE` statements.

before
parse sql in mssql
```
declare @a int;
set @a = 2;
select @a * 4;
```
raise error
```
Error during parsing: ParserError("Expected: end of statement, found: set at Line: 2, Column: 1")
```

This resolves issue https://github.com/apache/datafusion-sqlparser-rs/issues/1450

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5311,55 +5311,61 @@ impl<'a> Parser<'a> {
     /// ```
     /// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/declare-local-variable-transact-sql?view=sql-server-ver16
     pub fn parse_mssql_declare(&mut self) -> Result<Statement, ParserError> {
-        let mut stmts = vec![];
-
-        loop {
-            let name = {
-                let ident = self.parse_identifier(false)?;
-                if !ident.value.starts_with('@') {
-                    Err(ParserError::TokenizerError(
-                        "Invalid MsSql variable declaration.".to_string(),
-                    ))
-                } else {
-                    Ok(ident)
-                }
-            }?;
-
-            let (declare_type, data_type) = match self.peek_token().token {
-                Token::Word(w) => match w.keyword {
-                    Keyword::CURSOR => {
-                        self.next_token();
-                        (Some(DeclareType::Cursor), None)
-                    }
-                    Keyword::AS => {
-                        self.next_token();
-                        (None, Some(self.parse_data_type()?))
-                    }
-                    _ => (None, Some(self.parse_data_type()?)),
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -5335,7 +5335,7 @@ impl<'a> Parser<'a> {
                 for_query: None,
             });
 
-            if self.next_token() != Token::Comma {
+            if self.peek_token().token == Token::SemiColon || self.next_token() != Token::Comma {
```
iffyio, line 5338: Oh could this be simplified if we reused the `self.parse_comma_separated()` variant? I think if thats' possible it would be more conventional to use in the codebase (and would've prevented the edge case)

tests/sqlparser_mssql.rs
```diff
@@ -626,6 +626,72 @@ fn parse_mssql_declare() {
         }],
         ast
     );
+
+    let sql = "DECLARE @bar INT;SET @bar = 2;SELECT @bar * 4";
+    let ast = Parser::parse_sql(&MsSqlDialect {}, sql).unwrap();
```
iffyio, line 631: Can this use `ms().parse_sql_statements(sql)`?

src/parser/mod.rs
iffyio: @wugeer (just double checking since there was a requested review on this MR), did you get a chance to visit this comment yet?

just in case, to clarify I was thinking as the fix we could just move the body of the loop (minus line) into a method and call that method via parse_comma_separated. for example
```rust
    fn parse_mssql_declare_stmt(&mut self) -> Result<Declare, ParserError> {
        let stmts = self.parse_comma_separated(Parser::parse_mssql_declare_stmt)?;
        Ok(Statement::Declare { stmts })
    }
    fn parse_mssql_declare_stmt(&mut self) -> Result<Declare, ParserError> {
        let name = { ...};
        // ... remaining body verbatim
        let assignment = self.parse_mssql_variable_declaration_expression()?;
        Ok(Declare { names, ... })
    }
```

iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

iffyio:
@wugeer in the meantime when you get the chance could you take a look at the compile issue? i think it likely just needs the latest changes from main incorporated

wugeer:
@iffyio Thank you for your guidance. I have already fixed it. :)

alamb:
Thanks again @wugeer and @iffyio

Issue #1494, opened 2024-10-31
Not support mssql `OPENJSON WITH` syntax


gaoqiangz:
```sql
SELECT B.kind,B.id_list
FROM table1 A
CROSS APPLY OPENJSON(A.param,'$.config') WITH (kind VARCHAR(20) '$.kind',id_list NVARCHAR(MAX) '$.id_list' AS JSON) B
```

```
sql parser error: Expected: ), found: VARCHAR at Line: 3, Column: 53
```

<https://learn.microsoft.com/en-us/sql/t-sql/functions/openjson-transact-sql?view=sql-server-ver16#syntax>

PR #1498, opened 2024-11-04
Add support for MSSQL's `OPENJSON WITH` clause

gaoqiangz:
Resolves issue #1494


src/parser/mod.rs
```diff
@@ -10000,7 +10000,7 @@ impl<'a> Parser<'a> {
                     table_with_joins: Box::new(table_and_joins),
                     alias,
                 })
-            } else if dialect_of!(self is SnowflakeDialect | GenericDialect) {
+            } else if dialect_of!(self is SnowflakeDialect | GenericDialect | MsSqlDialect) {
```
lovasoa, line 10003: You may want to update the comments below too.

src/ast/query.rs
```diff
@@ -2346,6 +2386,40 @@ impl fmt::Display for JsonTableColumnErrorHandling {
...
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct OpenJsonTableColumn {
+    /// The name of the column to be extracted.
+    pub name: Ident,
+    /// The type of the column to be extracted.
+    pub r#type: DataType,
+    /// The path to the column to be extracted. Must be a literal string.
+    pub path: Option<Value>,
```
lovasoa, line 2405: If this needs to be a string, then maybe store an Option<String> ?

lovasoa (approved):
No blocker for me ! This would be of great use for us at SQLPage, where the use of sql json functions is common.

src/parser/mod.rs
```diff
@@ -10133,6 +10134,30 @@ impl<'a> Parser<'a> {
...
+            } else {
+                Vec::new()
+            };
+            let alias = self.parse_optional_table_alias(keywords::RESERVED_FOR_TABLE_ALIAS)?;
+            Ok(TableFactor::OpenJsonTable {
+                json_expr,
+                json_path,
+                columns,
+                alias,
+            })
```
iffyio, line 10160: One request (I think the other branches in this function don't do a good job of it :) could we move this to its own function e.g `self.parse_open_json_table_factor()` and call that function from this branch instead? thinking in order to keep the branches shorter/more-focused

src/parser/mod.rs
iffyio: Yeah to @lovasoa's suggestion, one double check on my end @gaoqiangz is the parenthesis around the joined table supported syntax for the openjson syntax? Thinking if its not, then we would be able to skip this change entirely. If it is then indeed lets update the comments to account for mssql.
Additionally, can we add a test case covering the behavior (as in a roundtrip test like `SELECT ... FROM T AS A CROSS APPLY (OPENJSON(...) WITH ...) AS B` so that we catch any regression on the behavior going forward

tests/sqlparser_mssql.rs
```diff
@@ -192,6 +192,35 @@ fn parse_mssql_apply_join() {
     );
 }
 
+#[test]
+fn parse_mssql_cross_apply_json() {
```
iffyio, line 196: Could we add one scenario that checks the returned AST (i.e that we get a TableFactor::OpenJSON containing the intended components of the input)?

tests/sqlparser_mssql.rs
```diff
@@ -192,6 +192,35 @@ fn parse_mssql_apply_join() {
...
+    let _ = ms().verified_only_select(
+        "SELECT B.kind, B.id_list \
+        FROM t_test_table AS A \
+        CROSS APPLY OPENJSON(A.param, '$.config') WITH (kind VARCHAR(20) '$.kind', [id_list] NVARCHAR(MAX) '$.id_list' AS JSON) AS B",
+    );
```
iffyio, line 201:
```suggestion
    ms_and_generic().verified_only_select(
        "SELECT B.kind, B.id_list \
        FROM t_test_table AS A \
        CROSS APPLY OPENJSON(A.param, '$.config') WITH (kind VARCHAR(20) '$.kind', [id_list] NVARCHAR(MAX) '$.id_list' AS JSON) AS B",
    );
```

src/parser/mod.rs
gaoqiangz: I have checked that MSSQL doesn't support parenthesis around the openjson syntax.
I'll revert this change.

tests/sqlparser_mssql.rs
gaoqiangz: `GenericDialect` doesn't support brackets around the identifier.

iffyio (approved):
LGTM! cc @alamb

alamb:
🚀

PR #1500, opened 2024-11-06
Add support for MSSQL's `XQuery` methods

gaoqiangz:
Supports  `XQuery` methods:

```sql
SELECT STUFF((SELECT ',' + name FROM sys.objects FOR XML PATH(''), TYPE).value('.', 'NVARCHAR(MAX)'), 1, 1, '') AS T
SELECT CAST(column AS XML).value('.', 'NVARCHAR(MAX)') AS T
```

See <https://learn.microsoft.com/en-us/sql/t-sql/xml/xml-data-type-methods?view=sql-server-ver16>.


src/parser/mod.rs
```diff
@@ -1269,6 +1269,50 @@ impl<'a> Parser<'a> {
             _ => self.expected("an expression", next_token),
         }?;
 
+        if dialect_of!(self is MsSqlDialect) {
+            // Convert `CompositeAccess` to `CompositeFunction` (MSSQL doesn't support `CompositeAccess` syntax)
+            // ```sql
+            // SELECT (SELECT ',' + name FROM sys.objects  FOR XML PATH(''), TYPE).value('.','NVARCHAR(MAX)')
```
iffyio, line 1275: is it possible to represent this syntax as the existing `Expr::JsonAccess`? I'm assuming that isn't the case since otherwise the the feature will be already be supported. If not its not super clear to me what exactly is the composite access we're introducing with this PR - My assumption is that its adding support for syntax like `<arbitrary-expr>.<arbitrary-expr>.<arbitrary-expr>...` and so on where `arbitrary-expr` can be any expr including a function call. Like in the example `SELECT CONVERT(XML,'<Book>abc</Book>').value('.','NVARCHAR(MAX)')`

gaoqiangz: `CompositeFunction` is similar to `CompositeAccess` but supports method call syntax, that supports method calls on subquery results (must be `XML` data type in MSSQL).
Maybe it would be better to make  `CompositeFunction` a generic feature?

src/ast/mod.rs
```diff
@@ -560,6 +560,13 @@ pub enum Expr {
         expr: Box<Expr>,
         key: Ident,
     },
+    /// CompositeFunction (mssql) e.g. SELECT (SELECT ',' + name FROM sys.objects  FOR XML PATH(''), TYPE).value('.','NVARCHAR(MAX)')   
+    /// <https://learn.microsoft.com/en-us/sql/t-sql/xml/xml-data-type-methods?view=sql-server-ver16>
```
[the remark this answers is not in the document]
gaoqiangz: There is no special syntax definition, just ordinary method calling syntax.

gaoqiangz:
I redesigned `CompositeFunction`, please take a look.

tests/sqlparser_common.rs
```diff
@@ -11387,6 +11387,372 @@ fn test_try_convert() {
...
+                    within_group: vec![]
+                }
+            }),
+            alias: Ident {
+                value: "T".into(),
+                quote_style: None
+            }
+        }],
+        sql.projection
+    );
```
lovasoa, line 11666: I find these tests hard to follow. And they will need to be updated even when the ast changes in places completely unrelated to Expr::CompositeFunction.

Maybe you could just test the new syntax introduced here with small precise asserts that test just a single thing ? Like:

```rs
assert!eq!(verified_expr("convert(XML, '<x/>').query('x')", some_small_ast);
```

If you want to ensure some more complex queries always parse, you could add some verified_only_select, without adding an assert on the result.

This should make the tests easier to understand, this pr easier to review, and reduce the burden of maintaining these tests in the future.

src/ast/mod.rs
```diff
@@ -808,6 +808,21 @@ pub enum Expr {
...
+    ///
+    /// > `arbitrary-expr` can be any expression including a function call.
+    ///
+    /// Example:
+    ///
+    /// ```sql
+    /// SELECT (SELECT ',' + name FROM sys.objects  FOR XML PATH(''), TYPE).value('.','NVARCHAR(MAX)')   
+    /// SELECT CONVERT(XML,'<Book>abc</Book>').value('.','NVARCHAR(MAX)')
+    /// ```
+    CompositeFunction(CompositeFunction),
```
lovasoa, line 825: Maybe we can find a better name ? Something like `MethodCall  { expr, method }` instead of `CompositeFunction { left, right }`

src/parser/mod.rs
```diff
@@ -1271,10 +1271,22 @@ impl<'a> Parser<'a> {
...
                     };
-                    Ok(Expr::CompositeAccess {
-                        expr: Box::new(expr),
-                        key,
-                    })
+                    if self.consume_token(&Token::LParen) {
+                        self.prev_token();
+                        let func = match self.parse_function(ObjectName(vec![key]))? {
+                            Expr::Function(func) => func,
+                            _ => unreachable!(),
```
iffyio, line 1278: Oh I think this panic won't be desirable here in any case given we don't have an invariant and `self.parse_function` could be in its own right to return something other than `Expr::Function` in a different situation.

Also given the nature of the syntax that we don't have any special keyword to match on, feels a bit uneasy that we could end up parsing sql incorrectly in other dialects in ambiguous cases.

could we do similar to e.g. [try_parse_lambda](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/parser/mod.rs#L1332-L1346) where this is both specific to mssql dialect, as well as potentially relying on `self.maybe_parse` to bail out if the lookahead isn't what we expect?

src/parser/mod.rs
```diff
@@ -1288,6 +1300,35 @@ impl<'a> Parser<'a> {
             _ => self.expected("an expression", next_token),
         }?;
 
+        // Composite function chain
+        while matches!(
+            expr,
+            Expr::Function(_)
+                | Expr::CompositeFunction { .. }
+                | Expr::Cast { .. }
+                | Expr::Convert { .. }
```
iffyio, line 1309: hmm some of the entries seem a bit arbitary (like cast, convert)? I wonder if instead of having this block as a special case, piggy-backing on my other comment to do similar to the lambda parsing, we could probably have the chain continuation logic in the same helper function that gets called once here (as in the function knows to parse not just a single function but also continuation/suffix if present of the full expression)

iffyio (commented):
Took a quick look and left some highlevel comment, will make time soon to take a closer look. Thanks for clarifying the behavior @gaoqiangz together with the rework the intention is much clearer to me!

gaoqiangz:
I added `supports_methods` to the `Dialect` trait and `try_parse_method` to the `Parser`, this might be better suited for different scenarios and doesn't break existing behavior.

src/ast/mod.rs
```diff
@@ -808,6 +808,21 @@ pub enum Expr {
     },
     /// Scalar function call e.g. `LEFT(foo, 5)`
     Function(Function),
+    /// Arbitrary expr method call
+    ///
+    /// Syntax:
+    ///
+    /// `<arbitrary-expr>.<arbitrary-expr>.<arbitrary-expr>...`
+    ///
+    /// > `arbitrary-expr` can be any expression including a function call.
```
iffyio, line 817: Ah if i understand correctly I think this comment might not be accurate. My understanding is that first expression is indeed an arbitrary expression but subsequent expressions must be function calls?

src/ast/mod.rs
```diff
@@ -808,6 +808,21 @@ pub enum Expr {
...
+    ///
+    /// Example:
+    ///
+    /// ```sql
+    /// SELECT (SELECT ',' + name FROM sys.objects  FOR XML PATH(''), TYPE).value('.','NVARCHAR(MAX)')   
+    /// SELECT CONVERT(XML,'<Book>abc</Book>').value('.','NVARCHAR(MAX)').value('.','NVARCHAR(MAX)')
```
iffyio, line 823: I saw this doc that seems to describe the feature and could be useful to include here (for a lack of syntax spec from mssql)?
https://learn.microsoft.com/en-us/sql/relational-databases/server-management-objects-smo/create-program/calling-methods?view=sql-server-ver16

src/ast/mod.rs
```diff
@@ -808,6 +808,21 @@ pub enum Expr {
...
+    /// ```
+    Method(Method),
```
iffyio, line 825: Should we maybe call this MethodInvocation vs Method? thinking the latter might suggest similar to Function that we're looking to represent a method body

src/ast/mod.rs
```diff
@@ -5593,6 +5609,21 @@ impl fmt::Display for FunctionArgumentClause {
...
 }
 
+/// A method call
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct Method {
+    pub expr: Box<Expr>,
+    pub method: Function,
+}
```
iffyio, line 5619: representation wise I'm thinking we could use something like
```rust
struct Method {
    pub prefix: Box<Expr>,
    pub suffix: Vec<Function>
}
```
Actually I'm not super sure naming wise of the properties, maybe we can come up with something better. But the main point/change I'm thinking would be the overall representation. Instead of using a nested datastructure for this, that doesn't seem required, the syntax is just a chain of expression so we can just use list to store that chain - and it maps nicely to how the parsing is done since its clear that the first part can be any expr, while subsequent components must be functions. wdyt?

src/parser/mod.rs
```diff
@@ -1258,23 +1258,27 @@ impl<'a> Parser<'a> {
...
-                            )
-                        }
-                    };
-                    Ok(Expr::CompositeAccess {
-                        expr: Box::new(expr),
-                        key,
-                    })
+                    if !self.consume_token(&Token::Period) {
+                        Ok(expr)
+                    } else {
```
iffyio, line 1266: I think the else block can be unnested one level to be?
```rust
else if !self.consume_token(...) {
   Ok(expr)
} else {
   let tok = ...
}
```

src/parser/mod.rs
```diff
@@ -1345,6 +1354,45 @@ impl<'a> Parser<'a> {
         })
     }
 
+    fn try_parse_method(&mut self, expr: &Expr) -> Result<Option<Expr>, ParserError> {
```
iffyio, line 1357: instead of passing a reference we can use owned values, such that here we return the same expression if we didn't parse anything. That would avoid cloning the expression. I think with that and the linear representation of the method expression, the parsing logic could be slightly simpler (since it won't need to mutate anything, given that the maybe_parse section would only deal with the suffix)

iffyio (commented):
@gaoqiangz took a look now and left some comments

src/ast/mod.rs
iffyio: Actually I think never mind this comment, realised this is indeed similar to Function as a function call, not sure why I was thinking of a different Function body

src/ast/mod.rs
gaoqiangz: good advice!

src/parser/mod.rs
```diff
@@ -3532,9 +3580,9 @@ impl<'a> Parser<'a> {
     }
 
     /// Run a parser method `f`, reverting back to the current position if unsuccessful.
-    pub fn maybe_parse<T, F>(&mut self, mut f: F) -> Result<Option<T>, ParserError>
+    pub fn maybe_parse<T, F>(&mut self, f: F) -> Result<Option<T>, ParserError>
     where
-        F: FnMut(&mut Parser) -> Result<T, ParserError>,
+        F: FnOnce(&mut Parser) -> Result<T, ParserError>,
```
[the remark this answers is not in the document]
gaoqiangz: This change is no longer needed.

gaoqiangz:
The code has been rewritten for the new `Method` definition, seems more reasonable:

```rs
pub struct Method {
    pub expr: Box<Expr>,
    // always non-empty
    pub method_chain: Vec<Function>,
}
```

Any other comments?

src/ast/mod.rs
gaoqiangz: This document is not about T-SQL.
Indeed, no exact syntax definition was found, another related documents like this:
https://learn.microsoft.com/en-us/sql/xquery/xquery-language-reference-sql-server?view=sql-server-ver16&redirectedfrom=MSDN

src/parser/mod.rs
```diff
@@ -1345,6 +1355,38 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// Parses method call expression
+    ///
+    /// Returns `Result<{parsed method-call-expr},{orig_expr}>`
+    fn try_parse_method(&mut self, orig_expr: Expr) -> Result<Result<Expr, Expr>, ParserError> {
+        if !self.dialect.supports_methods() {
+            return Ok(Err(orig_expr));
+        }
+        let method_chain = self.maybe_parse(|p| {
```
iffyio, line 1365: I think we can simplify the return type a bit if we do something like
```rust
let method_chain = self.maybe_parse(|p| {
  let mut method_chain = Vec::new();
  loop {
    // ...
    method_chain.push(func);
  }
  if method_chain.is_empty() {
    p.expected("function")
  } else {
    Ok(method_chain)
  }
})?;

if let Some(method_chain) {
  Ok(Expr::Method{Method { expr, method_chain }})
} else {
  Ok(expr)
}
```
that would remove the nested result type `Result<Result<Expr, Expr>>` -> `Result<Expr>` as usual

src/parser/mod.rs
```diff
@@ -1258,23 +1258,28 @@ impl<'a> Parser<'a> {
...
-                } else {
-                    let tok = self.next_token();
-                    let key = match tok.token {
-                        Token::Word(word) => word.to_ident(),
-                        _ => {
-                            return parser_err!(
-                                format!("Expected identifier, found: {tok}"),
-                                tok.location
-                            )
+                match self.try_parse_method(expr)? {
```
iffyio, line 1261: left a suggestion to avoid the nested result, thinking with that it should be possible to return the expr as is and then here we avoid the need to match: e.g.
```rust
let expr = self.try_parse_method(expr)?;
if !self.consume_token(...) {
}
```

src/ast/mod.rs
gaoqiangz: This link is what I mentioned in the topic. I'll add it to the comments.

iffyio: Ah indeed it was :+1:

iffyio (approved):
LGTM! cc @alamb

iffyio:
@gaoqiangz could you take a look at the doc building failure when you get the chance?

gaoqiangz:
@iffyio CI was fixed.

lovasoa (approved):

alamb:
Thanks again @gaoqiangz and @lovasoa and @iffyio

yoavcloud:
@gaoqiangz I've been working on a similar addition, it's great to see you beat me to it :-) One thing I added in my (unpublished) work is data-type specific methods. For example:
`SELECT geography::STGeomFromText('POLYGON((-122.358 47.653, -122.348 47.649, -122.348 47.658, -122.358 47.658, -122.358 47.653))', 4326).STAsText() FROM customers` 

How would you support the `geography` data type here? In my code I added it as a `namespace` field. I'm trying to avoid expanding ObjectName. Do you think that can work in your code too?

PR #1501, opened 2024-11-06
Add support for Snowflake SHOW DATABASES/SCHEMAS/TABLES/VIEWS/COLUMNS statements

yoavcloud:
This PR extends the existing support for SHOW statements to Snowflake


src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+            }
+        }
+        false
+    }
+
+    fn parse_show_opt_in(&mut self) -> Result<Option<ShowStatementIn>, ParserError> {
+        let clause = match self.parse_one_of_keywords(&[Keyword::FROM, Keyword::IN]) {
+            Some(Keyword::FROM) => ShowStatementInClause::FROM,
+            Some(Keyword::IN) => ShowStatementInClause::IN,
+            _ => return Ok(None),
```
iffyio, line 12378: we can return an error with e.g. `self.expected(...)` instead? since this would be a bug on our end if we hit the branch, so that the parser doesn't continue given it already consumed some tokens

src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+    /// Look for an expected keyword, without consuming it
+    fn peek_keyword(&self, expected: Keyword) -> bool {
+        match self.peek_token().token {
+            Token::Word(w) => expected == w.keyword,
+            _ => false,
+        }
+    }
+
+    /// Look for one of expected keyword, without consuming it
+    fn peek_keywords(&self, expected: &[Keyword]) -> bool {
```
iffyio, line 12365:
```suggestion
    fn peek_one_of_keywords(&self, expected: &[Keyword]) -> bool {
```
since parse_keywords might instead suggest that we expect to parse the specified keywords in sequence

src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+    }
+
+    fn parse_show_opt_in(&mut self) -> Result<Option<ShowStatementIn>, ParserError> {
+        let clause = match self.parse_one_of_keywords(&[Keyword::FROM, Keyword::IN]) {
+            Some(Keyword::FROM) => ShowStatementInClause::FROM,
+            Some(Keyword::IN) => ShowStatementInClause::IN,
+            _ => return Ok(None),
+        };
+
+        if self.parse_keyword(Keyword::DATABASE) {
```
iffyio, line 12381: the branches look identical, can we do the match once so that we reuse the same lookahead logic for all branches? thinking e.g roughly
```rust
        let (parent_type, parent_name) = match self.parse_one_of_keywords(&[Keyword::DATABASE,])? {
            Some(Keyword::DATABASE) => (Some(ShowStatementInParentType::Database), None)
           // ...
            None => { /* mysql */ }
        };
        if parent_type.is_some() && self.peek_keywords(&[Keyword::STARTS, Keyword::WITH]) {
            Ok(Some(ShowStatementIn {
                clause,
                parent_type,
                parent_name,
            }))
        } else {
            let parent_name = match self.parse_object_name(false) {
                Ok(n) => Some(n),
                Err(_) => None,
            };
            Ok(Some(ShowStatementIn {
                clause,
                parent_type: Some(ShowStatementInParentType::Database),
                parent_name,
            }))
        }
```

src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+            if self.peek_keywords(&[Keyword::STARTS, Keyword::WITH]) {
+                Ok(Some(ShowStatementIn {
+                    clause,
+                    parent_type: Some(ShowStatementInParentType::Database),
+                    parent_name: None,
+                }))
+            } else {
+                let parent_name = match self.parse_object_name(false) {
```
iffyio, line 12389:
```suggestion
                let parent_name = match self.parse_object_name(false)?;
```
we would want to return the error if there was one?

src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+
+            Ok(Some(ShowStatementIn {
+                clause,
+                parent_type: None,
+                parent_name: Some(parent_name),
+            }))
+        }
+    }
+
+    fn parse_show_opt_starts_with(&mut self) -> Result<Option<Value>, ParserError> {
```
iffyio, line 12466:
```suggestion
    fn maybe_parse_show_starts_with(&mut self) -> Result<Option<Value>, ParserError> {
```
thinking similar here and for the others, the 'opt' in between reads somewhat like its part of the type being parsed, might be easily recognizable with the maybe prefix

src/parser/mod.rs
```diff
@@ -12289,6 +12352,137 @@ impl<'a> Parser<'a> {
...
+        match self.parse_keywords(&[Keyword::STARTS, Keyword::WITH]) {
+            true => Ok(Some(self.parse_value()?)),
+            false => Ok(None),
+        }
```
iffyio, line 12470: nit: for these we could do an if/else instead of a match on a bool?

src/parser/mod.rs
yoavcloud: I don't think we can get to an unexpected state here because we will only get a Some if the parser consumed a FROM or an IN keyword. If it's neither, it means this show option does not exist in the statement and no other tokens were consumed, so we should just return the None.

src/parser/mod.rs
```diff
@@ -9685,58 +9711,95 @@ impl<'a> Parser<'a> {
...
-            _ => (None, None),
-        };
-        let filter = self.parse_show_statement_filter()?;
+        let history = !external && self.parse_keyword(Keyword::HISTORY);
+        let filter;
+        let show_in;
+        let filter_position;
+        if self.dialect.supports_show_like_before_in() {
+            filter = self.parse_show_statement_filter()?;
+            //YOAV: here we have a problem, the hint is DB-dependent (table in a schemas or some other object)
```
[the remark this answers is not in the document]
yoavcloud: Oops... fixed

yoavcloud:
@iffyio thanks for the great comments, it really helped me improve this code. New commit for you to look at.

src/parser/mod.rs
iffyio: Ah right, yeah I think in this commit both the input and the behavior of `parse_one_of_keywords` are in sync, with code changes there's a chance that could change. One could introduce a bug like updating the input keywords and forgetting to handle the branch. It would be desirable to return an error if such ever occurs. We could make the noop branch here explicit in that case e.g?
```rust
let clause = match self.parse_one_of_keywords(&[Keyword::FROM, Keyword::IN]) {
  Some(Keyword::FROM) => ShowStatementInClause::FROM,
  Some(Keyword::IN) => ShowStatementInClause::IN,
  None => return Ok(None),
  _ => return self.expected("")
}
```

src/parser/mod.rs
```diff
@@ -12315,6 +12310,124 @@ impl<'a> Parser<'a> {
...
+            Some(Keyword::DATABASE) if self.peek_keywords(&[Keyword::STARTS, Keyword::WITH]) => {
+                (Some(ShowStatementInParentType::Database), None)
+            }
+            Some(Keyword::SCHEMA) if self.peek_keywords(&[Keyword::STARTS, Keyword::WITH]) => {
+                (Some(ShowStatementInParentType::Schema), None)
+            }
+            Some(parent_kw) => {
+                let parent_name = match self.parse_object_name(false) {
+                    Ok(n) => Some(n),
+                    _ => None,
```
iffyio, line 12376: ah this one is ignoring an actual error that we would want to instead propagate or?

src/parser/mod.rs
```diff
@@ -12315,6 +12310,124 @@ impl<'a> Parser<'a> {
...
+                };
+                match parent_kw {
+                    Keyword::ACCOUNT => (Some(ShowStatementInParentType::Account), parent_name),
+                    Keyword::DATABASE => (Some(ShowStatementInParentType::Database), parent_name),
+                    Keyword::SCHEMA => (Some(ShowStatementInParentType::Schema), parent_name),
+                    Keyword::TABLE => (Some(ShowStatementInParentType::Table), parent_name),
+                    Keyword::VIEW => (Some(ShowStatementInParentType::View), parent_name),
+                    _ => unreachable!(),
```
iffyio, line 12384: I think similarly here, for unexpected states, ideally we can return an error to fail the parsing that specific input in order to avoid crashing

src/parser/mod.rs
```diff
@@ -12315,6 +12310,124 @@ impl<'a> Parser<'a> {
...
+    fn peek_keywords(&mut self, expected: &[Keyword]) -> bool {
+        let index = self.index;
+        for kw in expected {
+            if !self.parse_keyword(*kw) {
+                self.index = index;
+                return false;
+            }
+        }
+        self.index = index;
+        true
```
iffyio, line 12324:
```suggestion
        let matched = self.parse_keywords(expected);
        self.index = index;
        matched
```
thinking we could simplify this by reusing the existing parse_keywords functionality?

Also can we move this method to the same area of the code as we have the other `peek/parse_keyword*`? so that its easier to find when folks look for that type of functionality

src/parser/mod.rs
```diff
@@ -12315,6 +12310,124 @@ impl<'a> Parser<'a> {
...
+            }
+            None => {
+                // Parsing MySQL style FROM tbl_name FROM db_name
+                // which is equivalent to FROM tbl_name.db_name
+                let mut parent_name = self.parse_object_name(false)?;
+                if self
+                    .parse_one_of_keywords(&[Keyword::FROM, Keyword::IN])
+                    .is_some()
+                {
+                    parent_name.0.insert(0, self.parse_identifier(false)?);
```
iffyio, line 12395: what is the intent with the insert operation (tried to diff against the previous version but didn't see similar behavior)?

tests/sqlparser_snowflake.rs
```diff
@@ -2781,3 +2781,68 @@ fn test_parentheses_overflow() {
...
+    snowflake().verified_stmt("SHOW SCHEMAS IN DATABASE");
+    snowflake().verified_stmt("SHOW SCHEMAS IN DATABASE xyz");
+    snowflake().verified_stmt("SHOW SCHEMAS HISTORY LIKE '%xa%'");
+    snowflake().verified_stmt("SHOW SCHEMAS STARTS WITH 'abc' LIMIT 20");
+    snowflake().verified_stmt("SHOW SCHEMAS IN DATABASE STARTS WITH 'abc' LIMIT 20 FROM 'xyz'");
+}
+
+#[test]
+fn test_parse_show_tables() {
+    snowflake().verified_stmt("SHOW TABLES");
```
iffyio, line 2811: for these tests would it make sense to verify the statements using `let d =  all_dialects_where(|d| d.supports_show_like_before_in());`? so that any future dialect with the same behavior as snowflake in this regard gets automatic test coverage

src/ast/mod.rs
```diff
@@ -7357,6 +7354,99 @@ impl Display for UtilityOption {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+pub struct ShowStatementOptions {
```
iffyio, line 7358: For the added structs/enums, since they're public can we add a short description mentioning what they are? we could probably also link to one of the snowflake show docs to serve as a generic reference for someone that comes across it

iffyio (commented):
Left some comments, mostly small I think, the only one unclear to me was the mysql object_name insert behavior. I think this looks good overall, Thanks @yoavcloud!

src/parser/mod.rs
yoavcloud: Added another condition above to ensure we can expect an object name here

src/parser/mod.rs
yoavcloud: See here for the previous code: https://github.com/apache/datafusion-sqlparser-rs/blob/334a5bf354ac964a14f2e2dc3851ed4d853b28d7/src/parser/mod.rs#L9683

It basically transforms the SHOW COLUMNS FROM tbl1 FROM db1 to the canonical form SHOW COLUMNS FROM db1.tbl1

src/parser/mod.rs
yoavcloud: @iffyio actually, the issue here is that the optional parent name is located at the end of the statement. We need to differentiate between SHOW TABLES IN DATABASE and SHOW TABLES IN DATABASE db1. The failure to parse the object name is the indication.

Do you have a suggestion as to how to handle this case better?

tests/sqlparser_snowflake.rs
yoavcloud: Please see test_show_dbs_schemas_tables_views in sqlparser_common. Perhaps I should remove the non-Snowflake specific tests from this file.

src/parser/mod.rs
iffyio: Ah if its optional then we could use maybe_parse in that case?
```rust
let parent_name = self.maybe_parse(|p| p.parse_object_name(false))?;
```

src/parser/mod.rs
iffyio: Ah I see, that's interesting

tests/sqlparser_snowflake.rs
iffyio: Ah I see, yeah having them here would've meant those tests have to be snowflake agnostic to some degree which wouldn't be ideal either. I think moving some of them to common as is done currently is reasonable thanks!

src/ast/mod.rs
```diff
@@ -7357,6 +7354,102 @@ impl Display for UtilityOption {
     }
 }
 
+/// Represents the different options available for a SHOW <OBJECT>
+/// statement to filter the results. Example from Snowflake:
+/// https://docs.snowflake.com/en/sql-reference/sql/show-tables
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+pub struct ShowStatementOptions {
```
iffyio, line 7361: Oh one thing actually, we might be missing the serde and visit derive attributes on some of the new structs?

iffyio (approved):
LGTM! cc @alamb

alamb:
🚀

alamb:
Thanks again @iffyio  and @yoavcloud

Issue #1506, opened 2024-11-08
support for mssql `json_object` syntax 


lovasoa:
SQL server has a json_object function with a special syntax that uses colons instead of commas between json keys and values.

```sql
SELECT JSON_OBJECT('name':'value', 'type':1)
```

This is not yet supported in sqlparser. This would be useful in [SQLPage](https://sql-page.com) where we use json for advanced features.

Reference: https://learn.microsoft.com/en-us/sql/t-sql/functions/json-object-transact-sql?view=sql-server-ver16

@gaoqiangz this may be of interest to you :)

gaoqiangz:
👌

lovasoa:
Fixed by @gaoqiangz . Thank you !

PR #1507, opened 2024-11-09
Add support for MSSQL's `JSON_ARRAY`/`JSON_OBJECT` expr

gaoqiangz:
Solves #1506


lovasoa (approved):

src/ast/mod.rs
```diff
@@ -7317,6 +7367,87 @@ impl Display for UtilityOption {
...
+            write!(f, "{null_clause}")?;
+        }
+        write!(f, ")")
+    }
+}
+
+/// MSSQL's `JSON_OBJECT` constructor function
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
```
iffyio, line 7395: Ah I think piggy-backing on my comment on the JsonArray, we would probably support this syntax automatically  by extending the function_arguments_list representation? Did you see the existing [json functionality](https://github.com/apache/datafusion-sqlparser-rs/blob/a4c73f3c6b221d4db51c8126330e177c5d9bd820/src/parser/mod.rs#L2951), I would expect the parser currently supports `foo:bar` as an expression today iirc that's snowflake json syntax, which is identical to this and feels like should be supported already?

src/ast/mod.rs
```diff
@@ -931,6 +931,54 @@ pub enum Expr {
...
+    /// Example:
+    ///
+    /// ```sql
+    /// SELECT JSON_ARRAY('a', 1, 'b', 2) --["a",1,"b",2]
+    /// SELECT JSON_ARRAY('a', 1, NULL, 2 NULL ON NULL) --["a",1,null,2]
+    /// SELECT JSON_ARRAY('a', JSON_OBJECT('name':'value', 'type':1), JSON_ARRAY(1, null, 2 NULL ON NULL)) --["a",{"name":"value","type":1},[1,null,2]]
+    /// ```
+    ///
+    /// Reference: <https://learn.microsoft.com/en-us/sql/t-sql/functions/json-array-transact-sql?view=sql-server-ver16>
+    JsonArray(JsonArray),
```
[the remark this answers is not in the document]
gaoqiangz: I'll try it tomorrow.

gaoqiangz:
I modified existing `Function` parsing,including a break change,because `JSON_OBJECT` support `FUNC(<expr> : <expr>)` syntax:

```sql
DECLARE @id_key nvarchar(10) = N'id',@id_value nvarchar(64) = NEWID();
SELECT JSON_OBJECT(LOWER('USER_NAME') : USER_NAME(), LOWER(@id_key) : @id_value, 'sid':(SELECT @@SPID));
```

Break change:

```rs
pub enum FunctionArg {
    Named {
        name: Expr, // <<
        arg: FunctionArgExpr,
        operator: FunctionArgOperator,
    },
    Unnamed(FunctionArgExpr),
}
```

Please take a look!

src/parser/mod.rs
```diff
@@ -11096,6 +11096,19 @@ impl<'a> Parser<'a> {
...
+            if self.consume_token(&Token::Colon) {
+                let arg = self.parse_expr()?.into();
+                Ok(FunctionArg::Named {
+                    name,
+                    arg,
+                    operator: FunctionArgOperator::Colon,
+                })
+            } else {
+                Ok(FunctionArg::Unnamed(name.into()))
+            }
```
iffyio, line 11111:
```suggestion
        } else if self.dialect.supports_json_object_functions() && self.peek_nth_token(1) == Token::Colon {
            let name = self.parse_wildcard_expr(false)?;
            let _ = self.expect_token(&Token::Colon)?;
            let arg = self.parse_expr()?.into();
            Ok(FunctionArg::Named {
                    name,
                    arg,
                    operator: FunctionArgOperator::Colon,
                })
            }
```
Something like this so that we only enter the branch when we know that there is a valid expr to parse? We can use a method on the dialect for this feature instead of the `dialect_of!` macro (the latter is essentially deprecated)

tests/sqlparser_mssql.rs
```diff
@@ -448,6 +448,447 @@ fn parse_for_json_expect_ast() {
     );
 }
 
+#[test]
+fn parse_mssql_json_object() {
```
iffyio, line 452: noticed we used the parse_wildcard_expr, if that syntax is valid, can we include a test case with a wildcard argument?

src/parser/mod.rs
```diff
@@ -11141,19 +11154,34 @@ impl<'a> Parser<'a> {
...
+        let mut clauses = vec![];
+
+        if dialect_of!(self is MsSqlDialect) {
+            // JSON_ARRAY(NULL ON NULL)
+            if self.parse_keywords(&[Keyword::NULL, Keyword::ON, Keyword::NULL]) {
+                clauses.push(FunctionArgumentClause::JsonNullClause(
+                    JsonNullClause::NullOnNull,
+                ));
+            }
+            // JSON_ARRAY(ABSENT ON NULL)
```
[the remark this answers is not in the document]
iffyio: Ah noticed we have this code few lines below as well, so that the above comment applies to that one as well. Actually for this line, I'm not sure we need the changes here at all, since if we've hit `RParen` we don't have any clauses to parse (so that the one below I imagine should be sufficient)?

src/parser/mod.rs
```diff
@@ -11141,19 +11154,34 @@ impl<'a> Parser<'a> {
     /// FIRST_VALUE(x IGNORE NULL);
     /// ```
     fn parse_function_argument_list(&mut self) -> Result<FunctionArgumentList, ParserError> {
+        let mut clauses = vec![];
+
+        if dialect_of!(self is MsSqlDialect) {
```
iffyio, line 11159: similarly, we can use the `self.dialect.supports...()` method  to gate the behavior

src/parser/mod.rs
```diff
@@ -11096,6 +11096,19 @@ impl<'a> Parser<'a> {
...
+        } else if dialect_of!(self is MsSqlDialect) {
+            // FUNC(<expr> : <expr>)
+            let name = self.parse_wildcard_expr()?;
+            if self.consume_token(&Token::Colon) {
+                let arg = self.parse_expr()?.into();
+                Ok(FunctionArg::Named {
+                    name,
+                    arg,
+                    operator: FunctionArgOperator::Colon,
+                })
```
iffyio, line 11108: hmm yeah regarding the requirement I wondered a bit of pros/cons but I figure it might be the lesser evil to add a new variant like `FunctionArg::Association { target: Expr, arg, operator }` variant that is able to represent an arbitrary expression? 

The immediate advantage would be that its not a breaking change but we would also avoid hiding the identifier which is what users expect from a named function argument since that is a feature used in a lot of scenarios. So I figure we could be more explicit to represent this as a different class of functions, and then highlight in the doc the difference between the two

iffyio (commented):
Thanks @gaoqiangz I think this is looking good! Left some comments

src/parser/mod.rs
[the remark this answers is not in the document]
gaoqiangz: The first few lines of  `parse_keywords` are for `JSON_ARRAY(<json-null-clause>)`/``JSON_OBJECT(<json-null-clause>)`` which does not have any arguments only includes `json-null-clause` because T-SQL supports it.

playground:
<https://sqliteonline.com/#sqltext=%23url-sqlite%3Ddb-mssql%0D%0A%23tab-name%3DMS%20SQL%0D%0ASELECT%20JSON_ARRAY(NULL%20ON%20NULL)%2CJSON_OBJECT(ABSENT%20ON%20NULL)>

src/parser/mod.rs
gaoqiangz: `self.peek_nth_token(1)` cannot capture the case where the name is an expression.

iffyio: Ah that right! that makes sense, though can we collapse the last else if /else so that the unnamed default is constructed once?
```rust
} else  {
  let name = ...;
  if dialect_of!(self is MsSqlDialect) {
    return Ok(FunctionArg::Named{...})
  }
  Ok(FunctionArg::Unnamed(name.into()))
}
```

src/parser/mod.rs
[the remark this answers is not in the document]
iffyio: hmm yeah that was my understanding, but not sure I follow that the change is needed to handle that scenario, in that neither of the [other clauses](https://github.com/apache/datafusion-sqlparser-rs/blob/43b0652bb1469e253ef3e1d9e606f43c30dba845/src/parser/mod.rs#L11182-L11224) should kick in or?

src/parser/mod.rs
gaoqiangz: It might be better to add a new variant, but the naming is a headache.

gaoqiangz: How about this?

```rs
pub enum FunctionArg {
    Named {
        name: Ident,
        arg: FunctionArgExpr,
        operator: FunctionArgOperator,
    },
    ExprNamed {
        name: Expr, // <<
        arg: FunctionArgExpr,
        operator: FunctionArgOperator,
    },
    Unnamed(FunctionArgExpr),
}

pub trait Dialect {
  /// Some dialect supports `function(<arbitrary-expr>:<arbitrary-expr>,...)`,
  /// for this scenario, when returning `true` use the `FunctionArg::ExprNamed` variant in all cases to express named arguments,otherwise use the `FunctionArg::Named` variant (compatible reason).
  fn supports_function_expr_named { false }
}
```

src/parser/mod.rs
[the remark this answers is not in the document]
gaoqiangz: Maybe this is a useful scenario?

```sql
SELECT JSON_OBJECT($json_param ABSENT ON NULL);
```

`$json_param` can be replaced to empty by user input.

tests/sqlparser_mssql.rs
gaoqiangz: I used `parse_expr` instead.

gaoqiangz:
I refactored function parameter list parsing,add some `supports_named_fn_args_with_xxx` functions to the Dialect trait,please take a look!

src/dialect/mod.rs
```diff
@@ -231,11 +231,32 @@ pub trait Dialect: Debug + Any {
...
+    }
+
+    /// Returns true if the dialect supports named arguments of the form `FUN(a := '1', b := '2')`.
+    fn supports_named_fn_args_with_assignment_operator(&self) -> bool {
+        false
+    }
+
+    /// Returns true if the dialect supports named arguments of the form `FUN(a => '1', b => '2')`.
+    fn supports_named_fn_args_with_rarrow_operator(&self) -> bool {
+        true //for compatible
```
iffyio, line 251:
```suggestion
        true
```

src/dialect/mod.rs
```diff
@@ -231,11 +231,32 @@ pub trait Dialect: Debug + Any {
...
+    }
+
+    /// Returns true if dialect supports argument name as arbitrary expression (`FUN(<arbitrary-expr>:<arbitrary-expr>,...)`),and will use the `FunctionArg::ExprNamed` variant,
```
iffyio, line 254:
```suggestion
    /// Returns true if dialect supports argument name as arbitrary expression.
    /// e.g. `FUN(LOWER('a'):'1',  b:'2')`
    /// Such function arguments are represented in the AST by the `FunctionArg::ExprNamed` variant,
```

src/parser/mod.rs
```diff
@@ -11384,6 +11406,23 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parses MSSQL's json-null-clause
+    fn parse_json_null_clause(&mut self) -> Option<JsonNullClause> {
+        if !dialect_of!(self is MsSqlDialect) {
+            return None;
+        }
```
iffyio, line 11413: here we can use a dialect method as well?

src/parser/mod.rs
```diff
@@ -11384,6 +11406,23 @@ impl<'a> Parser<'a> {
...
+        if let Some(kw) = self.parse_one_of_keywords(&[Keyword::ABSENT, Keyword::NULL]) {
+            if self.parse_keywords(&[Keyword::ON, Keyword::NULL]) {
+                if kw == Keyword::ABSENT {
+                    return Some(JsonNullClause::AbsentOnNull);
+                } else if kw == Keyword::NULL {
+                    return Some(JsonNullClause::NullOnNull);
+                }
+            }
+        }
```
iffyio, line 11422: could we go back to the previous version of this logic? I think that was more readable since we avoided conditional state, and it read the same as the syntax. (e.g its no longer super clear what's happening if we match `ABSENT` but not the `ON NULL` part)
```rust
if self.parse_keywords(absent, on, null) {
   Some()
else if self.parse_keywords(null, on, null) {
   Some()
} else {
    None
}
```

src/parser/mod.rs
gaoqiangz: Maybe we can remove this predication?

iffyio: Yeah I think we can do that, since the syntax is unambiguous enough

iffyio (approved):
LGTM! cc @alamb

alamb:
Thanks again @gaoqiangz and @iffyio

lovasoa:
I opened a new PR to enable this syntax in postgres, which supports it too: https://github.com/apache/datafusion-sqlparser-rs/pull/1546

https://www.postgresql.org/docs/current/functions-json.html

Issue #1496, opened 2024-11-01
Expression prefix parsed instead of identifier


yoavcloud:
When a statement uses identifiers that match certain expression prefix keywords, the parser will try to parse the expression over parsing the token as an identifier. The following statements work on Snowflake but the parser errors on the SELECT. If a different identifier name is selected, for example, `interval1` then the parser works as expected.

```sql
create or replace temporary table test(interval int);
insert into test (interval) values (1);
insert into test (interval) values (2);
select max(interval) from test;
```

yoavcloud:
PR to fix this: https://github.com/apache/datafusion-sqlparser-rs/pull/1513

yoavcloud:
Fixed in #1513

PR #1513, opened 2024-11-11
Fallback to identifier parsing if expression parsing fails

yoavcloud:
This fixes the following issue: https://github.com/apache/datafusion-sqlparser-rs/issues/1496

The parser encounters situations when the next keyword indicates an expression, but in fact it should be parsed as an identifier. Example from Snowflake: `SELECT MAX(interval) from tbl`. When the parser encounters the `interval` word it tries to parse an `Expr::Interval` but it fails because in the context of this query, `interval` is an identifier that Snowflake (unlike most other dialects) allows in unquoted form.

The suggested approach is to try to parse the expression, but if that fails, fallback to parse an identifier under certain conditions. In addition, each dialect can now declare which keywords it reserves for use as an identifier in unquoted form.

Lastly, changed the parsing of the `DEFAULT` word to Expr::Default instead of an identifier.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -695,6 +695,8 @@ pub enum Expr {
         // https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
         format: Option<CastFormat>,
     },
+    /// `DEFAULT` value of a column e.g. INSERT INTO tbl (a, b) VALUES ('foo', DEFAULT)
+    Default,
     /// AT a timestamp to a different timezone e.g. `FROM_UNIXTIME(0) AT TIME ZONE 'UTC-06:00'`
     AtTimeZone {
         timestamp: Box<Expr>,
@@ -1449,6 +1451,7 @@ impl fmt::Display for Expr {
                     write!(f, "{expr}::{data_type}")
                 }
             },
+            Expr::Default => write!(f, "DEFAULT"),
             Expr::Extract {
                 field,
                 syntax,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -695,8 +695,6 @@ pub enum Expr {
         // https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
         format: Option<CastFormat>,
     },
-    /// `DEFAULT` value of a column e.g. INSERT INTO tbl (a, b) VALUES ('foo', DEFAULT)
-    Default,
     /// AT a timestamp to a different timezone e.g. `FROM_UNIXTIME(0) AT TIME ZONE 'UTC-06:00'`
     AtTimeZone {
         timestamp: Box<Expr>,
@@ -1451,7 +1449,6 @@ impl fmt::Display for Expr {
                     write!(f, "{expr}::{data_type}")
                 }
             },
-            Expr::Default => write!(f, "DEFAULT"),
             Expr::Extract {
                 field,
                 syntax,
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -38,6 +38,8 @@ use alloc::vec::Vec;
 #[cfg(not(feature = "std"))]
 use alloc::{format, vec};
 
+use super::keywords::RESERVED_FOR_IDENTIFIER;
+
 /// A [`Dialect`] for [Snowflake](https://www.snowflake.com/)
 #[derive(Debug, Default)]
 pub struct SnowflakeDialect;
@@ -213,6 +215,16 @@ impl Dialect for SnowflakeDialect {
     /// for example: <https://docs.snowflake.com/en/sql-reference/sql/show-views#syntax>
     fn supports_show_like_before_in(&self) -> bool {
         true
+    }
+
+    fn is_reserved_for_identifier(&self, kw: Keyword) -> bool {
+        // Unreserve some keywords that Snowflake accepts as identifiers
+        // See: https://docs.snowflake.com/en/sql-reference/reserved-keywords
+        if matches!(kw, Keyword::INTERVAL) {
+            false
+        } else {
+            RESERVED_FOR_IDENTIFIER.contains(&kw)
+        }
     }
 }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -45,6 +45,9 @@ pub enum ParserError {
     TokenizerError(String),
     ParserError(String),
     RecursionLimitExceeded,
+    /// Error indicating that the parsing branch taken
+    /// did not yield a meaningful result
+    BranchAbandoned,
 }
 
 // avoid clippy type_complexity warnings
@@ -174,6 +177,7 @@ impl fmt::Display for ParserError {
                 ParserError::TokenizerError(s) => s,
                 ParserError::ParserError(s) => s,
                 ParserError::RecursionLimitExceeded => "recursion limit exceeded",
+                ParserError::BranchAbandoned => "branch abandoned",
             }
         )
     }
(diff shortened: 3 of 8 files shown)


src/parser/mod.rs
```diff
@@ -1013,175 +1189,22 @@ impl<'a> Parser<'a> {
...
-                                    id_parts.push(Ident::with_quote('\'', s))
-                                }
-                                _ => {
-                                    return self
-                                        .expected("an identifier or a '*' after '.'", next_token);
-                                }
-                            }
+            // We first try to parse the word as the prefix of an expression.
+            // For example, the word INTERVAL in: SELECT INTERVAL '7' DAY
+            Token::Word(w) => match self.try_parse(|parser| parser.parse_expr_by_keyword(&w)) {
```
iffyio, line 1194: Is it possible to use `self.maybe_parse` instead for the optional expr parsing?

src/parser/mod.rs
```diff
@@ -3574,6 +3597,24 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Run a parser method `f`, reverting back to the current position if unsuccessful
+    /// but retaining the error message if such was raised by `f`
+    pub fn try_parse<T, F>(&mut self, mut f: F) -> Result<T, ParserError>
```
iffyio, line 3602: Ah I left a similar comment earlier, I'm suspecting I'm missing something, not sure I see the need for a custom try_parse vs maybe_parse.

src/ast/mod.rs
```diff
@@ -695,6 +695,8 @@ pub enum Expr {
         // https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
         format: Option<CastFormat>,
     },
+    /// `DEFAULT` value of a column e.g. INSERT INTO tbl (a, b) VALUES ('foo', DEFAULT)
+    Default,
```
iffyio, line 699: is default a reserved keyword in all dialects/sql-standard? wondering if its not this might break behavior for some folks

src/parser/mod.rs
```diff
@@ -965,6 +969,178 @@ impl<'a> Parser<'a> {
...
+            }
+            Keyword::MAP if self.peek_token() == Token::LBrace && self.dialect.support_map_literal_syntax() => {
+                self.parse_duckdb_map_literal()
+            }
+            Keyword::DEFAULT => Ok(Expr::Default),
+            _ => Err(ParserError::BranchAbandoned)
+        }
+    }
+
+    fn parse_ident_expr(&mut self, w: &Word) -> Result<Expr, ParserError> {
```
iffyio, line 1076: hmm thinking we would need to rename the function, it seems to be doing more than identifiers (and based on the name its unclear what the word `w` input argument is for)

tests/sqlparser_common.rs
```diff
@@ -5071,7 +5071,9 @@ fn parse_interval_dont_require_unit() {
 
 #[test]
 fn parse_interval_require_unit() {
-    let dialects = all_dialects_where(|d| d.require_interval_qualifier());
+    let dialects = all_dialects_where(|d| {
+        d.require_interval_qualifier() && d.is_reserved_for_identifier(Keyword::INTERVAL)
```
iffyio, line 5075: oh I would have expected that these the reserved condition is always true given the require_interval condition is true? in that, would it be correct for a dialect to not have INTERVAL as a keyword yet `d.require_interval_qualifier()` returns true for that dialect

src/parser/mod.rs
```diff
@@ -1013,175 +1189,22 @@ impl<'a> Parser<'a> {
...
+            // We first try to parse the word as the prefix of an expression.
+            // For example, the word INTERVAL in: SELECT INTERVAL '7' DAY
+            Token::Word(w) => match self.try_parse(|parser| parser.parse_expr_by_keyword(&w)) {
+                Ok(expr) => Ok(expr),
+                // Word does not indicate the start of a complex expression, try to parse as identifier
+                Err(ParserError::BranchAbandoned) => Ok(self.parse_ident_expr(&w)?),
+                // Word indicates the start of a complex expression, try to parse as identifier if the
+                // dialect does not reserve it, otherwise return the original error
+                Err(e) => {
+                    if !self.dialect.is_reserved_for_identifier(w.keyword) {
```
iffyio, line 1201: one thing I'm wondering regarding the original problem is whether the underlying issue is rather the keywords like interval not being properly gated to only the dialects that support them?
I suspect there are cases that could benefit from not having the parser mistaking keywords on dialects that don't support them though (like in this `self.dialect.is_reserved_for_identifier`) - but not super sure if there are cases in this scenario with exprs here that strictly require it

src/parser/mod.rs
yoavcloud: See above comment

src/parser/mod.rs
yoavcloud: I think the problem lies in dialects that interpret keywords differently in different contexts (I guess by looking ahead). For example, in Snowflake, the INTERVAL word can be a keyword for an INTERVAL expr or an ident:
```
create or replace temporary table test(interval int);
insert into test (interval) values (1), (2);
select current_date() + interval '1 day', max(interval) from test;
```

I see four possible cases for the parser to handle:
1. The word is part of a syntactically correct expression, e.g. `INTERVAL '1 day'` --> handled
2. The word is part of a syntactically incorrect expression, e.g. `INTERVAL '1 dai'` --> handled
3. The word is an identifier and the dialect does not reserve it, e.g. `SELECT MAX(interval) FROM tbl` --> not handled
4. The word is an identifier and the dialect reserves it, e.g. `SELECT MAX(from) FROM test;` --> not handled perfectly, but we are OK with the parser to be more accepting, as long as over time we give maintainers the tools to add reserved keywords as needed.

iffyio: ah got it! I was under the impression the issue was snowflake not supporting interval as a keyword, thanks for clarifying

src/parser/mod.rs
iffyio: Ah right, can we use a enum to represent the states returned by `parse_expr_by_keyword` instead? Im thinking that would be more appropriate here, we wouldnt want to flag this as an error since that's a public api, and it'll likely let us use maybe_parse as well so that we avoid the try_parse introduction

src/ast/mod.rs
yoavcloud: You are right, I will leave this one to a future PR

src/parser/mod.rs
yoavcloud: I ended up going with a Result<Option<Expr>> see next commit

src/parser/mod.rs
```diff
@@ -1057,176 +1231,33 @@ impl<'a> Parser<'a> {
...
+                // We first try to parse the word as the prefix of an expression.
+                // For example, the word INTERVAL in: SELECT INTERVAL '7' DAY
+                match self.parse_expr_prefix_by_reserved_word(&w) {
+                    // No expression prefix associated with this word
+                    Ok(None) => Ok(self.parse_expr_prefix_by_nonreserved_word(&w)?),
+                    // This word indicated an expression prefix and parsing was successful
+                    Ok(Some(expr)) => Ok(expr),
+                    // This word indicated an expression prefix but parsing failed. Two options:
+                    // 1. Malformed statement
+                    // 2. The dialect may allow this word as identifier as well as indicating an expression
```
iffyio, line 1246: The comments are indeed helpful to follow thanks! Thinking here we could clarify 2. even further with the `SELECT MAX(interval) from tbl` example?

src/parser/mod.rs
```diff
@@ -1057,176 +1231,33 @@ impl<'a> Parser<'a> {
...
+                    Err(e) => {
+                        let index_after_error = self.index;
+                        if !self.dialect.is_reserved_for_identifier(w.keyword) {
+                            // Rollback before trying to parse using a different approach
+                            self.index = index_before;
```
iffyio, line 1251: I'm thinking ideally we're somehow reusing maybe_parse for the logic, feels like we're duplicating that behavior for this custom use case with the manual stashing of indexes which makes it a bit harder to track the logic's correctness

yoavcloud: Look at the solution in this commit, breaking maybe_parse to two functions, one maintaining the current contract and the second allowing for the functionality I need.

src/parser/mod.rs
```diff
@@ -1057,176 +1234,258 @@ impl<'a> Parser<'a> {
...
-                                        .expected("an identifier or a '*' after '.'", next_token);
-                                }
+                    // This word indicated an expression prefix and parsing was successful
+                    Ok(Some(expr)) => Ok(expr),
+
+                    // No expression prefix associated with this word
+                    Ok(None) => Ok(self.parse_expr_prefix_by_unnreserved_word(&w)?),
+
+                    // If parsing of the word as a special expression failed, we are facing two options:
+                    // 1. The statement is malformed, e.g. `SELECT INTERVAL '1 DAI`
```
iffyio, line 1256:
```suggestion
                    // 1. The statement is malformed, e.g. `SELECT INTERVAL '1 DAY`
```

src/parser/mod.rs
```diff
@@ -1057,176 +1234,258 @@ impl<'a> Parser<'a> {
...
+                expr: Box::new(expr),
+                collation: self.parse_object_name(false)?,
+            })
+        } else {
+            Ok(expr)
+        }
+    }
+
+    /// Parse an expression prefix.
+    pub fn parse_prefix2(&mut self) -> Result<Expr, ParserError> {
```
iffyio, line 1404: Ah this version seems unused we can remove?

src/parser/mod.rs
```diff
@@ -1009,6 +1009,183 @@ impl<'a> Parser<'a> {
...
+            }
+            Keyword::MAP if self.peek_token() == Token::LBrace && self.dialect.support_map_literal_syntax() => {
+                Ok(Some(self.parse_duckdb_map_literal()?))
+            }
+            _ => Ok(None)
+        }
+    }
+
+    // Tries to parse an expression by a word that is not known to have a special meaning in the dialect.
+    fn parse_expr_prefix_by_unnreserved_word(&mut self, w: &Word) -> Result<Expr, ParserError> {
```
iffyio, line 1121:
```suggestion
    fn parse_expr_prefix_by_unreserved_word(&mut self, w: &Word) -> Result<Expr, ParserError> {
```

src/parser/mod.rs
```diff
@@ -1057,176 +1234,258 @@ impl<'a> Parser<'a> {
...
+
+                    // If parsing of the word as a special expression failed, we are facing two options:
+                    // 1. The statement is malformed, e.g. `SELECT INTERVAL '1 DAI`
+                    // 2. The word is used as an identifier, e.g. `SELECT MAX(interval) FROM tbl`
+                    // We first try to parse the word as an identifier and if that fails
+                    // we rollback and return the parsing error we got from trying to parse a
+                    // special expression (to maintain backwards compatibility of parsing errors).
+                    Err(e) => {
+                        if !self.dialect.is_reserved_for_identifier(w.keyword) {
+                            if let Ok(expr) = self.maybe_parse_internal(|parser| {
```
iffyio, line 1263:
```suggestion
                            if let Some(expr) = self.maybe_parse(|parser| {
```
it looks like we can use the normal maybe_parse here since it doesn't have the special requirement?

src/parser/mod.rs
```diff
@@ -3641,18 +3900,30 @@ impl<'a> Parser<'a> {
...
             Ok(t) => Ok(Some(t)),
-            // Unwind stack if limit exceeded
             Err(ParserError::RecursionLimitExceeded) => Err(ParserError::RecursionLimitExceeded),
-            Err(_) => {
+            _ => Ok(None),
+        }
+    }
+
+    /// Run a parser method `f`, reverting back to the current position if unsuccessful.
+    pub fn maybe_parse_internal<T, F>(&mut self, mut f: F) -> Result<T, ParserError>
```
iffyio, line 3916:
```suggestion
    pub fn try_parse<T, F>(&mut self, mut f: F) -> Result<T, ParserError>
```
We can probably call try_parse which could hint that it returns an error (vs maybe_parse which hints at an option)

src/dialect/snowflake.rs
```diff
@@ -214,6 +216,16 @@ impl Dialect for SnowflakeDialect {
     fn supports_show_like_before_in(&self) -> bool {
         true
     }
+
+    fn is_reserved_for_identifier(&self, kw: Keyword) -> bool {
```
iffyio, line 220: oh are we missing tests for the new behavior (I can't seem to find them in the current PR if so)?

iffyio (commented):
The changes look good to me overall! Left some minor comments then one regarding tests to demonstrate the new behavior

src/parser/mod.rs
yoavcloud: This is intentional, DAI is what makes this statement malformed :-)

src/parser/mod.rs
yoavcloud: Shit... Yes, on it.

iffyio (approved):
LGTM! cc @alamb

alamb:
This PR appears to have a small conflict

yoavcloud:
@alamb resolved

alamb:
Thanks @yoavcloud

PR #1521, opened 2024-11-13
recursive select calls are parsed with bad trailing_commas parameter

tomershaniii:
The parse_projection function modifies the parsers 'options' flag,  this creates different behavior when nesting select calls i.e. select(select()) which can manifest in failure to parse (see unit test)

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3532,16 +3532,11 @@ impl<'a> Parser<'a> {
         // e.g. `SELECT 1, 2, FROM t`
         // https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#trailing_commas
         // https://docs.snowflake.com/en/release-notes/2024/8_11#select-supports-trailing-commas
-        //
-        // This pattern could be captured better with RAII type semantics, but it's quite a bit of
-        // code to add for just one case, so we'll just do it manually here.
-        let old_value = self.options.trailing_commas;
-        self.options.trailing_commas |= self.dialect.supports_projection_trailing_commas();
 
-        let ret = self.parse_comma_separated(|p| p.parse_select_item());
-        self.options.trailing_commas = old_value;
+        let trailing_commas =
+            self.options.trailing_commas | self.dialect.supports_projection_trailing_commas();
 
-        ret
+        self.parse_comma_separated_with_trailing_commas(|p| p.parse_select_item(), trailing_commas)
     }
 
     pub fn parse_actions_list(&mut self) -> Result<Vec<ParsedAction>, ParserError> {
@@ -3568,11 +3563,12 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse the comma of a comma-separated syntax element.
+    /// Allows for control over trailing commas
     /// Returns true if there is a next element
-    fn is_parse_comma_separated_end(&mut self) -> bool {
+    fn is_parse_comma_separated_end_with_trailing_commas(&mut self, trailing_commas: bool) -> bool {
         if !self.consume_token(&Token::Comma) {
             true
-        } else if self.options.trailing_commas {
+        } else if trailing_commas {
             let token = self.peek_token().token;
             match token {
                 Token::Word(ref kw)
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -3538,15 +3538,35 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse the comma of a comma-separated syntax element.
+    /// Returns true if there is a next element
+    fn is_parse_comma_separated_end(&mut self) -> bool {
+        //default implementation
```
iffyio, line 3544:
```suggestion
```

src/parser/mod.rs
```diff
@@ -3516,11 +3515,12 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse the comma of a comma-separated syntax element.
+    /// Allows for control over trailing commas
     /// Returns true if there is a next element
-    fn is_parse_comma_separated_end(&mut self) -> bool {
+    fn is_parse_comma_separated_end_with_trailing(&mut self, trailing_commas: bool) -> bool {
```
iffyio, line 3520:
```suggestion
    fn is_parse_comma_separated_end_with_trailing_commas(&mut self, trailing_commas: bool) -> bool {
```

src/parser/mod.rs
```diff
@@ -3538,15 +3538,35 @@ impl<'a> Parser<'a> {
...
+    pub fn parse_comma_separated<T, F>(&mut self, f: F) -> Result<Vec<T>, ParserError>
+    where
+        F: FnMut(&mut Parser<'a>) -> Result<T, ParserError>,
+    {
+        self.parse_comma_separated_with_trailing(f, self.options.trailing_commas)
+    }
+
+    /// Parse a comma-separated list of 1+ items accepted by `F`
+    /// Allows for control over trailing commas
+    pub fn parse_comma_separated_with_trailing<T, F>(
```
iffyio, line 3558:
```suggestion
    fn parse_comma_separated_with_trailing_commas<T, F>(
```

tests/sqlparser_snowflake.rs
```diff
@@ -2783,6 +2783,14 @@ fn test_parentheses_overflow() {
 }
 
 #[test]
+
+fn test_nested_select_with_lateral_flatten() {
```
iffyio, line 2787: Since the functionality doesnt seem to be related to the lateral flatten, can we rename the test to e.g. `test_projection_with_nested_trailing_commas`? And we can include a few more cases (e.g. doubly nested, multiple set expressions that have trailing commas)?

tests/sqlparser_snowflake.rs
```diff
@@ -2783,6 +2783,14 @@ fn test_parentheses_overflow() {
...
+    let sql = "SELECT a FROM b, LATERAL FLATTEN(input => events)";
+    let _ = snowflake().parse_sql_statements(&sql).unwrap();
+
+    let sql = "SELECT (SELECT a FROM b, LATERAL FLATTEN(input => events))";
+    let _ = snowflake().parse_sql_statements(&sql).unwrap();
```
iffyio, line 2792: hmm, the test function looks incomplete, missing a closing `}`? unless github is masking that line somehow :thinking:

src/parser/mod.rs
```diff
@@ -3481,13 +3481,12 @@ impl<'a> Parser<'a> {
         // https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#trailing_commas
         // https://docs.snowflake.com/en/release-notes/2024/8_11#select-supports-trailing-commas
         //
```
iffyio, line 3483:
```suggestion
```

iffyio (commented):
Thanks @tomershaniii! Left some comments, I think the changes look reasonable overall

tests/sqlparser_snowflake.rs
tomershaniii: don't see anything unusual and test are passing, guess this is some github artifact.

tests/sqlparser_snowflake.rs
tomershaniii: can you elaborate on "multiple set expressions that have trailing commas" ?

tomershaniii:
Thanks for the review @iffyio, changes committed per your suggestions

tests/sqlparser_snowflake.rs
iffyio: Ah I was thinking something like this where the `trailing_commas` functionality is in use with the `supports_projection_trailing_commas` flag
```rust
SELECT a, b, FROM c, (SELECT d, e, FROM f, LATERAL FLATTEN(input => events))
```

tests/sqlparser_snowflake.rs
```diff
@@ -2846,3 +2846,17 @@ fn test_parse_show_columns_sql() {
...
+    let sql = "SELECT a FROM b, LATERAL FLATTEN(input => events)";
+    let _ = snowflake().parse_sql_statements(&sql).unwrap();
+
+    //Single nesting
+    let sql = format!("SELECT ({sql})");
+    let _ = snowflake().parse_sql_statements(&sql).unwrap();
+
+    //Double nesting
+    let sql = format!("SELECT ({sql})");
+    let _ = snowflake().parse_sql_statements(&sql).unwrap();
```
iffyio, line 2861: Can we instead write out the sql explicitly? I think that makes it easier to figure out what is being tested, and easier to change if needed going forward

tomershaniii: 👍

test.sql
```diff
@@ -0,0 +1,229 @@
+    WITH EventData AS (
```
iffyio, line 1: oh I think this and the other files were mistakenly checked in?

tomershaniii: Fixed, sorry about that

iffyio (approved):
LGTM! cc @alamb

src/parser/mod.rs
```diff
@@ -3532,16 +3532,11 @@ impl<'a> Parser<'a> {
...
         // https://docs.snowflake.com/en/release-notes/2024/8_11#select-supports-trailing-commas
-        //
-        // This pattern could be captured better with RAII type semantics, but it's quite a bit of
-        // code to add for just one case, so we'll just do it manually here.
-        let old_value = self.options.trailing_commas;
-        self.options.trailing_commas |= self.dialect.supports_projection_trailing_commas();
 
-        let ret = self.parse_comma_separated(|p| p.parse_select_item());
-        self.options.trailing_commas = old_value;
+        let trailing_commas =
```
alamb, line 3536: this is a nice improvement

alamb (approved):
Thank you @tomershaniii and @iffyio

Issue #984, opened 2023-09-26
Stack overflow in `Statement::to_string` for deeply nested expresions


jjbayer:
It seems that `Statement::to_string` overflows for deep expressions even if the default [recursion limit](https://github.com/sqlparser-rs/sqlparser-rs/pull/764) is able to parse the expression without error:

```rust
#[test]
fn render_overflow() {
    let expr = "SELECT 1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1";
    let statements = sqlparser::parser::Parser::parse_sql(&GenericDialect {}, expr).unwrap();
    println!("Was able to parse");
    assert_eq!(&statements[0].to_string(), expr);
}
```

Outputs

```
Was able to parse

thread 'normalize::span::description::sql::parser::render_overflow' has overflowed its stack
fatal runtime error: stack overflow
```

Version: 0.37.0

alamb:
Thank you @jjbayer  -- this would indeed be nice to fix

Issue #1517, opened 2024-11-11
Release sqlparser-rs version `0.53.0` / sqlparser_derive `0.3.0`


alamb:
Follow on to 
- https://github.com/apache/datafusion-sqlparser-rs/issues/1423

This ticket tracks creating the next sqlparser release (mostly so others can follow along)

**Target Release Date**: Approximately 2024-12-11

Release instructions: https://github.com/apache/datafusion-sqlparser-rs/blob/main/dev/release/README.md

## Note we will also have to release a new version of sqlparser-deriver due to https://github.com/apache/datafusion-sqlparser-rs/pull/1556

Steps:
- [x] Make a PR to update version / `CHANGELOG.md`: https://github.com/apache/datafusion-sqlparser-rs/pull/1592
- [x] Create a Release Candidate (RC) and start vote: https://lists.apache.org/thread/5pl93y0og0lldvgrlkzdyb7zh9ymywrn
- [x] (When vote is approved): Release to crates.io
- [x] Make ticket for next release: https://github.com/apache/datafusion-sqlparser-rs/issues/1597
- [x] Note for this time: make a tweet / announcement about the addition of spans

alamb:
Changelog PR ready: https://github.com/apache/datafusion-sqlparser-rs/pull/1592

alamb:
BTW I have verified this release works with DataFusion, see PR: 
- https://github.com/apache/datafusion/pull/13546

alamb:
https://crates.io/crates/sqlparser/0.53.0 published

alamb:
The release is complete 🎉  finally!

PR #1522, opened 2024-11-14
Add `#[recursive]`

blaginin:
Closes https://github.com/apache/datafusion-sqlparser-rs/issues/984, related to https://github.com/apache/datafusion/issues/9375#issuecomment-2469909998


blaginin:
I was wondering if we should remove RecursionCounter with this PR. In my opinion, we shouldn't, because of:
- Backwards compatibility
- The ability to limit max recursion is useful, regardless of our ability to handle large queries without crashing.

blaginin:
Looks like this PR won't slow down anything

```
critcmp main recursive                        
group                                                    main                                   recursive
-----                                                    ----                                   ---------
sqlparser-rs parsing benchmark/format_large_query        1.00   309.1±12.72µs        ? ?/sec    1.01    312.4±7.45µs        ? ?/sec
sqlparser-rs parsing benchmark/parse_large_query         1.02      6.3±0.35ms        ? ?/sec    1.00      6.2±0.40ms        ? ?/sec
sqlparser-rs parsing benchmark/parse_sql_large_query     1.00      6.1±0.23ms        ? ?/sec  
sqlparser-rs parsing benchmark/sqlparser::select         1.02  1893.8±82.19ns        ? ?/sec    1.00  1855.1±25.08ns        ? ?/sec
sqlparser-rs parsing benchmark/sqlparser::with_select    1.03     11.7±1.20µs        ? ?/sec    1.00     11.4±0.28µs        ? ?/sec
```

blaginin:
FYI, I marked this ready for review. @peter-toth @Eason0729, if you guys want to take a look 🌻

iffyio:
@blaginin thanks for looking to fix this!

Currently the preference is to avoid a third-party dependency for this issue, ideally fixing up the parser behavior instead to properly handle deeply nested input. See [comment here](https://github.com/apache/datafusion-sqlparser-rs/pull/1468#issuecomment-2414747574) for a bit more context on rationale
From a quick look at recursive it seems to be a procmacro on top of the stacker library so that the same considerations should apply I imagine.

sqlparser_bench/benches/sqlparser_bench.rs
```diff
@@ -42,6 +42,46 @@ fn basic_queries(c: &mut Criterion) {
     group.bench_function("sqlparser::with_select", |b| {
         b.iter(|| Parser::parse_sql(&dialect, with_query));
     });
+
```
Eason0729, line 45: For `large_statement`, making separated test would make differentiating potential(future) regression easier.
It's a suggestion(fine to leave it as it is).

tests/sqlparser_common.rs
```diff
@@ -11748,3 +11748,16 @@ fn parse_create_table_select() {
         );
     }
 }
+
+#[test]
+fn overflow() {
+    let expr = std::iter::repeat("1")
+        .take(1000)
```
Eason0729, line 11755: 1000 is reasonable for general usage, but 1000 won't overflow without recursive macro. Maybe we could increase it.

src/ast/visitor.rs
```diff
@@ -884,4 +884,29 @@ mod tests {
...
+
+    struct QuickVisitor; // [`TestVisitor`] is too slow to iterate over thousands of nodes
+
+    impl Visitor for QuickVisitor {
+        type Break = ();
+    }
+
+    #[test]
+    fn overflow() {
+        let cond = (0..1000)
```
Eason0729, line 896: Same reason as `sqlparser_common::overflow` test. Maybe we could increase it(increase the number `1000`).

Eason0729 (commented):
Thanks @blaginin. LGTM except for some number in unit test.

I left some comments, and any number above 40000 seems to overflow stack without `#[recursive]`.

peter-toth:
> Currently the preference is to avoid a third-party dependency for this issue, ideally fixing up the parser behavior instead to properly handle deeply nested input. See [comment here](https://github.com/apache/datafusion-sqlparser-rs/pull/1468#issuecomment-2414747574) for a bit more context on rationale From a quick look at recursive it seems to be a procmacro on top of the stacker library so that the same considerations should apply I imagine.

@iffyio, yeah we did some research on stacker / recursive in https://github.com/apache/datafusion/pull/13310 before adding it to datafusion: https://github.com/apache/datafusion/pull/13310#pullrequestreview-2424590466 / https://github.com/apache/datafusion/pull/13177#issuecomment-2465457730

Eason0729:
> > Currently the preference is to avoid a third-party dependency for this issue, ideally fixing up the parser behavior instead to properly handle deeply nested input. See [comment here](https://github.com/apache/datafusion-sqlparser-rs/pull/1468#issuecomment-2414747574) for a bit more context on rationale From a quick look at recursive it seems to be a procmacro on top of the stacker library so that the same considerations should apply I imagine.
> 
> @iffyio, we had done some research on stacker / recursive in https://github.com/apache/datafusion/pull/13310 to verify concerns before we added it to datafusion: https://github.com/apache/datafusion/pull/13310#pullrequestreview-2424590466 / https://github.com/apache/datafusion/pull/13177#issuecomment-2465457730

Sorry for missing that context when reviewing. So we would like to avoid using recursive for stablility, maybe try...
1. vender the code from recursive
2. use stacker directly, like https://github.com/apache/datafusion-sqlparser-rs/pull/1468

tests/sqlparser_common.rs
blaginin: I think it depends on the platform. IMO, the most important thing is that it overflows the stack [in the CI](https://github.com/blaginin/datafusion-sqlparser-rs/actions/runs/11870319597/job/33081591792)

sqlparser_bench/benches/sqlparser_bench.rs
blaginin: Not sure I understood your comment, sorry. I added tests for parsing large statements in tests/sqlparser_common.rs. Do you think we should test something else?

blaginin:
Thanks for the review!! 🙂

> Sorry for missing that context when reviewing. So we would like to avoid using recursive for stablility, maybe try...

We ended up using `#[recursive]` in Datafusion, as Peter highlighted. Feels like it’s good to stay consistent across projects?

iffyio:
Ah I see, thanks for the context @peter-toth!

cc @alamb for overall thoughts on adding this dependency to sqlparser?

sqlparser_bench/benches/sqlparser_bench.rs
Eason0729: I thinks it would be better to add as separated test.

tests/sqlparser_common.rs
Eason0729: Sorry I didn't see Ci fail on that, I reproduce it on my laptop and 1000 seems fined :smile:

alamb:
> Ah I see, thanks for the context @peter-toth!
> 
> cc @alamb for overall thoughts on adding this dependency to sqlparser?

While adding new dependencies in general ls 🤮  I don't think there is any viable alternative in this case 

We have tried to avoid doing something like `stacker` for several years with sqlparser and datafusion, but I feel like we have only been able to band aid over the problem, not fix the problem once and for all.

I am hopeful that if we adopt this particular crate we won't have to worry about it again 🤞

alamb:
I think this PR needs a bit more documentation and we shoudl figure out how to rationalize with the existing `recursion_limit` argument. 


https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.with_recursion_limit

blaginin:
@alamb, I feel like `recursion_limit` should stay because:

- `recursive` doesn't solve the issue when `std` isn’t enabled.  
- Removing it would be a breaking change.  
- It can be a useful feature even with `recursive` protection, for example, if lib users have additional constraints on the parsed query. 

I added notes in the methods I touched; should be better now ✋

Eason0729 (approved):
LGTM

Eason0729:
It seems like we reached the decision to add `recursive` instead of using underlying dependency(stacker).

blaginin:
Just for transparency, there's https://github.com/apache/datafusion/issues/13513 raised in Datafusion but I believe it shouldn't be the reason not to merge this one (happy to be challenged)

alamb:
> Just for transparency, there's [apache/datafusion#13513](https://github.com/apache/datafusion/issues/13513) raised in Datafusion but I believe it shouldn't be the reason not to merge this one (happy to be challenged)

Maybe we could make it an optional dependency 🤔

blaginin:
nice idea actually! will do

README.md
```diff
@@ -63,12 +63,12 @@ The following optional [crate  features](https://doc.rust-lang.org/cargo/referen
...
 * `serde`: Adds [Serde](https://serde.rs/) support by implementing  `Serialize` and `Deserialize` for all AST nodes.
 * `visitor`: Adds a `Visitor` capable of recursively walking the AST tree.
-
+* `recursive-protection` (enabled by default), uses [recursive](https://docs.rs/recursive/latest/recursive/) for stack overflow protection. 
 
 ## Syntax vs Semantics
 
 This crate provides only a syntax parser, and tries to avoid applying
-any SQL semantics, and accepts queries that specific databases would
+any SQL semantics, and accepts queries that specific databases wouldc
```
iffyio, line 71:
```suggestion
any SQL semantics, and accepts queries that specific databases would
```

iffyio (approved):
LGTM!

blaginin:
resolved conflicts, should be good to merge now 🤗

alamb:
Given the potential for unintended side effects with this change, I think we should merge it in asap after we have released

Eason0729:
Should we merge this? or something is missing in this PR.
I have no write access, so cc @alamb .

blaginin:
hey, based on the previous comment I think we want to make a release first 🙂

Eason0729:
> hey, based on the previous comment I think we want to make a release first 🙂

Thanks.
I am sorry for missing that! :eyes:

lovasoa:
Hi @alamb , @iffyio ! Just wanted to report that I just received a crash report that seems to come from here: https://github.com/sqlpage/SQLPage/issues/814

alamb:
> Hi @alamb , @iffyio ! Just wanted to report that I just received a crash report that seems to come from here: [sqlpage/SQLPage#814](https://github.com/sqlpage/SQLPage/issues/814)

Thaks @lovasoa 

Looks like you have also reported a bug to stacker:
 - https://github.com/rust-lang/stacker/issues/115

I am glad there is a way to disable the stacker dependency in sqlparser

Is there anything else you think we should do here?

Thanks again

lovasoa:
Yes, I discovered error handling is completely absent from stacker, and it just crashes the entire program with a cryptic error message when the underlying pthread library returns any error. I submitted a PR, but in the meantime, I wouldn't recommend including it by default in sqlparser, especially if the goal was to avoid crashes in the first place.

lovasoa:
SQLPage does not have a huge install base, and it took just a few days before the first crash report.

alamb:
> SQLPage does not have a huge install base, and it took just a few days before the first crash report.

DataFusion does have a pretty large user base and we haven't gotten crash reports yet that I know of.  I was somewhat worried about using stacker in datafusion too until @peter-toth  pointed out that it was used by rustc itself which allayed my concerns.

it does seem like the usecase 

> glibc targets where procfs isn't mounted at /proc is

Is somewhat uncommon. 

I don't really have a strong opinion one way or the other.

peter-toth:
I didn't notice the missing error handling either. Your https://github.com/rust-lang/stacker/pull/116 seems like a nice improvement. If it doesn't get accepted for some reason, then probably we could do error handling in DF. Maybe adjust the recursive macro to run some tests before using stacker...

alamb:
> I didn't notice the missing error handling either. Your [rust-lang/stacker#116](https://github.com/rust-lang/stacker/pull/116) seems like a nice improvement. But if it doesn't get accepted for some reason, then probably we could handle errors in DF. Maybe adjust the recursive macro to run some tests before using stacker...

- Looks like  https://github.com/rust-lang/stacker/pull/116 was accepted 🎉

Issue #1524, opened 2024-11-15
support column definition list in table alias for postgres


lovasoa:
Postgres supports specifying column names AND types in a table alias (after AS) for functions that return values of type "record".

The following syntax is valid in postgres

```sql
SELECT * FROM jsonb_to_record('{"a": "x", "b": 2}'::jsonb) AS x(a TEXT, b INT);
```

But in sqlparser, it returns

```
Expected: ), found: TEXT
```

This was initially reported in SQLPage, as: https://github.com/sqlpage/SQLPage/discussions/690

lovasoa:
From https://www.postgresql.org/docs/17/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS : 

>  In some cases it is useful to define table functions that can return different column sets depending on how they are invoked. To support this, the table function can be declared as returning the pseudo-type record with no OUT parameters. When such a function is used in a query, the expected row structure must be specified in the query itself, so that the system can know how to parse and plan the query. This syntax looks like:

```sql
function_call [AS] alias (column_definition [, ... ])
function_call AS [alias] (column_definition [, ... ])
ROWS FROM( ... function_call AS (column_definition [, ... ]) [, ... ] )
```

PR #1526, opened 2024-11-15
support column type definitions in table aliases

lovasoa:
```sql
SELECT * FROM some_fun() AS x (a TEXT, b INT)
```

fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1524

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -1627,7 +1627,7 @@ pub struct TableAliasColumnDef {
 
 impl TableAliasColumnDef {
     /// Create a new table alias column definition with only a name and no type
-    pub fn from_column_name(name: &str) -> Self {
+    pub fn from_name<S: Into<String>>(name: S) -> Self {
         TableAliasColumnDef {
             name: Ident::new(name),
             data_type: None,
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -533,9 +533,9 @@ fn parse_select_with_table_alias() {
                 alias: Some(TableAlias {
                     name: Ident::new("l"),
                     columns: vec![
-                        TableAliasColumnDef::from_column_name("A"),
-                        TableAliasColumnDef::from_column_name("B"),
-                        TableAliasColumnDef::from_column_name("C"),
+                        TableAliasColumnDef::from_name("A"),
+                        TableAliasColumnDef::from_name("B"),
+                        TableAliasColumnDef::from_name("C"),
                     ],
                 }),
                 args: None,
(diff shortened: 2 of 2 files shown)


src/parser/mod.rs
```diff
@@ -8607,6 +8607,23 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a parenthesized comma-separated list of unqualified, possibly quoted identifiers
+    pub fn parse_table_alias_column_defs(
```
iffyio, line 8611:
```suggestion
    fn parse_table_alias_column_defs(
```

src/parser/mod.rs
```diff
@@ -8607,6 +8607,23 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a parenthesized comma-separated list of unqualified, possibly quoted identifiers
```
iffyio, line 8610: The comment doesn't seem to match the behavior?

tests/sqlparser_common.rs
```diff
@@ -5576,6 +5580,40 @@ fn parse_table_function() {
     );
 }
 
+#[test]
+fn parse_table_valued_function_with_alias_and_column_defs() {
```
iffyio, line 5584:
```suggestion
fn parse_select_with_alias_and_column_defs() {
```
If I follow correctly this is only testing the column defs functionality and the `FROM T` part hasn't changed with this PR so that it can be any table factor?

src/ast/query.rs
```diff
@@ -1610,6 +1610,40 @@ impl fmt::Display for TableAlias {
...
+    pub data_type: Option<DataType>,
+}
+
+impl TableAliasColumnDef {
+    pub fn from_column_name(name: &str) -> Self {
+        TableAliasColumnDef {
+            name: Ident::new(name),
+            data_type: None,
+        }
+    }
```
iffyio, line 1634: is the function only being exposed publicly in order to use in tests? Thinking if so we could instead replace with a helper function or similar in the tests

lovasoa: It is useful in tests, but also for other library users. Since we add a level of indirection between TableAlias and the actual column names, this makes it easier to instantiate TableAlias in the overwhelming majority of cases when there is only a column name and no type

tests/sqlparser_common.rs
lovasoa: Correct. I updated the function name as per your suggestion.

src/ast/query.rs
iffyio: hmm exposing a helper for Ident I think is reasonable since its quite pervasive, and its simple/dialect-agonstic enough. Here doesnt feel the same, like the increased api surface area is worth it. I'm ok with keeping it, deferring to @alamb 

If we do keep it we should probably rename to something like `from_name` or maybe just `new` like Ident does (thinking a bit unclear what column name refers to, since this is more an alias for the column)?. Also for the parameter to accept e.g. `name: Into<String>` so that folks arent forced to pass a `&str` if they have something else on hand

iffyio (approved):
LGTM! cc @alamb

src/ast/query.rs
lovasoa: Ok, I renamed it to `from_name` and made it generic !

alamb:
nice!

PR #1528, opened 2024-11-16
Support ANTI and SEMI joins without LEFT/RIGHT

### LINKED ISSUE #1504: ANTI join without `LEFT/RIGHT` creates an invalid AST.
if you try to do a `ANTI JOIN` without `<LEFT|RIGHT> ANTI`, it treats `ANTI` as an alias for the table instead of an anti join

example query. This works fine in duckdb. 
```sql
        SELECT * from table1
        ANTI JOIN table2 t2 on t1.id = t2.id 
```

Full Example. 
```rs
use sqlparser::dialect::GenericDialect;

fn main() {
    let sql = r#"
        SELECT * from table1
        ANTI JOIN table2 t2 on t1.id = t2.id 
        "#;

    println!("AST: {:?}", ast);
}
```

delamarch3:
Closes #1504 

User-facing change: `ANTI` and `SEMI` have been added to the reserved keywords for table aliases

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -1694,6 +1694,13 @@ impl fmt::Display for Join {
                 suffix(constraint)
             ),
             JoinOperator::CrossJoin => write!(f, " CROSS JOIN {}", self.relation),
+            JoinOperator::Semi(constraint) => write!(
+                f,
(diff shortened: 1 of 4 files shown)


iffyio (approved):
LGTM! cc @alamb

PR #1531, opened 2024-11-18
Add support for PostgreSQL `UNLISTEN` syntax and Add support for Postgres `LOAD extension` expr

wugeer:
This PR supports two PostgreSQL syntaxes，one is
- `UNLISTEN` clause,. For more information, please refer to:
https://www.postgresql.org/docs/current/sql-unlisten.html

and the other is 
-  `LOAD extension` . For more information, please refer to:
https://www.postgresql.org/docs/current/sql-load.html

It also introduces the following keywords:
- UNLISTEN

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -633,18 +633,8 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if the dialect supports the `LISTEN` statement
-    fn supports_listen(&self) -> bool {
-        false
-    }
-
-    /// Returns true if the dialect supports the `UNLISTEN` statement
-    fn supports_unlisten(&self) -> bool {
-        false
-    }
-
-    /// Returns true if the dialect supports the `NOTIFY` statement
-    fn supports_notify(&self) -> bool {
+    /// Returns true if the dialect supports the `LISTEN`, `UNLISTEN` and `NOTIFY` statements
+    fn supports_listen_notify(&self) -> bool {
         false
     }
 
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -638,6 +638,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports the `UNLISTEN` statement
+    fn supports_unlisten(&self) -> bool {
+        false
+    }
+
     /// Returns true if the dialect supports the `NOTIFY` statement
     fn supports_notify(&self) -> bool {
         false
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -191,17 +191,9 @@ impl Dialect for PostgreSqlDialect {
     }
 
     /// see <https://www.postgresql.org/docs/current/sql-listen.html>
-    fn supports_listen(&self) -> bool {
-        true
-    }
-
     /// see <https://www.postgresql.org/docs/current/sql-unlisten.html>
-    fn supports_unlisten(&self) -> bool {
-        true
-    }
-
     /// see <https://www.postgresql.org/docs/current/sql-notify.html>
-    fn supports_notify(&self) -> bool {
+    fn supports_listen_notify(&self) -> bool {
         true
     }
 
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -195,6 +195,11 @@ impl Dialect for PostgreSqlDialect {
         true
     }
 
+    /// see <https://www.postgresql.org/docs/current/sql-unlisten.html>
+    fn supports_unlisten(&self) -> bool {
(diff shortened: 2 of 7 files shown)


tests/sqlparser_common.rs
```diff
@@ -11864,6 +11895,56 @@ fn parse_load_data() {
     );
 }
 
+#[test]
+fn test_load_extension() {
+    let dialects = all_dialects_where(|d| d.supports_load_extension());
+    let only_supports_load_data_dialects =
+        all_dialects_where(|d| !d.supports_load_extension() && d.supports_load_data());
+    let not_supports_load_dialects =
+        all_dialects_where(|d| !d.supports_load_data() && !d.supports_load_extension());
```
iffyio, line 11904: given we're only looking to test the load_extension syntax, would it make sense to only keep the test scoped to only `all_dialects_where(|d| d.supports_load_extension())` in order to keep the test focused?
Otherwise its not clear if/why the load_data syntax is relevant here. Then for the `not_supports_load_dialects` I think we can skip verifying the parser error on other dialects, since by definition they don't support the syntax so its not super important

src/parser/mod.rs
```diff
@@ -999,6 +1000,21 @@ impl<'a> Parser<'a> {
...
 
+    pub fn parse_unlisten(&mut self) -> Result<Statement, ParserError> {
+        let channel = if self.consume_token(&Token::Mul) {
+            Ident::new(Expr::Wildcard.to_string())
+        } else {
+            match self.parse_identifier(false) {
+                Ok(expr) => expr,
+                _ => {
+                    self.prev_token();
+                    return self.expected("wildcard or identent", self.peek_token());
```
iffyio, line 1011:
```suggestion
                    return self.expected("wildcard or identifier", self.peek_token());
```

src/dialect/mod.rs
```diff
@@ -638,6 +638,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports the `UNLISTEN` statement
+    fn supports_unlisten(&self) -> bool {
+        false
+    }
```
iffyio, line 644: Oh actually I was initially wondering if it makes sense to reuse the existing `supports_listen()` flag for this but then also realised we could introduce a `dialect.supports_listen_notify()` to mark the feature itself, replacing `dialect.supports_listen()`, `dialect.supports_notify()` and `dialect.supports_unlisten()`. Thinking since it would be expected for example that a dialect supporting listen should also support notify?

src/dialect/postgresql.rs
```diff
@@ -209,6 +214,11 @@ impl Dialect for PostgreSqlDialect {
     fn supports_comment_on(&self) -> bool {
         true
     }
+
+    /// See <https://www.postgresql.org/docs/current/sql-load.html>
+    fn supports_load_extension(&self) -> bool {
+        true
+    }
```
iffyio, line 221: Oh, I'm not sure I followed this change, I can see this flag and tests for it but there doesn't seem to be any changes to the parser itself?

iffyio (commented):
Thanks @wugeer! left a couple comments

src/dialect/mod.rs
wugeer: Yes, although this part of the syntax might be closely related, I think it's better to separate them as independent features. This is because `listen`, `unlisten` and `notify` might not necessarily be supported together. On this issue, perhaps we could get more input from others to consider it more comprehensively. What do you think? @iffyio

src/dialect/postgresql.rs
wugeer: As I understand it, this change simply adds support for the `LOAD  extension` statement to the `PostgreSQL` dialect, while the  `DuckDB` and `Generic` dialects both already support it. Therefore, no changes are needed to the parser. In my local environment, this modification is sufficient. If my understanding is incorrect, please feel free to correct me. ：）

tests/sqlparser_common.rs
wugeer: Good idea! Have Done!

wugeer:
@iffyio Thank you for reviewing the code. I've pushed the changes again. Looking forward to your next review ：）

src/dialect/postgresql.rs
iffyio: Ah yes ofc my bad! I missed that the feature already existed and was just being enabled for postgres :+1:

src/dialect/mod.rs
wugeer: @iffyio You make a good point. Currently, there are no dialects that support LISTEN but not NOTIFY, so let's go ahead and combine them for now.Have Done!  :)

iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

PR #1532, opened 2024-11-18
Parse byte/bit string literals in MySQL and Postgres

mvzink:


tests/sqlparser_postgres.rs
```diff
@@ -5098,3 +5098,34 @@ fn parse_create_type_as_enum() {
...
+            SetExpr::Values(Values {
+                explicit_row: false,
+                rows: vec![vec![Expr::Value(Value::SingleQuotedByteStringLiteral(
+                    "111".to_string()
+                ))]]
+            })
+        ),
+        _ => unreachable!(),
+    }
+}
```
iffyio, line 5131:
```suggestion
#[test]
fn parse_bitstring_literal() {
    let select = pg_and_generic().verified_only_select("SELECT B'111'");
    assert_eq!(
        select.projection,
        vec![SelectItem::UnnamedExpr(Expr::Value(
            Value::SingleQuotedByteStringLiteral("111".to_string())
        ))]
    );
}
```
Thinking its probably enough to show that the dialect is enabled to parse bitstrings, and that its not specific to select/insert. (Same fore mysql)

mvzink: Fair enough!

alamb (approved):
Makes sense to me -- thank you @mvzink and @iffyio

iffyio (approved):
LGTM!

PR #1534, opened 2024-11-19
PartiQL queries in Redshift

yoavcloud:
This PR adds support for parsing PartiQL queries in Redshift


tests/sqlparser_redshift.rs
```diff
@@ -196,3 +199,118 @@ fn test_create_view_with_no_schema_binding() {
     redshift_and_generic()
         .verified_stmt("CREATE VIEW myevent AS SELECT eventname FROM event WITH NO SCHEMA BINDING");
 }
+
+#[test]
+fn test_redshift_json_path() {
+    let sql = "SELECT cust.c_orders[0].o_orderkey FROM customer_orders_lineitem";
+    let select = redshift().verified_only_select(sql);
```
iffyio, line 206: Could we use `let dialect = all_dialects_where(|d| d.supports_partiql())` for the tests? thinking so any dialect that gets this feature can potentially reuse the same tests

src/ast/query.rs
```diff
@@ -974,6 +974,8 @@ pub enum TableFactor {
         with_ordinality: bool,
         /// [Partition selection](https://dev.mysql.com/doc/refman/8.0/en/partitioning-selection.html), supported by MySQL.
         partitions: Vec<Ident>,
+        /// Optional PartiQL JsonPath: <https://partiql.org/dql/from.html>
+        partiql: Option<JsonPath>,
```
iffyio, line 978: Maybe we can call this `json_path` to be generic (in the doc we mention partiql as one example)?

tests/sqlparser_redshift.rs
```diff
@@ -196,3 +199,118 @@ fn test_create_view_with_no_schema_binding() {
...
+                ]
+            }
+        },
+        expr_from_projection(only(&select.projection))
+    );
+}
+
+#[test]
+fn test_parse_json_path_from() {
+    let select = redshift().verified_only_select("SELECT * FROM src[0].a AS a");
```
iffyio, line 254: Could we add a test for something containing a qualified name e.g. `a.b.c[0].d` (it wasn't clear from the code how that's represented or rejected if its not valid partiql)?

iffyio (approved):
LGTM! cc @alamb

alamb:
Thanks @yoavcloud and @iffyio

PR #1535, opened 2024-11-19
Add support for data type specific methods

yoavcloud:
This PR adds support for static method invocations of data type-specific methods. For example: 
`SELECT geography::STGeomFromText('LINESTRING(-122.360 47.656, -122.343 47.656 )', 4326)`

See more here: https://learn.microsoft.com/en-us/sql/t-sql/spatial-geography/spatial-types-geography?view=sql-server-ver16

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1224,6 +1224,17 @@ impl<'a> Parser<'a> {
                             body: Box::new(self.parse_expr()?),
                         }));
                     }
+                    Token::DoubleColon if self.dialect.supports_methods() => {
+                        let namespace = w.value;
+                        self.expect_token(&Token::DoubleColon)?;
+                        let name_with_namespace = match self.next_token().token {
+                            Token::Word(func_name) => {
+                                ObjectName(vec![Ident::new(format!("{}::{}", namespace, func_name.value))])
+                            },
+                            _ => return self.expected("identifier", self.peek_token())
+                        };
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -1224,6 +1224,17 @@ impl<'a> Parser<'a> {
...
+                    Token::DoubleColon if self.dialect.supports_methods() => {
+                        let namespace = w.value;
+                        self.expect_token(&Token::DoubleColon)?;
+                        let name_with_namespace = match self.next_token().token {
+                            Token::Word(func_name) => {
+                                ObjectName(vec![Ident::new(format!("{}::{}", namespace, func_name.value))])
+                            },
+                            _ => return self.expected("identifier", self.peek_token())
+                        };
+                        Ok(self.parse_function(name_with_namespace)?)
```
iffyio, line 1236: Can we move the body to a `self.parse_namespaced_method(w)?` or similar function? Also if we can add an example syntax to the function doc with the link to clarify the syntax

tests/sqlparser_common.rs
```diff
@@ -11550,6 +11550,14 @@ fn parse_method_expr() {
         }
         _ => unreachable!(),
     }
+
+    dialects.verified_stmt(
```
iffyio, line 11554: Can we add a scenario in the style of `a::b::c().d()`?

PR #1536, opened 2024-11-20
Allow example CLI to read from stdin

mvzink:


iffyio (approved):
LGTM! cc @alamb

alamb (commented):
Thanks @mvzink  and @iffyio 

I tried it out

```
andrewlamb@Andrews-MacBook-Pro-2:~/Software/sqlparser-rs$ cargo run --example cli -
```

The only thing that was somewhat weird was I had to type `Ctrl-D` , but maybe we can treat ttys differently in the future

PR #1538, opened 2024-11-20
Fix MySQL parsing of GRANT, REVOKE, and CREATE VIEW

mvzink:
- Support MySQL style `'user'@'host'` syntax.
- Support wildcards on grant objects: `*.*`.
- Make `CASCADE`/`RESTRICT` in revoke statements optional since MySQL doesn't accept it.
- Support optional MySQL parameters for create view: algorithm, definer (requires `user@host` syntax), and security context.


mvzink:
Would appreciate feedback on the naming the MySQL-specific params types (would a generic/non-branded name be better?) and the API change for wildcard support (didn't want to change all 107 `parse_object_name` callsites for an option that is so niche, but happy to be convinced otherwise).

src/tokenizer.rs
```diff
@@ -1202,6 +1202,9 @@ impl<'a> Tokenizer<'a> {
                             }
                         }
                         Some(' ') => Ok(Some(Token::AtSign)),
+                        Some('\'') => Ok(Some(Token::AtSign)),
+                        Some('\"') => Ok(Some(Token::AtSign)),
+                        Some('`') => Ok(Some(Token::AtSign)),
```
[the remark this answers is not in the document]
mvzink: I thought so at first too, but variables in MySQL start with `@` and I don't think there's a way to differentiate them during tokenization. We would want to tokenize `@foo` as a single `Token::Word`, but `@'foo'` (in my understanding) as `[Token::AtSign, Token::SingleQuotedString]`. In this check during tokenization, we aren't differentiating between "schema names etc." and "variable names", just "identifiers". Maybe there is something more correct to do here, like checking for `self.is_identifier_continuation(sch)` and ruling out `'` there. But my approach of just adding cases for quotes is based on the fact that as far as I can tell, no dialect allows quote marks within an identifier unless it's in an outer quotation, which doesn't get tokenized by this path.

I think an approach like `self.is_identifier_continuation(sch)` would effectively be the same but could allow customization per dialect, but it has another pitfall: say in addition to quote marks, a dialect doesn't allow `_` in an identifier. Then should we fail to tokenize `@_` with an error, or tokenize them separately as `[@, _]`? If the former, we would then need another special case for quote marks, and the latter seems wrong to me.

One other example in support of tokenizing like `[Token::AtSign, Token::SingleQuotedString]`, just in case it's not clear why I think that's right: in Postgres, `@` is absolute value, and it will implicitly cast the operand to a numeric type, so that `select @'1';` (no space after `@`) treats `'1'` as a string and casts it to double.

All that in mind, I still think my current approach is the simplest correct I can think of; let me know what you think!

src/parser/mod.rs
```diff
@@ -4693,9 +4699,59 @@ impl<'a> Parser<'a> {
...
+            Some(
+                match self.expect_one_of_keywords(&[
+                    Keyword::UNDEFINED,
+                    Keyword::MERGE,
+                    Keyword::TEMPTABLE,
+                ])? {
+                    Keyword::UNDEFINED => MySQLViewAlgorithm::Undefined,
+                    Keyword::MERGE => MySQLViewAlgorithm::Merge,
+                    Keyword::TEMPTABLE => MySQLViewAlgorithm::TempTable,
+                    _ => unreachable!(),
```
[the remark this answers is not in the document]
mvzink: Did so be rewinding the token to then call `self.expected` (otherwise I don't see how to get the location that it wants). Seems a bit clunky when this is exactly the error that `self.expect_one_of_keywords` should have already returned. The only way we would hit this branch is if someone added a keyword to the `self.expect_one_of_keywords` call but forgot to add a corresponding `match` arm. One overengineered solution which could be used here and elsewhere would be to wrap it in a macro that looks like a `match` but collects the keywords from the match arms into the `self.expect_one_of_keywords` for you.

src/parser/mod.rs
```diff
@@ -8353,20 +8409,40 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a possibly qualified, possibly quoted identifier, optionally allowing for wildcards,
+    /// e.g. *, `foo`.*, or "foo"."bar"
+    pub fn parse_object_name_with_wildcards(
```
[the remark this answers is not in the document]
mvzink: Still having a think about the best way to do this; posting the other updates for further feedback.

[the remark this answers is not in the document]
mvzink: Sorry it's taken me a while to get to this. It's a bit tricky, partly because the expression form shouldn't accept `*.*`. I will try to come back to this later, but have had higher priorities the last few weeks and probably will until later in January. In my opinion, it might be better to avoid dragging on with more merge conflicts etc. and merge without this suggested refactor. Let me know your thoughts.

src/parser/mod.rs
```diff
@@ -3869,11 +3873,16 @@ impl<'a> Parser<'a> {
             .is_some();
         let persistent = dialect_of!(self is DuckDbDialect)
             && self.parse_one_of_keywords(&[Keyword::PERSISTENT]).is_some();
+        let create_view_params = if dialect_of!(self is MySqlDialect | GenericDialect) {
+            self.parse_create_view_params()?
+        } else {
+            None
+        };
```
iffyio, line 3880: maybe this could be accepted without the dialect check if that works just as well `let create_view_paras = self.parse_create_view_params()`?  thinking since the view params are optional

src/parser/mod.rs
```diff
@@ -11375,7 +11466,11 @@ impl<'a> Parser<'a> {
         } else {
             let object_type =
                 self.parse_one_of_keywords(&[Keyword::SEQUENCE, Keyword::SCHEMA, Keyword::TABLE]);
-            let objects = self.parse_comma_separated(|p| p.parse_object_name(false));
+            let objects = if dialect_of!(self is MySqlDialect | GenericDialect) {
+                self.parse_comma_separated(|p| p.parse_object_name_with_wildcards(false, true))
+            } else {
+                self.parse_comma_separated(|p| p.parse_object_name(false))
+            };
```
iffyio, line 11473: Thinking similarly here, if there aren't any conflicting behavior we could skip the dialect check and always parse the object name with wildcard?

src/parser/mod.rs
```diff
@@ -4842,9 +4852,68 @@ impl<'a> Parser<'a> {
             if_not_exists,
             temporary,
             to,
+            params: create_view_params,
         })
     }
 
+    /// Parse optional algorithm, definer, and security context parameters for [MySQL]
```
iffyio, line 4859:
```suggestion
    /// Parse optional parameters for the `CREATE VIEW` statement. [MySQL]
```
or something similar, thinking so since this can be extended later beyond mysql today

src/parser/mod.rs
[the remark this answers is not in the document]
iffyio: Ah not a problem I think this sounds reasonable! Could we maybe make the function non-public in case it needs to change later on so that we don't consider downstream impact?

iffyio (commented):
Thanks @mvzink! Left some minor comments I think with those and resolving conflicts on the branch this looks good to me!

src/parser/mod.rs
mvzink: There's no "conflicting behavior" per se, so the tests pass without these 2 checks, and I definitely see the value of reducing switches on dialect (especially using `dialect_of!` rather than a method on the dialect). But to be clear, this will mean that using `PostgreSqlDialect` for example will accept syntax that Postgres itself doesn't accept. If that's worth the tradeoff to reduce switching in here, I'm okay with that but just want to be clear that's okay.

iffyio: Ah yeah that would be fine, the dialect impls can accept a superset of the actual sql spec

iffyio (approved):
LGTM! Thanks @mvzink
cc @alamb

iffyio:
@mvzink it seems there's a couple conflicts in the branch could you take a look when you have some time?

iffyio:
Hi sorry I just noticed this was since rebased, could I ask for another rebase and if you could just ping me when its done I'll prio merging it to avoid further conflicts :pray:

mvzink:
@iffyio rebased, but there were some significant changes to grantee parsing by @yoavcloud so maybe he should take a look too.

src/parser/mod.rs
```diff
@@ -11808,23 +11899,32 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_grantee_name(&mut self) -> Result<GranteeName, ParserError> {
+        let name = self.parse_object_name(false)?;
+        if self.dialect.supports_user_host_grantee()
+            && name.0.len() == 1
+            && self.consume_token(&Token::AtSign)
+        {
+            let user = name.0[0].clone();
```
mvzink, line 11908: @yoavcloud I believe in #1640 you changed grantee parsing from `parse_identifier` to `parse_object_name`; is that actually used by Snowflake? Would like to find a way to avoid this clone to get the identifier out of `name`.

yoavcloud: I see at least one case in Snowflake where you grant privileges to a database role, which can be a fully qualified object name. See here: https://docs.snowflake.com/en/sql-reference/sql/grant-privilege under database_role_name.

mvzink:
@iffyio I think this is good to go.

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1539, opened 2024-11-20
Enhance object name path segments

ayman-sigma:
Right now `ObjectName` is just list of identifiers. We parse each object name path segment as a string identifier.  Some dialects has more rich types for each path segment. This PR rework the object name to allow different types for each path segment.

Examples this PR will make it easier to support:
1. Databricks `IDENTIFIER` clause. Example: `SELECT * FROM myschema.IDENTIFIER(:mytab)`. The `(:mytab)` is wrongly parsed right now as `TableFunctionArgs`.  More details: https://docs.databricks.com/en/sql/language-manual/sql-ref-names-identifier-clause.html
2. Snowflake double-dot notation. Example `SELECT * FROM db..table_name`. This indicates that use of default schema `PUBLIC`.  With this PR, we can use `DefaultSchema` variant for the path segment instead of using empty identifier. More details: https://docs.snowflake.com/en/sql-reference/name-resolution#resolution-when-schema-omitted-double-dot-notation

Most changes are mechanical except couple of locations I commented on below, in addition to the `ast/mod.rs`.


src/parser/mod.rs
```diff
@@ -4294,7 +4312,9 @@ impl<'a> Parser<'a> {
         let mut data_type = self.parse_data_type()?;
         if let DataType::Custom(n, _) = &data_type {
             // the first token is actually a name
-            name = Some(n.0[0].clone());
+            match n.0[0].clone() {
+                ObjectNamePart::Identifier(ident) => name = Some(ident),
```
ayman-sigma, line 4316: Once we start adding more to the `ObjectNamePart` enum, we will return parsing error for the other variants here.

src/ast/mod.rs
```diff
@@ -195,14 +199,36 @@ impl fmt::Display for Ident {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum ObjectNamePart {
+    Identifier(Ident),
+}
+
+impl fmt::Display for ObjectNamePart {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ObjectNamePart::Identifier(ident) => write!(f, "{}", ident),
```
mvzink, line 227: nit: I prefer `ident.fmt(f)`

src/parser/mod.rs
```diff
@@ -12579,7 +12599,9 @@ impl<'a> Parser<'a> {
                     .parse_one_of_keywords(&[Keyword::FROM, Keyword::IN])
                     .is_some()
                 {
-                    parent_name.0.insert(0, self.parse_identifier(false)?);
+                    parent_name
+                        .0
+                        .insert(0, self.parse_identifier(false)?.to_object_name_part());
```
mvzink, line 12604: Based on this one usage (and I don't see another one in your snowflake PR), couldn't `to_object_name_part` just take ownership and rewrap instead of cloning?

ayman-sigma: You are right. I actually was planning to remove it but forgot. I thought I would use it more

src/parser/mod.rs
```diff
@@ -10816,7 +10836,7 @@ impl<'a> Parser<'a> {
         self.expect_token(&Token::LParen)?;
         let aggregate_functions = self.parse_comma_separated(Self::parse_aliased_function_call)?;
         self.expect_keyword(Keyword::FOR)?;
-        let value_column = self.parse_object_name(false)?.0;
+        let value_column = self.parse_period_separated_identifiers()?;
```
iffyio, line 10839:
```suggestion
        let value_column = `self.parse_period_separated(|p| p.parse_identifier(false))?;
```
Can we inline the code? thinking since the current method is only a wrapper around it

src/parser/mod.rs
```diff
@@ -3605,6 +3605,24 @@ impl<'a> Parser<'a> {
         Ok(values)
     }
 
+    pub fn parse_period_separated<T, F>(&mut self, mut f: F) -> Result<Vec<T>, ParserError>
```
iffyio, line 3608:
```suggestion
    fn parse_period_separated<T, F>(&mut self, mut f: F) -> Result<Vec<T>, ParserError>
```
Can we add a short description mentioning what the function does?

src/parser/mod.rs
```diff
@@ -3605,6 +3605,24 @@ impl<'a> Parser<'a> {
...
+        loop {
+            values.push(f(self)?);
+            if !self.consume_token(&Token::Period) {
+                break;
+            }
+        }
+        Ok(values)
+    }
+
+    pub fn parse_period_separated_identifiers(&mut self) -> Result<Vec<Ident>, ParserError> {
```
iffyio, line 3622:
```suggestion
    fn parse_period_separated_identifiers(&mut self) -> Result<Vec<Ident>, ParserError> {
```

iffyio (commented):
Thanks @ayman-sigma! left some minor comments, this looks good to me overall

iffyio (approved):
LGTM! cc @alamb

alamb:
Hi @ayman-sigma  this PR appears to have some conflicts. Is there any chance you can resolve them so we can merge it in?

Thank you!

ayman-sigma:
> Hi @ayman-sigma this PR appears to have some conflicts. Is there any chance you can resolve them so we can merge it in?
> 
> Thank you!

Done.

ayman-sigma:
> Hi @ayman-sigma this PR appears to have some conflicts. Is there any chance you can resolve them so we can merge it in?
> 
> Thank you!

@alamb, Done.

alamb (commented):
I started trying to update DataFusion to use this change -- it turns out to be fairly invasive.

You can try here: https://github.com/apache/datafusion/pull/13546

(the issue is that we have a bunch of handling of ObjectName --> Indents code).

I think we can make the DataFusion code better / easier to follow

alamb:
Given the potential for non trivial downstream conflicts due to this change (look at the list of conflicts it has already collected) I would like to consider it for the next release

ayman-sigma:
> Given the potential for non trivial downstream conflicts due to this change (look at the list of conflicts it has already collected) I would like to consider it for the next release
> 
> * [Release sqlparser-rs version `0.53.0` / sqlparser_derive `0.3.0` #1517](https://github.com/apache/datafusion-sqlparser-rs/issues/1517)

Sounds good. Thanks @alamb!

iffyio:
@alamb just wanted to double check status of this PR if there were reservations you had or if you feel this is something we would be able to land?

alamb:
> @alamb just wanted to double check status of this PR if there were reservations you had or if you feel this is something we would be able to land?

My biggest reservation was that it would cause substantial downstream churn (I tried to make the changes to DataFUsion briefly and it was painful). So I just haven't had the heart to click the merge button

I mentially was prepared if you merged it I would sort it out downstraem but I couldn't get myself to inflict the main on myself ...

iffyio:
Ah yeah this is indeed an invasive change. Alright that makes sense! 

In that case @ayman-sigma please take a look at resolving the conflicts when you have some time to pick this back up and we can look to merge it? Sorry for the delay in getting to it

alamb:
FWIW I did a test upgrade to DataFusion to prepare for the next release and it already had some non trivial changes needed (changes to FieldAccess specifically)

- https://github.com/apache/datafusion/pull/14198

ayman-sigma:
> FWIW I did a test upgrade to DataFusion to prepare for the next release and it already had some non trivial changes needed (changes to FieldAccess specifically)
> 
> * [Test upgrade to sqlparser-rs 0.54 datafusion#14198](https://github.com/apache/datafusion/pull/14198)

I'm fine to drop this PR if will cause too much pain downstream. Let me know if we still want to go with this PR and I can fix the conflicts.

iffyio:
@ayman-sigma `FieldAccess` being referred to was part of the 0.54 release, those changes were breaking but the syntax isn't tied to the objectName changes/this PR.

I think it would be good to get this PR in since it'll be good to support the `Identifier` syntax, unfortunately that syntax isn't uncommon in Snowflake especially, and I couldn't come up with a variant that improves upon this PR in terms of being backwards compatible.
I figured an alternative could be to let the parser accept the syntax but drop the `Identifier` keyword so that the actual AST isn't affected. But afaict it would only be able to support `Identifier(<string_literal> | <variable>)` variants. So I figure it would make sense with the changes as in this PR that lets us support the full syntax going forward.

So my thinking was to get this PR in for the next 0.55 release, and then any further extensions to the representation can come in later releases afterwards

ayman-sigma:
@iffyio I rebased to main and resolved all conflicts. I made some changes in the last commit to best of my knowledge. Please make sure to review the last commit. Thanks!

iffyio:
Thanks @ayman-sigma!

PR #1540, opened 2024-11-20
Support snowflake double dot notation for object name

ayman-sigma:
Snowflake allows double-dot notation in the object name. The two dots indicate that the schema name is not specified. The PUBLIC default schema is always referenced. See more details here: https://docs.snowflake.com/en/sql-reference/name-resolution#resolution-when-schema-omitted-double-dot-notation

Right now I'm parsing the default schema segment as an empty identifier. Depending on the acceptance of [this PR](https://github.com/apache/datafusion-sqlparser-rs/pull/1539), we can change this to be something like `DefaultSchema` instead.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8384,9 +8384,9 @@ impl<'a> Parser<'a> {
         loop {
             if self.dialect.supports_object_name_double_dot_notation()
                 && idents.len() == 1
-                && self.peek_token() == Token::Period
+                && self.consume_token(&Token::Period)
             {
-                self.next_token();
+                // Empty string here means default schema
                 idents.push(Ident::new(""));
             }
             idents.push(self.parse_identifier(in_table_clause)?);
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -2869,13 +2869,7 @@ fn test_projection_with_nested_trailing_commas() {
 fn test_sf_double_dot_notation() {
     snowflake().verified_stmt("SELECT * FROM db_name..table_name");
     snowflake().verified_stmt("SELECT * FROM x, y..z JOIN a..b AS b ON x.id = b.id");
-}
 
-#[test]
-fn test_sf_double_dot_notation_wrong_position() {}
-
-#[test]
-fn test_parse_double_dot_notation_wrong_position() {
     assert_eq!(
         snowflake()
             .parse_sql_statements("SELECT * FROM X.Y..")
@@ -2883,7 +2877,13 @@ fn test_parse_double_dot_notation_wrong_position() {
             .to_string(),
         "sql parser error: Expected: identifier, found: ."
     );
-
+    assert_eq!(
+        snowflake()
+            .parse_sql_statements("SELECT * FROM X..Y..Z")
+            .unwrap_err()
+            .to_string(),
+        "sql parser error: Expected: identifier, found: ."
+    );
     assert_eq!(
         // Ensure we don't parse leading token
         snowflake()
@@ -2893,3 +2893,6 @@ fn test_parse_double_dot_notation_wrong_position() {
         "sql parser error: Expected: identifier, found: ."
     );
 }
+
+#[test]
+fn test_parse_double_dot_notation_wrong_position() {}


src/parser/mod.rs
```diff
@@ -8349,6 +8349,13 @@ impl<'a> Parser<'a> {
     pub fn parse_object_name(&mut self, in_table_clause: bool) -> Result<ObjectName, ParserError> {
         let mut idents = vec![];
         loop {
+            if self.dialect.supports_object_name_double_dot_notation()
+                && !idents.is_empty()
```
ayman-sigma, line 8353: I'm debating if I should force it only for missing schema. Which means only accepts `x..y` and rejects stuff like `x.y..z`. Forcing this seems like a semantic validation that we are trying to avoid.

tests/sqlparser_snowflake.rs
```diff
@@ -2846,3 +2846,32 @@ fn test_parse_show_columns_sql() {
...
 }
+
+#[test]
+fn test_sf_double_dot_notation() {
+    snowflake().verified_stmt("SELECT * FROM db_name..table_name");
+    snowflake().verified_stmt("SELECT * FROM x, y..z JOIN a..b AS b ON x.id = b.id");
+}
+
+#[test]
+fn test_sf_double_dot_notation_wrong_position() {}
```
iffyio, line 2857: ah did we plan to put in a test scenario for this behavior?

tests/sqlparser_snowflake.rs
```diff
@@ -2846,3 +2846,32 @@ fn test_parse_show_columns_sql() {
...
+
+    assert_eq!(
+        // Ensure we don't parse leading token
+        snowflake()
+            .parse_sql_statements("SELECT * FROM .X.Y")
+            .unwrap_err()
+            .to_string(),
+        "sql parser error: Expected: identifier, found: ."
+    );
+}
```
iffyio, line 2877: can we merge the methods into `test_sf_double_dot_notation()` so that the latter contains the different scenarios? thinking that would help keep the tests focused to the feature since the feature is limited enough in scope

tests/sqlparser_snowflake.rs
```diff
@@ -2846,3 +2846,32 @@ fn test_parse_show_columns_sql() {
     snowflake().verified_stmt("SHOW COLUMNS IN TABLE abc");
     snowflake().verified_stmt("SHOW COLUMNS LIKE '%xyz%' IN TABLE abc");
 }
+
+#[test]
+fn test_sf_double_dot_notation() {
```
iffyio, line 2851: can we add a test case containing multiple double dots? e.g. `select * from x, y..z..w`

src/parser/mod.rs
```diff
@@ -8349,6 +8349,13 @@ impl<'a> Parser<'a> {
     pub fn parse_object_name(&mut self, in_table_clause: bool) -> Result<ObjectName, ParserError> {
         let mut idents = vec![];
         loop {
+            if self.dialect.supports_object_name_double_dot_notation()
+                && idents.len() == 1
+                && self.peek_token() == Token::Period
+            {
+                self.next_token();
```
iffyio, line 8356:
```suggestion
                && self.consume_token(&Token::Period)
            {
```

src/parser/mod.rs
```diff
@@ -8349,6 +8349,13 @@ impl<'a> Parser<'a> {
...
+                idents.push(Ident::new(""));
```
iffyio, line 8357: Oh is this line correct, I imagined we would push `Ident::new(".")`?

iffyio: Ah I realised this is empty due to the display adding the dot on its behalf. Could we add a comment what this line does? since its not obvious otherwise

tests/sqlparser_snowflake.rs
ayman-sigma: Ya, I left it by mistake. Will remove.

tests/sqlparser_snowflake.rs
ayman-sigma: Sounds good

src/parser/mod.rs
ayman-sigma: Yes, I'm also waiting on #1539 to merge and then I will change this to something like `DefaultSchema` instead.

iffyio (approved):
LGTM! cc @alamb

alamb (approved):
Makes sense to me -- thanks @ayman-sigma  and @iffyio

PR #1542, opened 2024-11-21
Support Databricks struct literal

ayman-sigma:
Databricks supports struct literal (struct function). Struct literal is already added for BigQuery. Databricks' syntax is very similar to the BigQuery one, except the fields definition part.
I added `supports_struct_literal` to `Dialect` trait to specify if the dialect support this literal or not

Databricks docs: https://docs.databricks.com/en/sql/language-manual/functions/struct.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1123,9 +1123,8 @@ impl<'a> Parser<'a> {
             Keyword::MATCH if dialect_of!(self is MySqlDialect | GenericDialect) => {
                 Ok(Some(self.parse_match_against()?))
             }
-            Keyword::STRUCT if dialect_of!(self is BigQueryDialect | GenericDialect) => {
-                self.prev_token();
-                Ok(Some(self.parse_bigquery_struct_literal()?))
+            Keyword::STRUCT if self.dialect.supports_struct_literal() => {
+                Ok(Some(self.parse_struct_literal()?))
             }
             Keyword::PRIOR if matches!(self.state, ParserState::ConnectBy) => {
                 let expr = self.parse_subexpr(self.dialect.prec_value(Precedence::PlusMinus))?;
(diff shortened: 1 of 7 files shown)


src/parser/mod.rs
```diff
@@ -2328,19 +2327,25 @@ impl<'a> Parser<'a> {
...
     /// ```
-    fn parse_bigquery_struct_literal(&mut self) -> Result<Expr, ParserError> {
-        let (fields, trailing_bracket) =
-            self.parse_struct_type_def(Self::parse_struct_field_def)?;
-        if trailing_bracket.0 {
-            return parser_err!("unmatched > in STRUCT literal", self.peek_token().location);
+    fn parse_struct_literal(&mut self) -> Result<Expr, ParserError> {
+        let mut fields = vec![];
+        // Typed struct syntax is only supported by BigQuery
+        // https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#typed_struct_syntax
```
iffyio, line 2340:
```suggestion
```

src/parser/mod.rs
```diff
@@ -2328,19 +2327,25 @@ impl<'a> Parser<'a> {
         }
     }
 
-    /// Bigquery specific: Parse a struct literal
     /// Syntax
     /// ```sql
-    /// -- typed
+    /// -- typed, specific to bigquery
```
iffyio, line 2332:
```suggestion
    /// -- typed
```
for the comments mentioning bigquery specific I think we can skip since the definition is now encoded in the dialect method. Otherwise it'll be easier to go out of sync if other dialects support it

src/parser/mod.rs
```diff
@@ -2328,19 +2327,25 @@ impl<'a> Parser<'a> {
...
-    fn parse_bigquery_struct_literal(&mut self) -> Result<Expr, ParserError> {
-        let (fields, trailing_bracket) =
-            self.parse_struct_type_def(Self::parse_struct_field_def)?;
-        if trailing_bracket.0 {
-            return parser_err!("unmatched > in STRUCT literal", self.peek_token().location);
+    fn parse_struct_literal(&mut self) -> Result<Expr, ParserError> {
+        let mut fields = vec![];
+        // Typed struct syntax is only supported by BigQuery
+        // https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#typed_struct_syntax
+        if self.dialect.supports_typed_struct_syntax() {
```
iffyio, line 2341: Oh actually I'm not sure i follow the changes entirely, it seems Bigquery struct syntax is a superset of databricks struct? If so I'm thinking we ideally wouldnt need the extra `self.dialect.supports_typed_struct_syntax()` method, only guarding on `supports_struct_literal` should be enough?

It would technically mean that the typed syntax gets accepted by the databricks dialect but thats fine for the parser behavior

tests/sqlparser_databricks.rs
```diff
@@ -277,3 +277,48 @@ fn parse_use() {
         );
     }
 }
+
+#[test]
+fn parse_databricks_struct_function() {
+    assert_eq!(
+        databricks()
```
iffyio, line 284:
```suggestion
        databricks_and_generic()
```
thinking since generic accepts the same syntax? (same for the others)

src/parser/mod.rs
ayman-sigma: You are right, BigQuery struct is a superset of the Databricks struct. I wasn't sure if `supports_typed_struct_syntax` is too much or not. It sounds odd though that we would accept the syntax in Databricks even if it is invalid for this dialect. But I'm happy to change it if that is acceptable behavior for the parser.

tests/sqlparser_databricks.rs
ayman-sigma: Yes, these are already tested with BigQuery, but I think it wouldn't hurt to add it here also.

src/parser/mod.rs
ayman-sigma: done as suggested.

iffyio (approved):
LGTM! Thanks @ayman-sigma!
cc @alamb

alamb:
Thanks @ayman-sigma and @iffyio

Issue #1545, opened 2024-11-24
support postgres `json_object` syntax


lovasoa:
Postgresql supports the following syntax for the json_object function:

```sql
json_object('label' VALUE 'value')
json_object('label': 'value')
```

Reference: https://www.postgresql.org/docs/current/functions-json.html

Related: the `key: value` syntax was already implemented for mssql by @gaoqiangz in  https://github.com/apache/datafusion-sqlparser-rs/pull/1507 . We probably just need to enable it (+ a new VALUE separator) for postgres too

https://github.com/apache/datafusion-sqlparser-rs/issues/1506

Initially reported in https://github.com/sqlpage/SQLPage/issues/708

PR #1546, opened 2024-11-24
support `json_object('k':'v')` in postgres

lovasoa:
This just updates the postgres dialect definition to add support for the json_object syntax added in https://github.com/apache/datafusion-sqlparser-rs/pull/1507

The syntax was initially added for mssql but also works in postgres

See https://www.postgresql.org/docs/current/functions-json.html

See https://github.com/apache/datafusion-sqlparser-rs/issues/1545

support for `json_object('k' VALUE 'v')` is implemented [in a separate PR](https://github.com/apache/datafusion-sqlparser-rs/pull/1547)

PR #1547, opened 2024-11-24
`json_object('k' VALUE 'v')` in postgres

lovasoa:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1545

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -11478,6 +11478,9 @@ impl<'a> Parser<'a> {
     }
 
     fn parse_function_named_arg_operator(&mut self) -> Result<FunctionArgOperator, ParserError> {
+        if self.parse_keyword(Keyword::VALUE) {
+            return Ok(FunctionArgOperator::Value);
+        }
         let tok = self.next_token();
         match tok.token {
             Token::RArrow if self.dialect.supports_named_fn_args_with_rarrow_operator() => {
@@ -11495,9 +11498,6 @@ impl<'a> Parser<'a> {
             }
             Token::Colon if self.dialect.supports_named_fn_args_with_colon_operator() => {
                 Ok(FunctionArgOperator::Colon)
-            }
-            Token::Word(w) if w.value.eq_ignore_ascii_case("value") => {
-                Ok(FunctionArgOperator::Value)
             }
             _ => {
                 self.prev_token();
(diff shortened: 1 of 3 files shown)


lovasoa:
The new test will fail until https://github.com/apache/datafusion-sqlparser-rs/pull/1546 is merged

src/parser/mod.rs
```diff
@@ -11411,6 +11411,9 @@ impl<'a> Parser<'a> {
             Token::Colon if self.dialect.supports_named_fn_args_with_colon_operator() => {
                 Ok(FunctionArgOperator::Colon)
             }
+            Token::Word(w) if w.value.eq_ignore_ascii_case("value") => {
```
iffyio, line 11414: would it make sense to have value be a keyword so that this can potentially become `if self.parse_keyword(Keyword::VALUE)`?

tests/sqlparser_postgres.rs
```diff
@@ -2812,6 +2812,19 @@ fn test_json() {
     );
 }
 
+#[test]
+fn test_json_object() {
```
iffyio, line 2816:
```suggestion
fn test_fn_arg_with_value_operator() {
```
thinking or something more generic? since the feature can apply in other cases and json_object has other variants

src/parser/mod.rs
lovasoa: Yes, I guess so, but I think this is out of the scope of this PR...

tests/sqlparser_postgres.rs
lovasoa: I applied your suggestion. But this feature really should not apply to other functions. The syntax `'key' VALUE 'value'` is specific to json_object and will not be parsed by actual databases in arbitrary functions. The decision to parse this as a generic function call instead of a particular expression was taken in https://github.com/apache/datafusion-sqlparser-rs/pull/1507#discussion_r1835440360 , so I did not come back on the design.

src/parser/mod.rs
iffyio: hmm not sure I follow by it being out of scope. Took a quick look and VALUE is also already a keyword so that the PR doesn't need to touch that file either. thinking the changes required in the pr could be something like this to use the convention vs comparing the raw string?
```rust
if self.parse_keyword(Keyword::VALUE) {
    return Ok(FunctionArgOperator::Value)
}
```

iffyio (approved):
LGTM! cc @alamb

alamb:
Thanks again @iffyio  and @lovasoa

PR #1551, opened 2024-11-25
Consolidate `MapAccess`, and `Subscript` into `CompoundExpr` to handle the complex field access chain 

### LINKED ISSUE #1533: Improve the complex field access chain for the struct type

goldmedal:
# Description
close #1533

This PR introduces a new Expr, `Expr::CompoundExpr`, which handles the complex field access chain like `a.b[1].c` or `func()[1]`. As the previous discussion https://github.com/apache/datafusion-sqlparser-rs/pull/1541#discussion_r1861002348, I found we have many similar expressions for the similar purpose (e.g. `MapAccess`, `Subscript`, `JsonAcess`, ...). `CompoundExpr` tries to represent a structure for all of them. As the suggestion from @iffyio (https://github.com/apache/datafusion-sqlparser-rs/pull/1541#discussion_r1860879446), `CompoundExpr` is a linear structure, it has a root expression and a list of access fields, instead of a nested representation.
```rust
    /// Multi-part Expression accessing. It's used to represent an access chain from a root expression.
    /// e.g. `expr[0]`, `expr[0][0]`, or `expr.field1.filed2[1].field3`, ...
    CompoundExpr {
        root: Box<Expr>,
        chain: Vec<AccessField>,
    },
```

To avoid too many breaking changes, this PR only replaces `Expr::MapAccess` and `Expr::Subscript` with `Expr::CompoundExpr`. Potentially, it can represent more things.

## Array
A 1-dim array `a[1]` will be represented like
```
CompoundExpr(Ident('a'), vec![Subscript(1)]
```
A 2-dim array `a[1][2]` will be represented like
```
CompoundExpr(Ident('a'), vec![Subscript(1), Subscript(2)]
```

## Map or Struct (Bracket-style)
A map `a['fied1']` will be represented like
```
CompoundExpr(Ident('a'), vec![Subscript('field')]
```
A 2-dim map `a['field1']['field2']` will be represented like
```
CompoundExpr(Ident('a'), vec![Subscript('field1'), Subscript('field2')]
```

## Struct (Period-style) (only effect when the chain contains both `subscript` and `expr`)
A struct access `a[field1].field2` will be represented
```
CompoundExpr(Ident('a'), vec![Subscript('field1'), Ident('field2')]
```
If a struct access likes `a.field1.field2`, it will be represented by `CompoundIdentifer([a, field1, field2])`

# Breaking Change
- `Expr::Subscript` and `Expr::MapAccess` are removed.
- The result of the array, map, or struct will be changed.

# Additional context
- https://github.com/apache/datafusion-sqlparser-rs/pull/1541#issuecomment-2490464568


goldmedal:
@iffyio This PR is ready for review. Could you take a look at it? Thanks 🙇 
This PR involves some breaking changes to the array syntax. I'd appreciate it if @alamb could check it as well. 🙇

src/parser/mod.rs
```diff
@@ -1427,6 +1426,112 @@ impl<'a> Parser<'a> {
...
+                }
+            }
+        }
+
+        // if dialect supports partiql, we need to go back one Token::LBracket for the JsonAccess parsing
+        if self.dialect.supports_partiql() && ending_lbracket {
+            self.prev_token();
+        }
+
+        if let Some(wildcard_token) = ending_wildcard {
```
iffyio, line 1480: I'm wondering if it is required that `parse_compound_expr` track the wildcard and other cases that follow. I imagined the behavior somewhat similar to what the doc comment suggests, in that the method only looks to parse a chain if one exists (it looks like the function and outer join parsing is duplicated here which doesnt seem ideal).

Would it be feasible to have this function only take in the root and then look to parse only the chain? e.g. roughly
```rust
let mut chain = vec![];
loop {
  match self.peek().token {
    Token::Period => {
      self.next_token(); // `.`
      if let Some(expr) = self.maybe_parse(Self::parse_sub_expr(
          // Guessing a bit offhand that involving precedence here
          // could help avoid the need to flatten afterwards. I _think_ that was
          // the intent with `exprs_to_idents`? idea is that in this context,
          // the dot is essentially an operator similar to `[]`
          self.dialect.prec_value(Percedence::Dot)
      ))? {
        chain.push(AccessField::Expr(expr));
      } else {
        self.prev_token(); // `.`
        break
      }
    }
    Token::LBracket => {
      // We can probably let the subscript method handle the `[` and `]`
      if let Some(subscript) = self.maybe_parse(Self::parse_subscript)? {
        chain.push(AccessField::Subscript(subscript));
      } else {
        break
      }
    }
  }
}
```

src/ast/mod.rs
```diff
@@ -624,6 +590,12 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// Multi-part Expression accessing. It's used to represent an access chain from a root expression.
+    /// e.g. `expr[0]`, `expr[0][0]`, or `expr.field1.filed2[1].field3`, ...
```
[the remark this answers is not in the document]
alamb: I think it would be awesome to update this comment with the information from this PR's description:

Screenshot 2024-12-06 at 4 16 31 PM

src/ast/mod.rs
```diff
@@ -624,6 +590,12 @@ pub enum Expr {
     Identifier(Ident),
     /// Multi-part identifier, e.g. `table_alias.column` or `schema.table.col`
     CompoundIdentifier(Vec<Ident>),
+    /// Multi-part Expression accessing. It's used to represent an access chain from a root expression.
```
alamb, line 593:
```suggestion
    /// Multi-part expression access. 
    ///
    /// This structure represents an access chain in structured / nested types
    /// such as maps, arrays, and lists
```

src/ast/mod.rs
```diff
@@ -624,6 +590,12 @@ pub enum Expr {
...
+    /// e.g. `expr[0]`, `expr[0][0]`, or `expr.field1.filed2[1].field3`, ...
+    CompoundExpr {
```
alamb, line 595: Also, it seems to me that the name `Compound` is somewhat misleading in this context

It isn't a compound expression per se, it seems more like an a compoound access / multi level access

What do you think about names like `CompoundAccess` ?(that would also fit nicely with @iffyio 's suggestion at https://github.com/apache/datafusion-sqlparser-rs/pull/1551/files#r1871788945 for `AccessExpr`. Maybe something like 🤔  


```rust
...
    CompoundAccess {
      root: Box<Expr>,
      accesses: Vec<AccessExprs>
...
```

I am not sure I really like `accesses` (too many `s` 🤔 )

Maybe `CompoundFieldAccess` 🤔

src/ast/mod.rs
```diff
@@ -1094,6 +1053,25 @@ impl fmt::Display for Subscript {
     }
 }
 
+/// The contents inside the `.` in an access chain.
```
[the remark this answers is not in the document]
alamb: Maybe we could add a doc link to `Expr::CompoundExpr` like

```suggestion
/// An element of a [`Expr::CompoundExpr`]

```

src/ast/mod.rs
```diff
@@ -1289,12 +1267,19 @@ impl fmt::Display for Expr {
...
-                write!(f, "{column}{}", display_separated(keys, ""))
-            }
             Expr::Wildcard(_) => f.write_str("*"),
             Expr::QualifiedWildcard(prefix, _) => write!(f, "{}.*", prefix),
             Expr::CompoundIdentifier(s) => write!(f, "{}", display_separated(s, ".")),
+            Expr::CompoundExpr { root, chain } => {
+                write!(f, "{}", root)?;
+                for field in chain {
+                    match field {
+                        AccessField::Expr(expr) => write!(f, ".{}", expr)?,
```
alamb, line 1277: I would have expected the `,` or the `[]` be in the `AccessField` so it can be self-displaying

tests/sqlparser_common.rs
```diff
@@ -2995,6 +2995,31 @@ fn parse_window_function_null_treatment_arg() {
...
+#[test]
+fn test_compound_expr() {
+    let supported_dialects = TestedDialects::new(vec![
+        Box::new(GenericDialect {}),
+        Box::new(DuckDbDialect {}),
+        Box::new(BigQueryDialect {}),
+    ]);
+    let sqls = [
+        "SELECT abc[1].f1 FROM t",
+        "SELECT abc[1].f1.f2 FROM t",
```
alamb, line 3007: this is pretty cool

tests/sqlparser_common.rs
```diff
@@ -10299,20 +10324,38 @@ fn parse_map_access_expr() {
...
                     expr: Expr::Value(number("1")).into(),
                 },
-                syntax: MapAccessSyntax::Bracket,
-            },
-            MapAccessKey {
-                key: call("safe_offset", [Expr::Value(number("2"))]),
-                syntax: MapAccessSyntax::Bracket,
-            },
+            }),
+            AccessField::Subscript(Subscript::Index {
```
alamb, line 10339: this seems like a very reasonable structure

alamb (commented):
Thanks @goldmedal and @iffyio 

I didn't study the parser changes closely, but I reviewed the `Expr` representation and left some comments. 

In terms of the potential downstream impacts (and avoiding regressions), one thing that I suggest to help evaluate the potential downstream impact is of this trying it out in DataFusion and see if it works / passes all the tests; 

You could potentially use https://github.com/apache/datafusion/pull/13546 as a base to test with

src/parser/mod.rs
```diff
@@ -1144,53 +1144,52 @@ impl<'a> Parser<'a> {
...
+                    Ok(Expr::OuterJoin(Box::new(
+                        match <[Ident; 1]>::try_from(id_parts) {
+                            Ok([ident]) => Expr::Identifier(ident),
+                            Err(parts) => Expr::CompoundIdentifier(parts),
+                        },
+                    )))
+                } else {
+                    let mut expr = self.parse_function(ObjectName(id_parts))?;
+                    // consume all period if it's a method chain
+                    if self.dialect.supports_methods() {
```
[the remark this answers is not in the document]
goldmedal: No, if this condition was removed, the test for method parsing will fail:
```
    parse_method_expr
    parse_method_select
```
Because the following `parse_compound_expr` will try to consume all the dots for an expression, we need to parse the method here to avoid consume `(` in `parse_compound_expr`.

src/parser/mod.rs
```diff
@@ -1144,53 +1144,52 @@ impl<'a> Parser<'a> {
...
-                                Err(parts) => Expr::CompoundIdentifier(parts),
-                            },
-                        )))
-                    } else {
-                        self.prev_token();
-                        self.parse_function(ObjectName(id_parts))
+                    let mut fields = vec![];
+                    // if the function returns an array, it can be subscripted
+                    if self.consume_token(&Token::LBracket) {
+                        self.parse_multi_dim_subscript(&mut fields)?;
```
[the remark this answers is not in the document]
goldmedal: You're right. We can remove this parsing.

src/ast/mod.rs
goldmedal: > I am not sure I really like `accesses` (too many `s` 🤔 )
>
If not prefer `accesses`, how about `access_chain` or `access_path`?

> Maybe `CompoundFieldAccess` 🤔
>
`CompoundFieldAccess` is good to me.

src/ast/mod.rs
goldmedal: Indeed. It's better.

src/dialect/snowflake.rs
```diff
@@ -234,6 +234,10 @@ impl Dialect for SnowflakeDialect {
             RESERVED_FOR_IDENTIFIER.contains(&kw)
         }
     }
+
+    fn supports_partiql(&self) -> bool {
+        true
+    }
```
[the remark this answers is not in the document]
goldmedal: I tried to integrate the conditions at 
https://github.com/apache/datafusion-sqlparser-rs/blob/d0fcc06652ba9880622d0ef8b426c809cee752fe/src/parser/mod.rs#L2998-L3000

Then, we can only check `supports_partiql` in `parse_compound_expr`. 🤔 
```
                    if self.consume_token(&Token::LBracket) {
                        if self.dialect.supports_partiql() {
                            ending_lbracket = true;
                            break;
                        } else {
                            self.parse_multi_dim_subscript(&mut chain)?
                        }
                    }
```

Indeed, the name is a little weird for `Snowflake` but I think they mean the same thing 🤔

src/parser/mod.rs
goldmedal: I guess it's hard to do it 🤔  because the structure of `QualifiedWildcard` is
```
    QualifiedWildcard(ObjectName, AttachedToken),
```
It's similar to `CompoundIdentifier` and `CompoundExpr`. All of them will consume all the dots.

goldmedal:
Thanks, @alamb and @iffyio for reviewing. 

> In terms of the potential downstream impacts (and avoiding regressions), one thing that I suggest to help evaluate the potential downstream impact is of this trying it out in DataFusion and see if it works / passes all the tests;
> 
> You could potentially use [apache/datafusion#13546](https://github.com/apache/datafusion/pull/13546) as a base to test with

Sounds good 👍 I'll test this PR based on it tomorrow.

src/parser/mod.rs
iffyio: Ah right indeed `CompoundExpr` is a superset of `CompoundIdentifier` so there likely wouldn't be a way to avoid that `exprs_to_idents` flattening, like in the scenario where a `foo.bar.baz` expr shows up. That case seems reasonable to flatten since we'd want to keep the representations distinct!

The QualifiedWildcard I think could be fine this same loop as above can handle it I imagine if we peek the next token after the period? i.e if we consume a `.*` then we build up the wildcard expr as we do today with the root+chain consumed so far. So that essentially the loop parses the chain delimited by period and the `CompoundIdentifier` vs `QualifiedWildcard` are the same scenarios only differing in the terminating token

goldmedal:
> Thanks, @alamb and @iffyio for reviewing.
> 
> > In terms of the potential downstream impacts (and avoiding regressions), one thing that I suggest to help evaluate the potential downstream impact is of this trying it out in DataFusion and see if it works / passes all the tests;
> > You could potentially use [apache/datafusion#13546](https://github.com/apache/datafusion/pull/13546) as a base to test with
> 
> Sounds good 👍 I'll test this PR based on it tomorrow.

Created https://github.com/apache/datafusion/pull/13734 for testing. The good news is that It only impacts a few parts of the planner and the unparser 👍

alamb (approved):
Thanks @goldmedal  -- this looks good to me. 🙏

alamb:
Given the potential for non trivial downstream conflicts due to this change (look at the list of conflicts it has already collected) I would like to merge it in right after we have done the next release

src/parser/mod.rs
goldmedal: @iffyio 
I feel the idea of this loop is good but I found a problem 🤔 . If we try to invoke `parse_sub_expr` recursively, we will still get the nested representation. Consider the following cases:

## The representation of abc[1].f1.f2
The last field is a `CompoundIdentifier`.
```
&root = Identifier(
    Ident {
        value: "abc",
        quote_style: None,
        span: Span(Location(0,0)..Location(0,0)),
    },
)
&chain = [
    Subscript(
        Index {
            index: Value(
                Number(
                    "1",
                    false,
                ),
            ),
        },
    ),
    Dot(
        CompoundIdentifier(
            [
                Ident {
                    value: "f1",
                    quote_style: None,
                    span: Span(Location(0,0)..Location(0,0)),
                },
                Ident {
                    value: "f2",
                    quote_style: None,
                    span: Span(Location(0,0)..Location(0,0)),
                },
            ],
        ),
    ),
]
```
## The representation of abc[1].f1.f2[1]
The last field is a `CompoundFieldAccess` with another chain.
```
&root = Identifier(
    Ident {
        value: "abc",
        quote_style: None,
        span: Span(Location(0,0)..Location(0,0)),
    },
)
&chain = [
    Subscript(
        Index {
            index: Value(
                Number(
                    "1",
                    false,
                ),
            ),
        },
    ),
    Dot(
        CompoundFieldAccess {
            root: Identifier(
                Ident {
                    value: "f1",
                    quote_style: None,
                    span: Span(Location(0,0)..Location(0,0)),
                },
            ),
            access_chain: [
                Dot(
                    CompoundFieldAccess {
                        root: Identifier(
                            Ident {
                                value: "f2",
                                quote_style: None,
                                span: Span(Location(0,0)..Location(0,0)),
                            },
                        ),
                        access_chain: [
                            Subscript(
                                Index {
                                    index: Value(
                                        Number(
                                            "1",
                                            false,
                                        ),
                                    ),
                                },
                            ),
                        ],
                    },
                ),
            ],
        },
    ),
]
```
It seems we should consider `CompoundIdentifier` and `CompoundFieldAccess` as a kind of the nested struct. 
Ideally, the access_cahin shouldn't contain another nested struct. 
The `parse_subexpr` shouldn't produce the nested struct 🤔. It makes the thing more complicated.
I don't prefer to do It in this PR. The side effects will be too large. Maybe we can have a follow-up PR to refactor here.
WDYT?

src/parser/mod.rs
```diff
@@ -3095,15 +3201,29 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse an multi-dimension array accessing like `[1:3][1][1]`
```
iffyio, line 3204:
```suggestion
    /// Parses a multi-dimension array accessing like `[1:3][1][1]`
```

src/parser/mod.rs
```diff
@@ -3095,15 +3201,29 @@ impl<'a> Parser<'a> {
...
+    ///
+    /// Parser is right after the first `[`
```
iffyio, line 3206: given this is a pub function it would be nice if the api was symmetric so that the function also consumed the leading `[`? I think within the parser itself that would mean calling `prev_token()` or replacing with `peek()` before invoking this function which would be a fair price to pay?

iffyio (commented):
Thanks @goldmedal! Left some minor comments, the changes look good to me!

src/parser/mod.rs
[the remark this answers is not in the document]
goldmedal: Indeed. It's more simple. Thanks!

src/parser/mod.rs
```diff
@@ -3167,42 +3287,23 @@ impl<'a> Parser<'a> {
 
     pub fn parse_map_access(&mut self, expr: Expr) -> Result<Expr, ParserError> {
         let key = self.parse_expr()?;
+        let result = match key {
```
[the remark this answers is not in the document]
goldmedal: You're right. We can remove it entirely.

src/parser/mod.rs
```diff
@@ -1427,6 +1424,112 @@ impl<'a> Parser<'a> {
...
+                return Ok(root);
+            }
+            Ok(Expr::CompoundFieldAccess {
+                root: Box::new(root),
+                access_chain: chain.clone(),
+            })
+        }
+    }
+
+    fn exprs_to_idents(root: &Expr, fields: &[AccessExpr]) -> Option<Vec<Ident>> {
```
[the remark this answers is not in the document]
goldmedal: Nice catch. I think we can save the cloning by separating it into two functions. One for check and another for transfer.

src/parser/mod.rs
goldmedal: Make sense.

goldmedal: fixed by [67cd877](https://github.com/apache/datafusion-sqlparser-rs/pull/1551/commits/67cd8778224439aa2f1ddd6605fe15ed0cf12279)

src/parser/mod.rs
```diff
@@ -1427,6 +1424,112 @@ impl<'a> Parser<'a> {
...
+            };
+            if dialect_of!(self is SnowflakeDialect | MsSqlDialect)
+                && self.consume_tokens(&[Token::Plus, Token::RParen])
+            {
+                Ok(Expr::OuterJoin(Box::new(
+                    match <[Ident; 1]>::try_from(id_parts) {
+                        Ok([ident]) => Expr::Identifier(ident),
+                        Err(parts) => Expr::CompoundIdentifier(parts),
+                    },
+                )))
```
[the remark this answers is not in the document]
goldmedal: fixed by [4b3818c](https://github.com/apache/datafusion-sqlparser-rs/pull/1551/commits/4b3818cfc10cc2a15b985e50a0579b932a9220f8)

src/parser/mod.rs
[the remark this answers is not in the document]
goldmedal: fixed by [5c54d1b](https://github.com/apache/datafusion-sqlparser-rs/pull/1551/commits/5c54d1b26cee61683ab4d382090c80c4ac84a6b3)

iffyio (approved):
LGTM! Thanks @goldmedal!

goldmedal:
> Given the potential for non trivial downstream conflicts due to this change (look at the list of conflicts it has already collected) I would like to merge it in right after we have done the next release
> 
> * [Release sqlparser-rs version `0.53.0` / sqlparser_derive `0.3.0` #1517](https://github.com/apache/datafusion-sqlparser-rs/issues/1517)

@alamb @iffyio, `0.53.0` has been released. Could we merge this PR?
I want to merge it to avoid some potential conflicts.

iffyio:
Thanks for working on this @goldmedal!

goldmedal:
Thanks @iffyio and @alamb 👍

Issue #1550, opened 2024-11-25
JOIN should require ON condition


demetribu:
### Describe the bug

`SELECT ... FROM l JOIN r` is not a valid SQL query, the `ON` condition should not be optional

### Additional context

https://github.com/apache/datafusion/issues/13486

demetribu:
take

demetribu:
Proposed to wait for use case.
https://github.com/apache/datafusion-sqlparser-rs/pull/1552#discussion_r1866317958

PR #1552, opened 2024-11-25
Fix: JOIN should require ON condition

demetribu:
Closes: #1550


mvzink:
MySQL actually does allow joins with no ON/USING, so this could break some (admittedly weird) workflows.

demetribu:
What I found

> In MySQL, JOIN, CROSS JOIN, and INNER JOIN are syntactic equivalents (they can replace each other). In standard SQL, they are not equivalent. INNER JOIN is used with an ON clause, CROSS JOIN is used otherwise. 

https://dev.mysql.com/doc/refman/8.4/en/join.html

cc @findepi

findepi:
so maybe we should fix this in DF only?

demetribu:
as I know, in df, we don’t parse SQL directly; we only work with statements.

Dandandan:
It could be implemented for the dialects that (don't) support the joins without `ON` (i.e. PostgreSQL / ansi, ...)?

iffyio:
Ah yeah using a dialect method for this makes sense in that case

demetribu:
@iffyio It seems the main branch is currently failing Clippy lint checks against Rust 1.83.0.

cc @alamb

src/dialect/mod.rs
```diff
@@ -687,6 +687,41 @@ pub trait Dialect: Debug + Any {
...
+
+    /// Verifies whether the provided `JoinOperator` is supported by this SQL dialect.
+    /// Returns `true` if the `JoinOperator` is supported, otherwise `false`.
+    fn verify_join_operator(&self, _join_operator: &JoinOperator) -> bool {
+        true
+    }
+
+    /// Verifies if the given `JoinOperator`'s constraint is valid for this SQL dialect.
+    /// Returns `true` if the join constraint is valid, otherwise `false`.
+    fn verify_join_constraint(&self, join_operator: &JoinOperator) -> bool {
```
iffyio, line 699: Ah so since the behavior seems to vary a bit across dialects, I'm thinking it could make sense after all if we let the parse continue to be permissive in syntax and downstream crates can perform the additional checks in the cases where specific combinations need to be enforced?

demetribu: I understand your idea, and it makes sense. But the "issue" is based on the idea that the parser should restrict cases like “INNER JOIN” without an ON condition.

For me, as a developer, it would feel strange if I pick a parser, for example, the PostgreSQL dialect, and it allows “ANTI JOIN” without any error. This would mean I have to check all the statements again to find mistakes. It seems like a trade-off, and we need to decide which approach works best.

iffyio: Yeah it is indeed a tradeoff, the expectation currently is that downstream crates further validate the output of the parser against any dialect specific requirements/invariants, see [note here for example](https://github.com/apache/datafusion-sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics)

demetribu: Currently, DataFusion processes `SELECT ... FROM l JOIN r` as a `CROSS JOIN` for all dialects, including DuckDB, where an `INNER JOIN` explicitly requires an `ON `clause.

@Dandandan proposed addressing this issue in `sqlparser-rs`, which seems reasonable to me. However, after discussing it with @iffyio, I see

> indeed a tradeoff, the expectation currently is that downstream crates further validate the output of the parser

My questions are:
1. Should this behavior be considered a bug?
2. If yes, at which level should it be addressed: `datafusion` or `sqlparser-rs`?

alamb: > * Should this behavior be considered a bug?

I don't have a strong opinion -- is it causing anyone problems? It seems like the ramification of allowing `JOIN`... as `CROSS JOIN` 's largest implication is that now DataFusion has some new dialect. 

If it isn't causing problems, my suggestion is do nothing until someone has a concrete usecase

alamb:
Converting to draft to show we aren't necessairly planning to merge this

PR #1555, opened 2024-11-26
Document micro benchmarks

alamb:
@Dandandan  asked about benchmarks on https://github.com/apache/datafusion-sqlparser-rs/pull/1435#issuecomment-2500664144

There is actually a parser benchmark but I didn't even know it existed. Let's document it a bit so it may be more discoverable.


Dandandan (approved):

alamb:
Thanks for the review @Dandandan

PR #1556, opened 2024-11-26
Support relation visitor to visit the `Option` field

### LINKED ISSUE #1554: Relation visitor fails to visit the `SHOW COLUMNS` statement in the latest commit of the main branch
# Description
This problem can be reproduced by adding a test case at
https://github.com/apache/datafusion-sqlparser-rs/blob/525d1780e8f5c7ba6b7be327eaa788b6c8c47716/src/ast/visitor.rs#L736-L738

Consider a case
```rust
            (
                "SHOW COLUMNS FROM t1",
                vec![
                    "PRE: STATEMENT: SHOW COLUMNS FROM t1",
                    "PRE: RELATION: t1",
                    "POST: RELATION: t1",
                    "POST: STATEMENT: SHOW COLUMNS FROM t1",
                ],
            ),
```

The test case is passed in the previous release tag `v0.52.0-rc3`. However, in the latest commit of the main branch 525d1780e8f5c7ba6b7be327eaa788b6c8c47716 , the test fails and the result is
```rust
     "PRE: STATEMENT: SHOW COLUMNS FROM t1",
     "POST: STATEMENT: SHOW COLUMNS FROM t1",
```

It may cause the upstream project, DataFusion, to fail to execute the `SHOW COLUMNS FROM xxx` SQL.
https://github.com/apache/datafusion/blob/18fc103a403ab0efe5245dd4352f3f3b93c2a4fe/datafusion/core/src/execution/session_state.rs#L540

goldmedal:
close #1554 

If the field is a `Option` and add `#[with = "visit_xxx"]` to the field, the generated code
will try to access the field only if it is `Some`:

```rust
pub struct ShowStatementIn {
    pub clause: ShowStatementInClause,
    pub parent_type: Option<ShowStatementInParentType>,
    pub parent_name: Option<ObjectName>,
}
```

This will generate

```rust
impl sqlparser::ast::Visit for ShowStatementIn {
    fn visit<V: sqlparser::ast::Visitor>(
        &self,
        visitor: &mut V,
    ) -> ::std::ops::ControlFlow<V::Break> {
        sqlparser::ast::Visit::visit(&self.clause, visitor)?;
        sqlparser::ast::Visit::visit(&self.parent_type, visitor)?;
        if let Some(value) = &self.parent_name {
            visitor.pre_visit_relation(value)?;
            sqlparser::ast::Visit::visit(value, visitor)?;
            visitor.post_visit_relation(value)?;
        }
        ::std::ops::ControlFlow::Continue(())
    }
}

impl sqlparser::ast::VisitMut for ShowStatementIn {
    fn visit<V: sqlparser::ast::VisitorMut>(
        &mut self,
        visitor: &mut V,
    ) -> ::std::ops::ControlFlow<V::Break> {
        sqlparser::ast::VisitMut::visit(&mut self.clause, visitor)?;
        sqlparser::ast::VisitMut::visit(&mut self.parent_type, visitor)?;
        if let Some(value) = &mut self.parent_name {
            visitor.pre_visit_relation(value)?;
            sqlparser::ast::VisitMut::visit(value, visitor)?;
            visitor.post_visit_relation(value)?;
        }
        ::std::ops::ControlFlow::Continue(())
    }
}
```

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -18,11 +18,8 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::spanned::Spanned;
-use syn::{
-    parse::{Parse, ParseStream},
-    parse_macro_input, parse_quote, Attribute, Data, DeriveInput, Fields, GenericParam, Generics,
-    Ident, Index, LitStr, Meta, Token,
-};
+use syn::{parse::{Parse, ParseStream}, parse_macro_input, parse_quote, Attribute, Data, DeriveInput, Fields, GenericParam, Generics, Ident, Index, LitStr, Meta, Token, Type, TypePath};
+use syn::{Path, PathArguments};
 
 /// Implementation of `[#derive(Visit)]`
 #[proc_macro_derive(VisitMut, attributes(visit))]
@@ -182,9 +179,21 @@ fn visit_children(
             Fields::Named(fields) => {
                 let recurse = fields.named.iter().map(|f| {
                     let name = &f.ident;
+                    let is_option = is_option(&f.ty);
                     let attributes = Attributes::parse(&f.attrs);
-                    let (pre_visit, post_visit) = attributes.visit(quote!(&#modifier self.#name));
-                    quote_spanned!(f.span() => #pre_visit sqlparser::ast::#visit_trait::visit(&#modifier self.#name, visitor)?; #post_visit)
+                    if is_option && attributes.with.is_some() {
+                        let (pre_visit, post_visit) = attributes.visit(quote!(value));
+                        quote_spanned!(f.span() =>
+                            if let Some(value) = &#modifier self.#name {
+                                #pre_visit sqlparser::ast::#visit_trait::visit(value, visitor)?; #post_visit
+                            }
+                        )
+                    } else {
+                        let (pre_visit, post_visit) = attributes.visit(quote!(&#modifier self.#name));
+                        quote_spanned!(f.span() =>
+                            #pre_visit sqlparser::ast::#visit_trait::visit(&#modifier self.#name, visitor)?; #post_visit
+                        )
+                    }
                 });
                 quote! {
                     #(#recurse)*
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -7653,6 +7653,7 @@ impl fmt::Display for ShowStatementInParentType {
 pub struct ShowStatementIn {
     pub clause: ShowStatementInClause,
     pub parent_type: Option<ShowStatementInParentType>,
+    #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
```
alamb, line 7656: I took a look through the code, and it appears there are oehter fields that hold relation names that are not visited

For example:
https://github.com/alamb/sqlparser-rs/blob/a8432b57db6b8efc635e4903d712e76f7adb6e6d/src/ast/ddl.rs#L712-L713

https://github.com/apache/datafusion-sqlparser-rs/blob/4c629e8520b68eb289b34882aa326d80a6f8e022/src/ast/dml.rs#L580-L581

https://github.com/apache/datafusion-sqlparser-rs/blob/3c8fd748043188957d2dfcadb4bfcfb0e1f70c82/src/ast/mod.rs#L2432-L2431

https://github.com/apache/datafusion-sqlparser-rs/blob/3c8fd748043188957d2dfcadb4bfcfb0e1f70c82/src/ast/mod.rs#L2495-L2494

https://github.com/apache/datafusion-sqlparser-rs/blob/3c8fd748043188957d2dfcadb4bfcfb0e1f70c82/src/ast/mod.rs#L2666-L2665

Maybe we can file a ticket to annotate these ones too

derive/src/lib.rs
```diff
@@ -256,3 +265,16 @@ fn visit_children(
         Data::Union(_) => unimplemented!(),
     }
 }
+
+fn is_option(ty: &Type) -> bool {
+    if let Type::Path(TypePath { path: Path { segments, .. }, .. }) = ty {
+        if let Some(segment) = segments.last() {
+            if segment.ident == "Option" {
```
alamb, line 272: it seems in theory this would match anything called `Option` (even if it wasn't `std::Option`) but I think that seems ok to me (I don't think we'll have anything else realistically)

alamb (approved):
Thank you @goldmedal  . I think there were some other fields that have some wrapping structure simply because this feature was missing

I think we will also have to release a new version of sqlparser derive (which we don't always do) as part of this release given the change to the proc macro; I left a note to remind us here

I will also test this with DataFusion as well and report back

derive/src/lib.rs
goldmedal: Agreed, I think we won't have another struct called `Option` in this project 🤔. Or we can match the full path with `std::option::Option` or `core::option::Option` if required.

iffyio (approved):
Thanks @goldmedal!

alamb:
I double checked with https://github.com/apache/datafusion/pull/13546 and this works great. Thank you again @goldmedal and @iffyio

Issue #1558, opened 2024-11-26
Improve performance by not copying `Token`s as much


alamb:
Part of https://github.com/apache/datafusion-sqlparser-rs/issues/1557

While looking at the flamegraph posted to that ticket, one obvious source of improvement is to stop copying each token so much.

Functions like `peek_token()`, and `expect_token()` copy the `Token` (which often includes a string)

```rust
impl Parser {
    // returns an OWNED TokenWithLocation
    pub fn peek_token(&self) -> TokenWithLocation {
   ...
    }
}
```

For example
https://github.com/apache/datafusion-sqlparser-rs/blob/2e90e105a74bf9f50f2bad6c22992759ddb06880/src/parser/mod.rs#L3314-L3334

In the above code, the `non_whitespace.cloned()` call actually copies the token (and string) even when many callsites only need to check what it is and could get by with a &

## Suggestions for improvements

The biggest suggestion is to stop copying each token unless it actually needed a clone. For example instead of this

```rust
impl Parser {
    // returns an OWNED TokenWithLocation
    pub fn peek_token(&self) -> TokenWithLocation {
   ...
    }
}
```

Make it like this

```rust
impl Parser {
    // returns an reference to TokenWithLocation
    // (the & means no copying!)
    pub fn peek_token_ref(&self) -> &TokenWithLocation {
   ...
    }
}
```

I think we could do this without massive breaking changes by adding new functions like `peek_token_ref()` that returned a reference to the token rather than a clone

## Ideas

I played around with it a bit and here is what I came up with. I think a similar pattern could be applied to other places

```rust
impl Parser {
...
    /// Return the first non-whitespace token that has not yet been processed
    /// (or None if reached end-of-file) and mark it as processed. OK to call
    /// repeatedly after reaching EOF.
    pub fn next_token(&mut self) -> TokenWithLocation {
        self.next_token_ref().clone()
    }

    /// Return the first non-whitespace token that has not yet been processed
    /// (or None if reached end-of-file) and mark it as processed. OK to call
    /// repeatedly after reaching EOF.
    pub fn next_token_ref(&mut self) -> &TokenWithLocation {
        loop {
            self.index += 1;
            // skip whitespace
            if let Some(TokenWithLocation {
                token: Token::Whitespace(_),
                location: _,
            }) = self.tokens.get(self.index - 1) {
                continue
            }
            break;
        }
        if (self.index - 1) < self.tokens.len() {
[...]

PR #1561, opened 2024-11-26
POC to show performance improvements of not copying token

alamb:
This PR was hoping to show that the approach described in this ticket would be effective
- https://github.com/apache/datafusion-sqlparser-rs/issues/1558

While I had this all loaded into my head, I wanted to bash out a PR / POC, but ran out of time. 

This PR shows how we could avoid copying tokens 

Goals;
- [x] Sketch out `peek_token_ref()` and similar functions
- [ ] Use those functions to rewrite the hot paths shown in the flamegraph (`parse_prefix`, etc. as shown below)
- [ ] Demonstrate improved performance with the benchmark 

[fixed-flamegraph](https://github.com/user-attachments/assets/7ceae3a0-dec2-462d-9a00-0b49afd9d2bc)

fixed-flamegraph

It will take some restructring of `parse_prefix` to avoid the use of the copying APIs, but I absolutely think it is possible

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -27,6 +27,7 @@ use core::{
 use helpers::attached_token::AttachedToken;
 
 use log::debug;
+use std::sync::OnceLock;
 
 use recursion::RecursionCounter;
 use IsLateral::*;
@@ -3328,7 +3329,12 @@ impl<'a> Parser<'a> {
     }
 
     /// Return nth non-whitespace token that has not yet been processed
-    pub fn peek_nth_token(&self, mut n: usize) -> TokenWithLocation {
+    pub fn peek_nth_token(&self, n: usize) -> TokenWithLocation {
+        self.peek_nth_token_ref(n).clone()
+    }
+
+    /// Return nth non-whitespace token that has not yet been processed
+    pub fn peek_nth_token_ref(&self, mut n: usize) -> &TokenWithLocation {
         let mut index = self.index;
         loop {
             index += 1;
@@ -3339,10 +3345,11 @@ impl<'a> Parser<'a> {
                 }) => continue,
                 non_whitespace => {
                     if n == 0 {
-                        return non_whitespace.cloned().unwrap_or(TokenWithLocation {
-                            token: Token::EOF,
-                            span: Span::empty(),
-                        });
+                        if let Some(tok) = non_whitespace {
+                            return tok;
+                        } else {
+                            return eof_token();
+                        }
                     }
                     n -= 1;
                 }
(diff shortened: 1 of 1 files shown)


davisp:
https://github.com/apache/datafusion-sqlparser-rs/compare/main...davisp:datafusion-sqlparser-rs:pd/experiment/less-token-cloning

tl;dr - Roughly 28% speedup on the microbenchmarks compared to main. 18% of that by tweaking a few of the main token methods in the last commit.

I ended up poking at this today and have managed to get about a 28% improvement on the benchmarks over `main`. Roughly 10% of that is checking for hot paths in the flame graphs and slowly converting those functions over to avoid cloning tokens.

On a whim, I also poked at optimizing the Token::make_word method that does a binary search over all keywords. That appears to have saved about 400ms (of 1.2s total originally attributed to it). I skimmed the docs for [phf](https://crates.io/crates/phf) but I'm not certain what this project's appetite is for new dependencies. It currently seems quite lean so I didn't pursue that just yet.

Lastly, the biggest single commit was the latest that updates a few of the token handling methods themselves to avoid clones. This was the biggest jump at 18% improvement for a total just around 28% (based on my very non-scientific measurements).

Also, one thing I did different than you @alamb, I noted CI flagged your EOF_TOKEN, so I just made mine a static since it doesn't have any initialization.

All in all, this approach certainly seems feasible as well as actually productive. It turns out the hardest part is accounting for error reporting with the `expected("message", token)` API. I ended up adding an `expected_current` which doesn't take a token parameter and uses the current index for reporting. That works for now, but going forward I think we're gonna want to figure out a much different error reporting API as its quite fragile (you have to pay attention to which token is being reported and whether or not the token index has been modified). There are also a number of places where the token being reported in errors is incorrect.

alamb:
> On a whim, I also poked at optimizing the Token::make_word method that does a binary search over all keywords. That appears to have saved about 400ms (of 1.2s total originally attributed to it). I skimmed the docs for [phf](https://crates.io/crates/phf) but I'm not certain what this project's appetite is for new dependencies. It currently seems quite lean so I didn't pursue that just yet.

I agree we try to keep the dependency set down to a minimum. \

I wonder if we could automatically generate a jumptable with some trickery (maybe we could make a custom match block keyed on individual letters)L

Something like
```rust
match word[0] {
  'a' | 'A' => match word[1] {
    'b'| 'B' = > //match the third letter here
```

We could then make a script, etc to generate this jumptable from the list of keywords 🤔

davisp:
> Would you be willing to start making some smaller PRs with parts of your findings?

I've opened #1587 for the token cloning work and #1588 for investigating ways to speed up `Token::make_word`. Looks like I broke an error message though so time to fix that.

alamb:
> > Would you be willing to start making some smaller PRs with parts of your findings?
> 
> I've opened #1587 for the token cloning work and #1588 for investigating ways to speed up `Token::make_word`. Looks like I broke an error message though so time to fix that.

AMAZING!!! Thank you

PR #1564, opened 2024-11-28
Support MySQL size variants for BLOB and TEXT columns

mvzink:


iffyio (approved):
LGTM! cc @alamb

alamb:
Thanks @mvzink and @iffyio

PR #1566, opened 2024-11-28
feat: Supporting `SAMPLE` parsing

seve-martinez:
Snowflake, Databricks, Postgres, and others all support some from of table sampling. 

This commonly takes on the form

```sql
SELECT ...
FROM ... 
{ SAMPLE | TABLESAMPLE } [ samplingMethod ]
```

where 

```console
samplingMethod ::= { { BERNOULLI | ROW } ( { <probability> | <num> ROWS } ) |
                     { SYSTEM | BLOCK } ( <probability> ) [ { REPEATABLE | SEED } ( <seed> ) ] }
```

Different dialects support one, the other, or both sample keywords and only some of the methods. The current PR follows the snowflake paradigm.

## Outstanding questions
- [ ] This method is not currently called anywhere. Review of the `Parser` code would make me think it should be part of the `parse_table_factor()` logic since it always trails a relation. However this would then require each variant of `TableFactor` that supports `SAMPLE` to have a placeholder for it. Is that the right approach?


iffyio:
Hi @seve-martinez! It sounds like the parsing can be done while parsing the select body instead? Probably [here before looking for a WHERE clause](https://github.com/apache/datafusion-sqlparser-rs/blob/d0fcc06/src/parser/mod.rs#L9599)?

yoavcloud:
I did not notice this PR when I started working on parsing the SAMPLE option, but since then it was merged: https://github.com/apache/datafusion-sqlparser-rs/pull/1580

I can share that we used to parse the SAMPLE option as part of the TableFactor but it wasn't ideal, and moving it to the select made more sense.

seve-martinez:
> I did not notice this PR when I started working on parsing the SAMPLE option, but since then it was merged: #1580
> 
> I can share that we used to parse the SAMPLE option as part of the TableFactor but it wasn't ideal, and moving it to the select made more sense.

Oh awesome! Thanks for taking care of it, I wasn't in love with my solution and what you have there looks very sound. 👍

PR #1571, opened 2024-11-29
Increase version of sqlparser_derive from 0.2.2 to 0.3.0

alamb:
- Part of  https://github.com/apache/datafusion-sqlparser-rs/issues/1517

https://github.com/apache/datafusion-sqlparser-rs/pull/1556 from @goldmedal requires a newer version of sqlparser-derive (in order to visit `Option`) so I would like to increase the version now so I don't forget to do so before releasing


iffyio (approved):
LGTM!

alamb:
Thank you for the review @iffyio

PR #1574, opened 2024-11-30
Add support of the ENUM8|ENUM16 for ClickHouse dialect

git-hulk:
For the documentation, please refer to:

https://clickhouse.com/docs/en/sql-reference/data-types/enum


src/parser/mod.rs
```diff
@@ -1049,18 +1049,18 @@ impl<'a> Parser<'a> {
             | Keyword::CURRENT_USER
             | Keyword::SESSION_USER
             | Keyword::USER
-                if dialect_of!(self is PostgreSqlDialect | GenericDialect) =>
```
git-hulk, line 1052: The following parts are auto formatted by the command: `cargo format`.

git-hulk:
cc @iffyio

src/parser/mod.rs
```diff
@@ -7979,6 +7979,27 @@ impl<'a> Parser<'a> {
...
+        Ok(values)
+    }
+
+    pub fn parse_enum_value(&mut self) -> Result<EnumValue, ParserError> {
+        let str = self.parse_literal_string()?;
+        let value = match self.peek_token().token {
+            Token::Eq => {
+                // Consume the `=` token
+                self.next_token();
+                let value = self.parse_number_value()?;
```
iffyio, line 7995: it seems from the docs the value can be any integer whereas `parse_number_value()` doesn't support negative/signed integers, maybe we'd want to use `parse_number()` here instead?

src/parser/mod.rs
```diff
@@ -7979,6 +7979,27 @@ impl<'a> Parser<'a> {
...
+    pub fn parse_enum_values(&mut self) -> Result<Vec<EnumValue>, ParserError> {
+        self.expect_token(&Token::LParen)?;
+        let values = self.parse_comma_separated(Parser::parse_enum_value)?;
+        self.expect_token(&Token::RParen)?;
+        Ok(values)
+    }
+
+    pub fn parse_enum_value(&mut self) -> Result<EnumValue, ParserError> {
+        let str = self.parse_literal_string()?;
+        let value = match self.peek_token().token {
```
iffyio, line 7991: This can be simplified as
```rust
if self.consume_token(&Token::Eq) {
    Enum::Pair()
} else {
    Enum::String()
}
```
one other more of a nit, since the method is small enough and if its only used once, it could probably be inlined within the `parse_enum_values()`

src/ast/data_type.rs
```diff
@@ -300,7 +311,7 @@ pub enum DataType {
     /// [clickhouse]: https://clickhouse.com/docs/en/sql-reference/data-types/nested-data-structures/nested
     Nested(Vec<ColumnDef>),
     /// Enums
-    Enum(Vec<String>),
+    Enum(Vec<EnumValue>, Option<i64>),
```
iffyio, line 314:
```suggestion
    Enum(Vec<EnumValue>, Option<i64>),
```
is a negative size allowed (wondering if this should be u64 otherwise)?

src/ast/data_type.rs
```diff
@@ -25,10 +25,21 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum EnumValue {
+    String(String),
+    /// ClickHouse allows to specify an integer value for each enum value.
+    ///
+    /// [clickhouse](https://clickhouse.com/docs/en/sql-reference/data-types/enum)
+    Pair(String, Value),
+}
```
iffyio, line 41:
```suggestion
pub enum EnumMember {
    Name(String),
    /// ClickHouse allows to specify an integer value for each enum value.
    ///
    /// [clickhouse](https://clickhouse.com/docs/en/sql-reference/data-types/enum)
    NamedValue(String, Value),
}
```
Would something like this be reasonable? thinking member is convention and the variant would be a bit more descriptive

src/dialect/clickhouse.rs
```diff
@@ -50,4 +50,10 @@ impl Dialect for ClickHouseDialect {
     fn supports_limit_comma(&self) -> bool {
         true
     }
+
+    /// ClickHouse supports `Enum8` and `Enum16` types.
+    /// See [ClickHouse](https://clickhouse.com/docs/en/sql-reference/data-types/enum)
+    fn supports_enum_type_with_bits(&self) -> bool {
+        true
+    }
```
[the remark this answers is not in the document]
git-hulk: I think it's good to allow this in all dialects. Will remove this.

src/parser/mod.rs
git-hulk: Good point.

src/ast/data_type.rs
git-hulk: Yes, That's a good suggestion.

src/parser/mod.rs
git-hulk: Has been Changed to `parse_number`. I initially think the enum value won't be a negative value.

src/parser/mod.rs
alamb: I wonder why it just happened now 🤔  Is it possible you are using nightly rust 🤔

alamb (commented):
Thanks @git-hulk

src/parser/mod.rs
git-hulk: @alamb aha, I'm using 1.82.0 on my side.

tests/sqlparser_common.rs
```diff
@@ -12440,3 +12441,83 @@ fn test_reserved_keywords_for_identifiers() {
...
     dialects.parse_sql_statements(sql).unwrap();
 }
+
+#[test]
+fn parse_create_table_with_enum_types() {
+    let sql = "CREATE TABLE t0 (foo ENUM8('a' = 1, 'b' = 2), bar ENUM16('a' = 1, 'b' = 2), baz ENUM('a', 'b'))";
+    match all_dialects().verified_stmt(sql) {
+        Statement::CreateTable(CreateTable { name, columns, .. }) => {
+            std::assert_eq!(name.to_string(), "t0");
+            std::assert_eq!(
```
iffyio, line 12451: oh for the asserts was the `std::assert_eq` auto imported or was it a need to use them specifically for the tests?

iffyio (commented):
Thanks @git-hulk! The changes look good to me, just one minor comment/question on the assert import, also double checking the formatting changes sounded from the thread that they were  local and need a follow up?

tests/sqlparser_common.rs
git-hulk: Yes, `std::` is unnecessary here. 👍🏿

git-hulk:
> also double checking the formatting changes sounded from the thread that they were local and need a follow up?

It seems the fmt behavior changed in the newer version. To see if @alamb has any comments.

alamb (approved):
Looks good to me - thanks @git-hulk

iffyio (approved):
LGTM!

PR #1576, opened 2024-12-03
Redshift: Fix parsing for quoted numbered columns

7phs:
Quoted identifiers allow the use of any characters and can start with any character, including digits. For example:

```sql
SELECT 1 AS "1" FROM a
```

However, the tokenizer cannot parse such types of quoted identifiers in the case of Redshift. This issue is related to the handling of potential identifiers in PartiQL.

This merge request (MR) clarifies the behavior of quoted identifiers in PartiQL and fixes the parsing of quoted identifiers for Redshift.

I added an additional case of a valid JSON path in a specific unit test to ensure that JSON path parsing is not broken.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -128,16 +128,22 @@ pub trait Dialect: Debug + Any {
         ch == '"' || ch == '`'
     }
 
-    /// Return the character used to quote identifiers.
-    fn identifier_quote_style(&self, _identifier: &str) -> Option<char> {
-        None
+    /// Determine if a character starts a potential nested quoted identifier.
+    /// RedShift support old way of quotation with `[` and it can cover even nested quoted identifier.
+    fn is_nested_delimited_identifier_start(&self, _ch: char) -> bool {
+        false
     }
 
-    /// Determine if special way quoted characters are presented
-    fn special_delimited_identifier_start(
+    /// Determine if nested quoted characters are presented
+    fn nested_delimited_identifier(
         &self,
         mut _chars: Peekable<Chars<'_>>,
     ) -> Option<(char, Option<char>)> {
+        None
+    }
+
+    /// Return the character used to quote identifiers.
+    fn identifier_quote_style(&self, _identifier: &str) -> Option<char> {
         None
     }
 
@@ -851,19 +857,23 @@ mod tests {
                 self.0.is_delimited_identifier_start(ch)
             }
 
+            fn is_nested_delimited_identifier_start(&self, ch: char) -> bool {
+                self.0.is_nested_delimited_identifier_start(ch)
+            }
+
+            fn nested_delimited_identifier(
+                &self,
+                chars: std::iter::Peekable<std::str::Chars<'_>>,
+            ) -> Option<(char, Option<char>)> {
+                self.0.nested_delimited_identifier(chars)
+            }
+
             fn identifier_quote_style(&self, identifier: &str) -> Option<char> {
                 self.0.identifier_quote_style(identifier)
             }
 
             fn supports_string_literal_backslash_escape(&self) -> bool {
                 self.0.supports_string_literal_backslash_escape()
-            }
-
-            fn special_delimited_identifier_start(
-                &self,
-                chars: std::iter::Peekable<std::str::Chars<'_>>,
-            ) -> Option<(char, Option<char>)> {
-                self.0.special_delimited_identifier_start(chars)
             }
 
             fn supports_filter_during_aggregation(&self) -> bool {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -138,6 +138,15 @@ pub trait Dialect: Debug + Any {
         true
     }
 
+    /// Determine if nested quote start is presented and return it
+    fn nested_quote_start(
+        &self,
+        _quote_start: char,
+        mut _chars: Peekable<Chars<'_>>,
+    ) -> Option<char> {
+        None
+    }
+
     /// Determine if a character is a valid start character for an unquoted identifier
     fn is_identifier_start(&self, ch: char) -> bool;
 
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -129,13 +129,28 @@ pub trait Dialect: Debug + Any {
     }
 
     /// Determine if a character starts a potential nested quoted identifier.
-    /// RedShift support old way of quotation with `[` and it can cover even nested quoted identifier.
+    /// Example: RedShift supports the following quote styles to all mean the same thing:
+    /// ```sql
+    /// SELECT 1 AS foo;
+    /// SELECT 1 AS "foo";
+    /// SELECT 1 AS [foo];
+    /// SELECT 1 AS ["foo"];
+    /// ```
     fn is_nested_delimited_identifier_start(&self, _ch: char) -> bool {
         false
     }
 
-    /// Determine if nested quoted characters are presented
-    fn nested_delimited_identifier(
+    /// Only applicable whenever [`Self::is_nested_delimited_identifier_start`] returns true
+    /// If the next sequence of tokens potentially represent a nested identifier, then this method
+    /// returns a tuple containing the outer quote style, and if present, the inner (nested) quote style.
+    ///
+    /// Example (Redshift):
+    /// ```text
+    /// `["foo"]` => (Some(`[`), Some(`"`))
+    /// `[foo]` => (Some(`[`), None)
+    /// `"foo"` => None
+    /// ```
+    fn peek_nested_delimited_identifier_quotes(
         &self,
         mut _chars: Peekable<Chars<'_>>,
     ) -> Option<(char, Option<char>)> {
@@ -861,11 +876,11 @@ mod tests {
                 self.0.is_nested_delimited_identifier_start(ch)
             }
 
-            fn nested_delimited_identifier(
+            fn peek_nested_delimited_identifier_quotes(
                 &self,
                 chars: std::iter::Peekable<std::str::Chars<'_>>,
             ) -> Option<(char, Option<char>)> {
-                self.0.nested_delimited_identifier(chars)
+                self.0.peek_nested_delimited_identifier_quotes(chars)
             }
 
             fn identifier_quote_style(&self, identifier: &str) -> Option<char> {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -146,8 +146,9 @@ pub trait Dialect: Debug + Any {
     ///
     /// Example (Redshift):
     /// ```text
-    /// `["foo"]` => (Some(`[`), Some(`"`))
-    /// `[foo]` => (Some(`[`), None)
+    /// `["foo"]` => Some(`[`, Some(`"`))
+    /// `[foo]` => Some(`[`, None)
+    /// `[0]` => None
     /// `"foo"` => None
     /// ```
     fn peek_nested_delimited_identifier_quotes(
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -133,17 +133,11 @@ pub trait Dialect: Debug + Any {
         None
     }
 
-    /// Determine if quoted characters are proper for identifier
-    fn is_proper_identifier_inside_quotes(&self, mut _chars: Peekable<Chars<'_>>) -> bool {
-        true
-    }
-
-    /// Determine if nested quote start is presented and return it
-    fn nested_quote_start(
+    /// Determine if special way quoted characters are presented
+    fn special_delimited_identifier_start(
         &self,
-        _quote_start: char,
         mut _chars: Peekable<Chars<'_>>,
-    ) -> Option<char> {
+    ) -> Option<(char, Option<char>)> {
         None
     }
 
@@ -865,11 +859,11 @@ mod tests {
                 self.0.supports_string_literal_backslash_escape()
             }
 
-            fn is_proper_identifier_inside_quotes(
+            fn special_delimited_identifier_start(
                 &self,
                 chars: std::iter::Peekable<std::str::Chars<'_>>,
-            ) -> bool {
-                self.0.is_proper_identifier_inside_quotes(chars)
+            ) -> Option<(char, Option<char>)> {
+                self.0.special_delimited_identifier_start(chars)
             }
 
             fn supports_filter_during_aggregation(&self) -> bool {
--- a/src/dialect/redshift.rs
+++ b/src/dialect/redshift.rs
@@ -45,7 +45,7 @@ impl Dialect for RedshiftSqlDialect {
         // a start character and a quote is a beginning of quoted identifier.
         // Skipping analyzing token such as `"a"` and analyze only token that
         // can be part of json path potentially.
-        // For ex., `[0]`, `['a']` (seems part of json path) or `["a"]` (normal quoted identifier)
+        // For ex., `[0]` (seems part of json path) or `["a"]` (normal quoted identifier)
         if let Some(quote_start) = chars.peek() {
             if *quote_start == '"' {
                 return true;
@@ -61,6 +61,26 @@ impl Dialect for RedshiftSqlDialect {
             return ch == '"' || self.is_identifier_start(ch);
         }
         false
+    }
+
+    /// RedShift support nested quoted identifier like `["a"]`.
+    /// Determine if nested quote started and return it.  
+    fn nested_quote_start(
+        &self,
+        quote_start: char,
+        mut chars: Peekable<Chars<'_>>,
+    ) -> Option<char> {
+        if quote_start != '[' {
+            return None;
+        }
+
+        chars.next(); // skip opening quote start
+
+        if chars.skip_while(|ch| ch.is_whitespace()).peekable().peek() == Some(&'"') {
+            Some('"')
+        } else {
+            None
+        }
     }
 
     fn is_identifier_start(&self, ch: char) -> bool {
--- a/src/dialect/redshift.rs
+++ b/src/dialect/redshift.rs
@@ -53,7 +53,7 @@ impl Dialect for RedshiftSqlDialect {
             // PartiQL uses single quote as starting identification inside a quote
             // It is a normal identifier if it has no single quote at the beginning.
             // Additionally square bracket can contain quoted identifier.
-            return ch == '"' || ch != '\'' && self.is_identifier_start(ch);
+            return ch == '"' || self.is_identifier_start(ch);
         }
         false
     }
--- a/src/dialect/redshift.rs
+++ b/src/dialect/redshift.rs
@@ -32,55 +32,32 @@ pub struct RedshiftSqlDialect {}
 // in the Postgres dialect, the query will be parsed as an array, while in the Redshift dialect it will
 // be a json path
 impl Dialect for RedshiftSqlDialect {
-    fn is_delimited_identifier_start(&self, ch: char) -> bool {
-        ch == '"' || ch == '['
-    }
-
-    /// Determine if quoted characters are proper for identifier
+    /// Determine if quoted characters are looks like special case of quotation begining with `[`.
     /// It's needed to distinguish treating square brackets as quotes from
     /// treating them as json path. If there is identifier then we assume
     /// there is no json path.
-    fn is_proper_identifier_inside_quotes(&self, mut chars: Peekable<Chars<'_>>) -> bool {
-        // PartiQL (used as json path query language in Redshift) uses square bracket as
-        // a start character and a quote is a beginning of quoted identifier.
-        // Skipping analyzing token such as `"a"` and analyze only token that
-        // can be part of json path potentially.
-        // For ex., `[0]` (seems part of json path) or `["a"]` (normal quoted identifier)
-        if let Some(quote_start) = chars.peek() {
-            if *quote_start == '"' {
-                return true;
-            }
-        };
-        chars.next();
-        let mut not_white_chars = chars.skip_while(|ch| ch.is_whitespace()).peekable();
-        if let Some(&ch) = not_white_chars.peek() {
-            // PartiQL uses single quote as starting identification inside a quote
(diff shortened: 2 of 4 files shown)


src/dialect/redshift.rs
```diff
@@ -41,10 +41,19 @@ impl Dialect for RedshiftSqlDialect {
...
+            }
+        };
         chars.next();
         let mut not_white_chars = chars.skip_while(|ch| ch.is_whitespace()).peekable();
         if let Some(&ch) = not_white_chars.peek() {
-            return self.is_identifier_start(ch);
+            // PartiQL uses single quote as starting identification inside a quote
+            // It is a normal identifier if it has no single quote at the beginning.
+            // Additionally square bracket can contain quoted identifier.
+            return ch == '"' || ch != '\'' && self.is_identifier_start(ch);
```
bombsimon, line 56: Since the precedence here will group the `&&` expression, why do we need the `ch != '\''`, isn't that already checked in `is_identifier_start`?

Maybe I'm not understanding this part:
> PartiQL uses single quote as starting identification inside a quote

Would that be `"'foo'"`?

7phs: Good point! I've removed the extra check.

The expression can be `foo.field[0]['ab']`, but `foo.field[0]["ab"]` is invalid (I verified this on Redshift).

An identifier like `"'foo'"` never reaches this point. A quoted identifier starting with `"` is already checked in the condition above.

This method is designed to handle special cases of JSON paths, and it expects the input to start with a square bracket. I'm working to clarify this in this MR.

bombsimon: Gotcha! Maybe that's what confused me because it used to always consume the first token (a `[`?) but now the first token can be a `"` it seems like so I guess we reach this code even with `"A"` as in the PR description?

7phs: This method is called by [tokeniser](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/tokenizer.rs#L1078):
```
                quote_start
                    if self.dialect.is_delimited_identifier_start(ch)
                        && self
                            .dialect
                            .is_proper_identifier_inside_quotes(chars.peekable.clone()) =>
```
It applies to both delimiters of identifiers (`"` and `[`) to determine whether the current character sequence marks the beginning of a quoted token. This code is executed regardless of the specific characters—be it `[0]` or `"a"`.

However, only `[` is considered the start of a quoted identifier in json path case. This PR clarifies this behavior with a more explicit condition at the start of the method.

The highlighted line can only be triggered for `[0]` and never for `"a"`. However, there is a possibility that a double-quoted identifier like `["a"]` could be parsed. An additional check for `"` addresses this scenario. Note that JSON path expressions use single quotes `['a']` and never double quotes.

7phs: I added comments with examples to make a reason of conditions little bit more explicit.

src/dialect/redshift.rs
```diff
@@ -41,10 +41,24 @@ impl Dialect for RedshiftSqlDialect {
...
+        // PartiQL (used as json path query language in Redshift) uses square bracket as
+        // a start character and a quote is a beginning of quoted identifier.
+        // Skipping analyzing token such as `"a"` and analyze only token that
+        // can be part of json path potentially.
+        // For ex., `[0]`, `['a']` (seems part of json path) or `["a"]` (normal quoted identifier)
+        if let Some(quote_start) = chars.peek() {
+            if *quote_start == '"' {
+                return true;
+            }
+        };
```
bombsimon, line 53: On `main`, an expression like `SELECT foo["bar"]` evaluates to a `JsonAccess` with the path containing an ident named `bar` but I guess that's wrong. What's a bit interesting though is that with this change, it will evaluate to an expression with alias (`SELECT foo AS ["bar"]`). I know this library doesn't care about what's actually valid SQL, but I wonder if it's an unwanted side effect? I guess it just drops the `[]` and parses `SELECT foo "bar"` which is short for `SELECT foo AS "bar"`.

Another thing, is it intended to end up with `[]` as the quoting style and the value containing the quotes (`"a"`)?  Right now `["a"]` evaluates to an ident which I guess is what's causing above. Is `"a"` different from `a` as an identifier?

```rust
Ident {
    value: "\"1\"",
    quote_style: Some(
        '[',
    ),
},
```

If we encounter a `"` as the first character in `[]` quotes, should we somehow verify what's quoted or ensure we have a closing `"`? If we don't, anything will go, e.g. `["=[]` would parse to an `Ident` with value `"[=` but maybe that's valid PartiQL?

7phs: Thank you for feedback!

A goal of this MR is improving implementation of RedShift dialect. The method `is_proper_identifier_inside_quotes`  is implemented only for RedShift and I assume that it added only for RedShift specific case.

I verified one more time and found that `SELECT foo["bar"] FROM table_name` or `SELECT foo['bar'] FROM table_name` is not valid in RedShift case. `SELECT foo."bar" FROM table_name` is only valid for this certain case. We don't need to care about this case. I updated comments.

I'm going to improve verification of quoted identifier in this special case that you mentioned. Thank you for highlighting this edge case!

src/tokenizer.rs
```diff
@@ -1081,19 +1081,34 @@ impl<'a> Tokenizer<'a> {
...
                 {
-                    let error_loc = chars.location();
-                    chars.next(); // consume the opening quote
-                    let quote_end = Word::matching_end_quote(quote_start);
-                    let (s, last_char) = self.parse_quoted_ident(chars, quote_end);
+                    let word = if let Some(nested_quote_start) = self
+                        .dialect
+                        .nested_quote_start(quote_start, chars.peekable.clone())
+                    {
+                        chars.next(); // consume the opening quote
```
7phs, line 1088: This code is RedShift specific. The same time tokenisation of quoted identifier is a method of tokeniser and this is a reason of implemented it here.

Let me know if you have a good idea how makes this code more explicit.

src/dialect/redshift.rs
7phs: I added more unit-tests and updated a code of tokeniser to handle mentioned edge cases.

src/dialect/redshift.rs
```diff
@@ -41,14 +41,48 @@ impl Dialect for RedshiftSqlDialect {
     /// treating them as json path. If there is identifier then we assume
     /// there is no json path.
     fn is_proper_identifier_inside_quotes(&self, mut chars: Peekable<Chars<'_>>) -> bool {
```
iffyio, line 43: Wondering why redshift is special with this, Im thinking since there are other dialects that support jsonpath (which implements partiql in the parser) it shouldn't be. Would it not be enough to remove this `is_proper_identifier_inside_quotes` method entirely and the special `[` character from `is_delimited_identifier_start`?
From what I can tell in [their docs](https://docs.aws.amazon.com/redshift/latest/dg/r_names.html), `[` isn't an identifier delimiter so it seems that having the dialect `is_delimited_identifier_start` flag otherwise is incorrect?

7phs: It is a reasonable question!

I verified what happened if `is_proper_identifier_inside_quotes` is removed and `[` is removed from RedShift dialect - a couple tests failed `brackets_over_db_schema_table_name_with_whites_paces` and `test_square_brackets_over_db_schema_table_name` + my new one. Both are related to an old fashion way of using `[` in RedShift.

I verified that these queries work in RedShift and we can find it in query history or in end-user usages:
1. `SELECT 1 AS ["1"] FROM foo`
2. `SELECT [a] FROM [ schema ].[foo]`

Perhaps, needs to add a special case for `[` in RedShift dialect instead of improving identifying a possible identifier inside quotes.

7phs: @iffyio I have updated the PR based on your suggestion. Now, a common quoted identifier is processed without additional conditions, and RedShift uses the same set of starting characters for quoted identifiers as PostgreSQL.

Additionally, I added a specific match to handle the RedShift-specific case for [. All tests passed without requiring any changes.

src/dialect/mod.rs
```diff
@@ -128,14 +128,23 @@ pub trait Dialect: Debug + Any {
...
 
-    /// Return the character used to quote identifiers.
-    fn identifier_quote_style(&self, _identifier: &str) -> Option<char> {
+    /// Determine if a character starts a potential nested quoted identifier.
+    /// RedShift support old way of quotation with `[` and it can cover even nested quoted identifier.
+    fn is_nested_delimited_identifier_start(&self, _ch: char) -> bool {
+        false
+    }
+
+    /// Determine if nested quoted characters are presented
```
iffyio, line 137:
```suggestion
    /// Only applicable whenever [`Self::is_nested_delimited_identifier_start`] returns true
    /// If the next sequence of tokens potentially represent a nested identifier, then this method
    /// returns a tuple containing the outer quote style, and if present, the inner (nested) quote style.
    ///
    /// Example (Redshift):
    /// ```text
    /// `["foo"]` => (Some(`[`), Some(`"`))
    /// `[foo]` => (Some(`[`), None)
    /// `"foo"` => None
    /// ```
```
Similarly, to clarify the behavior with an example since it wouldn't be intuitive otherwise

src/dialect/mod.rs
```diff
@@ -128,14 +128,23 @@ pub trait Dialect: Debug + Any {
         ch == '"' || ch == '`'
     }
 
-    /// Return the character used to quote identifiers.
-    fn identifier_quote_style(&self, _identifier: &str) -> Option<char> {
+    /// Determine if a character starts a potential nested quoted identifier.
+    /// RedShift support old way of quotation with `[` and it can cover even nested quoted identifier.
```
iffyio, line 132:
```suggestion
    /// Example: RedShift supports the following quote styles to all mean the same thing:
    /// ```sql
    /// SELECT 1 AS foo;
    /// SELECT 1 AS "foo";
    /// SELECT 1 AS [foo];
    /// SELECT 1 AS ["foo"];
    /// ```
```
Thinking something like this to clarify the behavior?

src/dialect/mod.rs
```diff
@@ -128,14 +128,23 @@ pub trait Dialect: Debug + Any {
...
+    fn is_nested_delimited_identifier_start(&self, _ch: char) -> bool {
+        false
+    }
+
+    /// Determine if nested quoted characters are presented
+    fn nested_delimited_identifier(
```
iffyio, line 138:
```suggestion
    fn peek_nested_delimited_identifier_quotes(
```

src/dialect/redshift.rs
```diff
@@ -32,21 +32,36 @@ pub struct RedshiftSqlDialect {}
...
 // be a json path
 impl Dialect for RedshiftSqlDialect {
-    fn is_delimited_identifier_start(&self, ch: char) -> bool {
-        ch == '"' || ch == '['
+    fn is_nested_delimited_identifier_start(&self, ch: char) -> bool {
+        ch == '['
     }
 
-    /// Determine if quoted characters are proper for identifier
+    /// Determine if quoted characters are looks like special case of quotation begining with `[`.
```
iffyio, line 39: Maybe we can copy over the suggested dialect method comment here since its already describing redshift.

src/tokenizer.rs
```diff
@@ -1075,25 +1075,56 @@ impl<'a> Tokenizer<'a> {
                     Ok(Some(Token::DoubleQuotedString(s)))
                 }
                 // delimited (quoted) identifier
+                quote_start if self.dialect.is_delimited_identifier_start(ch) => {
+                    let word = self.tokenize_quoted_identifier(quote_start, chars)?;
+                    Ok(Some(Token::make_word(&word, Some(quote_start))))
+                }
+                // special (quoted) identifier
```
iffyio, line 1082:
```suggestion
                // Potentially nested delimited (quoted) identifier
```

src/tokenizer.rs
```diff
@@ -1075,25 +1075,56 @@ impl<'a> Tokenizer<'a> {
...
-                            .is_proper_identifier_inside_quotes(chars.peekable.clone()) =>
+                            .nested_delimited_identifier(chars.peekable.clone())
+                            .is_some() =>
                 {
-                    let error_loc = chars.location();
-                    chars.next(); // consume the opening quote
+                    let (quote_start, nested_quote_start) = self
+                        .dialect
+                        .nested_delimited_identifier(chars.peekable.clone())
+                        .unwrap();
```
iffyio, line 1095:
```suggestion
```
we can return an error here?

tests/sqlparser_redshift.rs
```diff
@@ -353,3 +378,20 @@ fn test_parse_json_path_from() {
...
+    redshift().verified_stmt(r#"SELECT 1 AS ["=]"] FROM a"#);
+    redshift().verified_stmt(r#"SELECT 1 AS ["a[b]"] FROM a"#);
+}
+
+#[test]
+fn test_parse_create_numbered_columns() {
+    redshift_and_generic().verified_stmt(
+        r#"CREATE TABLE test_table_1 ("1" INT, "d" VARCHAR(155), "2" DOUBLE PRECISION)"#,
+    );
+}
```
iffyio, line 397: Can we incorporate this as a test case under the `parse_delimited_identifiers` test?

tests/sqlparser_redshift.rs
```diff
@@ -353,3 +378,20 @@ fn test_parse_json_path_from() {
         _ => panic!(),
     }
 }
+
+#[test]
+fn test_parse_select_numbered_columns() {
```
iffyio, line 383: thinking a couple cases we can include?
```sql
SELECT 1 AS ["1];
SELECT 1 AS [ " 1 " ];
```

tests/sqlparser_redshift.rs
```diff
@@ -353,3 +378,20 @@ fn test_parse_json_path_from() {
...
+    redshift_and_generic().verified_stmt(r#"SELECT 1 AS "1" FROM a"#);
+    // RedShift specific case - quoted identifier inside square bracket
```
iffyio, line 385:
```suggestion
```
this one seems to be already covered by `test_parse_create_numbered_columns`?

src/tokenizer.rs
```diff
@@ -1075,25 +1075,56 @@ impl<'a> Tokenizer<'a> {
...
-                        self.tokenizer_error(
+                    chars.next(); // skip the first delimiter
+                    word.push(peeking_take_while(chars, |ch| ch.is_whitespace()));
+                    if chars.peek() != Some(&nested_quote_start) {
+                        return self.tokenizer_error(
+                            error_loc,
+                            format!("Expected nested delimiter '{nested_quote_start}' before EOF."),
+                        );
+                    }
+                    word.push(format!("{nested_quote_start}"));
```
iffyio, line 1115:
```suggestion
                    word.push(nested_quote_start.into()));
```

src/tokenizer.rs
```diff
@@ -1075,25 +1075,56 @@ impl<'a> Tokenizer<'a> {
...
+                    word.push(self.tokenize_quoted_identifier(nested_quote_end, chars)?);
+                    word.push(format!("{nested_quote_end}"));
```
iffyio, line 1117:
```suggestion
                    word.push(nested_quote_end.into());
```

tests/sqlparser_redshift.rs
```diff
@@ -279,6 +279,31 @@ fn test_redshift_json_path() {
...
+            value: Box::new(Expr::CompoundIdentifier(vec![
+                Ident::new("db1"),
+                Ident::new("sc1"),
+                Ident::new("tbl1"),
+                Ident::new("col1")
+            ])),
+            path: JsonPath {
+                path: vec![
+                    JsonPathElem::Bracket {
+                        key: Expr::Value(Value::Number("0".parse().unwrap(), false))
```
iffyio, line 296:
```suggestion
                        key: Expr::Value(number("0"))
```

iffyio (commented):
Thanks @7phs! Left some minor comments I think the approach looks reasonable!

src/dialect/mod.rs
7phs: Updated.

src/dialect/mod.rs
7phs: Updated.

src/dialect/redshift.rs
7phs: Comments copied.

src/dialect/mod.rs
7phs: Renamed

src/tokenizer.rs
7phs: Updated

src/tokenizer.rs
7phs: Fixed.

tests/sqlparser_redshift.rs
7phs: I split the test into two to make it more explicit:
1. `test_parse_select_numbered_columns` - it is a reason for this MR
2. `test_parse_nested_quoted_identifier`

I added suggested cases into `test_parse_nested_quoted_identifier`

src/tokenizer.rs
7phs: Updated

src/tokenizer.rs
7phs: Updated

tests/sqlparser_redshift.rs
7phs: All similar lines in `sqlparser_redshift.rs` are updated.

tests/sqlparser_redshift.rs
7phs: I split the test into two to make it more explicit:

1. `test_parse_select_numbered_columns` - it is a reason for this MR
2. `test_parse_nested_quoted_identifier`

src/dialect/mod.rs
```diff
@@ -128,14 +128,38 @@ pub trait Dialect: Debug + Any {
...
+
+    /// Only applicable whenever [`Self::is_nested_delimited_identifier_start`] returns true
+    /// If the next sequence of tokens potentially represent a nested identifier, then this method
+    /// returns a tuple containing the outer quote style, and if present, the inner (nested) quote style.
+    ///
+    /// Example (Redshift):
+    /// ```text
+    /// `["foo"]` => (Some(`[`), Some(`"`))
+    /// `[foo]` => (Some(`[`), None)
+    /// `"foo"` => None
```
bombsimon, line 151: nit: Looks like this doesn't line up with the return type.

```suggestion
    /// `["foo"]` => Some(('[', Some('"'))
    /// `[foo]` => Some(('[', None))
    /// `"foo"` => None
```

src/dialect/redshift.rs
```diff
@@ -32,21 +32,50 @@ pub struct RedshiftSqlDialect {}
...
-    fn is_proper_identifier_inside_quotes(&self, mut chars: Peekable<Chars<'_>>) -> bool {
+    /// Only applicable whenever [`Self::is_nested_delimited_identifier_start`] returns true
+    /// If the next sequence of tokens potentially represent a nested identifier, then this method
+    /// returns a tuple containing the outer quote style, and if present, the inner (nested) quote style.
+    ///
+    /// Example (Redshift):
+    /// ```text
+    /// `["foo"]` => (Some(`[`), Some(`"`))
+    /// `[foo]` => (Some(`[`), None)
+    /// `"foo"` => None
```
bombsimon, line 55: Same here right?

```suggestion
    /// `["foo"]` => Some(('[', Some('"'))
    /// `[foo]` => Some(('[', None))
    /// `"foo"` => None
```
```

**Review by bombsimon — APPROVED** [review]
🚀

File: src/dialect/mod.rs (hunk 17 above)
Comment (7phs) [reply src/dialect/mod.rs]: Nice catch.
Updated and I added one more example `[0]`

File: tests/sqlparser_redshift.rs [hunk 19]
```diff
@@ -353,3 +380,23 @@ fn test_parse_json_path_from() {
...
+fn test_parse_select_numbered_columns() {
+    // An alias starting with a number
+    redshift_and_generic().verified_stmt(r#"SELECT 1 AS "1" FROM a"#);
+    redshift_and_generic().verified_stmt(r#"SELECT 1 AS "1abc" FROM a"#);
+}
+
+#[test]
+fn test_parse_nested_quoted_identifier() {
+    redshift().verified_stmt(r#"SELECT 1 AS ["1"] FROM a"#);
+    redshift().verified_stmt(r#"SELECT 1 AS [ " 1 " ]"#);
```
iffyio, line 394: Ah I think the expected behavior would be that the parser strips out whitespace:
```rust
    redshift().one_statement_parses_to(
        r#"SELECT 1 AS [ " 1 " ]"#,
        r#"SELECT 1 AS [" 1 "]"#,
    );
```

7phs: It depends on what we want to achieve. I have implemented parsing of nested quotations, guided by the test case [brackets_over_db_schema_table_name_with_whites_paces](https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_redshift.rs#L69). For example, the following query:
```
SELECT [   col1  ] FROM [  test_schema].[ test_table]
```
is successfully parsed using `verified_stmt`.

I don't have a strong opinion on this implementation and can adjust the parsing if needed.

iffyio: Oh yeah I think that the linked test would be incorrect as well? from what I could tell and tested the following seemed to all be equivalent in redshift
```sql
SELECT 1 AS foo;
SELECT 1 AS [ foo ];
SELECT 1 AS [ "foo" ];
```
so that whitespace within the brackets isn't part of the identifier. Then it would be incorrect to have an identifier in the AST that contains the string ` foo ` for any of the above.

On the other hand this would actually be referencing an identifier containing whitespace
```sql
SELECT 1 AS [" foo "];
```

7phs: I have updated the PR to trim spaces in cases of nested quotations, e.g., `[  "foo"  ]` -> `["foo"]`. Implementing similar behavior for cases like `[  foo  ]` would require more extensive changes to the existing code. In my opinion, this could introduce some risk, at least in the context of the current PR.

iffyio (approved):
LGTM! Thanks @7phs @bombsimon 
cc @alamb

PR #1578, opened 2024-12-03
Parse Snowflake USE ROLE and USE SECONDARY ROLES

yoavcloud:
This PR adds support for Snowflake `USE ROLE <role>` and `USE SECONDARY ROLES` as described here: https://docs.snowflake.com/en/sql-reference/sql/use-secondary-roles


alamb (approved):
Makes sense to me -- thank you @yoavcloud 

FYI @iffyio

tests/sqlparser_snowflake.rs
```diff
@@ -2709,6 +2709,20 @@ fn parse_use() {
...
             ])))
         );
+        std::assert_eq!(
+            snowflake().verified_stmt(&format!("USE ROLE {0}my_role{0}", quote)),
+            Statement::Use(Use::Role(ObjectName(vec![Ident::with_quote(
+                quote,
+                "my_role".to_string(),
+            )])))
+        );
+        std::assert_eq!(
```
iffyio, line 2719:
```suggestion
        assert_eq!(
            snowflake().verified_stmt(&format!("USE ROLE {0}my_role{0}", quote)),
            Statement::Use(Use::Role(ObjectName(vec![Ident::with_quote(
                quote,
                "my_role".to_string(),
            )])))
        );
        assert_eq!(
```
if these were autoimported?

iffyio (commented):
Thanks @yoavcloud! Just one minor comment on the std assert_eq import

tests/sqlparser_snowflake.rs
yoavcloud: Fixed across the board

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #1579, opened 2024-12-04
Snowflake ALTER TABLE clustering options

yoavcloud:
This PR adds support for parsing the following `ALTER TABLE` statements in Snowflake:
`ALTER TABLE <tbl> CLUSTER BY`
`ALTER TABLE <tbl> SUSPEND | RESUME RECLUSTER`
`ALTER TABLE <tbl> DROP CLUSTERING KEY`

https://docs.snowflake.com/en/sql-reference/sql/alter-table#clustering-actions-clusteringaction


src/ast/ddl.rs
```diff
@@ -152,31 +159,43 @@ pub enum AlterTableOperation {
...
+    },
     /// `ENABLE ROW LEVEL SECURITY`
     ///
     /// Note: this is a PostgreSQL-specific operation.
     EnableRowLevelSecurity,
     /// `ENABLE RULE rewrite_rule_name`
     ///
     /// Note: this is a PostgreSQL-specific operation.
-    EnableRule { name: Ident },
+    EnableRule {
```
alamb, line 190: it is weird that these got reformatted

alamb (approved):
Looks good to me as well (thanks @yoavcloud !)

cc @iffyio

src/parser/mod.rs
```diff
@@ -7273,6 +7273,10 @@ impl<'a> Parser<'a> {
                 let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
                 let name = self.parse_identifier(false)?;
                 AlterTableOperation::DropProjection { if_exists, name }
+            } else if dialect_of!(self is SnowflakeDialect | GenericDialect)
+                && self.parse_keywords(&[Keyword::CLUSTERING, Keyword::KEY])
```
iffyio, line 7277: for the gating can we use a method on the dialect trait in place of the `dialect_of!`? I'm thinking as well if it would make sense to one method to represent these options since they belong to the same feature , e.g. `supports_alter_table_clustering_action()` or similar method?

yoavcloud: Perhaps we should drop the gate altogether in this case and introduce gating as necessary? As of now, the gate will only help reject statements the dialect shouldn't accept, vs. parsing incorrectly. Just a thought though...

iffyio: Yeah i think that sounds reasonable!

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #1580, opened 2024-12-07
Add support for TABLESAMPLE

yoavcloud:
This PR adds support for the `TABLESAMPLE` option in the following dialects:
 - Standard SQL: https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#sample-clause
 - Snowflake: https://docs.snowflake.com/en/sql-reference/constructs/sample
 - Hive: https://cwiki.apache.org/confluence/display/hive/languagemanual+sampling
 - Clickhouse: https://clickhouse.com/docs/en/sql-reference/statements/select/sample

Collateral work includes expanding the use of a constructor function to create `Table` structs in unit tests to avoid modifying many files when adding default options to the `Table` struct.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10655,7 +10655,7 @@ impl<'a> Parser<'a> {
         }
     }
 
-    fn maybe_parse_table_sample(&mut self) -> Result<Option<TableSample>, ParserError> {
+    fn maybe_parse_table_sample(&mut self) -> Result<Option<Box<TableSample>>, ParserError> {
         if self
             .parse_one_of_keywords(&[Keyword::SAMPLE, Keyword::TABLESAMPLE])
             .is_none()
@@ -10663,7 +10663,7 @@ impl<'a> Parser<'a> {
             return Ok(None);
         }
 
-        if self
+        let sample = if self
             .parse_one_of_keywords(&[Keyword::BERNOULLI, Keyword::ROW])
             .is_some()
         {
@@ -10678,11 +10678,11 @@ impl<'a> Parser<'a> {
                 (Some(expr), None, None)
             };
             self.expect_token(&Token::RParen)?;
-            Ok(Some(TableSample::Bernoulli(TableSampleBernoulli {
+            TableSample::Bernoulli(TableSampleBernoulli {
                 probability,
                 value,
                 unit,
-            })))
+            })
         } else if self
             .parse_one_of_keywords(&[Keyword::SYSTEM, Keyword::BLOCK])
             .is_some()
@@ -10701,10 +10701,10 @@ impl<'a> Parser<'a> {
             } else {
                 None
             };
-            Ok(Some(TableSample::System(TableSampleSystem {
+            TableSample::System(TableSampleSystem {
                 probability,
                 repeatable: seed,
-            })))
+            })
         } else if self.peek_token().token == Token::LParen {
             self.expect_token(&Token::LParen)?;
             if self.parse_keyword(Keyword::BUCKET) {
@@ -10717,11 +10717,11 @@ impl<'a> Parser<'a> {
                     None
                 };
                 self.expect_token(&Token::RParen)?;
-                Ok(Some(TableSample::Bucket(TableSampleBucket {
+                TableSample::Bucket(TableSampleBucket {
                     bucket,
                     total,
                     on,
-                })))
+                })
             } else {
                 let value = match self.try_parse(|p| p.parse_number_value()) {
                     Ok(num) => num,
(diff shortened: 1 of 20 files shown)


src/parser/mod.rs
```diff
@@ -10584,6 +10598,116 @@ impl<'a> Parser<'a> {
...
+                    Some(self.parse_expr()?)
+                } else {
+                    None
+                };
+                self.expect_token(&Token::RParen)?;
+                TableSampleMethod::Bucket(TableSampleBucket { bucket, total, on })
+            } else {
+                let value = match self.maybe_parse(|p| p.parse_number_value()) {
+                    Ok(Some(num)) => num,
+                    _ => {
```
iffyio, line 10669:
```suggestion
                let value = match self.maybe_parse(|p| p.parse_number_value())? {
                    Some(num) => num,
                    None => {
```
I think this is usually the recursion limit or similar fatal error we can propagate

src/parser/mod.rs
```diff
@@ -10584,6 +10598,116 @@ impl<'a> Parser<'a> {
...
+                        } else {
+                            return parser_err!(
+                                "Expecting number or byte length e.g. 100M",
+                                self.peek_token().span.start
+                            );
+                        }
+                    }
+                };
+                if self.peek_token().token == Token::RParen
+                    && !self.dialect.supports_implicit_table_sample_method()
```
iffyio, line 10681: maybe this could be simplified as `if self.dialect.supports_implicit_table_sample_method() && self.consume_token(Token::RParen)` it would also let us skip the expect_token that follows as well?

src/parser/mod.rs
```diff
@@ -10584,6 +10598,116 @@ impl<'a> Parser<'a> {
...
+                Some(seed)
+            } else {
+                None
+            };
+            TableSampleMethod::System(TableSampleSystem {
+                probability,
+                repeatable: seed,
+            })
+        // Try to parse without an explicit table sample method keyword
+        } else if self.peek_token().token == Token::LParen {
```
iffyio, line 10653:
```suggestion
        } else if self.consume_token(Token::LParen) {
```

tests/sqlparser_snowflake.rs
```diff
@@ -2952,3 +2950,33 @@ fn test_sf_double_dot_notation() {
...
+
+#[test]
+fn test_table_sample() {
+    snowflake_and_generic()
+        .verified_stmt("SELECT * FROM testtable AS t TABLESAMPLE BERNOULLI (10)");
+
+    // In Snowflake we translate implicit table sample method to bernoulli
+    snowflake().one_statement_parses_to(
+        "SELECT * FROM testtable SAMPLE (10)",
+        "SELECT * FROM testtable TABLESAMPLE BERNOULLI (10)",
```
iffyio, line 2962: for the scenarios that currently rely on `one_statement_parse_to`, could we represent them faithfully when displaying? e.g this and the `ROW` vs `BERNOULLI` variants etc

yoavcloud: Not sure if I understand the question, but if you mean whether the variants I added as interchangeable are really interchangeable in the dialect, as far as I understand yes. If not, please elaborate?

iffyio: Ah yeah so I was rather thinking that ideally we preserve the syntax roundtrip, even though they'r interchangeable in some dialects 
```rust
snowflake_and_generic().verified_stmt("SELECT * FROM testtable TABLESAMPLE ROW (20.3)");
snowflake_and_generic().verified_stmt("SELECT * FROM testtable SAMPLE BLOCK (3) SEED (82)");
```

yoavcloud:
@iffyio please see latest commit

tests/sqlparser_snowflake.rs
```diff
@@ -2952,3 +2950,19 @@ fn test_sf_double_dot_notation() {
 
 #[test]
 fn test_parse_double_dot_notation_wrong_position() {}
+
+#[test]
+fn test_table_sample() {
+    snowflake_and_generic().one_statement_parses_to(
+        "SELECT * FROM testtable SAMPLE (10)",
+        "SELECT * FROM testtable TABLESAMPLE (10)",
+    );
```
iffyio, line 2959: Oh could we do the same here as well? clickhouse for example doesn't have has `SAMPLE` but not `TABLESAMPLE` so that the roundtrip stays correct

iffyio (commented):
Thanks @yoavcloud! Left a comment to distinguish sample vs tablesample. Otherwise this looks good to me!

tests/sqlparser_snowflake.rs
yoavcloud: Added support for clickhouse as well, with a different approach to how to model the it in the AST

iffyio (approved):
LGTM! Thanks @yoavcloud 
cc @alamb

alamb:
Thanks @yoavcloud and @iffyio

PR #1584, opened 2024-12-09
Support INSERT OVERWRITE INTO syntax

yuval-illumex:
The following query is valid in Snowflake:
`INSERT OVERWRITE INTO schema.table SELECT a FROM b`


alamb:
Thanks @yuval-illumex and @iffyio

PR #1585, opened 2024-12-11
Add support for ODBC functions

iffyio:
Adds support for dialects like Snowflake and MsSQL that support function calls using the ODBC syntax.

```sql
SELECT {fn CONCAT('foo', 'bar')}
```

https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/scalar-function-calls?view=sql-server-2017


iffyio:
cc @alamb

alamb (approved):
Thank you @iffyio -- I couldn't find any suggestions for this PR. Very nice 🙏

PR #1586, opened 2024-12-11
Parse `INSERT` with subquery when lacking column names

iffyio:
Adds support to parse the example scenario:
```sql
INSERT INTO t1 (SELECT 1)
```

The parser previously looked to parse the subquery as a parenthesized list of column names which was
incorrect in this case.


alamb (approved):
Thanks @iffyio

Issue #1588, opened 2024-12-11
Optimize `Token::make_word`


davisp:
While working on #1587 I noticed that Instruments is showing `Token::make_word` as the second hottest single function, right after `alloc::raw_vec::finish_grow`.

Looking into the implementation I saw that its just doing a binary search across all keywords to find if its a known keyword or not. This is a fairly classical case where we have a known set of strings and want to check if a given string is in that list. There are a bunch of ways that we could speed this up. This issue is to figure out a good compromise between those possible speedups and other project constraints like maintaining a `no_std` ability.

My [first approach](https://github.com/apache/datafusion-sqlparser-rs/commit/4551933dc0a9e892e412be5ca0022a124859dad0) at speeding this up was to create a table for the first byte in every keyword to reduce the number of entries that need to be searched. This small optimization managed to shave off about 400ms of time (of the 1.4ish seconds total).

However, there are other approaches that could speed this up even more. Either by generating parsing/lookup tables or using something like [phf](https://crates.io/crates/phf) to do the heavy lifting for us.

LorrensP-2158466:
I don't know if this is useful, but I remember watching a video from Strager: [Perfect Hash Tables](https://youtu.be/DMQ_HcNSOAI?si=pTcWhy_D2-wBIkHb). He had the same kind of problem and created a "custom" hash table by using the information he already has, the known keywords

In short, because he knew which words are keywords, he built his hash table around that by combining the first 2 and last 2 bytes and hashing that. And after that still compare. (This is of the top of my head) 

I did notice there are about ~800 keywords, so I don't know if it's feasible, but maybe it's worth looking into it?

LucaCappelletti94:
Completed by #2176 and #2226.

PR #1590, opened 2024-12-11
Find keywords using perfect hashing

davisp:
This updates the keyword lookup algorthm to use the phf crate.

This is one possible approach to solving #1588 at the cost of pulling in a new dependency on [phf](https://crates.io/crates/phf). I will note that phf is rather small and is being used in a `no_std` configuration, but it is a new dependency which may be enough to not go this route.

Local measuring with the sqlparser_bench shows roughly a 5% speedup. I'll open a second PR for comparison with my other approach to just scope the binary search to words starting with the same first letter.


src/keywords.rs
```diff
@@ -33,845 +33,13 @@ use serde::{Deserialize, Serialize};
...
-            $(Keyword::$ident),*
-        ];
-
-        $(kw_def!($ident $(= $string_keyword)?);)*
-        pub const ALL_KEYWORDS: &[&str] = &[
-            $($ident),*
-        ];
-    };
+pub fn lookup(keyword: &str) -> Keyword {
+    let keyword = keyword.to_ascii_uppercase();
```
alamb, line 39: If we are going to optimize lookup I think we should also avoid this call to `to_ascii_upprcase` as it allocates / copies the string which I believe is significiantly more costly than some comparisons

https://doc.rust-lang.org/std/ascii/trait.AsciiExt.html#tymethod.to_ascii_uppercase
https://doc.rust-lang.org/std/primitive.str.html#method.to_uppercase

Since all keywords are ASCII, what if we did the binary search character by character

Here is some brain 🤮 :

```
let first_char = string.first().to_upper() 
// find range of keywords within the
let range = binary_search(keywords, first_char, 0);
// if not unique location, proceed to second character, etc
...
```

build.rs
```diff
@@ -0,0 +1,101 @@
+use std::env;
+use std::fs::File;
+use std::io::{BufWriter, Write};
+use std::path::Path;
+
+fn read_keywords() -> Vec<(String, Option<String>)> {
+    let path = Path::new("src").join("keywords.txt");
```
alamb, line 7: this is quite clever

alamb (commented):
Thank you for thsi @davisp

alamb:
As much as I love not rewriting code, I am quite loath to take on a new dependency in this crate. My concerns are
1. The binary / dependency size
2.  (mostly) that every additional crate used results in another places to help maintain (see for example https://github.com/apache/arrow-rs/issues/6858)

That being said, `phf` appears to be widely used and not changing: https://crates.io/crates/phf/reverse_dependencies

Are you wiling to try the "compare without allocating approach" described in https://github.com/apache/datafusion-sqlparser-rs/pull/1591#discussion_r1882405987?

tobyhede:
Also an open question regarding maintenance of the phf crate https://github.com/rust-phf/rust-phf/issues/318
It does appear to be relatively stable, most of the open issues appear to be feature requests and enhancements rather than fixes.

alamb:

PR #1591, opened 2024-12-11
Slightly faster keyword lookups

davisp:
Its a micro optimization but seemed to give a bit of a boost to only search words starting with the correct letter.

An alternative to #1590 that doesn't involve adding a new dependency. This just relies on having a pre-calculated table of word ranges to binary search.

Relatively speaking, this approach gets about slightly less than half the speed up of using phf.

### DIFF
--- a/src/keywords.rs
+++ b/src/keywords.rs
@@ -973,3 +973,61 @@ pub const RESERVED_FOR_IDENTIFIER: &[Keyword] = &[
     Keyword::STRUCT,
     Keyword::TRIM,
 ];
+
+pub const NA: usize = usize::MAX;
+
+#[rustfmt::skip]
+pub const KEYWORD_LOOKUP_INDEX_ROOT: &[usize; 26] = &[
+      0,  42,  67, 148, 198, 241, 281, 294, 305, 350, 357, 360, 390,
+    430, 465, 497, 539, 543, 605, 683, 728, 761, 780, 793, 795, 796,
+];
+
+pub fn lookup(word: &str) -> Keyword {
+    if word.len() < 2 {
+        return Keyword::NoKeyword;
+    }
+
+    let word = word.to_uppercase();
+    let byte1 = word.as_bytes()[0];
+    if !byte1.is_ascii_uppercase() {
+        return Keyword::NoKeyword;
+    }
+
+    let start = KEYWORD_LOOKUP_INDEX_ROOT[(byte1 - b'A') as usize];
+
+    let end = if (byte1 + 1) <= b'Z' {
+        KEYWORD_LOOKUP_INDEX_ROOT[(byte1 - b'A' + 1) as usize]
+    } else {
+        ALL_KEYWORDS.len()
+    };
+
+    let keyword = ALL_KEYWORDS[start..end].binary_search(&word.as_str());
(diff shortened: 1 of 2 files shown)


src/keywords.rs
```diff
@@ -973,3 +973,61 @@ pub const RESERVED_FOR_IDENTIFIER: &[Keyword] = &[
...
+      0,  42,  67, 148, 198, 241, 281, 294, 305, 350, 357, 360, 390,
+    430, 465, 497, 539, 543, 605, 683, 728, 761, 780, 793, 795, 796,
+];
+
+pub fn lookup(word: &str) -> Keyword {
+    if word.len() < 2 {
+        return Keyword::NoKeyword;
+    }
+
+    let word = word.to_uppercase();
```
alamb, line 990: if we can figure out how to remove this to_uppercase call I think this approach will likely be very fast.

I dug around in the rust API docs and it seems like we could change the comparison to ignore ascii case. See comment below 🤔

src/keywords.rs
```diff
@@ -973,3 +973,61 @@ pub const RESERVED_FOR_IDENTIFIER: &[Keyword] = &[
...
+
+    let start = KEYWORD_LOOKUP_INDEX_ROOT[(byte1 - b'A') as usize];
+
+    let end = if (byte1 + 1) <= b'Z' {
+        KEYWORD_LOOKUP_INDEX_ROOT[(byte1 - b'A' + 1) as usize]
+    } else {
+        ALL_KEYWORDS.len()
+    };
+
+    let keyword = ALL_KEYWORDS[start..end].binary_search(&word.as_str());
```
alamb, line 1004: I wonder if we could use something like

```suggestion
    let keyword = ALL_KEYWORDS[start..end].binary_search_by(|s| s.eq_ignore_ascii_case(&word))
```

https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by
https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case

alamb (commented):
Thank you (again) @davisp  -- I think this is a promising approach.

src/keywords.rs
```diff
@@ -973,3 +973,61 @@ pub const RESERVED_FOR_IDENTIFIER: &[Keyword] = &[
     Keyword::STRUCT,
     Keyword::TRIM,
 ];
+
+pub const NA: usize = usize::MAX;
+
+#[rustfmt::skip]
+pub const KEYWORD_LOOKUP_INDEX_ROOT: &[usize; 26] = &[
```
Dandandan, line 980: I think this is not super maintainable (needs to be updated manually when adding keywords).

alamb: I agree - it we go with this type of table driven approach we should have some sort of update script (or build.rs) that builds the table from the enum.

src/keywords.rs
alamb: I actually tried to do this approach, and I found the `to_uppercase` call is being done as part of the token creation. So to save the allocation we need to to avoid

https://github.com/alamb/sqlparser-rs/blob/4ab3ab91473d152c652e6582b63abb13535703f9/src/tokenizer.rs#L347-L346 

I think we can change it to `make_ascii_uppercase` perhaps

alamb:

PR #1598, opened 2024-12-12
Fix BigQuery hyphenated ObjectName with numbers

ayman-sigma:
We currently support hyphenated identifiers for BigQuery. The current code expects the number segment to be the last segment ex: `foo-123` and should be followed by whitespace. That is true except when this identifier is part of an ObjectName. Ex: `SELECT * FROM foo-123.bar`. 
The issue is that tokenizer parse the previous string as: `[Word("foo"), Minus, Number("123."), Word("bar")]`
This PR:
1. Fixes the tokenizer to tokenize `foo-123.bar` as `[Word('"foo"), Minus, Number("123"), Period, Word("bar")]`
2. Fixes the parser to parse `foo-123.bar` as `ObjectName([Ident("foo-123"), Ident("bar")])` instead of erroring out.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8747,7 +8747,9 @@ impl<'a> Parser<'a> {
                             }
                             Token::Number(s, false) if s.chars().all(|c| c.is_ascii_digit()) => {
                                 ident.value.push_str(&s);
-                                true
+                                // If next token is period, then it is part of an ObjectName and we don't expect whitespace
+                                // after the number.
+                                !matches!(self.peek_token().token, Token::Period)
                             }
                             _ => {
                                 return self
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1119,7 +1119,7 @@ impl<'a> Tokenizer<'a> {
                         {
                             s.push('.');
                             chars.next();
-                        } else if s.is_empty() {
+                        } else if !s.is_empty() {
                             // Number might be part of period separated construct. Keep the period for next token
                             // e.g. a-12.b
                             return Ok(Some(Token::Number(s, false)));
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1108,15 +1108,29 @@ impl<'a> Tokenizer<'a> {
 
                     // match one period
                     if let Some('.') = chars.peek() {
-                        s.push('.');
-                        chars.next();
+                        // Check if this actually is a float point number
+                        let mut char_clone = chars.peekable.clone();
+                        char_clone.next();
+                        // Next char should be a digit, otherwise, it is not a float point number
+                        if char_clone
+                            .peek()
+                            .map(|c| c.is_ascii_digit())
+                            .unwrap_or(false)
+                        {
+                            s.push('.');
+                            chars.next();
+                        } else if s.is_empty() {
+                            // Number might be part of period separated construct. Keep the period for next token
+                            // e.g. a-12.b
+                            return Ok(Some(Token::Number(s, false)));
+                        } else {
+                            // No number -> Token::Period
+                            chars.next();
+                            return Ok(Some(Token::Period));
+                        }
                     }
-                    s += &peeking_take_while(chars, |ch| ch.is_ascii_digit());
 
-                    // No number -> Token::Period
-                    if s == "." {
-                        return Ok(Some(Token::Period));
-                    }
+                    s += &peeking_take_while(chars, |ch| ch.is_ascii_digit());
 
                     let mut exponent_part = String::new();
                     // Parse exponent as number
@@ -2125,6 +2139,23 @@ mod tests {
             Token::Number(String::from(".1"), false),
         ];
 
+        compare(expected, tokens);
+    }
+
+    #[test]
+    fn tokenize_select_float_hyphenated_identifier() {
+        let sql = String::from("SELECT a-12.b");
+        let dialect = GenericDialect {};
+        let tokens = Tokenizer::new(&dialect, &sql).tokenize().unwrap();
+        let expected = vec![
+            Token::make_keyword("SELECT"),
+            Token::Whitespace(Whitespace::Space),
+            Token::make_word("a", None),
+            Token::Minus,
+            Token::Number(String::from("12"), false),
+            Token::Period,
+            Token::make_word("b", None),
+        ];
         compare(expected, tokens);
     }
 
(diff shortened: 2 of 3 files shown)


ayman-sigma:
I think I messed up something while fixing clippy. I will fix.

iffyio (approved):
LGTM! Thanks @ayman-sigma!
cc @alamb

goldmedal:
Hi @ayman-sigma, @iffyio,  

I'm concerned about this change. Is `SELECT * FROM foo-123.bar` valid SQL or a real use case for BigQuery? 🤔  

Based on the [BigQuery documentation](https://cloud.google.com/bigquery/docs/datasets#dataset-naming), dataset names cannot contain spaces or special characters such as `-`, `&`, `@`, or `%`.  

This PR modifies the tokenizer for numbers, which breaks SQL (It parsed to a wrong result, see #1619 ) like:  
```sql
SELECT 0. AS c1
```  
This syntax is valid in BigQuery.  

Additionally, I'm concerned about unquoted hyphenated identifiers,  made in #1109 by @jmhain
I can't execute the following SQL on my BigQuery project:  
```sql
SELECT * FROM foo-bar
```  
Generally, `-` is not a valid character for identifiers. The table name should be quoted, like this:  
```sql
SELECT * FROM `foo-bar`
```  

Is there any documentation that explicitly covers this 🤔 ? The only relevant information I found was about [unquoted identifiers](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#unquoted_identifiers):  

> Must begin with a letter or an underscore (_) character.  
> Subsequent characters can be letters, numbers, or underscores (_).  

also c.c. @alamb who reviewed #1109

ayman-sigma:
> Hi @ayman-sigma, @iffyio,
> 
> I'm concerned about this change. Is `SELECT * FROM foo-123.bar` valid SQL or a real use case for BigQuery? 🤔
> 
> Based on the [BigQuery documentation](https://cloud.google.com/bigquery/docs/datasets#dataset-naming), dataset names cannot contain spaces or special characters such as `-`, `&`, `@`, or `%`.

I think my example here was bad. I meant to write something like `SELECT * FROM foo-123.mydataset.mytable` which is a valid table syntax according to the docs you mentioned above.

> This PR modifies the tokenizer for numbers, which breaks SQL (It parsed to a wrong result, see #1619 ) like:
> 
> ```sql
> SELECT 0. AS c1
> ```
> 
> This syntax is valid in BigQuery.

Sorry about breaking that case. I'm actually surprised we don't have a test case for that.
I think it easy to fix the case of `SELECT 1. AS c1`, but it will be tricky to handle `SELECT 1.e5` for example.

goldmedal:
> I think my example here was bad. I meant to write something like `SELECT * FROM foo-123.mydataset.mytable` which is a valid table syntax according to the docs you mentioned above.
> 

I see. So you mean the project name in BigQuery.  I think it backs to my another concern about unquoted hyphenated identifiers. I agree `foo-123.mydatset.mytable` is a valid table name but I don't think we can query it without quoting it in BigQuery.

I tried a similar case in BigQuery (sorry for the non-English UI but I think it's easy to know which one works)
Screenshot 2024-12-27 at 9 18 14 AM
and
Screenshot 2024-12-27 at 9 18 29 AM
or
Screenshot 2024-12-27 at 9 21 09 AM

Or did I miss something? Can we use it in BigQuery by enabling some configuration?

> Sorry about breaking that case. I'm actually surprised we don't have a test case for that. I think it easy to fix the case of `SELECT 1. AS c1`, but it will be tricky to handle `SELECT 1.e5` for example.
>

Never mind. I'm also surprised about it 😢. We should have tests to protect it (I'll do it in #1619)
I think the problem is: **should we allow unquoted hyphenated identifiers for BigQuery or any dialect by default?**

If it's a BigQuery official valid syntax, I prefer to limit it to be a BigQuery-specific behavior in the tokenizer.
I pretty sure other dialects don't allow this kind of identifier.

However, it's not a valid syntax for BigQUery but it's used by some downstream projects (maybe in your case?). I prefer to let it be an optional behavior for the dialect. Maybe add a method in the Dialect trait to control it. 🤔 

WDYT?

ayman-sigma:
> I see. So you mean the project name in BigQuery. I think it backs to my another concern about unquoted hyphenated identifiers. I agree `foo-123.mydatset.mytable` is a valid table name but I don't think we can query it without quoting it in BigQuery.
> 

I'm not sure about the table name, but I know for sure that project name allows unquoted hyphenated identifier for BigQuery. 

> 
> > Sorry about breaking that case. I'm actually surprised we don't have a test case for that. I think it easy to fix the case of `SELECT 1. AS c1`, but it will be tricky to handle `SELECT 1.e5` for example.
> 
> Never mind. I'm also surprised about it 😢. We should have tests to protect it (I'll do it in #1619) I think the problem is: **should we allow unquoted hyphenated identifiers for BigQuery or any dialect by default?**
> 
> If it's a BigQuery official valid syntax, I prefer to limit it to be a BigQuery-specific behavior in the tokenizer. I pretty sure other dialects don't allow this kind of identifier.
> 
> However, it's not a valid syntax for BigQUery but it's used by some downstream projects (maybe in your case?). I prefer to let it be an optional behavior for the dialect. Maybe add a method in the Dialect trait to control it. 🤔
> 
> WDYT?

I believe the original PR from @jmhain was scoped to BigQuery, so we should be fine there. We just need to fix the tokenizer again for your case. 
The tokenizer change in this PR allow parsing `foo-123.bar` as  `[Word('"foo"), Minus, Number("123"), Period, Word("bar")]` instead of `[Word("foo"), Minus, Number("123."), Word("bar")]`. I believe the former is more correct than the latter.
You can scope that change to BigQuery, but I prefer that we fix it to allow the `1.` and `1.e5`, ... cases.

goldmedal:
> I'm not sure about the table name, but I know for sure that project name allows unquoted hyphenated identifier for BigQuery.
> 

Indeed, I have confirmed it: the project name is allowed to be an unquoted hyphenated identifier. 👍 
Screenshot 2024-12-27 at 10 44 20 AM


> I believe the original PR from @jmhain was scoped to BigQuery, so we should be fine there. We just need to fix the tokenizer again for your case. The tokenizer change in this PR allow parsing `foo-123.bar` as `[Word('"foo"), Minus, Number("123"), Period, Word("bar")]` instead of `[Word("foo"), Minus, Number("123."), Word("bar")]`. I believe the former is more correct than the latter. You can scope that change to BigQuery, but I prefer that we fix it to allow the `1.` and `1.e5`, ... cases.
>

I see. I think it's a BigQuery-specific behavior, not a common rule for others. I will add a method `Dialect::support_unquoted_hyphenated_identifiers` to scope it. 

Thanks for your explanation. 🙇

alamb:
- BTW this seems to have caused a regression: https://github.com/apache/datafusion-sqlparser-rs/pull/1621. Thanks to @goldmedal and @iffyio for fixing

Issue #1583, opened 2024-12-08
sqlite analyze don't require a table keyword??


yuyang-ok:
https://docs.rs/sqlparser/0.52.0/sqlparser/ast/enum.Statement.html#variant.Analyze don't this struct contains the sqlite analyze statement,  and sqlite don't require keywords `table` ,
Is this a bug or ??

alamb:
I am not sure what you are asking here. If a syntax you are looking for is not supported, we would welcome a PR to add support and tests.

iffyio:
Resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/1599

alamb:
🎉

PR #1599, opened 2024-12-13
Support optional table for `ANALYZE` statement

yuyang-ok:
I am trying to fix issue https://github.com/apache/datafusion-sqlparser-rs/issues/1583 
I search the document, I think there only some dialect require the `table` keywords.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -851,7 +851,9 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_analyze(&mut self) -> Result<Statement, ParserError> {
-        self.expect_keyword(Keyword::TABLE)?;
+        if dialect_of!(self is MySqlDialect | DatabricksDialect | HiveDialect | SnowflakeDialect) {
+            self.expect_keyword(Keyword::TABLE)?;
+        }
         let table_name = self.parse_object_name(false)?;
         let mut for_columns = false;
         let mut cache_metadata = false;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -851,9 +851,7 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_analyze(&mut self) -> Result<Statement, ParserError> {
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -851,7 +851,9 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_analyze(&mut self) -> Result<Statement, ParserError> {
-        self.expect_keyword(Keyword::TABLE)?;
+        if dialect_of!(self is MySqlDialect | DatabricksDialect | HiveDialect | SnowflakeDialect) {
+            self.expect_keyword(Keyword::TABLE)?;
+        }
```
iffyio, line 856: in order to avoid flagging on individual dialects, could we flag on the statement whether the keyword was present in the sql? e.g.
```rust
let has_table_keyword = self.parse_keyword(Keyword::TABLE);
...
Ok(Statement::Analyze { has_table_keyword ...})
```
See [Statement::Explain](https://github.com/apache/datafusion-sqlparser-rs/blob/8507e452399605cfca98eed11ce4123a5e738333/src/ast/mod.rs#L3219) for example. Also [here](https://github.com/apache/datafusion-sqlparser-rs/blob/8507e452399605cfca98eed11ce4123a5e738333/src/ast/mod.rs#L3460-L3462) for where we serialize the SQL.

That would make it so that the input sql is identical to the output sql.

Also can we add some tests to `sqlparser_common.rs` demonstrate the behavior?
See [hive](https://github.com/apache/datafusion-sqlparser-rs/blob/8507e452399605cfca98eed11ce4123a5e738333/tests/sqlparser_hive.rs#L80-L84) for example

yuyang-ok: Ok , I will try

yuyang-ok: ok @iffyio  thanks.

iffyio (approved):
LGTM! Thanks @yuyang-ok 
cc @alamb

PR #1600, opened 2024-12-13
Support arbitrary composite access expressions

ayman-sigma:
Right now the `CompositeAccess` expression is limited to support the PostgreSQL syntax like `SELECT (on_hand.item).name FROM on_hand WHERE (on_hand.item).price > 9`.
This PR generalize it to support arbitrary composite access expressions.  `CompositeAccess` expression is any prefix expression followed by a period then an identifier. The `CompositeAccess` itself can also be nested.

This change fixes issues with dialects that can access fields from another expression. For example dialects that supports `struct` literal. Something like `SELECT struct(1 as a).a` wouldn't work before this PR.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -640,7 +640,7 @@ pub enum Expr {
         /// The path to the data to extract.
         path: JsonPath,
     },
-    /// CompositeAccess (postgres) eg: SELECT (information_schema._pg_expandarray(array['i','i'])).n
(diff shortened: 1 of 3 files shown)


iffyio (approved):
LGTM!
cc @alamb

alamb:
- BTW this seems to have caused a regression: https://github.com/apache/datafusion-sqlparser-rs/pull/1621. Thanks to @goldmedal and @iffyio for fixing

PR #1602, opened 2024-12-13
Add support for BigQuery `ANY TYPE` data type

### LINKED ISSUE #1601: Fails to parse BigQuery UDF with "ANY TYPE" parameters
EDIT: The "proper" name for this pattern is "Templated, SQL UDF Parameters", which can be further elaborated in the BQ Docs: https://cloud.google.com/bigquery/docs/user-defined-functions#templated-sql-udf-parameters

trying to parse this simple function like this: 
```
fn main() {
    let sql = "
CREATE TEMP FUNCTION my_function(are_map ANY TYPE)
AS (
  (SELECT 1)
);
    ";

    let ast = Parser::parse_sql(&BigQueryDialect, sql).unwrap();
    for stmt in &ast {
        println!("{:?}", stmt);
    }
}
```

Yields this error:
```
thread 'main' panicked at src/main.rs:18:56:
called `Result::unwrap()` on an `Err` value: ParserError("Expected: ), found: TYPE at Line: 2, Column: 46")
```
I can try to have a look at this , seems relatively straightforward to fix!

MartinSahlen:
As referenced in #1601 , here is a PR, with a test, that verifies correct parsing of BigQuery User-defined functions using templated SQL parameters (what a mouthful :D). 

More docs on the actual feature/functionality in BQ can be found here: https://cloud.google.com/bigquery/docs/user-defined-functions

Thanks in advance for the consideration!

EDIT: This closes #1601

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8382,6 +8382,10 @@ impl<'a> Parser<'a> {
                     Ok(DataType::Tuple(field_defs))
                 }
                 Keyword::TRIGGER => Ok(DataType::Trigger),
+                Keyword::ANY => {
+                    self.expect_keyword(Keyword::TYPE)?;
+                    Ok(DataType::AnyType)
+                }
                 _ => {
                     self.prev_token();
                     let type_name = self.parse_object_name(false)?;
@@ -8391,6 +8395,7 @@ impl<'a> Parser<'a> {
                         Ok(DataType::Custom(type_name, vec![]))
                     }
                 }
+
             },
             _ => self.expected("a data type name", next_token),
         }?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8395,7 +8395,6 @@ impl<'a> Parser<'a> {
                         Ok(DataType::Custom(type_name, vec![]))
                     }
                 }
-
             },
             _ => self.expected("a data type name", next_token),
         }?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8382,8 +8382,8 @@ impl<'a> Parser<'a> {
                     Ok(DataType::Tuple(field_defs))
                 }
                 Keyword::TRIGGER => Ok(DataType::Trigger),
-                Keyword::ANY => {
-                    self.expect_keyword(Keyword::TYPE)?;
+                Keyword::ANY if self.peek_keyword(Keyword::TYPE) => {
+                    let _ = self.parse_keyword(Keyword::TYPE);
                     Ok(DataType::AnyType)
                 }
                 _ => {
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -2099,11 +2099,11 @@ fn test_bigquery_create_function() {
         ),
         // Templated
         concat!(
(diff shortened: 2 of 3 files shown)


src/parser/mod.rs
```diff
@@ -8382,6 +8382,10 @@ impl<'a> Parser<'a> {
                     Ok(DataType::Tuple(field_defs))
                 }
                 Keyword::TRIGGER => Ok(DataType::Trigger),
+                Keyword::ANY => {
```
iffyio, line 8385:
```suggestion
                Keyword::ANY if self.peek_keyword(Keyword::TYPE) => {
```
Thinking we can add this guard to ensure that it doesn't conflict with a custom data type that happens to be called `ANY`?

tests/sqlparser_bigquery.rs
```diff
@@ -2097,6 +2097,14 @@ fn test_bigquery_create_function() {
...
             "OPTIONS(a = [1, 2])",
         ),
+        // Templated
+        concat!(
+            "CREATE OR REPLACE TEMPORARY FUNCTION ",
+            "my_function(param1 ANY TYPE) ",
+            "AS (",
+            "(SELECT 1)",
+            ")",
+        ),
```
iffyio, line 2107: Since the functionality itself is the data type, I'm thinking we can use a dedicated test for it? e.g `fn test_parse_any_type()`? Also can we add a test case to demonstrate the behavior in the other comment what happens in a sql like `CREATE TABLE foo (x ANY);`

iffyio (commented):
Thanks @MartinSahlen! The changes look reasonable to me, left a couple minor comments

tests/sqlparser_bigquery.rs
MartinSahlen: Good point, I was in fact doing that but reverted to inlining the test with other UDFs. Not being sure what was the preferred pattern but makes sense!

src/parser/mod.rs
MartinSahlen: Ah yes that is much more in line with the desired / intended behavior. As the current approach would result in a parsing error should we come across an ANY type I think.

MartinSahlen: I tried implementing this change but it seems to break. hmm

MartinSahlen: As this is the first time I'm touching the codebase, I needed to understand the general mechanisms. Your change worked in conjunction with `let _ = self.parse_keyword(Keyword::TYPE);`, which makes intitutive sense after looking at the code in general.

MartinSahlen:
> Thanks @MartinSahlen! The changes look reasonable to me, left a couple minor comments

Thanks for the review and suggestions! I have now added the suggested changes and improvements, along with two seperate tests
- test that the usage of "ANY TYPE" works in a function definition for BigQuery
- test that the usage of the "ANY" type still works as a standalone type to support arbitrary or user-defined types without being restrictive. This also makes a lot of sense as the goal of this codebase is not to provide an opinionated or semantic validating parser.

iffyio (approved):
LGTM! Thanks @MartinSahlen!
cc @alamb

MartinSahlen:
> LGTM! Thanks @MartinSahlen! cc @alamb

Thanks!

PR #1603, opened 2024-12-14
Avoid cloning during token creation

alamb:
# Rationale
Inspired by @davisp's https://github.com/apache/datafusion-sqlparser-rs/pull/1591 I was looking at `Token::make_word` and I noticed it made *2* clones (owned strings). Each copy slows things down so let's avoid them

# Changes
1. Change `Token::make_word` to reuse its argument and thus save a clone

# API Changes
This changes Token::make_word and Token::make_keyword to take `String` instead of `&str`. This would be a breaking API change. If it is a problem I can make owned variants (like `Token::make_word_owned` or something) 🤔 

# Performance benchmarks: 

```
++ critcmp main alamb_faster_keyword_lookup
group                                                    alamb_faster_keyword_lookup            main
-----                                                    ---------------------------            ----
sqlparser-rs parsing benchmark/sqlparser::select         1.00      6.4±0.03µs        ? ?/sec    1.00      6.4±0.03µs        ? ?/sec
sqlparser-rs parsing benchmark/sqlparser::with_select    1.02     31.5±0.10µs        ? ?/sec    1.00     31.0±0.13µs        ? ?/sec
```

🤔  it appears not to make much/any difference

### DIFF
--- a/src/keywords.rs
+++ b/src/keywords.rs
@@ -70,7 +70,8 @@ macro_rules! define_keywords {
     };
 }
 
-// The following keywords should be sorted to be able to match using binary search
+// Notes: The following keywords should be sorted to be able to match using
+// binary search
 define_keywords!(
     ABORT,
     ABS,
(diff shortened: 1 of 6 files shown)


alamb:
I was looking at our benchmarks, and they are very limited. If we want to pursue optimizing sqlparser more I think it would probably be a good idea to add many  queries (like TPCDS and TPCH) into the benchmarks

alamb:
We have begin work on the real thing, so closing this draft

PR #1604, opened 2024-12-16
Support Snowflake Update-From-Select

yuval-illumex:
The following query works in Snowflake:

`UPDATE
    t1
FROM
    (
        SELECT
            name,
            id
        FROM
            t1
        GROUP BY
            id
    ) AS t2
SET
    name = t2.name
WHERE
    t1.id = t2.id
`

Today the parser only supports update-set-from

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2791,13 +2791,15 @@ impl fmt::Display for ValueTableMode {
     }
 }
 
-/// The update table from options
+/// The `FROM` clause of an `UPDATE TABLE` statement
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum UpdateTableFromKind {
-    /// Table sample located before the table alias option
+    /// Update Statment where the 'FROM' clause is before the 'SET' keyword (Supported by Snowflake)
+    /// For Example: UPDATE FROM t1 SET t1.name='aaa'
     BeforeSet(TableWithJoins),
-    /// Table sample located after the table alias option
+    /// Update Statment where the 'FROM' clause is after the 'SET' keyword (Which is the standard way)
+    /// For Example: UPDATE SET t1.name='aaa' FROM t1
     AfterSet(TableWithJoins),
 }
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2797,9 +2797,9 @@ impl fmt::Display for ValueTableMode {
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum UpdateTableFromKind {
     /// Update Statment where the 'FROM' clause is before the 'SET' keyword (Supported by Snowflake)
-    /// For Example: UPDATE FROM t1 SET t1.name='aaa'
+    /// For Example: `UPDATE FROM t1 SET t1.name='aaa'`
     BeforeSet(TableWithJoins),
     /// Update Statment where the 'FROM' clause is after the 'SET' keyword (Which is the standard way)
-    /// For Example: UPDATE SET t1.name='aaa' FROM t1
+    /// For Example: `UPDATE SET t1.name='aaa' FROM t1`
     AfterSet(TableWithJoins),
 }
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2788,3 +2788,14 @@ impl fmt::Display for ValueTableMode {
         }
     }
 }
+
+/// The update table from options
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum UpdateTableFromKind {
+    /// Table sample located before the table alias option
+    BeforeSet(TableWithJoins),
+    /// Table sample located after the table alias option
+    AfterSet(TableWithJoins),
+}
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -32,8 +32,8 @@ use super::{
     PivotValueSource, ProjectionSelect, Query, ReferentialAction, RenameSelectItem,
     ReplaceSelectElement, ReplaceSelectItem, Select, SelectInto, SelectItem, SetExpr, SqlOption,
     Statement, Subscript, SymbolDefinition, TableAlias, TableAliasColumnDef, TableConstraint,
-    TableFactor, TableOptionsClustered, TableWithJoins, Use, Value, Values, ViewColumnDef,
-    WildcardAdditionalOptions, With, WithFill,
+    TableFactor, TableOptionsClustered, TableWithJoins, UpdateTableFromKind, Use, Value, Values,
+    ViewColumnDef, WildcardAdditionalOptions, With, WithFill,
 };
 
 /// Given an iterator of spans, return the [Span::union] of all spans.
@@ -2095,6 +2095,15 @@ impl Spanned for SelectInto {
         } = self;
 
         name.span()
+    }
+}
+
+impl Spanned for UpdateTableFromKind {
+    fn span(&self) -> Span {
+        match self {
+            UpdateTableFromKind::BeforeSet(vec) => vec.span(),
+            UpdateTableFromKind::AfterSet(vec) => vec.span(),
+        }
     }
 }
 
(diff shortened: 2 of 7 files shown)


src/parser/mod.rs
```diff
@@ -11679,14 +11679,21 @@ impl<'a> Parser<'a> {
...
+        let from_before_set = if self.parse_keyword(Keyword::FROM)
+            && dialect_of!(self is GenericDialect | PostgreSqlDialect | DuckDbDialect | BigQueryDialect | SnowflakeDialect | RedshiftSqlDialect | MsSqlDialect | SQLiteDialect )
+        {
+            Some(self.parse_table_and_joins()?)
+        } else {
+            None
+        };
         self.expect_keyword(Keyword::SET)?;
         let assignments = self.parse_comma_separated(Parser::parse_assignment)?;
         let from = if self.parse_keyword(Keyword::FROM)
```
iffyio, line 11691: thinking here in order to be correct we might need to check `if from_before_set.is_none() && self.parse_keyword(FROM) ...`? I'm assuming that the intention is that it is invalid sql to provide both?

src/parser/mod.rs
```diff
@@ -11679,14 +11679,21 @@ impl<'a> Parser<'a> {
...
             && dialect_of!(self is GenericDialect | PostgreSqlDialect | DuckDbDialect | BigQueryDialect | SnowflakeDialect | RedshiftSqlDialect | MsSqlDialect | SQLiteDialect )
         {
             Some(self.parse_table_and_joins()?)
         } else {
-            None
+            from_before_set
```
iffyio, line 11696: Could we instead introduce a wrapper type to encode the position information?
Something like
```rust
enum UpdateTableFromKind {
     BeforeSet(TableWithJoins),
     AfterSet(TableWithJoins)
}
...
Statement::Update {... from: Option<UpdateTableFromKind> }
```
See [TableSampleKind](https://github.com/yoavcloud/datafusion-sqlparser-rs/blob/eae5629fb86f5e262063c0bc99ff628a5855168f/src/ast/query.rs#L1157-L1162) for an example on this approach.

That would ensure that we can produce the same sql as in the input.

tests/sqlparser_snowflake.rs
```diff
@@ -2974,3 +2974,10 @@ fn test_table_sample() {
     snowflake_and_generic().verified_stmt("SELECT id FROM mytable TABLESAMPLE (10) REPEATABLE (1)");
     snowflake_and_generic().verified_stmt("SELECT id FROM mytable TABLESAMPLE (10) SEED (1)");
 }
+
+#[test]
+fn parse_update_from_before_select() {
```
iffyio, line 2979: note if we drop the dialect guard, we can move the tests to sqlparser_common.rs to cover all dialects.

Also could we add a test case demonstrating the behavior mentioned above where we have two `FROM` clauses? e.g.
```sql
UPDATE t1 FROM (SELECT ... ) AS t2 SET name = t2.name FROM (SELECT ...) AS t2;
```

src/parser/mod.rs
yuval-illumex: I will join your assumption, as I really don't know if it's possible or not.

src/parser/mod.rs
yuval-illumex: @iffyio My problem with your approach that it will break backwards compatibility, and I don't feel it "worth it" in this case. What do you think?

yuval-illumex:
Thank you @iffyio for your feedback, I asked your opinion in one of the comments.

src/parser/mod.rs
iffyio: Ah yeah I think that would be ok in this case and the change required downstream would be mechanical

yuval-illumex: @iffyio I will change that but I have to say that I disagree. I have a system on top of sqlparser and every kind of this braking change is causing bugs and takes time to migrate.

yuval-illumex: @alamb fyi

yuval-illumex:
@iffyio appreciate your feedback again 🙏 (Tests that are failing are not related to my PR):

https://github.com/apache/datafusion-sqlparser-rs/actions/runs/12455101228/job/34767107925

tests/sqlparser_common.rs
```diff
@@ -12577,9 +12577,20 @@ fn overflow() {
...
+#[test]
+fn parse_update_from_before_select() {
+    all_dialects()
+    .verified_stmt("UPDATE t1 FROM (SELECT name, id FROM t1 GROUP BY id) AS t2 SET name = t2.name WHERE t1.id = t2.id");
+
+    let res = all_dialects().parse_sql_statements(
+        "UPDATE t1 FROM (SELECT name, id FROM t1 GROUP BY id) AS t2 SET name = t2.name FROM (SELECT name from t2) AS t2",
+    );
+
+    assert!(res.is_err());
```
iffyio, line 12595: could we assert in [this style](https://github.com/apache/datafusion-sqlparser-rs/blob/6cdf66457006a3fbfc569dc262e78f404522821b/tests/sqlparser_common.rs#L219-L235) to expect a specific error message?

src/ast/query.rs
```diff
@@ -2790,3 +2790,14 @@ impl fmt::Display for ValueTableMode {
         }
     }
 }
+
+/// The update table from options
```
iffyio, line 2794:
```suggestion
/// The `FROM` clause of an `UPDATE TABLE` statement
```

src/ast/query.rs
```diff
@@ -2790,3 +2790,14 @@ impl fmt::Display for ValueTableMode {
...
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum UpdateTableFromKind {
+    /// Table sample located before the table alias option
+    BeforeSet(TableWithJoins),
+    /// Table sample located after the table alias option
+    AfterSet(TableWithJoins),
```
iffyio, line 2802: Could we include example sql with each variant? thinking that would help folks differentiate easily.
Also the descriptions sound incorrect to be referencing table sample/alias?

src/keywords.rs
```diff
@@ -941,6 +941,7 @@ pub const RESERVED_FOR_TABLE_ALIAS: &[Keyword] = &[
     // Reserved for Snowflake table sample
     Keyword::SAMPLE,
     Keyword::TABLESAMPLE,
+    Keyword::FROM,
```
iffyio, line 944: is this change related to the PR (wasn't clear to me if so)?

iffyio (commented):
Left a couple comments, otherwise this looks good to me overall!

src/keywords.rs
yuval-illumex: Yes.
When removing this as a keyword the following error occurs:
`Error during parsing: ParserError("Expected: ), found: t1 at Line: 9, Column: 13")`

src/ast/query.rs
```diff
@@ -2790,3 +2790,16 @@ impl fmt::Display for ValueTableMode {
...
+/// The `FROM` clause of an `UPDATE TABLE` statement
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum UpdateTableFromKind {
+    /// Update Statment where the 'FROM' clause is before the 'SET' keyword (Supported by Snowflake)
+    /// For Example: UPDATE FROM t1 SET t1.name='aaa'
+    BeforeSet(TableWithJoins),
+    /// Update Statment where the 'FROM' clause is after the 'SET' keyword (Which is the standard way)
+    /// For Example: UPDATE SET t1.name='aaa' FROM t1
```
iffyio, line 2803:
```suggestion
    /// For Example: `UPDATE FROM t1 SET t1.name='aaa'`
    BeforeSet(TableWithJoins),
    /// Update Statment where the 'FROM' clause is after the 'SET' keyword (Which is the standard way)
    /// For Example: `UPDATE SET t1.name='aaa' FROM t1`
```

src/ast/spans.rs
```diff
@@ -2106,6 +2106,15 @@ impl Spanned for SelectInto {
     }
 }
 
+impl Spanned for UpdateTableFromKind {
+    fn span(&self) -> Span {
+        match self {
+            UpdateTableFromKind::BeforeSet(vec) => vec.span(),
+            UpdateTableFromKind::AfterSet(vec) => vec.span(),
```
iffyio, line 2113:
```suggestion
            UpdateTableFromKind::BeforeSet(from) => from.span(),
            UpdateTableFromKind::AfterSet(from) => from.span(),
```

iffyio (approved):
LGTM! Thanks @yuval-illumex!
cc @alamb

PR #1606, opened 2024-12-16
Add the alter table ON COMMIT option to Snowflake

yoavcloud:
It doesn't seem like this option is documented. However, Snowflake supports it, and we see such queries "in the wild."

Syntax:
`CREATE TABLE ... ON COMMIT PRESERVE ROWS`
`CREATE TABLE ... ON COMMIT DELETE ROWS`
`CREATE TABLE ... ON COMMIT DROP`

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6215,7 +6215,7 @@ impl<'a> Parser<'a> {
             .build())
     }
 
-    pub fn parse_create_table_on_commit(&mut self) -> Result<OnCommit, ParserError> {
+    pub(crate) fn parse_create_table_on_commit(&mut self) -> Result<OnCommit, ParserError> {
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -6226,6 +6215,21 @@ impl<'a> Parser<'a> {
             .build())
     }
 
+    pub fn parse_create_table_on_commit(&mut self) -> Result<OnCommit, ParserError> {
```
iffyio, line 6218:
```suggestion
    fn parse_create_table_on_commit(&mut self) -> Result<OnCommit, ParserError> {
```

iffyio (approved):
LGTM! cc @alamb

src/parser/mod.rs
iffyio: Ah right that sounds good!

PR #1608, opened 2024-12-18
Allow foreign table constraint without columns

ramnivas:
Postgres allows skipping the `(referenced-columns)` part in `ALTER TABLE ... FOREIGN KEY .. REFERENCES <tablename> (referenced-columns)` per https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES.

Since the generic SQL grammar requires referenced columns (https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#referential-constraint-definition), this change applies only to Postgres.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6830,7 +6830,15 @@ impl<'a> Parser<'a> {
                 let columns = self.parse_parenthesized_column_list(Mandatory, false)?;
                 self.expect_keyword(Keyword::REFERENCES)?;
                 let foreign_table = self.parse_object_name(false)?;
-                let referred_columns = self.parse_parenthesized_column_list(Mandatory, false)?;
+                // PostgreSQL allows foreign key columns to be optional
+                // (https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES)
+                let parenthesized_column_list_optional = if dialect_of!(self is PostgreSqlDialect) {
+                    Optional
+                } else {
+                    Mandatory
+                };
+                let referred_columns = self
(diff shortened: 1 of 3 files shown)


ramnivas:
An alternative design would be to add a new method in the `Dialect` trait with a default implementation and have `PostgreSqlDialect` override it. But, I see `dialect_of!(self is ...` usages , so I chose that. Let me know if I should implement that alternative.

src/parser/mod.rs
```diff
@@ -6830,7 +6830,15 @@ impl<'a> Parser<'a> {
                 let columns = self.parse_parenthesized_column_list(Mandatory, false)?;
                 self.expect_keyword(Keyword::REFERENCES)?;
                 let foreign_table = self.parse_object_name(false)?;
-                let referred_columns = self.parse_parenthesized_column_list(Mandatory, false)?;
+                // PostgreSQL allows foreign key columns to be optional
+                // (https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES)
+                let parenthesized_column_list_optional = if dialect_of!(self is PostgreSqlDialect) {
```
iffyio, line 6835: Since this behavior is a superset of the previous I think we can do without the dialect check and always set to `Optional`?

Also could we add a test case to [sqlparse_common.rs](https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_common.rs) demonstrating the new behavior?

ramnivas: Good point. Made the change.

iffyio (approved):
LGTM! Thanks @ramnivas!
cc @alamb

PR #1610, opened 2024-12-18
Add support for DROP EXTENSION

ramnivas:
```
DROP EXTENSION [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]
```

https://www.postgresql.org/docs/current/sql-dropextension.html

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2759,6 +2759,15 @@ pub enum Statement {
         version: Option<Ident>,
     },
     /// ```sql
+    /// DROP EXTENSION [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]
+    /// ```
+    DropExtension {
+        names: Vec<Ident>,
+        if_exists: bool,
+        /// `CASCADE` or `RESTRICT`
+        cascade_or_restrict: Option<ReferentialAction>,
+    },
+    /// ```sql
     /// FETCH
     /// ```
     /// Retrieve rows from a query using a cursor
@@ -4023,6 +4032,21 @@ impl fmt::Display for Statement {
                     }
                 }
 
+                Ok(())
+            }
+            Statement::DropExtension {
+                names,
+                if_exists,
+                cascade_or_restrict,
+            } => {
+                write!(f, "DROP EXTENSION")?;
+                if *if_exists {
+                    write!(f, " IF EXISTS")?;
+                }
+                write!(f, " {}", display_comma_separated(names))?;
+                if let Some(cascade_or_restrict) = cascade_or_restrict {
+                    write!(f, " {cascade_or_restrict}")?;
+                }
                 Ok(())
             }
             Statement::CreateRole {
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -2759,6 +2759,15 @@ pub enum Statement {
         version: Option<Ident>,
     },
     /// ```sql
+    /// DROP EXTENSION [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]
+    /// ```
```
iffyio, line 2763: Could we also include the link in the description here? would help folks find the syntax quicker in the future if needed

src/parser/mod.rs
```diff
@@ -5861,6 +5863,22 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_drop_extension(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 5866: Since the function is being made a public function, could we include a description for it?

src/parser/mod.rs
```diff
@@ -5861,6 +5863,22 @@ impl<'a> Parser<'a> {
...
+        let names = self.parse_comma_separated(|p| p.parse_identifier(false))?;
+        let cascade_or_restrict =
+            self.parse_one_of_keywords(&[Keyword::CASCADE, Keyword::RESTRICT]);
+        Ok(Statement::DropExtension {
+            names,
+            if_exists,
+            cascade_or_restrict: cascade_or_restrict.map(|k| match k {
+                Keyword::CASCADE => ReferentialAction::Cascade,
+                Keyword::RESTRICT => ReferentialAction::Restrict,
+                _ => unreachable!(),
```
iffyio, line 5877: in place of panicking here in case we missed something/bug, we can return an explicit error `return self.expected(..)`?

tests/sqlparser_postgres.rs
```diff
@@ -662,6 +662,22 @@ fn parse_create_extension() {
         .verified_stmt("CREATE EXTENSION extension_name WITH SCHEMA schema_name VERSION version");
 }
 
+#[test]
+fn parse_drop_extension() {
+    pg_and_generic().verified_stmt("DROP EXTENSION extension_name");
```
iffyio, line 667: Since the feature introduces a new node to the ast can we add one scenario that asserts the ast in [this style](https://github.com/apache/datafusion-sqlparser-rs/blob/04d8502677a85b2c166aabedf25c6bc94a039a0b/tests/sqlparser_postgres.rs#L567-L579)?

src/ast/mod.rs
ramnivas: Done!

src/parser/mod.rs
ramnivas: Added

src/parser/mod.rs
ramnivas: Done

tests/sqlparser_postgres.rs
ramnivas: Done, but in a slightly different style (more succinct), and did so for all scenarios.

iffyio (approved):
LGTM! Thanks @ramnivas!
cc @alamb

ramnivas:
Thanks @iffyio for your review!

PR #1613, opened 2024-12-20
Handle empty projection in Postgres SELECT  statements

tobyhede:
Handle empty projections in Postgres and the generic dialect

Statements in the form `SELECT FROM table` are valid in Postgres

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9604,7 +9604,13 @@ impl<'a> Parser<'a> {
             top = Some(self.parse_top()?);
         }
 
-        let projection = self.parse_projection()?;
+        let projection = if dialect_of!(self is PostgreSqlDialect | GenericDialect)
+            && self.peek_keyword(Keyword::FROM)
+        {
+            vec![]
+        } else {
+            self.parse_projection()?
+        };
 
         let into = if self.parse_keyword(Keyword::INTO) {
             let temporary = self
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9604,13 +9604,12 @@ impl<'a> Parser<'a> {
             top = Some(self.parse_top()?);
         }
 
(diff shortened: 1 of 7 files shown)


freshtonic:
Nice work @tobyhede - that will unblock use of database adapters like Elixir's `postgrex` with CipherStash Proxy.

`postgrex` uses `SELECT FROM <table>` to assert a table's existence and currently the proxy chokes on the SQL parsing.

src/parser/mod.rs
```diff
@@ -9604,7 +9604,13 @@ impl<'a> Parser<'a> {
             top = Some(self.parse_top()?);
         }
 
-        let projection = self.parse_projection()?;
+        let projection = if dialect_of!(self is PostgreSqlDialect | GenericDialect)
```
iffyio, line 9607: could we use a dialect method for the feature flag? so that this becomes e.g. `if self.dialect.supports_empty_projections() && self.peek_keyword(Keyword::FROM)`

tests/sqlparser_postgres.rs
```diff
@@ -5195,3 +5195,8 @@ fn parse_bitstring_literal() {
         ))]
     );
 }
+
+#[test]
+fn parse_select_without_projection() {
```
iffyio, line 5200: if we use a dialect method suggesed above we'd likely want to move this to common and be able to write the  test using [all_dialects_where()](https://github.com/apache/datafusion-sqlparser-rs/blob/bcdc790b0135dc64ec4d40eaa837316530a5f5f3/tests/sqlparser_common.rs#L4981)

tobyhede:
Hadn't used the dialect flags before, is much cleaner. Thanks @iffyio

iffyio (approved):
LGTM! Thanks @tobyhede! 
cc @alamb

alamb:
Thanks @tobyhede  and @iffyio

PR #1615, opened 2024-12-22
Merge composite and compound expr test cases

iffyio:
Fixes test failure on main due to conflicting representations of compound expressions after landing from #1600 and #1551


iffyio:
cc @alamb

FYI @ayman-sigma @goldmedal

goldmedal (approved):
Thanks @iffyio 👍

alamb:
❤️

Issue #1582, opened 2024-12-08
The syntax of mysql RENAME TABLE tb1 TO tb2 is not supported.


charmfocus:
```rust
    let sql = "RENAME TABLE `test`.`test` TO `test`.`test2`";

    let dialect = MySqlDialect {};
    let ast = Parser::parse_sql(&dialect, sql).unwrap();

    println!("AST: {:#?}", ast);
```

output error:
```
called `Result::unwrap()` on an `Err` value: ParserError("Expected: an SQL statement, found: RENAME at Line: 1, Column: 1")
```

LucaCappelletti94:
Closing as completed in https://github.com/apache/datafusion-sqlparser-rs/pull/1616

PR #1616, opened 2024-12-24
Add support for MYSQL's `RENAME TABLE`

wugeer:
This PR supports `RENAME TABLE` clause for mysql dialect,. For more information, please refer to:
https://dev.mysql.com/doc/refman/9.1/en/rename-table.html

This resolves issue https://github.com/apache/datafusion-sqlparser-rs/issues/1582

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4947,12 +4947,8 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
-            Statement::RenameTable(rename_object_defs) => {
-                write!(
-                    f,
-                    "RENAME TABLE {}",
-                    display_comma_separated(rename_object_defs)
-                )
+            Statement::RenameTable(rename_tables) => {
+                write!(f, "RENAME TABLE {}", display_comma_separated(rename_tables))
             }
         }
     }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3407,7 +3407,7 @@ pub enum Statement {
     /// Renames one or more tables
     ///
     /// See Mysql <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
-    RenameTable(Vec<RenameObjectDef>),
+    RenameTable(Vec<RenameTable>),
 }
 
 impl fmt::Display for Statement {
@@ -7660,12 +7660,12 @@ impl Display for JsonNullClause {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub struct RenameObjectDef {
+pub struct RenameTable {
     pub old_name: ObjectName,
     pub new_name: ObjectName,
 }
 
-impl fmt::Display for RenameObjectDef {
+impl fmt::Display for RenameTable {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{} TO {}", self.old_name, self.new_name)?;
         Ok(())
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3404,10 +3404,10 @@ pub enum Statement {
     /// ```sql
     /// Rename TABLE tbl_name TO new_tbl_name[, tbl_name2 TO new_tbl_name2] ...
     /// ```
-    /// renames one or more tables
+    /// Renames one or more tables
     ///
     /// See Mysql <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
-    RenameTable { operations: Vec<RenameObjectDef> },
+    RenameTable(Vec<RenameObjectDef>),
 }
 
 impl fmt::Display for Statement {
@@ -4947,8 +4947,12 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
-            Statement::RenameTable { operations } => {
-                write!(f, "RENAME TABLE {}", display_comma_separated(operations))
+            Statement::RenameTable(rename_object_defs) => {
+                write!(
+                    f,
+                    "RENAME TABLE {}",
+                    display_comma_separated(rename_object_defs)
+                )
             }
         }
     }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3401,6 +3401,13 @@ pub enum Statement {
         partitioned: Option<Vec<Expr>>,
         table_format: Option<HiveLoadDataFormat>,
     },
+    /// ```sql
+    /// Rename TABLE tbl_name TO new_tbl_name[, tbl_name2 TO new_tbl_name2] ...
+    /// ```
+    /// renames one or more tables
+    ///
+    /// See Mysql <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
+    RenameTable { operations: Vec<RenameObjectDef> },
 }
 
 impl fmt::Display for Statement {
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -6682,6 +6694,14 @@ impl<'a> Parser<'a> {
         Ok(Tag::new(name, value))
     }
 
+    pub(crate) fn parse_rename_object_def(&mut self) -> Result<RenameObjectDef, ParserError> {
```
iffyio, line 6697:
```suggestion
    fn parse_rename_object_def(&mut self) -> Result<RenameObjectDef, ParserError> {
```
maybe better since this method is small and only used once, we can inline it into the callsite instead of having it as a standalone method? e.g.
```rust
let operations = self.parse_comma_separated(|p| {
  let old_name = self.parse_object_name(false)?;
  self.expect_keyword(Keyword::TO)?;
  let new_name = self.parse_object_name(false)?;
  Ok((old_name, new_name))
})?;
```

src/dialect/mod.rs
```diff
@@ -715,6 +715,11 @@ pub trait Dialect: Debug + Any {
         true
     }
 
+    // Returns true if the dialect supports the `RENAME TABLE` statement
+    fn supports_rename_table(&self) -> bool {
+        false
+    }
```
iffyio, line 721: I'm thinking it could be easier to skip the dialect method in this case and have the parser always accept a rename table statement. since this is a full statement I imagine its not bound to conflict with other dialects

src/ast/mod.rs
```diff
@@ -3401,6 +3401,13 @@ pub enum Statement {
         partitioned: Option<Vec<Expr>>,
         table_format: Option<HiveLoadDataFormat>,
     },
+    /// ```sql
+    /// Rename TABLE tbl_name TO new_tbl_name[, tbl_name2 TO new_tbl_name2] ...
+    /// ```
+    /// renames one or more tables
+    ///
+    /// See Mysql <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
+    RenameTable { operations: Vec<RenameObjectDef> },
```
iffyio, line 3410: Could we do something like?
```rust
struct RenameTable {
    old_table: ObjectName,
    new_table: ObjectName
}
Statement::RenameTable(RenameTable)
```
using a dedicated struct makes it easier to extend (e.g with spans later on) and for downstream crates to work with the statement

tests/sqlparser_common.rs
```diff
@@ -4026,6 +4026,120 @@ fn parse_alter_table() {
...
+    match dialects.verified_stmt("RENAME TABLE `test`.`test1` TO `test_db`.`test2`") {
+        Statement::RenameTable { operations } => {
+            assert_eq!(
+                vec![RenameObjectDef {
+                    old_name: ObjectName(vec![
+                        Ident {
+                            value: "test".to_string(),
+                            quote_style: Some('`'),
+                            span: Span::empty(),
+                        },
```
iffyio, line 4042: in order to shorten the test case can we replace occurrences of `Ident {...}` with `Ident::new()` and `Ident::with_quote()` where applicable?

src/ast/mod.rs
```diff
@@ -3401,6 +3401,13 @@ pub enum Statement {
         partitioned: Option<Vec<Expr>>,
         table_format: Option<HiveLoadDataFormat>,
     },
+    /// ```sql
+    /// Rename TABLE tbl_name TO new_tbl_name[, tbl_name2 TO new_tbl_name2] ...
+    /// ```
+    /// renames one or more tables
```
iffyio, line 3407:
```suggestion
    /// Renames one or more tables
```

src/parser/mod.rs
```diff
@@ -1044,6 +1046,16 @@ impl<'a> Parser<'a> {
         Ok(Statement::NOTIFY { channel, payload })
     }
 
+    pub fn parse_rename(&mut self) -> Result<Statement, ParserError> {
+        if self.dialect.supports_rename_table() && self.peek_keyword(Keyword::TABLE) {
+            self.expect_keyword(Keyword::TABLE)?;
+            let operations = self.parse_comma_separated(Parser::parse_rename_object_def)?;
+            Ok(Statement::RenameTable { operations })
+        } else {
+            self.expected("an object type after RENAME", self.peek_token())
```
iffyio, line 1055: does it make sense to say e.g. 'table name' instead of 'object type'? Thinking the object type/def terminology in the sounds a bit generic if we're only dealing with tables in the context of this statement

tests/sqlparser_common.rs
wugeer: Good idea! Have Done.

src/ast/mod.rs
wugeer: Good idea! Have Done.

src/parser/mod.rs
wugeer: Good idea! Have Done.

src/parser/mod.rs
wugeer: Good idea! Have Done.

src/dialect/mod.rs
wugeer: Although many dialects support the syntax for rename table, they mainly use ` ALTER TABLE .... RENAME ...`. The method here refers to the switch of the `RENAME TABLE` statement. Therefore, I think it is necessary to keep this function.

iffyio: Ah so its not only a matter of whether the dialect supports the syntax in this case, I also meant that in order to avoid introducing the `supports_rename_table` flag, we can allow the parser always accept the `RENAME TABLE` syntax. Yes it will indeed mean that all dialects accept the syntax but that is fine given that it does not conflict with any other syntax the dialect supports

src/ast/mod.rs
```diff
@@ -3401,6 +3401,13 @@ pub enum Statement {
         partitioned: Option<Vec<Expr>>,
         table_format: Option<HiveLoadDataFormat>,
     },
+    /// ```sql
+    /// Rename TABLE tbl_name TO new_tbl_name[, tbl_name2 TO new_tbl_name2] ...
+    /// ```
+    /// Renames one or more tables
+    ///
+    /// See Mysql <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
+    RenameTable(Vec<RenameObjectDef>),
```
iffyio, line 3410:
```suggestion
    RenameTable(Vec<RenameTable>),
```
Could we call `RenameTable`? that would match the variant name, and the object def terminology I think would become too specific or no longer descriptive if other fields are added to the struct

src/dialect/mod.rs
wugeer: Done!

src/ast/mod.rs
wugeer: Done!

src/parser/mod.rs
```diff
@@ -547,6 +547,8 @@ impl<'a> Parser<'a> {
                 // `PRAGMA` is sqlite specific https://www.sqlite.org/pragma.html
                 Keyword::PRAGMA => self.parse_pragma(),
                 Keyword::UNLOAD => self.parse_unload(),
+                // `RENAME TABLE` is mysql specific https://dev.mysql.com/doc/refman/9.1/en/rename-table.html
```
iffyio, line 550:
```suggestion
```

src/parser/mod.rs
```diff
@@ -1044,6 +1046,24 @@ impl<'a> Parser<'a> {
         Ok(Statement::NOTIFY { channel, payload })
     }
 
+    /// Parses a `RENAME TABLE` statement. See [Statement::RenameTable]
+    /// see <https://dev.mysql.com/doc/refman/9.1/en/rename-table.html>
```
iffyio, line 1050:
```suggestion
```

iffyio (approved):
LGTM!
cc @alamb

PR #1587, opened 2024-12-11
Improve parsing performance by reducing token cloning

davisp:
This basically just takes the approach proposed by @alamb in #1561 and runs with it.

- Part of https://github.com/apache/datafusion-sqlparser-rs/issues/1558

The main changes involved:

1. A number of token parsing methods introduced that return a `&TokenWithSpan` instead of a cloned `TokenWithSpan`.
2. A bunch of updates to make use of these new ref returning functions.
3. Refactoring some of the core peek/expect/consume methods to avoid unnecessary clones
4. Replace all but one use of expect_keyword with expect_keyword_is.

Results on my M1 MacBook Pro:

```
# main:00abaf218

❯ cargo bench -- --save-baseline main
    Finished `bench` profile [optimized] target(s) in 0.11s
     Running benches/sqlparser_bench.rs (target/release/deps/sqlparser_bench-9f7a6e6e193d8f5c)
sqlparser-rs parsing benchmark/sqlparser::select
                        time:   [4.2036 µs 4.2197 µs 4.2372 µs]
                        change: [-2.4618% -2.1421% -1.8071%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
Benchmarking sqlparser-rs parsing benchmark/sqlparser::with_select: Collecting 100 samples in estimated 5.0715 s (278k iteratio
sqlparser-rs parsing benchmark/sqlparser::with_select
                        time:   [18.360 µs 18.425 µs 18.486 µs]
                        change: [-5.3763% -2.5663% -0.8662%] (p = 0.02 < 0.05)
                        Change within noise threshold.

# reduce-token-cloning:1ffa2aa4

❯ cargo bench -- --baseline main
   Compiling sqlparser v0.52.0 (/Users/davisp/github/davisp/datafusion-sqlparser-rs)
   Compiling sqlparser_bench v0.1.0 (/Users/davisp/github/davisp/datafusion-sqlparser-rs/sqlparser_bench)
    Finished `bench` profile [optimized] target(s) in 15.76s
     Running benches/sqlparser_bench.rs (target/release/deps/sqlparser_bench-9f7a6e6e193d8f5c)
sqlparser-rs parsing benchmark/sqlparser::select
                        time:   [2.9006 µs 2.9020 µs 2.9034 µs]
                        change: [-31.254% -31.013% -30.769%] (p = 0.00 < 0.05)
[...]

PR #1617, opened 2024-12-24
Improve Parser documentation

alamb:
- While reviewing https://github.com/apache/datafusion-sqlparser-rs/pull/1587 and preparing to improve documentation more it occurs to me the `Parser` struct could be documented better (especially given the large number of `parse_*` type methods that make it a bit confusing to navigate

I thought it would help to add some more overview information as well as internals

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -276,19 +276,58 @@ enum ParserState {
     ConnectBy,
 }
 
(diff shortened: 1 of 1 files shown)


iffyio (approved):
LGTM!

alamb:
Thanks @iffyio

PR #1618, opened 2024-12-24
Refactor advancing token to avoid duplication, avoid borrow checker issues

alamb:
- Follow on to https://github.com/apache/datafusion-sqlparser-rs/pull/1587

@davisp made some great improvements to the `Parser` API to avoid copies, but now the API for managing tokens is a bit unwieldy. Before we release a new version I would like to improve the API

Speicfically some of the APIs take a `&mut self` to advance `index` but return a read only reference to the token. This means the borrow checker is overly stringent and will sometimes think the parser has a mutable borrow outstanding when it doesnt.

I think we can make the APIs easier to use (and thus more likely to use) with a bit of finagling and documentation

See definitions of previous/current/next on https://github.com/apache/datafusion-sqlparser-rs/pull/1617

I am hoping that we can get the API in shape a bit more and then deprecate some of the older style APIs. 

Changes:
1. Introduce `advance_token()` to advance to the next token and reduce redundancy
2. Introduce `get_current_token()`, `get_next_token()` `get_prev_token`
2. Introduce `current_token_index`
3. Remove `next_token_ref_with_index`  and `next_token_ref` in favor of the above

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1276,7 +1276,9 @@ impl<'a> Parser<'a> {
 
         let dialect = self.dialect;
 
-        let (next_token, next_token_index) = self.next_token_ref_with_index();
+        self.advance_token();
+        let next_token_index = self.get_current_index();
+        let next_token = self.get_current_token();
         let span = next_token.span;
         let expr = match &next_token.token {
             Token::Word(w) => {
@@ -2914,7 +2916,9 @@ impl<'a> Parser<'a> {
 
         let dialect = self.dialect;
 
-        let (tok, tok_index) = self.next_token_ref_with_index();
+        self.advance_token();
+        let tok = self.get_current_token();
+        let tok_index = self.get_current_index();
         let span = tok.span;
         let regular_binary_operator = match &tok.token {
             Token::Spaceship => Some(BinaryOperator::Spaceship),
@@ -2994,7 +2998,8 @@ impl<'a> Parser<'a> {
                     // See https://www.postgresql.org/docs/current/sql-createoperator.html
                     let mut idents = vec![];
                     loop {
-                        idents.push(self.next_token_ref().to_string());
+                        self.advance_token();
+                        idents.push(self.get_current_token().to_string());
                         if !self.consume_token(&Token::Period) {
                             break;
                         }
@@ -3441,6 +3446,8 @@ impl<'a> Parser<'a> {
 
     /// Return the first non-whitespace token that has not yet been processed
     /// or Token::EOF
+    ///
+    /// See [`Self::peek_token_ref`] to avoid the copy.
     pub fn peek_token(&self) -> TokenWithSpan {
         self.peek_nth_token(0)
     }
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -1276,7 +1276,9 @@ impl<'a> Parser<'a> {
 
         let dialect = self.dialect;
 
-        let (next_token, next_token_index) = self.next_token_ref_with_index();
+        self.advance_token();
```
alamb, line 1279: This has two benefits: 
1. I think it is more explicit
2. It doesn't not result in a `mut` borrow of `self`(this is the key difference)

src/parser/mod.rs
```diff
@@ -3641,22 +3670,31 @@ impl<'a> Parser<'a> {
     #[must_use]
     pub fn parse_keyword(&mut self, expected: Keyword) -> bool {
         if self.peek_keyword(expected) {
-            self.next_token_ref();
+            self.advance_token();
```
alamb, line 3673: I think this is nicer too as it makes clear the token is unused (it is just being advanced)

src/parser/mod.rs
```diff
@@ -3557,45 +3564,67 @@ impl<'a> Parser<'a> {
...
+    pub fn next_token_no_skip(&mut self) -> Option<&TokenWithSpan> {
+        self.index += 1;
+        self.tokens.get(self.index - 1)
+    }
+
+    /// Advances the current token to the next non-whitespace token
     ///
-    /// # Notes:
-    /// OK to call repeatedly after reaching EOF.
-    pub fn next_token_ref_with_index(&mut self) -> (&TokenWithSpan, usize) {
```
alamb, line 3572: Note thse APIs were added in https://github.com/apache/datafusion-sqlparser-rs/pull/1587 (so not yet released) so this isn't a breaking API change

src/parser/mod.rs
```diff
@@ -3641,22 +3671,31 @@ impl<'a> Parser<'a> {
...
+    ///
+    /// See [`Self::parse_keyword_token_ref`] to avoid the copy.
     #[must_use]
     pub fn parse_keyword_token(&mut self, expected: Keyword) -> Option<TokenWithSpan> {
         self.parse_keyword_token_ref(expected).cloned()
     }
 
+    /// If the current token is the `expected` keyword, consume it and returns a reference to the next token.
+    ///
+    // TODO remove this in favor of advance??
```
iffyio, line 3691: Not sure I followed the intent of the comment, did it mean potentially replacing this api with something similar called `advance` or would that have a different behavior?

iffyio (approved):
Very nice! Thanks @alamb!

src/parser/mod.rs
alamb: That is a good question -- yes I think I meant changing this API somehow so it doesn't end up preserving a mutable borrow to return the option. 

I will see if I can find a better way

It is probably better to remove the comment though for now to avoid leaving a mess in the codebase:
 ```suggestion
```

alamb (commented):
Thanks @iffyio  -- I will try and find time to refine the `parse_keyword_token_ref` function over the next day or two

src/parser/mod.rs
iffyio: Ah I see that sounds good!

alamb:
> Not sure I followed the intent of the comment, did it mean potentially replacing this api with something similar called advance or would that have a different behavior?

> That is a good question -- yes I think I meant changing this API somehow so it doesn't end up preserving a mutable borrow to return the option.

BTW I spent some time last night and I found indeed it is possible to avoid these APIs. Here is what I came up with:
- https://github.com/apache/datafusion-sqlparser-rs/pull/1626#

PR #1620, opened 2024-12-26
SQLite: Allow dollar signs in placeholder names

hansott:
Relevant: https://github.com/apache/datafusion-sqlparser-rs/pull/1402

Screenshot 2024-12-26 at 14 46 58

Not sure if this is the best way to fix, open for feedback.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -636,10 +636,10 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if this dialect allows dollar quoted strings
-    /// e.g. `SELECT $$Hello, world!$$` or `SELECT $tag$Hello, world!$tag$`
-    fn supports_dollar_quoted_string(&self) -> bool {
-        true
+    /// Returns true if this dialect allows dollar placeholders
+    /// e.g. `SELECT $var` (SQLite)
+    fn supports_dollar_placeholder(&self) -> bool {
+        false
     }
 
     /// Does the dialect support with clause in create index statement?
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -636,6 +636,8 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allows dollar quoted strings
+    /// e.g. `SELECT $$Hello, world!$$` or `SELECT $tag$Hello, world!$tag$`
     fn supports_dollar_quoted_string(&self) -> bool {
         true
     }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -636,6 +636,10 @@ pub trait Dialect: Debug + Any {
         false
     }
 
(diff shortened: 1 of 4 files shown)


hansott:
cc @iffyio @alamb

src/dialect/mod.rs
```diff
@@ -636,6 +636,10 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    fn supports_dollar_quoted_string(&self) -> bool {
```
goldmedal, line 639: It's better to add some documents for this method. (Just like others).

hansott: Added!

src/tokenizer.rs
```diff
@@ -1557,10 +1559,13 @@ impl<'a> Tokenizer<'a> {
...
             value.push_str(&peeking_take_while(chars, |ch| {
-                ch.is_alphanumeric() || ch == '_'
+                ch.is_alphanumeric()
+                    || ch == '_'
+                    || matches!(ch, '$' if !self.dialect.supports_dollar_quoted_string())
             }));
 
-            if let Some('$') = chars.peek() {
+            let next_is_dollar = matches!(chars.peek(), Some('$'));
+            if next_is_dollar && self.dialect.supports_dollar_quoted_string() {
```
iffyio, line 1568:
```suggestion
            if matches!(chars.peek(), Some('$')) && self.dialect.supports_dollar_quoted_string() {
```

src/tokenizer.rs
```diff
@@ -2604,6 +2609,30 @@ mod tests {
         );
     }
 
+    #[test]
+    fn tokenize_dollar_placeholder_sqlite() {
```
iffyio, line 2613:
```suggestion
    fn tokenize_dollar_placeholder() {
```

src/dialect/mod.rs
```diff
@@ -636,6 +636,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allows dollar quoted strings
+    /// e.g. `SELECT $$Hello, world!$$` or `SELECT $tag$Hello, world!$tag$`
+    fn supports_dollar_quoted_string(&self) -> bool {
```
iffyio, line 641: I wondering would it make more sense to flip the condition to e.g. `supports_dollar_placeholder()`, setting that to true for sqlite and default false for the others? Thinking that might be the more conservative change of the two given that its the placeholder behavior that is more subtle/specific, (most of the other dialects technically don't support dollar quoted string so its less misleading if they don't have to explicitly flag otherwise)

iffyio (commented):
Thanks @hansott! The changes look reasonable to me overall, left some comments

hansott:
@iffyio Addressed all your comments, great feedback 🥇

iffyio (approved):
LGTM! Thanks @hansott!

alamb:
> @iffyio Addressed all your comments, great feedback 🥇

@iffyio ++

PR #1621, opened 2024-12-27
Fix the parsing result for the special double number

### LINKED ISSUE #1619: Wrong result for parsing special double number value
# Description
I tried to use the nightly commit in the upstream DataFusion (nightly commit) and the tests failed.
- https://github.com/goldmedal/datafusion/pull/2

~~I think it's a bug caused by https://github.com/apache/datafusion-sqlparser-rs/pull/1551~~. Consider the following SQL:
```
select 0. as c1
```
It should be parsed to 
```
ExprWithAlias { 
  expr: Value(Number("0.", false)), 
  alias: Ident { value: "c1", quote_style: None }
}
```
but it seems to be parsed to 
```
CompositeAccess { 
  expr: Value(Number("0", false)), 
  key: Ident { value: "as", quote_style: None, span: Span(Location(0,0)..Location(0,0))
}
```

We should handle this case in `parse_compound_field_access`, too.

### LINKED ISSUE #1622: Fix the special double number for the dialect supporting unquoted_hyphenated_identifiers
# Description
#1621 fixed the parsing special double number for general dialects but it doesn't cover the dialect supporting unquoted_hyphenated_identifiers. However, the double values below are also supported by them (e.g. BigQuery)
```
0.
.0
.0e-0
0.e-0
```
We should fix them and enable the tests `sqlparser_common::test_double_value`.

goldmedal:
close #1619 
close #1622 

# Changed
- Handle the Struct literal using `CompoundFieldAccess`
- ~~Introduce `support_unquoted_hyphenated_identifiers` to scope the change from #1598 to BigQuery only. See [previous discussions](https://github.com/apache/datafusion-sqlparser-rs/pull/1598#issuecomment-2563260997) for the details.~~
- As the suggestion https://github.com/apache/datafusion-sqlparser-rs/pull/1621#discussion_r1898558274, move the special case handling for BigQuery to the parser.

# Breaking change
- Change the arguments of `Parser::parse_identifier`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8769,17 +8769,28 @@ impl<'a> Parser<'a> {
     /// in this context on BigQuery.
     pub fn parse_object_name(&mut self, in_table_clause: bool) -> Result<ObjectName, ParserError> {
         let mut idents = vec![];
-        loop {
-            if self.dialect.supports_object_name_double_dot_notation()
-                && idents.len() == 1
-                && self.consume_token(&Token::Period)
-            {
-                // Empty string here means default schema
-                idents.push(Ident::new(""));
+
+        if dialect_of!(self is BigQueryDialect) && in_table_clause {
+            loop {
+                let (ident, end_with_period) = self.parse_unquoted_hyphenated_identifier()?;
+                idents.push(ident);
+                if !self.consume_token(&Token::Period) && !end_with_period {
+                    break;
+                }
             }
-            idents.push(self.parse_identifier(in_table_clause)?);
-            if !self.consume_token(&Token::Period) {
-                break;
+        } else {
+            loop {
+                if self.dialect.supports_object_name_double_dot_notation()
+                    && idents.len() == 1
+                    && self.consume_token(&Token::Period)
+                {
+                    // Empty string here means default schema
+                    idents.push(Ident::new(""));
+                }
+                idents.push(self.parse_identifier(in_table_clause)?);
+                if !self.consume_token(&Token::Period) {
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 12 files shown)


tests/sqlparser_common.rs
```diff
@@ -12470,6 +12578,41 @@ fn parse_composite_access_expr() {
     all_dialects_where(|d| d.supports_struct_literal()).verified_stmt(
         "SELECT * FROM t WHERE STRUCT(STRUCT(1 AS a, NULL AS b) AS c, NULL AS d).c.a IS NOT NULL",
     );
+    let support_struct = all_dialects_where(|d| d.supports_struct_literal());
+    let stmt = support_struct
+        .verified_only_select("SELECT STRUCT(STRUCT(1 AS a, NULL AS b) AS c, NULL AS d).c.a");
+    let expected = SelectItem::UnnamedExpr(Expr::CompoundFieldAccess {
```
goldmedal, line 12584: The access chain for the struct literal is represented by `CompoundFieldAccess` currently. See #1551 for the details.
cc @ayman-sigma

src/tokenizer.rs
```diff
@@ -1141,33 +1141,44 @@ impl<'a> Tokenizer<'a> {
...
-                        // Check if this actually is a float point number
-                        let mut char_clone = chars.peekable.clone();
-                        char_clone.next();
-                        // Next char should be a digit, otherwise, it is not a float point number
-                        if char_clone
-                            .peek()
-                            .map(|c| c.is_ascii_digit())
-                            .unwrap_or(false)
-                        {
+                    if self.dialect.support_unquoted_hyphenated_identifiers() {
```
iffyio, line 1145: @goldmedal @ayman-sigma
I'm wondering if it would make sense to move this special case handling from the tokenizer and into the parser. Since this is a BigQuery specific feature and only applies in a specific context, I feel like it would be more complicated to pull off within the tokenizer without affecting other use cases/dialects. Thinking in the parser we have more context on when this syntax is valid and a lot of that work was done in https://github.com/apache/datafusion-sqlparser-rs/pull/1109
(I'm imagining since the tokenizer would be left to parse decimals as usual, we won't have the consideration in https://github.com/apache/datafusion-sqlparser-rs/issues/1622 ?)

Essentially I'm wondering if we can extend [the work](https://github.com/apache/datafusion-sqlparser-rs/blob/6daa4b059cde8b77b67a3699b174ef0f8edff350/src/parser/mod.rs#L9015-L9063) to cover this scenario doublle number scenario. Using the example from https://github.com/apache/datafusion-sqlparser-rs/pull/1598 it would essentially mean that e.g. the following token stream

`[Word("foo"), Minus, Number("123."), Word("bar")]`

gets combined into

`ObjectName(vec![Word("foo-123"), Word("bar")])`

would something like that be feasible?

goldmedal: > Essentially I'm wondering if we can extend [the work](https://github.com/apache/datafusion-sqlparser-rs/blob/6daa4b059cde8b77b67a3699b174ef0f8edff350/src/parser/mod.rs#L9015-L9063) to cover this scenario doublle number scenario. Using the example from #1598 it would essentially mean that e.g. the following token stream
> 
> `[Word("foo"), Minus, Number("123."), Word("bar")]`
> 
> gets combined into
> 
> `ObjectName(vec![Word("foo-123"), Word("bar")])`
> 
> would something like that be feasible?

It's a good point. "Keeping tokenizer parse decimals as usual" makes sense to me.
I'll try it. Thanks.

goldmedal: Addressed it in [b1d02f2](https://github.com/apache/datafusion-sqlparser-rs/pull/1621/commits/b1d02f2ab9f8299e21063533ebe88abe78c5741a)
The double-value tests are enabled for BigQuery.

src/parser/mod.rs
```diff
@@ -9002,29 +9021,32 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a simple one-word identifier (possibly quoted, possibly a keyword)
-    ///
-    /// The `in_table_clause` parameter indicates whether the identifier is a table in a FROM, JOIN, or
-    /// similar table clause. Currently, this is used only to support unquoted hyphenated identifiers in
-    //  this context on BigQuery.
-    pub fn parse_identifier(&mut self, in_table_clause: bool) -> Result<Ident, ParserError> {
+    pub fn parse_identifier(&mut self) -> Result<Ident, ParserError> {
```
goldmedal, line 9024: It would be a breaking change for the downstream project using this public method.

iffyio (approved):
Looks great! Thanks for fixing this @goldmedal!
cc @alamb

goldmedal:
Thanks @iffyio 👍

alamb:
🚀

ayman-sigma:
Thanks @goldmedal , @iffyio  for the quick fix! 🙏

PR #1624, opened 2024-12-28
Improve parsing speed by avoiding some clones in parse_identifier

alamb:

I made a flamegraph of the `sqlparser_bench` and found a few unnecessary clones

flamegraph

This PR avoids a clone (string copy) in parse_identifier. 

I don't think this will make things much faster (yet), but it makes clear where the cloning is happening.
I hope to optimize the speed of parsing identifiers more as a follow on PR.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13477,7 +13477,7 @@ impl<'a> Parser<'a> {
 }
 
 impl Word {
-    #[deprecated(since = "0.55.0", note = "please use `into_ident` instead")]
+    #[deprecated(since = "0.54.0", note = "please use `into_ident` instead")]
     pub fn to_ident(&self, span: Span) -> Ident {
         Ident {
             value: self.value.clone(),
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -970,7 +970,7 @@ impl<'a> Parser<'a> {
             t @ (Token::Word(_) | Token::SingleQuotedString(_)) => {
                 if self.peek_token().token == Token::Period {
                     let mut id_parts: Vec<Ident> = vec![match t {
-                        Token::Word(w) => w.to_ident(next_token.span),
+                        Token::Word(w) => w.into_ident(next_token.span),
                         Token::SingleQuotedString(s) => Ident::with_quote('\'', s),
                         _ => unreachable!(), // We matched above
                     }];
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -970,15 +970,15 @@ impl<'a> Parser<'a> {
             t @ (Token::Word(_) | Token::SingleQuotedString(_)) => {
                 if self.peek_token().token == Token::Period {
                     let mut id_parts: Vec<Ident> = vec![match t {
-                        Token::Word(w) => w.to_ident(next_token.span),
+                        Token::Word(w) => w.into_ident(next_token.span),
```
alamb, line 973: `w.to_ident` cloned (copied) the string. `info_ident` simply reuses the string

src/parser/mod.rs
```diff
@@ -1108,7 +1108,7 @@ impl<'a> Parser<'a> {
             if dialect_of!(self is PostgreSqlDialect | GenericDialect) =>
                 {
                     Ok(Some(Expr::Function(Function {
-                        name: ObjectName(vec![w.to_ident(w_span)]),
+                        name: ObjectName(vec![w.clone().into_ident(w_span)]),
```
alamb, line 1111: in this case I couldn't figure out (yet) how to avoid this clone given that `&w` is passed in

This PR doesn't increase the number of clones done, but it makes it more explicit when they are happening

src/parser/mod.rs
```diff
@@ -13475,13 +13477,23 @@ impl<'a> Parser<'a> {
 }
 
 impl Word {
+    #[deprecated(since = "0.55.0", note = "please use `into_ident` instead")]
     pub fn to_ident(&self, span: Span) -> Ident {
         Ident {
             value: self.value.clone(),
```
alamb, line 13483: note `to_indent` clones self.value

src/parser/mod.rs
```diff
@@ -13475,13 +13477,23 @@ impl<'a> Parser<'a> {
 }
 
 impl Word {
+    #[deprecated(since = "0.55.0", note = "please use `into_ident` instead")]
```
alamb, line 13480:
```suggestion
    #[deprecated(since = "0.54.0", note = "please use `into_ident` instead")]
```

src/parser/mod.rs
```diff
@@ -13475,13 +13477,23 @@ impl<'a> Parser<'a> {
 }
 
 impl Word {
+    #[deprecated(since = "0.54.0", note = "please use `into_ident` instead")]
```
alamb, line 13480: I am trying to help any downstream crates that may also use this function with a deprecation notice rather than simply removing the function

iffyio: Sounds good!

iffyio (approved):
LGTM!

PR #1625, opened 2024-12-28
Fix `sqlparser_bench` benchmark compilation

alamb:
If you try to run the sqlparser benchmarks it fails to compile:

```shell
cd sqlparser_bench
cargo clippy --all-targets
```

```
error[E0277]: can't compare `std::string::String` with `sqlparser::ast::Statement`
  --> benches/sqlparser_bench.rs:82:13
   |
82 |             assert_eq!(formatted_query, large_statement);
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `std::string::String == sqlparser::ast::Statement`
   |
```

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -78,8 +78,7 @@ fn basic_queries(c: &mut Criterion) {
 
     group.bench_function("format_large_statement", |b| {
         b.iter(|| {
-            let formatted_query = large_statement.to_string();
-            assert_eq!(formatted_query, large_statement);
+            let _formatted_query = large_statement.to_string();
         });
     });
 }


sqlparser_bench/benches/sqlparser_bench.rs
```diff
@@ -78,8 +78,7 @@ fn basic_queries(c: &mut Criterion) {
 
     group.bench_function("format_large_statement", |b| {
         b.iter(|| {
-            let formatted_query = large_statement.to_string();
-            assert_eq!(formatted_query, large_statement);
+            let _formatted_query = large_statement.to_string();
```
alamb, line 81: I am not sure what this assert was meant to show, so I removed it.

alamb:
It does occur to me that this is not covered by CI. Maybe we should move the benchmark binary into the normal sqlparser crate (which I think would be more standard) rather than having an entirely separate crate (`sqlparser_bench`) -- that would also ensure the benchmark is covered by CI tests 🤔

alamb:
> It does occur to me that this is not covered by CI. Maybe we should move the benchmark binary into the normal sqlparser crate (which I think would be more standard) rather than having an entirely separate crate (`sqlparser_bench`) -- that would also ensure the benchmark is covered by CI tests 🤔

- Added a CI check here: https://github.com/apache/datafusion-sqlparser-rs/pull/1627

iffyio (approved):

iffyio:
Thanks @alamb!

PR #1626, opened 2024-12-28
Simplify `parse_keyword_apis` more

alamb:
- Follow on to https://github.com/apache/datafusion-sqlparser-rs/pull/1618

Rationale:

@davisp made some great improvements to performance, but now the API for managing tokens is a bit unwieldy. Before we release a new version I am working to slim down the APIs (ideally so we can migrate more code to avoid cloning)

Inspired by @iffyo's comments on https://github.com/apache/datafusion-sqlparser-rs/pull/1618

> Not sure I followed the intent of the comment, did it mean potentially replacing this api with something similar called advance or would that have a different behavior?

I spent some time playing around with the API and I found these APIs are not necessary, so I propose removing them to simplify the parser interface

Changes:
1. Remove `parse_keyword_token` and `parse_keyword_token_ref` in favor of get_current_token

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3708,7 +3708,7 @@ impl<'a> Parser<'a> {
         )
     }
 
-    /// Report that the current token was found instead of `expected`.
+    /// Report that the token at `index` was found instead of `expected`.
     pub fn expected_at<T>(&self, expected: &str, index: usize) -> Result<T, ParserError> {
         let found = self.tokens.get(index).unwrap_or(&EOF_TOKEN);
         parser_err!(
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -3729,27 +3729,6 @@ impl<'a> Parser<'a> {
...
-    /// See [`Self::parse_keyword_token_ref`] to avoid the copy.
-    #[must_use]
-    pub fn parse_keyword_token(&mut self, expected: Keyword) -> Option<TokenWithSpan> {
-        self.parse_keyword_token_ref(expected).cloned()
-    }
-
-    /// If the current token is the `expected` keyword, consume it and returns a reference to the next token.
-    ///
-    #[must_use]
-    pub fn parse_keyword_token_ref(&mut self, expected: Keyword) -> Option<&TokenWithSpan> {
```
alamb, line 3743: as discussed on https://github.com/apache/datafusion-sqlparser-rs/pull/1618, the fact this takes `&mut self` makes avoiding clones challenging (because as long as &TokenWithSpan) is around, rust treats the parser as borrowed mutably (and you can't call other APIs)

src/parser/mod.rs
```diff
@@ -3832,9 +3811,11 @@ impl<'a> Parser<'a> {
 
     /// If the current token is the `expected` keyword, consume the token.
     /// Otherwise, return an error.
+    ///
+    // todo deprecate infavor of expected_keyword_is
     pub fn expect_keyword(&mut self, expected: Keyword) -> Result<TokenWithSpan, ParserError> {
-        if let Some(token) = self.parse_keyword_token_ref(expected) {
-            Ok(token.clone())
+        if self.parse_keyword(expected) {
```
alamb, line 3817: I actually think using `parse_keyword` and `get_current_token` is clearer than the original formulation of `paser_keyword_token_ref` and checking for `Some`

iffyio (approved):
LGTM!

alamb:
Thanks @iffyio

PR #1629, opened 2024-12-30
Correctly tokenize nested comments

hansott:
The tokenizer currently throws EOF error for `select 'foo' /*/**/*/`

`last_ch` causes problems when tokenizing nested comments, we have to consume the combination of `/*` or `*/`

The existing `tokenize_nested_multiline_comment` test fails after fixing this logic:

```
/*multi-line\n* \n/* comment \n /*comment*/*/ */ /comment*/
^^ Start                                      ^^ End nested comment
```

proof:

```
psql (14.15 (Homebrew), server 14.11)
Type "help" for help.

main_db=# SELECT 'foo' /*/**/*/;
 ?column?
----------
 foo
(1 row)

main_db=# SELECT 'foo' /*multi-line\n* \n/* comment \n /*comment*/*/ */ /comment*/;
ERROR:  syntax error at or near ";"
LINE 1: .../*multi-line\n* \n/* comment \n /*comment*/*/ */ /comment*/;
                                                                      ^
main_db=# SELECT 'foo' /*multi-line\n* \n/* comment \n /*comment*/*/ */;
 ?column?
----------
 foo
(1 row)
```

Relevant: https://github.com/apache/datafusion-sqlparser-rs/pull/726

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -2727,48 +2727,52 @@ mod tests {
     #[test]
     fn tokenize_nested_multiline_comment() {
         let dialect = GenericDialect {};
-
-        let sql = String::from("0/*multi-line\n* \n/* comment \n /*comment*/*/ */ /comment*/1");
-        let tokens = Tokenizer::new(&dialect, &sql).tokenize().unwrap();
-        let expected = vec![
-            Token::Number("0".to_string(), false),
-            Token::Whitespace(Whitespace::MultiLineComment(
-                "multi-line\n* \n/* comment \n /*comment*/*/ ".into(),
-            )),
-            Token::Whitespace(Whitespace::Space),
-            Token::Div,
-            Token::Word(Word {
-                value: "comment".to_string(),
-                quote_style: None,
-                keyword: Keyword::COMMENT,
-            }),
-            Token::Mul,
-            Token::Div,
-            Token::Number("1".to_string(), false),
+        let test_cases = vec![
+            (
+                "0/*multi-line\n* \n/* comment \n /*comment*/*/ */ /comment*/1",
+                vec![
+                    Token::Number("0".to_string(), false),
+                    Token::Whitespace(Whitespace::MultiLineComment(
+                        "multi-line\n* \n/* comment \n /*comment*/*/ ".into(),
+                    )),
+                    Token::Whitespace(Whitespace::Space),
+                    Token::Div,
+                    Token::Word(Word {
+                        value: "comment".to_string(),
+                        quote_style: None,
+                        keyword: Keyword::COMMENT,
+                    }),
+                    Token::Mul,
+                    Token::Div,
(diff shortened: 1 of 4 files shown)


hansott:
cc @iffyio

src/tokenizer.rs
```diff
@@ -1855,22 +1855,31 @@ impl<'a> Tokenizer<'a> {
...
-        let mut last_ch = ' ';
+        let supports_nested_comments = self.dialect.supports_nested_comments();
 
         loop {
             match chars.next() {
                 Some(ch) => {
-                    if last_ch == '/' && ch == '*' {
+                    if ch == '/' && matches!(chars.peek(), Some('*')) && supports_nested_comments {
+                        s.push(ch);
+                        s.push(chars.next().unwrap()); // consume the '*'
```
iffyio, line 1865: can we rewrite this to return an error in order to avoid the unwrap?

src/tokenizer.rs
```diff
@@ -1855,22 +1855,31 @@ impl<'a> Tokenizer<'a> {
...
+                    if ch == '*' && matches!(chars.peek(), Some('/')) {
+                        s.push(ch);
+                        let slash = chars.next();
                         nested -= 1;
                         if nested == 0 {
-                            s.pop();
+                            s.pop(); // remove the last '*'
                             break Ok(Some(Token::Whitespace(Whitespace::MultiLineComment(s))));
                         }
+                        s.push(slash.unwrap());
```
iffyio, line 1878: same here, it would be nice if we can avoid unwraping in code

src/tokenizer.rs
```diff
@@ -1855,22 +1855,31 @@ impl<'a> Tokenizer<'a> {
...
         let mut s = String::new();
         let mut nested = 1;
-        let mut last_ch = ' ';
+        let supports_nested_comments = self.dialect.supports_nested_comments();
 
         loop {
             match chars.next() {
                 Some(ch) => {
-                    if last_ch == '/' && ch == '*' {
+                    if ch == '/' && matches!(chars.peek(), Some('*')) && supports_nested_comments {
```
iffyio, line 1863: could we move some of the condition out to the match statement? thinking that way we avoid extra nesting + continue. And the different cases are clearer. e.g.
```rust
Some('/') if matches!(chars.peek(), Some('*')) && supports_nested_comments => { ... }
Some('*') if matches!(chars.peek(), Some('/')) => { ... }
Some(ch) => { ... }
None => { ... }
```

iffyio (commented):
Thanks @hansott! Left a couple comments

src/tokenizer.rs
hansott: Makes sense, will do!

iffyio (approved):
LGTM! Thanks @hansott!
cc @alamb

PR #1630, opened 2024-12-31
Support pluralized time units

wugeer:
add time unit `SECONDS/MINUTES/HOURS/DAYS/WEEKS/MONTHS/YEARS` for `INTERVAL` type

see 
- https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27838462#LanguageManualTypes-Intervals
- https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT


src/parser/mod.rs
```diff
@@ -2353,14 +2355,30 @@ impl<'a> Parser<'a> {
...
+                        && self.consume_token(&Token::LParen)
+                    {
+                        let week_day = self.parse_identifier()?;
+                        self.expect_token(&Token::RParen)?;
+                        Some(week_day)
+                    } else {
+                        None
+                    };
+                    Ok(DateTimeField::Weeks(week_day))
+                }
```
iffyio, line 2369: This syntax does seem supported by [bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions)?

wugeer: Sorry, my mistake, it has been corrected.

src/parser/mod.rs
```diff
@@ -2353,14 +2355,30 @@ impl<'a> Parser<'a> {
...
```
iffyio, line 2369: @wugeer I think the code still references BigQuery, realized the tests don't seem to cover this scenario, but actually I'm also wondering which dialect supports these pluralized time units (do you have a link to the docs)?

tests/sqlparser_common.rs
```diff
@@ -5374,10 +5396,49 @@ fn parse_interval_all() {
...
+    );
+
+    // keep Generic/BigQuery extract week with weekday syntax success
+    let supported_dialects = TestedDialects::new(vec![
+        Box::new(GenericDialect {}),
+        Box::new(BigQueryDialect {}),
+    ]);
+
+    let sql = "SELECT EXTRACT(WEEK(a) FROM date)";
+    supported_dialects.verified_stmt(sql);
```
iffyio, line 5422: Oh not sure we need to test this here, its [already tested here](https://github.com/apache/datafusion-sqlparser-rs/blob/ba3acb243007555bfc90885fcc9aad52642c06ca/tests/sqlparser_bigquery.rs#L2181-L2193) for example

tests/sqlparser_common.rs
```diff
@@ -5374,10 +5396,49 @@ fn parse_interval_all() {
...
+        Box::new(SQLiteDialect {}),
+        Box::new(DuckDbDialect {}),
+    ]);
+
+    assert_eq!(
+        ParserError::ParserError("Expected 'FROM' or ','".to_owned()),
+        all_other_dialects
+            .parse_sql_statements("SELECT EXTRACT(WEEK(a) FROM date)")
+            .unwrap_err()
+    );
```
iffyio, line 5441: similarly I think we can skip this test, if a dialect doesnt support a feature no need to test that feature with the dialect

src/parser/mod.rs
iffyio: That sounds reasonable thanks for checking! It seems like `Weeks(Option<Ident>)` isn't supported by any dialect? I'm thinking if so we can change that to `Weeks`?

wugeer: Indeed, you are right! Have Done!

tests/sqlparser_common.rs
wugeer: Have Done! :)

tests/sqlparser_common.rs
wugeer: Have Done!

iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

alamb:
Wow we have really been picking up steam on reviews in sqlparser. The next relaese is going to be sweet

Issue #1607, opened 2024-12-17
Parser rejects column aliases that Snowflake accepts


yoavcloud:
For example:
`SELECT cluster FROM tbl` --> accepted by both Snowflake and the parser
`SELECT col cluster FROM tbl` --> accepted by Snowflake but not by the parser
`SELECT col col_alias FROM tbl` --> accepted by both Snowflake and the parser
`SELECT col AS cluster FROM tbl` --> accepted by both Snowflake and the parser (this one really bakes my noodle...)

yoavcloud:
The underlying issue here is that when a column alias is provided implicitly, i.e. without the use of the `AS` keyword, the parser has to decide if the next word after a column ident is an alias or the beginning of some other construct that should be parsed differently.

For example:
`SELECT 1 LIMIT` on Snowflake produces a result set with one column named `limit`, but the statement `SELECT 1 LIMIT 5` produces a result set with one column named `1` (the limit is parsed as a limit on the number of rows in the result set).

A possible solution is to make this decision more accurate. Instead of looking at just the next word, the parser should look further ahead to identify constructs that should not be parsed as an alias.

iffyio:
Done in https://github.com/apache/datafusion-sqlparser-rs/pull/1632

PR #1632, opened 2025-01-01
Add support for Snowflake column aliases that use SQL keywords

yoavcloud:
Attempts at solving issue #1607 

Snowflake is not very strict with using SQL keywords as select list item aliases. Added support for the dialect to control if a word is parsed as an alias or not. Parsing of aliases is quite centralized now, and this PR breaks that up a bit, so feedback is welcomed if this is a good approach or alternative solutions are preferred.


iffyio (commented):
Thanks @yoavcloud! The approach sounds reasonable to me, left a comment

yoavcloud:
@iffyio in this iteration I re-used the logic but also aligned parsing of table aliases to column aliases, enabling customization by dialects. LMKWYT

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1637, opened 2025-01-03
Add support for USE SECONDARY ROLE (vs. ROLES)

yoavcloud:


alamb (approved):
Thanks @yoavcloud  -- makes sense to me

FYI @iffyio

PR #1638, opened 2025-01-03
Add support for the SQL OVERLAPS predicate

yoavcloud:
https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#overlaps-predicate

I did not gate this one by dialect, as it doesn't seem to conflict with anything and is documented as part of the standard.


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1641, opened 2025-01-03
Add support for MySQL's INSERT INTO ... SET syntax

yoavcloud:
https://dev.mysql.com/doc/refman/8.4/en/insert.html


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

alamb:
Thank you for these PRs @yoavcloud  and thank you for reviewing and keeping them moving @iffyio

PR #1645, opened 2025-01-06
Support trailing commas in `FROM` clause

barsela1:
support trailing commas in FROM clause.

- add a const global list of keyworads that are allowed to appear after FROM, called "ALLOWED_KEYWORD_AFTER_FROM"
- Every dialict implement:
--get_reserved_keyword_after_from function that returns ALLOWED_KEYWORD_AFTER_FROM as default. each dialect can customize the list if needed.
-- supports_from_trailing_commas function that returns false  as default but for each dialect that supports trailing commas in the  FROM clause can customize the return value to true (e.g., Snowflake).
- add a function "parse_from" to handle the FROM keyword. parse_from set and pass trailing_commas according to the return value of self.dialect.supports_from_trailing_commas().
- change signature of parse_comma_separated_with_trailing_commas and is_parse_comma_separated_end_with_trailing_commas by adding parameter "reserved_keywords" (&[Keyword]) 
-at  is_parse_comma_separated_end_with_trailing_commas: keywords::RESERVED_FOR_COLUMN_ALIAS.contains(&kw.keyword) replaced by reserved_keywords so that the reserved_keywords can be determined by the current parsing state.
- In other cases where is_parse_comma_separated_end_with_trailing_commas is used, keywords::RESERVED_FOR_COLUMN_ALIAS can be obtained by calling the default_reserved_keywords function and passing it through the reserved_keywords parameter.

tests:
-added new tests.
-all tests passed.


src/parser/mod.rs
```diff
@@ -3974,15 +3989,26 @@ impl<'a> Parser<'a> {
...
-        self.is_parse_comma_separated_end_with_trailing_commas(self.options.trailing_commas)
+        self.is_parse_comma_separated_end_with_trailing_commas(
+            self.options.trailing_commas,
+            self.default_reserved_keywords(),
+        )
+    }
+
+    fn default_reserved_keywords(&self) -> &'static [Keyword] {
+        keywords::RESERVED_FOR_COLUMN_ALIAS
     }
```
iffyio, line 4000: the function name `default_reserved_keywords` sounds a bit misleading out of context, I wonder would it make sense to have the parameter optional and use it internally as the default? e.g.
```rust
fn is_parse_comma_separated_end_with_trailing_commas(
        &mut self,
        trailing_commas: bool,
        reserved_keywords: Option<&[Keyword]>,
    ) {
    let reserved_keywords = reserved_keywords.unwrap_or(keywords::RESERVED_FOR_COLUMN_ALIAS)
}
```

tests/sqlparser_snowflake.rs
```diff
@@ -2483,6 +2483,37 @@ fn test_sf_trailing_commas() {
...
+fn test_sf_trailing_commas_multiple_subqueries() {
+    snowflake().verified_only_select_with_canonical(
+        "SELECT 1, 2 FROM (SELECT * FROM t1), (SELECT * FROM t2),",
+        "SELECT 1, 2 FROM (SELECT * FROM t1), (SELECT * FROM t2)",
+    );
+}
+
+#[test]
+fn test_from_with_nested_trailing_commas() {
+    let sql = "SELECT 1, 2 FROM (SELECT * FROM t,),";
```
iffyio, line 2513: Could we merge these test scenarios into a single test case? since they're testing the same feature

tests/sqlparser_snowflake.rs
```diff
@@ -2483,6 +2483,37 @@ fn test_sf_trailing_commas() {
...
+
+#[test]
+fn test_sf_trailing_commas_multiple_sources() {
+    snowflake()
+        .verified_only_select_with_canonical("SELECT 1, 2 FROM t1, t2,", "SELECT 1, 2 FROM t1, t2");
+}
+
+#[test]
+fn test_from_trailing_commas_with_limit() {
+    let sql = "SELECT a, FROM b, LIMIT 1";
```
iffyio, line 2499: Can we add a similar scenario for the other entries in the allowed keyword list? (i.e. INTO, HAVING, WHERE)

src/parser/mod.rs
```diff
@@ -3947,20 +3951,31 @@ impl<'a> Parser<'a> {
         Ok(values)
     }
 
+    //Parse a 'FROM' statment. support trailing comma
+    pub fn parse_from(&mut self) -> Result<Vec<TableWithJoins>, ParserError> {
```
iffyio, line 3955:
```suggestion
    /// Parse a list of [TableWithJoins]
    fn parse_table_with_joins(&mut self) -> Result<Vec<TableWithJoins>, ParserError> {
```
Thinking no need to make this function public. And `parse_from` sounded generic since the actual `FROM` keyword isn't part of the parsed elements

tests/sqlparser_snowflake.rs
```diff
@@ -2483,6 +2483,37 @@ fn test_sf_trailing_commas() {
     snowflake().verified_only_select_with_canonical("SELECT 1, 2, FROM t", "SELECT 1, 2 FROM t");
 }
 
+#[test]
```
iffyio, line 2486: Could we move the tests to common and use the [all_dialects_where](https://github.com/apache/datafusion-sqlparser-rs/blob/5f3de315842963e6db47c9f30ddc914dc7c605c9/tests/sqlparser_common.rs#L8254) function? That would make it automatic to add test coverage for other dialects that have this feature enabled in the future

src/keywords.rs
```diff
@@ -975,6 +975,16 @@ pub const RESERVED_FOR_COLUMN_ALIAS: &[Keyword] = &[
     Keyword::END,
 ];
 
+// Global list of reserved keywords alloweed after FROM.
+// Parser should call Dialect::get_reserved_keyword_after_from
+// to allow for each dialect to customize the list.
+pub const ALLOWED_KEYWORD_AFTER_FROM: &[Keyword] = &[
```
iffyio, line 981:
```suggestion
pub const RESERVED_FOR_TABLE_FACTOR: &[Keyword] = &[
```
Thinking something like this to keep the naming somewhat similar to the existing lists?

If this changes the dialect method `get_reserved_keyword_after_from` would be updated to match the naming

src/dialect/mod.rs
```diff
@@ -399,6 +399,11 @@ pub trait Dialect: Debug + Any {
         self.supports_trailing_commas()
     }
 
+    /// Does the dialect support trailing commas in the FROM clause list?
```
iffyio, line 402:
```suggestion
    /// Returns true if the dialect supports trailing commas in the `FROM` clause of a `SELECT` statement.
    /// Example: `SELECT 1 FROM T, U, LIMIT 1`
```

src/dialect/mod.rs
```diff
@@ -758,6 +763,11 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_IDENTIFIER.contains(&kw)
     }
 
+    // Returns list of keyword allowed after from
```
iffyio, line 766:
```suggestion
    /// Returns reserved keywords when looking to parse a [TableFactor].
    /// See [Self::supports_from_trailing_commas]
```

iffyio (commented):
Thanks @barsela1! The changes look reasonable to me overall, left some comments

tests/sqlparser_snowflake.rs
```diff
@@ -2483,6 +2483,34 @@ fn test_sf_trailing_commas() {
...
+fn test_sf_from_trailing_commas() {
```
iffyio, line 2487: hmm are the tests relying on any snowflake specific feature, I imagined they should be part of the `test_trailing_commas_in_from` in common instead? essentially moving the tests to that file

iffyio (approved):
LGTM! Thanks @barsela1!
cc @alamb

iffyio:
@barsela1 could you take a look to resolve the conflicts on the branch?

alamb:
I also restarted the checks

src/dialect/mod.rs
```diff
@@ -404,6 +404,12 @@ pub trait Dialect: Debug + Any {
         self.supports_trailing_commas()
     }
 
+    /// Returns true if the dialect supports trailing commas in the `FROM` clause of a `SELECT` statement.
+    /// /// Example: `SELECT 1 FROM T, U, LIMIT 1`
+    fn supports_from_trailing_commas(&self) -> bool {
```
alamb, line 409: very nice

alamb:
Thanks @barsela1  and @iffyio  ❤️

PR #1646, opened 2025-01-06
MsSQL SET for session params

yoavcloud:


alamb:
Hi @yoavcloud  -- it looks like this PR has some conflicts -- is there any chance you can resolve them?

yoavcloud:
@alamb done

src/parser/mod.rs
```diff
@@ -10307,11 +10307,67 @@ impl<'a> Parser<'a> {
...
+        } else if self.parse_keywords(&[Keyword::STATISTICS, Keyword::PROFILE]) {
+            return Ok("STATISTICS PROFILE".to_string());
+        } else if self.parse_keywords(&[Keyword::STATISTICS, Keyword::TIME]) {
+            return Ok("STATISTICS TIME".to_string());
+        }
+        let next_token = self.next_token();
+        if let Token::Word(w) = next_token.token {
+            return Ok(w.to_string());
+        }
+        self.expected("Session param name", next_token)
```
alamb, line 10368: maybe it would be nice to list the expected keywords (IO/XML/PROFILE/TIME, etc) here

src/parser/mod.rs
```diff
@@ -10307,11 +10307,67 @@ impl<'a> Parser<'a> {
...
+        Ok(Statement::SetSessionParam {
+            names,
+            identity_insert_obj,
+            offsets_keywords,
+            value,
+        })
+    }
+
+    pub fn parse_set_session_param_name(&mut self) -> Result<String, ParserError> {
+        if self.parse_keywords(&[Keyword::STATISTICS, Keyword::IO]) {
```
alamb, line 10355: Given these are hard coded it seems like it would be cleaner to eventually return an enum for what setting was set rather than arbitrary strings that the using library needs to parse / interpret

However, I also think we could do that as a follow on PR

alamb (approved):
Seems good to me -- thanks @yoavcloud  -- it would be nice to avoid using Strings quite as much but that is something we could do as a follow on PR if anyone cares

alamb:
fyi @iffyio

yoavcloud:
@alamb @iffyio please see a more structured approach to the parsing of the session parameters

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #1647, opened 2025-01-07
Start new line if \r in Postgres dialect

hansott:
Currently the tokenizer throws an error for

```js
insert into cats_2 (petname) values ('foo'),--\r(version()||'\n');
```

this is because postgres treats \r as a separate new line character, see https://github.com/postgres/postgres/blob/master/src/backend/parser/scan.l

> In order to make the world safe for Windows and Mac clients as well as Unix ones, we accept either \n or \r as a newline.
> A DOS-style \r\n sequence will be seen as two successive newlines, but that doesn't cause any problems.
> non_newline			[^\n\r]
> comment			("--"{non_newline}*)

Let's make sure we start a new line if we encounter a \r when tokenizing a comment.


alamb (approved):
Thank you @hansott  - this makes sense to me and is a nice contribution

I'll leave it open for a while in case @iffyio  would also like a chance to review

iffyio (approved):
LGTM! Thanks @hansott!

PR #1649, opened 2025-01-07
Add support for MS-SQL BEGIN/END TRY/CATCH

yoavcloud:
MS-SQL has additional modifiers on tx control statements that implement exception handling, i.e. try/catch behavior: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/try-catch-transact-sql?view=sql-server-ver16.

Added support for parsing of such modifiers.


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1650, opened 2025-01-07
Correctly look for end delimiter dollar quoted string

hansott:
Currently the tokenizer throws an error for

```sql
SELECT $abc$x$ab$abc$
```

The logic is also quite difficult to read so I made it a bit simpler.

Before

```rust
'searching_for_end: loop {
    s.push_str(&peeking_take_while(chars, |ch| ch != '$'));
    match chars.peek() {
        Some('$') => {
            chars.next();
            let mut maybe_s = String::from("$");
            for c in value.chars() {
                if let Some(next_char) = chars.next() {
                    maybe_s.push(next_char);
                    if next_char != c {
                        // This doesn't match the dollar quote delimiter so this
                        // is not the end of the string.
                        s.push_str(&maybe_s);
                        continue 'searching_for_end;
                    }
                } else {
                    return self.tokenizer_error(
                        chars.location(),
                        "Unterminated dollar-quoted, expected $",
                    );
                }
            }
            if chars.peek() == Some(&'$') {
                chars.next();
                maybe_s.push('$');
                // maybe_s matches the end delimiter
                break 'searching_for_end;
            } else {
                // This also doesn't match the dollar quote delimiter as there are
                // more characters before the second dollar so this is not the end
                // of the string.
                s.push_str(&maybe_s);
                continue 'searching_for_end;
            }
        }
        _ => {
            return self.tokenizer_error(
                chars.location(),
                "Unterminated dollar-quoted, expected $",
            )
        }
    }
}
```

After
```rust
let mut temp = String::new();
let end_delimiter = format!("${}$", value);

loop {
    match chars.next() {
        Some(ch) => {
            temp.push(ch);

            if temp.ends_with(&end_delimiter) {
                s.push_str(&temp[..temp.len() - end_delimiter.len()]);
                break;
            }
        }
        None => {
            if temp.ends_with(&end_delimiter) {
                s.push_str(&temp[..temp.len() - end_delimiter.len()]);
                break;
            }

            return self.tokenizer_error(
                chars.location(),
                "Unterminated dollar-quoted, expected $",
            );
        }
    }
}
```

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1554,34 +1554,25 @@ impl<'a> Tokenizer<'a> {
             if matches!(chars.peek(), Some('$')) && !self.dialect.supports_dollar_placeholder() {
                 chars.next();
 
-                let mut end_tag_name = String::new();
+                let mut temp = String::new();
+                let end_delimiter = format!("${}$", value);
 
                 loop {
                     match chars.next() {
-                        Some('$') => {
-                            if !end_tag_name.is_empty() {
-                                if end_tag_name == value {
-                                    // We found the end delimiter
-                                    // Let's remove the tag name from the string content
-                                    // (plus the dollar sign of the end delimiter)
-                                    s.truncate(s.len() - value.len() - 1);
-                                    break;
-                                } else {
-                                    // Reset the tag name
-                                    end_tag_name.clear();
-                                }
-                            } else {
-                                // Start a new tag name
-                                end_tag_name.clear();
-                            }
-
-                            s.push('$');
-                        }
                         Some(ch) => {
-                            end_tag_name.push(ch);
-                            s.push(ch);
+                            temp.push(ch); // Collect all characters into `temp`
+
+                            if temp.ends_with(&end_delimiter) {
+                                s.push_str(&temp[..temp.len() - end_delimiter.len()]);
+                                break;
+                            }
                         }
                         None => {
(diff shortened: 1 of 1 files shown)


hansott:
Adding some more tests first

hansott:
@iffyio let me know if see some improvements

src/tokenizer.rs
```diff
@@ -1554,46 +1554,29 @@ impl<'a> Tokenizer<'a> {
...
+                let mut temp = String::new();
+                let end_delimiter = format!("${}$", value);
+
+                loop {
+                    match chars.next() {
+                        Some(ch) => {
+                            temp.push(ch);
+
+                            if temp.ends_with(&end_delimiter) {
+                                s.push_str(&temp[..temp.len() - end_delimiter.len()]);
```
iffyio, line 1566: ah I think indexing based on `String::len` won't necessarily produce correct results on arbitrary UTF8 strings, we could something like `strip_suffix` that doesn't rely on the byte count? e.g
```rust
if let Some(temp) = temp.strip_suffix(&end_delimiter) {
    s.push_str(temp);
    break
}
```

iffyio (commented):
Looks good to me overall thanks @hansott! Left one comment regarding the suffix matching

hansott:
@iffyio Updated, thanks for your feedback! 🥇

iffyio (approved):
LGTM! Thanks @hansott!
cc @alamb

PR #1652, opened 2025-01-09
Add support for the Snowflake MINUS set operator

yoavcloud:
Similar to EXCEPT, see here: https://docs.snowflake.com/en/sql-reference/operators-query#minus-except


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

alamb:
🚀

Issue #1644, opened 2025-01-06
fail to parse `grouping sets` in Hive Dialect


wugeer:
when parse sql `SELECT a, b, SUM(c) FROM tab1 GROUP BY a, b GROUPING SETS ( (a,b) )`, it fail in hive dialect.
![image](https://github.com/user-attachments/assets/dfcc1131-9711-4bd1-b759-010683a0fa87)

See
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=30151323#EnhancedAggregation,Cube,GroupingandRollup-GROUPINGSETSclause

PR #1653, opened 2025-01-09
Add supports for Hive's `SELECT ... GROUP BY .. GROUPING SETS` syntax

wugeer:
This PR supports `SELECT ... GROUP BY .. GROUPING SETS`  syntax for hive dialect,. For more information, please refer to:
https://cwiki.apache.org/confluence/display/Hive/Enhanced+Aggregation%2C+Cube%2C+Grouping+and+Rollup#EnhancedAggregation,Cube,GroupingandRollup-GROUPINGSETSclause

Also support hive `with rollup/cube` clause.
This resolves issue https://github.com/apache/datafusion-sqlparser-rs/issues/1644


src/ast/query.rs
```diff
@@ -328,6 +328,10 @@ pub struct Select {
     pub value_table_mode: Option<ValueTableMode>,
     /// STARTING WITH .. CONNECT BY
     pub connect_by: Option<ConnectBy>,
+    /// Hive syntax: `SELECT ... GROUP BY .. GROUPING SETS`
+    ///
+    /// [Hive](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=30151323#EnhancedAggregation,Cube,GroupingandRollup-GROUPINGSETSclause)
+    pub grouping_sets: Option<Expr>,
```
iffyio, line 334: I'm wondering is it possible to cover the lacking cases using the `group_by` property vs introducing a new property? Its not currently clear to me why a new field was introduced.

wugeer: Indeed, another idea is to attribute this situation to the `group by` property. It may be better. I will try to make it. :)

alamb:

wugeer:
@iffyio @alamb Sorry for replying so late. :(
I’ve resubmitted the code. Could you please take some time to review it?

src/parser/mod.rs
```diff
@@ -9092,6 +9092,16 @@ impl<'a> Parser<'a> {
                     });
                 }
             }
+            if self.dialect.supports_group_by_special_grouping_sets()
+                && self.parse_keywords(&[Keyword::GROUPING, Keyword::SETS])
+            {
+                self.expect_token(&Token::LParen)?;
+                let result = self.parse_comma_separated(|p| p.parse_tuple(true, true))?;
```
iffyio, line 9099: should this be comma separate expr or similar instead of tuple?
here's one example from the [hive doc](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=30151323#EnhancedAggregation,Cube,GroupingandRollup-GROUPINGSETSclause) that we seem to not be able to support as  a result of using parse_tuple
```sql
SELECT a, b, SUM( c ) FROM tab1 GROUP BY a, b GROUPING SETS ( (a, b), a, b, ( ) )
```

src/parser/mod.rs
```diff
@@ -9092,6 +9092,16 @@ impl<'a> Parser<'a> {
                     });
                 }
             }
+            if self.dialect.supports_group_by_special_grouping_sets()
```
iffyio, line 9095: I wonder if this dialect method is needed? the clause isn't special from what I can tell its the normal grouping set clause so I'm thinking we can skip the method and parse unconditionally

(the current setup seems to have similar impls/repr for group by clauses, other being `parse_group_by_expr`, which isn't ideal but I think that's out of scope for this PR).

src/dialect/mod.rs
```diff
@@ -245,6 +245,16 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialects supports `with rollup/cube/all` expressions.
```
iffyio, line 248:
```suggestion
    /// Returns true if the dialects supports `GROUP BY` modifiers prefixed by a `WITH` keyword.
    /// Example: `GROUP BY value WITH ROLLUP`.
```

src/ast/query.rs
```diff
@@ -2537,6 +2542,19 @@ impl fmt::Display for GroupByWithModifier {
...
+                    write!(f, "GROUPING SETS (")?;
+                    let mut sep = "";
+                    for set in sets {
+                        write!(f, "{sep}")?;
+                        sep = ", ";
+                        write!(f, "({})", display_comma_separated(set))?;
+                    }
+                    write!(f, ")")
+                }
+                _ => unreachable!(),
```
iffyio, line 2556: I think the panic here would be undesirable in any case. Can we reuse the existing impl on expr? its not clear to me why the display cares about the kind of expression here. e.g.
```rust
GroupByWithModifier::GroupingSets(expr) => {
    write!(f, "{expr}")
}
```

wugeer: Good idea! Have Done. :)

src/parser/mod.rs
wugeer: Have Done. :)

wugeer:
@iffyio Thank you for your guidance! I have resubmitted the code and hope to receive your review again. :)

iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

PR #1654, opened 2025-01-13
Support single line comments starting with '#' for Hive

wugeer:
The code of Apache Hive reveals that beeline also supports shell-style "#" prefix.  Currently, Hive mainly uses Beeline to execute SQL, so it might be necessary to support this here as well.
See https://github.com/apache/hive/blob/62ad7d42d6fd4e10ef5b86545ae41011554fd2f5/common/src/java/org/apache/hive/common/util/HiveStringUtils.java#L1170


iffyio (approved):
LGTM! Thanks @wugeer!
cc @alamb

PR #1655, opened 2025-01-15
Add support for `IS [NOT] [form] NORMALIZED`

alexander-beedie:
Adds parsing support for `IS [NOT] [form] NORMALIZED → bool` syntax:

Details from the PostgreSQL string function docs:
https://www.postgresql.org/docs/current/functions-string.html
```
Checks whether the string is in the specified Unicode normalization
form. The optional 'form' keyword specifies the form: NFC (the default),
NFD, NFKC, or NFKD. This expression can only be used when the server
encoding is UTF8. Note that checking for normalization using this
expression is often faster than normalizing possibly already
normalized strings.
```

* `NFC`: Canonical Decomposition, followed by Canonical Composition.
* `NFD`: Canonical Decomposition.
* `NFKC`: Compatibility Decomposition, followed by Canonical Composition.
* `NFKD`: Compatibility Decomposition.

As the normalised forms are fixed (there are only these four), it seemed reasonable to return the parsed form as a new `Option<NormalizationForm>` Enum (which helps the caller as they don't have to check the string or case-normalise it, and can then jump straight into some associated match block, etc).

(**Also**: fixed a few minor typos).

## Examples

Default/omitted `form`:
```sql
strcol IS NORMALIZED
strcol IS NOT NORMALIZED
```
Specific `form`:
```sql
strcol IS NFKC NORMALIZED
strcol IS NOT NFKD NORMALIZED
```


src/ast/mod.rs
```diff
@@ -1118,7 +1124,7 @@ impl fmt::Display for LambdaFunction {
 /// `OneOrManyWithParens` implements `Deref<Target = [T]>` and `IntoIterator`,
 /// so you can call slice methods on it and iterate over items
 /// # Examples
-/// Acessing as a slice:
+/// Accessing as a slice:
```
alamb, line 1127: Thank you for these cleanups

src/ast/value.rs
```diff
@@ -270,6 +270,27 @@ impl fmt::Display for DateTimeField {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum NormalizationForm {
+    NFC,
```
alamb, line 277: Bonus points if you added some comments here (maybe referencing the postgres docs and defining what the `NFC` and `NFD`, etc acronyms meant

I think you could copy from your very nice PR description 🙏

alamb (approved):
Thank you @alexander-beedie  -- I think this looks quite nice and well tested

fyi @iffyio

src/ast/value.rs
alexander-beedie: Good idea; done :)

src/parser/mod.rs
```diff
@@ -8452,6 +8458,42 @@ impl<'a> Parser<'a> {
...
+                    return self.expected("unicode normalization", next_token);
+                }
+            }
+        } else {
+            None
+        };
+        if self.parse_keyword(Keyword::NORMALIZED) {
+            return Ok((normalized_form, neg));
+        }
+        self.index = index;
```
iffyio, line 8493: Thinking it might be simpler to declare the strings as keywords, also instead of manually tracking the index in the function would it be possible to use `maybe_parse`? roughly thinking something like:
```rust
let normalized_form = self.maybe_parse(|parser| {
	match parser.parse_one_of_keywords(&[
		Keyword::NFC, ...
	]) {
		Some(Keyword::NFC) = Some(NormalizationForm::NFC)
		...
	}
})?;
if self.parse_keyword(Keyword::NORMALIZED) {
	return Ok((normalized_form, neg));
}
```

iffyio (commented):
Thanks @alexander-beedie!

src/parser/mod.rs
alexander-beedie: Nice hadn't previously spotted `maybe_parse`; I wasn't sure if making the normalization forms keywords would be seen as "noisy", but the end result definitely seems cleaner. Have updated the code accordingly ✌️

src/parser/mod.rs
```diff
@@ -8452,6 +8458,31 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a literal unicode normalization clause
+    pub fn parse_unicode_is_normalized(
+        &mut self,
+    ) -> Result<(Option<NormalizationForm>, bool), ParserError> {
```
iffyio, line 8464: Oh the return type becomes a bit surprising by returning a bool, since this is a public API would it make sense to have it return the expr instead `Expr::IsNormalized` ?

iffyio (commented):
Just one comment regarding the API signature, otherwise this looks good to me!

src/parser/mod.rs
alexander-beedie: Sure; updated it as `-> Result<Expr, ParserError>` 👍

iffyio (approved):

PR #1657, opened 2025-01-15
Allow empty options for BigQuery

MartinSahlen:
In BigQuery, it's possible to have statements like these (added tests for them):

```sql
CREATE TABLE foo (x INT64) OPTIONS()
```

```sql
CREATE TABLE db.schema.test (x INT64 OPTIONS(description = 'An optional INTEGER field')) OPTIONS()
```

While this may look silly (and clearly is), it's very common that companies are using tools (like dbt) to generate the SQL based on a certain configuration defined at runtime using macros and other abominations (but that is off-topic). 

This PR, with tests, allows specifying empty options. Seeing that all tests run, this can hopefully be merged quickly. Also ran `fmt` and `clippy`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7341,6 +7341,10 @@ impl<'a> Parser<'a> {
     pub fn parse_options(&mut self, keyword: Keyword) -> Result<Vec<SqlOption>, ParserError> {
         if self.parse_keyword(keyword) {
             self.expect_token(&Token::LParen)?;
+            if self.peek_token() == Token::RParen {
+                self.next_token();
+                return Ok(vec![]);
+            }
             let options = self.parse_comma_separated(Parser::parse_sql_option)?;
             self.expect_token(&Token::RParen)?;
             Ok(options)
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -2244,3 +2244,15 @@ fn test_any_type() {
 fn test_any_type_dont_break_custom_type() {
     bigquery_and_generic().verified_stmt("CREATE TABLE foo (x ANY)");
 }
+
+#[test]
+fn parse_create_table_with_empty_table_options() {
+    let sql = "CREATE TABLE foo (x INT64) OPTIONS()";
+    bigquery().verified_stmt(sql);
+}
+
+#[test]
+fn parse_create_table_with_empty_table_options_and_column_options() {
+    let sql = "CREATE TABLE db.schema.test (x INT64 OPTIONS(description = 'An optional INTEGER field')) OPTIONS()";
+    bigquery().verified_stmt(sql);
+}


src/parser/mod.rs
```diff
@@ -7341,6 +7341,10 @@ impl<'a> Parser<'a> {
     pub fn parse_options(&mut self, keyword: Keyword) -> Result<Vec<SqlOption>, ParserError> {
         if self.parse_keyword(keyword) {
             self.expect_token(&Token::LParen)?;
+            if self.peek_token() == Token::RParen {
+                self.next_token();
+                return Ok(vec![]);
+            }
             let options = self.parse_comma_separated(Parser::parse_sql_option)?;
             self.expect_token(&Token::RParen)?;
```
iffyio, line 7349:
```suggestion
            let options = self.parse_comma_separated0(Parser::parse_sql_option, &Token::RParen)?;
```
Would something like this work to simplify the condition?

tests/sqlparser_bigquery.rs
```diff
@@ -2244,3 +2244,15 @@ fn test_any_type() {
...
+fn parse_create_table_with_empty_table_options() {
+    let sql = "CREATE TABLE foo (x INT64) OPTIONS()";
+    bigquery().verified_stmt(sql);
+}
+
+#[test]
+fn parse_create_table_with_empty_table_options_and_column_options() {
+    let sql = "CREATE TABLE db.schema.test (x INT64 OPTIONS(description = 'An optional INTEGER field')) OPTIONS()";
+    bigquery().verified_stmt(sql);
+}
```
iffyio, line 2258: Can we add these as scenarios into the [existing tests for the feature](https://github.com/apache/datafusion-sqlparser-rs/blob/c8242d935d763f115cf7cb3efef98880da07f13d/tests/sqlparser_bigquery.rs#L475) vs having them as dedicated tests?

src/parser/mod.rs
MartinSahlen: I was not aware of this one, but it seems to break here:
```
thread 'test_bigquery_create_function' panicked at src/test_utils.rs:155:61:
CREATE OR REPLACE TEMPORARY FUNCTION project1.mydataset.myfunction(x FLOAT64) RETURNS FLOAT64 OPTIONS(x = 'y') AS 42: ParserError("Expected: AS, found: )")
```

when it looks like this
```rust
  pub fn parse_options(&mut self, keyword: Keyword) -> Result<Vec<SqlOption>, ParserError> {
        if self.parse_keyword(keyword) {
            let options = self.parse_comma_separated0(Parser::parse_sql_option, Token::RParen)?;
            Ok(options)
        } else {
            Ok(vec![])
        }
    }

```

I will work a bit to see what is going wrong

MartinSahlen: I tried changing both 
`parse_options` and `parse_options_with_keywords` as they excibit similar behaviour but I cannot get the tests to pass, although my original use case is working. It could be that I've stumbled upon some deeper issue or bug though (or I dont know enough about the code base which is clearly the case). The question is whether this potential bug should be addressed in this PR or not. Based on the expected bevahour of the function you suggested this seems like a bug.

MartinSahlen: Ok, I think I found the bug, if I do this: 
```rust
  pub fn parse_options(&mut self, keyword: Keyword) -> Result<Vec<SqlOption>, ParserError> {
        if self.parse_keyword(keyword) {
            let options = self.parse_comma_separated0(Parser::parse_sql_option, Token::RParen)?;
            Ok(options)
        } else {
            Ok(vec![])
        }
    }
```

its is working. And this leads me to look at the `parse_comma_separated0` implementation. Should this not also __consume__ the `end_token` parameter?

```rust
 pub fn parse_comma_separated0<T, F>(
        &mut self,
        f: F,
        end_token: Token,
    ) -> Result<Vec<T>, ParserError>
```

Ok I tried making this function consume the end_token but it seems to cause more issues. I also saw that there is a `parse_options_with_keywords` that at the moment will not accept an empty options, but I will not change or touch that in this PR.

tests/sqlparser_bigquery.rs
MartinSahlen: It shall be done! How did I miss that, I was searching but my bad!

iffyio (approved):
LGTM! Thanks @MartinSahlen!
cc @alamb

Issue #1660, opened 2025-01-16
`INSERT INTO ... SELECT ... RETURNING `


lovasoa:
The following is a valid and useful sql query:

```sql
INSERT INTO table1 SELECT * FROM table2 RETURNING id
```

However, we currently fail to parse it with: `ParserError("Expected: end of statement, found: id")`.

lovasoa:
This is because we currently allow `RETURNING` as an unquoted table alias identifier. Most DBs don't.

```sql
select * from demo returning
```

fails in
 - sqlite: `SQLITE_ERROR: sqlite3 result code 1: near "returning": syntax error` : https://sqliteonline.com/#sqltext=%23url-sqlite%3Ddb-sqlite%0D%0A%23tab-name%3DSQLite.3%0D%0Aselect%20*%20from%20demo%20returning
 - mariadb: `Help: Error 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'returning' at line 1`
 - postgres: `Help: Error 42601 syntax error at or near "returning"`

We should not allow `RETURNING` in that position either

lovasoa:
I am going to need this in the sqlpage spreadsheet component: https://github.com/sqlpage/sqlpage-spreadsheet

PR #1661, opened 2025-01-16
fix parsing of `INSERT INTO ... SELECT ... RETURNING `

lovasoa:
fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1660


iffyio (approved):
LGTM! Thanks @lovasoa!
cc @alamb

Issue #1658, opened 2025-01-16
where clause contains user column error


chenkovsky:
```
select * from t where user = 'test';
```

here user is treated as keyword.

https://github.com/apache/datafusion-sqlparser-rs/blob/474150006fc8dd299bfa212321206263de0f4290/src/parser/mod.rs#L1113

crepererum:
I think it's fine that this is a keyword, however as shown in https://github.com/apache/datafusion/issues/14141 it is parsed as a "function" -- which IMHO is wrong, no matter if the word `user` is reserved or not.

crepererum:
ah, I see you can indeed call functions w/o `()` in some dialects, well that's fun. So I think this case here might be fine.

PR #1662, opened 2025-01-17
feat: customize function keywords

chenkovsky:
this pr enables caller to decide whether user, current_user are columns or function calls.

close: #1658

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -340,7 +340,7 @@ impl<'a> Parser<'a> {
     /// ```
     /// # use sqlparser::{parser::{Parser, ParserError}, dialect::GenericDialect};
     /// # fn main() -> Result<(), ParserError> {
-    /// let dialect = GenericDialect{};
+    /// let dialect = GenericDialect::default();
     /// let statements = Parser::new(&dialect)
     ///   .try_with_sql("SELECT * FROM foo")?
     ///   .parse_statements()?;
@@ -368,7 +368,7 @@ impl<'a> Parser<'a> {
     /// ```
     /// # use sqlparser::{parser::{Parser, ParserError}, dialect::GenericDialect};
     /// # fn main() -> Result<(), ParserError> {
-    /// let dialect = GenericDialect{};
+    /// let dialect = GenericDialect::default();
     /// let result = Parser::new(&dialect)
     ///   .with_recursion_limit(1)
     ///   .try_with_sql("SELECT * FROM foo WHERE (a OR (b OR (c OR d)))")?
@@ -395,7 +395,7 @@ impl<'a> Parser<'a> {
     /// ```
     /// # use sqlparser::{parser::{Parser, ParserError, ParserOptions}, dialect::GenericDialect};
     /// # fn main() -> Result<(), ParserError> {
-    /// let dialect = GenericDialect{};
+    /// let dialect = GenericDialect::default();
     /// let options = ParserOptions::new()
     ///    .with_trailing_commas(true)
     ///    .with_unescape(false);
@@ -452,7 +452,7 @@ impl<'a> Parser<'a> {
     /// ```
     /// # use sqlparser::{parser::{Parser, ParserError}, dialect::GenericDialect};
     /// # fn main() -> Result<(), ParserError> {
-    /// let dialect = GenericDialect{};
+    /// let dialect = GenericDialect::default();
     /// let statements = Parser::new(&dialect)
     ///   // Parse a SQL string with 2 separate statements
     ///   .try_with_sql("SELECT * FROM foo; SELECT * FROM bar;")?
@@ -500,7 +500,7 @@ impl<'a> Parser<'a> {
     /// ```
     /// # use sqlparser::{parser::{Parser, ParserError}, dialect::GenericDialect};
     /// # fn main() -> Result<(), ParserError> {
-    /// let dialect = GenericDialect{};
+    /// let dialect = GenericDialect::default();
     /// let statements = Parser::parse_sql(
     ///   &dialect, "SELECT * FROM foo"
     /// )?;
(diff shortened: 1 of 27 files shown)


src/ast/mod.rs
```diff
@@ -5807,6 +5807,11 @@ pub struct Function {
     /// <aggregate_function>(expression) WITHIN GROUP (ORDER BY key [ASC | DESC], ...)
     /// ```
     pub within_group: Vec<OrderByExpr>,
+
+    /// Sometimes function names may also be column names,
+    /// for example, user, current_user,
+    /// it requires the user to determine whether it is a column name or a function name.
+    pub maybe_column: bool,
```
iffyio, line 5814: I think the desired behavior is that the keywords `USER`, `CURRENT_USER` etc are represented as identifiers in the scenarios where they are not functions. Having a flag on the function saying that the flag is in fact not a function sounds like it would be misleading/confusing behavior to the user

src/parser/mod.rs
```diff
@@ -14578,4 +14584,26 @@ mod tests {
 
         assert!(Parser::parse_sql(&MySqlDialect {}, sql).is_err());
     }
+
+    #[test]
+    fn test_keyword() {
+        let sql = "select * from t where user = 'test';";
```
iffyio, line 14590: oh actually the problem being solved isn't currently clear to me, my understanding is that in postgres for example, given this statement, `user` is indeed a function. What's the undesirable behavior being solved for by this PR?

MartinSahlen:
This is semantics, isnt it? Just like `current_timestamp` vs `current_timestamp` (i did a [post](https://www.linkedin.com/feed/update/urn:li:ugcPost:7275614401888440320?commentUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7275614401888440320%2C7275622633885335552%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287275622633885335552%2Curn%3Ali%3AugcPost%3A7275614401888440320%29) on linkedin about this, different engines interpret this differently. 

Isnt this a job for the AST traversal that is up the user of the library?

chenkovsky:
@MartinSahlen  @iffyio Please review it again. I know maybe it's not a good idea to solve it in this library.

chenkovsky:
there are so many `dialect_of!` in code, so it's hard to implement a dialect that behaviors similar to generic dialect except some differences. this pr make generic dialect customizable.

MartinSahlen:
> @MartinSahlen @iffyio Please review it again. I know maybe it's not a good idea to solve it in this library.

Given how I've shown that whether this is evaluated as a function or identifier/column depending on not just if it has parentheses or not, but also by what identifiers/columns are in scope I think this PR is not a good idea, and a bit of a slippery slope. It is easy to deal with this using a visitor pattern, which is what I think it is designed for - so this library can focus on syntax.

chenkovsky:
@MartinSahlen thanks for your suggestion!

chenkovsky:
> > @MartinSahlen @iffyio Please review it again. I know maybe it's not a good idea to solve it in this library.
> 
> Given how I've shown that whether this is evaluated as a function or identifier/column depending on not just if it has parentheses or not, but also by what identifiers/columns are in scope I think this PR is not a good idea, and a bit of a slippery slope. It is easy to deal with this using a visitor pattern, which is what I think it is designed for - so this library can focus on syntax.

@MartinSahlen  but i have a question, as you said,  `whether this is evaluated as a function or identifier/column depending on not just if it ...`, if this is correct. this library should not parse it into function call. it should leave it to caller.

MartinSahlen:
> > > @MartinSahlen @iffyio Please review it again. I know maybe it's not a good idea to solve it in this library.
> > 
> > 
> > Given how I've shown that whether this is evaluated as a function or identifier/column depending on not just if it has parentheses or not, but also by what identifiers/columns are in scope I think this PR is not a good idea, and a bit of a slippery slope. It is easy to deal with this using a visitor pattern, which is what I think it is designed for - so this library can focus on syntax.
> 
> @MartinSahlen but i have a question, as you said, `whether this is evaluated as a function or identifier/column depending on not just if it ...`, if this is correct. this library should not parse it into function call. this lib should leave it to caller.

I have not checked out the details of this. if the "user" function in postgres is indeed parsed as as function and not an identifier yes that goes against what I said. At the same time, if this is always true in postgres then maybe it does make sense. But following this logic then, we should not allow users to decide how the library should define functions. Then all dialects should have this built in according to their own semantics. Meaning: we should not expose this functionality to callers of the library, but rather point to rewriting the AST. 

All this being said, it's not fully black and white I guess - if there is indeed some instances of this dialect specific logic already my argument is of course weaker.

chenkovsky:
> > > > @MartinSahlen @iffyio Please review it again. I know maybe it's not a good idea to solve it in this library.
> > > 
> > > 
> > > Given how I've shown that whether this is evaluated as a function or identifier/column depending on not just if it has parentheses or not, but also by what identifiers/columns are in scope I think this PR is not a good idea, and a bit of a slippery slope. It is easy to deal with this using a visitor pattern, which is what I think it is designed for - so this library can focus on syntax.
> > 
> > 
> > @MartinSahlen but i have a question, as you said, `whether this is evaluated as a function or identifier/column depending on not just if it ...`, if this is correct. this library should not parse it into function call. this lib should leave it to caller.
> 
> I have not checked out the details of this. if the "user" function in postgres is indeed parsed as as function and not an identifier yes that goes against what I said. At the same time, if this is always true in postgres then maybe it does make sense. But following this logic then, we should not allow users to decide how the library should define functions. Then all dialects should have this built in according to their own semantics. Meaning: we should not expose this functionality to callers of the library, but rather point to rewriting the AST.
> 
> All this being said, it's not fully black and white I guess - if there is indeed some instances of this dialect specific logic already my argument is of course weaker.

for postgres, i'm agree it's good to convert it into function call here. but for generic dialect. it's a bit tricky. ☺️

PR #1663, opened 2025-01-17
Add support for qualified column names in JOIN ... USING

yoavcloud:
It appears that Snowflake allows specifying qualified column names in the USING column list. For example:
`SELECT * FROM tbl1 AS t1 JOIN tbl2 AS t2 USING(t2.col1)`

BTW, it's worthwhile exploring changing all occurrences to parse column lists as Vec<ObjectName> instead of Vec<Ident>, there are likely other statement patterns that get rejected...

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9336,7 +9336,8 @@ impl<'a> Parser<'a> {
         })
     }
 
-    /// Parse a parenthesized comma-separated list of unqualified, possibly quoted identifiers
+    /// Parses a parenthesized comma-separated list of unqualified, possibly quoted identifiers.
+    /// For example: `(col1, "col 2", ...)`
     pub fn parse_parenthesized_column_list(
         &mut self,
         optional: IsOptional,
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -9342,12 +9342,37 @@ impl<'a> Parser<'a> {
...
+    }
+
+    /// Parses a parenthesized comma-separated list of unqualified, possibly quoted identifiers
+    pub fn parse_parenthesized_qualified_column_list(
+        &mut self,
+        optional: IsOptional,
+        allow_empty: bool,
+    ) -> Result<Vec<ObjectName>, ParserError> {
+        self.parse_parenthesized_column_list_inner(optional, allow_empty, |p| {
+            p.parse_object_name(false)
```
iffyio, line 9355: Should the `in_table_clause` be configurable here? I thinking it's supposed to be set to true when parsing a `USING` clause given the behavior being fixed

tests/sqlparser_snowflake.rs
```diff
@@ -2650,6 +2650,9 @@ fn asof_joins() {
               "ON s.state = p.state ",
           "ORDER BY s.observed",
     ));
+
+    // Snowflake allows fully-qualified column names inside USING
+    snowflake().verified_stmt("SELECT * FROM tbl1 AS t1 JOIN tbl2 AS t2 USING(t2.col1)");
```
iffyio, line 2655: maybe we can add this scenario to the `parse_joins_using` test in common.rs instead?

src/parser/mod.rs
```diff
@@ -9342,12 +9342,37 @@ impl<'a> Parser<'a> {
         optional: IsOptional,
         allow_empty: bool,
     ) -> Result<Vec<Ident>, ParserError> {
+        self.parse_parenthesized_column_list_inner(optional, allow_empty, |p| p.parse_identifier())
+    }
+
+    /// Parses a parenthesized comma-separated list of unqualified, possibly quoted identifiers
```
iffyio, line 9348: The doc comment doesn't seem to match the behavior (unqualified vs qualified)? also it could be nice if the doc highlights that the parenthesis is also expected in the input (maybe with an example)

iffyio (commented):
Thanks @yoavcloud! Left some comments

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1664, opened 2025-01-17
Add support for Create Iceberg Table statement for Snowflake parser

Vedin:
Add support for Snowflake `CREATE ICEBERG TABLE` statement. https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table. 
I decided not to extend the `CREATE TABLE` statement because it seems to be different and `CREATE ICEBERG TABLE` only supported in Snowflake.


src/ast/mod.rs
```diff
@@ -2516,6 +2516,35 @@ pub enum Statement {
     /// CREATE TABLE
     /// ```
     CreateTable(CreateTable),
+    /// ``` sql
+    /// CREATE ICEBERG TABLE
+    /// Snowflake-specific statement
+    /// <https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table>
+    /// ```
+    CreateIcebergTable {
```
iffyio, line 2524: The preferred path would be to reuse the `CreateTable` variant, by e.g. adding a `is_iceberg_table: bool` to flag the table type and then including any iceberg table fields as optional/empty

Vedin: Ok, sure. I just noticed that `CreateVirtualTable` for SQLLite is a separate statement.

iffyio (approved):
LGTM! Thanks @Vedin!
cc @alamb

PR #1669, opened 2025-01-20
Parse Snowflake COPY INTO <location>

yoavcloud:
This PR adds support for parsing Snowflake's `COPY INTO <location>`, makes a distinction between `COPY INTO <table>`, and reuses much of its logic.

### DIFF
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -330,7 +330,7 @@ impl Spanned for Statement {
                 legacy_options: _,
                 values: _,
             } => source.span(),
-            Statement::CopyIntoSnowflake {
+            Statement::CopyIntoSnowflakeTable {
                 into: _,
                 from_stage: _,
                 from_stage_alias: _,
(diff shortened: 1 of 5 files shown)


src/ast/spans.rs
```diff
@@ -342,6 +343,15 @@ impl Spanned for Statement {
                 copy_options: _,
                 validation_mode: _,
             } => Span::empty(),
+            Statement::CopyIntoSnowflakeLocation {
```
alamb, line 346: `CopyIntoSnowflake` shares `into`, `partition`, `file_format`, `copy_options` fields  -- did you consider keeping `Statement::CopyIntoSnowflake` and add an enum like `target` to distinguish the two forms?

alamb (commented):
Thanks @yoavcloud

yoavcloud:
@alamb reminder to look at the updated PR based on your suggestion

iffyio (approved):
LGTM! Thanks @yoavcloud!

alamb:
🚀

PR #1672, opened 2025-01-21
National strings: check if dialect supports backslash escape

hansott:


src/tokenizer.rs
```diff
@@ -3543,4 +3546,17 @@ mod tests {
         ];
         compare(expected, tokens);
     }
+
+    #[test]
+    fn test_national_strings() {
+        let dialect = PostgreSqlDialect {};
```
iffyio, line 3552: could we test the behavior both for dialects that support and don't support backslash? (We can probably reuse the `all_dialects_where` helper function).
Also maybe the test could be called `test_national_strings_backslash_escape` or similarly specific to the test is covering?

hansott: Great suggestion! How does it look now?

iffyio (approved):
LGTM! Thanks @hansott!
cc @alamb

PR #1675, opened 2025-01-22
Add support for mysql table hints

AvivDavid-Satori:


src/dialect/mysql.rs
```diff
@@ -25,6 +25,11 @@ use crate::{
     parser::{Parser, ParserError},
 };
 
+use super::keywords;
+
+pub const RESERVED_FOR_TABLE_ALIAS_MYSQL: &[Keyword] =
```
yoavcloud, line 30: Maybe keep this one private? It's not used outside of this module for now.

src/parser/mod.rs
```diff
@@ -8888,6 +8888,64 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_table_index_hints(&mut self) -> Result<Vec<TableIndexHints>, ParserError> {
+        let mut hints = vec![];
+        while let Some(hint_type) =
+            self.parse_one_of_keywords(&[Keyword::USE, Keyword::IGNORE, Keyword::FORCE])
```
iffyio, line 8894: Given the following which is valid BigQuery for example `SELECT * FROM T USE LIMIT 1` - it seems such queries will fail to parse with the current impl? If so, `self.maybe_parse` might be useful here to try to lookahead to parse a hint and keep going on failure. Also it would be good to test the behavior on mysql and another dialect (e.g. BigQuery) that doesn't support index hints

src/parser/mod.rs
```diff
@@ -8888,6 +8888,64 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_table_index_hints(&mut self) -> Result<Vec<TableIndexHints>, ParserError> {
```
iffyio, line 8891:
```suggestion
    fn parse_table_index_hints(&mut self) -> Result<Vec<TableIndexHints>, ParserError> {
```
I wonder if we need to have this function public, if we do feel free to leave as is and we can also include a doc string describing its behavior in that case?

src/parser/mod.rs
AvivDavid-Satori: Thanks, added a test to verify it

yoavcloud: I would consider gating the parsing of table hints for both MySQL and MSSQL using a Dialect.supports_table_hints method.

tests/sqlparser_bigquery.rs
```diff
@@ -1601,6 +1602,34 @@ fn parse_join_constraint_unnest_alias() {
     );
 }
 
+#[test]
+fn parse_select_with_use() {
+    let sql = "SELECT * FROM T USE LIMIT 1";
```
iffyio, line 1607: Regarding the test, it might be a bit odd to have this test in this file, specific to bigquery since the behavior isn't bigquery specific. This potentially relates to @yoavcloud's comment on gating this with a dialect flag, thinking that might also make the test easier to setup

I think what we could do in that case would be e.g.
- introduce a `dialects.supports_table_hints()` flag to gate the feature (we could still use the `maybe_parse` or if the grammar is unambiguous for the dialects that support table hints, then we the flag could replace `maybe_parse` entirely`
- We move the `parse_select_table_with_index_hints` from mysql to `sqlparser_common.rs` and to use the [all_dialects_where](https://github.com/apache/datafusion-sqlparser-rs/blob/ef072be9e1b1ecbf8032bd2040131a9d5b00de5d/tests/sqlparser_common.rs#L125) format
- We move this test case into that `parse_select_table_with_index_hints` test where it uses e.g.
   ```rust
   let unsupported_dialects = all_dialects_where(|d| !d.supports_table_index_hints());
   unsupported_dialects.verified_stmt(sql);
   ...
   ```

tests/sqlparser_mysql.rs
```diff
@@ -2898,6 +2900,21 @@ fn parse_lock_tables() {
     mysql().verified_stmt("UNLOCK TABLES");
 }
 
+#[test]
+fn parse_select_table_with_index_hints() {
+    mysql()
```
iffyio, line 2905: for one of these scenarios can we assert that the `index_hints` property of the returned statement was in fact populated with the expected result?

iffyio (commented):
Thanks @AvivDavid-Satori! The changes look reasonable to me overall, left a couple comments regarding the test setup

tests/sqlparser_mysql.rs
AvivDavid-Satori: Thanks, inserted changes

iffyio (approved):
LGTM! Thanks @AvivDavid-Satori!
cc @alamb

iffyio:
@AvivDavid-Satori could you merge in latest main? There's since been some api changes to `ObjectName` that would be needed in this branch before merging

PR #1677, opened 2025-01-22
Support underscore separators in numbers for Clickhouse. Fixes #1659

### LINKED ISSUE #1659: Clickhouse: Support underscore in numbers
See https://clickhouse.com/docs/en/sql-reference/syntax#numeric

```sql
SELECT 10_000_000
```

```
[
    Word(
        Word {
            value: "SELECT",
            quote_style: None,
            keyword: SELECT,
        },
    ),
    Whitespace(
        Space,
    ),
    Number(
        "10",
        false,
    ),
    Word(
        Word {
            value: "_000_000",
            quote_style: None,
            keyword: NoKeyword,
        },
    ),
]

### PR DESCRIPTION
**Pull request by graup**
Fixes #1659

Adds support for underscore separators in numeric values by allowing them to appear among digits during tokenization.

This comes out of the parser as a Value with the same string content. I considered removing the underscore during parsing but since this project errs more on the side of "no semantics" I think it's maybe better to just keep it.

Not sure that additional test on the AST level is really needed, it basically just tests that the parser doesn't do anything with or throw on these expressions.

### CODE REVIEW

File: src/dialect/mod.rs [hunk 1]
```diff
@@ -304,6 +304,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports numbers containing underscores
```
iffyio, line 307:
```suggestion
    /// Returns true if the dialect supports numbers containing underscores. e.g. `10_000_000`
```
Thinking an example would clarify even further what it implies

src/dialect/mod.rs
```diff
@@ -304,6 +304,11 @@ pub trait Dialect: Debug + Any {
...
+    fn supports_underscore_separator(&self) -> bool {
```
iffyio, line 308:
```suggestion
    fn supports_numeric_literal_underscores(&self) -> bool {
```
something like this? thinking `underscore_operator` wouldn't be descriptive on its own

src/tokenizer.rs
```diff
@@ -2223,6 +2227,22 @@ mod tests {
         compare(expected, tokens);
     }
 
+    #[test]
+    fn tokenize_clickhouse_underscore_separator() {
```
iffyio, line 2231:
```suggestion
    fn tokenize_numeric_literal_underscore() {
```

iffyio (commented):
Thanks @graup! Left some minor comments, this looks good to me overall!

graup:
Thanks for the review @iffyio, I have applied your suggestions.

iffyio (approved):
LGTM! Thanks @graup!
cc @alamb

iffyio:
@graup could you take a look at the ci failures when you get a chance?

graup:
Thanks, should be fixed. (I found the helpful ` HINT: use test_utils::number` 😄)

graup:
Yeah, seems like this number behavior is still failing under the bignumbers feature. I can reproduce it locally with `cargo test --all-features`. I don't know enough to fix this sadly :/

src/tokenizer.rs
```diff
@@ -1147,7 +1147,11 @@ impl<'a> Tokenizer<'a> {
                         s.push('.');
                         chars.next();
                     }
-                    s += &peeking_take_while(chars, |ch| ch.is_ascii_digit());
+
+                    s += &peeking_take_while(chars, |ch| {
+                        ch.is_ascii_digit()
+                            || self.dialect.supports_numeric_literal_underscores() && ch == '_'
```
hansott, line 1153: What happens if starts or ends with `_`? 

I think Clickhouse rejects that: https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/StringUtils.h#L171-L182

mwylde:
This is great! Underscores are also supported in Postgres 16 (https://www.postgresql.org/docs/16/release-16.html#RELEASE-16-DATATYPES), so maybe it could be added to the postgres dialect as well?

graup:
Thanks for the comments! Seems Postgres ([patch](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=faff8f8e4)) is using the same syntax rule as Clickhouse ([code](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/StringUtils.h#L171-L182)). I'll fix this up on Monday.

In the meantime, if anyone has an idea how to address the test failure, I'd love some help with that.

Maybe it's better to drop the underscores in the parsing stage, after all?

tests/sqlparser_clickhouse.rs
```diff
@@ -1646,6 +1646,16 @@ fn parse_table_sample() {
     clickhouse().verified_stmt("SELECT * FROM tbl SAMPLE 1 / 10 OFFSET 1 / 2");
 }
 
+#[test]
+fn parse_numbers_with_underscore() {
+    let select = clickhouse().verified_only_select("SELECT 10_000");
```
iffyio, line 1651: @graup regarding the test failure, its due to the bigdecimal representation when the feature flag it enabled, could you try something like this for the assertion?
```rust
let canonical = if cfg!(feature = "bigdecimal") {
    "SELECT 10000"
} else {
    "SELECT 10_000"
};
let select = clickhouse().one_statement_parses_to(
    "SELECT 10_000",
    canonical,
);
// ...
```

src/tokenizer.rs
iffyio: It could be worth having a test to demonstrate the behavior with a trailing underscore

graup:
I fixed the test and implemented the syntax logic (one underscore at time, must be followed by another digit). We could consider throwing a syntax error (as the engines that supports this do), but on the other hand, we also didn't throw an error before – I have added a test that verifies the previous behavior).

src/tokenizer.rs
```diff
@@ -2227,6 +2260,43 @@ mod tests {
...
+            Token::Comma,
+            Token::Whitespace(Whitespace::Space),
+            Token::Number("10_00".to_string(), false),
+            Token::make_word("_", None), // trailing underscores tokenizes as a word (syntax error in some dialects)
+            Token::Comma,
+            Token::Whitespace(Whitespace::Space),
+            Token::Number("10".to_string(), false),
+            Token::make_word("___0", None), // multiple underscores tokenizes as a word (syntax error in some dialects)
+        ];
+        compare(expected, tokens);
```
hansott, line 2297: There's a utility `.tokenizes_to(...)` that might be useful for tests like this :)

src/tokenizer.rs
```diff
@@ -1946,6 +1961,24 @@ fn peeking_take_while(chars: &mut State, mut predicate: impl FnMut(char) -> bool
     s
 }
 
+// Same as peeking_take_while, but also passes the next character to the predicate.
```
iffyio, line 1964:
```suggestion
/// Same as peeking_take_while, but also passes the next character to the predicate.
```

iffyio (commented):
Thanks @graup! One minor edit on the doc formatting otherwise I think this looks good!

PR #1678, opened 2025-01-22
BigQuery: Fix column identifier reserved keywords list

iffyio:
Parser currently uses the same `RESERVED_FOR_COLUMN_ALIAS` keywords list to avoid lookahead when parsing column identifiers. This assumed that all listed keywords were reservered across all dialects which isn't the case.
So that the following valid BigQuery statement previously failed due to `OFFSET` being flagged as a keyword.
```sql
SELECT 1, OFFSET FROM T
```
This updates the parser to support dialect specific `RESERVED_FOR_COLUMN_ALIAS` list

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -788,7 +788,7 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_IDENTIFIER.contains(&kw)
     }
 
-    // Returns reserved keywords when looking to parse a [TableFactor].
+    /// Returns reserved keywords when looking to parse a `TableFactor`.
     /// See [Self::supports_from_trailing_commas]
     fn get_reserved_keywords_for_table_factor(&self) -> &[Keyword] {
         keywords::RESERVED_FOR_TABLE_FACTOR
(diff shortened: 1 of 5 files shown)


src/dialect/mod.rs
```diff
@@ -821,11 +821,24 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns reserved keywords when looking to parse column
+    /// identifiers.
+    /// i.e. (unquoted) words that may not be used as column identifiers
+    /// for the dialect.
+    /// For example given: `SELECT <col> AS <alias> FROM T`. The returned
+    /// keywords are not allowed in `<col>` and `<alias>`
+    fn get_reserved_keywords_for_column_identifier(&self) -> &'static [Keyword] {
```
yoavcloud, line 830: Theoretically, a dialect may have a different list of reserved keywords for column identifiers and column aliases, and in some cases, the parser will need to look ahead to make that decision accurately. 

Perhaps we should adopt a unified approach to this check along the lines of `Dialect::is_select_item_alias`?

tests/sqlparser_bigquery.rs
```diff
@@ -213,6 +213,15 @@ fn parse_raw_literal() {
     );
 }
 
+#[test]
+fn parse_big_query_non_reserved_column_alias() {
+    let sql = r#"SELECT OFFSET, EXPLAIN, ANALYZE, SORT, TOP, VIEW FROM T"#;
```
alamb, line 218: 🤯

src/dialect/mod.rs
alamb: I agree this would be nice -- perhaps we can do so in a follow on PR

alamb (approved):
Thanks @iffyio  and @yoavcloud  -- this looks good to me.

src/dialect/mod.rs
iffyio: I think this sounds good! I'll take a look at it tomorrow and I'll ping to take another look at the PR

iffyio: I've updated this now to be similar to `is_select_item_alias`, taking in a parser, PTAL!

alamb (approved):
👨‍🍳 👌

PR #1679, opened 2025-01-23
Make TypedString preserve quote style

### LINKED ISSUE #1673: Fails to parse TypedString expressions in BigQuery with triple quoted strings
i.e 

`select JSON """{"name": "Alice", "age": 30}"""`

fails. More examples can also be found here. 

I also tried to do things like `SELECT timestamp """{"name": "Alice", "age": 30}"""` which also fails. 

Error message is:

```
called `Result::unwrap()` on an `Err` value: ParserError("Expected: end of statement, found: \"\"\"{\"name\": \"Alice\", \"age\": 30}\"\"\" at Line: 7, Column: 13")
```

graup:
Fixes #1673. This is a breaking change.

TypedString contained a `String` without any knowledge of the used quote style. The parser used `parse_literal_string` to construct this, which doesn't support any quote styles other than single or double quotes. Namely, it doesn't support triple quotes from BigQuery, causing the issue reported in #1673. Additionally, it doesn't round-trip properly, always formatting its string using single quotes.

I think the most proper fix is to have `TypedString` contain a `Value` instead, similar to IntroducedString and others. This gives us immediate support for other quote styles and fixes the formatting to make it roundtrippable.

This is a breaking change but should be an easy fix in users' codebases, just (un)wrapping the value. Migration path:

1. When constructing an AST node

```diff
Expr::TypedString {
    data_type: DataType::JSON,
--  value: r#"{"class" : {"students" : [{"name" : "Jane"}]}}"#.to_string()
++  value: Value::SingleQuotedString(
++      r#"{"class" : {"students" : [{"name" : "Jane"}]}}"#.to_string()
++  )
},],
```

2. When using AST parser results

```diff
if let Expr::TypedString { data_type, value } = expr {
--  let string_value: String = value;
++  let string_value: String = value.into_string().unwrap();
}
```

For convenience, I have added a method `into_string -> Option(String)` to Value to get the underlying string value.


src/ast/value.rs
```diff
@@ -97,6 +97,32 @@ pub enum Value {
...
+            Value::SingleQuotedRawStringLiteral(s) => s,
+            Value::DoubleQuotedRawStringLiteral(s) => s,
+            Value::TripleSingleQuotedRawStringLiteral(s) => s,
+            Value::TripleDoubleQuotedRawStringLiteral(s) => s,
+            Value::NationalStringLiteral(s) => s,
+            Value::HexStringLiteral(s) => s,
+            Value::DoubleQuotedString(s) => s,
+            Value::Placeholder(s) => s,
+            Value::DollarQuotedString(s) => s.value,
+            _ => panic!("not a string value"),
```
graup, line 121: Not sure about this one. This panics for Number, Boolean, Null. Should be safe if users only access String values where they are supposed to be?

tests/sqlparser_bigquery.rs
```diff
@@ -2214,6 +2214,30 @@ fn test_select_as_value() {
...
+    if let Expr::TypedString { data_type, value } = expr {
+        assert_eq!(DataType::JSON, data_type);
+        let string_value: String = value.into();
+        assert_eq!(r#"{"foo":"bar's"}"#, string_value);
+    }
+
+    // SingleQuotedString and DoubleQuotedString are currently not correctly formatted by `fmt::Display for Value`.
+    // BigQuery does not support double escaping, should be \' or \" instead.
+    //bigquery().verified_expr(r#"JSON '{"foo":"bar\'s"}'"#);
+    //bigquery().verified_expr(r#"JSON "{\"foo\":\"bar's\"}""#);
```
iffyio, line 2238: is this part a TODO for this PR?

src/ast/value.rs
graup: Strictly speaking no, but without it, users that used the string value before have to put code like this in their code. Thought it would be nice offering a more easy migration by just adding `.into()` to the existing usage.

tests/sqlparser_bigquery.rs
```diff
@@ -48,34 +48,34 @@ fn parse_literal_string() {
     let select = dialect.verified_only_select(sql);
     assert_eq!(10, select.projection.len());
     assert_eq!(
-        &Expr::Value(Value::SingleQuotedString("single".to_string())),
+        &Expr::Value(Value::SingleQuotedString("single".into())),
```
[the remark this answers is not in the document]
graup: Yeah unrelated, was just tidying up to make the usage all consistent in the same file and not jump back and forth between into() and to_string(). Can revert this to make the PR simpler.

tests/sqlparser_bigquery.rs
```diff
@@ -39,7 +39,7 @@ fn parse_literal_string() {
         r#"'''triple-single'unescaped''', "#,
         r#""double\"escaped", "#,
         r#""""triple-double\"escaped""", "#,
-        r#""""triple-double"unescaped""""#,
+        r#""""triple-double"un'escaped""""#,
```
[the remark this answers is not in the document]
graup: Yes, because currently there is no test for a single quote inside a triple quoted string. Thought it's a good idea to test this formatting behavior. (Considering the other bug I found in the formatting.)

tests/sqlparser_bigquery.rs
graup: No, like I mentioned in the description, this would be better for another PR. I can remove those lines if you think that's better but I thought it's good to document this unintended behavior. (The test should be there, it's just failing at the moment.)

graup: The issue I ran into is this SQL `SELECT JSON """{"foo":"bar's"}"""` gets parsed and then re-printed as `SELECT JSON '{"foo":"bar's"}'` (note the not escaped `'`). But let me do some more tests – for now I'll take this out of this PR

src/ast/value.rs
graup: Thank you, that's a great suggestion. Have a look at my latest commit, does this make sense?

tests/sqlparser_bigquery.rs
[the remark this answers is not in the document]
graup: Good point, I changed it to add two cases instead of altering the existing one.

tests/sqlparser_bigquery.rs
MartinSahlen: I think we shouldnt need to set up unescape or other special modes to handle this case. The parser ingests valid input SQL and outputs buggy SQL in the case that @graup brings up so it's a pretty clear cut case of a bug?

src/ast/value.rs
```diff
@@ -97,6 +97,37 @@ pub enum Value {
...
 }
 
+impl Value {
+    /// Get the string value of a `Value` wrapping a string.
+    /// This includes all quotation styles
+    /// `{Single,Double,TripleSingle,TripleDouble}{QuotedString,RawString,ByteString}`
+    /// as well as `EscapedStringLiteral`, `UnicodeStringLiteral`, `NationalStringLiteral`,
+    /// `HexStringLiteral`, `DollarQuotedString`.
+    /// This will panic if called on a non-string variant like `Value::Number`` or `Value::Null`.
+    pub fn as_str(self) -> String {
```
iffyio, line 107:
```suggestion
    pub fn into_string(self) -> Option<String> {
```
The `into` prefix since this is consuming self and returning a new value (vs reference).
Then we want this to return None vs panic if the underlying literal isn't a string

src/ast/value.rs
```diff
@@ -97,6 +97,37 @@ pub enum Value {
     Placeholder(String),
 }
 
+impl Value {
+    /// Get the string value of a `Value` wrapping a string.
+    /// This includes all quotation styles
+    /// `{Single,Double,TripleSingle,TripleDouble}{QuotedString,RawString,ByteString}`
+    /// as well as `EscapedStringLiteral`, `UnicodeStringLiteral`, `NationalStringLiteral`,
+    /// `HexStringLiteral`, `DollarQuotedString`.
```
iffyio, line 105: hmm seems it might not be worth enumerating the types after all since we have a lot of them. We could say something like
```text
/// If the underlying literal is a string, regardless of quote style, returns the associated string value
```

graup:
Thanks @iffyio, appreciate your suggestions. I've changed the method to `into_string -> Option(String)`. Also updated the PR description.

PrettyWood:
Thank you so much for working on that 🙏  We have https://github.com/PRQL/prql/issues/5099 on prql side. Very much appreciated

iffyio (approved):
LGTM! Thanks @graup!
cc @alamb

alamb:
Werd! The PR / code 🚂  keeps on running. Thanks again @iffyio for all you do to keep this repo moving forward

PR #1680, opened 2025-01-23
BigQuery: Add support for select expr star

iffyio:
Adds support for wildcard expression on arbitrary
expressions in select lists.

Example:
```sql
SELECT l.LOCATION[offset(0)].* FROM l
```
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_expression_star

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -586,6 +586,20 @@ impl fmt::Display for Cte {
     }
 }
 
+/// Represents an expression behind a wildcard expansion in a projection.
+/// `SELECT T.* FROM T;
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum SelectItemQualifiedWildcardKind {
+    /// Expression is an object name.
+    /// e.g. `alias.*` or even `schema.table.*`
+    ObjectName(ObjectName),
+    /// Select star on an arbitrary expression.
+    /// e.g. `STRUCT<STRING>('foo').*`
+    Expr(Expr),
+}
+
 /// One item of the comma-separated list following `SELECT`
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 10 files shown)


src/ast/query.rs
```diff
@@ -586,6 +586,20 @@ impl fmt::Display for Cte {
...
+/// Represents an expression behind a wildcard expansion in a projection.
+/// `SELECT T.* FROM T;
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum SelectItemQualifiedWildcardKind {
+    /// Expression is an object name.
+    /// e.g. `alias.*` or even `schema.table.*`
+    ObjectName(ObjectName),
+    /// Select star on an arbitrary expression.
```
alamb, line 598: 👍

tests/sqlparser_common.rs
```diff
@@ -1050,6 +1053,84 @@ fn parse_column_aliases() {
...
 }
 
+#[test]
+fn parse_select_expr_star() {
+    let dialects = all_dialects_where(|d| d.supports_select_expr_star());
+
+    // Identifier wildcard expansion.
+    let select = dialects.verified_only_select("SELECT foo.bar.* FROM T");
+    let SelectItem::QualifiedWildcard(SelectItemQualifiedWildcardKind::ObjectName(object_name), _) =
+        only(&select.projection)
```
alamb, line 1063: `only` is a neat trick

alamb (approved):
Thanks @iffyio  -- this looks good to me. What mind blowing syntax

tests/sqlparser_common.rs
```diff
@@ -1050,6 +1053,84 @@ fn parse_column_aliases() {
...
+            ["foo", "bar"]
+                .into_iter()
+                .map(Ident::new)
+                .collect::<Vec<_>>()
+        ),
+        object_name
+    );
+
+    // Arbitrary compound expression with wildcard expansion.
+    let select = dialects.verified_only_select("SELECT foo - bar.* FROM T");
```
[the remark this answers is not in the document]
iffyio: :sweat_smile: thankfully the query isn't semantically valid, maybe the test is misleading in that sense, I added it only to verify that the parser accepts arbitrary expressions and wildcard on variables, but as far as binary operators specifically combined with the wildcard syntax wouldn't have meaning in BigQuery either

PR #1681, opened 2025-01-23
Support multiple tables in `UPDATE FROM` clause

iffyio:
Adds support for multiple tables in the `FROM` clause of an update statement.

```sql
UPDATE A
SET ..
FROM T, U, V
```


alamb (approved):
LGTM

FYI @yoavcloud

PR #1683, opened 2025-01-25
Enable GROUP BY exp for Snowflake dialect

yoavcloud:
Snowflake supports `GROUP BY <exp>` so just turning it on for this dialect


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1685, opened 2025-01-28
Make numeric literal underscore test dialect agnostic

iffyio:
Minor follow up from #1677
Support was added for both Clickhouse and Postgres but the test only covered Clickhouse. This move the test to common and to use the `all_dialects_where` for auto test coverage.


graup (approved):
LGTM

alamb (approved):
Thanks @iffyio  and @graup

Issue #1273, opened 2024-05-16
Support Snowflake lambda functions


jmhain:
Snowflake recently gained support for [higher-order functions](https://medium.com/snowflake/snowflake-supports-higher-order-functions-dfa4b7682f7a) like [`TRANSFORM`](https://docs.snowflake.com/en/sql-reference/functions/transform) and [`FILTER`](https://docs.snowflake.com/en/sql-reference/functions/filter) which utilize a new [lambda function syntax](https://docs.snowflake.com/en/user-guide/querying-semistructured#lambda-expressions) similar to that of [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-lambda-functions.html) for which I [recently added support](https://github.com/sqlparser-rs/sqlparser-rs/pull/1257).

The two syntaxes look compatible so we can mostly reuse what we have. The one difference I noted is that Snowflake lambdas are permitted to specify types, e.g. `(x INT, y INT) -> (x + y)`.

gstvg:
Support added by https://github.com/apache/datafusion-sqlparser-rs/pull/2192

LucaCappelletti94:
Closing as completed by https://github.com/apache/datafusion-sqlparser-rs/pull/2192

PR #1686, opened 2025-01-29
Extend lambda support for ClickHouse and DuckDB dialects

gstvg:
Extend lambda support landed on #1257 to more dialects (but not Snowflake, see #1273)

Returns true for `supports_lambda_functions` for ClickHouse and DuckDB ~~and Generic~~ dialects
~~Adds `supports_parensless_lambda_functions` to `Dialect`~~
~~Returns true for supports_parensless_lambda_functions for ClickHouse, Databricks and DuckDB dialects, but not Generic to no conflict with Postgres JSON~~

~~This is a breaking change because now to parse a parensless lambda  x -> x + 1 a dialect must also implement supports_parensless_lambda_functions~~

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -804,7 +804,7 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_IDENTIFIER.contains(&kw)
     }
 
-    // Returns reserved keywords when looking to parse a [TableFactor].
+    /// Returns reserved keywords when looking to parse a `TableFactor`.
     /// See [Self::supports_from_trailing_commas]
     fn get_reserved_keywords_for_table_factor(&self) -> &[Keyword] {
         keywords::RESERVED_FOR_TABLE_FACTOR
@@ -844,11 +844,17 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the specified keyword should be parsed as a column identifier.
+    /// See [keywords::RESERVED_FOR_COLUMN_ALIAS]
+    fn is_column_alias(&self, kw: &Keyword, _parser: &mut Parser) -> bool {
+        !keywords::RESERVED_FOR_COLUMN_ALIAS.contains(kw)
+    }
+
     /// Returns true if the specified keyword should be parsed as a select item alias.
     /// When explicit is true, the keyword is preceded by an `AS` word. Parser is provided
     /// to enable looking ahead if needed.
-    fn is_select_item_alias(&self, explicit: bool, kw: &Keyword, _parser: &mut Parser) -> bool {
-        explicit || !keywords::RESERVED_FOR_COLUMN_ALIAS.contains(kw)
+    fn is_select_item_alias(&self, explicit: bool, kw: &Keyword, parser: &mut Parser) -> bool {
+        explicit || self.is_column_alias(kw, parser)
     }
 
     /// Returns true if the specified keyword should be parsed as a table factor alias.
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -340,9 +340,18 @@ pub trait Dialect: Debug + Any {
     /// Returns true if the dialect supports lambda functions, for example:
     ///
     /// ```sql
-    /// SELECT transform(array(1, 2, 3), x -> x + 1); -- returns [2,3,4]
+    /// SELECT transform(array(1, 2, 3), (x) -> x + 1); -- returns [2,3,4]
     /// ```
     fn supports_lambda_functions(&self) -> bool {
+        false
+    }
+
+    /// Returns true if the dialect supports lambda functions without parentheses for a single argument, for example:
+    ///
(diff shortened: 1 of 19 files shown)


src/dialect/mod.rs
```diff
@@ -340,12 +340,21 @@ pub trait Dialect: Debug + Any {
...
     ///
     /// ```sql
-    /// SELECT transform(array(1, 2, 3), x -> x + 1); -- returns [2,3,4]
+    /// SELECT transform(array(1, 2, 3), (x) -> x + 1); -- returns [2,3,4]
     /// ```
     fn supports_lambda_functions(&self) -> bool {
         false
     }
 
+    /// Returns true if the dialect supports lambda functions without parentheses for a single argument, for example:
```
gstvg, line 349: What do you think, @jmhain ?

tests/sqlparser_common.rs
```diff
@@ -13285,3 +13285,98 @@ fn test_trailing_commas_in_from() {
         "SELECT 1, 2 FROM (SELECT * FROM t1), (SELECT * FROM t2)",
     );
 }
+
+#[test]
+fn test_lambdas() {
```
gstvg, line 13290: Moved from [databricks tests](https://github.com/apache/datafusion-sqlparser-rs/pull/1686/files#diff-6b67304a4e2562e3ae4f6adcc37e87530a5b65984f664f60bf776f93f47e7f28)

tests/sqlparser_databricks.rs
```diff
@@ -83,69 +83,6 @@ fn test_databricks_exists() {
     );
 }
 
-#[test]
```
gstvg, line 86: Moved into [common tests](https://github.com/apache/datafusion-sqlparser-rs/pull/1686/files#diff-ae2dab2c924f4018c1daf0743b8fd912c44c5b63ae7144dd8b951bb3bd254349)

tests/sqlparser_postgres.rs
```diff
@@ -2749,7 +2749,7 @@ fn test_json() {
     );
```
gstvg, line 2750: Assert that postgres json `->` still works

src/dialect/mod.rs
```diff
@@ -340,12 +340,21 @@ pub trait Dialect: Debug + Any {
...
+    ///
+    /// ```sql
+    /// SELECT transform(array(1, 2, 3), x -> x + 1); -- returns [2,3,4]
+    /// ```
+    fn supports_parensless_lambda_functions(&self) -> bool {
```
iffyio, line 354: If I understand correctly the idea with introducing this method is to avoid the Generic dialect's syntax conflict due to its pg json syntax support? If so I'm thinking it could more sense to turn off `supports_lambda_functions` for the Generic dialect instead, idea with the dialect is that it gets feature support by default only if there aren't conflicting syntax. So that if its not expected that a dialect supports `(x) -> y` but not `x -> y` then maybe Generic dialect shouldn't support lambdas after all

gstvg: You got it right. I think your point makes sense and I'm fine with turning off `supports_lambda_functions` for the Generic dialect, but should `supports_parensless_lambda_functions` be removed from the Dialect trait too? I want to run Datafusion with support for both lambdas (even if with limited syntax) and pg json syntax from [datafusion-functions-json](https://github.com/datafusion-contrib/datafusion-functions-json), and `supports_parensless_lambda_functions` allows me to use a custom dialect to do so. In the worst case I can run without pg syntax and support only direct function calls (`json_get`, etc)

cc @samuelcolvin in case you have interest in this too

samuelcolvin: I don't have much context here, personally I think we'll want to switch off lambdas.

iffyio: hmm, given expressions `x->y` or `(x)->y`: if a dialect supports both the pg json `->` operator and the lambda syntax then it seems to suggest an ambiguous grammar (it doesn't seem like there's a way to tell what either expression should be parsed into)?

My thinking was indeed to potentially turn off generic dialect and remove `supports_parensless_lambda_functions`. Latter sounds like we'd be introducing a behavior to the parser that is only relevant in certain cominbations, neither covered by any of the parser's supported dialects nor a sql spec for reference which spontaneously feels like a slippery slope.

samuelcolvin: Agreed. The lambda syntax looks identical to Pg JSON lookup syntax with a completely different meaning.

gstvg: Agreed. `supports_parensless_lambda_functions` is removed and lambda support restricted to clickhouse, databricks and duckdb. I actually forgot the nested expr `(expr)`, so yes, this conflicted with pg, it just wasn't caught by any test, plain wrong. Thanks :pray:

gstvg:
Looks like the CI errors are related to #1693

iffyio (approved):
LGTM! Thanks @gstvg and @samuelcolvin!
cc @alamb

PR #1687, opened 2025-01-29
Add RETURNS TABLE() support for CREATE FUNCTION in Postgresql

remysaissy:
Currently CREATE FUNCTION doesn't support the RETURNS TABLE datatype (https://www.postgresql.org/docs/15/sql-createfunction.html).

This PR adds it for the PostgresSQL and Generic dialect.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4563,14 +4563,7 @@ impl<'a> Parser<'a> {
         self.expect_token(&Token::RParen)?;
 
         let return_type = if self.parse_keyword(Keyword::RETURNS) {
-            if dialect_of!(self is PostgreSqlDialect | GenericDialect)
-                && self.parse_keyword(Keyword::TABLE)
-            {
-                let columns = self.parse_parenthesized_columns()?;
-                Some(DataType::Table(columns))
-            } else {
-                Some(self.parse_data_type()?)
-            }
+            Some(self.parse_data_type()?)
         } else {
             None
         };
@@ -8873,6 +8866,10 @@ impl<'a> Parser<'a> {
                 Keyword::ANY if self.peek_keyword(Keyword::TYPE) => {
                     let _ = self.parse_keyword(Keyword::TYPE);
                     Ok(DataType::AnyType)
+                }
+                Keyword::TABLE => {
+                    let columns = self.parse_returns_table_columns()?;
+                    Ok(DataType::Table(columns))
                 }
                 _ => {
                     self.prev_token();
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4563,7 +4563,14 @@ impl<'a> Parser<'a> {
         self.expect_token(&Token::RParen)?;
 
         let return_type = if self.parse_keyword(Keyword::RETURNS) {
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -4535,7 +4535,14 @@ impl<'a> Parser<'a> {
...
 
         let return_type = if self.parse_keyword(Keyword::RETURNS) {
-            Some(self.parse_data_type()?)
+            if dialect_of!(self is PostgreSqlDialect | GenericDialect)
+                && self.parse_keyword(Keyword::TABLE)
+            {
+                let columns = self.parse_parenthesized_columns()?;
+                Some(DataType::Table(columns))
+            } else {
+                Some(self.parse_data_type()?)
```
iffyio, line 4544: I'm thinking the behavior here could be a bit misleading, given the variant `DataType::Table`, the expectation is that `self.parse_data_type()` can parse that variant. Would that be reasonable to do here instead (i.e. the table type logic is moved into `parse_data_type()` and this part becomes unchanged [likely no need to guard by dialect either since the syntax doesn't conflict] I imagine as a result)?

src/parser/mod.rs
```diff
@@ -8866,6 +8873,24 @@ impl<'a> Parser<'a> {
...
+        let data_type = self.parse_data_type()?;
+        Ok(ColumnDef {
+            name,
+            data_type,
+            collation: None,
+            options: Vec::new(), // No constraints expected here
+        })
+    }
+
+    fn parse_parenthesized_columns(&mut self) -> Result<Vec<ColumnDef>, ParserError> {
```
iffyio, line 8887:
```suggestion
    fn parse_returns_table_columns(&mut self) -> Result<Vec<ColumnDef>, ParserError> {
```

src/parser/mod.rs
remysaissy: Yes sure, I was not sure that adding it to the parse_data_type() would not have side effects on other parts of the parser as the function is generic and not focused only on the function case.

iffyio (approved):
LGTM! Thanks @remysaissy!
cc @alamb

iffyio:
@remysaissy when you get the chance could you take a look to resolve the conflicts on the branch?

remysaissy:
Yes, I’ll do it today.

alamb:
Thanks @remysaissy  and @iffyio

alamb:
It looks like there are some real test failures

remysaissy:
@alamb, I don't see how the `parse_join_constraint_unnest_alias` (nightly build failure) is related to this commit since the difference is related to a Join becoming an Inner in the AST and the initial query is not a CREATE FUNCTION nor it uses the TABLE in it.

Regarding the parse_trigger_related_functions, I think it is related to another PR: https://github.com/apache/datafusion-sqlparser-rs/pull/1690 since the `parse_create_function` from beta and nightly failed on a query related to Hive. 
This has been fixed after my commit was merged but since the log stops at the first error so maybe there was other errors as well.

I propose to fix these failures in another PR that I will mark related to the #1690 if that's ok for you.

PR #1688, opened 2025-01-29
Fix bug when parsing a Snowflake stage with `;` suffix

yoavcloud:
…t after


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1689, opened 2025-01-29
Add parsing for GRANT ROLE and GRANT DATABASE ROLE in Snowflake dialect

yoavcloud:


src/ast/mod.rs
```diff
@@ -5912,6 +5924,7 @@ impl fmt::Display for GrantObjects {
                     display_comma_separated(schemas)
                 )
             }
+            GrantObjects::None => Ok(()),
```
iffyio, line 5927: Can we instead make `GrantObjects` optional at the usage sites that require it? i.e. `objects: Option<GrantObjects>`. Thinking so that we get the None representation by convention

yoavcloud: Good suggestion, fixed in the last commit

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

alamb:
Rolling along ☸️

PR #1690, opened 2025-01-29
Fix DDL generation in case of an empty arguments function.

remysaissy:
Functions with an empty argument list are properly parsed into the AST but the string representation of such function removes the parenthesis required after the function name, causing an error at the database level.

For eg.
`CREATE OR REPLACE FUNCTION no_arg() RETURNS VOID LANGUAGE plpgsql AS $$ BEGIN DELETE FROM my_table; END; $$`
Becomes:
`CREATE OR REPLACE FUNCTION no_arg RETURNS VOID LANGUAGE plpgsql AS $$ BEGIN DELETE FROM my_table; END; $$ 
`
And causes Postgres to reject the query with a 
`[96579] ERROR:  syntax error at or near "RETURNS"`

This PR fixes the string representation in this specific case.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2011,6 +2011,8 @@ impl fmt::Display for CreateFunction {
         )?;
         if let Some(args) = &self.args {
             write!(f, "({})", display_comma_separated(args))?;
+        } else {
+            write!(f, "()")?;
         }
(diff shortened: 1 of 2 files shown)


iffyio (approved):
Thanks @remysaissy!
cc @alamb

PR #1692, opened 2025-01-30
Allow plain JOIN without turning it into INNER

mvzink:
Although these are syntactic equivalents in some dialects, we can preserve the input syntax without less rewriting this way.


iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1693, opened 2025-01-30
Fix `CREATE FUNCTION` round trip for Hive dialect

iffyio:
Fixes the test failure in #1690


iffyio:
My bad, I somehow managed to miss that the test was failing before merging

iffyio:
Thanks @alamb!

PR #1696, opened 2025-01-31
Parse SET GLOBAL variable modifier for MySQL

### LINKED ISSUE #1694: Parse MySQL `SET GLOBAL` variables
In addition to `SESSION` and `LOCAL` qualifiers, MySQL allows `GLOBAL` for modifying system variables:

```
mysql> SET GLOBAL max_connections = 1000;
```

sqlparser currently interprets `global` as a variable name and fails to parse:

```
$ echo -n "SET GLOBAL max_connections = 1000;" | cargo run --example cli - --mysql
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
2025-01-31T00:44:22.858Z DEBUG [sqlparser::parser] Parsing sql 'SET GLOBAL max_connections = 1000;'...
Error during parsing: ParserError("Expected: equals sign or TO, found: max_connections at Line: 1, Column: 12")
```

mvzink:
This also stops rewriting `SESSION` away.

Closes #1694

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2889,7 +2889,7 @@ pub enum Statement {
     /// least MySQL and PostgreSQL. Not all MySQL-specific syntactic forms are
     /// supported yet.
     SetVariable {
-        local: bool,
+        scope: SetVariableScope,
         hivevar: bool,
         variables: OneOrManyWithParens<ObjectName>,
         value: Vec<Expr>,
@@ -4510,15 +4510,12 @@ impl fmt::Display for Statement {
                 write!(f, "SET{context_modifier} ROLE {role_name}")
             }
             Statement::SetVariable {
-                local,
+                scope,
                 variables,
                 hivevar,
                 value,
             } => {
-                f.write_str("SET ")?;
-                if *local {
-                    f.write_str("LOCAL ")?;
-                }
+                write!(f, "SET{scope} ")?;
                 let parenthesized = matches!(variables, OneOrManyWithParens::Many(_));
                 write!(
                     f,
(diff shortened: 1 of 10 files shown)


src/ast/mod.rs
```diff
@@ -8433,6 +8430,31 @@ impl fmt::Display for SessionParamValue {
...
+    None,
+}
+
+impl fmt::Display for SetVariableScope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Local => write!(f, " LOCAL"),
+            Self::Session => write!(f, " SESSION"),
+            Self::Global => write!(f, " GLOBAL"),
+            Self::None => write!(f, ""),
```
iffyio, line 8453: does it work the same to define scope as an optional field (`scope: Option<SetVariableScope>`)? in order to represent `None` by convention?

src/parser/mod.rs
```diff
@@ -10533,8 +10533,17 @@ impl<'a> Parser<'a> {
     }
 
     pub fn parse_set(&mut self) -> Result<Statement, ParserError> {
-        let modifier =
-            self.parse_one_of_keywords(&[Keyword::SESSION, Keyword::LOCAL, Keyword::HIVEVAR]);
+        let modifier_keywords = if self.dialect.supports_global_variable_modifier() {
```
iffyio, line 10536: I'm thinking having the lone `global` keyword as a dialect flag might not scale API wise, to make it dialect specific maybe it might make sense to have delegate to the dialect? like with [parse_column_option](https://github.com/apache/datafusion-sqlparser-rs/blob/c0b5b20d11cee97c9541e6671a9833e1c2f926e7/src/dialect/mod.rs#L622) as an example except with a default implementation that others like hive and mysql can override. 
e.g. 
```rust
let scope = self.dialect.parse_set_variable_scope(self)?;
```

mvzink:
I actually think between the point about API scalability and the further/future need for substantially different parsing for MySQL (#1697), this particular approach that just shoehorns in `GLOBAL` is not worthwhile. So I'm going to close this PR, and if I end up needing this in the future I'll try to do it right & cover the more general MySQL case.

PR #1698, opened 2025-02-01
Do not parse ASOF and MATCH_CONDITION as table factor aliases

yoavcloud:
Example 1 - `SELECT * FROM tbl ASOF`
`ASOF` is parsed as the alias for `tbl`, but on Snowflake, it's reserved for the `ASOF JOIN` option.

Example 2 - `SELECT * FROM tbl1 ASOF JOIN tbl2 MATCH_CONDITION(...)`
`MATCH_CONDITION` is parsed as the alias for `tbl2`, but on Snowflake, it's reserved for the `ASOF JOIN` option.

Added `ASOF` AND `MATCH_CONDITION` as reserved keywords for table alias and added a test for the ASOF JOIN functionality without explicit aliases.

### DIFF
--- a/src/keywords.rs
+++ b/src/keywords.rs
@@ -1000,6 +1000,8 @@ pub const RESERVED_FOR_TABLE_ALIAS: &[Keyword] = &[
     Keyword::ANTI,
     Keyword::SEMI,
     Keyword::RETURNING,
+    Keyword::ASOF,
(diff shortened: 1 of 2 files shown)


iffyio (approved):
Thanks @yoavcloud!
cc @alamb

PR #1702, opened 2025-02-04
Add suppport for Show Objects statement for the Snowflake parser

DanCodedThis:
It's a naive implementation, since `ShowStatementOptions` allow for cases that are not supported by Snowflake `SHOW OBJECTS` like:
- `SHOW TERSE OBJECTS WHERE ... ...`
- Instead of `SHOW TERSE OBJECTS LIKE '%test%' ...`

Snowflake spec: https://docs.snowflake.com/en/sql-reference/sql/show-objects


src/ast/mod.rs
```diff
@@ -2980,6 +2980,32 @@ pub enum Statement {
...
+    ///                              }
+    ///                        ]
+    ///                        [ STARTS WITH '<name_string>' ]
+    ///                        [ LIMIT <rows> [ FROM '<name_string>' ] ]
+    /// ```
+    /// Snowflake-specific statement
+    /// <https://docs.snowflake.com/en/sql-reference/sql/show-objects>
+    ShowObjects {
+        terse: bool,
+        show_options: ShowStatementOptions,
```
iffyio, line 3006: Could we use a struct to wrap the arguments? we're currently trying to move away from the anonymous struct pattern (https://github.com/apache/datafusion-sqlparser-rs/issues/1204)
Something like
```rust
struct ShowObjects {
         terse: bool,
        show_options: ShowStatementOptions,
}
Statement::ShowObjects(ShowObjects)
```

src/ast/mod.rs
```diff
@@ -2980,6 +2980,32 @@ pub enum Statement {
...
+    ///                                SCHEMA <schema_name>                            |
+    ///                                <schema_name>
+    ///
+    ///                                APPLICATION <application_name>                  |
+    ///                                APPLICATION PACKAGE <application_package_name>  |
+    ///                              }
+    ///                        ]
+    ///                        [ STARTS WITH '<name_string>' ]
+    ///                        [ LIMIT <rows> [ FROM '<name_string>' ] ]
+    /// ```
```
iffyio, line 3001: Thinking since this is quite verbose we can either leave it out, relying on the link to the documentation and a description or we can use a shorter summary of the syntax like an example sql statement?

src/parser/mod.rs
```diff
@@ -14172,7 +14172,7 @@ impl<'a> Parser<'a> {
         false
     }
 
-    fn parse_show_stmt_options(&mut self) -> Result<ShowStatementOptions, ParserError> {
+    pub fn parse_show_stmt_options(&mut self) -> Result<ShowStatementOptions, ParserError> {
```
iffyio, line 14175:
```suggestion
    pub(crate) fn parse_show_stmt_options(&mut self) -> Result<ShowStatementOptions, ParserError> {
```

src/dialect/snowflake.rs
```diff
@@ -182,6 +183,15 @@ impl Dialect for SnowflakeDialect {
             return Some(parse_file_staging_command(kw, parser));
         }
 
+        if parser.parse_keyword(Keyword::SHOW) {
+            let terse = parser.parse_keyword(Keyword::TERSE);
+            if parser.parse_keyword(Keyword::OBJECTS) {
+                return Some(parse_show_objects(terse, parser));
+            } else {
+                return Some(parser.parse_show());
```
iffyio, line 191: what would be the behavior if `terse` is true here, can we maybe include a test case demonstrating that scenario?

tests/sqlparser_snowflake.rs
```diff
@@ -2975,6 +2975,25 @@ fn test_parse_show_schemas() {
     snowflake().verified_stmt("SHOW SCHEMAS IN DATABASE STARTS WITH 'abc' LIMIT 20 FROM 'xyz'");
 }
 
+#[test]
+fn test_parse_show_objects() {
+    snowflake().verified_stmt("SHOW OBJECTS");
```
iffyio, line 2980: Since this introduces a new node in the AST, can we add an assertion for one of these test cases covering that the AST looks like what is expected? i.e a test in [this syle](https://github.com/apache/datafusion-sqlparser-rs/blob/e13a1b844b176746dbc5f04987e0b97d78df88b4/tests/sqlparser_snowflake.rs#L51-L59)

iffyio (commented):
Thanks @DanCodedThis! Left some comments

src/dialect/snowflake.rs
DanCodedThis: I think I fixed this the right way, shouldn't need tests now.

DanCodedThis: Added the `terse` test in other relevant snowflake `SHOW` statement tests, just in case (checked, they all worked and the `ShowObjects` also)

iffyio (approved):
LGTM! Thanks @DanCodedThis!
cc @alamb

PR #1705, opened 2025-02-04
Require space after -- to start single line comment in MySQL

hansott:
- Docs https://dev.mysql.com/doc/refman/8.4/en/ansi-diff-comments.html
- Lexer https://github.com/mysql/mysql-server/blob/trunk/sql/sql_lex.cc#L1465-L1472
- Char checks
```cpp
static constexpr uint8_t MY_CHAR_CTR =  040; /* Control character */

inline bool my_iscntrl(const CHARSET_INFO *cs, char ch) {
  return ((cs->ctype + 1)[static_cast<uint8_t>(ch)] & MY_CHAR_CTR) != 0;
}

static constexpr uint8_t MY_CHAR_SPC =  010; /* Spacing character */

inline bool my_isspace(const CHARSET_INFO *cs, char ch) {
  return ((cs->ctype + 1)[static_cast<uint8_t>(ch)] & MY_CHAR_SPC) != 0;
}
```

It's not super clear which characters belong to `my_iscntrl` and `my_isspace` but we cannot use `char.is_whitespace()` as it includes `'\n'`?

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -881,12 +881,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns whether it's the start of a single line comment
-    /// e.g. MySQL requires a space after `--` to be a single line comment
-    /// Otherwise it's interpreted as a double minus operator
+    /// Returns true if this dialect requires a whitespace character after `--` to start a single line comment.
     ///
     /// MySQL: <https://dev.mysql.com/doc/refman/8.4/en/ansi-diff-comments.html>
-    fn requires_whitespace_to_start_comment(&self) -> bool {
+    /// e.g. UPDATE account SET balance=balance--1
+    //       WHERE account_id=5752             ^^^ will be interpreted as two minus signs instead of a comment
+    fn requires_single_line_comment_whitespace(&self) -> bool {
         false
     }
 }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -880,6 +880,15 @@ pub trait Dialect: Debug + Any {
     fn supports_table_hints(&self) -> bool {
         false
     }
+
+    /// Returns whether it's the start of a single line comment
+    /// e.g. MySQL requires a space after `--` to be a single line comment
+    /// Otherwise it's interpreted as a double minus operator
+    ///
+    /// MySQL: <https://dev.mysql.com/doc/refman/8.4/en/ansi-diff-comments.html>
+    fn is_start_of_single_line_comment(&self, _ch: char) -> bool {
+        true
+    }
 }
 
 /// This represents the operators for which precedence must be defined
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -886,8 +886,8 @@ pub trait Dialect: Debug + Any {
     /// Otherwise it's interpreted as a double minus operator
     ///
     /// MySQL: <https://dev.mysql.com/doc/refman/8.4/en/ansi-diff-comments.html>
-    fn is_start_of_single_line_comment(&self, _ch: char) -> bool {
-        true
+    fn requires_whitespace_to_start_comment(&self) -> bool {
+        false
     }
 }
 
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -125,6 +125,15 @@ impl Dialect for MySqlDialect {
     fn supports_table_hints(&self) -> bool {
         true
     }
+
+    /// Returns whether it's the start of a single line comment
+    /// e.g. MySQL requires a space after `--` to be a single line comment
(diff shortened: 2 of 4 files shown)


src/dialect/mod.rs
```diff
@@ -880,6 +880,15 @@ pub trait Dialect: Debug + Any {
     fn supports_table_hints(&self) -> bool {
         false
     }
+
+    /// Returns whether it's the start of a single line comment
+    /// e.g. MySQL requires a space after `--` to be a single line comment
+    /// Otherwise it's interpreted as a double minus operator
+    ///
+    /// MySQL: <https://dev.mysql.com/doc/refman/8.4/en/ansi-diff-comments.html>
+    fn requires_whitespace_to_start_comment(&self) -> bool {
```
iffyio, line 889:
```suggestion
    fn requires_single_line_comment_whitespace(&self) -> bool {
```
thinking something like this to flag that this is only for the `--` comment style

src/dialect/mod.rs
```diff
@@ -880,6 +880,15 @@ pub trait Dialect: Debug + Any {
     fn supports_table_hints(&self) -> bool {
         false
     }
+
+    /// Returns whether it's the start of a single line comment
```
iffyio, line 884:
```suggestion
    /// Returns true if this dialect requires a whitespace after `--` for single line comments.
```

src/dialect/mysql.rs
```diff
@@ -125,6 +125,15 @@ impl Dialect for MySqlDialect {
     fn supports_table_hints(&self) -> bool {
         true
     }
+
+    /// Returns whether it's the start of a single line comment
+    /// e.g. MySQL requires a space after `--` to be a single line comment
+    /// Otherwise it's interpreted as a double minus operator
+    ///
```
iffyio, line 132: Since this is already duplicated at the definition of the dialect method, we could consider not including it (The link to the doc on the other hand is useful in any case)

src/tokenizer.rs
```diff
@@ -1229,15 +1229,31 @@ impl<'a> Tokenizer<'a> {
...
-                        Some('-') => {
-                            chars.next(); // consume the second '-', starting a single-line comment
+
+                    if let Some('-') = chars.peek() {
+                        let mut is_comment = true;
+                        if self.dialect.requires_whitespace_to_start_comment() {
+                            // MySQL requires a space after the -- for a single-line comment
+                            // Otherwise it's interpreted as two minus signs
+                            // e.g. UPDATE account SET balance=balance--1
+                            //      WHERE account_id=5752;
```
iffyio, line 1239: Can we move the example to the doc string in the dialect method? That would help clarify the behavior to folks without having to dig deeper in the docs or the parser code.
Then for the rest of the comment here we can probably skip since we've already documented as much in the dialect method

src/tokenizer.rs
```diff
@@ -1229,15 +1229,31 @@ impl<'a> Tokenizer<'a> {
                 // operators
                 '-' => {
                     chars.next(); // consume the '-'
-                    match chars.peek() {
-                        Some('-') => {
-                            chars.next(); // consume the second '-', starting a single-line comment
+
+                    if let Some('-') = chars.peek() {
```
iffyio, line 1233: would it be possible to have this remain as a match statement, calling `chars.peek()` only once? i.e.
```rust
match chars.peek() {
    Some('-') => { ... }
    Some('>') => { ... }
 }
```

hansott:
@iffyio Addressed all your comments, great feedback.

src/tokenizer.rs
```diff
@@ -1229,14 +1229,26 @@ impl<'a> Tokenizer<'a> {
...
                         Some('-') => {
-                            chars.next(); // consume the second '-', starting a single-line comment
-                            let comment = self.tokenize_single_line_comment(chars);
-                            Ok(Some(Token::Whitespace(Whitespace::SingleLineComment {
-                                prefix: "--".to_owned(),
-                                comment,
-                            })))
+                            let mut is_comment = true;
+                            if self.dialect.requires_single_line_comment_whitespace() {
+                                is_comment = Some(' ') == chars.peekable.clone().nth(1);
```
vimko, line 1237: Usually, `--` is used independently to separate segments of SQL statements. For example:

```sql
BEGIN;
--
-- Add field field_1 to table1
--
ALTER TABLE `table1` ADD COLUMN `field_1` datetime(6) NULL;

--
-- Add field field_2 to table2
--
ALTER TABLE `table2` ADD COLUMN `field_2` datetime(6) NULL;

COMMIT;

```

In this case, `--` will not be recognized as a comment.

hansott: Yes, that's right!

It was still on my planning to create a follow-up PR... I was doing more research but then the PR got merged

alamb: A follow on PR would be most appreciated

Issue #1706, opened 2025-02-05
Failure while parsing GIN Index


LucaCappelletti94:
The parsing of the `CREATE INDEX` case, is currently incomplete. Here follows a couple of examples:

```sql
CREATE INDEX users_name_trgm_idx ON users USING gin (concat_users_name(first_name, last_name) gin_trgm_ops);
```

which raises:

```rust
CREATE INDEX projects_name_description_trgm_idx ON projects USING gin (concat_projects_name_description(name, description) gin_trgm_ops);: ParserError("Expected: ), found: gin_trgm_ops")
```

Or analogously:

```sql
CREATE INDEX sample_containers_barcode_trgm_idx ON sample_containers USING gin (barcode gin_trgm_ops);
```

which raises:

```rust
CREATE INDEX sample_containers_barcode_trgm_idx ON sample_containers USING gin (barcode gin_trgm_ops);: ParserError("Expected: ), found: gin_trgm_ops")
```

I will be attempting shortly to do a PR adding support for these use cases.

LucaCappelletti94:
Right now, an Index only supports columns as defined by a `OrderByExpr`, which I do not believe is general enough to also handle the aforementioned cases. Is this correct?

LucaCappelletti94:
Closing as solved in PR #1707

PR #1707, opened 2025-02-05
Extend support for INDEX parsing

LucaCappelletti94:
This pull request addresses issue #1706.

- [x] Added tests reproducing issue #1706 
- [x] Added enumerations and keywords to properly represent operator classes
- [x] Extended Index struct to be able to characterize indices such as those presented in the issue

Summary of changes:

- Added a new document [`operator_classes.rs`](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-3a5b1c35e141ef8a4949ca9a150e0870d31afe954dc4b7d5d64dd002c4f0faa9) containing the classes
- Added all of the [necessary keywords](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-4861c0e30ea971e804df22762d8304abfc9a199cf54153b291517f1b3df5295a) to characterize the parsing of the operator classes
- Extended enumeration of available Indices [`ddl.rs`](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-ea0460af06320c3ef6a3778d344251fd254ddd6f04169e29ddda9b77dba4d010)
- Introduced a new struct [`IndexColumn`](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-ea0460af06320c3ef6a3778d344251fd254ddd6f04169e29ddda9b77dba4d010)
- [Extended parser](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-ea0460af06320c3ef6a3778d344251fd254ddd6f04169e29ddda9b77dba4d010) to properly handle the coupled parsing of different Indices and associated Operator Classes
- Updated accordingly the [test suite](https://github.com/apache/datafusion-sqlparser-rs/pull/1707/files#diff-ea0460af06320c3ef6a3778d344251fd254ddd6f04169e29ddda9b77dba4d010)


src/ast/ddl.rs
```diff
@@ -1030,20 +1030,29 @@ impl fmt::Display for KeyOrIndexDisplay {
 /// [1]: https://dev.mysql.com/doc/refman/8.0/en/create-table.html
 /// [2]: https://dev.mysql.com/doc/refman/8.0/en/create-index.html
 /// [3]: https://www.postgresql.org/docs/14/sql-createindex.html
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
```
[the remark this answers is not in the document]
LucaCappelletti94: Postgres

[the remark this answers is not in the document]
LucaCappelletti94: Should I do anything regarding this comment?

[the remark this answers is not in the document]
alamb: Sorry -- I haven't had a chance to review this PR in detail. I was mostly trying to gather information about what it was doing / what it was adding so that when someone had time to do a review they didn't have to guess/research the why and usecase.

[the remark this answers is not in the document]
LucaCappelletti94: I tried to describe those points in the first PR post, though I am perfectly fine with extending it further - is there any part I should be documenting more, in particular?

src/ast/dml.rs
```diff
@@ -32,14 +32,42 @@ use sqlparser_derive::{Visit, VisitMut};
...
+}
+
+impl From<OrderByExpr> for IndexColumn {
+    fn from(column: OrderByExpr) -> Self {
+        Self {
+            column,
+            operator_class: None,
+        }
+    }
+}
```
iffyio, line 59: hmm I'm thinking having an `OrderByExpr` -> `IndexColumn` could be arbitrary and potentially cause issues/mismatches if either struct evolves going forward, it might be better to manually construct `IndexColumn` where this was used?

src/parser/mod.rs
```diff
@@ -273,6 +274,7 @@ enum ParserState {
     ConnectBy,
 }
 
+#[cfg_attr(feature = "std", derive(Debug))]
```
iffyio, line 277: can we remove the introduced Debug statements if they're unrelated to the PR?

src/parser/mod.rs
```diff
@@ -13337,6 +13376,42 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse an expression, optionally followed by ASC or DESC (used in ORDER BY)
+    pub fn parse_create_index_expr<OPS: OperatorClass>(
```
iffyio, line 13380:
```suggestion
    /// Parse an expression, optionally followed by ASC or DESC (used in ORDER BY)
    fn parse_create_index_expr<OPS: OperatorClass>(
```
The description for the function sounds incorrect?

src/parser/mod.rs
```diff
@@ -13337,6 +13376,42 @@ impl<'a> Parser<'a> {
...
+            None
+        };
+
+        let with_fill = if dialect_of!(self is ClickHouseDialect | GenericDialect)
+            && self.parse_keywords(&[Keyword::WITH, Keyword::FILL])
+        {
+            Some(self.parse_with_fill()?)
+        } else {
+            None
+        };
```
iffyio, line 13402: since this part is identical to parse_order_by_expr, can we reuse the same code for both? e.g. split out to its own function

src/parser/mod.rs
```diff
@@ -13337,6 +13376,42 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse an expression, optionally followed by ASC or DESC (used in ORDER BY)
+    pub fn parse_create_index_expr<OPS: OperatorClass>(
+        &mut self,
+    ) -> Result<IndexColumn, ParserError> {
+        let expr = self.parse_expr()?;
+
+        let operator_class: Option<OPS> = self.parse_one_of_keywords(OPS::KEYWORDS).map(Into::into);
```
iffyio, line 13385: this seems to be the only usage of the introduced `operator_classes.rs`? I'm thinking it would ideal if we can avoid modelling the postgres spec for it, would it be possible to represent the operator_class as a opaque string/identifier instead?

tests/sqlparser_postgres.rs
```diff
@@ -2477,6 +2477,257 @@ fn parse_create_anonymous_index() {
     }
 }
 
+#[test]
+fn parse_create_users_name_trgm_index() {
+    let sql = "CREATE INDEX users_name_trgm_idx ON users USING GIN (concat_users_name(first_name, last_name) gin_trgm_ops)";
```
iffyio, line 2482: If I understood the syntax, its possible to have multiple indexed columns, could we have a test covering that if so? Also it would be ideal if we included test coverage for the various index types and operator classes introduced (though for the latter I think if we switch to opaque identifiers that would not be necessary).

(Also note that it would probably be easier to add further tests without asserting the full AST, only relying on verified_stmt roundtrip)

src/parser/mod.rs
iffyio: I think its fine to allow downstream crates to validate that the operator class is compatible with the specified index, that expectation is already present when consuming an AST from the parser, see [syntax vs semantics section](https://github.com/apache/datafusion-sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics) for the parser.

I think in this case there is a lot of code introduced to power this detail such that its not worth the accompanying complexity.

Representing it opaquely also makes the parser forwards compatible with with postgres going forward. admittedly this is a minor pro but it could be desirable given that there's seems to be quite a few of classes

iffyio:
Marking this as draft in the meantime as its no longer awaiting review, @LucaCappelletti94 please feel free to undraft and ping when ready!

LucaCappelletti94:
At this time I am somewhat stuck as I would not know how to parse the class operator names if not as a keyword - do you have any examples that could be applied to this use case?

iffyio:
> At this time I am somewhat stuck as I would not know how to parse the class operator names if not as a keyword - do you have any examples that could be applied to this use case?


@LucaCappelletti94 (not sure I followed entirely) did you mean to parse and represent an operator class like `gist_oid_ops`? If so I imagine we could parse the next token and check that its an unquoted `Token::Word(_)` and represent the class in the AST as an opaque `String` for example

LucaCappelletti94:
While things may change in the future, to the best of my knowledge some indices can have operator classes and some cannot. Since we model the different indices, is it preferable if I continue to model this semantical difference between them? Or should I remove it?

I believe that different indices which accept different parametrizations are more similar to syntax than semantics. Do let me know what is your opinion on it.

iffyio:
Sorry im not sure I understood the question, could you clarify with a sql example if that's possible?

iffyio:
Ah I see thanks for clarifying!

1. I think introducing a `Custom` variant would be reasonable, it would be similar to the existing pattern of `DataType::Custom`, `BinaryOperator::Custom` etc that we have. The existing `Hash` and `BTree` I figure we can keep as is, represented explicitly, partily to avoid breaking changes more than necessary and also since those types are the most common index types across dialects
2. I don't think we need to check that, we can look to parse the operator class if one is present in the input.
3. I think we can keep `IndexType` as the enum in (1), then essentially operator class in (2) is an opaque `String`.

LucaCappelletti94:
I have tried to replace the parsing of keywords with the generic String Ident as proposed, but it leads to much more complex parsing.

Primarily, I don't understand how I could distinguish an operator class from a keyword, such as: `CREATE UNIQUE INDEX IF NOT EXISTS idx_name ON test USING BTREE (name gin_trgm_ops,age DESC)`, specifically how can you tell that `gin_trgm_ops` is to be parsed as an operator class and `DESC` should instead be a keyword? Should I try to parse it as a keyword, and it it fails I assume it is an operator class?

LucaCappelletti94:
I have managed to do it, but I really dislike the approach I used, which peeks whether the parser will next encounter one of the expected keywords and if not, assumes that the next token is an operator class. I much preferred the previous approach that distinctively treated operator classes as keywords which was much stricter, but at some point it boils down to personal opinion.

Let me know your opinion on it.

LucaCappelletti94:
I tried to do the merge using the GitHub tool but it was lagging immensely, I will now fix it offline.

src/ast/operator_classes.rs
```diff
@@ -0,0 +1,581 @@
...
+impl OperatorClass for BloomOperatorClass {
+    const KEYWORDS: &'static [Keyword] = &[Keyword::INT4_OPS, Keyword::TEXT_OPS];
+}
+
+impl From<Keyword> for BloomOperatorClass {
+    fn from(keyword: Keyword) -> Self {
+        match keyword {
+            Keyword::INT4_OPS => BloomOperatorClass::Int4,
+            Keyword::TEXT_OPS => BloomOperatorClass::Text,
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
LucaCappelletti94: Comment no longer applies as the Operator classes enumeration no longer exists.

src/ast/dml.rs
LucaCappelletti94: Removed.

src/parser/mod.rs
LucaCappelletti94: Removed.

src/parser/mod.rs
LucaCappelletti94: Resolved as discussed in the other comments of the issue.

src/parser/mod.rs
```diff
@@ -63,6 +63,7 @@ mod recursion {
 
     use super::ParserError;
 
+    #[derive(Debug)]
```
iffyio, line 66:
```suggestion
```

src/parser/mod.rs
```diff
@@ -263,7 +264,7 @@ impl ParserOptions {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
```
iffyio, line 267:
```suggestion
#[derive(Copy, Clone)]
```

src/parser/mod.rs
```diff
@@ -7629,24 +7643,34 @@ impl<'a> Parser<'a> {
...
 
-    /// Parse [USING {BTREE | HASH}]
+    /// Parse [USING {BTREE | HASH | GIN | GIST | SPGIST | BRIN | BLOOM | identifier}]
     pub fn parse_optional_using_then_index_type(
         &mut self,
     ) -> Result<Option<IndexType>, ParserError> {
-        if self.parse_keyword(Keyword::USING) {
-            Ok(Some(self.parse_index_type()?))
+        Ok(if self.parse_keyword(Keyword::USING) {
+            Some(self.parse_index_type()?)
```
iffyio, line 7670:
```suggestion
       if self.parse_keyword(Keyword::USING) {
            Ok(Some(self.parse_index_type()?))
```
Can we change this and `parse_index_type` to use the previous style of returning individual ok instead of wrapping the if statement within the OK? I think that style is more consistent with the codebase

src/parser/mod.rs
```diff
@@ -7629,24 +7643,34 @@ impl<'a> Parser<'a> {
...
+            IndexType::BRIN
+        } else if self.parse_keyword(Keyword::BLOOM) {
+            IndexType::Bloom
+        } else {
+            IndexType::Custom(self.parse_identifier()?)
+        })
     }
 
-    /// Parse [USING {BTREE | HASH}]
+    /// Parse [USING {BTREE | HASH | GIN | GIST | SPGIST | BRIN | BLOOM | identifier}]
```
iffyio, line 7665:
```suggestion
    /// Optionally parse the `USING` keyword, followed by an [IndexType]
    /// Example:
    /// ```sql
    //// USING BTREE (name, age DESC)
    /// ```
```
Can we say something like this? in order to avoid duplicating the list of supported index types

src/parser/mod.rs
```diff
@@ -13595,10 +13619,33 @@ impl<'a> Parser<'a> {
         }
     }
 
-    /// Parse an expression, optionally followed by ASC or DESC (used in ORDER BY)
+    /// Parse an OrderByExpr expression, optionally followed by ASC or DESC (used in ORDER BY)
```
iffyio, line 13622:
```suggestion
    /// Parse an [OrderByExpr] expression.
```
Since the ASC/DESC is part of the returned `OrderByExpr`

src/parser/mod.rs
```diff
@@ -13595,10 +13619,33 @@ impl<'a> Parser<'a> {
...
     pub fn parse_order_by_expr(&mut self) -> Result<OrderByExpr, ParserError> {
+        self.parse_create_index_expr::<false>()
+            .map(|index_column| index_column.column)
+    }
+
+    /// Parse an IndexColumn expression (used in CREATE INDEX)
```
iffyio, line 13628:
```suggestion
    /// Parse an [IndexColumn].
```

src/parser/mod.rs
```diff
@@ -13595,10 +13619,33 @@ impl<'a> Parser<'a> {
...
+    pub fn parse_create_index_expr<const PARSE_OPERATOR_CLASS: bool>(
```
iffyio, line 13629: Can we move the replace the generic parameter with a normal function parameter? That would be consistent with the codebase. Also it would be good to document the parameter since this is being exposed as a public function, maybe with a SQL example to clarify?

hmm actually I'm thinking this param shouldn't actually be required on this public api since an operator class should always be applicable when parsing a create index expr, can we do something like this to clarify the relationships

```rust
pub fn parse_order_by_expr(&mut self) -> Result<OrderByExpr, ParserError> {
    self.parse_index_expr_inner(false)
        .map(|(index_expr,)| index_expr)
}
pub fn parse_create_index_expr() {
    self.parse_index_expr_inner(false)
            .map(|(column, operator_class)| IndexColumn { column operator_class })
}
fn parse_order_by_expr_inner(&self, with_operator_class: bool) -> Result<(IndexExpr, Option<Ident>)> {
    ...
}

File: tests/sqlparser_postgres.rs [hunk 16]
```diff
@@ -2509,6 +2509,340 @@ fn parse_create_anonymous_index() {
...
+            assert!(include.is_empty());
+            assert!(with.is_empty());
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn parse_create_projects_name_description_trgm_index() {
+    let sql = "CREATE INDEX projects_name_description_trgm_idx ON projects USING GIN (concat_projects_name_description(name, description) gin_trgm_ops)";
```
iffyio, line 2590: this test scenario looks identical to `parse_create_users_name_trgm_index` in terms of coverage?
Im thinking generally for the tests we can group them into the same test function, it would be sufficient with the AST assertion on only one of the test scenarios then for the rest we can rely only on `verified_stmt()`, that would keep the tests code smaller and easier to track what part of the syntax is covered.

Then could we add scenarios for all the introduced index types (i.e. BRIN, BLOOM etc)? As well as the custom type which probably would live in common I imagine?

src/parser/mod.rs

iffyio (approved):
cc @alamb

PR #1708, opened 2025-02-05
Fix incorrect parsing of JsonAccess bracket notation after cast in Snowflae

yoavcloud:
Some dialects support a size parameter for array type definitions, using bracket notation. For example: `INT[100]`. For other dialects, the bracket notation should be parsed as JsonAccess, for example in Snowflake this statement means to extract the second element after casting the JSONPath `b.c` from column a to an array: `SELECT a:b:c::ARRAY[1]`.

This PR makes a distinction between these dialects to ensure the correct parsing of JsonAccess after a cast expression.

### DIFF
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -80,4 +80,9 @@ impl Dialect for DuckDbDialect {
     fn supports_load_extension(&self) -> bool {
         true
     }
+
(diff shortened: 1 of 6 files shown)


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

Issue #1710, opened 2025-02-06
Implement ALTER SESSION


osipovartem:
https://docs.snowflake.com/en/sql-reference/sql/alter-session

PR #1712, opened 2025-02-06
Implement SnowFlake ALTER SESSION

osipovartem:
Closes https://github.com/apache/datafusion-sqlparser-rs/issues/1710

https://docs.snowflake.com/en/sql-reference/sql/alter-session


src/ast/mod.rs
```diff
@@ -2722,6 +2722,17 @@ pub enum Statement {
...
     /// ```sql
+    /// ALTER SESSION SET sessionParam
+    /// ALTER SESSION UNSET <param_name> [ , <param_name> , ... ]
+    /// ```
+    /// See <https://docs.snowflake.com/en/sql-reference/sql/alter-session>
+    AlterSession {
+        /// true is to set for the session parameters, false is to unset
+        set: bool,
+        /// The session parameters to set or unset
+        session_params: DataLoadingOptions,
```
alamb, line 2733: DataLoading options come from here

https://github.com/apache/datafusion-sqlparser-rs/blob/86abbd6028d72cee5fa299bc171f492fb722157c/src/ast/helpers/stmt_data_loading.rs#L50

alamb (commented):
Thanks @osipovartem  -- I started the CI checks on this PR (looks like there are some failures)

@yoavcloud since you are more familiar with Snowflake syntax, would you be able to review this PR as well?

tests/sqlparser_snowflake.rs
```diff
@@ -3404,3 +3404,29 @@ fn test_grant_database_role_to() {
     snowflake_and_generic().verified_stmt("GRANT DATABASE ROLE r1 TO ROLE r2");
     snowflake_and_generic().verified_stmt("GRANT DATABASE ROLE db1.sc1.r1 TO ROLE db1.sc1.r2");
 }
+
+#[test]
+fn test_alter_session() {
+    snowflake().verified_stmt("ALTER SESSION SET");
```
yoavcloud, line 3410: This is not a valid Snowflake statement, Snowflake doesn't allow an empty list of params.

tests/sqlparser_snowflake.rs
```diff
@@ -3404,3 +3404,29 @@ fn test_grant_database_role_to() {
...
+    snowflake().verified_stmt("ALTER SESSION UNSET");
```
yoavcloud, line 3411: This is not a valid Snowflake statement, Snowflake doesn't allow an empty list of params.

tests/sqlparser_snowflake.rs
```diff
@@ -3404,3 +3404,29 @@ fn test_grant_database_role_to() {
...
+    snowflake().verified_stmt("ALTER SESSION SET AUTOCOMMIT=TRUE");
+    snowflake().verified_stmt("ALTER SESSION SET AUTOCOMMIT=FALSE QUERY_TAG='tag'");
+    snowflake().verified_stmt("ALTER SESSION UNSET AUTOCOMMIT");
+    snowflake().verified_stmt("ALTER SESSION UNSET AUTOCOMMIT QUERY_TAG");
```
yoavcloud, line 3415: Snowflake requires a comma for unsetting multiple params

src/dialect/snowflake.rs
```diff
@@ -120,6 +120,12 @@ impl Dialect for SnowflakeDialect {
     }
 
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
+        if parser.parse_keywords(&[Keyword::ALTER, Keyword::SESSION]) {
+            // ALTER SESSION
+            let set = parser.parse_keyword(Keyword::SET) | !parser.parse_keyword(Keyword::UNSET);
```
yoavcloud, line 125: This would accept any statements that are not acceptable by Snowflake such as `ALTER SESSION XYZ`. Suggest to rewrite as:
```rust
            let set = match parser.parse_one_of_keywords(&[Keyword::SET, Keyword::UNSET]) {
                Some(Keyword::SET) => true,
                Some(Keyword::UNSET) => false,
                _ => return Some(parser.expected("SET or UNSET", parser.peek_token()))
            };
```

src/dialect/snowflake.rs
```diff
@@ -960,6 +980,42 @@ fn parse_stage_params(parser: &mut Parser) -> Result<StageParamsObject, ParserEr
...
+            },
+            _ => {
+                if parser.peek_token().token == Token::EOF {
+                    break
+                }
+                return parser.expected("another option", parser.peek_token())
+            },
+        }
+    }
+    Ok(options)
```
yoavcloud, line 1015: Perhaps add a check if for empty options and raise error

src/ast/mod.rs
yoavcloud: DataLoadingOptions is a very useful struct IMO, I would rename it to KeyValueOptions and promote it to its own helper file.

yoavcloud (changes requested):
See my comments inline. Generally speaking, I support parsing of arbitrary options using DataLoadingOptions but would rename and move this struct to a more central location in the codebase to reflect its new purpose.

alamb:
> See my comments inline. Generally speaking, I support parsing of arbitrary options using DataLoadingOptions but would rename and move this struct to a more central location in the codebase to reflect its new purpose.

Sounds good -- @osipovartem  -- what do you think?

yoavcloud (approved):
LGTM @iffyio @alamb!

iffyio (approved):
Thanks @osipovartem!

iffyio:
@osipovartem could you take a look into the CI failure?

alamb:
Marking as draft as we sort out the CI failure

PR #1718, opened 2025-02-08
BigQuery: Add support for `BEGIN`

iffyio:
Adds support for the `BEGIN ... EXCEPTION ... END` syntax in BigQuery:

```sql
BEGIN
  SELECT 1;
  SELECT 2;
EXCEPTION WHEN ERROR THEN
  SELECT 3;
  SELECT 4;
END
```

https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#beginexceptionend

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3072,6 +3072,28 @@ pub enum Statement {
         begin: bool,
         transaction: Option<BeginTransactionKind>,
         modifier: Option<TransactionModifier>,
+        /// List of statements belonging to the `BEGIN` block.
+        /// Example:
+        /// ```sql
+        /// BEGIN
+        ///     SELECT 1;
+        ///     SELECT 2;
+        /// END;
+        /// ```
+        statements: Vec<Statement>,
+        /// Statements of an exception clause.
+        /// Example:
+        /// ```sql
+        /// BEGIN
+        ///     SELECT 1;
+        /// EXCEPTION WHEN ERROR THEN
+        ///     SELECT 2;
+        ///     SELECT 3;
+        /// END;
+        /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#beginexceptionend>
+        exception_statements: Option<Vec<Statement>>,
+        /// TRUE if the statement has an `END` keyword.
+        has_end_keyword: bool,
     },
     /// ```sql
     /// SET TRANSACTION ...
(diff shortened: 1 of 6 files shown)


yoavcloud (approved):
LKTM @iffyio. I recently added support for something similar in MS-SQL, but only for the case where there's a semi-colon after the BEGIN/END TRY/CATCH keywords. I think your approach of `Vec<Statement>` is better, and I'll align to that.

src/ast/mod.rs
```diff
@@ -3058,6 +3058,33 @@ pub enum Statement {
...
+        /// ```sql
+        /// BEGIN
+        ///     SELECT 1;
+        ///     SELECT 2;
+        /// EXCEPTION WHEN ERROR THEN
+        ///     SELECT 3;
+        /// END;
+        /// ```
+        /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#beginexceptionend>
+        has_exception_when_clause: bool,
```
alamb, line 3082: I wondered if this field redundant with when `exception_statements` is empty? However, given the test cases below, it seems it is possible to have a `EXCEPTION WHEN ERROR THEN END`  (without any statements) so distinguishing is necessary

BTW another more Rusty way might be to represent this as 
```rust
        exception_statements: Option<Vec<Statement>>,
```

I don't feel strongly about this however

alamb (approved):
Thanks @iffyio and @yoavcloud   -- looks like a nice improvement to me!

src/ast/mod.rs
iffyio: Yeah I think that makes sense, I initially figured to use an explicit flag in case it was possible to have exception statements with a different syntax but probably no need to do that unless necessary, will update to use an option!

alamb: I think using the somewhat akward `Option<Vec<..>>` would allow the different syntax.

I can help if that would be good

iffyio: Done! updated to use an `Option`

src/ast/mod.rs
```diff
@@ -4831,6 +4856,24 @@ impl fmt::Display for Statement {
                 if !modes.is_empty() {
                     write!(f, " {}", display_comma_separated(modes))?;
                 }
+                if !statements.is_empty() {
+                    write!(f, " {}", display_separated(statements, "; "))?;
+                    // We manually insert semicolon for the last statement,
+                    // since display_separated doesn't handle that case.
+                    write!(f, ";")?;
+                }
+                if let Some(exception_statements) = exception_statements {
```
alamb, line 4865: ❤️

PR #1724, opened 2025-02-12
Add support for `ORDER BY ALL`

PokIsemaine:
Add support for `ORDER BY ALL`
```
SELECT *
FROM addresses
ORDER BY ALL;
```
https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all-examples
https://clickhouse.com/docs/en/sql-reference/statements/select/order-by


src/ast/query.rs
```diff
@@ -2205,31 +2205,50 @@ pub enum JoinConstraint {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub struct OrderBy {
-    pub exprs: Vec<OrderByExpr>,
-    /// Optional: `INTERPOLATE`
-    /// Supported by [ClickHouse syntax]
+pub enum OrderBy {
```
iffyio, line 2208: Would it make sense to use this representation instead?
```rust
enum OrderByKind {
    Expr(Vec<OrderByExpr>),
    All
}
struct OrderBy {
    pub kind: OrderByKind,
    pub interpolate: Option<Interpolate>,
}
```
thinking in order to avoid duplicating interpolate (and other shared) parameters

src/ast/query.rs
```diff
@@ -2205,31 +2205,50 @@ pub enum JoinConstraint {
...
+    }
+    pub fn get_interpolate(&self) -> Option<&Interpolate> {
+        match self {
+            OrderBy::Expressions(exprs_with_interpolate) => {
+                exprs_with_interpolate.interpolate.as_ref()
+            }
+            OrderBy::All(_) => None,
+        }
+    }
+}
```
iffyio, line 2250: I'm thinking we would rather avoid these methods to avoid more api breakage than necessary when things change. Did we introduce them primarily for the tests? if so I think we could as usual have the test match the expected expr variant

src/parser/mod.rs
```diff
@@ -9183,17 +9183,19 @@ impl<'a> Parser<'a> {
 
     pub fn parse_optional_order_by(&mut self) -> Result<Option<OrderBy>, ParserError> {
         if self.parse_keywords(&[Keyword::ORDER, Keyword::BY]) {
-            let order_by_exprs = self.parse_comma_separated(Parser::parse_order_by_expr)?;
-            let interpolate = if dialect_of!(self is ClickHouseDialect | GenericDialect) {
-                self.parse_interpolations()?
+            let order_by = if self.parse_keyword(Keyword::ALL) {
```
iffyio, line 9186: could we introduce a dialect method `self.dialect.supports_order_by_all()` and guard this behavior with it? since it would be incorrect behavior if this branch returns true for a dialect that does not support this feature

src/ast/query.rs
PokIsemaine: This is worth considering, `ORDER BY ALL` supports `ASC`, `DSC`, but `FILL` and `INTERPOLATE` are not supported in `ClickHouse`.
[ORDER BY ALL WITH FILL playground](https://sql.clickhouse.com?query=U0VMRUNUIG51bWJlciBhcyBrZXksIDUgKiBudW1iZXIgdmFsdWUsICdvcmlnaW5hbCcgQVMgc291cmNlCkZST00gbnVtYmVycygxNikgV0hFUkUga2V5ICUgNSA9PSAwCk9SREVSIEJZIEFMTCBXSVRIIEZJTEw7&chart=eyJ0eXBlIjoibGluZSIsImNvbmZpZyI6eyJ4YXhpcyI6InllYXIiLCJ5YXhpcyI6Ijk5dGhfYXZnX3RlbXAiLCJzZXJpZXMiOiJjb3VudHJ5IiwidGl0bGUiOiJUZW1wZXJhdHVyZSBieSBjb3VudHJ5IGFuZCB5ZWFyIn19&)
[ORDER BY ALL WITH INTERPOLATE playground](https://sql.clickhouse.com?query=U0VMRUNUIG4sIHNvdXJjZSwgaW50ZXIgRlJPTSAoCiAgIFNFTEVDVCB0b0Zsb2F0MzIobnVtYmVyICUgMTApIEFTIG4sICdvcmlnaW5hbCcgQVMgc291cmNlLCBudW1iZXIgYXMgaW50ZXIKICAgRlJPTSBudW1iZXJzKDEwKSBXSEVSRSBudW1iZXIgJSAzID0gMQopIE9SREVSIEJZIEFMTCBXSVRIIEZJTEwgRlJPTSAwIFRPIDUuNTEgU1RFUCAwLjUgSU5URVJQT0xBVEUgKGludGVyIEFTIGludGVyICsgMSk7&chart=eyJ0eXBlIjoibGluZSIsImNvbmZpZyI6eyJ4YXhpcyI6InllYXIiLCJ5YXhpcyI6Ijk5dGhfYXZnX3RlbXAiLCJzZXJpZXMiOiJjb3VudHJ5IiwidGl0bGUiOiJUZW1wZXJhdHVyZSBieSBjb3VudHJ5IGFuZCB5ZWFyIn19&)

src/ast/query.rs
PokIsemaine: Yes, this is mainly for testing, thank you for your suggestion.

src/parser/mod.rs
PokIsemaine: Thank you for your detailed reminder. Since I am not very familiar with this project, I overlooked this point.

src/ast/query.rs
iffyio: I think that would be fine, in the `ALL` case where they aren't provided then `interpolate` could be None I imagine

src/ast/query.rs
```diff
@@ -2339,6 +2364,33 @@ impl fmt::Display for InterpolateExpr {
...
+            Some(false) => write!(f, " DESC")?,
+            None => (),
+        }
+        match self.nulls_first {
+            Some(true) => write!(f, " NULLS FIRST")?,
+            Some(false) => write!(f, " NULLS LAST")?,
+            None => (),
+        }
+        Ok(())
+    }
```
iffyio, line 2391: I think it would be ideal to reuse the same code as orderbyexpr, maybe we can rename it to something generic like `OrderByOptions` and then replace both `asc` and `nulls_first` on `OrderByExpr` with a field containing this struct?

src/dialect/duckdb.rs
```diff
@@ -89,4 +89,9 @@ impl Dialect for DuckDbDialect {
     fn supports_from_first_select(&self) -> bool {
         true
     }
+
+    // See DuckDB <https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all-examples>
```
iffyio, line 93:
```suggestion
    /// See DuckDB <https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all-examples>
```

src/dialect/clickhouse.rs
```diff
@@ -80,6 +80,11 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    // See <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by>
```
iffyio, line 83:
```suggestion
    /// See <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by>
```

src/dialect/mod.rs
```diff
@@ -918,6 +918,16 @@ pub trait Dialect: Debug + Any {
     fn supports_array_typedef_size(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports `ORDER BY ALL`.
+    /// `ALL` which means all columns of the SELECT clause.
+    ///
+    /// DuckDB: <https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all>
+    /// ClickHouse: <https://clickhouse.com/docs/en/sql-reference/statements/select/order-by>
```
iffyio, line 926:
```suggestion
```
might not need to list the dialects since they're already listed in the impls. That way this list doesnt go out of sync if future dialects are added

src/parser/mod.rs
```diff
@@ -9191,17 +9191,31 @@ impl<'a> Parser<'a> {
...
+                if !self.dialect.supports_order_by_all() {
+                    return parser_err!(
+                        "ALL is not supported in ORDER BY",
+                        self.peek_token().span.start
+                    );
+                }
```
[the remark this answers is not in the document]
PokIsemaine: If dialect supports `ORDER BY ALL`, what kind of behavior do you think is appropriate? Should it be like before, parser_err, or parsed as a special OrderBy?

src/parser/mod.rs
```diff
@@ -13405,6 +13419,19 @@ impl<'a> Parser<'a> {
         })
     }
 
+    pub fn parse_order_by_all(&mut self) -> Result<OrderByAll, ParserError> {
```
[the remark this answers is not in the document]
PokIsemaine: You are right, this change is better. I plan to first change `pub fn parse_order_by_all` to `fn parse_order_by_options` and then reuse it in `parse_order_by_expr`.

src/dialect/duckdb.rs
PokIsemaine: I am not very familiar with Rust yet, and I want to know if there are any tools that can automatically unify the comment format of a project, because I saw both `//` and `///` in the same file, which is a bit confusing. Or does this project have any relevant specification documents?

src/parser/mod.rs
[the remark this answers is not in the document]
PokIsemaine: If an unsupported dialect uses `ORDER BY ALL`, should we return `parser_err` or an `OrderByKind::Expressions` type `OrderBy` with its expr as `ALL`?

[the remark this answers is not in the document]
iffyio: Yeah the latter, if an unsupporting dialect sees the query `ORDER BY ALL` it implies that there's a column/variable in scope called `ALL` i.e all is an identifier so it'll indeed be correct to parse it as an expression

src/dialect/duckdb.rs
iffyio: You're right, we don't run any lint for this in the project unfortunattely so there are discrepancies in the codebase. Ideally especially for public functions, in order for the comment to show up in the documentation `///` is required, whereas `//` is a internal-only code comment

iffyio (approved):
LGTM! Thanks @PokIsemaine!
cc @alamb

iffyio:
@PokIsemaine could you look into the merge conflicts on the branch when you get the time?

PR #1725, opened 2025-02-13
Extend Visitor trait for Value type

tomershaniii:
Enhancing the AST Visitor so it is possible to visit Values

### DIFF
--- a/src/ast/visitor.rs
+++ b/src/ast/visitor.rs
@@ -17,7 +17,7 @@
 
 //! Recursive visitors for ast Nodes. See [`Visitor`] for more details.
 
-use crate::ast::{Expr, ObjectName, Query, Statement, TableFactor};
+use crate::ast::{Expr, ObjectName, Query, Statement, TableFactor, Value};
 use core::ops::ControlFlow;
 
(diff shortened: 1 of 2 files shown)


src/ast/visitor.rs
```diff
@@ -889,34 +909,74 @@ mod tests {
...
-            let actual: Vec<_> = actual.iter().map(|x| x.as_str()).collect();
+            let mut visitor = TestVisitor::default();
+            let _ = do_visit(sql, &mut visitor);
+            let actual: Vec<_> = visitor.visited.iter().map(|x| x.as_str()).collect();
             assert_eq!(actual, expected)
         }
     }
+}
 
-    struct QuickVisitor; // [`TestVisitor`] is too slow to iterate over thousands of nodes
```
iffyio, line 898: is it reasonable to reserve the overflow test case? it sounded like it was supposed to test ability to cover large number of entries

tomershaniii: @iffyio  thanks for the feedback, indeed it was overlooked

iffyio (approved):
LGTM! Thanks @tomershaniii!
cc @alamb

alamb (commented):
THanks

PR #1726, opened 2025-02-14
Differentiate LEFT JOIN from LEFT OUTER JOIN

mvzink:
And same for RIGHT JOIN.

Although these are generally syntactically equivalent, I believe it is preferable to preserve the original input when rendering.


iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

alamb:
Makes sense -- thanks @mvzink

PR #1727, opened 2025-02-16
Add support for Postgres `ALTER TYPE`

jvatic:
And expose `CREATE TYPE AS ENUM` through `Generic` dialect.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -640,6 +640,95 @@ impl fmt::Display for AlterIndexOperation {
     }
 }
 
+/// An `ALTER TYPE` statement (`Statement::AlterType`)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct AlterType {
+    pub name: ObjectName,
+    pub operation: AlterTypeOperation,
+}
+
+/// An [AlterType] operation
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum AlterTypeOperation {
+    Rename(AlterTypeRename),
+    AddValue(AlterTypeAddValue),
+    RenameValue(AlterTypeRenameValue),
+}
+
+/// See [AlterTypeOperation::Rename]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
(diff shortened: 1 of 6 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -5320,6 +5320,117 @@ fn parse_create_type_as_enum() {
...
+                if_not_exists: false,
+                value: Ident::with_quote('\'', "label5"),
+                position: AlterTypeOperationAddValuePosition::Default,
+            },
+        },
+    ]
+    .into_iter()
+    .enumerate()
+    .for_each(|(index, tc)| {
+        let statement = pg_and_generic().one_statement_parses_to(tc.sql, tc.canonical);
```
iffyio, line 5424:
```suggestion
        let statement = pg_and_generic().verified_stmt(tc.sql);
```
can we use verified_stmt instead?

src/dialect/mod.rs
```diff
@@ -490,6 +490,26 @@ pub trait Dialect: Debug + Any {
...
+
+    /// Return true if the dialect supports ALTER TYPE statements
+    ///
+    /// Example:
+    /// ```sql
+    /// ALTER TYPE some_enum ADD VALUE 'some_value';
+    /// ```
+    fn supports_alter_type(&self) -> bool {
+        false
+    }
```
iffyio, line 511: Thinking we could skip the dialect flags for this case and have the parse the statement unconditionally, if it does not conflict with existing statements?

src/ast/ddl.rs
```diff
@@ -640,6 +640,68 @@ impl fmt::Display for AlterIndexOperation {
...
+    RenameValue {
+        from: Ident,
+        to: Ident,
+    },
+}
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum AlterTypeOperationAddValuePosition {
```
iffyio, line 665:
```suggestion
pub enum AlterTypeAddValuePosition {
```
maybe an opportunity to shorten the name, since AddValue is already an operation

src/ast/mod.rs
```diff
@@ -2691,6 +2692,14 @@ pub enum Statement {
         with_options: Vec<SqlOption>,
     },
     /// ```sql
+    /// ALTER TYPE
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-altertype.html)
+    /// ```
+    AlterType {
+        name: ObjectName,
+        operation: AlterTypeOperation,
+    },
```
iffyio, line 2701: Can we replace the anonymous enum variants (this and `AlterTypeOperation` with structs? we're slowly [moving a way from using this representation](https://github.com/apache/datafusion-sqlparser-rs/issues/1204)

jvatic:
Thanks for the review @iffyio! :) Comments addressed, let me know if there's any more changes you'd like.

iffyio (approved):
LGTM! Thanks @jvatic!
cc @alamb

PR #1731, opened 2025-02-19
Treat COLLATE like any other column option

mvzink:
This allows preserving the orderings of column options.

We also no longer hardcode a list of dialects which accept a column collation, as it's part of the SQL standard and some dialects which do support it were missing from the list (e.g. MSSQL).


iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

Issue #302, opened 2021-03-24
MySQL Compatibility Issues: mysqldump


DerSaidin:
I ran into multiple MySQL compatibility issues when attempting to parse a dump produced by `sudo mysqldump dbname > dbname_backup.sql` (also see [mysqldump docs](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) or [mariadb docs](https://mariadb.com/kb/en/mysqldump/)).

Here is an example of `mysqldump` generated SQL (this database is from a [WordPress](https://wordpress.com/) installation):
```
-- MySQL dump 10.16  Distrib 10.1.26-MariaDB, for debian-linux-gnu (x86_64)

--
-- Table structure for table `wp_users`
--

DROP TABLE IF EXISTS `wp_users`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!40101 SET character_set_client = utf8 */;
CREATE TABLE `wp_users` (
  `ID` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `user_login` varchar(60) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_pass` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_nicename` varchar(50) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_email` varchar(100) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_url` varchar(100) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_registered` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
  `user_activation_key` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `user_status` int(11) NOT NULL DEFAULT '0',
  `display_name` varchar(250) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  PRIMARY KEY (`ID`),
  KEY `user_login_key` (`user_login`),
  KEY `user_nicename` (`user_nicename`),
  KEY `user_email` (`user_email`)
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
/*!40101 SET character_set_client = @saved_cs_client */;

--
-- Dumping data for table `wp_users`
--

[...]

PR #1732, opened 2025-02-20
Add support column prefix index for MySQL

zzzdong:
This pull request add support for column prefixe and functional indexes in MySQL's `CREATE TABLE` and `ALTER TABLE` statements.
```sql
 ALTER TABLE `tbl_1` ADD KEY `idx_1` (col_1(10) DESC, (LOWER(col_2)));
```
May fix parts of #302  .

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1049,16 +1049,6 @@ pub enum Expr {
     /// [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-lambda-functions.html)
     /// [DuckDb](https://duckdb.org/docs/sql/functions/lambda.html)
     Lambda(LambdaFunction),
-    /// A ColumnPrefix used in MySQL indexes.
-    /// ```sql
-    /// CREATE INDEX ON tbl (col(10));
-    /// ```
-    ///
-    /// [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-index.html)
-    ColumnPrefix {
-        column: Ident,
-        length: u64,
-    },
 }
 
 /// The contents inside the `[` and `]` in a subscript expression.
@@ -1827,7 +1817,6 @@ impl fmt::Display for Expr {
             }
             Expr::Prior(expr) => write!(f, "PRIOR {expr}"),
             Expr::Lambda(lambda) => write!(f, "{lambda}"),
-            Expr::ColumnPrefix { column, length } => write!(f, "{column}({length})"),
         }
     }
 }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1049,6 +1049,16 @@ pub enum Expr {
     /// [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-lambda-functions.html)
     /// [DuckDb](https://duckdb.org/docs/sql/functions/lambda.html)
     Lambda(LambdaFunction),
+    /// A ColumnPrefix used in MySQL indexes.
+    /// ```sql
+    /// CREATE INDEX ON tbl (col(10));
+    /// ```
+    ///
+    /// [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-index.html)
+    ColumnPrefix {
+        column: Ident,
+        length: u64,
+    },
 }
 
 /// The contents inside the `[` and `]` in a subscript expression.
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -8591,6 +8591,61 @@ pub enum CopyIntoSnowflakeKind {
...
+    /// Mysql specific syntax
+    ///
+    /// See [Mysql documentation](https://dev.mysql.com/doc/refman/8.3/en/create-index.html)
+    /// for more details.
+    ColumnPrefix {
+        column: Ident,
+        length: u64,
+    },
+    Functional(Box<Expr>),
+}
```
iffyio, line 8637: would it be possible to replace this with an `Expr` instead? The docs seems to suggest that arbitrary expressions are allowed, from the docs
```sql
CREATE TABLE t1 (col1 INT, col2 INT, INDEX func_index ((ABS(col1))));
CREATE INDEX idx1 ON t1 ((col1 + col2));
CREATE INDEX idx2 ON t1 ((col1 + col2), (col1 - col2), col1);
ALTER TABLE t1 ADD INDEX ((col1 * 40) DESC);
```
 this would let us reuse existing code and representation, maybe even replace `IndexField` with `OrderByExpr`

zzzdong: I agree that using OrderByExpr could be a good approach. However, I am somewhat concerned that users might find it challenging to distinguish between the different `Expr` when working with it.

iffyio: The different Expr I imagine are encoded by the Expr variants, I think it would be preferrable to using a custom repr here which looks to require duplicating a lot of Expr in order to be comprehensive

zzzdong: That's right. I believe adding only ColumnPrefix will suffice. We can extend the Expr enum as follows:

```rust
enum Expr {
    ColumnPrefix { column: Ident, length: u64 }
}
```

iffyio: Hmm it looks like `ColumnPrefix` is equivalent to a function call with one argument? that can already be expressed by an `Expr` so that it doesn't seem like that should be necessary?

src/parser/mod.rs
```diff
@@ -7646,6 +7646,39 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_index_exprs(&mut self) -> Result<Vec<IndexExpr>, ParserError> {
+        self.parse_parenthesized(|p| p.parse_comma_separated(Parser::parse_index_expr))
+    }
+
+    pub fn parse_index_expr(&mut self) -> Result<IndexExpr, ParserError> {
```
iffyio, line 7653: Since this is flagged as a public function can we add a short description mentioning what it does?

src/parser/mod.rs
```diff
@@ -7646,6 +7646,39 @@ impl<'a> Parser<'a> {
...
+        };
+
+        let (operator_class, order_options) = if self.peek_keyword(Keyword::ASC)
+            || self.peek_keyword(Keyword::DESC)
+            || self.peek_keyword(Keyword::NULLS)
+        {
+            let order_options = self.parse_order_by_options()?;
+            (None, order_options)
+        } else {
+            let operator_class = self.maybe_parse(|p| p.parse_expr())?;
```
iffyio, line 7668: hmm I couldn't see `operator_class` being used in the introduced tests or documented so its unclear what the syntax is?

src/parser/mod.rs
```diff
@@ -7646,6 +7646,39 @@ impl<'a> Parser<'a> {
...
+        let expr = self.parse_expr()?;
+        let collation = if self.parse_keyword(Keyword::COLLATE) {
+            Some(self.parse_object_name(false)?)
+        } else {
+            None
+        };
+
+        let (operator_class, order_options) = if self.peek_keyword(Keyword::ASC)
+            || self.peek_keyword(Keyword::DESC)
+            || self.peek_keyword(Keyword::NULLS)
```
iffyio, line 7663: can we introduce a test case that covers the `ASC/DESC` and `nulls_first` syntax?

iffyio, line 7663: Instead of using `peek`, we can probably use `self.maybe_parse()` so that we don't need to maintain this list of keywords in sync with the `parse_order_by_options` function e.g.
```rust
let (operator_class, order_options) = if let Some(order_options ) = self.maybe_parse(|p| p.parse_order_by_options())? {
  (None, order_options)
} else {
  let operator_class = self.maybe_parse(|p| p.parse_expr())?;
  let order_options = self.parse_order_by_options()?;
  (operator_class, order_options)
};
```

src/parser/mod.rs
```diff
@@ -14688,6 +14721,8 @@ impl Word {
 
 #[cfg(test)]
 mod tests {
+    use std::vec;
+
```
iffyio, line 14725: hmm is this needed?

zzzdong:
Considering the work on the index parsing in #1707, this will be closed.

PR #1734, opened 2025-02-20
feat: adjust create and drop trigger for mysql dialect

invm:
Hi guys!

Mysql triggers are also a thing!

Cant parse `drop trigger a;` on mysql dialect, also should be able to parse create trigger for mysql dialect.

Appreciate all the hard work!

Let me know if I need to fix or cleanup anything.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4976,12 +4976,10 @@ impl<'a> Parser<'a> {
         }
         let if_exists = self.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
         let trigger_name = self.parse_object_name(false)?;
-        let table_name = match dialect_of!(self is PostgreSqlDialect | GenericDialect) {
-            true => {
-                self.expect_keyword_is(Keyword::ON)?;
-                Some(self.parse_object_name(false)?)
-            }
-            false => None,
+        let table_name = if self.parse_keyword(Keyword::ON) {
+            Some(self.parse_object_name(false)?)
+        } else {
+            None
         };
         let option = self
             .parse_one_of_keywords(&[Keyword::CASCADE, Keyword::RESTRICT])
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -5061,20 +5066,19 @@ impl<'a> Parser<'a> {
...
-                    .expect_keyword_is(Keyword::OF)
-                    .map(|_| TriggerPeriod::InsteadOf)?,
-                _ => unreachable!(),
-            },
-        )
+        let allowed_keywords = if dialect_of!(self is MySqlDialect) {
+            vec![Keyword::BEFORE, Keyword::AFTER]
+        } else {
+            vec![Keyword::BEFORE, Keyword::AFTER, Keyword::INSTEAD]
+        };
```
iffyio, line 5073: maybe it might not be necessary to single out mysql here, was the previous behavior conficting with the mysql syntax?

src/parser/mod.rs
```diff
@@ -4970,14 +4970,19 @@ impl<'a> Parser<'a> {
...
         let trigger_name = self.parse_object_name(false)?;
-        self.expect_keyword_is(Keyword::ON)?;
-        let table_name = self.parse_object_name(false)?;
+        let table_name = match dialect_of!(self is PostgreSqlDialect | GenericDialect) {
+            true => {
+                self.expect_keyword_is(Keyword::ON)?;
+                Some(self.parse_object_name(false)?)
+            }
+            false => None,
+        };
```
iffyio, line 4985: similar here wondering if the syntax doesn't conflict with mysql then it might be reasonable to parse this unconditionally, (otherwise its a bit more of a hassle to keep all the dialect checks in sync with control flow)

invm: Hiya!

Yes, it does conflict, thats why the `ON` keyword is skipped for mysql.

This is the definition from their docs: `DROP TRIGGER [IF EXISTS] [schema_name.]trigger_name`

src/parser/mod.rs
invm: Same here, the `INSTEAD` keyword is not a thing in mysql.

I'm happy to hear other cleaner suggestions, I just tried to modify as little code as possible

src/parser/mod.rs
iffyio: Ah yeah, I think what I meant was that since the table name is now optional that we can look to parse the `ON` suffix if it appears in the text, decoupled from the dialect. i.e.
```rust
let table_name = if self.parse_keyword(Keyword::ON) {
    Some(self.parse_object_name(false)?)
} else {
    None
}
```

src/parser/mod.rs
iffyio: I think cleanest would be to leave as the previous version which parses the `INSTEAD` keyword unconditionally, the parser doesn't strive to be exact in terms of dialect support in order to keep things tractable, rather downstream crates may further validate the syntax tree if necessary

invm: I understand what you mean, I reverted this part to the original implementation, just be aware that this parser will pass an invalid drop trigger statement for mysql, also remove this test:
```rust
let sql_drop_trigger_invalid = "DROP TRIGGER emp_stamp on emp;";
assert!(mysql()
    .parse_sql_statements(sql_drop_trigger_invalid)
    .is_err());
```

iffyio (approved):
LGTM! Thanks @invm!
cc @alamb

PR #1735, opened 2025-02-20
Ignore escaped LIKE wildcards in MySQL

mvzink:
MySQL has a special case for escaped LIKE wildcards appearing in string literals: the escaping is ignored, whereas normally for any other (non-special) character, the backslash would be stripped.

This is implemented with a new flag on dialect which gets passed into the tokenizer, because I don't know if any other dialects have similar behavior and wanted to make it easy to add them if they do. I can't currently test Snowflake, BigQuery, or Clickhouse on this point, so I'm just going off my best guess based on docs and examples.

[MySQL docs](https://dev.mysql.com/doc/refman/8.4/en/string-literals.html) (see "Table 11.1 Special Character Escape Sequences" and note following)

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -201,6 +201,33 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Determine whether the dialect strips the backslash when escaping LIKE wildcards (%, _).
+    ///
+    /// [MySQL] has a special case when escaping single quoted strings which leaves these unescaped
+    /// so they can be used in LIKE patterns without double-escaping (as is necessary in other
+    /// escaping dialects, such as [Snowflake]). Generally, special characters have escaping rules
+    /// causing them to be replaced with a different byte sequences (e.g. `'\0'` becoming the zero
+    /// byte), and the default if an escaped character does not have a specific escaping rule is to
+    /// strip the backslash (e.g. there is no rule for `h`, so `'\h' = 'h'`). MySQL's special case
+    /// for ignoring LIKE wildcard escapes is to *not* strip the backslash, so that `'\%' = '\\%'`.
+    /// This applies to all string literals though, not just those used in LIKE patterns.
+    ///
+    /// ```text
+    /// mysql> select '\_', hex('\\'), hex('_'), hex('\_');
+    /// +----+-----------+----------+-----------+
+    /// | \_ | hex('\\') | hex('_') | hex('\_') |
+    /// +----+-----------+----------+-----------+
+    /// | \_ | 5C        | 5F       | 5C5F      |
+    /// +----+-----------+----------+-----------+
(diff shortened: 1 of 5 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -2530,6 +2530,16 @@ fn parse_rlike_and_regexp() {
     }
 }
 
+#[test]
+fn parse_like_with_escape() {
+    mysql().verified_only_select(r#"SELECT * FROM customer WHERE name LIKE 'a\%c'"#);
```
iffyio, line 2535: I took a look at the comments and the tests but Im not sure I currently understood the behavior/problem being solved for.

Figured here to try ask with a concrete example, given this test case with the string `'a\%c'` what was the previous vs current behavior was it the case that we failed to parse the string in the previous behavior or that it produced a different kind of output?

iffyio: Ah yeah that makes sense, thanks for clarifying!

src/dialect/mod.rs
```diff
@@ -201,6 +201,33 @@ pub trait Dialect: Debug + Any {
...
+    /// | \_ | hex('\\') | hex('_') | hex('\_') |
+    /// +----+-----------+----------+-----------+
+    /// | \_ | 5C        | 5F       | 5C5F      |
+    /// +----+-----------+----------+-----------+
+    /// 1 row in set (0.00 sec)
+    /// ```
+    ///
+    /// [MySQL]: https://dev.mysql.com/doc/refman/8.4/en/string-literals.html
+    /// [Snowflake]: https://docs.snowflake.com/en/sql-reference/functions/like#usage-notes
+    fn ignores_like_wildcard_escapes(&self) -> bool {
```
iffyio, line 227:
```suggestion
    fn ignores_wildcard_escapes(&self) -> bool {
```
maybe we drop the `like` part? as the comment suggests if its nothing special about the `LIKE` syntax and more of a general string literal escape behavior

src/tokenizer.rs
```diff
@@ -807,6 +807,9 @@ pub struct Tokenizer<'a> {
     /// If true (the default), the tokenizer will un-escape literal
     /// SQL strings See [`Tokenizer::with_unescape`] for more details.
     unescape: bool,
+    /// If true, the tokenizer will not escape % and _, for use in in LIKE patterns. See
+    /// [`Dialect::ignores_like_wildcard_escapes`] for more details.
+    ignore_like_wildcard_escapes: bool,
```
iffyio, line 812: was it a reason to store this value here vs relying solely on the dialect via `self.dialect.ignores_like_wildcard_escapes()` when needed?

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1736, opened 2025-02-21
feat: use Edition 2024

iajoiner:
The only warning I got was
```bash
warning: relative drop order changing in Rust 2024
   --> src/test_utils.rs:305:11
    |
304 |     let mut iter = v.into_iter();
    |         --------
    |         |
    |         `iter` calls a custom destructor
    |         `iter` will be dropped later as of Edition 2024
305 |     match (iter.next(), iter.next()) { (Some(item), None) => {
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |           |
    |           this value will be stored in a temporary; let us call it `#1`
    |           up until Edition 2021 `#1` is dropped last but will be dropped earlier in Edition 2024
...
310 | }
    | - now the temporary value is dropped here, before the local variables in the block or statement
    |
    = warning: this changes meaning in Rust 2024
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-tail-expr-scope.html>
    = note: most of the time, changing drop order is harmless; inspect the `impl Drop`s for side effects like releasing locks or sending messages
    = note: `--force-warn tail-expr-drop-order` implied by `--force-warn rust-2024-compatibility`
```


iffyio (approved):
Thanks @iajoiner! This looks reasonable to me!
cc @alamb

iffyio:
@iajoiner could you take a look at the ci failure?

iajoiner:
@iffyio Yes!

Cargo.toml
```diff
@@ -30,7 +30,7 @@ include = [
     "Cargo.toml",
     "LICENSE.TXT",
 ]
-edition = "2021"
+edition = "2024"
```
alamb, line 33: my only concern is if this will force users of slparser to use newer versions of the rust compiler (as in once we release sqlparser with this change, will it require the use of rust 1.85?)

I haven't studied the implications yet

alamb (commented):
Thank you @iajoiner and @iffyio  -- the code changes that I skimmed quickly look good to me. As I mentioned my only real concern is the implication for required rust versions (but maybe there aren't any)

iffyio:
Oh good point, yeah it does look like MSRV is 1.85 for 2024 edition, did a quick test
```
Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.84.0 (66221abde 2024-11-19)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```
In which case I'm thinking this might not be a desirable change to make currently

alamb:
🤔  interestingly we don't seem to have an MSRV policy in this crate (at least not that I could find in https://github.com/apache/datafusion-sqlparser-rs/blob/main/README.md) 

We do have such a thing in DataFusion here: https://github.com/apache/datafusion?tab=readme-ov-file#rust-version-compatibility-policy

Maybe it is worth considering adding some documentation / policy in the sqlparser crate too 🤔

iffyio:
Adding an MSRV sounds good to me!

alamb:
> Adding an MSRV sounds good to me!

Filed an issue to track:
- https://github.com/apache/datafusion-sqlparser-rs/issues/1744

PR #1739, opened 2025-02-21
Parse SIGNED INTEGER type in MySQL CAST

### LINKED ISSUE #1589: MySQL datatype discrepancy between DDL and `CAST`
In MySQL, creating a column expects `integer unsigned` (as an "attribute" on the integer type, see [docs](https://dev.mysql.com/doc/refman/9.1/en/numeric-type-attributes.html)), while `CAST` expects `unsigned integer` or even just `unsigned` (see [docs](https://dev.mysql.com/doc/refman/9.1/en/cast-functions.html)).

This can be seen in this transcript:

```
mysql> create table foo (x unsigned integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'unsigned integer)' at line 1

mysql> create table foo (x integer unsigned);
Query OK, 0 rows affected (0.03 sec)

mysql> select cast(x as unsigned integer) from foo;
Empty set (0.01 sec)

mysql> select cast(x as integer unsigned) from foo;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'integer unsigned) from foo' at line 1
```

sqlparser just has one `parse_data_type` procedure used in both contexts, which currently parses according to the column datatype rules, which means it fails to parse `CAST` correctly.

Incorrectly failing to parse `unsigned integer`:

```
echo 'select cast(x as unsigned integer) from foo;' | cargo run --example cli - --mysql
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
2024-12-11T19:37:28.945Z DEBUG [sqlparser::parser] Parsing sql 'select cast(x as unsigned integer) from foo;
'...
2024-12-11T19:37:28.945Z DEBUG [sqlparser::parser] parsing expr
2024-12-11T19:37:28.945Z DEBUG [sqlparser::parser] parsing expr
2024-12-11T19:37:28.946Z DEBUG [sqlparser::parser] prefix: Identifier(Ident { value: "x", quote_style: None, span: Span(Location(1,13)..Location(1,14)) })
2024-12-11T19:37:28.946Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithSpan { token: Word(Word { value: "as", quote_style: None, keyword: AS }), span: Span(Location(1,15)..Location(1,17)) }
2024-12-11T19:37:28.946Z DEBUG [sqlparser::parser] next precedence: 0
2024-12-11T19:37:28.946Z DEBUG [sqlparser::parser] parsing expr
2024-12-11T19:37:28.946Z DEBUG [sqlparser::parser] prefix: Identifier(Ident { value: "x", quote_style: None, span: Span(Location(1,13)..Location(1,14)) })
2024-12-11T19:37:28.946Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithSpan { token: Word(Word { value: "as", quote_style: None, keyword: AS }), span: Span(Location(1,15)..Location(1,17)) }
2024-12-11T19:37:28.946Z DEBUG [sqlparser::parser] next precedence: 0
Error during parsing: ParserError("Expected: ), found: integer at Line: 1, Column: 27")
```

Incorrectly succeeding in parsing `integer unsigned`:

```
echo 'select cast(x as integer unsigned) from foo;' | cargo run --example cli - --mysql
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
2024-12-11T19:38:23.359Z DEBUG [sqlparser::parser] Parsing sql 'select cast(x as integer unsigned) from foo;
'...
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] parsing expr
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] parsing expr
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] prefix: Identifier(Ident { value: "x", quote_style: None, span: Span(Location(1,13)..Location(1,14)) })
2024-12-11T19:38:23.360Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithSpan { token: Word(Word { value: "as", quote_style: None, keyword: AS }), span: Span(Location(1,15)..Location(1,17)) }
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] next precedence: 0
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] prefix: Cast { kind: Cast, expr: Identifier(Ident { value: "x", quote_style: None, span: Span(Location(1,13)..Location(1,14)) }), data_type: UnsignedInteger(None), format: None }
2024-12-11T19:38:23.360Z DEBUG [sqlparser::dialect] get_next_precedence_full() TokenWithSpan { token: Word(Word { value: "from", quote_style: None, keyword: FROM }), span: Span(Location(1,36)..Location(1,40)) }
2024-12-11T19:38:23.360Z DEBUG [sqlparser::parser] next precedence: 0
Round-trip:
'SELECT CAST(x AS INTEGER UNSIGNED) FROM foo'
Parse results:
[
    Query(
        Query {
            with: None,
            body: Select(
                Select {
                    select_token: TokenWithSpan {
                        token: Word(
                            Word {
                                value: "select",
                                quote_style: None,
                                keyword: SELECT,
                            },
                        ),
                        span: Span(Location(1,1)..Location(1,7)),
                    },
                    distinct: None,
                    top: None,
                    top_before_distinct: false,
                    projection: [
                        UnnamedExpr(
                            Cast {
                                kind: Cast,
                                expr: Identifier(
                                    Ident {
                                        value: "x",
                                        quote_style: None,
                                        span: Span(Location(1,13)..Location(1,14)),
                                    },
                                ),
                                data_type: UnsignedInteger(
                                    None,
                                ),
                                format: None,
                            },
                        ),
                    ],
                    into: None,
                    from: [
                        TableWithJoins {
                            relation: Table {
                                name: ObjectName(
                                    [
                                        Ident {
                                            value: "foo",
                                            quote_style: None,
                                            span: Span(Location(1,41)..Location(1,44)),
                                        },
                                    ],
                                ),
                                alias: None,
                                args: None,
                                with_hints: [],
                                version: None,
                                with_ordinality: false,
                                partitions: [],
                                json_path: None,
                            },
                            joins: [],
                        },
                    ],
                    lateral_views: [],
                    prewhere: None,
                    selection: None,
                    group_by: Expressions(
                        [],
                        [],
                    ),
                    cluster_by: [],
                    distribute_by: [],
                    sort_by: [],
                    having: None,
                    named_window: [],
                    qualify: None,
                    window_before_qualify: false,
                    value_table_mode: None,
                    connect_by: None,
                },
            ),
            order_by: None,
            limit: None,
            limit_by: [],
            offset: None,
            fetch: None,
            locks: [],
            for_clause: None,
            settings: None,
            format_clause: None,
        },
    ),
]
```

I don't know if it would be better to have a second, somewhat redundant `parse_data_type_for_cast` which we enter for mysql-alikes, or add some special casing in the existing `parse_data_type`. That may depend on whether there are other discrepancies between create and cast that need to be addressed. Not sure when I'll have time to address this, but any feedback on approach would be welcome.

mvzink:
MySQL doesn't have the same set of possible CAST types as for e.g. column definitions. For example, it raises a syntax error for `CAST(1 AS INTEGER SIGNED)` and instead expects `CAST(1 AS SIGNED INTEGER)`.

We retain the current somewhat permissive datatype parsing behavior (e.g. allowing `CAST(1 AS BIGINT)` even though MySQL would raise a syntax error), and add datatypes for this specific case (`SIGNED [INTEGER]` and `UNSIGNED [INTEGER]`).

To keep the naming consistent across datatypes, we also rename `UnsignedInt` to `IntUnsigned` and similar for other unsigned datatypes. This means they display in the same way they are written, i.e. `IntUnsigned = INT UNSIGNED` instead of `UnsignedInt = INT UNSIGNED`.

Closes #1589

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -132,19 +132,19 @@ pub enum DataType {
     /// Tiny integer with optional display width e.g. TINYINT or TINYINT(3)
     TinyInt(Option<u64>),
     /// Unsigned tiny integer with optional display width e.g. TINYINT UNSIGNED or TINYINT(3) UNSIGNED
-    UnsignedTinyInt(Option<u64>),
+    TinyIntUnsigned(Option<u64>),
     /// Int2 as alias for SmallInt in [postgresql]
     /// Note: Int2 mean 2 bytes in postgres (not 2 bits)
     /// Int2 with optional display width e.g. INT2 or INT2(5)
     ///
     /// [postgresql]: https://www.postgresql.org/docs/15/datatype.html
     Int2(Option<u64>),
-    /// Unsigned Int2 with optional display width e.g. INT2 Unsigned or INT2(5) Unsigned
-    UnsignedInt2(Option<u64>),
+    /// Unsigned Int2 with optional display width e.g. INT2 UNSIGNED or INT2(5) UNSIGNED
+    Int2Unsigned(Option<u64>),
     /// Small integer with optional display width e.g. SMALLINT or SMALLINT(5)
     SmallInt(Option<u64>),
     /// Unsigned small integer with optional display width e.g. SMALLINT UNSIGNED or SMALLINT(5) UNSIGNED
-    UnsignedSmallInt(Option<u64>),
+    SmallIntUnsigned(Option<u64>),
     /// MySQL medium integer ([1]) with optional display width e.g. MEDIUMINT or MEDIUMINT(5)
     ///
     /// [1]: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
@@ -152,7 +152,7 @@ pub enum DataType {
     /// Unsigned medium integer ([1]) with optional display width e.g. MEDIUMINT UNSIGNED or MEDIUMINT(5) UNSIGNED
     ///
     /// [1]: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
-    UnsignedMediumInt(Option<u64>),
+    MediumIntUnsigned(Option<u64>),
     /// Int with optional display width e.g. INT or INT(11)
     Int(Option<u64>),
     /// Int4 as alias for Integer in [postgresql]
@@ -197,11 +197,11 @@ pub enum DataType {
     /// Integer with optional display width e.g. INTEGER or INTEGER(11)
     Integer(Option<u64>),
     /// Unsigned int with optional display width e.g. INT UNSIGNED or INT(11) UNSIGNED
-    UnsignedInt(Option<u64>),
+    IntUnsigned(Option<u64>),
     /// Unsigned int4 with optional display width e.g. INT4 UNSIGNED or INT4(11) UNSIGNED
-    UnsignedInt4(Option<u64>),
+    Int4Unsigned(Option<u64>),
     /// Unsigned integer with optional display width e.g. INTEGER UNSIGNED or INTEGER(11) UNSIGNED
-    UnsignedInteger(Option<u64>),
+    IntegerUnsigned(Option<u64>),
     /// Unsigned integer type in [clickhouse]
     /// Note: UInt8 mean 8 bits in [clickhouse]
     ///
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -798,9 +798,15 @@ pub enum Expr {
         kind: CastKind,
         expr: Box<Expr>,
         data_type: DataType,
-        // Optional CAST(string_expression AS type FORMAT format_string_expression) as used by BigQuery
-        // https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
+        /// Optional CAST(string_expression AS type FORMAT format_string_expression) as used by [BigQuery]
+        ///
+        /// [BigQuery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#formatting_syntax
         format: Option<CastFormat>,
+        /// Whether this was parsed as a [MySQL]-style cast, which has a different syntax for
```
iffyio, line 805: would it be possible to instead add the `DataType` variants that mysql uses? It seems like one issue is that the existing signed/unsigned datatypes aren't matching their expected string output. for example [Datatype::UnsignedInteger](https://github.com/apache/datafusion-sqlparser-rs/blob/339239d0c54fd0fe12e05d8de868340d07030ffa/src/ast/data_type.rs#L203-L204) outputs itself as `INTEGER UNSIGNED`.

Thinking expectation behaviour could be that we introduce
`DataType::IntegerUnsigned` => `INTEGER UNSIGNED`
and update
`DataType::UnsignedInteger` => `UNSIGNED INTEGER`

mvzink: That's a much better idea. Thanks!

src/ast/data_type.rs
```diff
@@ -238,6 +238,26 @@ pub enum DataType {
...
+    /// Signed integer as used in [MySQL CAST] target types, with optional `INTEGER` suffix:
+    /// `SIGNED [INTEGER]`
+    ///
+    /// Note that this doesn't accept a display width and is reversed from the syntax used in column
+    /// definitions ([`DataType::Int`]): `INTEGER [SIGNED]`
+    ///
+    /// Semantically equivalent to `BIGINT`.
+    ///
+    /// [MySQL CAST]: https://dev.mysql.com/doc/refman/8.4/en/cast-functions.html
+    Signed(bool),
```
iffyio, line 250: thinking it feels a unexpected to use `Signed(true)` to represent `SIGNED INTEGER`? and the repr assumes  a mysql-only type which might not be the case going forward. Feels like it'd be clear to represent them explicitly as `SIGNED` and `SIGNED INTEGER` where both can be reused by other dialects later if needed?

mvzink: I'm not sure exactly what you mean. Different variants `Signed = SIGNED` and `SignedInteger = SIGNED INTEGER`? The only reason I didn't do that is there's already `UnsignedInt = INT[EGER] UNSIGNED` and it would be even more confusing to have `UnsignedInteger = UNSIGNED INTEGER`, and very disruptive to change the existing `UnsignedInt` to `IntUnsigned` or something.

mvzink: Ah I see, that is what you are suggesting. I misread your previous comment, sorry! I'll try that out.

mvzink: Fixed. Let me know if this is what you had in mind. Definitely could be a bit disruptive to downstream users.

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1741, opened 2025-02-24
Add `CASE` and `IF` statement support

iffyio:
Add support for scripting statements
```sql
CASE product_id
  WHEN 1 THEN SELECT 1;
  WHEN 2 THEN SELECT 2;
  ELSE SELECT 3;
END CASE;
```

```sql
IF EXISTS(SELECT 1) THEN SELECT 1;
  ELSEIF EXISTS(SELECT 2) THEN SELECT 2;
  ELSE SELECT 3;
END IF;
```

[BigQuery CASE](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#case)
[BigQuery IF](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#if)
[Snowflake CASE](https://docs.snowflake.com/en/sql-reference/snowflake-scripting/case)
[Snowflake IF](https://docs.snowflake.com/en/sql-reference/snowflake-scripting/if)

### DIFF
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -22,20 +22,21 @@ use crate::tokenizer::Span;
 
 use super::{
     dcl::SecondaryRoles, value::ValueWithSpan, AccessExpr, AlterColumnOperation,
-    AlterIndexOperation, AlterTableOperation, Array, Assignment, AssignmentTarget, CloseCursor,
-    ClusteredIndex, ColumnDef, ColumnOption, ColumnOptionDef, ConflictTarget, ConnectBy,
-    ConstraintCharacteristics, CopySource, CreateIndex, CreateTable, CreateTableOptions, Cte,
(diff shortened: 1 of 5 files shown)


src/ast/spans.rs
```diff
@@ -732,6 +735,53 @@ impl Spanned for CreateIndex {
...
+impl Spanned for CaseStatement {
+    fn span(&self) -> Span {
+        let CaseStatement {
+            match_expr,
+            when_blocks,
+            else_block,
+            has_end_case: _,
+        } = self;
+
+        union_spans(
```
alamb, line 747: this is quite fancy 👌

alamb (approved):
Thank you @iffyio  -- I found this PR really easy to read and review as it was well documented and well tested ❤️

iffyio:
Thanks for the reviews @alamb!

PR #1742, opened 2025-02-24
Add support for `CREATE SCHEMA` options

iffyio:
```sql
CREATE SCHEMA DEFAULT COLLATE 'foo' myschema OPTIONS(key='value');
```

[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_schema_statement)

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -985,6 +985,7 @@ fn parse_create_schema_if_not_exists() {
         Statement::CreateSchema {
             if_not_exists: true,
             schema_name,
+            ..
         } => assert_eq!("schema_name", schema_name.to_string()),
         _ => unreachable!(),
     }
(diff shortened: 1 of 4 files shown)


alamb (approved):

tests/sqlparser_postgres.rs
```diff
@@ -988,6 +988,7 @@ fn parse_create_schema_if_not_exists() {
         Statement::CreateSchema {
             if_not_exists: true,
             schema_name,
+            ..
```
alamb, line 991: I personally think adding explicit ignores here

```suggestion
            options: _, 
            default_collate_spec: _,
```

Is better as then if a new field is added then the compiler will point out this place as needing to be checked

That being said, I don't think it is critical in this case

iffyio: Done!

alamb:
🚀

PR #1745, opened 2025-02-25
Parse MySQL ALTER TABLE ALGORITHM option

mvzink:
This may semantically be an option for how to perform the other operations in the statement, but syntactically MySQL treats it like an operation (e.g. can be in any order relative to other operations, can even be the only 'operation'), so we parse it as an operation.


iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1746, opened 2025-02-26
Align SQL formatting and add all missing table options

benrsatori:


mvzink:
If you are considering overhauling and fleshing out support for table options, I would be strongly in favor of making them work like column options: i.e. a `Vec` of options that can be parsed and rendered in any order. See `Parser::parse_optional_column_option` and its various usages. It would also be easier to have parity with `CREATE TABLE` and `ALTER TABLE` that way (both could just include a `Vec<TableOption>` or whatever).

PR #1747, opened 2025-02-26
Add all missing table options to be handled in any order

benrsatori:


mvzink:
Ah, I'm guessing this supersedes #1746. As mentioned in my [comment](https://github.com/apache/datafusion-sqlparser-rs/pull/1746#issuecomment-2686444775) there, I strongly believe we should implement table options in the same way as column options, with a `Vec` preserving the input order.

src/ast/dml.rs
```diff
@@ -138,6 +143,30 @@ pub struct CreateTable {
     pub engine: Option<TableEngine>,
     pub comment: Option<CommentDef>,
     pub auto_increment_offset: Option<u32>,
+    pub key_block_size: Option<u32>,
```
iffyio, line 146: there's quite a few of these options and its not clear what dialect they're coming from, I assume there will be more across the various dialects.

I'm thinking to @mvzink's point, it will be good to represent this as a `Vec`, also think that we can represent this similar to the existing `with_options: Vec<SqlOption>` on this struct (maybe we can just reuse and merge this into that field if it makes sense, otherwise it could be a new field)? essentially that it would be ideal to not have to enumerate and keep in sync all possible table options across dialects but instead always accept them as opaque entries

tests/sqlparser_duckdb.rs
```diff
@@ -721,6 +721,30 @@ fn test_duckdb_union_datatype() {
             comment: Default::default(),
             auto_increment_offset: Default::default(),
             default_charset: Default::default(),
+            insert_method: Default::default(),
```
tomershaniii, line 724: can we use ..Default::default() for all these members?

tests/sqlparser_mssql.rs
```diff
@@ -1590,6 +1590,30 @@ fn parse_create_table_with_valid_options() {
                 comment: None,
                 auto_increment_offset: None,
                 default_charset: None,
+                insert_method: None,
```
tomershaniii, line 1593: ..Default::default() ?

iffyio:
Marking the PR as draft in the meantime as it is not currently pending review, @benrsatori feel free to undraft and ping when ready!

src/ast/dml.rs
mvzink: Seems like a reasonable approach to me. Thanks for pushing this forward! I would suggest that whatever parsing routine/subparser produces that vector of options be written in such a way that it could be used for `ALTER TABLE` parsing as well; I believe their syntax is basically the same, at least for MySQL, and that struct should be updated as well.

iffyio: > Add another parameter Vec which will include properties which do not require a preceding keyword (WITH/OPTIONS/...), e.g. plain_options

seeing as there's a few similar parameters, thinking it could be nice to merge the variants into an enum something like
```rust
enum CreateTableOptions {
  TableProperties(Vec<SqlOption>),
  WithOptions(Vec<SqlOption>),
  Options(Vec<SqlOption>),
  Plain(Vec<SqlOption>),
}
```
that doesn't have to be in this PR however, we could introduce a new field if it could be a hassle to merge the existing ones.


> Parameter parsing will be delegated to the respective Dialect to avoid clutter in the main flow

How do we mean by this part? I imagined it would be reusing `self.parse_options`

tomershaniii: Not sure it makes sense to reuse `parse_options`, it's designed to parse something in the format of `KEYWORD(opt1 , opt2, opt3)` while mysql syntax is `opt1 opt2 opt3`.

tomershaniii: I like the idea of an enum, will handle that on a separate PR

tomershaniii:
@iffyio @mvzink appreciate your feedback on the updated PR.
BTW, please pull this PR out of WIP (I do not have the permissions to do so)

src/parser/mod.rs
```diff
@@ -6928,13 +6874,122 @@ impl<'a> Parser<'a> {
...
+
+        // Parse table options in any order
+        if let Some(keyword) = self.parse_one_of_keywords(&[
+            Keyword::ENGINE,
+            Keyword::AUTO_INCREMENT,
+            Keyword::DEFAULT,
+            Keyword::CHARSET,
+            Keyword::COLLATE,
+            Keyword::INSERT_METHOD,
+        ]) {
```
mvzink, line 6900: I'm not sure I understand the motivation for making this a dialect parsing method. Is there a problem with putting all the supported options across all dialects in here? It especially raises my eyebrow to see the two lists have so much overlap, including with (afaik) MySQL-specific options like `ENGINE` and `AUTO_INCREMENT` in the generic set. If we are going to have Postgres and MSSQL parse those, why not have them parse `DATA DIRECTORY` too?

src/ast/mod.rs
```diff
@@ -6925,6 +6928,12 @@ pub enum SqlOption {
         range_direction: Option<PartitionRangeDirection>,
         for_values: Vec<Expr>,
     },
+
+    TableSpace(TablespaceOption),
+
+    Union(Vec<Ident>),
+
+    TableEngine(TableEngine),
```
mvzink, line 6936: tbh I haven't checked where else `SqlOption` is used, but it really seems like it should be called `TableOption`, no?

src/ast/helpers/stmt_create_table.rs
```diff
@@ -78,25 +79,22 @@ pub struct CreateTableBuilder {
...
+    pub options: Option<Vec<SqlOption>>,
+    pub plain_options: Vec<SqlOption>,
     pub file_format: Option<FileFormat>,
     pub location: Option<String>,
     pub query: Option<Box<Query>>,
     pub without_rowid: bool,
     pub like: Option<ObjectName>,
     pub clone: Option<ObjectName>,
-    pub engine: Option<TableEngine>,
     pub comment: Option<CommentDef>,
```
mvzink, line 90: I wonder if `comment` should be folded in to options as well (with a new `SqlOption::Comment(CommentDef)` variant)?

src/parser/mod.rs
```diff
@@ -6928,13 +6874,122 @@ impl<'a> Parser<'a> {
...
+            let _ = self.consume_token(&Token::Eq);
+            let value = self.next_token();
+
+            match (keyword, value.token) {
+                (Keyword::AUTO_INCREMENT, Token::Number(n, l)) => {
```
mvzink, line 6905: I think this is missing the optional `=` parsing that was present in the old version (and I believe was correct).

src/ast/mod.rs
iffyio, line 6936: Could we add a description (ideally with syntax example) and links to the docs where these variants come from?

src/dialect/mysql.rs
```diff
@@ -141,6 +149,280 @@ impl Dialect for MySqlDialect {
...
+    /// This method is called to parse the next table/view option.
+    ///
+    /// If `None` is returned, falls back to the default behavior.
+    /// <https://dev.mysql.com/doc/refman/8.4/en/create-table.html>
+    fn parse_plain_option(&self, parser: &mut Parser) -> Result<Option<SqlOption>, ParserError> {
+        //Some consts
+        const COMPRESSION_OPTS: [&str; 3] = ["ZLIB", "LZ4", "NONE"];
+        const INSERT_METHODS_OPTS: [&str; 3] = ["NO", "FIRST", "LAST"];
+
+        let keyword = match parser.parse_one_of_keywords(&[
```
iffyio, line 163: to clarify my [comment here](https://github.com/apache/datafusion-sqlparser-rs/pull/1747#discussion_r1972915911) my thinking is that the options are plain key value pairs and idealy would be parsed as such. So that I imagined the parser shouldnt look for explicit keywords, similar to other dialects like bigquery it only looks to parse a key and a value.

Essentially, that it looks to parse and represent something like this I was thinking
```rust
 SqlOption::Plain { // Similar to SqlOption::KeyValue except it doesnt format with =
        key: Ident,
        value: Expr,
    }
```

src/parser/mod.rs
tomershaniii: Note the  `let _ = self.consume_token(&Token::Eq);`  a could of lines earlier or is it something else i've missed?

src/ast/helpers/stmt_create_table.rs
tomershaniii: Good idea, will do

src/parser/mod.rs
mvzink: Nope, I totally missed that!

src/dialect/mysql.rs
tomershaniii: I thought it would be better to parse more accurately where is is easy to implement (i.e. only a limited set of values are allowed). it does provide some value and I do not see the downside given that it's in dialect specific code.
LMKWYT

src/parser/mod.rs
tomershaniii: It's a bigger question if the library should branch out to dialect specific parsing or maintain common code supporting all dialects, looking at `SnowflakeDialect` is seems like a specific implementation exists (`parse_create_table`) and looking at the common code i can see several places where the code diverges based on dialect (mostly hive related).
 
Having `parse_create_table` as part of the dialect is what guided the decision to pull this code in, I may have overlooked `ENGINE` and `AUTO_INCREMENT`, prefer to pull those into mysql code rather than take mysql code out.

LMKWYT

src/ast/helpers/stmt_create_table.rs
tomershaniii: See the latest commit, 
note that the hive dialect still has some specific requirements

src/ast/dml.rs
tomershaniii: @iffyio added a commit with merging of all variants into the (pre-existing) CreateTableOptions enum

src/ast/mod.rs
tomershaniii: Seems to be used all over (` ColumnOption`, `CreateConnector`, `CreateSchema`, `Cache` & more)
looks like it's mainly used for the key/value variant.
IMO it makes sense to split out into a key/value type and a TableOption type (which has a K/V variant).
I think we should save this for another PR, there is a lot of scope in this one as is.

tomershaniii:
@iffyio @mvzink thanks for the feedback, did some additional work per your comments (& commented on other)
LMK if makes sense

src/parser/mod.rs
mvzink: You're right, that's a bigger question. I don't have a strong suggestion at the moment and this certainly seems servicable to me. My 2¢ on the bigger question is that introducing more dialect overrides (and actually this is more similar to `parse_column_option`, which is global, than Snowflake's `parse_create_table` which is only called from `SnowflakeDialect::parse_statement`, the normal top-level override) is a slippery slope toward fully dynamically dispatched parsing, but it's more justifiable if it's parsing something with a substantially different syntactic structure. However, this is *mostly* just parsing a different list of acceptable keywords but doing it in the same way. In such cases, I personally think more permissive generic parsing is preferable; e.g. does it really matter if `SnowflakeDialect` accepts `ROW_FORMAT = COMPACT` even though it doesn't make sense, so long as it doesn't introduce syntactic ambiguity? The exceptions of special parsing routines for a few options like `DATA DIRECTORY` and `TABLESPACE` are a good argument though. If anything, I would probably favor putting all the not-so-special options in the default implementation and only using the override for these options that require special parsing. Honestly, if it wouldn't introduce ambiguity (mostly due to options not being comma separated), I would probably even prefer the default/fallback routine for non-special options just be "parse ident, optionally parse equals sign, parse expr"—that's all thats's going to end up in the `SqlOption::KeyValue` at the end anyway, so why do we care which specific list of idents are acceptable? That is more of a semantic than syntactic question IMO.

src/dialect/mysql.rs
mvzink: I philosophically agree with @iffyio's ask here, as I mention in my other big ranty comment (sorry 😛). However, I am not 100% convinced in practice, mostly because these options aren't comma separated and this could produce some really confusing parses if users make errors. I imagined the default parsing routine would be "parse ident, optionally parse equals sign, parse expr". But suppose a user intended to send `ROW_FORMAT COMPACT DEFAULT CHARACTER SET latin1`, but accidentally deleted `COMPACT`. Instead of a parse error like they would get if options were comma separated or hardcoded, they get a successful parse with `ROW_FORMAT = DEFAULT` followed by `CHARACTER SET = latin1`. Is it our fault? Not really. Is it better than the maintenance burden of keeping a list of valid one-word option names up to date? I'm honestly not sure.

tomershaniii:
Hi again @iffyio @mvzink, see the latest commit, per your suggestions:
- Permissive parsing (removed value validation for key/value parameters)
- Centralized all plain parameters into a single parsing function
- Simplified the Key/Value parsing code (aligned @iffyio suggestion)

src/parser/mod.rs
```diff
@@ -7081,18 +7029,243 @@ impl<'a> Parser<'a> {
...
+            Ident::new("INDEX DIRECTORY")
+        } else if self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET]) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("CHARACTER SET")
+        } else if self.parse_keyword(Keyword::CHARSET) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("CHARSET")
+        } else if self.parse_keyword(Keyword::COLLATE) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("COLLATE")
```
mvzink, line 7182: These comments are wrong.

src/parser/mod.rs
```diff
@@ -7081,18 +7029,243 @@ impl<'a> Parser<'a> {
...
+            }
+        }
+
+        // Key/Value parameter option
+        let key = if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARSET]) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("DEFAULT CHARSET")
+        } else if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARACTER, Keyword::SET]) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("DEFAULT CHARACTER SET")
```
mvzink, line 7164: Not for this PR IMO, but I'd be curious to hear your thoughts & @iffyio's: would it be appropriate to add another `SqlOption` variant like `TableSpace` and `TableEngine` for a few of these options to make inspecting them easier for users? I'm specifically thinking of `[DEFAULT] {CHARSET | CHARACTER SET}` because as it is if I wanted to extract charset info from DDL, I would have to check for 4 different hardcoded strings which aren't represented in the public API (i.e. there's no `DEFAULT_CHARACTER_SET_IDENT` or a set of idents I could expect to see). Does the fact that it has 4 possible representations make it worth special casing with a `SqlOption::CharSet { default, shorthand, value }`? Since I will in fact need to inspect character set info, I may post a PR for that.

For this PR though, to at least make it harder to miss one of those 4 variants, can we put all the charset related branches next to each other?

mvzink (commented):
Thanks for keeping on with this!

src/parser/mod.rs
```diff
@@ -7081,18 +7029,243 @@ impl<'a> Parser<'a> {
...
+            // PASSWORD [=] 'string'
+            Ident::new("PASSWORD")
+        } else if self.parse_keyword(Keyword::SECONDARY_ENGINE_ATTRIBUTE) {
+            // SECONDARY_ENGINE_ATTRIBUTE [=] 'string'
+            Ident::new("SECONDARY_ENGINE_ATTRIBUTE")
+        } else if self.parse_keyword(Keyword::INSERT_METHOD) {
+            // INSERT_METHOD [=] { NO | FIRST | LAST }
+            Ident::new("INSERT_METHOD")
+        } else if self.parse_keyword(Keyword::AUTO_INCREMENT) {
+            Ident::new("AUTO_INCREMENT")
```
iffyio, line 7242: Ah so for this part I think these can be handled transparently, its what I meant by this section in my previous comment
```rust
  let _ = self.consume_token('=')?;
  let value = self.parse_expr()?;
  Ok(Some(SqlOption::KeyValue {
    key, value
  }))
```
in that the ideal scenario avoids enumerating the various options where possible

src/parser/mod.rs
```diff
@@ -7081,18 +7029,243 @@ impl<'a> Parser<'a> {
...
+            let value = self.next_token();
+
+            match value.token {
+                // UNION [=] (tbl_name[,tbl_name]...)
+                Token::LParen => {
+                    let tables: Vec<Ident> =
+                        self.parse_comma_separated0(Parser::parse_identifier, Token::RParen)?;
+                    self.expect_token(&Token::RParen)?;
+
+                    return Ok(Some(SqlOption::Union(tables)));
```
iffyio, line 7151: Instead of calling this variant `Union` can make it reusable for future dialects or similar options? something like this maybe
```rust
SqlOption::NamedParenthesizedList {
     name: Ident, // Union, TABLEENGINE
     key: Option<Ident>, // table engine name
     values: Vec<Expr>, // (val1, val2 ...)
}
```

src/parser/mod.rs
tomershaniii: Thanks, fixed

src/parser/mod.rs
```diff
@@ -7082,18 +7030,243 @@ impl<'a> Parser<'a> {
...
+                                _ => self.expected("Token::Word", storage_token)?,
+                            }
+                        }
+                        false => None,
+                    };
+
+                    Ok(Some(SqlOption::TableSpace(TablespaceOption {
+                        name,
+                        storage,
+                    })))
```
iffyio, line 7130: thinking similar to `Union` it would be good to have a reusable representation for this, maybe?
```rust
SqlOption::NamedKeyValue {
     name: Ident, // TABLESPACE
     key: Ident,
     value: Expr
}
```

src/parser/mod.rs
tomershaniii: Grouped the branches by topic.
IMHO another `SqlOption` makes sense, in the same way `Expr::Substring` supports both the shorthand flavor (SUBSTR)

src/parser/mod.rs
tomershaniii: The Value part is handled transparently, not sure how we can prevent enumeration for the different keys (or maybe i've missed something ....)
Interesting to note that 'parse_expr' cannot be used in this context as it has explicit handling for the [COLLATE keyword](https://github.com/SatoriCyber/datafusion-sqlparser-rs/blob/62495f2f0d998690495f111063a0b0a3466ffa69/src/parser/mod.rs#L1622)

src/parser/mod.rs
tomershaniii: I like that, will do

src/parser/mod.rs
tomershaniii: IMO it's a little premature to generalize, not sure what future variations will look like.
(also, this type is different than others as it does not have the optional '=', see [tablespace_option](https://dev.mysql.com/doc/refman/8.4/en/create-table.html) )

tomershaniii:
I will have limited internet connectivity / dev time for the rest of this month, appreciate if we could get this PR merged today.

tomershaniii (approved):

src/parser/mod.rs
```diff
@@ -7081,18 +7029,243 @@ impl<'a> Parser<'a> {
...
+                    return Ok(Some(SqlOption::Union(tables)));
+                }
+                _ => {
+                    return self.expected("Token::LParen", value)?;
+                }
+            }
+        }
+
+        // Key/Value parameter option
+        let key = if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARSET]) {
```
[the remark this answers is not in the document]
mvzink: Please defer to @iffyio, avoiding parsers inside `Dialect` is just my personal preference 😛

tomershaniii:
@iffyio @mvzink could you please specify what you see as a blocker for this PR?

src/parser/mod.rs
iffyio: Ah so I meant essentially that the plain key value cases can be handled with e.g.
```rust
let key = self.parse_identifier()
```
without enumerating the individual keys.

It might be worth considering using Expr for the value part (the COLLATE scenario can be handled specially if it conflicts and is the lone edge case), note for example that Value doesnt support signed numbers

iffyio:
Hi @tomershaniii I havent made time to look closely at the diff, mostly skimmed the PR on a high level so far and it feels to me like its coming together. I left a comment now to that effect, sorry for the delay

mvzink:
> could you please specify what you see as a blocker for this PR?

No blockers IMO; like I said about the "putting subparsers inside `Dialect`" topic, that's just my opinion and you should defer to @iffyio. Looking forward to this landing, thanks for the work!

src/parser/mod.rs
tomershaniii: The key can be a complex object (as suggested in one of the previous comments), i.e. 
the term  like "DEFAULT CHARACTER SET" is parsed into an identifier ```Ident::new("DEFAULT CHARACTER SET")```, hence i do not believe blindly parsing an identifier is possible in this case.
(Another reason i would advise against it that there are several other table option formats which may clash)

tomershaniii: Regarding having Expr as the the Value, while I agree that this is a a more generic option i prefer saving this for future work for the following reasons:
- In the context of "plain table options" (mostly mysql), signed numbers. are not applicable
- My preference is to provide a uniform  implementation and avoid a one-off for the "collate" case, i have looked into moving the collate handling into a more specific location, but this seems beyond the scope of this PR .

iffyio: > the term like "DEFAULT CHARACTER SET" is parsed into an identifier

Ah so to clarify what I meant in case. I was reiterating my [previous comment here](https://github.com/apache/datafusion-sqlparser-rs/pull/1747#discussion_r2011373382), in that the two-three word keys like 'DEFAULT CHARACTER SET' are handled explicitly, whereas the majority are single identifiers and can be handled with the one-liner as above.

So essentially that in the code we handle the special options like `DEFAULT CHARACTER SET` explicitly, then the rest are covered with the catch all `let key = self.parse_identifier()`, would that not work do you mean?

Similarly for `COLLATE`

> In the context of "plain table options" (mostly mysql), signed numbers. are not applicable

I think this would be acceptable and downstream crates may validate signedness if neccessary.

>  i have looked into moving the collate handling into a more specific location, but this seems beyond the scope of this PR

Not sure I understand, I didn't imagine something like that would be necessary, what I figured was that the collate option would be handled explicitly similar to the current code `if self.parse_keyword(Keyword::COLLATE) { ... }`

tomershaniii: Thanks for the clarification, 
From a code readability perspective having both implicit & explicit table options doesn't make me too comfortable, if you think this is the better approach, i'm with you :-)

Can you suggest an elegant way to assert the end of the options list without explicitly identifying tokens:
e.g. `CREATE TABLE tbl_name KEY_BLOCK_SIZE 256 AS SELECT 1` we would attempt to parse `AS SELECT` as Key/Value

iffyio: hmm yeah not sure if there's a nice way to do that actually, we could probably check that the next token is the start of the next clause, from the docs those look like one of `PARTITION`, `AS`, `IGNORE`, `REPLACE`

tomershaniii: The list is a little longer:
                    Keyword::PARTITION,
                    Keyword::AS,
                    Keyword::IGNORE,
                    Keyword::REPLACE,
                    Keyword::ORDER,
                    Keyword::PRIMARY,
                    Keyword::ON,
                    Keyword::SELECT,
                    Keyword::STRICT,
Caveats to this approach:
- Some negative unit tests need revising (they are failing differently)
- The parse_plain_option is now concerned with keywords which are not option related (i see this as a mix of different layers/concerns)
- We save on keywords, but we may end up with a negative list which is just as long

Unless you have a strong preference, i prefer we leave the existing implementation, LMKWYT

iffyio: I think we can go with what we currently have in this PR

src/ast/helpers/stmt_create_table.rs
```diff
@@ -76,27 +78,20 @@ pub struct CreateTableBuilder {
...
     pub query: Option<Box<Query>>,
     pub without_rowid: bool,
     pub like: Option<ObjectName>,
     pub clone: Option<ObjectName>,
-    pub engine: Option<TableEngine>,
-    pub comment: Option<CommentDef>,
-    pub auto_increment_offset: Option<u32>,
-    pub default_charset: Option<String>,
-    pub collation: Option<String>,
+    pub comment_after_column_def: Option<CommentDef>,
```
iffyio, line 87: can we call it plainly `comment` as previously or was there a requirement for the naming?

src/ast/mod.rs
```diff
@@ -8543,27 +8617,13 @@ impl Display for CreateViewParams {
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.name)?;
-
-        if let Some(parameters) = self.parameters.as_ref() {
-            write!(f, "({})", display_comma_separated(parameters))?;
-        }
-
-        Ok(())
-    }
+pub struct NamedParenthesizedList {
```
iffyio, line 8623: Can we add a doc comment with an example of what the option looks like? in order to illustrate what the fields of the struct map to as well

src/ast/mod.rs
```diff
@@ -7320,10 +7351,53 @@ impl fmt::Display for SqlOption {
                     display_comma_separated(for_values)
                 )
             }
+            SqlOption::TableSpace(tablespace_option) => {
+                write!(f, "TABLESPACE {}", tablespace_option.name)?;
+                match tablespace_option.storage {
+                    Some(StorageType::Disk) => write!(f, " STORAGE DISK"),
+                    Some(StorageType::Memory) => write!(f, " STORAGE MEMORY"),
+                    _ => Ok(()),
```
iffyio, line 7359:
```suggestion
                    None => Ok(()),
```

src/ast/mod.rs
```diff
@@ -8543,27 +8617,13 @@ impl Display for CreateViewParams {
...
-        if let Some(parameters) = self.parameters.as_ref() {
-            write!(f, "({})", display_comma_separated(parameters))?;
-        }
-
-        Ok(())
-    }
+pub struct NamedParenthesizedList {
+    pub key: Ident,
+    pub value: Option<Ident>,
+    pub parameters: Vec<Ident>,
```
iffyio, line 8626: hmm not sure re the naming of the properties, they don't hint anything being named as the `NamedParenthesizeList` suggest and the parameters property is essentially a list of values as well so that the current struct is declaring a `value` and `values` which feels odd. maybe replacing `value` with `name`, and `parameters` with `values` would make more sense? Since currently if `value` is None, then the `Vec` isn't representing parameters for that option, rather a plain list of values

src/ast/mod.rs
```diff
@@ -7289,6 +7310,16 @@ pub enum SqlOption {
...
     },
+
+    Comment(CommentDef),
+
+    TableSpace(TablespaceOption),
+
+    // Advanced parameter formations
+    // UNION  = (tbl_name[,tbl_name]...)
+    // ENGINE = ReplicatedMergeTree('/table_name','{replica}', ver)
+    // ENGINE = SummingMergeTree([columns])
```
iffyio, line 7321:
```suggestion
    /// An option representing a key value pair, where the value is a parenthesized list and with an optional name.
    ///
    /// ```sql
    /// UNION  = (table1, table2)
    /// ENGINE = ReplicatedMergeTree('/table_name','{replica}', ver)
    /// ENGINE = SummingMergeTree(column1, column2)
    /// ```
```

src/ast/mod.rs
```diff
@@ -7289,6 +7310,16 @@ pub enum SqlOption {
         range_direction: Option<PartitionRangeDirection>,
         for_values: Vec<Expr>,
     },
+
+    Comment(CommentDef),
+
+    TableSpace(TablespaceOption),
```
iffyio, line 7316: Can we add descriptions for these variants, also it would be good at least with the tablespace version to link to the docs where the option comes from?

src/ast/mod.rs
```diff
@@ -7320,10 +7351,53 @@ impl fmt::Display for SqlOption {
...
             }
+            SqlOption::TableSpace(tablespace_option) => {
+                write!(f, "TABLESPACE {}", tablespace_option.name)?;
+                match tablespace_option.storage {
+                    Some(StorageType::Disk) => write!(f, " STORAGE DISK"),
+                    Some(StorageType::Memory) => write!(f, " STORAGE MEMORY"),
+                    _ => Ok(()),
+                }
+            }
+
```
iffyio, line 7362:
```suggestion
```

src/ast/mod.rs
```diff
@@ -7320,10 +7351,53 @@ impl fmt::Display for SqlOption {
...
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum StorageType {
+    Disk,
+    Memory,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
iffyio, line 7393: Can we add a description and link to the docs?

src/parser/mod.rs
```diff
@@ -4217,7 +4217,7 @@ impl<'a> Parser<'a> {
         }
     }
 
-    fn parse<T: FromStr>(s: String, loc: Location) -> Result<T, ParserError>
+    pub fn parse<T: FromStr>(s: String, loc: Location) -> Result<T, ParserError>
```
iffyio, line 4220:
```suggestion
    fn parse<T: FromStr>(s: String, loc: Location) -> Result<T, ParserError>
```
was there a requirement for this change to pub? if needed elsewhere inthe crate we can use `pub(crate)``instead

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
+            .map(Expr::Value)
+        {
+            Some(expr) => expr,
+            None => Expr::Identifier(self.parse_identifier()?),
+        };
+
+        Ok(Some(SqlOption::KeyValue { key, value }))
+    }
+
+    pub fn parse_plain_options(&mut self) -> Result<Vec<SqlOption>, ParserError> {
```
iffyio, line 7269:
```suggestion
    fn parse_plain_options(&mut self) -> Result<Vec<SqlOption>, ParserError> {
```

iffyio, line 7269: Should this rather be called `parse_sql_options` / `parse_sql_option`? not sure its clear what we mean by plain option in this context?

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("CHARSET")
+        } else if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARACTER, Keyword::SET]) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("DEFAULT CHARACTER SET")
+        } else if self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET]) {
+            // [DEFAULT] CHARACTER SET [=] charset_name
+            Ident::new("CHARACTER SET")
+        } else if self.parse_keywords(&[Keyword::DEFAULT, Keyword::COLLATE]) {
+            // [DEFAULT] COLLATE [=] collation_name
```
iffyio, line 7182:
```suggestion
            // [DEFAULT] CHARACTER SET [=] charset_name
            Ident::new("DEFAULT CHARSET")
        } else if self.parse_keyword(Keyword::CHARSET) {
            // [DEFAULT] CHARACTER SET [=] charset_name
            Ident::new("CHARSET")
        } else if self.parse_keywords(&[Keyword::DEFAULT, Keyword::CHARACTER, Keyword::SET]) {
            Ident::new("DEFAULT CHARACTER SET")
        } else if self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET]) {
            Ident::new("CHARACTER SET")
        } else if self.parse_keywords(&[Keyword::DEFAULT, Keyword::COLLATE]) {
```
similar for the others, I think we can skip the comments describing grammar for the key and value. Since the value doesn't affect the parser, and the key is already being described by the parse_keywords function call.

On a related note, I think its important if we can find a way to flag the documentation to mysql for the applicable options, ideally without mentioning it per-option but maybe that's the easiest way if the sources across the options are mixed

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
+
+            return tablespace;
+        }
+
+        if self.parse_keyword(Keyword::UNION) {
+            let _ = self.consume_token(&Token::Eq);
+            let value = self.next_token();
+
+            match value.token {
+                // UNION [=] (tbl_name[,tbl_name]...)
```
iffyio, line 7148: Can we also mention where this syntax comes from?

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
+
+            return engine;
+        }
+
+        if self.parse_keywords(&[Keyword::TABLESPACE]) {
+            let _ = self.consume_token(&Token::Eq);
+            let value = self.next_token();
+
+            let tablespace = match value.token {
+                // TABLESPACE tablespace_name [STORAGE DISK] | [TABLESPACE tablespace_name] STORAGE MEMORY
```
iffyio, line 7111: Can we include a link to where the syntax comes from?

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
+                    Ok(Some(SqlOption::Comment(CommentDef::WithoutEq(s))))
+                }
+                (_, token) => {
+                    self.expected("Token::SingleQuotedString", TokenWithSpan::wrap(token))
+                }
+            };
+            return comment;
+        }
+
+        if self.parse_keywords(&[Keyword::ENGINE]) {
```
iffyio, line 7078: Can we include a link to where this syntax comes from?

src/parser/mod.rs
```diff
@@ -7082,18 +7030,252 @@ impl<'a> Parser<'a> {
...
             partition_by,
             cluster_by,
-            options,
+            table_options,
         })
     }
 
+    fn parse_plain_option(&mut self) -> Result<Option<SqlOption>, ParserError> {
+        // Single parameter option
+        if self.parse_keywords(&[Keyword::START, Keyword::TRANSACTION]) {
```
iffyio, line 7055: Can we include a link to where this syntax comes from?

tests/sqlparser_mysql.rs
```diff
@@ -859,29 +873,276 @@ fn parse_create_table_comment() {
...
     }
 }
 
+#[test]
+fn parse_create_table_with_all_table_options() {
+    let sql =
+        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE = InnoDB AUTO_INCREMENT = 123 DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_0900_ai_ci INSERT_METHOD = FIRST KEY_BLOCK_SIZE = 8 ROW_FORMAT = DYNAMIC DATA DIRECTORY = '/var/lib/mysql/data' INDEX DIRECTORY = '/var/lib/mysql/index' PACK_KEYS = 1 STATS_AUTO_RECALC = 1 STATS_PERSISTENT = 0 STATS_SAMPLE_PAGES = 128 DELAY_KEY_WRITE = 1 COMPRESSION = 'ZLIB' ENCRYPTION = 'Y' MAX_ROWS = 10000 MIN_ROWS = 10 AUTOEXTEND_SIZE = 64 AVG_ROW_LENGTH = 128 CHECKSUM = 1 CONNECTION = 'mysql://localhost' ENGINE_ATTRIBUTE = 'primary' PASSWORD = 'secure_password' SECONDARY_ENGINE_ATTRIBUTE = 'secondary_attr' START TRANSACTION TABLESPACE my_tablespace STORAGE DISK UNION = (table1, table2, table3)";
+
+    let x = mysql().verified_stmt(sql);
+    println!("{x:?}");
```
iffyio, line 965:
```suggestion
```

tests/sqlparser_mysql.rs
```diff
@@ -859,29 +873,276 @@ fn parse_create_table_comment() {
...
         }
     }
 }
 
+#[test]
+fn parse_create_table_with_all_table_options() {
+    let sql =
+        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE = InnoDB AUTO_INCREMENT = 123 DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_0900_ai_ci INSERT_METHOD = FIRST KEY_BLOCK_SIZE = 8 ROW_FORMAT = DYNAMIC DATA DIRECTORY = '/var/lib/mysql/data' INDEX DIRECTORY = '/var/lib/mysql/index' PACK_KEYS = 1 STATS_AUTO_RECALC = 1 STATS_PERSISTENT = 0 STATS_SAMPLE_PAGES = 128 DELAY_KEY_WRITE = 1 COMPRESSION = 'ZLIB' ENCRYPTION = 'Y' MAX_ROWS = 10000 MIN_ROWS = 10 AUTOEXTEND_SIZE = 64 AVG_ROW_LENGTH = 128 CHECKSUM = 1 CONNECTION = 'mysql://localhost' ENGINE_ATTRIBUTE = 'primary' PASSWORD = 'secure_password' SECONDARY_ENGINE_ATTRIBUTE = 'secondary_attr' START TRANSACTION TABLESPACE my_tablespace STORAGE DISK UNION = (table1, table2, table3)";
+
+    let x = mysql().verified_stmt(sql);
```
iffyio, line 964:
```suggestion
```
given the match version below is identical

iffyio (commented):
Thanks @tomershaniii! The changes look reasonable to me overall, left some minor comments/questions

src/ast/helpers/stmt_create_table.rs
tomershaniii: All dialects have the comment as part of the options, hence they are part of the generic comment list.
Hive has a different definition of where the comment should be, previously this was defined by an enum variant see [here](https://github.com/apache/datafusion-sqlparser-rs/blob/a5b9821d1d2fa9c0b8ee73b698a2b0e5f138beaf/src/ast/mod.rs#L8847)
The previous 'generic comment' is now part of the generic list (SqlOption), i believe the name should reflect this context (open to alternative names).

iffyio: Ah I see, I think we can leave it as `comment` and include a description on the field to highlight the difference. I'm suspecting that `comment_after_column_def` or similar naming could raise more questions than it answers

src/parser/mod.rs
tomershaniii: Check out `CreateTableOptions` it explain the different variations for create table options.
Open to naming changes...

tomershaniii:
@iffyio  See last commit, we should be god to go :-)

tomershaniii:
@iffyio Note that i am unable to resolve the issues on the PR (as i am not the owner of the original PR), please consider all issues resolved.

iffyio (approved):
LGMT! Thanks @tomershaniii and @benrsatori!
cc @alamb

alamb:
This is very nice -- thank you

PR #1751, opened 2025-02-27
Standardize CREATE TABLE options equals signs

mvzink:
* Make spaces around equals signs canonical: `ENGINE = InnoDB` instead of `ENGINE=InnoDB`
* Make equals signs canonical: `AUTO_INCREMENT = 100` instead of `AUTO_INCREMENT 100`
* Make equals signs optional for charset, engine, and collation to match MySQl: `COLLATE utf8mb4_unicode_ci` works as well as `COLLATE = utf8mb4_unicode_ci`

In some dialects, some options might require the equals signs, so will be more permissive in some situations. For example, [ClickHouse] requires the equals signs for `ENGINE`. It didn't seem worth special casing this to me.

[ClickHouse]: https://clickhouse.com/docs/en/sql-reference/statements/create/table/#create-table-options

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -860,11 +860,11 @@ fn parse_create_table_comment() {
 #[test]
 fn parse_create_table_auto_increment_offset() {
     let canonical =
-        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE=InnoDB AUTO_INCREMENT 123";
-    let with_equal =
-        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE=InnoDB AUTO_INCREMENT=123";
+        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE = InnoDB AUTO_INCREMENT = 123";
+    let without_equal =
+        "CREATE TABLE foo (bar INT NOT NULL AUTO_INCREMENT) ENGINE = InnoDB AUTO_INCREMENT 123";
 
-    for sql in [canonical, with_equal] {
+    for sql in [canonical, without_equal] {
         match mysql().one_statement_parses_to(sql, canonical) {
             Statement::CreateTable(CreateTable {
                 name,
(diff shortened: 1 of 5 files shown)


mvzink:
Note that a version of #1747 could which standardizes all options a la `SqlOption` could supersede this.

tests/sqlparser_mysql.rs
```diff
@@ -1047,6 +1047,174 @@ fn parse_create_table_gencol() {
...
+
+    // Test mixing options with and without equals sign
+    mysql_and_generic().one_statement_parses_to(
+            "CREATE TABLE foo (id INT) ENGINE InnoDB AUTO_INCREMENT = 100 CHARSET = utf8mb4 COLLATE utf8mb4_unicode_ci",
+            "CREATE TABLE foo (id INT) ENGINE = InnoDB AUTO_INCREMENT = 100 DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_unicode_ci",
+        );
+}
+
+#[test]
+#[ignore = "not yet supported"]
```
mvzink, line 1096: Happy to remove this, especially if it ends up being covered by #1747 or something.

tests/sqlparser_mysql.rs
```diff
@@ -1047,6 +1047,174 @@ fn parse_create_table_gencol() {
...
+#[test]
+fn parse_create_table_without_equals() {
+    mysql_and_generic().one_statement_parses_to(
+        "CREATE TABLE foo (id INT) CHARSET utf8mb4",
+        "CREATE TABLE foo (id INT) DEFAULT CHARSET = utf8mb4",
+    );
+
+    mysql_and_generic().one_statement_parses_to(
+        "CREATE TABLE foo (id INT) COLLATE utf8mb4_unicode_ci",
+        "CREATE TABLE foo (id INT) COLLATE = utf8mb4_unicode_ci",
```
iffyio, line 1059: I'm thinking the mismatches aren't ideal if we would rather represent the input SQL faithfully?

tests/sqlparser_mysql.rs
iffyio: Yeah I think we can remove and add it back when we have support for the syntax

tests/sqlparser_mysql.rs
mvzink: Well, I think ideally these options would all be in a vector similar to `Vec<SqlOption>`, but with a flag for the equals sign like I added for `ALGORITHM` and `AUTO_INCREMENT`, which would avoid these mismatches. Maybe I will see how #1747 shakes out and then revisit this, and mark this draft in the meantime.

PR #1752, opened 2025-02-28
Parse SET NAMES syntax in Postgres

mvzink:
Postgres also supports a limited version of the `SET NAMES` syntax already supported for MySQL (but doesn't support the `COLLATE` part).

This also switches the charset name to an `Ident` simply to preserve the quoting from the input.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10962,14 +10962,14 @@ impl<'a> Parser<'a> {
             OneOrManyWithParens::One(self.parse_object_name(false)?)
         };
 
-        if matches!(&variables, OneOrManyWithParens::One(variable) if variable.to_string().eq_ignore_ascii_case("NAMES")
-            && dialect_of!(self is MySqlDialect | GenericDialect))
-        {
+        let names = matches!(&variables, OneOrManyWithParens::One(variable) if variable.to_string().eq_ignore_ascii_case("NAMES"));
+
+        if names && self.dialect.supports_set_names() {
             if self.parse_keyword(Keyword::DEFAULT) {
                 return Ok(Statement::SetNamesDefault {});
             }
 
-            let charset_name = self.parse_literal_string()?;
+            let charset_name = self.parse_identifier()?;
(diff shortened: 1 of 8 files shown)


src/parser/mod.rs
```diff
@@ -10962,15 +10962,17 @@ impl<'a> Parser<'a> {
...
+
+        if names && self.dialect.supports_set_names() {
             if self.parse_keyword(Keyword::DEFAULT) {
                 return Ok(Statement::SetNamesDefault {});
             }
 
-            let charset_name = self.parse_literal_string()?;
-            let collation_name = if self.parse_one_of_keywords(&[Keyword::COLLATE]).is_some() {
+            let charset_name = self.parse_identifier()?;
+            let collation_name = if self.dialect.supports_set_names_collation()
```
iffyio, line 10973: wondering if the `supports_set_names_collation()` flag necessary or if we could skip it? the syntax looks like we might be able to parse it unconditionally/unambiguous given it shows up after an identifier

tests/sqlparser_postgres.rs
```diff
@@ -5533,3 +5533,11 @@ fn parse_varbit_datatype() {
         _ => unreachable!(),
     }
 }
+
+#[test]
+fn parse_set_names() {
+    pg_and_generic().verified_stmt("SET NAMES 'UTF8'");
+    pg_and_generic().verified_stmt("SET NAMES 'utf8'");
+    pg().parse_sql_statements("SET NAMES 'UTF8' COLLATE bogus")
+        .expect_err("Postgres does not support COLLATE");
```
iffyio, line 5542: can we move this test to common and to use the `all_dialects_where` style for coverage?

src/parser/mod.rs
mvzink: The only reason to add the flag was to fail parsing in Postgres; but you're right that it doesn't introduce ambiguity, so it would probably be fine to be more permissive.

iffyio (approved):
Thanks @mvzink!
cc @alamb

PR #1754, opened 2025-02-28
re-add support for nested comments in mssql

### LINKED ISSUE #1753: Nested comments were broken by the sqlparser v0.54 release
Nested `/*` comments used to be parsed as such by sqlparser, but have been broken by the 0.54 release.

The following works in postgresql and mssql:

```
select /* /* comment */ */ 1;
```

and it used to be parsed correctly by sqlparser, but now is not.

Initially reported in https://github.com/sqlpage/SQLPage/issues/834

lovasoa:
fix #1753

see https://learn.microsoft.com/en-us/sql/t-sql/language-elements/slash-star-comment-transact-sql?view=sql-server-ver16

### DIFF
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -95,4 +95,9 @@ impl Dialect for MsSqlDialect {
     fn supports_timestamp_versioning(&self) -> bool {
         true
     }
+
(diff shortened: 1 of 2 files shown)


iffyio (approved):
Thanks @lovasoa!
cc @alamb

PR #1755, opened 2025-03-04
Snowflake: Support dollar quoted comments

7phs:
Snowflake [supports](https://docs.snowflake.com/en/sql-reference/literals-table) dollar-quoted comments as string literals when creating tables, views, and their fields.

For example:
```
CREATE OR REPLACE TEMPORARY VIEW foo.bar.baz (
    "COL_1" COMMENT $$comment 1$$
) COMMENT = $$view comment$$ AS (
    SELECT 1
)
```

This pull request adds support for this type of comments.

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -245,14 +245,6 @@ impl Dialect for SnowflakeDialect {
                     .map(|p| Some(ColumnOption::Policy(ColumnPolicy::ProjectionPolicy(p)))))
             } else if parser.parse_keywords(&[Keyword::TAG]) {
                 Ok(parse_column_tags(parser, with).map(|p| Some(ColumnOption::Tags(p))))
-            } else if parser.parse_keywords(&[Keyword::COMMENT]) {
-                let next_token = parser.next_token();
-                match next_token.token {
-                    Token::DollarQuotedString(value, ..) => {
-                        Ok(Ok(Some(ColumnOption::Comment(value.value))))
-                    }
-                    _ => Err(ParserError::ParserError("not found match".to_string())),
-                }
             } else {
                 Err(ParserError::ParserError("not found match".to_string()))
             }
@@ -430,7 +422,7 @@ pub fn parse_create_table(
                 Keyword::COMMENT => {
                     // Rewind the COMMENT keyword
                     parser.prev_token();
-                    builder = builder.comment(parser.parse_optional_inline_comment(true)?);
+                    builder = builder.comment(parser.parse_optional_inline_comment()?);
                 }
                 Keyword::AS => {
                     let query = parser.parse_query()?;
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -644,11 +644,7 @@ pub fn parse_create_stage(
     // [ comment ]
     if parser.parse_keyword(Keyword::COMMENT) {
         parser.expect_token(&Token::Eq)?;
-        comment = Some(match parser.next_token().token {
-            Token::SingleQuotedString(word) => Ok(word),
-            Token::DollarQuotedString(word) => Ok(word.value),
-            _ => parser.expected("a comment statement", parser.peek_token()),
-        }?)
+        comment = Some(parser.parse_comment_value()?);
     }
 
     Ok(Statement::CreateStage {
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -6922,18 +6924,23 @@ impl<'a> Parser<'a> {
...
             let next_token = self.next_token();
-            match next_token.token {
-                Token::SingleQuotedString(str) => Some(if has_eq {
-                    CommentDef::WithEq(str)
-                } else {
-                    CommentDef::WithoutEq(str)
-                }),
+            let comment = match next_token.token {
+                Token::SingleQuotedString(str) => str,
+                Token::DollarQuotedString(str) if support_dollar_quoted_comment => str.value,
```
iffyio, line 6936: I think no need to guard this with a flag, we can always accept a dollar quoted string if one shows up. Also noticed we have similar [function here](https://github.com/apache/datafusion-sqlparser-rs/blob/24a2968b33787626fe63fc626c19dbfb50ec48f5/src/parser/mod.rs#L5323-L5328) could we maybe pull that part out into a helper and reuse in both places?

tests/sqlparser_snowflake.rs
```diff
@@ -976,6 +976,27 @@ fn parse_sf_create_or_replace_with_comment_for_snowflake() {
...
+        )"#
+        )
+        .is_ok());
+
+    assert!(snowflake()
+        .parse_sql_statements(
+            r#"CREATE TABLE my_table (
+            a STRING COMMENT $$comment 1$$
+        ) COMMENT = $$table comment$$"#
+        )
```
iffyio, line 996: can we use `snowflake().verified_stmt(sql)` for the tests?

src/dialect/snowflake.rs
```diff
@@ -245,6 +245,14 @@ impl Dialect for SnowflakeDialect {
                     .map(|p| Some(ColumnOption::Policy(ColumnPolicy::ProjectionPolicy(p)))))
             } else if parser.parse_keywords(&[Keyword::TAG]) {
                 Ok(parse_column_tags(parser, with).map(|p| Some(ColumnOption::Tags(p))))
+            } else if parser.parse_keywords(&[Keyword::COMMENT]) {
+                let next_token = parser.next_token();
+                match next_token.token {
+                    Token::DollarQuotedString(value, ..) => {
+                        Ok(Ok(Some(ColumnOption::Comment(value.value))))
+                    }
+                    _ => Err(ParserError::ParserError("not found match".to_string())),
```
iffyio, line 254: hmm the error message sounds like it'd be confusing what we mean by 'not found match'.

Left a comment below about reusing the parsing code, we probably want to reuse it here as well (any reason not to cover the single quote case here)?

src/parser/mod.rs
7phs: A parameter `support_dollar_quoted_comment` is removed.

tests/sqlparser_snowflake.rs
7phs: 1. There is no code for storing a comment quotation. Could you clarify whether it makes sense to update all the code that handles comment?
2. Snowflake converts a dollar-quoted comment into a single-quoted comment in the DDL representation of table/view creation.

src/dialect/snowflake.rs
7phs: That makes sense. I have moved this code to the parser level.

tests/sqlparser_snowflake.rs
iffyio: Ah that makes sense, I think we can leave as is in that case in order to limit the scope of the PR

src/parser/mod.rs
```diff
@@ -6926,14 +6927,16 @@ impl<'a> Parser<'a> {
...
-                    CommentDef::WithEq(str)
-                } else {
-                    CommentDef::WithoutEq(str)
-                }),
+            let comment = match next_token.token {
+                Token::SingleQuotedString(str) => str,
+                Token::DollarQuotedString(str) => str.value,
                 _ => self.expected("comment", next_token)?,
-            }
+            };
```
iffyio, line 6934: can we move this part to be reusable? something like
```rust
fn parse_comment_value() -> Result<String> {
          let value = match next_token.token {
                Token::SingleQuotedString(str) => str,
                Token::DollarQuotedString(str) => str.value,
                _ => self.expected("comment", next_token)?,
            };
         Ok(value)
}
```
and then to call the function from both places

iffyio (approved):
LGTM! Thanks @7phs!
cc @alamb

alamb:
🚀

PR #1757, opened 2025-03-06
Update assignment list

### LINKED ISSUE #1694: Parse MySQL `SET GLOBAL` variables

MohamedAbdeen21:
Closes #1694

- Minor refactoring of the `parse_set` parser (using match, extracting functions, etc ..)
- Support multiple comma-separated assignments (`SET a = 1, b = 2, ...`)


src/dialect/mysql.rs
```diff
@@ -141,6 +141,10 @@ impl Dialect for MySqlDialect {
     fn supports_set_names(&self) -> bool {
         true
     }
+
+    fn supports_comma_separated_set_assignments(&self) -> bool {
+        true
```
mvzink, line 146: Probably should set this to true in `GenericDialect` too.

src/parser/mod.rs
```diff
@@ -10961,127 +10961,184 @@ impl<'a> Parser<'a> {
...
-            self.expect_keywords(&[Keyword::AS, Keyword::TRANSACTION])?;
-            Ok(Statement::SetTransaction {
-                modes: self.parse_transaction_modes()?,
-                snapshot: None,
-                session: true,
-            })
-        } else if variable.to_string() == "TRANSACTION" && modifier.is_none() {
-            if self.parse_keyword(Keyword::SNAPSHOT) {
-                let snapshot_id = self.parse_value()?.value;
+        match variable.to_string().to_ascii_uppercase().as_str() {
```
mvzink, line 11088: Would be nice to avoid allocating a new `String` for every variable for this check (though only doing it once with this refactor is an improvement). I think you could match on the `ObjectName` and only apply this check if there's a single `Ident`, and just directly compare to its value (with a series of if-else `ident.value.eq_ignore_ascii_case`).

That said—unless it turns out that these different specially named set statements can also be done multiply (e.g. `SET NAMES utf8, TIMEZONE utc`), which I doubt and we already don't support—I actually think it would be simpler to just handle these cases individually up front, instead of passing them through this intermediary `variable`. As an example, the non-locality introduced by checking for `TIME ZONE` and turning it into `TIMEZONE` earlier in the function, knowing it will be handled specially here, just makes this difficult to read, has a needless performance hit, and doesn't buy us anything as far as I can tell. I don't thinking starting `parse_set` with a series of checks for these keywords would be a problem, before jumping into arbitrarily-named variable assignment parsing.

src/parser/mod.rs
```diff
@@ -10961,127 +10961,184 @@ impl<'a> Parser<'a> {
...
+            "CHARACTERISTICS" => {
+                self.expect_keywords(&[Keyword::AS, Keyword::TRANSACTION])?;
+                return Ok(Statement::SetTransaction {
+                    modes: self.parse_transaction_modes()?,
+                    snapshot: None,
+                    session: true,
+                });
+            }
+            "TRANSACTION" if modifier.is_none() => {
+                if self.parse_keyword(Keyword::SNAPSHOT) {
```
mvzink, line 11121: This could probably be something more like:

```rust
let snapshot = if self.parse_keyword(Keyword::SNAPSHOT) {
    self.parse_value()?.value
} else {
    None
}
Ok(Statement::SetTransaction {
    snapshot,
    ..
}
```

tests/sqlparser_common.rs
```diff
@@ -14654,3 +14640,23 @@ fn parse_set_names() {
     dialects.verified_stmt("SET NAMES 'utf8'");
     dialects.verified_stmt("SET NAMES UTF8 COLLATE bogus");
 }
+
+#[test]
+fn parse_multiple_set_statements() -> Result<(), ParserError> {
+    let dialects = all_dialects_where(|d| d.supports_comma_separated_set_assignments());
+    let stmt = dialects.parse_sql_statements("SET @a = 1, b = 2")?;
```
mvzink, line 14647: I think it would be preferable to have a `verified_stmt` here, and do more explicit checks for the variables and values (not just length checks) below. For one thing, if my other comment about MySQL style getting rendered as Snowflake style is correct, using `verified_stmt` here would catch that.

src/parser/mod.rs
```diff
@@ -10961,127 +10961,182 @@ impl<'a> Parser<'a> {
...
+        }
 
-        if self.consume_token(&Token::Eq) || self.parse_keyword(Keyword::TO) {
-            if parenthesized_assignment {
-                self.expect_token(&Token::LParen)?;
-            }
+        if self.dialect.supports_comma_separated_set_assignments() {
+            if let Ok(v) =
+                self.try_parse(|parser| parser.parse_comma_separated(Parser::parse_set_assignment))
+            {
```
mvzink, line 11037: If I'm reading this `if` branch correctly, this will actually translate MySQL-style syntax into Snowflake-style syntax (i.e. the test string below will parse the same as, and get `format`ted as, `SET (@a, b) = (1, 2)`, because we still just have a pair of `OneOrManyWithParens` in the `SetVariable`, and nothing to tell us which style was used. Please let me know if that's wrong.

Otherwise, I don't know what would be the best way to modify/replace `SetVariable` to represent this distinction, but we'll need something. In line with a recent PR to improve `CASE` handling (#1733), I would suggest that if we parse the MySQL style, we should have an entirely different `Statement` variant along the lines of `SetVariables { assignments: Vec<(Scope, Ident, Expr)> }`. (Note that with scope, I'm looking ahead to #1694 because each variable assignment has its own scope, and it could be either `@@global.var` or `GLOBAR var` style, which isn't currently supported.)

(And I would maybe even replace `Statement::SetVariable` with `Statement::Set(SetStatement)` to encapsulate all of these by having `parse_set` return `SetStatement`, but I don't really know whether others would find that desirable.)

mvzink (changes requested):

src/parser/mod.rs
```diff
@@ -10961,127 +10961,182 @@ impl<'a> Parser<'a> {
...
+                let (variables, values): (Vec<_>, Vec<_>) = v.into_iter().unzip();
 
-            let mut values = vec![];
-            loop {
-                let value = if let Some(expr) = self.try_parse_expr_sub_query()? {
-                    expr
-                } else if let Ok(expr) = self.parse_expr() {
-                    expr
+                let variables = if variables.len() == 1 {
+                    variables.into_iter().next().unwrap()
```
iffyio, line 11041: Can we replace this unwrap with an error? that way we don't panic if we get our assumptions wrong

MohamedAbdeen21: I agree, but it's wrapped in an if-statement that checks the length so it will never panic

src/dialect/mysql.rs
MohamedAbdeen21: That will cause some tests to fail. A dialect should either allow snowflake syntax or MySQL syntax.

For comma-separated lists, values are parsed using the `parse_expr` parser, which stops at commas.

For all other `SET` statements, we use a loop (extracted to a func called `parse_set_values` in this PR) that collects all comma-separated tokens as `Vec<value>`, because some dialects allow that (MSSQL and Hive for example). And by the way, that's what caused the output reported in the issue.

src/parser/mod.rs
MohamedAbdeen21: Updated, moved special cases to the top.

Can be done multiply IF the dialect supports the `supports_set_stmt_without_operator`, or just add assignment operators `SET NAMES TO utf8, TIMEZONE = 'utc'`

tests/sqlparser_common.rs
MohamedAbdeen21: You're other comment is correct; it gets translated to snowflake-style. Either update the enum/variants to reflect the syntax used or use `dialects.one_statement_parses_to("SET @a = 1, b = 2", "SET (@a, b) = (1, 2)");`

src/parser/mod.rs
MohamedAbdeen21: I prefer doing this in a follow-up. This PR refactored the entire function and may be annoying to review as-is. Plus, the discussion about the approach will be more visible in a separate follow-up

src/parser/mod.rs
MohamedAbdeen21: will require a second condition to also parse the modes field.

src/parser/mod.rs
mvzink: It should still probably be `next().expect("Only one variable")` or have a `// SAFETY:` comment.

src/parser/mod.rs
mvzink: I think if you want to unify all the `Statement::Set*` variants into `Statement::Set(SetStatement)` (or similar), that might be better in a separate PR. However, adding a new variant for the new syntax I don't think will add that much complexity to the PR—not enough to warrant merging a partial implementation of the new syntax IMO. Nothing else here changes/touches the AST type definitions, and just making this one branch return the new variant won't make this any harder to review; in fact, it will be a lot clearer IMO.

src/dialect/mysql.rs
MohamedAbdeen21: Updated, but the code is a little messier now

mvzink (approved):
Awesome, thank you. The only thing missing I can think of is the per-variable scope support represented by #1694, which should be a separate PR IMO. However, it will require changes to this new variant (and probably moving the `modifier` parsing around, not sure exactly how), so it would be good to keep in mind while reviewing this.

src/ast/mod.rs
```diff
@@ -2947,6 +2947,17 @@ pub enum Statement {
...
+    /// ```sql
+    /// SET <variable> = expression [, <variable> = expression]*;
+    /// ```
+    ///
+    /// Note: this is a MySQL-specific statement.
+    /// Refer to [`Dialect.supports_comma_separated_set_assignments`]
+    SetVariables {
+        variables: Vec<ObjectName>,
+        values: Vec<Expr>,
+    },
```
iffyio, line 2960: ah can we merge this with `Statement::SetVariable`? Having both `SetVariable` and `SetVariables` as different representations in the tree doesn't seem ideal, since they're similar and some dialects might need elements from both

mvzink: It seems @MohamedAbdeen21 was right that this aspect should be a different PR to have a more comprehensive discussion. I suggested this new variant in [another comment](https://github.com/apache/datafusion-sqlparser-rs/pull/1757#discussion_r1984144488) because I think that this is a different enough syntactic construct from the other two existing styles (single-variable assignment a la Postgres and parenthesized multiple-variable assignment a la Snowflake) that adding another flag to differentiate them would be unclear and cumbersome. Already, the use of `OneOrManyWithParens` to differentiate the Postgres and Snowflake styles is, in my opinion, inferior to a higher level enum variant differeniating the two styles. Furthermore, especially in the case of the new MySQL-style multiple-assignments-of-single-variables, I don't think that there enough in common or a way in which dialects would need elements from both, that warrants having them in the same variant.

However, I think my opinion would be clearer with two changes. First, when I initially suggested adding a new variant, I was imagining something more similar to `Case` with its `Vec<CaseWhen>`: i.e. having a `Vec<VariableAssignment>` or `Vec<(ObjectName, Expr)>` instead of two vecs which you later have to `zip` together to do anything with (i.e. in the `Display` implementation). This loses the structure that was parsed and only makes sense for the Snowflake style. Again, having that syle also be used for the Postgres single-assignment style is cumbersome and unclear, and this is similar. I actually didn't notice this difference from my original suggestion when re-reviewing this PR or I would have raised it.

The second change that might make my thinking clearer is the other one I suggested in that comment, which is to unify some or all `SET` statements in another struct/enum. I originally suggested `Statement::Set(SetStatement)` to include `SetNames` etc., but I actually think `Statement::SetVariable(SetVariable)` only covering these 3 cases would be better. Then any common handling between different types of variable assignments could still be handled centrally, but the 3 distinct styles could be represented more concretely, similar to this:

```rust
enum SetVariable {
    /// SQL Standard-style
    /// SET a = 1;
    SingleAssignment {
        local: bool,
        hivevar: bool,
        variable: ObjectName,
        value: Expr,
    }
    /// Snowflake-style
    /// SET (a, b) = (1, 2);
    ParenthesizedAssignments {
        variables: Vec<ObjectName>,
        values: Vec<Expr>,
    }
    /// MySQL-style
    /// SET a = 1, b = 2;
    MultipleAssignments {
        assignments: Vec<(ObjectName, Expr)>
    }
}
```

Shoe-horning all 3 of these into one struct/variant via a combination of `OneOrManyWithParens` and a new boolean flag differentiating Snowflake-style from MySQL-style that only applies when `variables` matches `OneOrManyWithParens::Many(_))` seems really annoying without, as far as I can tell, buying us anything.

If y'all think it would be clear enough to do it that way, I won't stand in the way, but I really think some kind of new variant/structure (if not precisely this one) would be a huge improvement.

iffyio: Thanks for clarify @mvzink! Introducing the `SetVariable` enum as above to the handle the different representations sounds reasonable to me.
I'm not sure it necessarily needs to be done as a follow up PR given the current PR would be introducing a new statement variant that is already on its way out, maybe we can make changes to the effect in this PR? 
@MohamedAbdeen21 wdyt?

MohamedAbdeen21: The main reason I wanted to have a follow-up was to give the discussion more visibility, but since we agree on the approach now, I'm down to implement the enum suggested above as part of this change.

I'll push the change before EOD

src/parser/mod.rs
```diff
@@ -10955,134 +10955,226 @@ impl<'a> Parser<'a> {
...
+        {
+            if self.consume_token(&Token::Eq) || self.parse_keyword(Keyword::TO) {
+                return Ok(Set::SingleAssignment {
+                    local: modifier == Some(Keyword::LOCAL),
+                    hivevar: modifier == Some(Keyword::HIVEVAR),
+                    variable: ObjectName::from(vec!["TIMEZONE".into()]),
+                    values: self.parse_set_values(false)?,
+                }
+                .into());
+            } else if self.dialect.is::<PostgreSqlDialect>() {
```
iffyio, line 11046: was there a requirement to gate this to the pg dialect? seems this is tied to the comment in the `parse_set_time_zone_alias` test, thinking we can probably allow the parser continue to work as before and reenable that test case?

iffyio (commented):
Thanks @MohamedAbdeen21! The changes look reasonable to me overall, left some comments

src/ast/mod.rs
```diff
@@ -6148,12 +6218,12 @@ impl fmt::Display for GrantObjects {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub struct Assignment {
+pub struct UpdateAssignment {
```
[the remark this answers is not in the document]
MohamedAbdeen21: No,  it's just to distinguish between the new SetAssignment and the existing Assignment

src/parser/mod.rs
MohamedAbdeen21: It's not a requirement, but we'd be successfully parsing something that is not supported by the dialect.

If there are more dialects that support this shorthand, then we can turn it into a `supports_` method.

MohamedAbdeen21: Now that I think about it, we can add a field that specifies if the shorthand syntax was used and only set it for pg.

src/ast/mod.rs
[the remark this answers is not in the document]
MohamedAbdeen21: I thought about merging them into one and calling it `Assignment`, but we will need a new field in SetAssignment for #1694 so I prefer having two separate structs

src/parser/mod.rs
MohamedAbdeen21:
```rs
    /// ```sql
    /// SET TIME ZONE <value>
    /// ```
    ///
    /// Note: this is a PostgreSQL-specific statements
    /// `SET TIME ZONE <value>` is an alias for `SET timezone TO <value>` in PostgreSQL
    SetTimeZone { local: bool, value: Expr },
```
We can 
1. enable it for all dialects
2. add a dialect method `supports_set_timezone_alias`
3. keep the current change (only for psql)
4. add a field that specifies if an assignment op was used.

I'm okay with any tbf as it doesn't affect my use case, so up to you.

iffyio: I think we can go with 1. that its fine to enable it for all dialects

src/ast/mod.rs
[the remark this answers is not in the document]
MohamedAbdeen21: Sounds reasonable. reverted.

iffyio (approved):
Also thanks for the reviews @mvzink!
cc @alamb

Issue #1758, opened 2025-03-09
Support SQL pipe operator


simonvandel:
Google BigQuery is releasing support for a pipe operator in their supported Sql dialect. See https://cloud.google.com/bigquery/docs/pipe-syntax-guide and https://research.google/pubs/sql-has-problems-we-can-fix-them-pipe-syntax-in-sql/

Databricks seem to be supporting a similar operator https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-pipeop.html

Relates to https://github.com/apache/datafusion/issues/14660 which is the Datafusion part

Progress

- [x] Limit (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Where (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] OrderBy (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Select (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Extend (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Set (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Drop (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] As (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Aggregate (https://github.com/apache/datafusion-sqlparser-rs/pull/1759)
- [x] Tablesample (https://github.com/apache/datafusion-sqlparser-rs/pull/1860)
- [ ] Rename (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Join (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Union (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Intersect (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Except (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Call (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Pivot (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)
- [ ] Unpivot (https://github.com/apache/datafusion-sqlparser-rs/pull/1879)

PR #1759, opened 2025-03-09
Support some of pipe operators

simonvandel:
Part of https://github.com/apache/datafusion-sqlparser-rs/issues/1758

Still missing (ran out of time today, can be done in follow-ups perhaps)
- join
- union|intersect|except
- call
- tablesample
- pivot
- unpivot

I'm a first time contributor - please let me know if there are better places to store the `PipeOperator` in the AST


src/ast/query.rs
```diff
@@ -2407,6 +2413,98 @@ impl fmt::Display for OffsetRows {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum PipeOperator {
```
iffyio, line 2419: Can we add a description and link to the docs https://cloud.google.com/bigquery/docs/pipe-syntax-guide#basic_syntax

src/parser/mod.rs
```diff
@@ -10319,6 +10321,104 @@ impl<'a> Parser<'a> {
...
+                            None
+                        };
+                        pipe_operators.push(PipeOperator::Limit { expr, offset })
+                    }
+                    // AGGREGATE <agg_expr> [[AS] alias], ...
+                    //
+                    // and
+                    //
+                    // AGGREGATE [<agg_expr> [[AS] alias], ...]
+                    // GROUP BY <grouping_expr> [AS alias], ...
```
iffyio, line 10386: Ah so I think these comments would be better placed on in the doc strings for each variant (left similar comment at that site)

src/parser/mod.rs
```diff
@@ -10319,6 +10321,104 @@ impl<'a> Parser<'a> {
                 None
             };
 
+            let mut pipe_operators = Vec::new();
+
+            // Syntax from "SQL Has Problems. We Can Fix Them: Pipe Syntax In SQL"
+            // https://storage.googleapis.com/gweb-research2023-media/pubtools/1004848.pdf
+            while self.consume_token(&Token::VerticalBarRightAngleBracket) {
```
iffyio, line 10328: can we move the logic to a dedicated function like `parse_pipe_operators()`?

src/parser/mod.rs
```diff
@@ -10319,6 +10321,104 @@ impl<'a> Parser<'a> {
                 None
             };
 
+            let mut pipe_operators = Vec::new();
+
+            // Syntax from "SQL Has Problems. We Can Fix Them: Pipe Syntax In SQL"
+            // https://storage.googleapis.com/gweb-research2023-media/pubtools/1004848.pdf
```
iffyio, line 10327:
```suggestion
```

src/dialect/mod.rs
```diff
@@ -508,6 +508,20 @@ pub trait Dialect: Debug + Any {
...
+    /// Return true if the dialect supports pipe operator.
+    ///
+    /// Example:
+    /// ```sql
+    /// SELECT *
+    /// FROM table
+    /// |> limit 1
+    /// ```
+    ///
+    /// See "SQL Has Problems. We Can Fix Them: Pipe Syntax In SQL" https://research.google/pubs/sql-has-problems-we-can-fix-them-pipe-syntax-in-sql/
```
iffyio, line 520: I think we can link to a concrete syntax like this instead
https://cloud.google.com/bigquery/docs/pipe-syntax-guide#basic_syntax

The paper wouldn't really help as much in the context of someone working on the codebase or trying to use the library

iffyio (commented):
Thanks @simonvandel! The changes look reasonable to me overall, left some comments

simonvandel:
Hi @iffyio 

I'm sorry for the long response time.
I have now revised the code according to your comments.
Adding more tests found some bugs, so I had to change the implementation a bit

src/parser/mod.rs
```diff
@@ -10574,11 +10598,96 @@ impl<'a> Parser<'a> {
...
+                        group_by_expr,
+                    })
+                }
+                Keyword::ORDER => {
+                    self.expect_one_of_keywords(&[Keyword::BY])?;
+                    let exprs = self.parse_comma_separated(Parser::parse_order_by_expr)?;
+                    pipe_operators.push(PipeOperator::OrderBy { exprs })
+                }
+                unhandled => {
+                    unreachable!("`expect_one_of_keywords` further up allowed unhandled keyword: {unhandled:?}");
```
iffyio, line 10684: can we instead return an error instead of panicking? to avoid crashing on bugs/mistakes

iffyio (approved):
Thanks @simonvandel! left a common re error handling, otherwise the changes look good to me

src/parser/mod.rs
simonvandel: Resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/1759/commits/6d4e722ad4e105ed6f57585d700c375d19bcd2b9

simonvandel:
Thanks @iffyio 

I resolved the merge conflict, fixed your comment on error handling.
Additionally, I added a dialect check before parsing pipe operators. See https://github.com/apache/datafusion-sqlparser-rs/pull/1759/commits/f01301dc7a85ce9139346c27e9b8d295e5225a8f

iffyio:
@simonvandel could you take a look at the CI issues when you get the time?

simonvandel:
> @simonvandel could you take a look at the CI issues when you get the time?

I think CI should be green with https://github.com/apache/datafusion-sqlparser-rs/pull/1759/commits/89309c1d48342d21497fd77b104e976d0c2cdb8e

iffyio (approved):
LGTM! Thanks @simonvandel!
cc @alamb

Issue #1763, opened 2025-03-10
CTEs (`WITH` clauses) in `DELETE`  statements


lovasoa:
SQLite, Postgres, and Mssql support WITH clauses attached to data-modifying expressions.

```sql
with x(y) as (select 1) UPDATE demo SET Name = 'j'; 
```

```sql
with x(y) as (select 1) delete from demo; 
```

```sql
with x(y) as (select 1) INSERT INTO demo VALUES(1,'v','h');
```

However, sqlparser cannot `WITH ... DELETE`

Initially reported in https://github.com/sqlpage/SQLPage/discussions/852

PR #1764, opened 2025-03-10
add support for `with` clauses (CTEs) in `delete` statements

lovasoa:
fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1763

fixes https://github.com/sqlpage/SQLPage/discussions/852

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10209,17 +10209,6 @@ impl<'a> Parser<'a> {
         }
     }
 
-    /// Parse a `WITH` clause, i.e. a `WITH` keyword followed by a `RECURSIVE` keyword
-    /// and a comma-separated list of CTE declarations.
-    fn parse_with_clause(&mut self) -> Result<With, ParserError> {
-        let with_token = self.get_current_token();
-        Ok(With {
-            with_token: with_token.clone().into(),
-            recursive: self.parse_keyword(Keyword::RECURSIVE),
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -10202,19 +10209,25 @@ impl<'a> Parser<'a> {
...
+    /// Parse a `WITH` clause, i.e. a `WITH` keyword followed by a `RECURSIVE` keyword
+    /// and a comma-separated list of CTE declarations.
+    fn parse_with_clause(&mut self) -> Result<With, ParserError> {
+        let with_token = self.get_current_token();
+        Ok(With {
+            with_token: with_token.clone().into(),
+            recursive: self.parse_keyword(Keyword::RECURSIVE),
+            cte_tables: self.parse_comma_separated(Parser::parse_cte)?,
+        })
+    }
```
iffyio, line 10221: oh was there a requirement to not have the logic inlined as before?

iffyio: Ah I see, was initially wondering in the diff if i missed something and figured to double check. yeah I think it we can have it inlined instead, the new function partially contains the impl since the `WITH` part is parsed elsewhere which might be a bit confusing, might as well leave it inline i figure

lovasoa: i inlined it back

iffyio (approved):
LGTM! Thanks @lovasoa!
cc @alamb

PR #1765, opened 2025-03-11
Preserve MySQL-style `LIMIT <offset>, <limit>` syntax

mvzink:
We already parse it, but were rewriting it into standard `LIMIT <limit> OFFSET <offset>` syntax. Now we preserve the original syntax.


src/parser/mod.rs
```diff
@@ -10272,20 +10269,33 @@ impl<'a> Parser<'a> {
...
-                vec![]
+                None
+            };
+
+            let limit_clause = if let Some(offset) = limit_comma_offset {
+                Some(LimitClause::OffsetCommaLimit {
+                    offset,
+                    limit: limit.expect("Expected <limit> expression for LIMIT <offset>, <limit>"),
+                })
+            } else if limit.is_some() || offset.is_some() || limit_by.is_some() {
```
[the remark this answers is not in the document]
mvzink: I reorganized it even more because the loop approach was hard to reason about. And in truth, all it is accomplishing is trying to parse `OFFSET` both before and after trying to parse `LIMIT`. So I just... did that. It seems fine because parsing offset is the simplest part. Let me know if the new approach makes sense. Added some extra tests for cases like you are mentioning too.

src/parser/mod.rs
```diff
@@ -9490,6 +9490,55 @@ impl<'a> Parser<'a> {
...
+        let (limit, limit_by) = if self.parse_keyword(Keyword::LIMIT) {
+            let expr = self.parse_limit()?;
+
+            if self.dialect.supports_limit_comma()
+                && offset.is_none()
+                && expr.is_some() // ALL not supported with comma
+                && self.consume_token(&Token::Comma)
+            {
+                return Ok(Some(LimitClause::OffsetCommaLimit {
+                    offset: expr.unwrap(), // Just checked `expr.is_some()`
```
iffyio, line 9509: we can replace the unwrap with an explicit error?

src/parser/mod.rs
```diff
@@ -9490,6 +9490,55 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_optional_limit_clause(&mut self) -> Result<Option<LimitClause>, ParserError> {
```
iffyio, line 9493:
```suggestion
   fn parse_optional_limit_clause(&mut self) -> Result<Option<LimitClause>, ParserError> {
```

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

alamb:
Love it -- thank you @mvzink and @iffyio

PR #1769, opened 2025-03-18
Parse `SUBSTR` as alias for `SUBSTRING`

mvzink:


src/parser/mod.rs
```diff
@@ -2403,7 +2404,7 @@ impl<'a> Parser<'a> {
         }
     }
 
-    pub fn parse_substring_expr(&mut self) -> Result<Expr, ParserError> {
+    pub fn parse_substring_expr(&mut self, shorthand: bool) -> Result<Expr, ParserError> {
```
iffyio, line 2407: would it make sense to change this function to include parsing the keyword? thinking in order to avoid the flag as a parameter and it would better suit the function definition to avoid the split parsing
```rust
    pub fn parse_substring_expr(&mut self, shorthand: bool) -> Result<Expr, ParserError> {
         let shorthand = self.parse_one_of_keywords(SUBSTR SUBSTRING)   {
         };
         // ...
    }
```
on the caller side would do
```rust
Keyword::SUBSTR | Keyword::SUBSTRING => {
     self.prev_token()
     self.parse_substring_expr()
}
```

PR #1774, opened 2025-03-24
Support qualified column names in `MATCH AGAINST` clause

tomershaniii:


iffyio (approved):
LGTM! Thanks @tomershaniii!
cc @alamb

PR #1779, opened 2025-03-27
Mysql: Add support for := operator

barsela1:
add support for := operator-mysql
MySQL dialect supports the := operator to assign values to user-defined variables in expressions, typically used in SELECT or UPDATE queries.
see https://dev.mysql.com/doc/refman/8.4/en/assignment-operators.html#operator_assign-value


iffyio (approved):
LGTM! Thanks @barsela1!
cc @alamb

PR #1780, opened 2025-03-28
Snowflake: Add support for `CONNECT_BY_ROOT`

tomershaniii:
Support for the snowflake CONNECT_BY_ROOT operator
https://docs.snowflake.com/en/sql-reference/constructs/connect-by

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -888,6 +888,12 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_TABLE_FACTOR
     }
 
+    /// Returns reserved keywords that may prefix a select item expression
+    /// e.g. `SELECT CONNECT_BY_ROOT name FROM Tbl2` (Snowflake)
+    fn get_reserved_keywords_for_select_item_operator(&self) -> &[Keyword] {
+        &[]
+    }
+
     /// Returns true if this dialect supports the `TABLESAMPLE` option
     /// before the table alias option. For example:
     ///
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -44,6 +44,7 @@ use alloc::{format, vec};
 use super::keywords::RESERVED_FOR_IDENTIFIER;
 use sqlparser::ast::StorageSerializationPolicy;
 
+const RESERVED_KEYWORDS_FOR_SELECT_ITEM_OPERATOR: [Keyword; 1] = [Keyword::CONNECT_BY_ROOT];
 /// A [`Dialect`] for [Snowflake](https://www.snowflake.com/)
 #[derive(Debug, Default)]
 pub struct SnowflakeDialect;
@@ -345,6 +346,11 @@ impl Dialect for SnowflakeDialect {
     /// See: <https://docs.snowflake.com/en/sql-reference/constructs/group-by>
     fn supports_group_by_expr(&self) -> bool {
         true
+    }
+
+    /// See: <https://docs.snowflake.com/en/sql-reference/constructs/connect-by>
+    fn get_reserved_keywords_for_select_item_operator(&self) -> &[Keyword] {
+        &RESERVED_KEYWORDS_FOR_SELECT_ITEM_OPERATOR
     }
 }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -14026,10 +14026,10 @@ impl<'a> Parser<'a> {
                 .maybe_parse_select_item_alias()
                 .map(|alias| match alias {
                     Some(alias) => SelectItem::ExprWithAlias {
-                        expr: prefixed_expr(expr, prefix),
+                        expr: maybe_prefixed_expr(expr, prefix),
                         alias,
                     },
-                    None => SelectItem::UnnamedExpr(prefixed_expr(expr, prefix)),
+                    None => SelectItem::UnnamedExpr(maybe_prefixed_expr(expr, prefix)),
                 }),
         }
     }
@@ -15391,7 +15391,7 @@ impl<'a> Parser<'a> {
     }
 }
 
-fn prefixed_expr(expr: Expr, prefix: Option<Ident>) -> Expr {
+fn maybe_prefixed_expr(expr: Expr, prefix: Option<Ident>) -> Expr {
     if let Some(prefix) = prefix {
         Expr::Prefixed {
             prefix,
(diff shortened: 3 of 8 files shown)


src/parser/mod.rs
```diff
@@ -13732,6 +13749,10 @@ impl<'a> Parser<'a> {
 
     /// Parse a comma-delimited list of projections after SELECT
     pub fn parse_select_item(&mut self) -> Result<SelectItem, ParserError> {
+        let prefix = self
+            .maybe_parse(|parser| parser.parse_select_item_prefix_by_reserved_word())?
+            .flatten();
+
         match self.parse_wildcard_expr()? {
```
iffyio, line 13756: Thinking introducing the prefix field into the select item is a bit more invasive and would be nice to avoid if it makes sense. I noticed this use case is similar in representation to [IntroducedString](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/ast/mod.rs#L929-L934), maybe we can repurpose that one to be generic.

I'm thinking something like this?

changing that enum variant into
```rust
Expr::Prefixed { prefix: Ident, expr: Expr }
```

Then impl wise here we could wrap the `self.parse_wildcard_expr()` call with the logic to optionally parse the prefix

```rust
fn parse_select_item_expr() {
  let prefix = self.parse_one_of_keywords(self.dialect.get_reserved_keywords_for_select_item());

  let expr = self.parse_wildcard_expr()?

  if let Some(prefix) = prefix {
    Expr::Prefixed {
      prefix: Ident::new(prefix),
      expr,
    }
  } else {
    expr
  }
}
```

src/dialect/mod.rs
```diff
@@ -888,6 +888,12 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_TABLE_FACTOR
     }
 
+    /// Returns reserved keywords for projection item prefix operator
+    /// e.g. SELECT CONNECT_BY_ROOT name FROM Tbl2 (Snowflake)
```
iffyio, line 892:
```suggestion
    /// e.g. `SELECT CONNECT_BY_ROOT name FROM Tbl2` (Snowflake)
```

src/dialect/mod.rs
```diff
@@ -888,6 +888,12 @@ pub trait Dialect: Debug + Any {
...
+    fn get_reserved_keywords_for_select_item_operator(&self) -> &[Keyword] {
```
iffyio, line 893:
```suggestion
    fn get_reserved_keywords_for_select_item(&self) -> &[Keyword] {
```

src/dialect/mod.rs
```diff
@@ -888,6 +888,12 @@ pub trait Dialect: Debug + Any {
         keywords::RESERVED_FOR_TABLE_FACTOR
     }
 
+    /// Returns reserved keywords for projection item prefix operator
```
iffyio, line 891:
```suggestion
    /// Returns reserved keywords that may prefix a select item expression
```

src/dialect/snowflake.rs
```diff
@@ -346,6 +346,13 @@ impl Dialect for SnowflakeDialect {
     fn supports_group_by_expr(&self) -> bool {
         true
     }
+
+    /// See: <https://docs.snowflake.com/en/sql-reference/constructs/connect-by>
+    fn get_reserved_keywords_for_select_item_operator(&self) -> &[Keyword] {
+        const RESERVED_KEYWORDS_FOR_SELECT_ITEM_OPERATOR: [Keyword; 1] = [Keyword::CONNECT_BY_ROOT];
```
iffyio, line 352: could we move this to the top of the file maybe? thinking it would be more visible there and if the list happens to get long it doesnt add to the function's length

tests/sqlparser_postgres.rs
```diff
@@ -4911,17 +4965,17 @@ fn parse_array_agg() {
...
+    // // handles multi-part identifier with general code path
+    // let sql3 = "SELECT GREATEST(my_schema.sections_tbl.*) AS sections FROM sections_tbl";
+    // pg().verified_stmt(sql3);
 
-    // handles multi-part identifier with array_agg code path
-    let sql4 = "SELECT ARRAY_AGG(my_schema.sections_tbl.*) AS sections FROM sections_tbl";
-    pg().verified_stmt(sql4);
+    // // handles multi-part identifier with array_agg code path
+    // let sql4 = "SELECT ARRAY_AGG(my_schema.sections_tbl.*) AS sections FROM sections_tbl";
+    // pg().verified_stmt(sql4);
```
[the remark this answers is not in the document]
tomershaniii: yes, missed that

iffyio:
Marking as draft in the meantime as this PR is no longer pending review. @tomershaniii please feel free to undraft and ping when ready!

tomershaniii:
Hi @iffyio , Modified the PR per you suggestion, please see if this is aligned to what you have in mind.

src/parser/mod.rs
```diff
@@ -15375,6 +15391,17 @@ impl<'a> Parser<'a> {
     }
 }
 
+fn prefixed_expr(expr: Expr, prefix: Option<Ident>) -> Expr {
```
iffyio, line 15394:
```suggestion
fn maybe_prefixed_expr(expr: Expr, prefix: Option<Ident>) -> Expr {
```
thinking in order to flag that the prefixing is optional

tests/sqlparser_snowflake.rs
```diff
@@ -3983,3 +3983,38 @@ fn test_nested_join_without_parentheses() {
...
+        Statement::Query(query) => {
+            assert_eq!(
+                query.body.as_select().unwrap().projection[0],
+                SelectItem::UnnamedExpr(Expr::Prefixed {
+                    prefix: Ident::new("CONNECT_BY_ROOT"),
+                    value: Box::new(Expr::Identifier(Ident::new("name")))
+                })
+            );
+        }
+        _ => unreachable!(),
```
iffyio, line 4018: can we add a test case for e.g. `SELECT connect_by_root FROM T`? i.e. one that uses the keyword as the projected column name

iffyio (commented):
Thanks @tomershaniii! I think this looks good overall to me, left some minor comments

tomershaniii:
Hi @iffyio , please see latest commit.

iffyio (approved):
LGTM! Thanks @tomershaniii!
cc @alamb

PR #1782, opened 2025-03-30
Add cipherstash-proxy to list of users in README.md

coderdan:
Added cipherstash proxy to list of users


iffyio (approved):
Thanks @coderdan!
cc @alamb

PR #1783, opened 2025-03-30
MySQL create table options: optional DEFAULT keyword, CHARACTER SET longhand

MohamedAbdeen21:
As per [MySQL docs](https://dev.mysql.com/doc/refman/8.4/en/create-table.html) both `DEFAULT` and `=` are optional for `CHARACTER SET` and `COLLATE` options.

Also, CHARSET is an alias for CHARACTER SET, so both should be parsed.

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -948,6 +948,36 @@ fn parse_create_table_engine_default_charset() {
 }
 
 #[test]
+fn parse_create_table_charset_collate() {
+    let sql =
+        "CREATE TABLE foo (id INT(11)) CHARACTER SET utf8mb3 DEFAULT COLLATE utf8mb4_0900_ai_ci";
+    let expected =
+        "CREATE TABLE foo (id INT(11)) DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb4_0900_ai_ci";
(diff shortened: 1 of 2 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -947,6 +947,36 @@ fn parse_create_table_engine_default_charset() {
     }
 }
 
+#[test]
+fn parse_create_table_charset_collate() {
+    let sql =
+        "CREATE TABLE foo (id INT(11)) CHARACTER SET utf8mb3 DEFAULT COLLATE utf8mb4_0900_ai_ci";
+    let expected =
+        "CREATE TABLE foo (id INT(11)) DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb4_0900_ai_ci";
+    match mysql().one_statement_parses_to(sql, expected) {
```
MohamedAbdeen21, line 956: If we want to switch this to use `verified_stmt` instead, we will need to replace both collate and charset fields in `CreateTable` with structs to capture the exact syntax used (was default keyword and/or assignment operator used? shorthand `CHARSET` instead of `CHARACTER SET`?).

MohamedAbdeen21:
Closed in favor of #1747

Issue #1775, opened 2025-03-24
Snowflake ALTER SESSION cannot be followed by other statements.


romanb:
It seems that Snowflake `ALTER SESSION` statements cannot currently be followed by other statements. For example:
```sql
ALTER SESSION SET QUERY_TAG='tag'; SELECT 42
```
yields
```
Expected: another option, found: SELECT at Line: 1, Column: 36
```

PR #1786, opened 2025-03-31
Fix: Snowflake ALTER SESSION cannot be followed by other statements.

romanb:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1775.

Currently `parse_session_options` for Snowflake does not check for semicolons, which makes it impossible to parse multiple statements, where another statement comes after an `ALTER SESSION` statement.


iffyio (approved):
LGTM! Thanks @romanb!
cc @alamb

PR #1787, opened 2025-03-31
Enable double-dot-notation for mssql.

romanb:
Since the schema name is optional in MSQL table identifiers, double-dot notation is supported, in which case the default dbo schema is used. This PR enables double-dot notation for the MSSQL dialect.


iffyio (approved):
Thanks @romanb!
cc @alamb

PR #1788, opened 2025-04-01
Add GreptimeDB to the "Users" in README

MichaelScofield:
GreptimeDB uses sqlparser in its "RANGE" syntax, and various other places.


iffyio (approved):
Thanks @MichaelScofield!
cc @alamb

PR #1790, opened 2025-04-01
MSSQL: Add support for functionality `MERGE` output clause

### LINKED ISSUE #1789: Add support for OUTPUT Clause in TSQL (MS SQL) Merge statements
Add the end of a merge statement in T-SQL we can use an OUTPUT clause to write the changed rows to a table variable. This is often used in Data Warehouse workloads were we would like to track slowly changing dimensions. 

A merge statement with the output clause could look like this:
```sql

MERGE dso.products AS t
USING dsi.products AS s
	ON s.ProductID = t.ProductID
WHEN MATCHED 
	AND NOT (t.[ProductName] = s.[ProductName] OR (ISNULL(t.[ProductName], s.[ProductName]) IS NULL))
THEN
	UPDATE SET t.[ProductName] = s.[ProductName]

WHEN NOT MATCHED BY TARGET THEN
INSERT ([ProductID]
           ,[ProductName])
     VALUES
           (s.[ProductID]
           ,s.[ProductName])

WHEN NOT MATCHED BY SOURCE  THEN
	DELETE

OUTPUT $action, deleted.* INTO dsi.temp_products;
```

Currently this fails with the error `ParserError("Expected: WHEN, found: OUTPUT at Line: 112, Column: 1")`.

I will try to create a PR solving this issue, but any pointers on a good approach would be highly appreciated. 

Thanks for the awesome crate btw.

dilovancelik:
Hey I have added a feature to handle OUTPUT Statements in the end of merge statements, which is used in MS SQL fixes #1789  that I created. 

- I've added the OUTPUT Keyword
- I've added a struct in for Output in ast/mod.rs
- I've added an optional Output parameter to the Merge struct in ast/mod.rs
- added function for parsing the output parameter in parser/mod.rs
- updated parse merge clause function, to break on output keyword
- updated parse merge to handling of the output clause
- added and updated tests

This is my first Pull Request to an open source project so please let me know, if there is something missing, or something I did wrong. 

Thank you for a great crate.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -5408,8 +5408,8 @@ impl fmt::Display for Statement {
                 table,
                 source,
                 on,
-                clauses, 
-                output
+                clauses,
+                output,
             } => {
                 write!(
                     f,
@@ -7924,7 +7924,10 @@ pub struct Output {
 
 impl fmt::Display for Output {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let Output { select_items, into_table } = self;
+        let Output {
+            select_items,
+            into_table,
+        } = self;
 
         write!(
             f,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -7914,7 +7914,6 @@ impl Display for MergeClause {
 /// Example:
 /// OUTPUT $action, deleted.* INTO dbo.temp_products;
 /// [mssql](https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql)
-
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -5419,9 +5419,8 @@ impl fmt::Display for Statement {
                 )?;
                 write!(f, "ON {on} ")?;
                 write!(f, "{}", display_separated(clauses, " "))?;
-                if output.is_some() {
-                    let out = output.clone().unwrap();
-                    write!(f, " {out}")?;
+                if let Some(output) = output {
+                    write!(f, " {output}")?;
                 }
                 Ok(())
             }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -662,17 +662,17 @@ pub enum Expr {
     /// such as maps, arrays, and lists:
     /// - Array
     ///     - A 1-dim array `a[1]` will be represented like:
-    ///         `CompoundFieldAccess(Ident('a'), vec![Subscript(1)]`
+    ///       `CompoundFieldAccess(Ident('a'), vec![Subscript(1)]`
     ///     - A 2-dim array `a[1][2]` will be represented like:
-    ///         `CompoundFieldAccess(Ident('a'), vec![Subscript(1), Subscript(2)]`
+    ///       `CompoundFieldAccess(Ident('a'), vec![Subscript(1), Subscript(2)]`
     /// - Map or Struct (Bracket-style)
     ///     - A map `a['field1']` will be represented like:
-    ///         `CompoundFieldAccess(Ident('a'), vec![Subscript('field')]`
+    ///       `CompoundFieldAccess(Ident('a'), vec![Subscript('field')]`
     ///     - A 2-dim map `a['field1']['field2']` will be represented like:
-    ///         `CompoundFieldAccess(Ident('a'), vec![Subscript('field1'), Subscript('field2')]`
+    ///       `CompoundFieldAccess(Ident('a'), vec![Subscript('field1'), Subscript('field2')]`
     /// - Struct (Dot-style) (only effect when the chain contains both subscript and expr)
     ///     - A struct access `a[field1].field2` will be represented like:
-    ///         `CompoundFieldAccess(Ident('a'), vec![Subscript('field1'), Ident('field2')]`
+    ///       `CompoundFieldAccess(Ident('a'), vec![Subscript('field1'), Ident('field2')]`
     /// - If a struct access likes `a.field1.field2`, it will be represented by CompoundIdentifier([a, field1, field2])
     CompoundFieldAccess {
         root: Box<Expr>,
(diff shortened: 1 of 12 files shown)


src/parser/mod.rs
```diff
@@ -14586,6 +14589,29 @@ impl<'a> Parser<'a> {
         Ok(clauses)
     }
 
+    pub fn parse_output(&mut self) -> Result<Output, ParserError> {
```
iffyio, line 14592:
```suggestion
    fn parse_output(&mut self) -> Result<Output, ParserError> {
```

src/parser/mod.rs
```diff
@@ -14489,7 +14489,10 @@ impl<'a> Parser<'a> {
     pub fn parse_merge_clauses(&mut self) -> Result<Vec<MergeClause>, ParserError> {
         let mut clauses = vec![];
         loop {
-            if self.peek_token() == Token::EOF || self.peek_token() == Token::SemiColon {
+            if self.peek_token() == Token::EOF
+                || self.peek_token() == Token::SemiColon
+                || self.peek_keyword(Keyword::OUTPUT)
```
iffyio, line 14494: hmm I wonder in order to avoid accumulating conditions here over time, would it make sense to change to something like this? replacing the EOF, Semicolon and OUTPUT conditions
```rust
loop {
      if !self.parse_keyword(Keyword::WHEN) {
            break
      }
      // ...
}
```

src/parser/mod.rs
```diff
@@ -14586,6 +14589,29 @@ impl<'a> Parser<'a> {
...
+            .is_some();
+        let unlogged = self.parse_keyword(Keyword::UNLOGGED);
+        let table = self.parse_keyword(Keyword::TABLE);
+        let name = self.parse_object_name(false)?;
+        let into_table = SelectInto {
+            temporary,
+            unlogged,
+            table,
+            name,
+        };
```
iffyio, line 14607: This looks to be the same code as in `parse_select`? could we factor that into a function like `parse_into_clause()` that could be reused in both places?

tests/sqlparser_mssql.rs
```diff
@@ -1921,3 +1921,9 @@ fn ms() -> TestedDialects {
 fn ms_and_generic() -> TestedDialects {
     TestedDialects::new(vec![Box::new(MsSqlDialect {}), Box::new(GenericDialect {})])
 }
+
+#[test]
+fn parse_mssql_merge_with_output() {
+    let stmt = "MERGE dso.products AS t USING dsi.products AS s ON s.ProductID = t.ProductID WHEN MATCHED AND NOT (t.ProductName = s.ProductName OR (ISNULL(t.ProductName, s.ProductName) IS NULL)) THEN UPDATE SET t.ProductName = s.ProductName WHEN NOT MATCHED BY TARGET THEN INSERT (ProductID, ProductName) VALUES (s.ProductID, s.ProductName) WHEN NOT MATCHED BY SOURCE THEN DELETE OUTPUT $action, deleted.ProductID INTO dsi.temp_products";
```
iffyio, line 1927: to help with readability could we use line breaks for formatting here similar to the `test_merge_with_output` test case in?

src/ast/mod.rs
```diff
@@ -3828,6 +3829,8 @@ pub enum Statement {
         on: Box<Expr>,
         /// Specifies the actions to perform when values match or do not match.
         clauses: Vec<MergeClause>,
+        // Speccifies the output to save changes in MSSQL
```
iffyio, line 3832:
```suggestion
        // Specifies the output to save changes in MSSQL
```

iffyio (commented):
Thanks @dilovancelik! The changes look reasonable to me overall, left some comments

tests/sqlparser_mssql.rs
dilovancelik: Yeah that makes a lot of sense. Its should be updated to be multiline now

src/parser/mod.rs
dilovancelik: I definitely see your point. I implemented and had to remove the expect_keyword_is below since the keyword is consumed in the check

src/parser/mod.rs
dilovancelik: Yeah that makes sense. I made it, it's own function, which I called parse_select_into, since it returns the SelectInto structure, that way its consistent

src/parser/mod.rs
```diff
@@ -14586,6 +14585,34 @@ impl<'a> Parser<'a> {
...
+        self.expect_keyword_is(Keyword::INTO)?;
+        let into_table = self.parse_select_into()?;
+
+        Ok(OutputClause {
+            select_items,
+            into_table,
+        })
+    }
+
+    fn parse_select_into(&mut self) -> Result<SelectInto, ParserError> {
```
iffyio, line 14600: oh could we also update [this usage](https://github.com/apache/datafusion-sqlparser-rs/blob/01c7db690c01a84a5cf5d29445d3149889dbd02a/src/parser/mod.rs#L10913-L10924) to call this `parse_select_into()` function as well?

iffyio (commented):
left a comment to reuse the new select_into function, beyond that this looks good to me!

src/parser/mod.rs
dilovancelik: good idea, I've changed it to use the new function!

dilovancelik:
I updated the old usage to use the new parse_select_into function and ran all tests :)

iffyio (approved):
LGTM! Thanks @dilovancelik!
cc @alamb

iffyio:
@dilovancelik when you get the chance to could you merge in latest from main, in order to fix the CI issue?

dilovancelik:
should be added now :)

iffyio:
@dilovancelik Oh could you take a look at the added commits, not sure how the merge from main was done but there's some extraneous diffs that's been brought into the PR

dilovancelik:
Hey I did a rebase, and I think something went wrong, now I just did a classic merge, and et looks like the extra commits are gone. Sorry.

iffyio:
Oh yes so that failure is my bad and came from the last PR that landed on main yesterday, it fails on rust 1.86. To be clear its not related to this PR but its possible to resolve, @dilovancelik would you be able to resolve the lint failure in this PR and I can merge both in the same go? Sorry for the hassle

dilovancelik:
No worries, I just couldn't get the error locally so couldn't fix. That however was because my main was one commit behind. I synced and removed the borrow which threw the error. 

Again thanks for all the feedback, this has been a really pleasant first OS experience

iffyio:
Awesome thanks again @dilovancelik!

PR #1793, opened 2025-04-02
Add support for parenthesized subquery as `IN` predicate

### LINKED ISSUE #1792: PostgreSQL dialect fails to parse `IN` with union
Minimal example:

```sql
SELECT 1 FROM t WHERE t.id IN ((SELECT id FROM t) UNION (SELECT id FROM t))
```

Currently this fails with:

```rs
ParserError("Expected: ), found: UNION at Line: 1, Column: 51")
```

adamchainz:
Fixes #1792.

This is my attempt to hack in support with my minimal Rust skills. I stopped when I realized that my change of `InSubquery` to use `SetExpr` for its subquery broke many tests.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3742,15 +3742,13 @@ impl<'a> Parser<'a> {
             });
         }
         self.expect_token(&Token::LParen)?;
-        let in_op = if self.parse_keyword(Keyword::SELECT) || self.parse_keyword(Keyword::WITH) {
-            self.prev_token();
-            Expr::InSubquery {
+        let in_op = match self.maybe_parse(|p| p.parse_query_body(p.dialect.prec_unknown()))? {
+            Some(subquery) => Expr::InSubquery {
                 expr: Box::new(expr),
-                subquery: self.parse_query()?,
+                subquery,
                 negated,
-            }
-        } else {
-            Expr::InList {
+            },
+            None => Expr::InList {
                 expr: Box::new(expr),
                 list: if self.dialect.supports_in_empty_list() {
                     self.parse_comma_separated0(Parser::parse_expr, Token::RParen)?
@@ -3758,7 +3756,7 @@ impl<'a> Parser<'a> {
                     self.parse_comma_separated(Parser::parse_expr)?
                 },
                 negated,
-            }
+            },
         };
         self.expect_token(&Token::RParen)?;
         Ok(in_op)
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -2224,7 +2224,21 @@ fn parse_in_subquery() {
     assert_eq!(
         Expr::InSubquery {
             expr: Box::new(Expr::Identifier(Ident::new("segment"))),
-            subquery: Box::new(verified_query("SELECT segm FROM bar")),
+            subquery: verified_query("SELECT segm FROM bar").body,
+            negated: false,
+        },
+        select.selection.unwrap()
+    );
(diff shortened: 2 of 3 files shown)


src/parser/mod.rs
```diff
@@ -3744,10 +3744,18 @@ impl<'a> Parser<'a> {
...
+                    subquery: self.parse_query_body(self.dialect.prec_unknown())?,
+                    negated,
+                }
+            } else {
+                Expr::InSubquery {
+                    expr: Box::new(expr),
+                    subquery: Box::new(SetExpr::Query(self.parse_query()?)),
+                    negated,
+                }
             }
```
iffyio, line 3759: would it make sense to skip the `supports_in_query_body` flag entirely? thinking since the new representation `SetExpr` is a superset of the old query

adamchainz: I am not sure since that would allow parsing this syntax on dialects that don't actually support it, right?

iffyio: Ah yeah I think that would be fine to have the parser always accept the syntax

adamchainz: Okay, done, and fixed to work in all cases!

src/parser/mod.rs
```diff
@@ -3742,24 +3742,23 @@ impl<'a> Parser<'a> {
             });
         }
         self.expect_token(&Token::LParen)?;
-        let in_op = if self.parse_keyword(Keyword::SELECT) || self.parse_keyword(Keyword::WITH) {
-            self.prev_token();
-            Expr::InSubquery {
+        let in_op = match self.try_parse(|p| p.parse_query_body(p.dialect.prec_unknown())) {
```
adamchainz, line 3745: The previous approach peeked for a `SELECT` or `WITH`, but we cannot rely on that because there may be more than one parenthesis, as in `IN ((SELECT ...) UNION ())`.

The new approach tries to parse a subquery first, then an in list second.

tests/sqlparser_common.rs
```diff
@@ -2224,7 +2224,21 @@ fn parse_in_subquery() {
...
-            subquery: Box::new(verified_query("SELECT segm FROM bar")),
+            subquery: verified_query("SELECT segm FROM bar").body,
+            negated: false,
+        },
+        select.selection.unwrap()
+    );
+}
+
+#[test]
+fn parse_in_union() {
```
adamchainz, line 2235: I only added one test - happy to add more if deemed necessary.

src/parser/mod.rs
iffyio: Yeah this makes sense!

iffyio, line 3745: Can we use `self.maybe_parse()` instead of `try_parse()`? that would automatically handle the recursion exceeded case for example

iffyio (commented):
Thanks @adamchainz! This looks good to me overall, left one comment

iffyio:
Marking as draft in the meantime as this is no longer pending review, @adamchainz please feel free to undraft and ping when ready!

src/parser/mod.rs
adamchainz: Done! Thank you, that is nicer.

iffyio (approved):
LGTM! Thanks @adamchainz!
cc @alamb

adamchainz:
Yay, thank you!

alamb:
The code train keeps on moving in this repo. very impressive

adamchainz:
image

choo choo!

PR #1794, opened 2025-04-02
Extend snowflake grant options support

yoavcloud:
This PR adds parsing support for more Snowflake GRANT statement options, specifically, privileges on account global objects such as connections, external volumes, etc.


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1795, opened 2025-04-02
Allow single quotes in EXTRACT() for Redshift.

romanb:
Just like Postgres, Redshift supports enclosing the `datepart` parameter in single quotes. See also the examples in https://docs.aws.amazon.com/redshift/latest/dg/r_EXTRACT_function.html.


alamb:
the code keeps flowing -- thanks @iffyio and @romanb

PR #1799, opened 2025-04-06
Snowflake: support nested join without parentheses

barsela1:


src/parser/mod.rs
```diff
@@ -11823,7 +11828,16 @@ impl<'a> Parser<'a> {
...
-                let relation = self.parse_table_factor()?;
+                let mut relation = self.parse_table_factor()?;
+
+                if self.is_next_token_a_join() {
+                    let joins = self.parse_joins()?;
+                    relation = TableFactor::NestedJoin {
+                        table_with_joins: Box::new(TableWithJoins { relation, joins }),
+                        alias: None,
+                    };
+                }
```
iffyio, line 11839: would it make more sense to have this logic as part of the table factor parsing instead? thinking that would be by definition, and so that any caller of `self.parse_table_factor` gets the full behavior including nested joins.

i imagine essentially to split table factor parsing instead e.g.

```rust
fn parse_table_factor() {
    let table_factor = self.parse_table_factor_inner()?;
    if self.is_next_token_a_join() {
        // ...
   }
}
```

src/parser/mod.rs
```diff
@@ -11833,7 +11847,21 @@ impl<'a> Parser<'a> {
             };
             joins.push(join);
         }
-        Ok(TableWithJoins { relation, joins })
+        Ok(joins)
+    }
+
+    fn is_next_token_a_join(&self) -> bool {
```
iffyio, line 11853:
```suggestion
    fn peek_parens_less_nested_join(&self) -> bool {
```
maybe something like this naming wise? peek uses existing parser terminology. Also would it be possible with a doc comment of the function mentioning the use case? I think that would be helpful to future readers

src/parser/mod.rs
```diff
@@ -11833,7 +11847,21 @@ impl<'a> Parser<'a> {
...
+        matches!(
+            self.peek_token().token,
```
iffyio, line 11855:
```suggestion
            self.peek_token_ref().token,
```

tests/sqlparser_snowflake.rs
```diff
@@ -3516,3 +3516,87 @@ fn test_alter_session_followed_by_statement() {
         _ => panic!("Unexpected statements: {:?}", stmts),
     }
 }
+
+#[test]
+fn test_nested_join_without_parentheses() {
+    let query = "SELECT DISTINCT p.product_id FROM orders AS o INNER JOIN customers AS c INNER JOIN products AS p ON p.customer_id = c.customer_id ON c.order_id = o.order_id";
```
iffyio, line 3522: Can we add coverage for all the supported join types?

tests/sqlparser_snowflake.rs
```diff
@@ -3516,3 +3516,87 @@ fn test_alter_session_followed_by_statement() {
...
+    assert_eq!(
+        only(
+            snowflake()
+                .verified_only_select_with_canonical(query, "")
```
iffyio, line 3526: oh what does the empty string second argument imply?

barsela1: I want to skip the canonical comparison because parentheses were added there and it fails the test even though it is correct. Is there a better way to overcome this?

src/parser/mod.rs
barsela1: Can you please explain the example you brought? i didn't fully understood it

iffyio: Ah yeah so what I meant was that the current `parse_table_factor` gets renamed to something like `parse_table_factor_inner`, then the actual impl `parse_table_factor`, after parsing a lone `TableFactor` here
```rust
    let table_factor = self.parse_table_factor_inner()?;
```
it next peeks if this table factor is actually a nested join, similar to the current PR impl, and if it is, completes the parsing as a nested table factor
```rust
    let mut table_factor = self.parse_table_factor_inner()?;

    if self.peek_parens_less_nested_join() {
        let joins = self.parse_joins()?;
        table_factor = TableFactor::NestedJoin {
            table_with_joins: Box::new(TableWithJoins { relation: table_factor, joins }),
            alias: None,
        };
    }
    
    Ok(table_factor)
```

would this be reasonable?

barsela1: I’m not sure this approach is correct...test_join() test fails.
In parse_table_and_joins, there's an initial call to parse_table_factor that is unrelated to join parsing.
With this implementation, the first table will always be parsed as a NestedJoin, even when it shouldn’t be.
I tried to handle nested joins explicitly within the join logic to avoid that

iffyio: Ah I see, that sounds reasonable then thanks!

iffyio (approved):
LGTM! Thanks @barsela1!
cc @alamb

PR #1801, opened 2025-04-09
Allow literal backslash escapes for string literals in Redshift dialect.

romanb:
Redshift supports the literal backslash escape at least for a subset of characters, including single quotes ('), for example
```sql
SELECT 'l\'auto'
```
This is unfortunately not publicly documented (yet?) but used in successful queries.


iffyio (approved):
LGTM! Thanks @romanb!
cc @alamb

PR #1802, opened 2025-04-09
Add support for MySQL's STRAIGHT_JOIN join operator.

romanb:
This PR adds support for parsing the MySQL `STRAIGHT_JOIN` join operator.

See https://dev.mysql.com/doc/refman/8.4/en/join.html for reference.


src/ast/query.rs
```diff
@@ -2197,6 +2200,8 @@ pub enum JoinOperator {
         match_condition: Expr,
         constraint: JoinConstraint,
     },
+    /// STRAIGHT_JOIN (non-standard)
```
iffyio, line 2203: Could we include a link to the mysql docs for reference?

romanb: Of course, done!

iffyio (approved):
LGTM! Thanks @romanb!
cc @alamb

PR #1805, opened 2025-04-10
Snowflake COPY INTO target columns, select items and optional alias

yoavcloud:
This PR adds support for the following in Snowflake's COPY INTO:
1. Specifying a list of columns in the destination table:
```sql
COPY INTO [<namespace>.]<table_name> [ ( <col_name> [ , <col_name> ... ] ) ]
```
2. Using regular SQL expressions in the data transformation query, typically for constant or computed values:
```sql
SELECT t1.$1:st AS st, $1:index, t2.$1, 4, '5' AS const_str FROM @schema.general_finished
```
3. The `AS` keyword is optional when specifying an alias for the stage in the data transformation query
```sql
COPY INTO tbl FROM (SELECT t1.$1:st AS st, 4, '5' AS const_str FROM @stage S)
```

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -878,14 +878,10 @@ fn parse_select_items_for_data_load(
                 parser.parse_select_item()?,
             )),
         }
-        match parser.next_token().token {
-            Token::Comma => {
-                // continue
-            }
-            _ => {
-                parser.prev_token(); // need to move back
-                break;
-            }
+        if matches!(parser.peek_token_ref().token, Token::Comma) {
+            parser.advance_token();
+        } else {
+            break;
         }
     }
     Ok(Some(select_items))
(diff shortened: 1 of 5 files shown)


src/dialect/snowflake.rs
```diff
@@ -866,86 +867,97 @@ pub fn parse_copy_into(parser: &mut Parser) -> Result<Statement, ParserError> {
...
+        }
+        match parser.next_token().token {
+            Token::Comma => {
+                // continue
+            }
+            _ => {
+                parser.prev_token(); // need to move back
+                break;
+            }
+        }
```
iffyio, line 889: could we do something like this in order to avoid the token clone here?
```rust
if matches!(parser.peek_token_ref(), Token::Comma) {
    parser.advance()
} else {
   break
}
```

yoavcloud: Great idea! I included it in the latest commit.

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

Issue #1804, opened 2025-04-10
Missing support for `INHERITS` operation from PostgreSQL


LucaCappelletti94:
Hi all,

It looks like `sqlparser` does not currently support the [`INHERITS` operation](https://www.postgresql.org/docs/current/ddl-inherit.html).

```sql
CREATE TABLE child_table (child_column INT) INHERITS (parent_table);
```

as I get the following error:

```
CREATE TABLE child_table (child_column INT) INHERITS (parent_table);: ParserError("Expected: end of statement, found: INHERITS")
```

I will try to add support for this operation in a PR when I get the time, unless you tell me that for some reason you do not intend to support this syntax.

Best,
Luca

LucaCappelletti94:
Closing issue as relevant pull request has been merged.

PR #1806, opened 2025-04-10
Add support for `INHERITS` option in `CREATE TABLE` statement

LucaCappelletti94:
This pull request will add support for parsing the `INHERITS` keyword in the `CREATE TABLE` statements.

Solves issue #1804


src/parser/mod.rs
```diff
@@ -7070,13 +7071,22 @@ impl<'a> Parser<'a> {
...
+    /// Parse configuration like inheritance, partitioning, clustering information during the table creation.
     ///
     /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_2)
-    /// [PostgreSQL](https://www.postgresql.org/docs/current/ddl-partitioning.html)
+    /// [PostgreSQL Partitioning](https://www.postgresql.org/docs/current/ddl-partitioning.html)
+    /// [PostgreSQL Inheritance](https://www.postgresql.org/docs/current/ddl-inherit.html)
     fn parse_optional_create_table_config(
         &mut self,
     ) -> Result<CreateTableConfiguration, ParserError> {
+        let inherits = if dialect_of!(self is BigQueryDialect | PostgreSqlDialect | GenericDialect)
```
iffyio, line 7082: the `dialect_of` macro we're transitioning away from for new code, but I think in this case there's probably no need to gate this on dialect (also not sure BQ supports this syntax so it might be misleading to explicitly name BQ here) so that we can remove the check entirely and always parse the `INHERIT` option if it shows up in the statement?

tests/sqlparser_postgres.rs
```diff
@@ -2733,6 +2733,39 @@ fn parse_create_brin() {
...
+            assert_eq_vec(&["pg_catalog", "pg_settings"], &inherits[1].0);
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+#[should_panic]
+fn parse_create_table_with_empty_inherits_fails() {
+    pg().verified_stmt("CREATE TABLE child_table (child_column INT) INHERITS ()");
```
iffyio, line 2766: in place of the `should_panic`, we can rewrite the test case [in this form](https://github.com/apache/datafusion-sqlparser-rs/blob/165164bfc2d19fc61b3caec3fec2265c9173766a/tests/sqlparser_postgres.rs#L307-L310)
It would let us include the test case in the existing test function `parse_create_table_with_inherits` to help with grouping related test cases

src/parser/mod.rs
LucaCappelletti94: The dialect thing must have been a copy paste from the line below and then my brain decided to drop that pointer of information - fixed as described!

tests/sqlparser_postgres.rs
LucaCappelletti94: Refactored test as described!

iffyio (approved):
Thanks @LucaCappelletti94!
cc @alamb

Issue #1800, opened 2025-04-07
SQL Server: Increase support for multi statement object definitions


aharpervc:
It would be extremely useful if this library was able to properly parse multiple statements, particularly in a object definition body, without requiring semi-colons which are optional for SQL Server.

---

The following otherwise valid SQL fails parsing (MSSQL dialect):

```mssql
create or alter procedure test()
as
begin
    declare @x bit = 1

    if @x = 1
    begin
        select 1
    end
end
```

The error is at the `if @x = 1` line:

```
end of statement, found: if
```

The problem appears to be related to parsing multiple statements in the object definition body. The parsing concludes successfully if the code is changed to `declare @x bit = 1;` with a semi-colon.

There's also the similar case of two top level statements:

```mssql
declare @x bit = 1

if @x = 1
begin
    select 1
end
```

The API I was using was parse_sql:

```rust
static DIALECT: sqlparser::dialect::MsSqlDialect = sqlparser::dialect::MsSqlDialect {};
let sql_text = ...

let mut statements = Parser::parse_sql(&DIALECT, sql_text.as_str())?; // error: end of statement, found: if
```

There's a workaround for the multiple top level statements scenario, but that won't work for the stored procedure example:

```rust
fn parse_all_statements(sql_text: &str) -> Result<Vec<Statement>, Box<dyn std::error::Error>> {
    let mut parser = Parser::new(&DIALECT)
        .try_with_sql(&sql_text)?;

    let mut statements = Vec::new();

    loop {
        if let Token::EOF = parser.peek_token_ref().token {
            break;
        }

        statements.push(parser.parse_statement()?);

        while let Token::SemiColon = parser.peek_token_ref().token {
            parser.advance_token();
        }
    }

    return Ok(statements);
}
```

---

follow up from: https://github.com/apache/datafusion-sqlparser-rs/pull/1791#issuecomment-2784202219

aharpervc:
The related PR in this repo went stale, but the branch on my fork is still functional with all tests passing, if anyone else is interested: https://github.com/aharpervc/datafusion-sqlparser-rs/tree/expand-parse-without-semicolons

PR #1808, opened 2025-04-10
Add `CREATE FUNCTION` support for SQL Server

aharpervc:
Partially related: https://github.com/apache/datafusion-sqlparser-rs/issues/1800

---

For reference: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16

In SQL Server, functions look a lot like procedures, with begin/end, multiple statements, and a return. To accommodate that here in the parser, `BeginEndStatement` was extracted from the conditional logic to be reused here, along with a `RETURN` statement type

<details>
Formerly, this was a new struct, but was later consolidated in with BeginEndStatement. 
I'm open to suggestions about how this could be improve (now or later). For example, since it's a BEGIN (statements) END construct, should this (and procedures) be using the statement block logic introduced here? https://github.com/apache/datafusion-sqlparser-rs/pull/1791 
</details>

Then also while I was working on this, I noticed that only SQL Server (so far...?) supports `OR ALTER`, so I introduced a new struct field to track that. (Again, very similar to `CREATE PROCEDURE` logic).

I'm not aiming to enable parsing every function type here in this PR -- just trying to get something operational, and the additional capabilities like table expressions, etc, can be added in later.

---

To continue the recent discussion, I noticed that both these new multi statement functions & the existing procedure parsing fails unless the statements are semi-colon delimited. Semi-colons are optional in SQL Server, so for example, this code executes fine for real but fails to parse here in this library, unless that "SET" ends with a `;`...

```mssql
CREATE OR ALTER PROCEDURE test (@foo INT, @bar VARCHAR(256))
AS
BEGIN
    SET @foo = @foo + 1
    SELECT @foo
END
```

I think this is because the procedure body parsing logic is using `parse_statements()`, which requires it? I'm game to tackle the problem with some guidance on what's a good approach to make it parse 🤔

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2157,6 +2157,10 @@ impl fmt::Display for ClusteredBy {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct CreateFunction {
+    /// True if this is a `CREATE OR ALTER FUNCTION` statement
+    ///
+    /// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#or-alter)
+    pub or_alter: bool,
     pub or_replace: bool,
     pub temporary: bool,
     pub if_not_exists: bool,
@@ -2219,9 +2223,10 @@ impl fmt::Display for CreateFunction {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "CREATE {or_replace}{temp}FUNCTION {if_not_exists}{name}",
+            "CREATE {or_alter}{or_replace}{temp}FUNCTION {if_not_exists}{name}",
             name = self.name,
             temp = if self.temporary { "TEMPORARY " } else { "" },
+            or_alter = if self.or_alter { "OR ALTER " } else { "" },
             or_replace = if self.or_replace { "OR REPLACE " } else { "" },
             if_not_exists = if self.if_not_exists {
                 "IF NOT EXISTS "
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2277,11 +2277,9 @@ impl fmt::Display for CreateFunction {
         if let Some(CreateFunctionBody::AsAfterOptions(function_body)) = &self.function_body {
             write!(f, " AS {function_body}")?;
         }
-        if let Some(CreateFunctionBody::MultiStatement(statements)) = &self.function_body {
+        if let Some(CreateFunctionBody::AsBeginEnd(bes)) = &self.function_body {
             write!(f, " AS")?;
-            write!(f, " BEGIN")?;
-            write!(f, " {}", display_separated(statements, "; "))?;
-            write!(f, " END")?;
+            write!(f, " {}", bes)?;
         }
         Ok(())
     }
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2278,8 +2278,7 @@ impl fmt::Display for CreateFunction {
             write!(f, " AS {function_body}")?;
         }
         if let Some(CreateFunctionBody::AsBeginEnd(bes)) = &self.function_body {
-            write!(f, " AS")?;
-            write!(f, " {}", bes)?;
+            write!(f, " AS {bes}")?;
         }
         Ok(())
     }
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2272,6 +2272,12 @@ impl fmt::Display for CreateFunction {
         if let Some(CreateFunctionBody::AsAfterOptions(function_body)) = &self.function_body {
             write!(f, " AS {function_body}")?;
         }
+        if let Some(CreateFunctionBody::MultiStatement(statements)) = &self.function_body {
+            write!(f, " AS")?;
+            write!(f, " BEGIN")?;
+            write!(f, " {}", display_separated(statements, "; "))?;
+            write!(f, " END")?;
+        }
         Ok(())
     }
 }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2318,7 +2318,14 @@ impl fmt::Display for ConditionalStatements {
     }
 }
 
-/// A shared representation of `BEGIN`, multiple statements, and `END` tokens.
+/// Represents a list of statements enclosed within `BEGIN` and `END` keywords.
+/// Example:
+/// ```sql
+/// BEGIN
+///     SELECT 1;
+///     SELECT 2;
+/// END
+/// ```
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
@@ -9273,17 +9280,10 @@ impl fmt::Display for PrintStatement {
     }
 }
 
-/// Return (MsSql)
+/// Represents a `Return` statement.
 ///
-/// for Functions:
-/// RETURN scalar_expression
-///
-/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql>
-///
-/// for Triggers:
-/// RETURN
-///
-/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql>
+/// [MsSql triggers](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql)
+/// [MsSql functions](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql)
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2292,18 +2292,14 @@ pub enum ConditionalStatements {
     /// SELECT 1; SELECT 2; SELECT 3; ...
     Sequence { statements: Vec<Statement> },
     /// BEGIN SELECT 1; SELECT 2; SELECT 3; ... END
-    BeginEnd {
-        begin_token: AttachedToken,
-        statements: Vec<Statement>,
-        end_token: AttachedToken,
-    },
+    BeginEnd(BeginEndStatements),
 }
 
 impl ConditionalStatements {
     pub fn statements(&self) -> &Vec<Statement> {
         match self {
             ConditionalStatements::Sequence { statements } => statements,
-            ConditionalStatements::BeginEnd { statements, .. } => statements,
+            ConditionalStatements::BeginEnd(bes) => &bes.statements,
         }
     }
 }
@@ -2317,12 +2313,34 @@ impl fmt::Display for ConditionalStatements {
                 }
                 Ok(())
             }
-            ConditionalStatements::BeginEnd { statements, .. } => {
-                write!(f, "BEGIN ")?;
-                format_statement_list(f, statements)?;
-                write!(f, " END")
-            }
+            ConditionalStatements::BeginEnd(bes) => write!(f, "{}", bes),
         }
+    }
+}
+
+/// A shared representation of `BEGIN`, multiple statements, and `END` tokens.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct BeginEndStatements {
+    pub begin_token: AttachedToken,
+    pub statements: Vec<Statement>,
+    pub end_token: AttachedToken,
+}
+
+impl fmt::Display for BeginEndStatements {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let BeginEndStatements {
+            begin_token: AttachedToken(begin_token),
+            statements,
+            end_token: AttachedToken(end_token),
+        } = self;
+
+        write!(f, "{begin_token} ")?;
+        if !statements.is_empty() {
+            format_statement_list(f, statements)?;
+        }
+        write!(f, " {end_token}")
     }
 }
 
@@ -8406,7 +8424,7 @@ pub enum CreateFunctionBody {
     /// ```
     ///
     /// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
-    MultiStatement(Vec<Statement>),
+    AsBeginEnd(BeginEndStatements),
     /// Function body expression using the 'RETURN' keyword.
     ///
     /// Example:
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3614,6 +3614,7 @@ pub enum Statement {
     /// 1. [Hive](https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl#LanguageManualDDL-Create/Drop/ReloadFunction)
     /// 2. [PostgreSQL](https://www.postgresql.org/docs/15/sql-createfunction.html)
     /// 3. [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_function_statement)
+    /// 4. [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql)
     CreateFunction(CreateFunction),
     /// CREATE TRIGGER
     ///
@@ -4060,6 +4061,12 @@ pub enum Statement {
     ///
     /// See: <https://learn.microsoft.com/en-us/sql/t-sql/statements/print-transact-sql>
     Print(PrintStatement),
+    /// ```sql
+    /// RETURN [ expression ]
+    /// ```
+    ///
+    /// See [ReturnStatement]
+    Return(ReturnStatement),
 }
 
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -5752,6 +5759,7 @@ impl fmt::Display for Statement {
                 Ok(())
             }
             Statement::Print(s) => write!(f, "{s}"),
+            Statement::Return(r) => write!(f, "{r}"),
             Statement::List(command) => write!(f, "LIST {command}"),
             Statement::Remove(command) => write!(f, "REMOVE {command}"),
         }
@@ -8354,6 +8362,7 @@ impl fmt::Display for FunctionDeterminismSpecifier {
 ///
 /// [BigQuery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_11
 /// [PostgreSQL]: https://www.postgresql.org/docs/15/sql-createfunction.html
+/// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
@@ -8382,6 +8391,22 @@ pub enum CreateFunctionBody {
     ///
     /// [BigQuery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_11
     AsAfterOptions(Expr),
+    /// Function body with statements before the `RETURN` keyword.
+    ///
+    /// Example:
+    /// ```sql
+    /// CREATE FUNCTION my_scalar_udf(a INT, b INT)
+    /// RETURNS INT
+    /// AS
+    /// BEGIN
+    ///     DECLARE c INT;
+    ///     SET c = a + b;
+    ///     RETURN c;
+    /// END
+    /// ```
+    ///
+    /// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
+    MultiStatement(Vec<Statement>),
     /// Function body expression using the 'RETURN' keyword.
     ///
     /// Example:
@@ -9228,6 +9253,41 @@ impl fmt::Display for PrintStatement {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "PRINT {}", self.message)
     }
+}
+
+/// Return (MsSql)
+///
+/// for Functions:
+/// RETURN scalar_expression
+///
+/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql>
+///
+/// for Triggers:
+/// RETURN
+///
+/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql>
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ReturnStatement {
+    pub value: Option<ReturnStatementValue>,
+}
+
+impl fmt::Display for ReturnStatement {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self.value {
+            Some(ReturnStatementValue::Expr(expr)) => write!(f, "RETURN {}", expr),
+            None => write!(f, "RETURN"),
+        }
+    }
+}
+
+/// Variants of a `RETURN` statement
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum ReturnStatementValue {
+    Expr(Expr),
 }
 
 #[cfg(test)]
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -779,11 +779,9 @@ impl Spanned for ConditionalStatements {
             ConditionalStatements::Sequence { statements } => {
                 union_spans(statements.iter().map(|s| s.span()))
             }
-            ConditionalStatements::BeginEnd {
-                begin_token: AttachedToken(start),
-                statements: _,
-                end_token: AttachedToken(end),
-            } => union_spans([start.span, end.span].into_iter()),
+            ConditionalStatements::BeginEnd(bes) => {
+                union_spans([bes.begin_token.0.span, bes.end_token.0.span].into_iter())
+            }
         }
     }
 }
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -23,8 +23,8 @@ use crate::tokenizer::Span;
 use super::{
     dcl::SecondaryRoles, value::ValueWithSpan, AccessExpr, AlterColumnOperation,
     AlterIndexOperation, AlterTableOperation, Array, Assignment, AssignmentTarget, AttachedToken,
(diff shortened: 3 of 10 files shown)


src/ast/mod.rs
```diff
@@ -4050,6 +4051,16 @@ pub enum Statement {
         arguments: Vec<Expr>,
         options: Vec<RaisErrorOption>,
     },
+    /// Return (SQL Server)
+    ///
+    /// for Functions:
+    /// RETURN
+    /// RETURN scalar_expression
+    /// RETURN select_statement
```
aharpervc, line 4059: 1. Perhaps this example should be removed, until it's supported?
2. Do we want a "ReturnType" enum to describe what will be?

tests/sqlparser_hive.rs
```diff
@@ -423,7 +423,7 @@ fn parse_create_function() {
     }
 
     // Test error in dialect that doesn't support parsing CREATE FUNCTION
-    let unsupported_dialects = TestedDialects::new(vec![Box::new(MsSqlDialect {})]);
+    let unsupported_dialects = TestedDialects::new(vec![Box::new(AnsiDialect {})]);
```
aharpervc, line 426: MSSQL dialect supports this now, so I had to pick another dialect 😁

src/ast/ddl.rs
```diff
@@ -2157,6 +2157,10 @@ impl fmt::Display for ClusteredBy {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct CreateFunction {
+    /// Conditionally alters the function only if it already exists.
+    ///
+    /// [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#or-alter)
+    pub or_alter: bool,
     pub or_replace: bool,
```
aharpervc, line 2164: I wasn't sure if there was a conventional ordering for these struct fields so I put it on top, before "or_replace". However I just noticed that parse_create has or_replace above or_alter here: https://github.com/apache/datafusion-sqlparser-rs/blob/d090ad4ccfd6c76c87de676fae50bbaec36e752a/src/parser/mod.rs#L4536-L4537

Happy to adjust it here for consistency if anyone prefers

src/parser/mod.rs
```diff
@@ -5135,6 +5146,63 @@ impl<'a> Parser<'a> {
...
+
+        let return_type = if self.parse_keyword(Keyword::RETURNS) {
+            Some(self.parse_data_type()?)
+        } else {
+            None
+        };
+
+        self.expect_keyword_is(Keyword::AS)?;
+        self.expect_keyword_is(Keyword::BEGIN)?;
+        let function_body = Some(CreateFunctionBody::MultiStatement(self.parse_statements()?));
```
aharpervc, line 5183: Again, this will suffer the same fate as other multi statement syntax currently in this parser, which is that semi-colons are required to parse.

tests/sqlparser_mssql.rs
```diff
@@ -187,6 +187,145 @@ fn parse_mssql_create_procedure() {
...
+        }),
+    );
+
+    let multi_statement_function = "\
+        CREATE FUNCTION some_scalar_udf(@foo INT, @bar VARCHAR(256)) \
+        RETURNS INT \
+        AS \
+        BEGIN \
+            SET @foo = @foo + 1; \
+            RETURN @foo \
```
aharpervc, line 253: Hm, the return "should" have a semi-colon here... although I think if it did, parse_statements would produce a `COMMIT` for the end. I don't think this needs to block the PR, but could be some followup work.

src/parser/mod.rs
iffyio: Ah if the desired behavior is to have optional semicolon delimited statements for mssql, what we could do would be to make the behavior configurable when [parsing the statements list](https://github.com/apache/datafusion-sqlparser-rs/blob/2457d079daeb564782f8d408c19846fe45080caf/src/parser/mod.rs#L461). We could do something like this?

```rust
pub fn parse_statements(&mut self) -> Result<Vec<Statement>, ParserError> {
    self.parse_statements_inner(true)
}

fn parse_statements_inner(&mut self, require_semicolon_delimiter) -> Result<Vec<Statement>, ParserError> {
    // ... same as the body of `parse_statements()` on main
    // ... I _think_ except for this part to make the behavior configurable
    if require_semicolon_delimiter && expecting_statement_delimiter {
        return self.expected("end of statement", self.peek_token());
    }
    // ...
}
```
then the mssql code can call the inner function directly

src/parser/mod.rs
```diff
@@ -5135,6 +5146,63 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    /// Parse `CREATE FUNCTION` for [SQL Server]
+    ///
+    /// [SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
```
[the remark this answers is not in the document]
aharpervc: I'll do it for this PR to get it merged.

Would you be willing to have a separate discussion about normalizing that name to `SQL Server` library-wide? The product/database itself is "SQL Server", and the language is "Transact-SQL" (aka T-SQL). I noticed that it's mostly like "BigQuery" rather than "Google BigQuery" and "Postgres" (or PostgreSQL) but not "PL/PGSQL". Therefore it seems like the most "normal" way to name it would just be "SQL Server" (or `SQLServer` for identifiers). The string "mssql" doesn't even show up [here](https://learn.microsoft.com/en-us/sql/sql-server/?view=sql-server-ver16) or [here](https://en.wikipedia.org/wiki/Microsoft_SQL_Server), for reference.

Therefore it seems reasonable to me, to have SQL Server be more like BigQuery & PostgreSQL rather than it's own separate naming system...

src/ast/ddl.rs
```diff
@@ -2157,6 +2157,10 @@ impl fmt::Display for ClusteredBy {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct CreateFunction {
+    /// Conditionally alters the function only if it already exists.
```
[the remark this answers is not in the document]
aharpervc: 👍  I copy/pasted from the SQL Server docs page [here](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#or-alter), although I see your point about making the comment for the struct field rather than the semantics.

src/parser/mod.rs
aharpervc: I think I understand what you're getting at with regards to configuration, but I'm not yet sure how to actually implement that statement parsing logic.

Do you see the semicolon-less statement parsing question as a blocker for the PR? If not I'd prefer to get this in (per your other suggestions) and then tackle this separately

src/ast/mod.rs
aharpervc: Excellent. I will do this

tests/sqlparser_mssql.rs
```diff
@@ -255,6 +256,93 @@ fn parse_create_function() {
...
+            determinism_specifier: None,
+            options: None,
+            remote_connection: None,
+        }),
+    );
+}
+
+#[test]
+fn parse_mssql_create_function() {
+    let create_or_alter_function = "\
```
aharpervc, line 333: In the latest iteration, I moved this "create or alter" example to a separate test function. Not sure what the best pattern here is, trying to follow existing patterns as applicable 🤔

src/ast/mod.rs
```diff
@@ -8368,6 +8387,22 @@ pub enum CreateFunctionBody {
...
+    /// AS
+    /// BEGIN
+    ///     DECLARE c INT;
+    ///     SET c = a + b;
+    ///     RETURN c;
+    /// END;
+    /// ```
+    ///
+    /// [SQL Server]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
+    MultiStatement(Vec<Statement>),
```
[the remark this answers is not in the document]
aharpervc: Is the current iteration acceptable for this PR, or do you want me to do this now?

src/parser/mod.rs
```diff
@@ -15017,6 +15075,13 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_return(&mut self) -> Result<Statement, ParserError> {
+        let expr = self.parse_expr()?;
```
aharpervc, line 15079: Does this need to be a maybe_parse, like how parse_raise_stmt does it for Exprs? The approach I picked for the discussion of RETURN variants above was to only do what's needed for _this_ PR & _these_ tests, which always have an Expr after the return.

So I think perhaps what we'd need in a followup PR is:
1. test case for bare RETURN
2. update this to be use "maybe" logic to find an expr (perhaps we'd like `maybe_expr` to wrap up that pattern?)

src/ast/mod.rs
aharpervc: We'll want bare `RETURN` for SQL Server trigger bodies, so the followup on this is over here: https://github.com/apache/datafusion-sqlparser-rs/pull/1810

src/parser/mod.rs
```diff
@@ -5135,6 +5142,69 @@ impl<'a> Parser<'a> {
...
+            };
+        self.expect_token(&Token::LParen)?;
+        let args = self.parse_comma_separated0(parse_function_param, Token::RParen)?;
+        self.expect_token(&Token::RParen)?;
+
+        let return_type = if self.parse_keyword(Keyword::RETURNS) {
+            Some(self.parse_data_type()?)
+        } else {
+            None
+        };
```
aharpervc, line 5175: In SQL Server, all function types must have a `RETURNS ...` to be valid ([ref](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16)). With this current code, a function will successfully parse even without that keyword. Therefore, it'd be up to library consumers to figure out that the return type is `None` and handle that accordingly. This seems incorrect. Therefore, I think it'd be better to expect the returns keyword.

@iffyio thoughts?

src/ast/mod.rs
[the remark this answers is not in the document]
iffyio: Yeah I think we ideally have it in this PR since this adds a new API that will otherwise need to be changed

src/ast/mod.rs
```diff
@@ -4050,6 +4051,13 @@ pub enum Statement {
         arguments: Vec<Expr>,
         options: Vec<RaisErrorOption>,
     },
+    /// Return (Mssql)
+    ///
+    /// for Functions:
+    /// RETURN scalar_expression
+    ///
+    /// See: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
```
iffyio, line 4059: Maybe we can move this to the `ReturnStatement` struct definition, and here we can only point to the statement like `Parse a [ReturnStatement]`? Thinking having it on the struct will be more visible as the struct can be reused/reached from more places

src/ast/mod.rs
```diff
@@ -9203,6 +9232,30 @@ pub enum CopyIntoSnowflakeKind {
...
+impl fmt::Display for ReturnStatement {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self.value {
+            Some(ReturnStatementValue::Expr(expr)) => write!(f, "RETURN {}", expr),
+            None => write!(f, "RETURN")
+        }
+    }
+}
+
+/// Variants of the Mssql `RETURN` statement
```
iffyio, line 9251:
```suggestion
/// Variants of a `RETURN` statement
```
since this statement isn't mysql specific

src/ast/mod.rs
iffyio, line 4059:
```suggestion
    /// Return
    ///
    /// for Functions:
    /// RETURN scalar_expression
    ///
    /// [MsSql]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql
```
since the statement is supported by other dialects

src/parser/mod.rs
iffyio, line 5175: This part looks identical to the [bigquery implementation](https://github.com/apache/datafusion-sqlparser-rs/blob/1b50dd9d0764b96d9a6b4e81f77d77e7cae84097/src/parser/mod.rs#L5064-L5085) could we pull the logic out into a function to be reused by both?

iffyio: Yeah it sounds reasonable in that case to expect the returns clause

src/parser/mod.rs
iffyio: Maybe we can remove the return statement bits from this PR and tackle it separately as its own PR?

Otherwise indeed the expectation would be that this uses maybe_parse and we have the test cases in place as part of this PR

src/parser/mod.rs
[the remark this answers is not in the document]
iffyio: Yeah the mssql name used in the library is unfortunate, but it wouldn't be worth the breakage to rename the [dialect itself](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/mssql.rs#L29-L31) and the documentation  in the code follows from there, so that I think it would be preferable to remain consistent with the naming at this point even if the name isn't optimal

src/parser/mod.rs
iffyio: > but I'm not yet sure how to actually implement that statement parsing logic

hmm not sure I followed what we mean here? parsing a semicolon-less statement I imagine would follow from the above code snippet by calling `self.parse_statements_inner(false)`

It would indeed be good to resolve in this PR, currently in the PR we seem to have a workaround for this behavior [here](https://github.com/apache/datafusion-sqlparser-rs/blob/1b50dd9d0764b96d9a6b4e81f77d77e7cae84097/src/parser/mod.rs#L487-L493) and [here](https://github.com/apache/datafusion-sqlparser-rs/blob/1b50dd9d0764b96d9a6b4e81f77d77e7cae84097/src/parser/mod.rs#L5180-L5185) and another in the Go statement PR.

src/parser/mod.rs
aharpervc: > Maybe we can remove the return statement bits from this PR and tackle it separately as its own PR?

I don't think this is possible, since SQL Server requires that functions have a return, so create function goes together with return. I'll take another pass on the PR though

src/ast/mod.rs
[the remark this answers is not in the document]
aharpervc: > Yeah I think we ideally have it in this PR since this adds a new API that will otherwise need to be changed

Makes sense, and I've done it, per your suggestion above -- BeginEndStatements is extracted from ConditionalStatements as a re-usable struct, and then also used for parsing create function. It's much cleaner!

src/parser/mod.rs

src/ast/mod.rs

src/ast/ddl.rs
```diff
@@ -2272,6 +2277,10 @@ impl fmt::Display for CreateFunction {
         if let Some(CreateFunctionBody::AsAfterOptions(function_body)) = &self.function_body {
             write!(f, " AS {function_body}")?;
         }
+        if let Some(CreateFunctionBody::AsBeginEnd(bes)) = &self.function_body {
+            write!(f, " AS")?;
+            write!(f, " {}", bes)?;
```
iffyio, line 2282:
```suggestion
            write!(f, " AS {bes}")?;
```

src/ast/spans.rs
```diff
@@ -777,11 +778,9 @@ impl Spanned for ConditionalStatements {
             ConditionalStatements::Sequence { statements } => {
                 union_spans(statements.iter().map(|s| s.span()))
             }
-            ConditionalStatements::BeginEnd {
-                begin_token: AttachedToken(start),
-                statements: _,
-                end_token: AttachedToken(end),
-            } => union_spans([start.span, end.span].into_iter()),
+            ConditionalStatements::BeginEnd(bes) => {
+                union_spans([bes.begin_token.0.span, bes.end_token.0.span].into_iter())
```
iffyio, line 782: Can we `impl Spanned for BeginEndStatements` and then call `bes.span()` here instead? That would make the span impl reusable by nother nodes that embed the `BeginEndStatements`

src/ast/mod.rs
```diff
@@ -9211,6 +9253,41 @@ pub enum CopyIntoSnowflakeKind {
...
+///
+/// for Functions:
+/// RETURN scalar_expression
+///
+/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql>
+///
+/// for Triggers:
+/// RETURN
+///
+/// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql>
```
iffyio, line 9266:
```suggestion
/// Represents a `Return` statement.
///
/// [MsSql triggers](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql)
/// [MySql functions](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql)
```
this makes it possible to include docs to other dialects like bigquery/snowflake that support this statement

src/ast/mod.rs
```diff
@@ -2317,15 +2313,37 @@ impl fmt::Display for ConditionalStatements {
...
-                write!(f, "BEGIN ")?;
-                format_statement_list(f, statements)?;
-                write!(f, " END")
-            }
+            ConditionalStatements::BeginEnd(bes) => write!(f, "{}", bes),
         }
     }
 }
 
+/// A shared representation of `BEGIN`, multiple statements, and `END` tokens.
```
iffyio, line 2321:
```suggestion
/// Represents a list of statements enclosed within `BEGIN` and `END` keywords.
/// Example:
/// ```sql
/// BEGIN
///     SELECT 1;
///     SELECT 2;
/// END
/// ```
```

tests/sqlparser_mssql.rs
```diff
@@ -187,6 +188,386 @@ fn parse_mssql_create_procedure() {
...
+            ]),
+            return_type: Some(DataType::Int(None)),
+            function_body: Some(CreateFunctionBody::AsBeginEnd(BeginEndStatements {
+                begin_token: AttachedToken(TokenWithSpan::wrap(sqlparser::tokenizer::Token::Word(
+                    sqlparser::tokenizer::Word {
+                        value: "BEGIN".to_string(),
+                        quote_style: None,
+                        keyword: Keyword::BEGIN
+                    }
+                ))),
```
iffyio, line 239:
```suggestion
                begin_token: AttachedToken::empty(),
```
for tests in places where a token is needed, we can use this instead. since there is no Eq impl on the `AttachedToken` type

tests/sqlparser_mssql.rs
```diff
@@ -187,6 +188,386 @@ fn parse_mssql_create_procedure() {
...
+                span: Span::empty(),
+            }]),
+            args: Some(vec![
+                OperateFunctionArg {
+                    mode: None,
+                    name: Some(Ident {
+                        value: "@foo".into(),
+                        quote_style: None,
+                        span: Span::empty(),
+                    }),
```
iffyio, line 213:
```suggestion
                    name: Some(Ident::new("@foo")),
```
I think the idents in the tests can be simplified with the following

tests/sqlparser_mssql.rs
```diff
@@ -187,6 +188,386 @@ fn parse_mssql_create_procedure() {
...
+        BEGIN
+            IF 1=2
+            BEGIN
+                RETURN 1;
+            END
+
+            RETURN 0;
+        END
+    "#;
+    let create_stmt = ms().one_statement_parses_to(create_function_with_conditional, "");
```
iffyio, line 374: can we use `verified_stmt` here instead (what does the empty canonical string imply)?

tests/sqlparser_mssql.rs
```diff
@@ -187,6 +188,386 @@ fn parse_mssql_create_procedure() {
...
+            parallel: None,
+            using: None,
+            language: None,
+            determinism_specifier: None,
+            options: None,
+            remote_connection: None,
+        }),
+    );
+
+    let multi_statement_function = "\
```
iffyio, line 264: to simplify the tests, I think we can have one of the test cases like `return_expression_function` as is, which asserts the returned AST. Then for the others, it should be enough to rely on `verified_stmt`, without fully asserting the AST. the latter has significant overhead maintenance wise so that ideally we use them more sparingly.

So essentially thinking for `multi_statement_function`, `create_function_with_conditional` and `create_or_alter_function` we just call `verified_stmt(sql)` to assert behavior. Then the `create_or_alter_function` test case I think we can move into this function since its part of the same feature

src/parser/mod.rs
iffyio: Ah alright, in that case I think its only the test cases missing for the return statement we can add those to common.rs I imagine

src/parser/mod.rs
iffyio: @aharpervc just in case, note that this part of the PR is unresolved

aharpervc: Thanks for the reminder.

> It would indeed be good to resolve in this PR, currently in the PR we seem to have a workaround for this behavior [here](https://github.com/apache/datafusion-sqlparser-rs/blob/1b50dd9d0764b96d9a6b4e81f77d77e7cae84097/src/parser/mod.rs#L487-L493) and [here](https://github.com/apache/datafusion-sqlparser-rs/blob/1b50dd9d0764b96d9a6b4e81f77d77e7cae84097/src/parser/mod.rs#L5180-L5185) and another in the Go statement PR.

I was able to refactor the latter, and the former we'll need regardless of the semi colon situation. Does the latest iteration resolve your concerns here?

tests/sqlparser_mssql.rs
aharpervc: This is fixed

src/parser/mod.rs
```diff
@@ -580,10 +580,12 @@ impl<'a> Parser<'a> {
                 // `BEGIN` is a nonstandard but common alias for the
                 // standard `START TRANSACTION` statement. It is supported
                 // by at least PostgreSQL and MySQL.
+                // `BEGIN` is also used for multi-statement blocks in SQL Server
                 Keyword::BEGIN => self.parse_begin(),
                 // `END` is a nonstandard but common alias for the
                 // standard `COMMIT TRANSACTION` statement. It is supported
                 // by PostgreSQL.
+                // `END` is also used for multi-statement blocks in SQL Server
```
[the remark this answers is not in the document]

tests/sqlparser_mssql.rs
aharpervc: I think the main downside is we lose the newlines, right? So we can't write SQL text the conventional way with new lines with `r#"..."#;`. Unless, are you suggesting this example case be reduced to _only_ verified_stmt, and no ast assertion?

src/parser/mod.rs
aharpervc: I've added a test case for a bare `RETURN` to sqlparser_common.rs, although I'm unsure how useful it is. At least for SQL Server, a top level return _must_ be bare (eg, can't return anything). Therefore, the more useful tests are where return is inside a function body, trigger body, etc. And we have those tests in their respective branches.

src/ast/ddl.rs
aharpervc: Done 👍

src/ast/mod.rs

---

BTW, see how easy it is for "MsSql" to accidentally become "M**y**Sql"? Another vote for renaming the dialect 😆 

image

src/ast/mod.rs

tests/sqlparser_mssql.rs
aharpervc: Done 👍. I don't really understand the pro/con versus an empty token, though

tests/sqlparser_mssql.rs
aharpervc: Makes sense, I didn't know you could do that. Done 👍

tests/sqlparser_mssql.rs
aharpervc: Done 👍

If I'm understanding your thinking, we don't need a separate ast assertion for `or_alter`, because it's the preference of this project to pair parsing + writing, such that new properties like this can fully round trip, and therefore as long as the output matches the input we're appropriately covered. Is that accurate?

src/ast/spans.rs
aharpervc: I think I understand what you're asking for, I'll give the implementation a shot

aharpervc: Done 👍

src/ast/spans.rs
```diff
@@ -2280,6 +2277,12 @@ impl Spanned for TableObject {
     }
 }
 
+impl Spanned for BeginEndStatements {
+    fn span(&self) -> Span {
+        union_spans([self.begin_token.0.span, self.end_token.0.span].into_iter())
```
iffyio, line 2282: We can do an exhausitve match here, so that the code is revisite if fields get added or changed
```rust
let BeginEndStatements {
    begin,
    statements,
    end
} = self;
union_spans(begin, statements, end)
```
we can probably include statements in the union spans call if only for completeness

tests/sqlparser_common.rs
```diff
@@ -15015,3 +15015,8 @@ fn parse_set_time_zone_alias() {
         _ => unreachable!(),
     }
 }
+
+#[test]
+fn parse_return() {
+    all_dialects().verified_stmt("RETURN");
```
iffyio, line 15021: Can we have this check assert the AST? Also we can add another test case that includes a return statement having an expression

tests/sqlparser_mssql.rs
iffyio: Yeah that's correct!

src/parser/mod.rs
iffyio: does something like in [this suggestion](https://github.com/apache/datafusion-sqlparser-rs/pull/1808#discussion_r2040592844) work just as well? its currently unclear to me why the semicolon handling is a special case of an `IfCondition` statement

src/ast/spans.rs
aharpervc: I updated this, although I'm unsure if my implementation was what you're suggesting

tests/sqlparser_common.rs
aharpervc: > Can we have this check assert the AST?

Done 👍 

> Also we can add another test case that includes a return statement having an expression

Done, although this is invalid for SQL Server. I think having that test here would also mean we can't have a test that asserts an error when return with value is found as a top level statement. Thoughts? 🤔

src/parser/mod.rs
aharpervc: I'm not deeply opposed to making an inner parse function, I guess, but it also doesn't seem required for _this_ PR. 

> its currently unclear to me why the semicolon handling is a special case of an IfCondition statement

I reverted that change, I think either it was a problem while hacking on the branch or pollution from another branch 😬

src/parser/mod.rs
```diff
@@ -5134,6 +5128,76 @@ impl<'a> Parser<'a> {
...
+        or_replace: bool,
+        temporary: bool,
+    ) -> Result<Statement, ParserError> {
+        let (name, args) = self.parse_create_function_name_and_params()?;
+
+        let return_type = if self.parse_keyword(Keyword::RETURNS) {
+            Some(self.parse_data_type()?)
+        } else {
+            return parser_err!("Expected RETURNS keyword", self.peek_token().span.start);
+        };
```
iffyio, line 5146:
```suggestion
        self.expect_keyword(Keyword::RETURNS)?;
        let return_type = Some(self.parse_data_type()?);
```

src/parser/mod.rs
```diff
@@ -15063,6 +15127,15 @@ impl<'a> Parser<'a> {
             message: Box::new(self.parse_expr()?),
         }))
     }
+    /// Parse [Statement::Return]
```
iffyio, line 15130:
```suggestion

    /// Parse [Statement::Return]
```

src/parser/mod.rs
iffyio: we seem to have some of the related changes left in this codepath? https://github.com/apache/datafusion-sqlparser-rs/blob/6d0ea6462ead65ff598105d8cd44468d972667cb/src/parser/mod.rs#L4458-L4466

tests/sqlparser_common.rs
iffyio: hmm yeah given that its currently part of the parser behavior I think its reasonable to assert it. If the parser later stops accepting a top-level return value then we would change the test here to assert an error instead

iffyio (commented):
Thanks @aharpervc! Left a couple minor comments, I think beyond the if condition semicolon special handling the changes look reasonable to me!

src/parser/mod.rs
aharpervc: Done 👍

iffyio (approved):
LGTM! Thanks @aharpervc!
 cc @alamb

alamb:
Thanks again @iffyio and @aharpervc

PR #1809, opened 2025-04-11
Add support for `GO` batch delimiter in SQL Server

aharpervc:
Reference: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go

Lots of conventional SQL Server tooling supports `GO`, so it seems really useful & prudent to be able to parse it. I realize that strictly speaking, it isn't a SQL "statement". However, I think it makes sense to model here as a statement since it acts as one in all regards except documentation.

In terms of parsing, `GO` has some peculiarities I tried to express with the new tests, such as with regards to position and basically treating newline as a delimiter. This also starts to nudge on https://github.com/apache/datafusion-sqlparser-rs/issues/1800, where newlines may possibly be interpreted as statement delimiters in a general case. However, my approach here for `GO` is isolated to only that statement, at least for the time being.

---

Also, this will conflict with https://github.com/apache/datafusion-sqlparser-rs/pull/1808 when either is merged, because I'm adding things that end up being adjacent to each other which can confuse git, depending on merge strategy. There's no _actual_ conflict.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4097,6 +4097,12 @@ pub enum Statement {
     ///
     /// See [ReturnStatement]
     Return(ReturnStatement),
+    /// Go (MsSql)
+    ///
+    /// GO is not a Transact-SQL statement; it is a command recognized by various tools as a batch delimiter
+    ///
+    /// See: <https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go>
+    Go(GoStatement),
 }
 
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -5791,6 +5797,7 @@ impl fmt::Display for Statement {
                 Ok(())
             }
             Statement::Print(s) => write!(f, "{s}"),
+            Statement::Go(s) => write!(f, "{s}"),
             Statement::Return(r) => write!(f, "{r}"),
             Statement::List(command) => write!(f, "LIST {command}"),
             Statement::Remove(command) => write!(f, "REMOVE {command}"),
@@ -9313,6 +9320,26 @@ impl fmt::Display for ReturnStatement {
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum ReturnStatementValue {
     Expr(Expr),
+}
+
+/// Represents a `GO` statement.
+///
+/// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct GoStatement {
+    pub count: Option<u64>,
+}
+
+impl Display for GoStatement {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(count) = self.count {
+            write!(f, "GO {count}")
+        } else {
+            write!(f, "GO")
+        }
+    }
 }
 
 #[cfg(test)]
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -22,7 +22,7 @@ use crate::ast::{
 use crate::dialect::Dialect;
 use crate::keywords::{self, Keyword};
 use crate::parser::{Parser, ParserError};
-use crate::tokenizer::{Token, Whitespace};
+use crate::tokenizer::Token;
 #[cfg(not(feature = "std"))]
 use alloc::{vec, vec::Vec};
 
@@ -119,26 +119,14 @@ impl Dialect for MsSqlDialect {
     }
 
     fn is_column_alias(&self, kw: &Keyword, parser: &mut Parser) -> bool {
-        // if:
-        // - keyword is `GO`, and
-        // - looking backwards there's only (any) whitespace preceded by a newline
-        // then: `GO` iSN'T a column alias
-        if kw == &Keyword::GO {
-            let mut look_back_count = 2;
-            loop {
-                let prev_index = parser.index().saturating_sub(look_back_count);
-                if prev_index == 0 {
-                    break;
-                }
-                let prev_token = parser.token_at(prev_index);
-                match prev_token.token {
-                    Token::Whitespace(ref w) => match w {
-                        Whitespace::Newline => return false,
-                        _ => look_back_count += 1,
-                    },
-                    _ => break,
-                };
-            }
+        // if we find maybe whitespace then a newline looking backward, then `GO` ISN'T a column alias
+        // if we can't find a newline then we assume that `GO` IS a column alias
+        if kw == &Keyword::GO
+            && parser
+                .expect_previously_only_whitespace_until_newline()
+                .is_ok()
+        {
+            return false;
         }
 
         !keywords::RESERVED_FOR_COLUMN_ALIAS.contains(kw) && !RESERVED_FOR_COLUMN_ALIAS.contains(kw)
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -22,7 +22,7 @@ use crate::ast::{
 use crate::dialect::Dialect;
 use crate::keywords::{self, Keyword};
 use crate::parser::{Parser, ParserError};
-use crate::tokenizer::Token;
+use crate::tokenizer::{Token, Whitespace};
 #[cfg(not(feature = "std"))]
 use alloc::{vec, vec::Vec};
 
@@ -118,7 +118,29 @@ impl Dialect for MsSqlDialect {
         true
     }
 
-    fn is_column_alias(&self, kw: &Keyword, _parser: &mut Parser) -> bool {
+    fn is_column_alias(&self, kw: &Keyword, parser: &mut Parser) -> bool {
+        // if:
+        // - keyword is `GO`, and
+        // - looking backwards there's only (any) whitespace preceded by a newline
+        // then: `GO` iSN'T a column alias
+        if kw == &Keyword::GO {
+            let mut look_back_count = 2;
+            loop {
+                let prev_index = parser.index().saturating_sub(look_back_count);
+                if prev_index == 0 {
+                    break;
+                }
+                let prev_token = parser.token_at(prev_index);
+                match prev_token.token {
+                    Token::Whitespace(ref w) => match w {
+                        Whitespace::Newline => return false,
+                        _ => look_back_count += 1,
+                    },
+                    _ => break,
+                };
+            }
+        }
+
         !keywords::RESERVED_FOR_COLUMN_ALIAS.contains(kw) && !RESERVED_FOR_COLUMN_ALIAS.contains(kw)
     }
 
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -121,11 +121,7 @@ impl Dialect for MsSqlDialect {
     fn is_column_alias(&self, kw: &Keyword, parser: &mut Parser) -> bool {
         // if we find maybe whitespace then a newline looking backward, then `GO` ISN'T a column alias
         // if we can't find a newline then we assume that `GO` IS a column alias
-        if kw == &Keyword::GO
-            && parser
-                .expect_previously_only_whitespace_until_newline()
-                .is_ok()
-        {
+        if kw == &Keyword::GO && parser.prev_only_whitespace_until_newline() {
             return false;
         }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -475,12 +475,6 @@ impl<'a> Parser<'a> {
                     if expecting_statement_delimiter && word.keyword == Keyword::END {
                         break;
                     }
-                    // Treat batch delimiter as an end of statement
-                    if expecting_statement_delimiter && dialect_of!(self is MsSqlDialect) {
-                        if let Some(Statement::Go(GoStatement { count: _ })) = stmts.last() {
-                            expecting_statement_delimiter = false;
-                        }
-                    }
                 }
                 _ => {}
             }
@@ -497,6 +491,8 @@ impl<'a> Parser<'a> {
                     // parsing the statement sequence consumes the statement delimiter, so it shouldn't be expected here
                     ConditionalStatements::Sequence { .. } => false,
                 },
+                // Treat batch delimiter as an end of statement, so no additional statement delimiter expected here
+                Statement::Go(_) => false,
                 _ => true,
             };
             stmts.push(statement);
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -475,6 +475,10 @@ impl<'a> Parser<'a> {
                     if expecting_statement_delimiter && word.keyword == Keyword::END {
                         break;
                     }
+
+                    if expecting_statement_delimiter && word.keyword == Keyword::GO {
+                        expecting_statement_delimiter = false;
+                    }
                 }
                 _ => {}
             }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -490,8 +490,16 @@ impl<'a> Parser<'a> {
             }
 
             let statement = self.parse_statement()?;
+            expecting_statement_delimiter = match &statement {
+                Statement::If(s) => match &s.if_block.conditional_statements {
+                    // the `END` keyword doesn't need to be followed by a statement delimiter, so it shouldn't be expected here
+                    ConditionalStatements::BeginEnd { .. } => false,
+                    // parsing the statement sequence consumes the statement delimiter, so it shouldn't be expected here
+                    ConditionalStatements::Sequence { .. } => false,
+                },
+                _ => true,
+            };
             stmts.push(statement);
-            expecting_statement_delimiter = true;
         }
         Ok(stmts)
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15254,6 +15254,12 @@ impl<'a> Parser<'a> {
             match prev_token.token {
                 Token::Whitespace(ref w) => match w {
                     Whitespace::Newline => break,
+                    Whitespace::SingleLineComment { comment, prefix: _ } => {
+                        if comment.ends_with('\n') {
+                            break;
+                        }
+                        look_back_count += 1;
+                    }
                     _ => look_back_count += 1,
                 },
                 _ => {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15324,11 +15324,27 @@ impl<'a> Parser<'a> {
             };
         };
 
-        if self.peek_token().token == Token::SemiColon {
-            parser_err!(
-                "GO may not end with a semicolon",
-                self.peek_token().span.start
-            )?;
+        loop {
+            let next_token = self.peek_token_no_skip();
+            match next_token.token {
+                Token::EOF => break,
+                Token::Whitespace(ref w) => match w {
+                    Whitespace::Newline => break,
+                    Whitespace::SingleLineComment { comment, prefix: _ } => {
+                        if comment.ends_with('\n') {
+                            break;
+                        }
+                        _ = self.next_token_no_skip();
+                    }
+                    _ => _ = self.next_token_no_skip(),
+                },
+                _ => {
+                    parser_err!(
+                        "GO must be followed by a newline or EOF",
+                        self.peek_token().span.start
+                    )?;
+                }
+            };
         }
 
         Ok(Statement::Go(GoStatement { count }))
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15294,6 +15294,16 @@ impl<'a> Parser<'a> {
 
     /// Parse [Statement::Go]
     fn parse_go(&mut self) -> Result<Statement, ParserError> {
+        // disambiguate between GO as batch delimiter & GO as identifier (etc)
+        // compare:
+        // ```sql
+        // select 1 go
+        // ```
+        // vs
+        // ```sql
+        // select 1
+        // go
+        // ```
         self.expect_previously_only_whitespace_until_newline()?;
 
         let count = loop {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4065,6 +4065,44 @@ impl<'a> Parser<'a> {
         )
     }
 
+    /// Look backwards in the token stream and expect that there was only whitespace tokens until the previous newline
+    pub fn expect_previously_only_whitespace_until_newline(&mut self) -> Result<(), ParserError> {
+        let mut look_back_count = 2;
+        loop {
+            let prev_index = self.index.saturating_sub(look_back_count);
+            if prev_index == 0 {
+                break;
+            }
+            let prev_token = self.token_at(prev_index);
+            match prev_token.token {
+                Token::Whitespace(ref w) => match w {
+                    Whitespace::Newline => break,
+                    // special consideration required for single line comments since that string includes the newline
+                    Whitespace::SingleLineComment { comment, prefix: _ } => {
+                        if comment.ends_with('\n') {
+                            break;
+                        }
+                        look_back_count += 1;
+                    }
+                    _ => look_back_count += 1,
+                },
+                _ => {
+                    let current_token = self.get_current_token();
+                    if prev_token == current_token {
+                        // if we are at the start of the statement, we can skip this check
+                        break;
+                    }
+
+                    self.expected(
+                        &format!("newline before current token ({})", current_token),
+                        prev_token.clone(),
+                    )?
+                }
+            };
+        }
+        Ok(())
+    }
+
     /// If the current token is the `expected` keyword, consume it and returns
     /// true. Otherwise, no tokens are consumed and returns false.
     #[must_use]
@@ -15242,36 +15280,7 @@ impl<'a> Parser<'a> {
 
     /// Parse [Statement::Go]
     fn parse_go(&mut self) -> Result<Statement, ParserError> {
-        // previous token should be a newline (skipping non-newline whitespace)
-        // see also, `previous_token`
-        let mut look_back_count = 2;
-        loop {
-            let prev_index = self.index.saturating_sub(look_back_count);
-            if prev_index == 0 {
-                break;
-            }
-            let prev_token = self.token_at(prev_index);
-            match prev_token.token {
-                Token::Whitespace(ref w) => match w {
-                    Whitespace::Newline => break,
-                    Whitespace::SingleLineComment { comment, prefix: _ } => {
-                        if comment.ends_with('\n') {
-                            break;
-                        }
-                        look_back_count += 1;
-                    }
-                    _ => look_back_count += 1,
-                },
-                _ => {
-                    if prev_token == self.get_current_token() {
-                        // if we are at the start of the statement, we can skip this check
-                        break;
-                    }
-
-                    self.expected("newline before GO", prev_token.clone())?
-                }
-            };
-        }
+        self.expect_previously_only_whitespace_until_newline()?;
 
         let count = loop {
             // using this peek function because we want to halt this statement parsing upon newline
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4077,35 +4077,26 @@ impl<'a> Parser<'a> {
     }
 
     /// Look backwards in the token stream and expect that there was only whitespace tokens until the previous newline or beginning of string
-    pub(crate) fn expect_previously_only_whitespace_until_newline(
-        &mut self,
-    ) -> Result<(), ParserError> {
+    pub(crate) fn prev_only_whitespace_until_newline(&mut self) -> bool {
         let mut look_back_count = 1;
         loop {
             let prev_token = self.peek_prev_nth_token_no_skip_ref(look_back_count);
             match prev_token.token {
-                Token::EOF => break,
+                Token::EOF => break true,
                 Token::Whitespace(ref w) => match w {
-                    Whitespace::Newline => break,
+                    Whitespace::Newline => break true,
                     // special consideration required for single line comments since that string includes the newline
                     Whitespace::SingleLineComment { comment, prefix: _ } => {
                         if comment.ends_with('\n') {
-                            break;
+                            break true;
                         }
                         look_back_count += 1;
                     }
                     _ => look_back_count += 1,
                 },
-                _ => self.expected(
-                    &format!(
-                        "newline before current token ({})",
-                        self.get_current_token()
-                    ),
-                    prev_token.clone(),
-                )?,
+                _ => break false,
             };
         }
-        Ok(())
     }
 
     /// If the current token is the `expected` keyword, consume it and returns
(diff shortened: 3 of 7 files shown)


src/ast/mod.rs
```diff
@@ -4050,6 +4050,14 @@ pub enum Statement {
...
         options: Vec<RaisErrorOption>,
     },
+    /// Go (SQL Server)
+    ///
+    /// GO is not a Transact-SQL statement; it is a command recognized by various tools as a batch delimiter
+    ///
+    /// See: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go
+    Go {
+        count: Option<u64>,
+    },
```
iffyio, line 4060: Repr wise can we wrap the statement body in explicit structs? we're[ transitioning away](https://github.com/apache/datafusion-sqlparser-rs/issues/1204) from anonymous structs in order to make the API more ergonomic. Something like
```rust
struct GoStatement {
    count: Option<u64>
}
Statement::Go(GoStatement)
```

src/parser/mod.rs
```diff
@@ -475,6 +475,12 @@ impl<'a> Parser<'a> {
                     if expecting_statement_delimiter && word.keyword == Keyword::END {
                         break;
                     }
+                    // Treat batch delimiter as an end of statement
+                    if expecting_statement_delimiter && dialect_of!(self is MsSqlDialect) {
+                        if let Some(Statement::Go { count: _ }) = stmts.last() {
+                            expecting_statement_delimiter = false;
+                        }
+                    }
```
iffyio, line 483: not sure I followed this part, what was the intention?

src/parser/mod.rs
```diff
@@ -617,6 +623,9 @@ impl<'a> Parser<'a> {
                 }
                 // `COMMENT` is snowflake specific https://docs.snowflake.com/en/sql-reference/sql/comment
                 Keyword::COMMENT if self.dialect.supports_comment_on() => self.parse_comment(),
+                Keyword::GO if dialect_of!(self is MsSqlDialect) => {
```
iffyio, line 626:
```suggestion
                Keyword::GO => {
```
I think it should be fine to let the parser always accept the statement if it shows up. Technically if we wanted to gate it behind a feature then we could use a `self.dialect.supports` flag since the `dialect_of` macro is deprecated, but I think it makes more sense to let the parser accept unconditionally

src/parser/mod.rs
```diff
@@ -15017,6 +15026,48 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_go(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 15029:
```suggestion
    /// Parse [Statement::Go]
    fn parse_go(&mut self) -> Result<Statement, ParserError> {
```

iffyio, line 15029: One thing we could do, in the `parse_stmt()` function could we call `self.prev_token()` to rewind the `Go` keyword so that this function becomes self contained in being able to parse a full `Go` statement?

src/parser/mod.rs
```diff
@@ -15017,6 +15026,48 @@ impl<'a> Parser<'a> {
...
+        // previous token should be a newline (skipping non-newline whitespace)
```
iffyio, line 15030: hmm I didn't look too closely but not sure I followed the function body on a high level, from the docs and the representation in the parser the Go statement only contains an optional int, so that I imagined all the function needs to do would be to `maybe_parse(|parser| parser.parse_number_value())` or similar? The function is currently managing whitespace and semicolon but not super clear to me why that is required

tests/sqlparser_mssql.rs
```diff
@@ -2036,3 +2036,78 @@ fn parse_mssql_merge_with_output() {
...
+    let err = ms().parse_sql_statements(invalid_go_position);
+    assert!(err.is_err());
+    assert_eq!(
+        err.unwrap_err().to_string(),
+        "sql parser error: Expected: newline before GO, found: ;"
+    );
+
+    let invalid_go_count = "SELECT 1\nGO x";
+    let err = ms().parse_sql_statements(invalid_go_count);
+    assert!(err.is_err());
```
iffyio, line 2108:
```suggestion
    assert_eq!(
        err.unwrap_err().to_string(),
        "sql parser error: Expected: newline before GO, found: ;"
    );

    let invalid_go_count = "SELECT 1\nGO x";
    let err = ms().parse_sql_statements(invalid_go_count);
```
since we're already unwrapping the error

aharpervc: Ah yes great point

src/ast/mod.rs

src/parser/mod.rs

src/parser/mod.rs
aharpervc: I tried to explain this in the commit message & the "multiple_gos" example in the test. Basically, we want this to parse as 4 statements:

```mssql
SELECT 1;
GO
SELECT 2;
GO
```

The second `GO` is terminated by EOF, so no problem. But the first `GO` _cannot_ end with a semi colon (This overlaps with the ongoing "no semi-colons for statements" discussion but from a different angle). So we need to figure out how to parse it all out without the existing conventional semi-colon logic.

What I came up with is to basically just look for a GO and have that signal that we _aren't_ then also looking for statement delimiter (eg, semi-colon). This also makes sense semantically because statements can't extend between batches, so any batch delimiter also functions as a statement delimiter.

src/parser/mod.rs
aharpervc: Does [my comment above](https://github.com/apache/datafusion-sqlparser-rs/pull/1809#discussion_r2042359100) & the code comment below, "using this peek function because we want to halt this statement parsing upon newline" help?

```mssql
-- this should parse as `Go { Some(4) }`
GO 4`

-- this should parse as something like `Go { None }`, `Expr(4)`
GO
4
```

Eg, the whitespace is meaningful for parsing

src/parser/mod.rs
aharpervc: In terms of the code, I worked out a perhaps slightly less confusing way to set whether or not a semi-colon is expected, [here](https://github.com/apache/datafusion-sqlparser-rs/pull/1808/files#diff-fc6a8d66b8cb6bd48a119a70e489cbcef82e7f551e7cf08e8e972ef4e774ef49R487-R493).

This branch can be updated accordingly if we agree over there

aharpervc: I've now done that due to another problem I noticed, for IF statements (new test case also added...)

aharpervc:
Rebased on main & updated

src/parser/mod.rs
```diff
@@ -15058,6 +15069,57 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse [Statement::Go]
+    fn parse_go(&mut self) -> Result<Statement, ParserError> {
+        // previous token should be a newline (skipping non-newline whitespace)
+        // see also, `previous_token`
+        let mut look_back_count = 2;
+        loop {
```
aharpervc, line 15077: This can maybe be simplified. Also, needs a test when `go` starts a file but followed by other statements

src/parser/mod.rs
iffyio: Ah interesting, I think it would be very helpful to add this example to the code to illustrate the implications,

Then for the function, I think using the `peek_token_no_skip` could simplify the impl. Something roughly like the following? to avoid manually tracking token counts and lookbacks
```rust
let count = loop {
  match self.peek_token_no_skip() {
    Token::Whitespace(Whitespace::Newline) | Token::Whitespace(Whitespace::SingleLineComment) | Token::EOF => {
      self.advance_token();
      break None
    }
    Token::Whitespace(_) => {
      self.advance_token();
    }
    _ => {
      break Some(self.parse_number())
    }
  }
};
```

tests/sqlparser_mssql.rs
```diff
@@ -2053,3 +2054,171 @@ fn parse_drop_trigger() {
         }
     );
 }
+
+#[test]
+fn parse_mssql_go_keyword() {
+    let single_go_keyword = "USE some_database;\nGO";
```
iffyio, line 2060: Can we have one of the test cases use `ms().one_statement_parses_to(sql, canonical)`? looking at the tests its unclear what the behavior of the parser is in terms of serializing an AST that has GO statements in it, and that it needs special behavior to inject new lines manually I imagine

tests/sqlparser_mssql.rs
```diff
@@ -2053,3 +2054,171 @@ fn parse_drop_trigger() {
...
+    assert_eq!(stmts[1], Statement::Go(GoStatement { count: None }));
+
+    let multi_line_comment_preceding_go = "USE some_database; /* okay */\nGO";
+    let stmts = ms()
+        .parse_sql_statements(multi_line_comment_preceding_go)
+        .unwrap();
+    assert_eq!(stmts.len(), 2);
+    assert_eq!(stmts[1], Statement::Go(GoStatement { count: None }));
+
+    let comment_following_go = "USE some_database;\nGO -- okay";
```
iffyio, line 2117: maybe we can add a `multine_line_comment_following_go` test case? e.g. `GO /* whitespace */ 42`

src/ast/mod.rs
```diff
@@ -4054,6 +4054,12 @@ pub enum Statement {
         arguments: Vec<Expr>,
         options: Vec<RaisErrorOption>,
     },
+    /// Go (MSSQL)
+    ///
+    /// GO is not a Transact-SQL statement; it is a command recognized by various tools as a batch delimiter
+    ///
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go>
```
iffyio, line 4061: For visibility we can move the comment to the struct definition similar to the print statement?

src/dialect/mssql.rs
```diff
@@ -116,7 +116,29 @@ impl Dialect for MsSqlDialect {
...
     }
 
-    fn is_column_alias(&self, kw: &Keyword, _parser: &mut Parser) -> bool {
+    fn is_column_alias(&self, kw: &Keyword, parser: &mut Parser) -> bool {
+        // if:
+        // - keyword is `GO`, and
+        // - looking backwards there's only (any) whitespace preceded by a newline
+        // then: `GO` iSN'T a column alias
+        if kw == &Keyword::GO {
+            let mut look_back_count = 2;
```
iffyio, line 125: double checking, why does the index start at 2 vs 1?

src/dialect/mssql.rs
```diff
@@ -116,7 +116,29 @@ impl Dialect for MsSqlDialect {
...
+            loop {
```
iffyio, line 126: Can we have this logic as a helper function?
``` rust
fn prev_nth_token_no_skip(&self, mut n: usize) -> Option<&TokenWithSpan> {
  // ...
}
```
like a combination of [peek_nth_token](https://github.com/apache/datafusion-sqlparser-rs/blob/ed181fde55fa090b42f047bb0fbb84cf494904ef/src/parser/mod.rs#L3922C5-L3922C69) and [prev_token](https://github.com/apache/datafusion-sqlparser-rs/blob/ed181fde55fa090b42f047bb0fbb84cf494904ef/src/parser/mod.rs#L4034)

src/dialect/mssql.rs
aharpervc: It's similar to parse_go, which references the existing previous token logic which also starts by subtracting 2. My understanding is that 0 is the next token, -1 is the current token, and -2 is the first previous token.

src/parser/mod.rs
aharpervc: Yes, I have something like that already for parsing the batch count, which doesn't need to look backwards and already uses peek_token_no_skip. I think between here & your other comment on the alias function, the concern is regarding the newline logic which looks backwards. I'll try and move that into a helper so it can be simplified.

tests/sqlparser_mssql.rs
aharpervc: I'm unsure the best way to approach this question. `one_statement_parses_to` requires _one_ statement, right? But most of these SQL text examples are multiple statements. I didn't see anything in TestUtils for a series of statements, so I added a new helper accordingly with canonical strings, let me know if that covers it

tests/sqlparser_mssql.rs
aharpervc: Done 👍

src/dialect/mssql.rs
aharpervc: I added a new helper, done 👍

src/ast/mod.rs
iffyio: Ah indeed, ideally both would have the doc comments on their respective structs, I think I meant to reference return  statement in the other PR as the example

src/test_utils.rs
```diff
@@ -166,6 +168,32 @@ impl TestedDialects {
         only_statement
     }
 
+    /// The same as [`one_statement_parses_to`] but it works for a multiple statements
+    pub fn multiple_statements_parse_to(
```
iffyio, line 172:
```suggestion
    pub fn statements_parse_to(
```

tests/sqlparser_mssql.rs
iffyio: Ah yeah a new helper sounds reasonable to me thanks!

src/parser/mod.rs
```diff
@@ -4055,6 +4070,44 @@ impl<'a> Parser<'a> {
         )
     }
 
+    /// Look backwards in the token stream and expect that there was only whitespace tokens until the previous newline
+    pub fn expect_previously_only_whitespace_until_newline(&mut self) -> Result<(), ParserError> {
```
iffyio, line 4074:
```suggestion
    pub(crate) fn expect_previously_only_whitespace_until_newline(&mut self) -> Result<(), ParserError> {
```

src/parser/mod.rs
```diff
@@ -618,6 +632,7 @@ impl<'a> Parser<'a> {
                 // `COMMENT` is snowflake specific https://docs.snowflake.com/en/sql-reference/sql/comment
                 Keyword::COMMENT if self.dialect.supports_comment_on() => self.parse_comment(),
                 Keyword::PRINT => self.parse_print(),
+                Keyword::GO => self.parse_go(),
```
iffyio, line 635: can we rewind with `self.prev_token()` here before calling parse_go? so that the `parse_go` function is self contained and can parse a full GO statement. (I also realise now we missed that for print)

src/parser/mod.rs
```diff
@@ -15064,6 +15117,38 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    /// Parse [Statement::Go]
+    fn parse_go(&mut self) -> Result<Statement, ParserError> {
+        self.expect_previously_only_whitespace_until_newline()?;
+
+        let count = loop {
+            // using this peek function because we want to halt this statement parsing upon newline
```
iffyio, line 15125: Can we include the example you had in the comment earlier, explicitly highlighting why this statement is special? I think otherwise it would not be obvious to folks that come across the code later on why the current code is a special case

src/parser/mod.rs
```diff
@@ -15064,6 +15117,38 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    /// Parse [Statement::Go]
+    fn parse_go(&mut self) -> Result<Statement, ParserError> {
+        self.expect_previously_only_whitespace_until_newline()?;
```
iffyio, line 15122: hmm this call look equivalent to the loop below that peeks forward to skip newlines? unclear why we're peeking backwards here right after we parsed the GO keyword (so that this will always be infallible?)

src/dialect/mssql.rs
```diff
@@ -116,7 +116,17 @@ impl Dialect for MsSqlDialect {
...
     }
 
-    fn is_column_alias(&self, kw: &Keyword, _parser: &mut Parser) -> bool {
+    fn is_column_alias(&self, kw: &Keyword, parser: &mut Parser) -> bool {
+        // if we find maybe whitespace then a newline looking backward, then `GO` ISN'T a column alias
+        // if we can't find a newline then we assume that `GO` IS a column alias
+        if kw == &Keyword::GO
+            && parser
+                .expect_previously_only_whitespace_until_newline()
+                .is_ok()
```
iffyio, line 125: if the intent is for the condition to return a bool, then we can update this function to return a bool instead of a result (where the latter is flagging an error to be handled)

src/parser/mod.rs
iffyio, line 4074: hmm so i think it would be good to write this function in a reusable manner if possible since it feels similar to the `peek_*`, `prev_*` helper functions that we have. Would it be possible to introduce a generic `prev_nth_token_no_skip(n: usize)` that can be called by parser and dialect methods?

src/parser/mod.rs
aharpervc: I see that some of the statement types in parse_statement are doing that but most aren't. What's the strategy on this? How do you know when you're supposed to do it or not?

src/parser/mod.rs
aharpervc: Yes, I can add that comment. However there is also test coverage for this behavior, so it should be protected from future refactoring

src/dialect/mssql.rs
aharpervc: In this case, it's most useful to know if there was previously only whitespace until a newline. However in the other call of this function in parse_go, we legitimately do want a parser error. Therefore we have choices:

1. expect_previously_only_whitespace_until_newline raises a parse error which makes sense in parse_go but is converted to a boolean here
2. expect_previously_only_whitespace_until_newline returns a boolean which makes sense here and then parse_go needs to separately raise a parse error

Which is the best choice for the time being?

src/parser/mod.rs

src/parser/mod.rs
aharpervc: Done 👍

src/parser/mod.rs

src/test_utils.rs
aharpervc: Done 👍

src/parser/mod.rs
iffyio: Ah that makes sense! And the added comment clarifies as well thanks!

src/parser/mod.rs
iffyio: Indeed the code is inconsistent in that regard historically, for newer statements ideally the parser functions are self contained when possible

src/parser/mod.rs
```diff
@@ -484,8 +488,18 @@ impl<'a> Parser<'a> {
...
 
             let statement = self.parse_statement()?;
+            expecting_statement_delimiter = match &statement {
+                Statement::If(s) => match &s.if_block.conditional_statements {
+                    // the `END` keyword doesn't need to be followed by a statement delimiter, so it shouldn't be expected here
+                    ConditionalStatements::BeginEnd { .. } => false,
+                    // parsing the statement sequence consumes the statement delimiter, so it shouldn't be expected here
+                    ConditionalStatements::Sequence { .. } => false,
+                },
+                // Treat batch delimiter as an end of statement, so no additional statement delimiter expected here
```
iffyio, line 498: will this be replaced by similar solution as in #1808?

src/parser/mod.rs
```diff
@@ -4055,6 +4090,38 @@ impl<'a> Parser<'a> {
...
+        let mut look_back_count = 1;
+        loop {
+            let prev_token = self.peek_prev_nth_token_no_skip(look_back_count);
+            match prev_token.token {
+                Token::EOF => break,
+                Token::Whitespace(ref w) => match w {
+                    Whitespace::Newline => break,
+                    // special consideration required for single line comments since that string includes the newline
+                    Whitespace::SingleLineComment { comment, prefix: _ } => {
+                        if comment.ends_with('\n') {
```
iffyio, line 4106: double checking: is there a scenario where a single line comment doesn't end with a new line? spontaneously sounds like that should always hold true so that the manual newline check would not be required

src/parser/mod.rs
```diff
@@ -3939,6 +3954,26 @@ impl<'a> Parser<'a> {
             })
     }
 
+    /// Return nth previous token, possibly whitespace
+    /// (or [`Token::EOF`] when before the beginning of the stream).
+    pub fn peek_prev_nth_token_no_skip(&self, n: usize) -> TokenWithSpan {
```
iffyio, line 3959:
```suggestion
    pub(crate) fn peek_prev_nth_token_no_skip(&self, n: usize) -> &TokenWithSpan {
```
we can probably start with the API being non-public. Then can we switch to returning a reference to avoid the mandatory cloning?

src/dialect/mssql.rs
iffyio: Ah I see, 2. sounds desirable to return a boolean, in that the function is only checking if a condition holds, then interpretation of the condition is left to the caller of the function. in the other usage the caller does e.g.
```rust
if !self.prev_whitespace_only_is_newline() {
    return self.expected("newline ...")
}
```

aharpervc: Gotcha, I will update it to do that

aharpervc:
Here's another example case this PR should parse properly, before merging (on my todo list...)

```
USE some_database;
GO

;WITH cte AS (
  SELECT 1 x
)
SELECT *
FROM cte;
```

The peek logic in parse_go should be seeing that newline, but it's accidentally advancing past the whitespace, seeing the semi-colon, and failing.

src/parser/mod.rs
aharpervc: > double checking: is there a scenario where a single line comment doesn't end with a new line? spontaneously sounds like that should always hold true so that the manual newline check would not be required

I actually don't know, I was surprised to find that the newline is actually part of the comment text in this library. It seemed prudent to be defensive in this new code, since that comment parsing behavior maybe isn't particularly intentional.

src/dialect/mssql.rs
aharpervc: Done 👍

src/parser/mod.rs
aharpervc: Thanks, that makes sense. I've updated it to rewind before parsing (using raiserror as a reference)

src/parser/mod.rs

src/parser/mod.rs
aharpervc: I suspect so... need to investigate. (I need to rebuild my mental model on this first 😆).

aharpervc:
@iffyio anything else on this one?

src/parser/mod.rs
```diff
@@ -475,6 +475,10 @@ impl<'a> Parser<'a> {
                     if expecting_statement_delimiter && word.keyword == Keyword::END {
                         break;
                     }
+
```
iffyio, line 478: @aharpervc Sorry for the delay getting back to this, actually I'm leaning towards not merging this feature after all unfortunately. It feels to me like there's a bit too many special cases (like this extending the core parsing loop for example) for this feature that go against the normal of sql statement parsing. And combined with this feature not being part of any sql reference but rather a specific tool, it makes the tradeoff undesirable

iffyio:
Closing this PR per my comment in https://github.com/apache/datafusion-sqlparser-rs/pull/1809#discussion_r2072719447
Thanks again @aharpervc for working on this and sorry for the confusion :pray:

aharpervc:
> Closing this PR per my comment in [#1809 (comment)](https://github.com/apache/datafusion-sqlparser-rs/pull/1809#discussion_r2072719447) Thanks again @aharpervc for working on this and sorry for the confusion 🙏

I saw your comment but hadn't had a chance to get back to you. I'd like to get a clarification -- are you closing the PR because you're opposed to supporting batch delimiters _at all_ in this library, or is it that you don't want to accept this particular approach?

If it's the latter, I'm completely open to continuing to iterate the branch. However if you don't want batch delimiters at all, that's disappointing -- it means this library will never be able to process real world SQL Server code, since tools on that stack all know how to process batch delimiters.

iffyio:
> are you closing the PR because you're opposed to supporting batch delimiters at all in this library, or is it that you don't want to accept this particular approach

@aharpervc ah to clarify I'm not opposed to supporting the feature itself - my thinking is mainly that the feature is out of scope for the parser since it is not a sql construct, so that supporting it would be at best a nice to have, then had it been easy to include that support we could do so. But afaict so far in this PR, that unfortunately doesn't seem to be the case because there's been a few special case code being introduced to the parser.

Basically that the complexity being by the PR from my pov isn't worth the return given that this feature isn't meant to be supported by the parser in the first place.

Maybe if the code is much simpler and touches less areas (most importantly I think the core parser loop I highlighted in the previous comment), we could merge such a patch, but I'm unable to say for sure what that looks like, code wise, from my end.

But actually I think if we can come up with something that doesn't touch that core parser loop we might have something mergeable. I'll reopen this PR, feel free to iterate if you have ideas to try out and I'm happy to take a look when you do!

aharpervc:
> But actually I think if we can come up with something that doesn't touch that core parser loop we might have something mergeable. I'll reopen this PR, feel free to iterate if you have ideas to try out and I'm happy to take a look when you do!

Sounds good... I will brainstorm on this 👍.

---

It may also be useful to clarify what is my overall objective... basically, I think it'd be useful read a whole file into a string, then pass it to this library for parsing, then process the AST's. Since SQL code I work with (and in the industry) often includes `GO` batch delimiters in the file source code, and this library doesn't support batch delimiters outside of this PR, that presented a difficulty. Initially I was thinking that the smoothest approach would be to add batch delimiter support directly into the parser. However, this discussion is making me consider the pro/con more closely.

Something else I was thinking about is how the current "parse statements" API _implies_ that you're only giving the parser sql of a _single batch_. Eg, `single batch string` -> `Vec<Statement>` makes complete sense in this model. However this PR introduced a much more odd model, where a single string could actually be multiple batches. Eg, `Vec<Batch>` -> `Vec<Statements>` is a strange API.

So I think there could be a couple possible approaches going forward:

_Option 1: Embrace Batches_

We could introduce a Batch struct & new top level `String` -> `Vec<Batch>`, then `Batch` -> `Vec<Statements>`. `Batch` could be like

```rust
struct Batch {
  statements: Vec<Statement>,
  count: u16 // or whatever
}
```

This seems useful because it makes explicit what was formerly implicit. Probably all the existing top level API's could be left alone, so all existing users uninterested in Batches would be unaffected. Then we'd want something like `parse_batches` to return a `Vec<Batch>` and then do all the normal parsing.

The downside is, it would introduce a new concept & ongoing maintenance when perhaps there isn't a lot of interest. Also, not every dialect even has the concept of a batch, I think, so that's a difficulty here too.

_Option 2: Reject Batches_

Close this PR permanently & open a new PR, that would basically be a docs/readme update, to clarify what is the expected string users should provide to parse_sql & similar API's. The clarification is it needs to be the contents of a SQL batch, since that is the thing that can clearly be turned into a single `Vec<Statement>`. Dialects such as SQL Server which include batch delimiters in the SQL source code should be processed _outside_ of this library to split the batches, then the batch string can be given to sqlparser-rs.

The downside is back to that original objective, which is reading a whole SQL file to a string then passing it to this library for parsing. If library users have to do our own detection of the batch delimiter keyword, now we all have to implement a mini parser before using the actual parser. That seems really unfortunate.

_Option 3: Hold Batches out at arm's length_ (thesis, antithesis, ...synthesis)

I don't have a clear vision on this option yet, but it seems like this library is well positioned to do _some_ work on this topic, but perhaps not as much as full on Batch support. What if we had something like "parse_batches", but it returned a simple `Vec<String>`? I'm imagining something like:

```rust
fn split_batches -> Vec<String> {
  let mut result = vec![];
  while let Some(token) = self.tokens.next() {
    if peek_keyword(Keyword::GO) {
      self.prev_token();
      result.push(parse_go());
    }
  }
}

```

The concept here is the library has the ability to scan through the tokens & parse the GO keyword, and split a single string into Vec<String> but would otherwise remain unchanged. Dialects that support batches could use this API to turn a whole file string into something suitable for parse_statements. Other dialects would remain unaffected.

---

I'd be interested to hear anyone else's opinion on this too, for/against any option or to brainstorm other options.

iffyio:
> I think it'd be useful read a whole file into a string, then pass it to this library for parsing, then process the AST's

Not sure I necessarily agree with this sentiment. I think if there are constructs in a file that arent sql constructs, they should be stripped off from the file before parsing to the library. Parsing non-sql constructs would be out of scope for this library.

As a result, similar to ideally avoiding changes to the core parsing loop, I think it would also be undesirable to change or introduce parser APIs in order to support this feature. My current thinking is that if supporting the feature is a low hanging fruit that doesn't introduce too much complexity to the codebase, then we could look to support it as a one-off. Beyond that scenario afaict, downstream users of the library are expected strip off GO lines from input files or do similar preprocessing, before passing input onto the library.

aharpervc:
> Not sure I necessarily agree with this sentiment. I think if there are constructs in a file that arent sql constructs, they should be stripped off from the file before parsing to the library. Parsing non-sql constructs would be out of scope for this library.

I actually agree with your perspective here. The central question is, "is the `go` batch delimiter found in SQL files 'a SQL construct' or not"? (Then we have a separate question on the implementation, which is the options list I suggested above).

For illustration, something I think we'd both agree is _not_ a SQL construct & therefore _not_ appropriate for this parse is [Template Syntax](https://learn.microsoft.com/en-us/ssms/template/templates-ssms). That seems pretty clearly something that should be filled in prior to expecting a SQL document to parse.

For `go`, here's what I can say:

| evidence | a SQL construct | not a SQL construct |
|---|---|---|
| [documented as a "statement" alongside other SQL statements](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go) & indistinguishable from a statement in code | ✅ | |
| documented as ["commands that are not Transact-SQL statements"](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/sql-server-utilities-statements-go) & "GO is not a Transact-SQL statement" | | ✅ |
| [SQL Server's ManagedBatchParser parses it](https://learn.microsoft.com/en-us/dotnet/api/managedbatchparser.parser.setbatchdelimiter?view=sqlserver-2016) (actually, configurable (!)) | ✅ | |
| Apps need to parse SQL to find `go`'s (this library or another) | ✅ | ✅ |

src/parser/mod.rs
```diff
@@ -475,6 +475,10 @@ impl<'a> Parser<'a> {
...
+                    if expecting_statement_delimiter && word.keyword == Keyword::GO {
```
iffyio, line 479: Can we add a clarifying comment here, mentioning that for the `GO` statement, it is not necessary that the preceeding statement ends with a semicolon (ideally with a sql example)?

src/parser/mod.rs
```diff
@@ -484,8 +488,9 @@ impl<'a> Parser<'a> {
             }
 
             let statement = self.parse_statement()?;
+            // Treat batch delimiter as an end of statement, so no additional statement delimiter expected here
```
iffyio, line 491:
```suggestion
            // The `GO` statement optionally ends with a semicolon
```
thinking of a this more straightforward description, it would not be obvious to a reader unfamiliar with the statement that 'batch delimiter' is the same thing in this context

iffyio (commented):
Left a couple comments to document the changes to the parser loop

src/parser/mod.rs

src/parser/mod.rs
aharpervc: Done 👍

aharpervc:
Apologies for the (even longer!) delay on this PR. I have rebase on latest main and addressed the single remaining code comment discussion. I think this is finally ready for review @iffyio @alamb

aharpervc:
Still waiting on review 🤞

aharpervc:
Even though this PR went stale, the branch is still available on my fork and works as expected

PR #1810, opened 2025-04-11
Add `CREATE TRIGGER` support for SQL Server

aharpervc:
Adjacent to: https://github.com/apache/datafusion-sqlparser-rs/pull/1808 with similar considerations, and temporarily rebased on it (eg, this branch should probably wait for that branch to merge).

---

This PR introduces support for parsing `CREATE TRIGGER` for SQL Server.

The main concern is that for the existing dialects, there was an expectation of an `EXECUTE` keyword (PG: https://www.postgresql.org/docs/current/sql-createtrigger.html). However, SQL Server doesn't use this syntax and instead supports multi statement blocks (like a stored procedure).

The difficulty here in the codebase is what to do about CreateTrigger.exec_body & TriggerExecBody. In this iteration I made the property an `Option`, which seemed like the least impact on existing code. TriggerExecBody, etc can be left alone.

However in the future I think this could be improved with a more broad refactoring, such as a TriggerBody enum, which can consolidate TriggerExecBodyType's options & use `Vec<Statements>` for a `MultiStatement` variant and `FunctionDesc` for `Function` & `Procedure` variants. That'd basically remove the TriggerExecBody struct, too. Overall I think that'd be a cleaner approach for the CreateTrigger struct.

--- 

That's all speculative followup work, for now I just want to be able to parse most SQL Server triggers. Other SQL Server specific things like particular trigger options, DDL trigger stuff, etc can come later as needed.

### DIFF
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -246,16 +246,14 @@ impl MsSqlDialect {
             let statements = parser.parse_statement_list(&[Keyword::END])?;
             let end_token = parser.expect_keyword(Keyword::END)?;
 
-            BeginEndStatements {
+            ConditionalStatements::BeginEnd(BeginEndStatements {
                 begin_token: AttachedToken(begin_token),
                 statements,
                 end_token: AttachedToken(end_token),
-            }
+            })
         } else {
-            BeginEndStatements {
-                begin_token: AttachedToken::empty(),
+            ConditionalStatements::Sequence {
                 statements: parser.parse_statements()?,
-                end_token: AttachedToken::empty(),
             }
         };
 
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -240,22 +240,7 @@ impl MsSqlDialect {
         let events = parser.parse_comma_separated(Parser::parse_trigger_event)?;
 
         parser.expect_keyword_is(Keyword::AS)?;
-
-        let trigger_statements_body = if parser.peek_keyword(Keyword::BEGIN) {
-            let begin_token = parser.expect_keyword(Keyword::BEGIN)?;
-            let statements = parser.parse_statement_list(&[Keyword::END])?;
-            let end_token = parser.expect_keyword(Keyword::END)?;
-
-            ConditionalStatements::BeginEnd(BeginEndStatements {
-                begin_token: AttachedToken(begin_token),
-                statements,
-                end_token: AttachedToken(end_token),
-            })
-        } else {
-            ConditionalStatements::Sequence {
-                statements: parser.parse_statements()?,
-            }
-        };
+        let statements = Some(parser.parse_conditional_statements(&[Keyword::END])?);
 
         Ok(Statement::CreateTrigger {
             or_alter,
@@ -271,7 +256,7 @@ impl MsSqlDialect {
             include_each: false,
             condition: None,
             exec_body: None,
-            statements: Some(trigger_statements_body),
+            statements,
             characteristics: None,
         })
     }
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -254,7 +254,7 @@ impl MsSqlDialect {
         } else {
             BeginEndStatements {
                 begin_token: AttachedToken::empty(),
-                statements: vec![parser.parse_statement()?],
+                statements: parser.parse_statements()?,
                 end_token: AttachedToken::empty(),
             }
         };
--- a/src/dialect/mssql.rs
+++ b/src/dialect/mssql.rs
@@ -18,6 +18,7 @@
 use crate::ast::helpers::attached_token::AttachedToken;
 use crate::ast::{
     BeginEndStatements, ConditionalStatementBlock, ConditionalStatements, IfStatement, Statement,
+    TriggerObject,
 };
 use crate::dialect::Dialect;
 use crate::keywords::{self, Keyword};
@@ -125,6 +126,15 @@ impl Dialect for MsSqlDialect {
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.peek_keyword(Keyword::IF) {
             Some(self.parse_if_stmt(parser))
+        } else if parser.parse_keywords(&[Keyword::CREATE, Keyword::TRIGGER]) {
+            Some(self.parse_create_trigger(parser, false))
+        } else if parser.parse_keywords(&[
+            Keyword::CREATE,
+            Keyword::OR,
+            Keyword::ALTER,
+            Keyword::TRIGGER,
+        ]) {
+            Some(self.parse_create_trigger(parser, true))
         } else {
             None
         }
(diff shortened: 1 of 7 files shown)


src/parser/mod.rs
```diff
@@ -5265,18 +5271,71 @@ impl<'a> Parser<'a> {
...
+
+        let statements = if self.peek_keyword(Keyword::BEGIN) {
+            self.next_token();
+            let mut result = self.parse_statements()?;
+            // note: `parse_statements` will consume the `END` token & produce a Commit statement...
+            if let Some(Statement::Commit{ chain, end, modifier }) = result.last() {
+                if *chain == false && *end == true && *modifier == None {
+                    result = result[..result.len() - 1].to_vec();
+                }
+            }
```
aharpervc, line 5305: I fully realize this is goofy, but it seems like a temporarily reasonable workaround for the current difficulty on parsing multi statement begin/end blocks. Perhaps there's an obvious/easier way?

aharpervc:
This fails to parse on this branch, but should:

```mssql
CREATE TRIGGER some_trigger ON some_table FOR INSERT
AS
BEGIN
    IF 1=1
    BEGIN
        RAISERROR('Trigger fired', 10, 1);
    END

    RETURN;
END
```

aharpervc:
FYI for reviewers I rebased this on the create function branch, due to the return statement logic here: https://github.com/apache/datafusion-sqlparser-rs/pull/1808#discussion_r2042434307

aharpervc:
Rebased again now that https://github.com/apache/datafusion-sqlparser-rs/pull/1808 has been merged. Should be ready for review.

src/parser/mod.rs
```diff
@@ -5256,14 +5256,19 @@ impl<'a> Parser<'a> {
...
     ) -> Result<Statement, ParserError> {
-        if !dialect_of!(self is PostgreSqlDialect | GenericDialect | MySqlDialect) {
+        if !dialect_of!(self is PostgreSqlDialect | GenericDialect | MySqlDialect | MsSqlDialect) {
             self.prev_token();
             return self.expected("an object type after CREATE", self.peek_token());
         }
 
+        if dialect_of!(self is MsSqlDialect) {
+            return self.parse_mssql_create_trigger(or_alter, or_replace, is_constraint);
+        }
```
iffyio, line 5270: maybe this could be parsed via `self.dialect.parse_statement()` for the mssql version? since the dialect_of macro we try to stay away for new code

aharpervc: That seems tricky. Can you help me understand what that code would look like? parse_create_trigger is happening inside a complicated branch of parse_create, so it seems like that logic from src/parser/mod.rs would need to be copy/pasted to src/dialect/mssql.rs which seems unfortunate.

Otherwise I don't know how to have the dialect's parse_statement implement custom logic only for "create trigger" but not for all the other "create xxx"'s

iffyio: ah yeah it would need to manually parse the alter, replace for the trigger, I agree that its not ideal but its probably the lesser evil, see [snowflake here for an example](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/snowflake.rs#L144)

iffyio: ah yeah it would need to manually parse the alter, replace for the trigger, I agree that its not ideal but its probably the lesser evil, see [snowflake here for an example](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/snowflake.rs#L144)

iffyio: ah yeah it would need to manually parse the alter, replace for the trigger, I agree that its not ideal but its probably the lesser evil, see [snowflake here for an example](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/snowflake.rs#L144)r

src/dialect/mssql.rs
```diff
@@ -215,6 +218,78 @@ impl MsSqlDialect {
         }))
     }
 
+    /// Parse a SQL CREATE statement
+    fn parse_create(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
+        let original_index = parser.index();
+
+        if !parser.parse_keyword(Keyword::CREATE) {
+            parser.set_index(original_index);
```
iffyio, line 226: I think instead of manually tracking index we can rely on the `parser.maybe_parse()`? that will automatically handle the indexing

aharpervc: I've added a commit to the branch for this approach.

I'm not sure it's strictly better because it seems more difficult to extend if/when additional `CREATE` things are moved into dialects, if this is a general pattern. Perhaps you can clear up what I'm concerned about here -- with maybe_parse if parsing fails inside the dialect, it'll rewind the parser & report None up the chain. Then, parsing will continue as if it was common code. However that might not be what we prefer -- if we're already into parse_create_trigger and get a parsing error, we actually do want it to fail right there rather than rewinding. In the former approach that seemed clear. With this approach it seems like the dialect specific parser failure will get ignored (and you will probably get another parser error from the general code instead).

Thoughts?

iffyio: Ah yes you're right, if failure occurs while parsing the trigger body for example the ideal behavior in this case would be to report that error instead of returning None. In which case maybe_parse isn't ideal.

I think the main goal is avoiding a custom implementation to track indexing here, looking at the new diff it seems the  sqlserver version is simpler so maybe something like this suffices already?
```rust
if self.parse_keywords(CREATE, TRIGGER) {
  Some(self.parse_create_trigger(self, parser, false))
} else if self.parse_keywords(CREATE, OR, ALTER, TRIGGER) {
  Some(self.parse_create_trigger(self, parser, true))
} else {
  None
}
```

aharpervc: Great, I'll drop the maybe_parse commit & update the implementation to this two branch approach. I think that will meet both our expectations for the time being

src/dialect/mssql.rs
```diff
@@ -215,6 +225,59 @@ impl MsSqlDialect {
...
+    ) -> Result<Statement, ParserError> {
+        let name = parser.parse_object_name(false)?;
+        parser.expect_keyword_is(Keyword::ON)?;
+        let table_name = parser.parse_object_name(false)?;
+        let period = parser.parse_trigger_period()?;
+        let events = parser.parse_comma_separated(Parser::parse_trigger_event)?;
+
+        parser.expect_keyword_is(Keyword::AS)?;
+
+        let trigger_statements_body = if parser.peek_keyword(Keyword::BEGIN) {
```
aharpervc, line 244: Similar approach in the other PR for stored procedures that may/not have begin/end tokens https://github.com/apache/datafusion-sqlparser-rs/pull/1834/files#r2069458988

src/dialect/mssql.rs
```diff
@@ -215,6 +225,59 @@ impl MsSqlDialect {
...
+                begin_token: AttachedToken(begin_token),
+                statements,
+                end_token: AttachedToken(end_token),
+            }
+        } else {
+            BeginEndStatements {
+                begin_token: AttachedToken::empty(),
+                statements: parser.parse_statements()?,
+                end_token: AttachedToken::empty(),
+            }
```
iffyio, line 259: Oh I think instead of representing the begin optionality using tokens, we can use an enum, something similar to [ConditionalStatements](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/ast/mod.rs#L2294-L2299) where we have an explicit `BeginEnd` enum variant

aharpervc: Gotcha, yeah I could see it either way. I will update this to use the enum pattern

aharpervc: Done 👍

src/dialect/mssql.rs
aharpervc: I also updated the related stored procedure PR accordingly: https://github.com/apache/datafusion-sqlparser-rs/pull/1834#issuecomment-2847567460

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1811, opened 2025-04-11
Add support for `PRINT` statement for SQL Server

aharpervc:
Reference: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/print-transact-sql?view=sql-server-ver16

Making `message` a `Box<Expr>` instead of an enum of (national) string literal, variable, or expression is _slightly_ lazy and also seems completely acceptable for the time being.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4054,6 +4054,12 @@ pub enum Statement {
         arguments: Vec<Expr>,
         options: Vec<RaisErrorOption>,
     },
+    /// ```sql
+    /// PRINT msg_str | @local_variable | string_expr
+    /// ```
+    ///
+    /// See: <https://learn.microsoft.com/en-us/sql/t-sql/statements/print-transact-sql>
+    Print(PrintStatement),
 }
 
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
@@ -5745,7 +5751,7 @@ impl fmt::Display for Statement {
                 }
                 Ok(())
             }
-
+            Statement::Print(s) => write!(f, "{s}"),
             Statement::List(command) => write!(f, "LIST {command}"),
             Statement::Remove(command) => write!(f, "REMOVE {command}"),
         }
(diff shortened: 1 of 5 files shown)


tests/sqlparser_mssql.rs
```diff
@@ -2036,3 +2036,37 @@ fn parse_mssql_merge_with_output() {
         OUTPUT $action, deleted.ProductID INTO dsi.temp_products";
     ms_and_generic().verified_stmt(stmt);
 }
+
+#[test]
+fn parse_print() {
```
aharpervc, line 2041: Do we need a test that asserts a particular error for a bare `PRINT`? Not sure what the conventions are for that

src/ast/mod.rs
```diff
@@ -4050,6 +4050,14 @@ pub enum Statement {
...
         options: Vec<RaisErrorOption>,
     },
+    /// ```sql
+    /// PRINT msg_str | @local_variable | string_expr
+    /// ```
+    ///
+    /// See: https://learn.microsoft.com/en-us/sql/t-sql/statements/print-transact-sql
+    Print {
+        message: Box<Expr>,
+    },
```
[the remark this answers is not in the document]
iffyio: Yeah we can use a separate struct for repr

src/parser/mod.rs
```diff
@@ -617,6 +617,9 @@ impl<'a> Parser<'a> {
                 }
                 // `COMMENT` is snowflake specific https://docs.snowflake.com/en/sql-reference/sql/comment
                 Keyword::COMMENT if self.dialect.supports_comment_on() => self.parse_comment(),
+                Keyword::PRINT if dialect_of!(self is MsSqlDialect) => {
```
iffyio, line 620:
```suggestion
                Keyword::PRINT => {
```
we can parse unconditionally?

src/ast/mod.rs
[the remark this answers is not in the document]
aharpervc: Done

src/parser/mod.rs

aharpervc:
Rebased & updated

tests/sqlparser_mssql.rs
```diff
@@ -2053,3 +2053,37 @@ fn parse_drop_trigger() {
...
+    );
+
+    let print_variable = "PRINT @my_variable";
+    let print_stmt = ms().one_statement_parses_to(print_variable, "");
+    assert_eq!(
+        print_stmt,
+        Statement::Print(PrintStatement {
+            message: Box::new(Expr::Identifier(Ident::new("@my_variable"))),
+        })
+    );
```
iffyio, line 2088: oh was there a requirement to use `one_statement_parses_to` instead of `verified_stmt` for these tests?

Also with vierified_stmt we can simplify the tests except, for the hello world example, to skip assertion on the AST

aharpervc: I'm unaware of any requirement, it's just that I don't understand the project testing conventions & I'm trying to follow the patterns already present in the file. I'll switch it over to your suggestion

aharpervc: I think we might want to inspect the ast in the future if we do an enum of value params instead of an expr, but we can cross that bridge if/when we ever come to it. This suggestion is complete 👍

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1812, opened 2025-04-14
Fix `STRAIGHT_JOIN` constraint when table alias is absent

killertux:
In https://github.com/apache/datafusion-sqlparser-rs/pull/1802 was added support for MySQL STRAIGHT_JOIN, but it was not added as a RESERVED_FOR_TABLE_ALIAS. Because of this, it fails to parse queries where we are not using an alias.

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -3715,10 +3715,7 @@ fn parse_straight_join() {
     mysql().verified_stmt(
         "SELECT a.*, b.* FROM table_a AS a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id",
     );
-}
-
-#[test]
-fn parse_straight_join_without_alias() {
+    // Without table alias
     mysql()
         .verified_stmt("SELECT a.*, b.* FROM table_a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id");
 }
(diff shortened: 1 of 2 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -3716,3 +3716,9 @@ fn parse_straight_join() {
         "SELECT a.*, b.* FROM table_a AS a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id",
     );
 }
+
+#[test]
+fn parse_straight_join_without_alias() {
+    mysql()
+        .verified_stmt("SELECT a.*, b.* FROM table_a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id");
+}
```
iffyio, line 3724:
```suggestion
    // Without table alias
    mysql()
        .verified_stmt("SELECT a.*, b.* FROM table_a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id");
}
```
thinking we can merge this scenario into the existing test case for the straight join feature?

iffyio (commented):
Thanks @killertux! Left one minor comment to merge the test cases otherwise this looks good to me

tests/sqlparser_mysql.rs
killertux: Done.

iffyio (approved):
Thanks @killertux!
cc @alamb

PR #1815, opened 2025-04-17
enable `supports_filter_during_aggregation` for Generic dialect

goldmedal:
- Related to https://github.com/apache/datafusion/issues/15719

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -11642,6 +11642,20 @@ fn parse_connect_by() {
 
 #[test]
 fn test_selective_aggregation() {
+    let testing_dialects = all_dialects_where(|d| d.supports_filter_during_aggregation());
+    let expected_dialects: Vec<Box<dyn Dialect>> = vec![
+        Box::new(PostgreSqlDialect {}),
+        Box::new(DatabricksDialect {}),
+        Box::new(HiveDialect {}),
+        Box::new(SQLiteDialect {}),
+        Box::new(DuckDbDialect {}),
+        Box::new(GenericDialect {}),
+    ];
+    assert_eq!(testing_dialects.dialects.len(), expected_dialects.len());
+    expected_dialects
+        .into_iter()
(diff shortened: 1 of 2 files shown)


goldmedal:
@alamb, What do you think about this? I think this change is related to the default behavior of DataFusion.

goldmedal:
> The only thing I think this PR needs is a test and it should be good to go

oops , I forgot to mention the tests. I found it will be covered by the original case:
https://github.com/apache/datafusion-sqlparser-rs/blob/81d8909e006f90c9b2f06fdeb4bab22935bec827/tests/sqlparser_common.rs#L11644-L11655

I guess it's enough?

alamb:
> oops , I forgot to mention the tests. I found it will be covered by the original case:

I was thinking a test that would prevent against regressions in a future refactor -- namely a test that will fail if this code change is reverted, 

I think as it stands, if the code change is reverted all the tests will still pass 🤔

goldmedal:
> > oops , I forgot to mention the tests. I found it will be covered by the original case:
> 
> I was thinking a test that would prevent against regressions in a future refactor -- namely a test that will fail if this code change is reverted,
> 
> I think as it stands, if the code change is reverted all the tests will still pass 🤔

I see. Make sense. I'll add another test to protect it.

tests/sqlparser_common.rs
```diff
@@ -11642,16 +11642,28 @@ fn parse_connect_by() {
...
+        Box::new(DatabricksDialect {}),
+        Box::new(HiveDialect {}),
+        Box::new(SQLiteDialect {}),
+        Box::new(DuckDbDialect {}),
+        Box::new(GenericDialect {}),
+    ];
+    assert_eq!(testing_dialects.dialects.len(), expected_dialects.len());
+    expected_dialects
+        .into_iter()
+        .for_each(|d| assert!(d.supports_filter_during_aggregation()));
```
goldmedal, line 11657: I list all the dialects expected to support filtering during aggregation.

alamb (approved):
Thank you @goldmedal 

FYI @iffyio

PR #1818, opened 2025-04-21
Add `OR ALTER` support for `CREATE VIEW`

aharpervc:
Reference: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-view-transact-sql

This PR adds an `or_alter` field to the CreateView struct, which is false by default but true for `CREATE OR ALTER`. This syntax is used by SQL Server, but implemented here generically. There is a `verified_stmt` test added to the existing "create view" test, to assert this field round trips as expected.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3037,6 +3037,10 @@ pub enum Statement {
     /// CREATE VIEW
     /// ```
     CreateView {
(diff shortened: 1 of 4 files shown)


iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

alamb:
🚀

PR #1822, opened 2025-04-25
Prepare for 0.56.0 release: Version and CHANGELOG

iffyio:
Part of
- #1756 

1. Updates the version
2. Adds a changelog. See [rendered version here](https://github.com/validio-io/sqlparser-rs/blob/prepare_0.56.0/changelog/0.56.0.md)

PR #1823, opened 2025-04-25
Handle missing login in changelog generate script

iffyio:
Noticed during #1822 that the script initially failed due to [this
commit](https://github.com/apache/datafusion-sqlparser-rs/commit/10cf7c164ee0bae8a71e1d8f0af5851b96465692) no longer having the associated GitHub user.

Updates the script to skip the username referencing in such cases.

### DIFF
--- a/dev/release/generate-changelog.py
+++ b/dev/release/generate-changelog.py
@@ -28,7 +28,8 @@ def print_pulls(repo_name, title, pulls):
         print()
         for (pull, commit) in pulls:
             url = "https://github.com/{}/pull/{}".format(repo_name, pull.number)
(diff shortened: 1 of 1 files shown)


dev/release/generate-changelog.py
```diff
@@ -28,7 +28,8 @@ def print_pulls(repo_name, title, pulls):
         print()
         for (pull, commit) in pulls:
             url = "https://github.com/{}/pull/{}".format(repo_name, pull.number)
-            print("- {} [#{}]({}) ({})".format(pull.title, pull.number, url, commit.author.login))
+            author = f"({commit.author.login})" if commit.author else ''
```
alamb, line 31: FYI @andygrove

alamb (approved):
Thanks @iffyio

Issue #1825, opened 2025-04-27
Library fails to parse function


LucaCappelletti94:
Attempting to parse a function such as the following, which in PostgreSQL works fine, currently fails with `ParserError("Expected: ), found: INT")`

```sql
CREATE OR REPLACE FUNCTION check_values_different(int1 INT, int2 INT) RETURNS BOOLEAN AS $$
BEGIN
    IF int1 <> int2 THEN
        RETURN TRUE;
    ELSE
        RETURN FALSE;
    END IF;
END;
$$ LANGUAGE plpgsql;
```

Surprisingly enough, the following instead works:

```sql
CREATE OR REPLACE FUNCTION check_strings_different(str1 VARCHAR, str2 VARCHAR) RETURNS BOOLEAN AS $$
BEGIN
    IF str1 <> str2 THEN
        RETURN TRUE;
    ELSE
        RETURN FALSE;
    END IF;
END;
$$ LANGUAGE plpgsql;
```

And even more weird, the following works, so it must be something regarding the name of the attribute `int1` and `int2` colliding with a type:

```sql
CREATE OR REPLACE FUNCTION check_values_different(a INT, b INT) RETURNS BOOLEAN AS $$
BEGIN
    IF a <> b THEN
        RETURN TRUE;
    ELSE
        RETURN FALSE;
    END IF;
END;
$$ LANGUAGE plpgsql;
```

I will try to fix this error in a PR soon.

PR #1826, opened 2025-04-27
Handle optional datatypes properly in `CREATE FUNCTION` statements

LucaCappelletti94:
This pull request resolves the bug described in issue #1825, which was caused by an incorrect implementation of the named argument parsing. It also adds a few tests to verify that the new implementation is correct.

The previous implementation made the incorrect assumption that arguments name cannot have the same name as types, but the set of types that are parsed as types in `sqlparser` is a superset of the types that are present in each dialect. Therefore, it is correct syntax to use as argument name for instance `int2` for PostgreSQL, while this same argument name would be interpreted as a type elsewhere.

I have changed the parsing to determine via a look-ahead whether the name is a type or not.

Best,
Luca

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5210,12 +5210,9 @@ impl<'a> Parser<'a> {
         // To check whether the first token is a name or a type, we need to
         // peek the next token, which if it is another type keyword, then the
         // first token is a name and not a type in itself.
-        let potential_tokens = [Token::Eq, Token::RParen, Token::Comma];
-        if !self.peek_keyword(Keyword::DEFAULT)
-            && !potential_tokens.contains(&self.peek_token().token)
-        {
+        if let Some(next_data_type) = self.maybe_parse(|parser| parser.parse_data_type())? {
             name = Some(Ident::new(next_token.to_string()));
-            data_type = self.parse_data_type()?;
+            data_type = next_data_type;
         }
 
         let default_expr = if self.parse_keyword(Keyword::DEFAULT) || self.consume_token(&Token::Eq)
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5201,11 +5201,6 @@ impl<'a> Parser<'a> {
         let mut name = None;
         let mut data_type = self.parse_data_type()?;
 
-        // It may appear that the first token can be converted into a known
-        // type, but this could also be a collision as some types are only
-        // present in some dialects and therefore some type reserved keywords
-        // may be freely used as argument names in other dialects.
-
         // To check whether the first token is a name or a type, we need to
         // peek the next token, which if it is another type keyword, then the
         // first token is a name and not a type in itself.
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5199,12 +5199,22 @@ impl<'a> Parser<'a> {
 
         // parse: [ argname ] argtype
         let mut name = None;
+        let next_token = self.peek_token();
         let mut data_type = self.parse_data_type()?;
-        if let DataType::Custom(n, _) = &data_type {
-            // the first token is actually a name
-            match n.0[0].clone() {
-                ObjectNamePart::Identifier(ident) => name = Some(ident),
-            }
+
+        // It may appear that the first token can be converted into a known
+        // type, but this could also be a collision as some types are only
+        // present in some dialects and therefore some type reserved keywords
+        // may be freely used as argument names in other dialects.
+
+        // To check whether the first token is a name or a type, we need to
+        // peek the next token, which if it is another type keyword, then the
+        // first token is a name and not a type in itself.
+        let potential_tokens = [Token::Eq, Token::RParen, Token::Comma];
+        if !self.peek_keyword(Keyword::DEFAULT)
+            && !potential_tokens.contains(&self.peek_token().token)
+        {
+            name = Some(Ident::new(next_token.to_string()));
             data_type = self.parse_data_type()?;
         }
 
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -5199,12 +5199,22 @@ impl<'a> Parser<'a> {
...
+        // may be freely used as argument names in other dialects.
+
+        // To check whether the first token is a name or a type, we need to
+        // peek the next token, which if it is another type keyword, then the
+        // first token is a name and not a type in itself.
+        let potential_tokens = [Token::Eq, Token::RParen, Token::Comma];
+        if !self.peek_keyword(Keyword::DEFAULT)
+            && !potential_tokens.contains(&self.peek_token().token)
+        {
+            name = Some(Ident::new(next_token.to_string()));
```
iffyio, line 5217: wondering if something like this work instead?
```rust
if let DataType::Custom(n, _) = &data_type {
  if let Some(dt) = self.maybe_parse(|parser| parser.parse_data_type())? {
    match n.0[0].clone() {
      ObjectNamePart::Identifier(ident) => name = Some(ident),
    }
    data_type = dt;
  }
}
```
thinking if so it would closer match the desired goal to parse an optional datatype if the first token is regular identifier

LucaCappelletti94: I can try it out, I wasn't aware of `maybe_parse` which certainly seems to make it less confusing to read.

LucaCappelletti94: The code you proposed does not work as `Int2` (or any analogous such type) does not fall in `if let DataType::Custom(n, _) = &data_type {`, but other variants. I am trying to update my own version using `maybe_parse` instead of the keywords check.

src/parser/mod.rs
```diff
@@ -5199,13 +5199,20 @@ impl<'a> Parser<'a> {
 
         // parse: [ argname ] argtype
         let mut name = None;
+        let next_token = self.peek_token();
```
iffyio, line 5202: > The code you proposed does not work as Int2 (or any analogous such type) does not fall in if let DataType::Custom(n, _) = &data_type {

Oh how did you mean here by `Int2` in this example not being parsed as a custom datatype, do we get back a different type or does `parse_data_type` fail in that scenario?

I think ideally we will want to do without this `self.peek_token()` to avoid the cloning that it includes

LucaCappelletti94: The argument named `Int2` (as described in the [issue](https://github.com/apache/datafusion-sqlparser-rs/issues/1825)) is not parsed as `DataType::Custom`, but as a [`DataType::Int2`](https://github.com/apache/datafusion-sqlparser-rs/blob/a464f8e8d7a5057e4e5b8046b0f619acdf7fce74/src/ast/data_type.rs#L150). Analogously, any other such argument names that collides with data types from other SQL engines would be parsed into a type.

Now, if I were to convert back to string `DataType::Int2` I would get some arbitrary capitalization which in this case is `INT2` - without the `peek_token`, I am unsure how we can preserve the initial token from being lost.

iffyio: Ah I see that makes sense! Maybe something like this we can do to restrict the cloning to only when necessary?
```rust
let data_type_idx = self.get_current_index();
if let Some(next_data_type) = self.maybe_parse(|parser| {
    name = parser.token_at(data_type_idx).to_string();
   // ...
})
```
Coming to think about it, would we not need to sanity check that the first token is actually a `Token::Word` variant? current code seems to assume that to be the case which might not necessarily be true.
For example following how the following sql would be parsed, we can probably have a test case it
```rust
function(struct<a,b> int64)
```
we would call `to_string()` on only the first token which would be `struct` even though this query is technically invalid?

LucaCappelletti94: Could you provide a complete example of such a broken case, so that I may add it to the test suite?

iffyio: I think something like this potentially
```rust
let sql = r#"CREATE OR REPLACE FUNCTION foo(a TIMESTAMP WITH TIMEZONE, b VARCHAR) RETURNS BOOLEAN LANGUAGE plpgsql AS $$
BEGIN
	RETURN TRUE;
END;
$$"#;
pg().verified_stmt(sql);
```

src/parser/mod.rs
```diff
@@ -5200,12 +5200,26 @@ impl<'a> Parser<'a> {
...
-        if let DataType::Custom(n, _) = &data_type {
-            // the first token is actually a name
-            match n.0[0].clone() {
-                ObjectNamePart::Identifier(ident) => name = Some(ident),
+
+        // It may appear that the first token can be converted into a known
+        // type, but this could also be a collision as some types are only
+        // present in some dialects and therefore some type reserved keywords
+        // may be freely used as argument names in other dialects.
+
```
iffyio, line 5208:
```suggestion
```
I think we could remove this part to keep the comment smaller/focused. This part seems to be describing the current bug, which wouldn't be obvious when reading going forward since the bug would have been fixed. (smaller text also makes it more likely to be read in general)

tests/sqlparser_postgres.rs
```diff
@@ -4098,6 +4099,219 @@ fn parse_update_in_with_subquery() {
...
+            options: None,
+            remote_connection: None,
+        })
+    );
+}
+
+#[test]
+fn parser_create_function_with_invalid_args() {
+    let sql = "CREATE FUNCTION add(function(struct<a,b> int64), b INTEGER) RETURNS INTEGER LANGUAGE SQL IMMUTABLE STRICT PARALLEL SAFE AS 'select $1 + $2;'";
+    assert!(pg().parse_sql_statements(sql).is_err(),);
```
iffyio, line 4312: can we move this into the existing test function instead of having it as its own test function?

LucaCappelletti94: Done

iffyio (approved):
cc @alamb

Issue #1827, opened 2025-04-27
`DROP DOMAIN` is not supported


LucaCappelletti94:
The [`DROP DOMAIN`](https://www.postgresql.org/docs/current/sql-dropdomain.html) syntax is not currently supported.

Statements such as the following, therefore, cannot be parsed at this time.

```sql
DROP DOMAIN IF EXISTS jpeg_in CASCADE;
```

I will try to add support for this in a PR soon.

Best,
Luca

PR #1828, opened 2025-04-27
Added support for `DROP DOMAIN`

LucaCappelletti94:
This pull request:
* Adds support for [`DROP DOMAIN`](https://www.postgresql.org/docs/current/sql-dropdomain.html) syntax resolving issue #1827 
* Adds tests for `DROP DOMAIN`

Best,
Luca

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3325,11 +3325,7 @@ pub enum Statement {
     ///
     /// DROP DOMAIN [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]
     ///
-    DropDomain {
-        if_exists: bool,
-        name: ObjectName,
-        drop_behavior: Option<DropBehavior>,
-    },
+    DropDomain(DropDomain),
     /// ```sql
     /// DROP PROCEDURE
     /// ```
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -3319,6 +3319,18 @@ pub enum Statement {
...
+    /// ```
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-dropdomain.html)
+    ///
+    /// DROP DOMAIN [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]
+    ///
+    DropDomain {
+        if_exists: bool,
+        name: ObjectName,
+        drop_behavior: Option<DropBehavior>,
+    },
```
iffyio, line 3332: Could we wrap the new statement in a named struct? We're looking to move over to [that pattern](https://github.com/apache/datafusion-sqlparser-rs/issues/1204) for newer statements. e.g.
```rust
struct DropDomain {
    if_exist: bool 
    // ...
}
Statement::DropDomain(DropDomain)
```

LucaCappelletti94: Oki

LucaCappelletti94: Done in commit [10dab51](https://github.com/apache/datafusion-sqlparser-rs/pull/1828/commits/10dab5142da31e0fd0768318cc57b389ed6b1de2)

iffyio (approved):
cc @alamb

Issue #1829, opened 2025-04-27
`CREATE DOMAIN` is not supported


LucaCappelletti94:
[`CREATE DOMAIN`](https://www.postgresql.org/docs/current/sql-createdomain.html) is not supported.

Statements such as the following fail parsing at this time:

```sql
CREATE DOMAIN my_domain AS INTEGER CHECK (VALUE > 0)
```

I will try to add a PR adding support for this syntax.

Best,
Luca

LucaCappelletti94:
Added support for `CREATE DOMAIN` in PR #1830

PR #1830, opened 2025-04-27
Added support for `CREATE DOMAIN`

LucaCappelletti94:
This pull request adds support for the [`CREATE DOMAIN`](https://www.postgresql.org/docs/current/sql-createdomain.html) syntax and the tests to validate whether the implementation is correct. It closes issue #1829.

IMPORTANT: this pull request may collide with #1828, as I had to add the keyword `DOMAIN` in both of them.

Best,
Luca

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3979,13 +3979,7 @@ pub enum Statement {
     ///     { NOT NULL | NULL | CHECK (expression) }
     /// ```
     /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-createdomain.html)
-    CreateDomain {
-        name: ObjectName,
-        data_type: DataType,
-        collation: Option<Ident>,
-        default: Option<Expr>,
-        constraints: Vec<TableConstraint>,
-    },
+    CreateDomain(CreateDomain),
     /// ```sql
     /// CREATE TYPE <name>
     /// ```
@@ -4532,25 +4526,7 @@ impl fmt::Display for Statement {
                 Ok(())
             }
             Statement::CreateFunction(create_function) => create_function.fmt(f),
-            Statement::CreateDomain {
-                name,
-                data_type,
-                collation,
-                default,
-                constraints,
-            } => {
-                write!(f, "CREATE DOMAIN {name} AS {data_type}")?;
-                if let Some(collation) = collation {
-                    write!(f, " COLLATE {collation}")?;
-                }
-                if let Some(default) = default {
-                    write!(f, " DEFAULT {default}")?;
-                }
-                if !constraints.is_empty() {
-                    write!(f, " {}", display_separated(constraints, " "))?;
-                }
-                Ok(())
-            }
+            Statement::CreateDomain(create_domain) => create_domain.fmt(f),
             Statement::CreateTrigger {
                 or_replace,
                 is_constraint,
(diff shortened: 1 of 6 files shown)


LucaCappelletti94:
Refactored analogously to what was requested in pull request #1828

src/ast/mod.rs
```diff
@@ -3968,6 +3968,19 @@ pub enum Statement {
...
+    /// CREATE DOMAIN name [ AS ] data_type
+    ///         [ COLLATE collation ]
+    ///         [ DEFAULT expression ]
+    ///         [ domain_constraint [ ... ] ]
+    ///
+    ///     where domain_constraint is:
+    ///
+    ///     [ CONSTRAINT constraint_name ]
+    ///     { NOT NULL | NULL | CHECK (expression) }
+    /// ```
```
iffyio, line 3980: Instead of duplicating this doc comment, maybe it would suffice to only mention that this is a `CREATE DOMAIN` statement, [for example](https://github.com/apache/datafusion-sqlparser-rs/blob/d4868ba1316419ef3dc79942eaf1eaf662be359c/src/ast/mod.rs#L2980)

src/parser/mod.rs
```diff
@@ -5894,6 +5896,47 @@ impl<'a> Parser<'a> {
...
+    ///             [ DEFAULT expression ]
+    ///             [ domain_constraint [ ... ] ]
+    ///     
+    ///         where domain_constraint is:
+    ///     
+    ///         [ CONSTRAINT constraint_name ]
+    ///         { NOT NULL | NULL | CHECK (expression) }
+    /// ```
+    ///
+    /// [PostgreSQL Documentation](https://www.postgresql.org/docs/current/sql-createdomain.html)
```
iffyio, line 5911: same here regarding duplicating the doc comment, we can only link to `[Statement::CreateDomain]` which readers can follow for more context around the statement if needed.

src/parser/mod.rs
```diff
@@ -5894,6 +5896,47 @@ impl<'a> Parser<'a> {
...
+    pub fn parse_create_domain(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 5912:
```suggestion
    fn parse_create_domain(&mut self) -> Result<Statement, ParserError> {
```

tests/sqlparser_postgres.rs
```diff
@@ -5080,6 +5080,111 @@ fn test_escaped_string_literal() {
...
+                right: Box::new(Expr::Value(test_utils::number("0").into())),
+            }),
+        }],
+    });
+
+    assert_eq!(pg().verified_stmt(sql4), expected);
+}
+
+#[test]
+fn parse_create_domain_with_named_constraint() {
```
iffyio, line 5168: Can we merge the test cases into a single test covering this feature, vs having one test per test case? Also we can consider relying only on `verified_stmt` for most of the test cases since its usually verbose and harder to maintain tests that assert the full AST

src/ast/mod.rs
LucaCappelletti94: Updated

src/parser/mod.rs
LucaCappelletti94: Done

tests/sqlparser_postgres.rs
LucaCappelletti94: Done

iffyio (approved):
cc @alamb

alamb:
Werd!

PR #1831, opened 2025-04-28
Improve support for cursors for SQL Server

aharpervc:
This PR is a followup to https://github.com/apache/datafusion-sqlparser-rs/pull/1821 to address several difficulties I had parsing real world SQL files.

There are 3 related enhancements here:

1. Introduce support for parsing `OPEN` statements, eg: `OPEN my_cursor`
2. Expand existing `FETCH` statement parsing support the `FROM` keyword. Eg, `FETCH NEXT FROM my_cursor`. The logic formerly only supported "FETCH NEXT IN" syntax
3. Introduce support for parsing `WHILE` statements, which is commonly used in conjunction with cursors. Eg `WHILE @@fetch_status = 0...`. This is a conditional statement block, much like IF & CASE, so that code has been structured similarly and placed adjacent to those statements.

(4th thing -- a test helper introduced in this PR was brought over here, because it simplifies validating a test case. If the other PR merges first that commit can be dropped out of this branch).

The effect of these changes is that this cursor example from the [SQL Server FETCH documentation page](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/fetch-transact-sql?view=sql-server-ver16#a-using-fetch-in-a-simple-cursor) now successfully parses:

```mssql
DECLARE Employee_Cursor CURSOR FOR
SELECT LastName, FirstName
FROM AdventureWorks2022.HumanResources.vEmployee
WHERE LastName LIKE 'B%';

OPEN Employee_Cursor;
FETCH NEXT FROM Employee_Cursor;

WHILE @@FETCH_STATUS = 0
BEGIN
    FETCH NEXT FROM Employee_Cursor;
END;

CLOSE Employee_Cursor;
DEALLOCATE Employee_Cursor
```

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3447,10 +3447,7 @@ pub enum Statement {
         /// Cursor name
         name: Ident,
         direction: FetchDirection,
-        /// Differentiate between dialects that fetch `FROM` vs fetch `IN`
-        ///
-        /// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/fetch-transact-sql)
-        from_or_in: AttachedToken,
+        position: FetchPosition,
         /// Optional, It's possible to fetch rows form cursor to the table
         into: Option<ObjectName>,
     },
@@ -4273,25 +4270,10 @@ impl fmt::Display for Statement {
             Statement::Fetch {
                 name,
                 direction,
-                from_or_in,
+                position,
                 into,
             } => {
-                write!(f, "FETCH {direction} ")?;
-
-                match &from_or_in.0.token {
-                    Token::Word(w) => match w.keyword {
-                        Keyword::FROM => {
-                            write!(f, "FROM {name}")?;
-                        }
-                        Keyword::IN => {
-                            write!(f, "IN {name}")?;
-                        }
-                        _ => unreachable!(),
-                    },
-                    _ => {
-                        unreachable!()
-                    }
-                }
+                write!(f, "FETCH {direction} {position} {name}")?;
 
                 if let Some(into) = into {
                     write!(f, " INTO {into}")?;
@@ -6226,6 +6208,28 @@ impl fmt::Display for FetchDirection {
                 }
             }
             FetchDirection::BackwardAll => f.write_str("BACKWARD ALL")?,
+        };
+
+        Ok(())
+    }
+}
+
+/// The "position" for a FETCH statement.
+///
+/// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/fetch-transact-sql)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum FetchPosition {
+    From,
+    In,
+}
+
+impl fmt::Display for FetchPosition {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            FetchPosition::From => f.write_str("FROM")?,
+            FetchPosition::In => f.write_str("IN")?,
         };
 
         Ok(())
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2226,7 +2226,33 @@ impl fmt::Display for IfStatement {
     }
 }
 
-/// A block within a [Statement::Case] or [Statement::If]-like statement
+/// A `WHILE` statement.
+///
+/// Example:
+/// ```sql
+/// WHILE @@FETCH_STATUS = 0
+/// BEGIN
+///    FETCH NEXT FROM c1 INTO @var1, @var2;
+/// END
+/// ```
+///
+/// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/while-transact-sql)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct WhileStatement {
+    pub while_block: ConditionalStatementBlock,
+}
+
+impl fmt::Display for WhileStatement {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let WhileStatement { while_block } = self;
+        write!(f, "{while_block}")?;
+        Ok(())
+    }
+}
+
+/// A block within a [Statement::Case] or [Statement::If] or [Statement::While]-like statement
 ///
 /// Example 1:
 /// ```sql
@@ -2241,6 +2267,14 @@ impl fmt::Display for IfStatement {
 /// Example 3:
 /// ```sql
 /// ELSE SELECT 1; SELECT 2;
+/// ```
+///
+/// Example 4:
+/// ```sql
+/// WHILE @@FETCH_STATUS = 0
+/// BEGIN
+///    FETCH NEXT FROM c1 INTO @var1, @var2;
+/// END
 /// ```
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -2981,6 +3015,8 @@ pub enum Statement {
     Case(CaseStatement),
     /// An `IF` statement.
     If(IfStatement),
+    /// A `WHILE` statement.
+    While(WhileStatement),
     /// A `RAISE` statement.
     Raise(RaiseStatement),
     /// ```sql
@@ -4343,6 +4379,9 @@ impl fmt::Display for Statement {
                 write!(f, "{stmt}")
             }
             Statement::If(stmt) => {
+                write!(f, "{stmt}")
+            }
+            Statement::While(stmt) => {
                 write!(f, "{stmt}")
             }
             Statement::Raise(stmt) => {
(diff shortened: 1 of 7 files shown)


src/ast/mod.rs
```diff
@@ -2226,7 +2226,33 @@ impl fmt::Display for IfStatement {
...
+///    FETCH NEXT FROM c1 INTO @var1, @var2;
+/// END
+/// ```
+///
+/// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/while-transact-sql)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct WhileStatement {
+    pub while_block: ConditionalStatementBlock,
```
aharpervc, line 2244: We don't absolutely need a WhileStatement struct; we could be doing `Statement::While(ConditionalStatementBlock)` instead. I'm following the example of CASE & IF, which also do it this way.

src/ast/mod.rs
```diff
@@ -3032,6 +3068,14 @@ pub enum Statement {
         partition: Option<Box<Expr>>,
     },
     /// ```sql
+    /// OPEN cursor_name
+    /// ```
+    /// Opens a cursor.
+    Open {
```
aharpervc, line 3074: I placed this next to `CLOSE` because they're semantically paired, rather than alphabetical. Not sure what's preferred on this project

src/ast/mod.rs
```diff
@@ -3403,6 +3447,10 @@ pub enum Statement {
         /// Cursor name
         name: Ident,
         direction: FetchDirection,
+        /// Differentiate between dialects that fetch `FROM` vs fetch `IN`
+        ///
+        /// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/fetch-transact-sql)
+        from_or_in: AttachedToken,
```
aharpervc, line 3453: Not sure what's best here, it could also be two separate Optional<AttachedToken> fields

iffyio: Could we represent it with an explicit enum?
e.g
```rust
enum FetchPosition {
    From
    In
}
```

src/keywords.rs
```diff
@@ -1064,6 +1065,8 @@ pub const RESERVED_FOR_TABLE_ALIAS: &[Keyword] = &[
     Keyword::SAMPLE,
     Keyword::TABLESAMPLE,
     Keyword::FROM,
+    // Reserved for SQL Server cursors
```
iffyio, line 1068:
```suggestion
```
I think we can skip the comment, in case the keyword is used in other contexts in the future

tests/sqlparser_mssql.rs
```diff
@@ -1393,6 +1393,52 @@ fn parse_mssql_declare() {
...
+        \
+        CLOSE Employee_Cursor; \
+        DEALLOCATE Employee_Cursor\
+    ";
+    let _ = ms().statements_parse_to(full_cursor_usage, 6, "");
+}
+
+#[test]
+fn test_mssql_while_statement() {
+    let while_single_statement = "WHILE 1 = 0 PRINT 'Hello World';";
```
iffyio, line 1421: Can we include a test case with multiple statements in the while block?

Also since this introduces a new statement, can we include a test case that asserts the returned AST?

src/test_utils.rs
```diff
@@ -166,6 +168,30 @@ impl TestedDialects {
...
 
+    /// The same as [`one_statement_parses_to`] but it works for a multiple statements
+    pub fn statements_parse_to(
+        &self,
+        sql: &str,
+        statement_count: usize,
+        canonical: &str,
+    ) -> Vec<Statement> {
+        let statements = self.parse_sql_statements(sql).expect(sql);
+        assert_eq!(statements.len(), statement_count);
```
iffyio, line 179: this assertion seems to already be covered by the if/else below? so that we can skip the statement_count argument requirement?

src/parser/mod.rs
```diff
@@ -8735,6 +8779,14 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse [Statement::Open]
```
iffyio, line 8782: maybe I missed it, we seem to be lacking test cases for the open statement feature?

src/parser/mod.rs
```diff
@@ -4453,6 +4487,9 @@ impl<'a> Parser<'a> {
                     break;
                 }
             }
+            if let Token::EOF = self.peek_nth_token_ref(0).token {
+                break;
+            }
```
iffyio, line 4492: can we collapse this into above to use a match statement?
```rust
match &self.peek_nth_token_ref(0).token {
    Token::Word(n) if ...
    Token::Eof 
}
```

src/ast/mod.rs

tests/sqlparser_mssql.rs
aharpervc: > Can we include a test case with multiple statements in the while block?

This is covered by the additional subsequent examples

> Also since this introduces a new statement, can we include a test case that asserts the returned AST?

Yes, I'll do that here for the initial example case

aharpervc: Done 👍

src/test_utils.rs
aharpervc: Hm, I don't fully understand. Without this line you can't guarantee that the string you feed in has exactly the number of statements you intend it to parse. Also, `one_statement_parses_to` has this same assertion before the if/else.

src/parser/mod.rs
aharpervc: It's part of `test_mssql_cursor`, but I'll make a separate test function just for `OPEN` for clarity

aharpervc: Done 👍

src/parser/mod.rs
aharpervc: Done 👍. I was probably trying to minimize the diff for review here

src/ast/mod.rs
```diff
@@ -4225,11 +4267,10 @@ impl fmt::Display for Statement {
             Statement::Fetch {
                 name,
                 direction,
+                position,
                 into,
             } => {
-                write!(f, "FETCH {direction} ")?;
-
-                write!(f, "IN {name}")?;
+                write!(f, "FETCH {direction} {position} {name}")?;
```
aharpervc, line 4273: This could probably be ` write!(f, "{position} {name}")?;`, not sure what the pro/con is on that

tests/sqlparser_mssql.rs
```diff
@@ -1393,6 +1394,85 @@ fn parse_mssql_declare() {
...
+                    )),
+                }),
+                then_token: None,
+                conditional_statements: ConditionalStatements::Sequence {
+                    statements: vec![Statement::Print(PrintStatement {
+                        message: Box::new(Expr::Value(
+                            (Value::SingleQuotedString("Hello World".to_string()))
+                                .with_empty_span()
+                        )),
+                    }),],
```
aharpervc, line 1452: Oops... I will fix this

src/test_utils.rs
iffyio: Ah so I meant that in both cases when asserting :

in this case we're already explicitly check that both statements lists are identical
```rust
assert_eq!(self.parse_sql_statements(canonical).unwrap(), statements);
```

Then in this case, we're doing so implicitly, reconstructing the input sql based off the returned statement
```rust
assert_eq!(
                sql,
                statements
                    .iter()
                    .map(|s| s.to_string())
                    .collect::<Vec<_>>()
                    .join("; ")
            );
```

So that i imagine it shouldn't be possible for the count assertion to fail and either of the subsequent assertion to pass?

`one_statement_parse_to` uses the count internally and there it makes sense to sanity check that since the expected count is always one, in this case we're exposing the count as a function argument which makes for suboptimal API that the user has to manually increment/decrement a counter when the sql changes. So that if the count argument isn't strictly necessary we would want to skip it

aharpervc: > Then in this case, we're doing so implicitly, reconstructing the input sql based off the returned statement ... So that i imagine it shouldn't be possible for the count assertion to fail and either of the subsequent assertion to pass?

I will remove the assertion here to get this branch merged. However, I think removing it removes a level of safety that is beneficial. Part of my thinking here is motivated by my upcoming branch on making semi colon statement delimiters optional. So any code that is making assumptions about "number of statements" becomes even more useful.

But perhaps that branch can re-introduce that assertion if necessary.

aharpervc: BTW, this helper was introduced over on the [GO branch](https://github.com/apache/datafusion-sqlparser-rs/pull/1809), does your opinion change at all seeing the usage over there?

src/parser/mod.rs
```diff
@@ -716,20 +734,36 @@ impl<'a> Parser<'a> {
...
                 then_token = Some(AttachedToken(self.expect_keyword(Keyword::THEN)?));
                 Some(expr)
             }
         };
 
-        let statements = self.parse_statement_list(terminal_keywords)?;
+        let conditional_statements = if self.peek_keyword(Keyword::BEGIN) {
+            let begin_token = self.expect_keyword(Keyword::BEGIN)?;
+            let statements = self.parse_statement_list(terminal_keywords)?;
+            let end_token = self.expect_keyword(Keyword::END)?;
```
iffyio, line 751: We seem to have this pattern upcoming in a few places, like #1810 maybe it would be good to pull it out into a method and reuse it both here and the preexisting usage [here](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/parser/mod.rs#L5157-L5159)? We can probably do so in the former PR instead

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1834, opened 2025-04-30
Allow stored procedures to be defined without `BEGIN`/`END`

aharpervc:
For SQL Server, you can make a stored procedure without begin/end ([docs ref](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-procedure-transact-sql?view=sql-server-ver16)). Otherwise, it parses the same way.

To differentiate with/without in the parser, the stored procedure struct's `statements` field was changed from `Vec<Statement>` (where begin/end are required & implicit) to a `BeginEndStatements`, where the begin/end tokens are explicit. They're empty when missing & written when present.

This PR also includes the fix to allow EOF to end a statement list from https://github.com/apache/datafusion-sqlparser-rs/pull/1831 (so whichever merges first, I'll rebase accordingly)

The diff is perhaps larger than expected due to the question of canonical semicolons for procedure statement bodies. Formerly, a semicolon after the last statement in a procedure was non-canonical (because they were added via `join`... so perhaps not particular intentional for it to have been that way); a `BeginEndStatements` statements list will always write them out. 

An additional test case example without begin/end has been added as well.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3826,7 +3826,7 @@ pub enum Statement {
         or_alter: bool,
         name: ObjectName,
         params: Option<Vec<ProcedureParam>>,
-        body: Vec<Statement>,
+        body: BeginEndStatements,
     },
     /// ```sql
     /// CREATE MACRO
@@ -4705,11 +4705,8 @@ impl fmt::Display for Statement {
                         write!(f, " ({})", display_comma_separated(p))?;
                     }
                 }
-                write!(
-                    f,
-                    " AS BEGIN {body} END",
-                    body = display_separated(body, "; ")
-                )
+
+                write!(f, " AS {body}")
             }
             Statement::CreateMacro {
                 or_replace,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3826,7 +3826,7 @@ pub enum Statement {
         or_alter: bool,
         name: ObjectName,
         params: Option<Vec<ProcedureParam>>,
-        body: BeginEndStatements,
+        body: ConditionalStatements,
     },
     /// ```sql
     /// CREATE MACRO
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15506,27 +15506,13 @@ impl<'a> Parser<'a> {
         let params = self.parse_optional_procedure_parameters()?;
         self.expect_keyword_is(Keyword::AS)?;
 
-        let begin_token: AttachedToken = self
-            .expect_keyword(Keyword::BEGIN)
-            .map(AttachedToken)
-            .unwrap_or_else(|_| AttachedToken::empty());
-        let statements = self.parse_statement_list(&[Keyword::END])?;
-        let end_token = match &begin_token.0.token {
-            Token::Word(w) if w.keyword == Keyword::BEGIN => {
-                AttachedToken(self.expect_keyword(Keyword::END)?)
-            }
-            _ => AttachedToken::empty(),
-        };
+        let body = self.parse_conditional_statements(&[Keyword::END])?;
 
         Ok(Statement::CreateProcedure {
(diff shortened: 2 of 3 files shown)


src/parser/mod.rs
```diff
@@ -15095,14 +15099,28 @@ impl<'a> Parser<'a> {
...
+            .expect_keyword(Keyword::BEGIN)
+            .map(AttachedToken)
+            .unwrap_or_else(|_| AttachedToken::empty());
+        let statements = self.parse_statement_list(&[Keyword::END])?;
+        let end_token = match &begin_token.0.token {
+            Token::Word(w) if w.keyword == Keyword::BEGIN => {
+                AttachedToken(self.expect_keyword(Keyword::END)?)
+            }
+            _ => AttachedToken::empty(),
+        };
```
aharpervc, line 15113: Could also do it like this: https://github.com/apache/datafusion-sqlparser-rs/pull/1810/files#diff-6f6a082c3ddfc1f16cc4a455e3e2e2d2508f77b682ab18cabee69471bbb3edb3R244-R260, not sure which is best

src/ast/mod.rs
```diff
@@ -3744,7 +3750,7 @@ pub enum Statement {
         or_alter: bool,
         name: ObjectName,
         params: Option<Vec<ProcedureParam>>,
-        body: Vec<Statement>,
+        body: BeginEndStatements,
```
aharpervc, line 3753: It it an abuse of BeginEndStatements to potentially have the begin/end token be empty? This could alternately be an enum of BeginEndStatements and Sequence, similar to elsewhere

aharpervc:
I have temporarily rebased this branch on https://github.com/apache/datafusion-sqlparser-rs/pull/1810 to pick up the new helper function and similarly use the enum pattern to distinguish between Sequence & BeginEndStatements.

iffyio:
@aharpervc could you take a look at this PR, it seems to have commits from other PRs like the CREATE TRIGGER in it, so hard to tell what's being introduced this one

aharpervc:
> @aharpervc could you take a look at this PR, it seems to have commits from other PRs like the CREATE TRIGGER in it, so hard to tell what's being introduced this one

Yes, there is similarity between the branches, which is why I had rebased this on the other. Now that it's merged I rebased on main so this branch only includes the distinct commits.

src/ast/mod.rs
aharpervc: @iffyio based on the recent discussion on the other PR, guessing you will prefer this also to be an enum, to avoid the empty begin/end tokens?

tests/sqlparser_mssql.rs
```diff
@@ -100,48 +100,52 @@ fn parse_mssql_delimited_identifiers() {
 
 #[test]
 fn parse_create_procedure() {
-    let sql = "CREATE OR ALTER PROCEDURE test (@foo INT, @bar VARCHAR(256)) AS BEGIN SELECT 1 END";
+    let sql = "CREATE OR ALTER PROCEDURE test (@foo INT, @bar VARCHAR(256)) AS BEGIN SELECT 1; END";
```
aharpervc, line 103: I think the semicolon is coming from `format_statement_list` here: https://github.com/apache/datafusion-sqlparser-rs/blob/ac1c339666c68779ed7f20dc0fb3b7473b298f83/src/ast/mod.rs#L163-L165

Not sure if we need to change any of that in this PR. It seems odd that semicolons _weren't_ part of the canonical SQL previously 🤔

iffyio: I think this behavior is OK, we try to preserve roundtrip behavior where possible but I'm suspecting that in this case, the effort required to work around it would not be worth the return

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1835, opened 2025-04-30
Fix: parsing ident starting with underscore in certain dialects

MohamedAbdeen21:
The dialects that support underscore as a separator in numeric literals used to parse `._123` as a number, which I don't think is valid SQL. However, that means that something like `._abc` would be parsed as Number `._` and word `abc`, which is wrong.

This PR splits the tokenizer match branch for numbers and periods into two branches to make things easier, fixes the issue mentioned above, and adds tests.

CC: @mvzink

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1287,9 +1287,16 @@ impl<'a> Tokenizer<'a> {
                     chars.next(); // consume the dot
 
                     match chars.peek() {
-                        Some('_') => {
-                            // Handle "._" case as a period (special token) followed by identifier
+                        // Handle "._" case as a period followed by identifier
+                        // if the last token was a word
+                        Some('_') if matches!(prev_token, Some(Token::Word(_))) => {
                             Ok(Some(Token::Period))
+                        }
+                        Some('_') => {
+                            self.tokenizer_error(
+                                chars.location(),
+                                "Unexpected an underscore here".to_string(),
+                            )
                         }
                         Some(ch)
                             //  Hive and mysql dialects allow numeric prefixes for identifers
@@ -2504,6 +2511,16 @@ mod tests {
         ];
 
         compare(expected, tokens);
+
+        let sql = String::from("SELECT ._123");
+        if let Ok(tokens) = Tokenizer::new(&dialect, &sql).tokenize() {
+            panic!("Tokenizer should have failed on {sql}, but it succeeded with {tokens:?}");
+        }
+
+        let sql = String::from("SELECT ._abc");
+        if let Ok(tokens) = Tokenizer::new(&dialect, &sql).tokenize() {
+            panic!("Tokenizer should have failed on {sql}, but it succeeded with {tokens:?}");
+        }
     }
 
     #[test]
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1295,7 +1295,7 @@ impl<'a> Tokenizer<'a> {
                         Some('_') => {
                             self.tokenizer_error(
                                 chars.location(),
-                                "Unexpected an underscore here".to_string(),
+                                "Unexpected underscore here".to_string(),
                             )
                         }
                         Some(ch)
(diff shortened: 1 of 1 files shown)


src/tokenizer.rs
```diff
@@ -1281,20 +1262,91 @@ impl<'a> Tokenizer<'a> {
...
+                        // if the last token was a word
+                        Some('_') if matches!(prev_token, Some(Token::Word(_))) => {
+                            Ok(Some(Token::Period))
+                        }
+                        Some('_') => {
+                            self.tokenizer_error(
+                                chars.location(),
+                                "Unexpected underscore here".to_string(),
+                            )
+                        }
```
iffyio, line 1297: I wonder if its worth returning an error here or whether we lose anything by allowing the tokenizer continue? I'm guessing its still possible for the tokenizer to return `Token::Period` here as well?

src/tokenizer.rs
```diff
@@ -1281,20 +1262,91 @@ impl<'a> Tokenizer<'a> {
...
+                                if matches!(chars.peek(), Some(c) if c.is_ascii_digit()) {
+                                    exp += &peeking_take_while(chars, |c| c.is_ascii_digit());
+                                    s += &exp;
+                                }
+                            }
+
+                            // Handle "L" suffix for long numbers
+                            let long = if chars.peek() == Some(&'L') {
+                                chars.next();
+                                true
```
iffyio, line 1337: hmm most of this logic looks to already be duplicated on the [number parsing](https://github.com/apache/datafusion-sqlparser-rs/blob/f05913dd698a1fa2eeebf4f6a286f3cf19ddaff7/src/tokenizer.rs#L1224-L1276) code path, so that that side effect would be undesirable I think.


If I understood the issue being solved for, its only the case of `._` being parsed as a number, would it be possible/more-desirable to only update the existing logic to properly detect and handle that case or is the current logic not well equipped to handle that sanely?

MohamedAbdeen21: I know. I tried to fix it without splitting the branches, but I couldn't. 

Feel free to take a go at it if possible

MohamedAbdeen21: The problem is the match happens on a peek and you need to consume the dot in order to peek the underscore.

What if the second peek wasn't an underscore? You need to un-consume the dot for it to be parsed as part of the number.

src/tokenizer.rs
MohamedAbdeen21: AFAIK underscore after a dot can only be an identifier, so it makes sense to error here

src/tokenizer.rs
iffyio: Does the [peekable_clone](https://github.com/apache/datafusion-sqlparser-rs/blob/f05913dd698a1fa2eeebf4f6a286f3cf19ddaff7/src/tokenizer.rs#L1060-L1061) help in this case maybe? in order to lookahead without consuming characters

MohamedAbdeen21: Of course, but I prefer not cloning the entire token stream, especially since the dup code is only a few lines.

MohamedAbdeen21: Would you prefer pulling the dup code into functions, if possible?

iffyio: Oh is that really the case that cloning the peekable iterator would clone the entire stream? (iirc it should only clone a struct with pointer offsets), the functionality is used in a few places in the tokenizer so I think it would be preferrable to avoid dup logic (if it turns out its an expensive thing to do then maybe we need to do a pass through the tokenizer to avoid that pattern overall, but that would be unrelated to this PR ofc)

src/tokenizer.rs
MohamedAbdeen21: without the errors, '._123' as a whole will be parsed as a number, which is not valid in any standard AFAICT.
'._abc' as a whole would also be parsed as a word, which doesn't make sense because the prev token was not a word.

src/tokenizer.rs
MohamedAbdeen21: updated

iffyio (approved):
cc @alamb

PR #1836, opened 2025-05-01
Add support for `DENY` statements

aharpervc:
This is another statement for SQL Server: https://learn.microsoft.com/en-us/sql/t-sql/statements/deny-transact-sql, but implemented in a common way. Similar to GRANT & REVOKE, so using a lot of those patterns with a similar test.

Supplementary improvements:

1. While I was working on that, I also noticed that `EXEC` was missing from the privileges list, so that's added now too (with a test example).
2. Additionally, SQL Server supports identifiers quoted `[like this]`. That dialect function hadn't been implemented yet, which caused a test failure on example that use that quoting style. So I've added that now too with a couple example cases for grant/deny.
3. ~~Finally,~~ "public" is a built in role on SQL Server, so doing something like `grant select on mytable to public` should parse public normally. However, the former logic was reading it as a keyword. This has been adjusted to be not special on SQL Server and tests cases have been added as well.
4. Actually finally, `GRANT ... AS role` parsing is now supported (formerly only worked for REVOKE/DENY) and a new test case example

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13419,7 +13419,7 @@ impl<'a> Parser<'a> {
         let mut values = vec![];
         let mut grantee_type = GranteesType::None;
         loop {
-            grantee_type = if self.parse_keyword(Keyword::ROLE) {
+            let new_grantee_type = if self.parse_keyword(Keyword::ROLE) {
                 GranteesType::Role
             } else if self.parse_keyword(Keyword::USER) {
                 GranteesType::User
@@ -13427,21 +13427,27 @@ impl<'a> Parser<'a> {
                 GranteesType::Share
             } else if self.parse_keyword(Keyword::GROUP) {
                 GranteesType::Group
+            } else if self.parse_keyword(Keyword::PUBLIC) {
+                GranteesType::Public
             } else if self.parse_keywords(&[Keyword::DATABASE, Keyword::ROLE]) {
                 GranteesType::DatabaseRole
             } else if self.parse_keywords(&[Keyword::APPLICATION, Keyword::ROLE]) {
                 GranteesType::ApplicationRole
             } else if self.parse_keyword(Keyword::APPLICATION) {
                 GranteesType::Application
-            } else if self.peek_keyword(Keyword::PUBLIC) {
-                if dialect_of!(self is MsSqlDialect) {
-                    grantee_type
-                } else {
-                    GranteesType::Public
-                }
             } else {
-                grantee_type // keep from previous iteraton, if not specified
+                grantee_type.clone() // keep from previous iteraton, if not specified
             };
+
+            if self
+                .dialect
+                .get_reserved_grantees_types()
+                .contains(&new_grantee_type)
+            {
+                self.prev_token();
+            } else {
+                grantee_type = new_grantee_type;
+            }
 
             let grantee = if grantee_type == GranteesType::Public {
                 Grantee {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -583,6 +583,10 @@ impl<'a> Parser<'a> {
                 Keyword::SHOW => self.parse_show(),
                 Keyword::USE => self.parse_use(),
                 Keyword::GRANT => self.parse_grant(),
+                Keyword::DENY => {
+                    self.prev_token();
+                    self.parse_deny()
+                }
                 Keyword::REVOKE => self.parse_revoke(),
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -12987,23 +12991,34 @@ impl<'a> Parser<'a> {
...
 
         let with_grant_option =
             self.parse_keywords(&[Keyword::WITH, Keyword::GRANT, Keyword::OPTION]);
 
-        let granted_by = self
-            .parse_keywords(&[Keyword::GRANTED, Keyword::BY])
-            .then(|| self.parse_identifier().unwrap());
+        let as_grantor = if self.peek_keyword(Keyword::AS) {
+            self.parse_keywords(&[Keyword::AS])
+                .then(|| self.parse_identifier().unwrap())
```
iffyio, line 13004: Oh I realise this unwrap wasn't introduced by this PR, but can we switch it to return an error instead? it looks otherwise like an scenario where the parser will panic on invalid sql which would be undesirable

src/parser/mod.rs
```diff
@@ -12987,23 +12991,34 @@ impl<'a> Parser<'a> {
...
+        } else {
+            None
+        };
+
+        let granted_by = if self.peek_keywords(&[Keyword::GRANTED, Keyword::BY]) {
+            self.parse_keywords(&[Keyword::GRANTED, Keyword::BY])
+                .then(|| self.parse_identifier().unwrap())
```
iffyio, line 13011: similar here, we would want to return an error vs unwrap

src/parser/mod.rs
```diff
@@ -13409,9 +13430,40 @@ impl<'a> Parser<'a> {
...
+                )
+            }
+        };
+
+        self.expect_keyword_is(Keyword::TO)?;
+        let grantees = self.parse_grantees()?;
+        let cascade = self.parse_cascade_option();
+        let granted_by = self
+            .parse_keywords(&[Keyword::AS])
+            .then(|| self.parse_identifier().unwrap());
```
iffyio, line 13453: same comment here re unwrap

src/parser/mod.rs
```diff
@@ -13020,14 +13035,18 @@ impl<'a> Parser<'a> {
...
             } else if self.parse_keywords(&[Keyword::APPLICATION, Keyword::ROLE]) {
                 GranteesType::ApplicationRole
             } else if self.parse_keyword(Keyword::APPLICATION) {
                 GranteesType::Application
+            } else if self.peek_keyword(Keyword::PUBLIC) {
+                if dialect_of!(self is MsSqlDialect) {
+                    grantee_type
+                } else {
+                    GranteesType::Public
+                }
```
iffyio, line 13049: hmm not sure I followed the intention here, public is represented differently for mssql? (we would want to describe that behavior without the `dialect_of` macro I think)

aharpervc: > hmm not sure I followed the intention here, public is represented differently for mssql? (we would want to describe that behavior without the `dialect_of` macro I think)

My interpretation of the existing code is that "public" is being interpreted as a keyword. With SQL Server, "public" is a built in role ([ref](https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/database-level-roles?view=sql-server-ver16#public-database-role))

From a parsing perspective, there wouldn't be any difference between a built-in role & any other role. The consideration here is "keyword" vs "not keyword". The additional test case examples here cover this behavior.

Put simply, this (and GRANT) should be able to parse normally:

```mssql
DENY SELECT ON my_table TO public
```

---

With regards to the code, what is the preferred way to implement this behavior? Brainstorming...
1. Use `dialect_of`
2. Copy/paste this entire method into the dialect so it's fully duplicated except for this keyword
3. Implement a dialect function to describe "tokens that are listed as keywords but should be treated as identifiers", or similar, and then it's blank for most dialects but returns "public" for SQL Server
4. Something else...?

src/parser/mod.rs
aharpervc: I've refactored these three related methods to avoid unwrap 👍

src/parser/mod.rs
iffyio: Ah I see, yeah option 3. sounds reasonable and we could try that?

aharpervc:
This has also been rebased on main & merge conflict resolved

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1839, opened 2025-05-06
Add support for table valued functions for SQL Server

aharpervc:
This PR adds support for table valued functions for SQL Server, both inline & multi statement functions. For reference, that's the B & C documentation here: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#b-create-an-inline-table-valued-function

Inline TVF's are defined with `AS RETURN`, so we have a new `CreateFunctionBody::AsReturn` variant accordingly. Functions using "AS RETURN" don't have BEGIN/END, so that part of the parsing logic is now conditional. Additionally, the data type parser now supports "RETURNS TABLE" without a table definition.

Multi statement TVF's use named table expressions, so a new `NamedTable` data type variant was added. I didn't see a great way to integrate this into the existing data type parser (especially without rewinding), so creating this data type happens inside the parse create function logic first by parsing the identifier, then parsing the table definition, then using those elements to produce a NamedTable.

I also added a new test example for each of these scenarios.

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -51,12 +51,12 @@ pub enum DataType {
     /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
     Table(Option<Vec<ColumnDef>>),
     /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
-    NamedTable(
+    NamedTable {
         /// Table name.
-        ObjectName,
+        name: ObjectName,
         /// Table columns.
-        Vec<ColumnDef>,
-    ),
+        columns: Vec<ColumnDef>,
+    },
     /// Fixed-length character type, e.g. CHARACTER(10).
     Character(Option<CharacterLength>),
     /// Fixed-length char type, e.g. CHAR(10).
@@ -732,8 +732,8 @@ impl fmt::Display for DataType {
                     write!(f, "TABLE")
                 }
             },
-            DataType::NamedTable(name, fields) => {
-                write!(f, "{} TABLE ({})", name, display_comma_separated(fields))
+            DataType::NamedTable { name, columns } => {
+                write!(f, "{} TABLE ({})", name, display_comma_separated(columns))
             }
             DataType::GeometricType(kind) => write!(f, "{}", kind),
         }
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -49,7 +49,7 @@ pub enum DataType {
     ///
     /// [PostgreSQL]: https://www.postgresql.org/docs/15/sql-createfunction.html
     /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
-    Table(Vec<ColumnDef>),
+    Table(Option<Vec<ColumnDef>>),
     /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
     NamedTable(
         /// Table name.
@@ -724,12 +724,14 @@ impl fmt::Display for DataType {
             DataType::Unspecified => Ok(()),
             DataType::Trigger => write!(f, "TRIGGER"),
             DataType::AnyType => write!(f, "ANY TYPE"),
-            DataType::Table(fields) => {
-                if fields.is_empty() {
-                    return write!(f, "TABLE");
+            DataType::Table(fields) => match fields {
+                Some(fields) => {
+                    write!(f, "TABLE({})", display_comma_separated(fields))
                 }
-                write!(f, "TABLE({})", display_comma_separated(fields))
-            }
+                None => {
+                    write!(f, "TABLE")
+                }
+            },
             DataType::NamedTable(name, fields) => {
                 write!(f, "{} TABLE ({})", name, display_comma_separated(fields))
             }
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -51,6 +51,8 @@ pub enum DataType {
     /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
     Table(Option<Vec<ColumnDef>>),
     /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
+    ///
+    /// [MsSQl]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#table
     NamedTable {
         /// Table name.
         name: ObjectName,
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -48,6 +48,7 @@ pub enum DataType {
     /// Table type in [PostgreSQL], e.g. CREATE FUNCTION RETURNS TABLE(...).
     ///
     /// [PostgreSQL]: https://www.postgresql.org/docs/15/sql-createfunction.html
+    /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
     Table(Vec<ColumnDef>),
     /// Fixed-length character type, e.g. CHARACTER(10).
     Character(Option<CharacterLength>),
@@ -716,7 +717,12 @@ impl fmt::Display for DataType {
             DataType::Unspecified => Ok(()),
             DataType::Trigger => write!(f, "TRIGGER"),
             DataType::AnyType => write!(f, "ANY TYPE"),
-            DataType::Table(fields) => write!(f, "TABLE({})", display_comma_separated(fields)),
+            DataType::Table(fields) => {
+                if fields.is_empty() {
+                    return write!(f, "TABLE");
+                }
+                write!(f, "TABLE({})", display_comma_separated(fields))
+            }
             DataType::GeometricType(kind) => write!(f, "{}", kind),
         }
     }
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -50,6 +50,13 @@ pub enum DataType {
     /// [PostgreSQL]: https://www.postgresql.org/docs/15/sql-createfunction.html
     /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
     Table(Vec<ColumnDef>),
+    /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
+    NamedTable(
+        /// Table name.
+        ObjectName,
+        /// Table columns.
+        Vec<ColumnDef>,
+    ),
     /// Fixed-length character type, e.g. CHARACTER(10).
     Character(Option<CharacterLength>),
     /// Fixed-length char type, e.g. CHAR(10).
@@ -722,6 +729,9 @@ impl fmt::Display for DataType {
                     return write!(f, "TABLE");
                 }
                 write!(f, "TABLE({})", display_comma_separated(fields))
+            }
+            DataType::NamedTable(name, fields) => {
+                write!(f, "{} TABLE ({})", name, display_comma_separated(fields))
             }
             DataType::GeometricType(kind) => write!(f, "{}", kind),
         }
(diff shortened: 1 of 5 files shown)


tests/sqlparser_mssql.rs
```diff
@@ -283,6 +294,50 @@ fn parse_create_function() {
         END\
     ";
     let _ = ms().verified_stmt(create_function_with_return_expression);
+
+    let create_inline_table_value_function = "\
+        CREATE FUNCTION some_inline_tvf(@foo INT, @bar VARCHAR(256)) \
+        RETURNS TABLE \
+        AS \
+        RETURN (SELECT 1 AS col_1)\
```
aharpervc, line 302: Parentheses are optional for inline tvf `return` queries, although I think the subquery expr expects/requires them currently.

aharpervc: I added support for that syntax & added a new test case example

aharpervc: `UNION` is also supported in this syntax, but I'm comfortable leaving that for later

src/parser/mod.rs
```diff
@@ -5203,19 +5203,73 @@ impl<'a> Parser<'a> {
...
+        } else {
+            Some(self.parse_data_type()?)
+        };
 
-        let begin_token = self.expect_keyword(Keyword::BEGIN)?;
-        let statements = self.parse_statement_list(&[Keyword::END])?;
-        let end_token = self.expect_keyword(Keyword::END)?;
+        if self.peek_keyword(Keyword::AS) {
+            self.expect_keyword_is(Keyword::AS)?;
+        }
```
iffyio, line 5237:
```suggestion
        self.parse_keyword(Keyword::AS);
```
this looks like it could be simplified as above?

src/ast/data_type.rs
```diff
@@ -716,7 +724,15 @@ impl fmt::Display for DataType {
             DataType::Unspecified => Ok(()),
             DataType::Trigger => write!(f, "TRIGGER"),
             DataType::AnyType => write!(f, "ANY TYPE"),
-            DataType::Table(fields) => write!(f, "TABLE({})", display_comma_separated(fields)),
+            DataType::Table(fields) => {
+                if fields.is_empty() {
+                    return write!(f, "TABLE");
+                }
+                write!(f, "TABLE({})", display_comma_separated(fields))
```
iffyio, line 731: I think instead of using the `is_empty` we can make the `fields` value an `Option` and skip the parenthesis only if `fields` is `None`. Otherwise we could run into issues if empty fields is allowed in the other variant of this datatype

src/parser/mod.rs
```diff
@@ -5203,19 +5203,73 @@ impl<'a> Parser<'a> {
...
+                ObjectName(vec![ObjectNamePart::Identifier(return_table_name)]),
+                table_column_defs.clone(),
+            ))
+        })?;
+
+        let return_type = if return_table.is_some() {
+            return_table
+        } else {
+            Some(self.parse_data_type()?)
+        };
```
iffyio, line 5233: Could this be included in the `maybe_parse()` logic above? such that the logic returns both the return table and the datatype. based on this condition, they seem to go hand in hand?

iffyio (commented):
took a quick look and left a couple comments, @aharpervc could you rebase on main now that the other PR has landed, in order to remove the extra diff?

src/parser/mod.rs
aharpervc: Hm, I don't see how that would be done. The use of maybe_parse is to get the table name & other parts, but if we don't have that then we should rewind and just parse a typical data type. Do you mean like having another maybe_parse inside the first maybe_parse?

aharpervc:
> took a quick look and left a couple comments, @aharpervc could you rebase on main now that the other PR has landed, in order to remove the extra diff?

Done 👍

src/parser/mod.rs
iffyio: Ah indeed, in this case its one or the other, so should be fine to leave as is.

src/parser/mod.rs
```diff
@@ -5203,19 +5203,79 @@ impl<'a> Parser<'a> {
...
+        } else if self.parse_keyword(Keyword::RETURN) {
+            if self.peek_token() == Token::LParen {
+                let expr = self.parse_expr()?;
+                if !matches!(expr, Expr::Subquery(_)) {
+                    parser_err!(
+                        "Expected a subquery after RETURN",
+                        self.peek_token().span.start
+                    )?
+                }
+                Some(CreateFunctionBody::AsReturnSubquery(expr))
```
iffyio, line 5266:
```suggestion
                Some(CreateFunctionBody::AsReturnExpr(expr))
```
thinking since a subquery is already an expr, we can be more permissive here?

aharpervc: I think this is a "could" vs "should" situation. The question here is what is the parser expecting? I don't think it makes sense to parse arbitrary expr's if they wouldn't otherwise be allowed by any sql engine... that seems better as a parse failure.

src/ast/data_type.rs
```diff
@@ -48,7 +48,15 @@ pub enum DataType {
...
-    Table(Vec<ColumnDef>),
+    /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
+    Table(Option<Vec<ColumnDef>>),
+    /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
+    NamedTable(
+        /// Table name.
+        ObjectName,
+        /// Table columns.
+        Vec<ColumnDef>,
+    ),
```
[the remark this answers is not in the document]
aharpervc: Oops, yes we can. Not sure why I did it that way. Done 👍

src/ast/data_type.rs
```diff
@@ -48,7 +48,15 @@ pub enum DataType {
     /// Table type in [PostgreSQL], e.g. CREATE FUNCTION RETURNS TABLE(...).
     ///
     /// [PostgreSQL]: https://www.postgresql.org/docs/15/sql-createfunction.html
-    Table(Vec<ColumnDef>),
+    /// [MsSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver16#c-create-a-multi-statement-table-valued-function
+    Table(Option<Vec<ColumnDef>>),
+    /// Table type with a name, e.g. CREATE FUNCTION RETURNS @result TABLE(...).
```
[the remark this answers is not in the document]
aharpervc: Done 👍

src/ast/mod.rs
```diff
@@ -8660,6 +8660,28 @@ pub enum CreateFunctionBody {
...
+
+    /// Function body expression using the 'AS RETURN' keywords, with an un-parenthesized SELECT query
+    ///
+    /// Example:
+    /// ```sql
+    /// CREATE FUNCTION myfunc(a INT, b INT)
+    /// RETURNS TABLE
+    /// AS RETURN SELECT a + b AS sum;
+    /// ```
+    AsReturnSelect(Select),
```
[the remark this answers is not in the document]
aharpervc: Done 👍

src/parser/mod.rs
```diff
@@ -9766,8 +9833,12 @@ impl<'a> Parser<'a> {
                     Ok(DataType::AnyType)
                 }
                 Keyword::TABLE => {
-                    let columns = self.parse_returns_table_columns()?;
-                    Ok(DataType::Table(columns))
+                    if self.peek_token() != Token::LParen {
+                        Ok(DataType::Table(None))
```
iffyio, line 9837: can we add a comment here explaining what the LParen check is for?

src/parser/mod.rs
iffyio: Yeah ideally we would parse statements accordingy to spec but the parser is more permissive in some cases, especially in order to minimize complexity in the parser. In this case we would avoid introducing an extra `AsReturnSubquery` enum variant, as well as the `matches!()` check that does the validation after the fact of parsing an expression which isn't ideal. So I think we'd be better off [leaving the validation to the downstream crate in this scenario](https://github.com/apache/datafusion-sqlparser-rs?tab=readme-ov-file#syntax-vs-semantics)

aharpervc: Hm, yes I suppose that makes sense. It seems odd for the parser to produce output if it's known to be invalid, but I suppose you're saying, the validity question is in the _semantics_ rather than the _syntax_.

Loosening the implementation to any Expr would probably mean that `select .. union .. select` syntax would begin parsing, which addresses my concern above.

I'll investigate...

src/parser/mod.rs
```diff
@@ -5203,19 +5203,79 @@ impl<'a> Parser<'a> {
...
+            }
 
-        let function_body = Some(CreateFunctionBody::AsBeginEnd(BeginEndStatements {
-            begin_token: AttachedToken(begin_token),
-            statements,
-            end_token: AttachedToken(end_token),
-        }));
+            Ok(DataType::NamedTable(
+                ObjectName(vec![ObjectNamePart::Identifier(return_table_name)]),
+                table_column_defs.clone().unwrap(),
```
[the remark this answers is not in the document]
aharpervc: I reworked this code to address your several comments 👍

src/parser/mod.rs
aharpervc: Done, and I simplified the if/else 👍

src/parser/mod.rs
aharpervc: The primary reason to have both `AsReturnSubquery` and `AsReturnSelect` is to distinguish between `AS RETURN (SELECT a + b AS sum)` and `AS RETURN SELECT a + b AS sum`.

In other PR's, such as recent work on BeginEndStatements, I believe you had recommended that with/without wrapping tokens should be distinct enum variants instead of something like present/empty AttachedTokens. Based on that guidance, it seems like it'd also fit to have two distinct enum variants here as well.

There is your additional concern regarding "subquery" vs "expr". To address this I have renamed `AsReturnSubquery` to `AsReturnExpr` & removed the non-subquery parser error.

So I think this concern is now resolved

aharpervc:
@iffyio concerns addressed & rebased on latest master. I'm unsure how to address the linting concerns and I can't reproduce the warnings locally, so I may need some guidance on that.

aharpervc:
It seems the CI failure for the lint job was addressed by https://github.com/apache/datafusion-sqlparser-rs/pull/1856. I've rebased again, all good 👍

src/parser/mod.rs
```diff
@@ -5204,19 +5204,79 @@ impl<'a> Parser<'a> {
...
-        self.expect_keyword_is(Keyword::AS)?;
+        let return_table = self.maybe_parse(|p| {
+            let return_table_name = p.parse_identifier()?;
+
+            if !p.peek_keyword(Keyword::TABLE) {
+                parser_err!(
+                    "Expected TABLE keyword after return type",
+                    p.peek_token().span.start
+                )?
+            }
```
iffyio, line 5216:
```suggestion
            p.expect_keyword(Keyword::TABLE)?;
```

src/parser/mod.rs
```diff
@@ -5204,19 +5204,79 @@ impl<'a> Parser<'a> {
...
+                    None => parser_err!(
+                        "Expected table column definitions after TABLE keyword",
+                        p.peek_token().span.start
+                    )?,
+                },
+                _ => parser_err!(
+                    "Expected table data type after TABLE keyword",
+                    p.peek_token().span.start
+                )?,
+            };
```
iffyio, line 5239:
```suggestion
                DataType::Table(Some(table_column_defs)) !if table_column_defs.is_empty() => table_column_defs,
                _ => parser_err!(
                    "Expected table data type after TABLE keyword",
                    p.peek_token().span.start
                )?,
            };
```
looks like this condition can be simplified?
Also could we add a negative test to assert the behavior on invalid data_types, noticed it seems to be lacking coverage in the PR tests

src/parser/mod.rs
```diff
@@ -5204,19 +5204,79 @@ impl<'a> Parser<'a> {
...
+                Some(CreateFunctionBody::AsReturnExpr(self.parse_expr()?))
+            } else if self.peek_keyword(Keyword::SELECT) {
+                let select = self.parse_select()?;
+                Some(CreateFunctionBody::AsReturnSelect(select))
+            } else {
+                parser_err!(
+                    "Expected a subquery (or bare SELECT statement) after RETURN",
+                    self.peek_token().span.start
+                )?
+            }
```
iffyio, line 5276: can we add a test for this behavior? (e.g. one that passes a regular expression and the verifies that the parser rejects it)

src/parser/mod.rs
aharpervc: The reason I didn't do it this way is because expect_keyword (and expect_keyword_is) consume the token. That causes parse_data_type to break, because it uses the TABLE keyword to understand it should parse the table data type.

However, I suppose that I could just call `prev_token()` after expect to undo consuming the token. I will make this change.

src/parser/mod.rs
aharpervc: Done 👍

src/parser/mod.rs

iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1840, opened 2025-05-06
Add support for the MATCH and REGEXP binary operators

lovasoa:
In SQLite, MATCH and REGEXP are binary operators.
SQLite treats them like function calls, but this implementation preserves the original binary expressions.

**real-world examples**:

```sql
SELECT * FROM email WHERE email MATCH 'fts5';
```

```sql
select count(*) from messages where msg_text regexp '\d+';
```

**documentation**:

 - https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators
 - https://sqlite.org/fts5.html


iffyio (approved):
LGTM! Thanks @lovasoa!
cc @alamb

PR #1842, opened 2025-05-07
fix: allow arbitrary operators with ANY and ALL on Postgres

freshtonic:
In sqlparser PR #963 a check was introduced which limits which operators can be used with `ANY` and `ALL` expressions.

Postgres can parse more (possibly _all_ binary operators, investigation pending) in this location. Postgres only seems to care that the operator yields a boolean - which is a semantic error, not a syntax (parse) error.

Example of semantic error in Postgres:

```
select 123 % ANY(array[246]);
ERROR:  op ANY/ALL (array) requires operator to yield boolean
LINE 1: select 123 % ANY(array[246]);
                   ^
```

FWIW, the Postgres `pg_trgm` extension provides support for the `%` where left and right args are `text` and it returns a `boolean` - which makes it useable with `ANY` and `ALL`.

The following code in `src/parser/mod.rs:2893-2908` is where the allowlist of operators is enforced:

```rust
if !matches!(
    op,
    BinaryOperator::Gt
        | BinaryOperator::Lt
        | BinaryOperator::GtEq
        | BinaryOperator::LtEq
        | BinaryOperator::Eq
        | BinaryOperator::NotEq
) {
    return parser_err!(
        format!(
        "Expected one of [=, >, <, =>, =<, !=] as comparison operator, found: {op}"
    ),
        tok.span.start
    );
};
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3471,7 +3471,7 @@ impl<'a> Parser<'a> {
                     right
                 };
 
-                if !matches!(
+                if !dialect_of!(self is PostgreSqlDialect) && !matches!(
                     op,
                     BinaryOperator::Gt
                         | BinaryOperator::Lt
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -5937,6 +5937,36 @@ fn parse_bitstring_literal() {
 }
 
 #[test]
+fn parse_arbitrary_operator_with_any_and_all() {
+    // The following statement is semantically non-sensical on vanilla postgres but it should *parse* in order to
+    // support situations where PG operators have been overloaded.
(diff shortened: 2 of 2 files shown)


src/parser/mod.rs
```diff
@@ -3471,7 +3471,7 @@ impl<'a> Parser<'a> {
                     right
                 };
 
-                if !matches!(
+                if !dialect_of!(self is PostgreSqlDialect) && !matches!(
```
iffyio, line 3474: can we use the dialect trait for this logic (dialect_of is effectively deprecated internally)? something like `if self.dialect.supports_arbitrary_comparison_in_subquery_operator()`

iffyio:
> Maybe we can remove the check entirely ?

Ah indeed I think this would make more sense here, it should be fine to remove the check entirely

PR #1848, opened 2025-05-11
Fix big performance issue in string serialization

lovasoa:
The old code was handling string escaping character by character. For every literal string in the AST, it would push it to the underlying writer **character by character**, resulting in thousands of write calls for long strings.

The new code calls the write function only once, with the entire string, in most cases.

Only when the string is stored unescaped do we really need to call write multiple times; and even then, we don't need to call it more than the total number of characters to escape plus one.

Here are benchmark results for serializing the following sql statement: `"SELECT 'xxx...(x 10000)' as long_string"` to a string in memory:

```
sqlparser-rs parsing benchmark/format_long_string
                        time:   [10.544 µs 10.645 µs 10.743 µs]
                        change: [-84.871% -84.195% -83.553%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```

image

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -462,16 +462,20 @@ impl fmt::Display for EscapeQuotedString<'_> {
             match ch {
                 char if char == quote => {
                     if previous_char == '\\' {
+                        // the quote is already escaped with a backslash, skip
                         peekable_chars.next();
                         continue;
                     }
                     peekable_chars.next();
                     match peekable_chars.peek() {
                         Some((_, c)) if *c == quote => {
+                            // the quote is already escaped with another quote, skip
                             peekable_chars.next();
                         }
                         _ => {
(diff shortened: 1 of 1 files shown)


jayzhan211 (approved):

alamb:
When you say "old" code do you know what PR introduced this regression?

alamb (approved):
looks good to me -- thanks @lovasoa

lovasoa:
@alamb : I meant that the code before this PR handled the string char by char. I don't think this was a regression.

alamb:
Thanks @lovasoa and @jayzhan211  for the review

cc @iffyio

lovasoa:
thanks for merging !

Issue #1850, opened 2025-05-13
Add pretty printing to more sql constructs


lovasoa:
This is a followup on https://github.com/apache/datafusion-sqlparser-rs/pull/1847

Here are some constructs that are currently not handled by the pretty printer (they are displayed on a single line). I'm opening this to serve as a tracking issue to track implementation of these. 

* [x] Format `INSERT INTO` with both `VALUES` and `SELECT` forms
* [x] Support `UPDATE` and `DELETE` with proper clause indentation
* [x] indent `CREATE TABLE` column definitions and constraints
* [x] Format `CREATE VIEW` and nested `SELECT` cleanly
* [ ] Handle `CREATE FUNCTION`/`PROCEDURE` blocks with inner SQL formatting
* [ ] format `json_table` and `xmltable`
* [ ] Format `BEGIN`, `COMMIT`, `ROLLBACK` blocks clearly
* [ ] Support control flow structures (`IF`, `WHILE`, `LOOP`) with block indentation
* [ ] support multiline strings (the indentation should not be reflected inside the string)
* [ ] Handle `MERGE` statements with `WHEN MATCHED` and `WHEN NOT MATCHED` branches
* [ ] format `CREATE INDEX` with minimal structure
* [ ] Handle `EXPLAIN`, `EXPLAIN ANALYZE`, and `DESCRIBE` statements

cc @alamb

hengfeiyang:
For this sql, the format also is not correct:
```sql
SELECT cid,COUNT(*)AS cnt FROM tbl WHERE hostname IN('www.google.com','www.abc.com')AND cid IN(SELECT DISTINCT cid FROM tbl2 WHERE hostname='www.abc.com' AND graphqlop='test' AND status IN(SELECT DISTINCT status FROM tbl3)ORDER BY cid LIMIT 1000)GROUP BY cid ORDER BY cid;
```

currently version output is:
```sql
SELECT
  cid,
  COUNT(*) AS cnt
FROM
  tbl
WHERE
  hostname IN ('www.google.com', 'www.abc.com') AND cid IN (SELECT DISTINCT cid FROM tbl2 WHERE hostname = 'www.abc.com' AND graphqlop = 'test' AND status IN (SELECT DISTINCT status FROM tbl3) ORDER BY cid LIMIT 1000)
GROUP BY
  cid ORDER BY cid
```

PR #1851, opened 2025-05-13
pretty print improvements

lovasoa:
this is a first PR for https://github.com/apache/datafusion-sqlparser-rs/issues/1850
mainly implementing pretty printing for insert update and delete statements

- **fix pretty-printing of multi-line literal strings**
- **readme example for pretty printing**
- **Add pretty-printing tests for various SQL statements**
- **pretty print update**
- **pretty-print DELETE**
- **pretty-printing for INSERT statements and VALUES clause**

### DIFF
--- a/src/display_utils.rs
+++ b/src/display_utils.rs
@@ -68,12 +68,17 @@ impl Display for SpaceOrNewline {
 
 /// A value that displays a comma-separated list of values.
 /// When pretty-printed (using {:#}), it displays each value on a new line.
-pub(crate) struct DisplayCommaSeparated<'a, T: fmt::Display>(&'a [T]);
+pub(crate) struct DisplayCommaSeparated<I, T: fmt::Display>(I)
+where
+    I: IntoIterator<Item = T> + Clone;
 
-impl<T: fmt::Display> fmt::Display for DisplayCommaSeparated<'_, T> {
+impl<I, T: fmt::Display> fmt::Display for DisplayCommaSeparated<I, T>
+where
+    I: IntoIterator<Item = T> + Clone,
+{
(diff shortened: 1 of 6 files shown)


src/display_utils.rs
```diff
@@ -71,12 +68,17 @@ impl Display for SpaceOrNewline {
...
 
-impl<T: fmt::Display> fmt::Display for DisplayCommaSeparated<'_, T> {
+impl<I, T: fmt::Display> fmt::Display for DisplayCommaSeparated<I, T>
+where
+    I: IntoIterator<Item = T> + Clone,
+{
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut first = true;
-        for t in self.0 {
+        for t in self.0.clone() {
```
iffyio, line 81: was the clone needed here, we seem to only call `fmt` on `t`?

lovasoa: Yes, we are never actually cloning anything expensive here, but we need to add the `Clone` generic bound in order to be able to use the `IntoIterator` trait.

This `.clone()` concretely always clones
 - either just a reference
 - or an iterator object (but not the underlying structure being iterated on)

lovasoa: @iffyio , I can roll it back to the non-generic version, and just implement comma separation manually in the Values Display impl, if you want.

iffyio: Yeah I think it might be better to make it explicit, with the new impl seems like it wouldn't be obvious if we were to go on to later clone something expensive accidentally

lovasoa: @iffyio : I reverted the generic `indented_list` implementation and reworked the Values Display implementation not to use it.

PR #1852, opened 2025-05-15
Mysql: Add `SRID` column option

MohamedAbdeen21:
Add support for Mysql geometric `SRID {int}` column option in create table statements.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7754,6 +7754,10 @@ impl<'a> Parser<'a> {
             && dialect_of!(self is MySqlDialect | SQLiteDialect | DuckDbDialect | GenericDialect)
         {
             self.parse_optional_column_option_as()
+        } else if self.parse_keyword(Keyword::SRID)
+            && dialect_of!(self is MySqlDialect | GenericDialect)
+        {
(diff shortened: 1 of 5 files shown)


MohamedAbdeen21:
I give up on the CI. I understand the error, but idk why it's showing specifically on this PR.

Clippy passes locally, I guess that the CI doesn't use a fixed cargo/clippy version, and that these warnings were introduced recently.

src/parser/mod.rs
```diff
@@ -16571,6 +16575,23 @@ mod tests {
...
+        if let Statement::CreateTable(v) = &ast[0] {
+            assert_eq!(
+                v.columns[0].options[0].option,
+                ColumnOption::Srid(Box::new(Expr::value(Value::Number(
+                    "4326".parse().unwrap(),
+                    false
+                ))))
+            );
+        }
+    }
```
iffyio, line 16593: can we instead extend one of the create table tests in either common or mysql tests file? (essentially this test ideally would live in one of those files I think)

MohamedAbdeen21: Moved to mysql tests file. Rewritten to use `verified_stmt`

iffyio (approved):
cc @alamb

PR #1856, opened 2025-05-16
fix new rust 1.87 cargo clippy warnings

lovasoa:
all PRs are currently red because of the new rust release


iffyio (approved):
LGTM! Thanks @lovasoa!
cc @alamb

PR #1860, opened 2025-05-26
Add support for `TABLESAMPLE` pipe operator

hendrikmakait:
Part of #1758

This PR adds support for the `|> TABLESAMPLE ...` pipe operator.

Open question:

[BigQuery's `TABLESAMPLE` operator](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#tablesample_pipe_operator) does not support the full [SQL standard spec](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#sample-clause). This PR adds support for a superset of the BigQuery syntax which, e.g., includes the `REPEATABLE` clause and reuses the existing code for parsing `TABLESAMPLE` operators. Is this desired from the perspective of the project or is it preferred to stick closely to the BigQuery syntax? If so, I'm happy to change this.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2707,7 +2707,7 @@ impl fmt::Display for PipeOperator {
             PipeOperator::Limit { expr, offset } => {
                 write!(f, "LIMIT {}", expr)?;
                 if let Some(offset) = offset {
-                    write!(f, "OFFSET {}", offset)?;
+                    write!(f, " OFFSET {}", offset)?;
                 }
                 Ok(())
             }
@@ -2719,12 +2719,12 @@ impl fmt::Display for PipeOperator {
                 if !full_table_exprs.is_empty() {
                     write!(
                         f,
-                        "{}",
+                        " {}",
                         display_comma_separated(full_table_exprs.as_slice())
                     )?;
                 }
                 if !group_by_expr.is_empty() {
-                    write!(f, "GROUP BY {}", display_comma_separated(group_by_expr))?;
+                    write!(f, " GROUP BY {}", display_comma_separated(group_by_expr))?;
                 }
                 Ok(())
             }
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -104,7 +104,7 @@ impl fmt::Display for Query {
             format.fmt(f)?;
         }
         for pipe_operator in &self.pipe_operators {
-            f.write_str(" |>")?;
+            f.write_str(" |> ")?;
             pipe_operator.fmt(f)?;
         }
         Ok(())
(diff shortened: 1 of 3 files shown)


src/ast/query.rs
```diff
@@ -2680,28 +2680,32 @@ pub enum PipeOperator {
         full_table_exprs: Vec<ExprWithAliasAndOrderBy>,
         group_by_expr: Vec<ExprWithAliasAndOrderBy>,
     },
+    /// Selects a random sample of rows from the input table.
+    /// Syntax: `|> TABLESAMPLE <method> (<size> {ROWS | PERCENT})`
```
hendrikmakait, line 2684: This is the officially supported spec from the paper but doesn't cover everything that is technically supported in this PR. I'm wondering how to best deal with that (see the open question in the PR description).

src/ast/query.rs
```diff
@@ -2680,28 +2680,32 @@ pub enum PipeOperator {
...
+    /// See more at <https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#tablesample_pipe_operator>
+    TableSample { sample: Box <TableSample> },
 }
 
 impl fmt::Display for PipeOperator {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             PipeOperator::Select { exprs } => {
-                write!(f, "SELECT {}", display_comma_separated(exprs.as_slice()))
+                write!(f, " SELECT {}", display_comma_separated(exprs.as_slice()))
```
iffyio, line 2693: Oh I think we'd ideally undo these changes, its better for the operator display to be standalone where possible (i.e. they shouldn't assume surrounding space formatting, rather it should be left up to the caller), it makes it easier to compose nodes in the AST when displaying

src/ast/query.rs
iffyio: I think the doc comment doesn't necessarily need to spell out the full spec. usually its enough to give a rough example of what the statement looks like e.g.
```sql
Syntax: `|> TABLESAMPLE BERNOULLI(50)
```

tests/sqlparser_common.rs
```diff
@@ -15155,6 +15155,12 @@ fn parse_pipeline_operator() {
     dialects.verified_stmt("SELECT * FROM users |> ORDER BY id DESC");
     dialects.verified_stmt("SELECT * FROM users |> ORDER BY id DESC, name ASC");
 
+    // tablesample pipe operator
+    dialects.verified_stmt("SELECT * FROM tbl |> TABLESAMPLE BERNOULLI (50)");
+    dialects.verified_stmt("SELECT * FROM tbl |> TABLESAMPLE SYSTEM (50)");
+    // TODO: Technically, REPEATABLE is not available in BigQuery, but it is used with TABLESAMPLE in other dialects
```
iffyio, line 15161:
```suggestion
```
Yeah I think this behavior would be fine

iffyio (commented):
Thanks @hendrikmakait! The changes look reasonable to me overall, left a couple minor comments

iffyio:
@hendrikmakait could you take a look at the CI failures?

alamb:
I restarted the checks

alamb:
🚀

PR #1862, opened 2025-05-26
Keep the COLUMN keyword only if it exists when dropping the column

### LINKED ISSUE #1859: Postgres: `DROP COLUMN c` reformatted to `DROP c`
Given the following input:

```sql
ALTER TABLE logs DROP COLUMN details
```

the following code outputs the statement as:

```sql
ALTER TABLE logs DROP details
```

the code:

```rust
let stmts = Parser::parse_sql(&PostgreSqlDialect {}, sql).unwrap();

// grab the first statement
// let stmt = ...
stmt.to_string()
```

I understand that the two are functionally equivalent, however according to the README I would expect that the `COLUMN` keyword is preserved:

> This crate allows users to recover the original SQL text (with comments removed, normalized whitespace and keyword capitalization), which is useful for tools that analyze and manipulate SQL.

Would you classify this as a bug?

git-hulk:
This closes #1859.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -139,6 +139,7 @@ pub enum AlterTableOperation {
     },
     /// `DROP [ COLUMN ] [ IF EXISTS ] <column_name> [ CASCADE ]`
     DropColumn {
+        has_column_keyword: bool,
         column_name: Ident,
         if_exists: bool,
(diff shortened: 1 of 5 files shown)


iffyio (approved):
LGTM! Thanks @git-hulk!
cc @alamb

Issue #1863, opened 2025-05-27
mysql: `DROP INDEX idx_xxx on t` and an error occurs


vimko:
With v0.56.0 using the mysql dialect, use the `parse_sql` function to parse `DROP INDEX idx_name ON table_demo;` or `ALTER TABLE tbl_name DROP INDEX idx_name;` and an error occurs.

```rust
use sqlparser::dialect::MySqlDialect;
use sqlparser::parser::Parser;

fn main() {
    let sql = "DROP INDEX idx_name ON table_demo;";

    let dialect = MySqlDialect {}; // or AnsiDialect, or your own dialect ...

    let ast = Parser::parse_sql(&dialect, sql).unwrap();

    println!("AST: {:?}", ast);
}

```

Report the following error:

```
called `Result::unwrap()` on an `Err` value: ParserError("Expected: end of statement, found: ON at Line: 1, Column: 21")
```

In the official documentation of MySQL, when deleting an index, you need to specify the table where the index is located.

```sql
DROP INDEX index_name ON tbl_name
    [algorithm_option | lock_option] ...

algorithm_option:
    ALGORITHM [=] {DEFAULT | INPLACE | COPY}

lock_option:
    LOCK [=] {DEFAULT | NONE | SHARED | EXCLUSIVE}

```

https://dev.mysql.com/doc/refman/8.0/en/drop-index.html

Would you classify this as a bug?

LucaCappelletti94:
Closing as resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/1864

PR #1864, opened 2025-05-27
Adds support for mysql's drop index

dmzmk:
Addresses #1863

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6249,10 +6249,10 @@ impl<'a> Parser<'a> {
                 loc
             );
         }
-        // Mysql requires table specification on index drop
-        let table = match self.parse_keyword(Keyword::ON) {
-            true => Some(self.parse_object_name(false)?),
-            false => None,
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -6249,6 +6249,11 @@ impl<'a> Parser<'a> {
                 loc
             );
         }
+        // Mysql requires table specification on index drop
+        let table = match self.parse_keyword(Keyword::ON) {
+            true => Some(self.parse_object_name(false)?),
+            false => None,
+        };
```
iffyio, line 6256:
```suggestion
        let table = if self.parse_keyword(Keyword::ON) {
            Some(self.parse_object_name(false)?)
        } else {
            None
        };
```

dmzmk: @iffyio thanks for the review! I've applied your suggestion

alamb:
The code just keeps churning along in this repo. Thank you @iffyio  and @dmzmk

PR #1865, opened 2025-05-28
Add support for `ALTER TABLE DROP INDEX`

vimko:
This PR adds support for mysql's drop index:

- ~~`DROP INDEX idx_a ON table_a`~~
- `ALTER TABLE table_a DROP INDEX idx_a`

Addresses https://github.com/apache/datafusion-sqlparser-rs/issues/1863


src/ast/ddl.rs
```diff
@@ -187,6 +187,14 @@ pub enum AlterTableOperation {
     DropForeignKey {
         name: Ident,
     },
+    /// `DROP INDEX <index_name>`
+    ///
+    /// Note: this is a [MySQL]-specific operation.
+    ///
```
```suggestion
```
I think we can skip the note, since we're already linking to the mysql docs below

tests/sqlparser_common.rs
```diff
@@ -9132,7 +9132,9 @@ fn test_create_index_with_with_clause() {
 #[test]
 fn parse_drop_index() {
     let sql = "DROP INDEX idx_a";
-    match verified_stmt(sql) {
+    // MySql dialect doesn't support `DROP INDEX idx_a`,you need to specify a specific table, please refer:
+    // [MySql](https://dev.mysql.com/doc/refman/8.4/en/drop-index.html)
+    match all_dialects_except(|d| d.is::<MySqlDialect>()).verified_stmt(sql) {
```
iffyio: not sure I followed the intent here, from the code in this PR the parser doesnt seem to have started rejecting any `DROP INDEX` statement variants specifically for mysql, so that I imagined these should still have been passing for mysql?

vimko: That's right, I didn't think it through well.

src/ast/ddl.rs
vimko: Yes, this is more concise, so I don't need to specify that only MySQL supports it.

iffyio:
@vimko could you take a look at the cargo failure when you get the chance?

vimko:
> @vimko could you take a look at the cargo failure when you get the chance?

@iffyio Thank you for the reminder, I have already taken care of it.

PR #1866, opened 2025-05-28
Add support for parameter default values in SQL Server

aharpervc:
This PR adds support for default values for parameters, as documented here: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver17#--default-

The concept of a parameter default value already existed on the `OperateFunctionArg` struct, but was hardcoded to `None`. That logic has been updated to conditionally consume `=` & parse an Expr.

Since this is relevant for SQL Server, a new test was added to `sqlparser_mssql.rs` accordingly.


iffyio (approved):
LGTM! Thanks @aharpervc!
cc @alamb

PR #1867, opened 2025-05-30
Fix: GROUPING SETS accept values without parenthesis

Vedin:
`GROUPING SETS` currently accepts only tuples. It can be single elements as well. 
At least, I can see this behaviour in Hive and Snowflake:
https://hive.apache.org/docs/latest/30151323/
https://docs.snowflake.com/en/sql-reference/constructs/group-by-grouping-sets
This PR fixes this.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2273,6 +2273,16 @@ impl<'a> Parser<'a> {
         }
     }
 
+    // Helper function to parse an element within GROUPING SETS
+    // This element can either be a single expression or a parenthesized list of expressions.
+    fn parse_grouping_set_element(&mut self) -> Result<Vec<Expr>, ParserError> {
+        if self.peek_token_ref().token == Token::LParen {
+            self.parse_tuple(true, true)
+        } else {
+            Ok(vec![self.parse_expr()?])
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -10045,7 +10055,7 @@ impl<'a> Parser<'a> {
             }
             if self.parse_keywords(&[Keyword::GROUPING, Keyword::SETS]) {
                 self.expect_token(&Token::LParen)?;
-                let result = self.parse_comma_separated(|p| p.parse_tuple(true, true))?;
+                let result = self.parse_comma_separated(|p| p.parse_grouping_set_element())?;
```
iffyio: could we inline the logic vs declaring it as a function? since the logic is only used once and small enough,

```rust
let result = self.parse_comma_separated(|p| {
            self.parse_tuple(true, true)
        } else {
            Ok(vec![self.parse_expr()?])
        }
})?;
```

iffyio (commented):
Thanks @Vedin! left a minor comment otherwise this looks good to me

src/parser/mod.rs
Vedin: Sure, done!

Issue #1868, opened 2025-06-02
Snowflake: ALTER ICEBERG TABLE


osipovartem:
https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table
Icecerg keyword is not supported yet.

Options to implement:
- extent alter table with bool iceberg flag
- add new AlterTableSnowflake and put there only snowflake specific syntax ([doc link](https://docs.snowflake.com/en/sql-reference/sql/alter-table))

PR #1869, opened 2025-06-02
Add ICEBERG keyword support to ALTER TABLE statement

osipovartem:
Related to #1868

This change extends the existing ALTER TABLE parsing logic by introducing an optional ICEBERG keyword. Instead of creating a separate AlterTableIceberg variant, this approach adds a simple bool iceberg flag to the existing Statement::AlterTable structure.


alamb:
ICEBERG!

PR #1870, opened 2025-06-03
MySQL: `[[NOT] ENFORCED]` in CHECK constraint

MohamedAbdeen21:
Add support for MySQL's `[[NOT] ENFORCED]` option for CHECK constraints.

docs: https://dev.mysql.com/doc/refman/8.4/en/create-table.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8139,7 +8139,19 @@ impl<'a> Parser<'a> {
                 self.expect_token(&Token::LParen)?;
                 let expr = Box::new(self.parse_expr()?);
                 self.expect_token(&Token::RParen)?;
-                Ok(Some(TableConstraint::Check { name, expr }))
+
+                let enforced = match dialect_of!(self is GenericDialect | MySqlDialect) {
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -8134,7 +8134,19 @@ impl<'a> Parser<'a> {
                 self.expect_token(&Token::LParen)?;
                 let expr = Box::new(self.parse_expr()?);
                 self.expect_token(&Token::RParen)?;
-                Ok(Some(TableConstraint::Check { name, expr }))
+
+                let enforced = match dialect_of!(self is GenericDialect | MySqlDialect) {
```
iffyio: I think we can have the parser always accept the `NOT ENFORCED` keywords if they show up in the statement? Otherwise we'd want to turn this into a dialect method instead of using the dialect_of macro but that feels like it might not be worth it in this case

MohamedAbdeen21: Updated and rebased on main

iffyio (approved):
cc @alamb

PR #1872, opened 2025-06-03
Postgres: Apply `ONLY` keyword per table in TRUNCATE stmt

MohamedAbdeen21:
The `ONLY` keyword was previously applied to the entire stmt, the correct psql behaviour is to apply the keyword per table.


iffyio (approved):
cc @alamb

PR #1874, opened 2025-06-05
Fix `CASE` expression spans

eliaperantoni:
Previously, a `Expr::Case` implementation of `Spanned` would ignore the `CASE .. END` keywords. ie.

```sql
CASE a WHEN b THEN c ELSE d END
--   ^^^^^^^^^^^^^^^^^^^^^^
--   Only this would be highlighted
```

Because the only stored locations would be those of the `Ident` _inside_ the `CASE` expression.

This PR attaches two `AttachedToken` to the AST node, one for `CASE` and one for `END`, so that `Spanned::span` now returns the whole range. This was done to follow the footsteps of the `SELECT` implementation, which also stores an `AttachedToken` for the `SELECT` keyword. Also added a test for this.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -14469,16 +14469,6 @@ fn test_case_statement_span() {
 }
 
 #[test]
-fn test_case_expr_span() {
-    let sql = "CASE 1 WHEN 2 THEN 3 ELSE 4 END";
-    let mut parser = Parser::new(&GenericDialect {}).try_with_sql(sql).unwrap();
-    assert_eq!(
-        parser.parse_expr().unwrap().span(),
-        Span::new(Location::new(1, 1), Location::new(1, sql.len() as u64 + 1))
-    );
-}
-
-#[test]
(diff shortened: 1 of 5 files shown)


tests/sqlparser_common.rs
```diff
@@ -14464,6 +14468,16 @@ fn test_case_statement_span() {
     );
 }
 
+#[test]
```
iffyio: Could we move the test case to the [spans.rs](https://github.com/apache/datafusion-sqlparser-rs/blob/6cf531d4951545708a82b208bed8ea596f91a186/src/ast/spans.rs#L2333) tests?

iffyio (commented):
Thanks @eliaperantoni! The changes look reasonable to me, left a minor comment

tests/sqlparser_common.rs
eliaperantoni: Absolutely! I moved it and aligned its body with the approach taken in the rest of the tests in `spans.rs`. Thank you for the review 🥰!

PR #1876, opened 2025-06-09
chore: Replace archived actions-rs/install action 

assignUser:
The actions-rs org has been archived for years at this point and it doesn't look like it's coming back. For security reasons we are going to remove it from the allow list soon.

In this PR I have replaced the usage of `actions-rs/install` with a well known rust cache action and a normal `cargo install`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -88,11 +88,8 @@ jobs:
       uses: ./.github/actions/setup-builder
       with:
         rust-version: ${{ matrix.rust }}
+    - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2.7.8
     - name: Install Tarpaulin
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -85,11 +88,8 @@ jobs:
       uses: ./.github/actions/setup-builder
       with:
         rust-version: ${{ matrix.rust }}
+    - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2.7.8
```
alamb, line 91: I checked that we use the same action in DataFusion so this sounds like a good idea to me 👍 


https://github.com/apache/datafusion/blob/b41acf3c43ff259a22f9ecc2abdb17db58297fd8/.github/actions/setup-macos-aarch64-builder/action.yaml#L47

Thanks @assignUser and @iffyio

alamb (approved):

PR #1879, opened 2025-06-10
Support remaining pipe operators

simonvandel:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1758

This PR adds support for the remaining SQL pipe operators as defined by BigQuery.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2823,10 +2823,7 @@ impl fmt::Display for PipeOperator {
             } => Self::fmt_set_operation(f, "EXCEPT", set_quantifier, queries),
             PipeOperator::Call { function, alias } => {
                 write!(f, "CALL {}", function)?;
-                if let Some(alias) = alias {
-                    write!(f, " AS {}", alias)?;
-                }
-                Ok(())
+                Self::fmt_optional_alias(f, alias)
             }
             PipeOperator::Pivot {
                 aggregate_functions,
@@ -2841,10 +2838,7 @@ impl fmt::Display for PipeOperator {
                     Expr::CompoundIdentifier(value_column.to_vec()),
                     value_source
                 )?;
-                if let Some(alias) = alias {
-                    write!(f, " AS {}", alias)?;
-                }
-                Ok(())
+                Self::fmt_optional_alias(f, alias)
             }
             PipeOperator::Unpivot {
                 value_column,
@@ -2859,16 +2853,21 @@ impl fmt::Display for PipeOperator {
                     name_column,
                     display_comma_separated(unpivot_columns)
                 )?;
-                if let Some(alias) = alias {
-                    write!(f, " AS {}", alias)?;
-                }
-                Ok(())
+                Self::fmt_optional_alias(f, alias)
             }
         }
     }
 }
 
 impl PipeOperator {
+    /// Helper function to format optional alias for pipe operators
+    fn fmt_optional_alias(f: &mut fmt::Formatter<'_>, alias: &Option<Ident>) -> fmt::Result {
+        if let Some(alias) = alias {
+            write!(f, " AS {}", alias)?;
+        }
+        Ok(())
+    }
+
     /// Helper function to format set operations (UNION, INTERSECT, EXCEPT) with queries
     fn fmt_set_operation(
         f: &mut fmt::Formatter<'_>,
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2695,7 +2695,16 @@ pub enum PipeOperator {
     /// Syntax: `|> UNION [ALL|DISTINCT] (<query>), (<query>), ...`
     ///
     /// See more at <https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#union_pipe_operator>
-    Union { 
+    Union {
+        set_quantifier: SetQuantifier,
+        queries: Vec<Box<Query>>,
+    },
+    /// Returns only the rows that are present in both the input table and the specified tables.
+    ///
+    /// Syntax: `|> INTERSECT [DISTINCT] (<query>), (<query>), ...`
+    ///
+    /// See more at <https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#intersect_pipe_operator>
+    Intersect {
         set_quantifier: SetQuantifier,
         queries: Vec<Box<Query>>,
     },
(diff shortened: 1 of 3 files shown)


src/ast/query.rs
```diff
@@ -2739,7 +2812,91 @@ impl fmt::Display for PipeOperator {
...
+                write!(f, " {}", set_quantifier)?;
+            }
+        }
+        write!(f, " ")?;
+        for (i, query) in queries.iter().enumerate() {
+            if i > 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "({})", query)?;
+        }
```
iffyio: can this use the `display_comma_separated()` helper function?

src/parser/mod.rs
```diff
@@ -9947,6 +9947,51 @@ impl<'a> Parser<'a> {
         Ok(IdentWithAlias { ident, alias })
     }
 
+    /// Parse `identifier [AS] identifier` where the AS keyword is optional
+    pub fn parse_identifier_with_optional_alias(&mut self) -> Result<IdentWithAlias, ParserError> {
```
```suggestion
    fn parse_identifier_with_optional_alias(&mut self) -> Result<IdentWithAlias, ParserError> {
```

src/parser/mod.rs
```diff
@@ -9947,6 +9947,51 @@ impl<'a> Parser<'a> {
...
+        if self.parse_keywords(&[Keyword::DISTINCT, Keyword::BY, Keyword::NAME]) {
+            Ok(SetQuantifier::DistinctByName)
+        } else if self.parse_keyword(Keyword::DISTINCT) {
+            Ok(SetQuantifier::Distinct)
+        } else {
+            Err(ParserError::ParserError(format!(
+                "{} pipe operator requires DISTINCT modifier",
+                operator_name
+            )))
+        }
```
iffyio: Thinking since this block is a subset of the set_quantifier parsing logic [here](https://github.com/apache/datafusion-sqlparser-rs/blob/552def1f9ec99919222e41e18a8a517eaf582b93/src/parser/mod.rs#L11693-L11707), it might be good to reuse? one way could be we pull the latter out into a function and call it here, rejecting any returned upexpected variants. another would be that we update the latter to call this `parse_distinct_required_set_quantifier()` function

src/parser/mod.rs
```diff
@@ -9947,6 +9947,51 @@ impl<'a> Parser<'a> {
...
+    }
+
+    /// Parse optional alias (with or without AS keyword) for pipe operators
+    fn parse_optional_pipe_alias(&mut self) -> Result<Option<Ident>, ParserError> {
+        if self.parse_keyword(Keyword::AS) {
+            Some(self.parse_identifier()).transpose()
```
```suggestion
            Ok(Some(self.parse_identifier()?))
```
thinking this variant would be more explicit in this case? (the transpose seemed to obfuscate a bit what was going on)

src/parser/mod.rs
```diff
@@ -9947,6 +9947,51 @@ impl<'a> Parser<'a> {
...
+        } else {
+            Err(ParserError::ParserError(format!(
+                "{} pipe operator requires DISTINCT modifier",
+                operator_name
+            )))
+        }
+    }
+
+    /// Parse optional alias (with or without AS keyword) for pipe operators
+    fn parse_optional_pipe_alias(&mut self) -> Result<Option<Ident>, ParserError> {
```
```suggestion
    fn parse_identifier_optional_alias(&mut self) -> Result<Option<Ident>, ParserError> {
```
maybe we could rename this to something more generic? feels like it could be reusable in other contexts than pipe operators

src/parser/mod.rs
```diff
@@ -11142,6 +11200,217 @@ impl<'a> Parser<'a> {
...
+                    let alias = self.parse_optional_pipe_alias()?;
+
+                    pipe_operators.push(PipeOperator::Unpivot {
+                        value_column,
+                        name_column,
+                        unpivot_columns,
+                        alias,
+                    });
+                }
+                Keyword::JOIN => {
```
iffyio: for the joins, I'm wondering if its possible to pull out and reuse [the existing logic here](https://github.com/apache/datafusion-sqlparser-rs/blob/552def1f9ec99919222e41e18a8a517eaf582b93/src/parser/mod.rs#L12606-L12702)? e.g. here we'll `prev_token()` if needed whenever we see the next operator looks like a join and delegate to that helper function?

src/parser/mod.rs
```diff
@@ -11142,6 +11200,217 @@ impl<'a> Parser<'a> {
...
+                    let relation = self.parse_table_factor()?;
+                    let constraint = self.parse_join_constraint(false)?;
+                    if matches!(constraint, JoinConstraint::None) {
+                        return Err(ParserError::ParserError(
+                            "JOIN in pipe syntax requires ON or USING clause".to_string(),
+                        ));
+                    }
```
iffyio: thinking these constraint validations might not be ideal and we can skip them? in case the pipe syntax is later ported to other dialects, there the semantics might differ between dialects, [prev related behavior](https://github.com/apache/datafusion-sqlparser-rs/pull/1552)

src/ast/query.rs
simonvandel: Thanks, https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/18ea4ba7c84038c5285420c75d0bd53f44161412 and https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/330a9252afb0b6f637d92cc8b4df4e738ad9ffb3

src/parser/mod.rs
simonvandel: Thanks, https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/15e4e32beef24cc4bbe581dfe1a918f101cb6f8e

src/parser/mod.rs
simonvandel: Thanks, https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/370d119ebabe086d3363cf1c273b691d24ee993a

src/parser/mod.rs
simonvandel: Thanks, https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/774daf56618624c51f63e0d4fbce8c0011662ee5

src/parser/mod.rs
simonvandel: Thanks, https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/324ad14701756dc7b4e0c67a8384697a197527c7

src/parser/mod.rs
simonvandel: Removed in https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/324ad14701756dc7b4e0c67a8384697a197527c7

src/parser/mod.rs
simonvandel: Thanks https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/bfbd0086d541ac2d6a7279fd877646ccf071939a

simonvandel:
Hi @iffyio and thank you for the review.
I believe I have addressed all your comments in the latest commits

src/parser/mod.rs
```diff
@@ -11200,6 +11255,117 @@ impl<'a> Parser<'a> {
...
+                Keyword::JOIN
+                | Keyword::INNER
+                | Keyword::LEFT
+                | Keyword::RIGHT
+                | Keyword::FULL
+                | Keyword::CROSS => {
+                    self.prev_token();
+                    let mut joins = self.parse_joins()?;
+                    // Take first
+                    let join = joins.swap_remove(0);
```
iffyio: here can we add a condition to guard the vec indexing, so that we don't panic if our assumption or code is wrong for any reason? i.e.
```
if joins.len() != 1 {
   return Err(...)
}
let join = joins.swap_remove(0);
```
And maybe for the comment we can elaborate on if/why we expect exactly one entry in the returned join response? thinking since that might not be obvious for folks that stumble on this part of the code later on

iffyio (commented):
Thanks @simonvandel! Just one comment re vec index and I think this should be good!

src/parser/mod.rs
simonvandel: Good idea! https://github.com/apache/datafusion-sqlparser-rs/pull/1879/commits/552714aabf4ef17bf2e127631fcbde9e6e446d8e

iffyio (approved):
LGTM! Thanks @simonvandel!
cc @alamb

PR #1880, opened 2025-06-11
Support `DISTINCT AS { STRUCT | VALUE }` for BigQuery

bombsimon:
According to the query syntax at
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_list you can combine `[ { ALL | DISTINCT } ]` with `[ AS { STRUCT | VALUE } ]`:

```sh
SELECT
  [ WITH differential_privacy_clause ]
  [ { ALL | DISTINCT } ]
  [ AS { STRUCT | VALUE } ]
  select_list
```

This adds support to parse `DISTINCT` or `ALL` as the first keyword after `SELECT` and adds two new variants to the `ValueTableMode` if defined.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -3338,11 +3338,8 @@ impl fmt::Display for OpenJsonTableColumn {
 }
 
 /// BigQuery supports ValueTables which have 2 modes:
-/// `SELECT AS STRUCT`
-/// `SELECT AS VALUE`
-///
-/// They can be combined with `[ { ALL | DISTINCT } ]`, e.g.
-/// `SELECT DISTINCT AS STRUCT`
+/// `SELECT [ALL | DISTINCT] AS STRUCT`
+/// `SELECT [ALL | DISTINCT] AS VALUE`
 ///
 /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#value_tables>
 /// <https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_list>
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -11505,31 +11505,7 @@ impl<'a> Parser<'a> {
         }
 
         let select_token = self.expect_keyword(Keyword::SELECT)?;
-        let value_table_mode = if dialect_of!(self is BigQueryDialect) {
-            if self.parse_keywords(&[Keyword::DISTINCT, Keyword::AS]) {
-                if self.parse_keyword(Keyword::VALUE) {
-                    Some(ValueTableMode::DistinctAsValue)
-                } else if self.parse_keyword(Keyword::STRUCT) {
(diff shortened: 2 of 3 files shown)


src/ast/query.rs
```diff
@@ -3340,20 +3340,29 @@ impl fmt::Display for OpenJsonTableColumn {
 /// BigQuery supports ValueTables which have 2 modes:
 /// `SELECT AS STRUCT`
 /// `SELECT AS VALUE`
+///
+/// They can be combined with `[ { ALL | DISTINCT } ]`, e.g.
+/// `SELECT DISTINCT AS STRUCT`
+///
```
```suggestion
/// `SELECT [DISTINCT] AS STRUCT`
/// `SELECT [DISTINCT] AS VALUE`
///
```
thinking we can do this instead to flag optionality. Or we can include the specific examples as doc strings on each variant

src/parser/mod.rs
```diff
@@ -11516,7 +11524,10 @@ impl<'a> Parser<'a> {
                 }
             } else {
                 None
-            };
+            }
+        } else {
+            None
```
iffyio: would it make sense to unnest the logic with something like?
```rust
if self.parse_keywords([Keyword::DISTINCT, Keyword::AS, Keyword::VALUE]) {
    Some()
} else if self.parse_keywords([Keyword::DISTINCT, Keyword::AS, Keyword::STRUCT]) {
    Some()
} else if self.parse_keywords([Keyword::AS, Keyword::VALUE]) {
    Some()
} else if self.parse_keywords([Keyword::AS, Keyword::STRUCT]) {
    Some()
} else {
    None
};
```
Also maybe we can move the logic to its own method now like `let value_table_mode = self.parse_value_table_mode()`, since its no longer trivial

tests/sqlparser_bigquery.rs
```diff
@@ -2377,3 +2377,14 @@ fn test_any_type() {
...
+
+#[test]
+fn test_select_distinct_as_struct_or_value() {
+    for sql in [
+        "SELECT DISTINCT AS STRUCT a, ABS(b) FROM UNNEST(c) AS T",
+        "SELECT DISTINCT AS VALUE a, ABS(b) FROM UNNEST(c) AS T",
+        "SELECT ARRAY(SELECT DISTINCT AS STRUCT a, b, ABS(c) AS c, ABS(d) AS d FROM UNNEST(e) AS T)",
+    ] {
+        bigquery().verified_stmt(sql);
+    }
```
iffyio: thinking we can extend [this test case](https://github.com/apache/datafusion-sqlparser-rs/blob/6a774dabfc3573679a3ba392244f20461c0cb0f5/tests/sqlparser_bigquery.rs#L2322) instead of adding a new one I think, since they're the same feature

iffyio (approved):
LGTM! Thanks @bombsimon!
cc @alamb

alamb:
woohoo

PR #1881, opened 2025-06-11
Allow `IF NOT EXISTS` after table name for Snowflake

bombsimon:
Although not documented under
https://docs.snowflake.com/en/sql-reference/sql/create-table, it is valid to put the `IF NOT EXIST` _after_ the table name when creating a table for Snowflake.

This adds support to check for `IF NOT EXIST` again after parsing the table name if it wasn't found before the name.

---

I wish I could find _some_ references of this, but sadly it's not that easy. Best I can do is to provide a screenshot from their UI executing a query with `IF NOT EXIST` after the table and show that it succeeds.

image

### DIFF
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -446,6 +446,18 @@ fn test_snowflake_create_table_if_not_exists() {
         }
         _ => unreachable!(),
     }
+
+    for sql in [
+        r#"CREATE TABLE IF NOT EXISTS "A"."B"."C" (v VARIANT)"#,
+        r#"CREATE TABLE "A"."B"."C" IF NOT EXISTS (v VARIANT)"#,
+        r#"CREATE TRANSIENT TABLE IF NOT EXISTS "A"."B"."C" (v VARIANT)"#,
+        r#"CREATE TRANSIENT TABLE "A"."B"."C" IF NOT EXISTS (v VARIANT)"#,
+    ] {
+        // We don't use `verified_stmt` here because the parser won't produce a 1:1 result of the
+        // query. We will always put the `IF NOT EXIST` before the column name when displaying the
+        // query.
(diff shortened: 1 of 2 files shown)


bombsimon:
As seen by my second commit I just saw further down that this seems to be true for all Snowflake parameters so this basically just extends the existing loop that checks for parameters to include the `IF NOT EXIST` which can show up anywhere (even after the schema).

tests/sqlparser_snowflake.rs
```diff
@@ -446,6 +446,18 @@ fn test_snowflake_create_table_if_not_exists() {
...
+    for sql in [
+        r#"CREATE TABLE IF NOT EXISTS "A"."B"."C" (v VARIANT)"#,
+        r#"CREATE TABLE "A"."B"."C" IF NOT EXISTS (v VARIANT)"#,
+        r#"CREATE TRANSIENT TABLE IF NOT EXISTS "A"."B"."C" (v VARIANT)"#,
+        r#"CREATE TRANSIENT TABLE "A"."B"."C" IF NOT EXISTS (v VARIANT)"#,
+    ] {
+        // We don't use `verified_stmt` here because the parser won't produce a 1:1 result of the
+        // query. We will always put the `IF NOT EXIST` before the column name when displaying the
+        // query.
+        assert!(snowflake().parse_sql_statements(sql).is_ok());
```
iffyio: ah for this behavior I think we can use [one_statement_parse_to](https://github.com/apache/datafusion-sqlparser-rs/blob/06c6b35e9739221b7e5c7bf28e77fb6e23246875/tests/sqlparser_snowflake.rs#L1008-L1011) that provides the canonical form to assert on

bombsimon: Perfect, updated to that one!

iffyio (approved):
LGTM! Thanks @bombsimon!
cc @alamb

PR #1884, opened 2025-06-12
Extend exception handling

bombsimon:
- Exception handling can handle multiple `WHEN` arms
- Exception can re-raise with `RAISE` keyword
- Snowflake can now also parse `BEGIN ... EXCEPTION ... END`

Example:

```sql
BEGIN
    SELECT 1;
EXCEPTION
    WHEN EXCEPTION_1 THEN
        SELECT 1;
    WHEN EXCEPTION_2 OR EXCEPTION_3 THEN
        SELECT 2;
        SELECT 3;
    WHEN OTHER THEN
        SELECT 4;
RAISE;
END;
```

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -5558,7 +5558,7 @@ impl fmt::Display for Statement {
                 transaction,
                 modifier,
                 statements,
-                exception: exception_handling,
+                exception,
                 has_end_keyword,
             } => {
                 if *syntax_begin {
@@ -5580,7 +5580,7 @@ impl fmt::Display for Statement {
                     write!(f, " ")?;
                     format_statement_list(f, statements)?;
                 }
-                if let Some(exception_when) = exception_handling {
+                if let Some(exception_when) = exception {
                     write!(f, " EXCEPTION")?;
                     for when in exception_when {
                         write!(f, " {when}")?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15136,7 +15136,7 @@ impl<'a> Parser<'a> {
     pub fn parse_begin_exception_end(&mut self) -> Result<Statement, ParserError> {
         let statements = self.parse_statement_list(&[Keyword::EXCEPTION, Keyword::END])?;
 
-        let exception_handling = if self.parse_keyword(Keyword::EXCEPTION) {
+        let exception = if self.parse_keyword(Keyword::EXCEPTION) {
             let mut when = Vec::new();
 
             // We can have multiple `WHEN` arms so we consume all cases until `END`
@@ -15170,7 +15170,7 @@ impl<'a> Parser<'a> {
         Ok(Statement::StartTransaction {
             begin: true,
             statements,
-            exception: exception_handling,
+            exception,
             has_end_keyword: true,
             transaction: None,
             modifier: None,


src/ast/mod.rs
```diff
@@ -2982,6 +2982,63 @@ impl From<Set> for Statement {
...
+        if let Some(raises) = &self.raises {
+            write!(f, " {raises};")?;
+        }
+
+        Ok(())
+    }
+}
+
+/// A representation of a `WHEN` arm with all the identifiers catched and the statements to execute
+/// for the arm.
```
iffyio: Can we add a link to the docs here as well?

src/ast/mod.rs
```diff
@@ -2982,6 +2982,63 @@ impl From<Set> for Statement {
...
 }
 
+/// An exception representing exception handling with the `EXCEPTION` keyword.
+///
+/// Snowflake: <https://docs.snowflake.com/en/sql-reference/snowflake-scripting/exception>
+/// BigQuery: <https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#beginexceptionend>
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct Exception {
```
```suggestion
pub struct ExceptionClause {
```

src/ast/mod.rs
```diff
@@ -2982,6 +2982,63 @@ impl From<Set> for Statement {
...
+    pub when: Vec<ExceptionWhen>,
+    pub raises: Option<Box<Statement>>,
```
iffyio: Can we add a description what raises refers to here? thinking since its being represetned as a box statement it won't be obvious

src/ast/mod.rs
```diff
@@ -3670,17 +3727,24 @@ pub enum Statement {
...
-        ///     SELECT 2;
-        ///     SELECT 3;
+        /// EXCEPTION
+        ///     WHEN EXCEPTION_1 THEN
+        ///         SELECT 2;
+        ///     WHEN EXCEPTION_2 OR EXCEPTION_3 THEN
+        ///         SELECT 3;
+        ///     WHEN OTHER THEN
+        ///         SELECT 4;
+        /// RAISE;
```
iffyio: given the syntax example, Im not sure I see the requirement for the introduced `Exception` struct - here the `RAISE` looks like a normal statement that belongs either to the `EXCEPTION` clause, or the `BEGIN` statement itself? If the former, could it be represented as another statement in the `Vec<Statement>`?, if the latter seems like its out of place in the new `Exception` struct?

src/ast/mod.rs
```diff
@@ -2982,6 +2982,63 @@ impl From<Set> for Statement {
...
+}
+
+impl Display for ExceptionWhen {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let idents = self
+            .idents
+            .iter()
+            .map(ToString::to_string)
+            .collect::<Vec<_>>()
+            .join(" OR ");
```
iffyio: Can this use [display_separated](https://github.com/apache/datafusion-sqlparser-rs/blob/f69e0d928c2ee6a72261a3dace68228d7ad79e96/src/ast/mod.rs#L151)?

src/ast/mod.rs
```diff
@@ -2982,6 +2982,63 @@ impl From<Set> for Statement {
...
+
+        write!(f, " WHEN {idents} THEN", idents = idents)?;
```
```suggestion
        write!(f, " WHEN {idents} THEN")?;
```

src/parser/mod.rs
```diff
@@ -15096,12 +15096,16 @@ impl<'a> Parser<'a> {
...
             modifier: None,
             statements: vec![],
-            exception_statements: None,
+            exception: None,
             has_end_keyword: false,
         })
     }
 
     pub fn parse_begin(&mut self) -> Result<Statement, ParserError> {
+        if dialect_of!(self is SnowflakeDialect | BigQueryDialect) {
```
iffyio: the `dialect_of` macro is discouraged for new code, for dialect-specific functionality we would use a method on the Dialect trait - but I don't think that would work too well in this case, I think we can use the previous pattern of having the dialect call the method instead (i.e the parse_statement approach)

tests/sqlparser_common.rs
```diff
@@ -8593,7 +8593,7 @@ fn lateral_function() {
 fn parse_start_transaction() {
     let dialects = all_dialects_except(|d|
         // BigQuery does not support this syntax
-        d.is::<BigQueryDialect>());
+        d.is::<BigQueryDialect>() || d.is::<SnowflakeDialect>());
```
iffyio, line 8599: the comment above becomes outdated by this change, we should probably adjust it to be generic or mention sf

src/ast/mod.rs
bombsimon: Added same references!

src/ast/mod.rs
bombsimon: Added a description.

Initially I used `Option<Option<Ident>>` because you can either do bare `RAISE` which re-reases the exception currently handled _or_ you can do `RASIE SOME_EXCEPTION` which reaises `SOME_EXCEPTION`.

For BigQuery however this is a bit more complex, there you can something like this which was solved by using the already existing [`parse_raise_stmt`](https://github.com/apache/datafusion-sqlparser-rs/blob/23350bd8d8ee1ff87e2cacb6c9654a9ba1ab76a4/src/parser/mod.rs?plain=1#L786-L800).

```sql
RAISE USING MESSAGE = FORMAT('Some issue: %s', i)
```

src/ast/mod.rs
bombsimon: An exception is represented by an `ExceptionClause` which consists of a `Vec` of `ExceptionWhen` (each `WHEN` arm and its statements) and a potential `RAISE` statement.

For BigQuery, [the docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#raise) says this which I interpreted as it can't be used without `EXCEPTION`:

> **When USING MESSAGE isn't supplied**
> The RAISE statement must only be used within an EXCEPTION clause. The RAISE statement will re-raise the exception that was caught, and preserve the original stack trace.

For Snowflake we could likely do this, but there are other concerns. I know the parser isn't requiring _valid_ syntax, but since this specific `RAISE` is within error handling where you can omit the exception and re-raise what's being handled, I think it makes more sense to put it here. I would thing that if the `RAISE` was on the `StartTransaction` we would always have an exception as described in the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/snowflake-scripting/raise). It's mentioned [here](https://docs.snowflake.com/en/developer-guide/snowflake-scripting/exceptions#raising-the-same-exception-again-in-an-exception-handler).

If I move the `RAISE` to the `StartTransaction`, should I allow it both with and without re-raising and should I move the `Vec<ExceptionWhen>` directly under `StartTransaction` as well?

src/ast/mod.rs
bombsimon: Oh of course, only knew about `display_comma_separated` and didn't know this was a trait. Sorry for not looking further through the code! 🙈

src/parser/mod.rs
bombsimon: I removed the use of `dialect_of!` macro and added it back to `parse_statement` for Snowflake and BigQuery. This means that it will still not work on `GenericDialect` but I assume that is correct?

tests/sqlparser_common.rs
bombsimon: Updated comment with Snowflake and added links to respective docs.

src/ast/mod.rs
```diff
@@ -2982,6 +2982,74 @@ impl From<Set> for Statement {
...
+pub struct ExceptionWhen {
+    pub idents: Vec<Ident>,
+    pub statements: Vec<Statement>,
+}
+
+impl Display for ExceptionWhen {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            " WHEN {idents} THEN",
```
```suggestion
            "WHEN {idents} THEN",
```
display wise its better to have the display impl self contained (the caller should write the leading space when needed), that makes it easier to display the same struct in other contexts

src/ast/mod.rs
```diff
@@ -2982,6 +2982,74 @@ impl From<Set> for Statement {
...
+    ///
+    /// BigQuery: <https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#raise>
+    /// Snowflake: <https://docs.snowflake.com/en/sql-reference/snowflake-scripting/raise>
+    // pub raises: Option<Option<Ident>>,
+    pub raises: Option<Box<Statement>>,
+}
+
+impl Display for ExceptionClause {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, " EXCEPTION")?;
```
```suggestion
        write!(f, "EXCEPTION")?;
```
similar comment re display below

src/ast/mod.rs
```diff
@@ -2982,6 +2982,74 @@ impl From<Set> for Statement {
...
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ExceptionClause {
+    /// When represents each `WHEN` case
+    pub when: Vec<ExceptionWhen>,
+    /// The exception that is being raised or the current exception being handled if `None`.
+    ///
+    /// Example
+    /// RAISE;
+    /// RAISE MY_EXCEPTION;
+    /// RAISE USING MESSAGE = "Some error";
```
iffyio: Can we wrap this in a sql codeblock? Also I think it would be useful to show the example in the context of the exception when (i.e. illustrating the part of the statement does this property map to) vs a standalone statement (which is already documented elsewhere)

src/ast/mod.rs
```diff
@@ -2982,6 +2982,74 @@ impl From<Set> for Statement {
...
+    ///
+    /// BigQuery: <https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#raise>
+    /// Snowflake: <https://docs.snowflake.com/en/sql-reference/snowflake-scripting/raise>
+    // pub raises: Option<Option<Ident>>,
+    pub raises: Option<Box<Statement>>,
```
iffyio: hmm actually would it make more sense to represent this as a `RaiseStmt` instead of an opaque `Statement`?

```suggestion
    pub raise: Option<Box<Statement>>,
```
we only have a single value right or?

src/ast/mod.rs
iffyio: looking at snowflake and bigquery docs there doest seem to be a need for treat raise specially from what i can tell.

snowflake: their example [here](https://docs.snowflake.com/en/developer-guide/snowflake-scripting/exceptions#raising-the-same-exception-again-in-an-exception-handler) shows raise as a regular statement in the statement list.

bigquery: their [spec](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#beginexceptionend) also only mentions a statement list.

Also looking at this [example test case](https://github.com/apache/datafusion-sqlparser-rs/blob/4848e91bf3f9d2d2a1cb53cc898764abfa953d42/tests/sqlparser_snowflake.rs#L4097), its not clear to me where the `RAISE` statement applies to, and formatting wise looks different from whats in the snowflake doc like in the example above

I dont think it matters too much in which context the raise statement applies or the exact behavior of the statement in those scenarios, thats more towards semantics.

So that it seems to me that we should be able to leave as is in the previous behavior of treating `RAISE` as a regular statement. Was there an example sql  that wasn't covered by that approach?

bombsimon: So what's your conclusion around the quote I pasted from [the BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#raise)

> The RAISE statement must only be used within an EXCEPTION clause. The RAISE statement will re-raise the exception that was caught, and preserve the original stack trace.

We just treat that as semantics and put the raise directly on the `StartTransaction` struct together with a `Vec<ExceptionWhen>`?

> So that it seems to me that we should be able to leave as is in the previous behavior of treating RAISE as a regular statement. 

You mean the `RAISE` should be a regular statement in the last `ExceptionWhen`?

bombsimon: I went ahead and dropped the new `ExceptionClasue` which then also includes the custom handling of `RAISE`. `RAISE` will now be a regular statement in the block which I actually THINK is correct given something like this:

```sql
BEGIN
    SELECT 1;
EXCEPTION
    WHEN EXCEPTION_1 THEN
        SELECT 1;
        RAISE SOME_OTHER_EX;
RAISE;
END;
```

With this I moved all the `ExceptionWhen` directly under the `StartTransaction` similar to previous statement list.

alamb:
🚀

PR #1891, opened 2025-06-18
Snowflake: support multiple column options in `CREATE VIEW`

eliaperantoni:
A Snowflake query like:

```sql
CREATE VIEW X (COL WITH TAG (pii='email') COMMENT 'foobar') AS SELECT * FROM Y
```

Would've previously failed because of two reasons:
- It contains two options for the `COL` column, but only one is currently parsed for `CREATE VIEW` statements.
- It uses an option other than `COMMENT`, which is the only one currently supported for the Snowflake dialect.

This PR fixes both.

Additionally, it records in the AST whether the options are comma separated or not. Otherwise, `ToString` wouldn't produce a syntactically valid query for Snowflake.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -28,6 +28,7 @@ use helpers::{
     stmt_data_loading::{FileStagingCommand, StageLoadSelectItemKind},
 };
 
+use core::cmp::Ordering;
 use core::ops::Deref;
 use core::{
     fmt::{self, Display},
@@ -173,7 +174,7 @@ fn format_statement_list(f: &mut fmt::Formatter, statements: &[Statement]) -> fm
 }
 
 /// An identifier, decomposed into its value or character data and the quote style.
-#[derive(Debug, Clone, PartialOrd, Ord)]
+#[derive(Debug, Clone)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct Ident {
@@ -214,6 +215,35 @@ impl core::hash::Hash for Ident {
 }
 
 impl Eq for Ident {}
+
+impl PartialOrd for Ident {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Ident {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let Ident {
+            value,
+            quote_style,
+            // exhaustiveness check; we ignore spans in ordering
+            span: _,
+        } = self;
+
+        let Ident {
+            value: other_value,
+            quote_style: other_quote_style,
+            // exhaustiveness check; we ignore spans in ordering
+            span: _,
+        } = other;
+
+        // First compare by value, then by quote_style
+        value
+            .cmp(other_value)
+            .then_with(|| quote_style.cmp(other_quote_style))
+    }
+}
 
 impl Ident {
     /// Create a new identifier with the given value and no quotes and an empty span.
@@ -4215,7 +4245,7 @@ pub enum Statement {
     /// ```sql
     /// NOTIFY channel [ , payload ]
     /// ```
-    /// send a notification event together with an optional “payload” string to channel
+    /// send a notification event together with an optional "payload" string to channel
     ///
     /// See Postgres <https://www.postgresql.org/docs/current/sql-notify.html>
     NOTIFY {
@@ -9772,6 +9802,8 @@ impl fmt::Display for NullInclusion {
 
 #[cfg(test)]
 mod tests {
+    use crate::tokenizer::Location;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -10589,9 +10580,23 @@ impl<'a> Parser<'a> {
             name,
             data_type,
             options,
+            options_comma_separated: !dialect_of!(self is SnowflakeDialect),
```
iffyio: Instead of introducing a new flag, can we change the datatype of `options` to be an enum that encodes the same information? Something like the following
```rust
enum ColumnOptions {
    CommaSeparated(Vec<ColumnOption>),
    SpaceSeparated(Vec<ColumnOption>)
}

options: Option<ColumnOptions>
```

tests/sqlparser_snowflake.rs
```diff
@@ -4082,3 +4084,16 @@ fn parse_connect_by_root_operator() {
...
+    let create_view_with_tag = r#"CREATE VIEW X (COL WITH TAG (pii='email')) AS SELECT * FROM Y"#;
+    snowflake().verified_stmt(create_view_with_tag);
+}
+
+#[test]
+fn test_snowflake_create_view_with_tag_and_comment() {
+    let create_view_with_tag_and_comment =
+        r#"CREATE VIEW X (COL WITH TAG (pii='email') COMMENT 'foobar') AS SELECT * FROM Y"#;
+    snowflake().verified_stmt(create_view_with_tag_and_comment);
+}
```
iffyio: Can we merge both scenarios into a single test? e.g called `test_snowflake_create_view_options`

eliaperantoni:
Thank you for your feedback @iffyio, I have addressed it and pushed the changes 😊

src/ast/ddl.rs
```diff
@@ -1436,7 +1453,11 @@ impl fmt::Display for ViewColumnDef {
             write!(f, " {}", data_type)?;
         }
         if let Some(options) = self.options.as_ref() {
-            write!(f, " {}", display_comma_separated(options.as_slice()))?;
+            if matches!(options, ColumnOptions::CommaSeparated(_)) {
+                write!(f, " {}", display_comma_separated(options.as_slice()))?;
+            } else {
+                write!(f, " {}", display_separated(options.as_slice(), " "))?
+            }
```
iffyio: Can we make this an exhaustive match? that way if new variants are added the compiler enforces that this code is revisited. e.g.
```rust
match options {
    ColumnOptions::CommaSeparated(_) => {}
    ColumnOptions::SpaceSeparated(_) => {}
}
```

src/ast/spans.rs
```diff
@@ -992,8 +992,11 @@ impl Spanned for ViewColumnDef {
         } = self;
 
         union_spans(
-            core::iter::once(name.span)
-                .chain(options.iter().flat_map(|i| i.iter().map(|k| k.span()))),
+            core::iter::once(name.span).chain(
+                options
+                    .iter()
+                    .flat_map(|i| i.as_slice().iter().map(|k| k.span())),
```
iffyio: Can we `impl Spanned for ColumnOptions` instead? Then here we'll call the `span()` function, and we can drop the `as_slice()` custom function for the struct

src/parser/mod.rs
```diff
@@ -10602,6 +10593,26 @@ impl<'a> Parser<'a> {
...
+            }
+        }
+        Ok(options
+            .is_empty()
+            .not()
+            .then_some(if dialect_of!(self is SnowflakeDialect) {
+                ColumnOptions::SpaceSeparated(options)
+            } else {
+                ColumnOptions::CommaSeparated(options)
+            }))
```
iffyio: Can we write this to not use the `not()` method? I think it currently obfuscates what's going on. Instead we can use an if/else `if !options.is_empty() {} else {}`

src/parser/mod.rs
```diff
@@ -10602,6 +10593,26 @@ impl<'a> Parser<'a> {
...
+            if let Some(option) = option {
+                options.push(option);
+            } else {
+                break;
+            }
+        }
+        Ok(options
+            .is_empty()
+            .not()
+            .then_some(if dialect_of!(self is SnowflakeDialect) {
```
iffyio: Instead of using the `dialect_of` macro, we can add a `if self.dialect.supports_space_separated_column_options()` method. The macro is effectively deprecated and we try to avoid it in new code

tests/sqlparser_common.rs
```diff
@@ -7964,7 +7964,7 @@ fn parse_create_view_with_columns() {
     let sql = "CREATE VIEW v (has, cols) AS SELECT 1, 2";
     // TODO: why does this fail for ClickHouseDialect? (#1449)
     // match all_dialects().verified_stmt(sql) {
-    match all_dialects_except(|d| d.is::<ClickHouseDialect>()).verified_stmt(sql) {
+    match all_dialects_where(|d| !d.is::<ClickHouseDialect>()).verified_stmt(sql) {
```
iffyio: Was there a requirement to change this? on a glance it doesn't look like any functionality is changed and if so we can likely revert this diff

tests/sqlparser_snowflake.rs
```diff
@@ -4165,3 +4167,10 @@ fn test_snowflake_fetch_clause_syntax() {
         canonical,
     );
 }
+
+#[test]
+fn test_snowflake_create_view_with_tag() {
+    let create_view_with_tag =
+        r#"CREATE VIEW X (COL WITH TAG (pii='email') COMMENT 'foobar') AS SELECT * FROM Y"#;
```
iffyio, line 4174: is the intent to verify the behavior of tag or multiple comments support? my understanding is the latter so that we could probably rename the test to match?

tests/sqlparser_common.rs
eliaperantoni: Ah whoops! Not really, it's only a relic of a previous commit which I partially reverted.

eliaperantoni:
Updated @iffyio :) Thanks for your feedback!

PR #1892, opened 2025-06-18
Change tag and policy names to `ObjectName`

eliaperantoni:
In e.g. Snowflake, tags can have qualifying elements in their name:

```sql
CREATE VIEW foo AS (SELECT 1 AS COL WITH TAG foo.bar.baz)
```

But the parser currently expects an `Ident`, which is only one-part.

The same is true for policies names.

This PR fixes both.


eliaperantoni:
It would be nice if #1891 was merged first so that I can write a Snowflake test like the example in this PR's description, since `WITH TAG ...` is not currently supported in `CREATE VIEW` statements for the Snowflake dialect.

src/parser/mod.rs
```diff
@@ -7866,7 +7866,7 @@ impl<'a> Parser<'a> {
     }
 
     pub(crate) fn parse_tag(&mut self) -> Result<Tag, ParserError> {
-        let name = self.parse_identifier()?;
+        let name = self.parse_object_name(false)?;
```
iffyio, line 7887: Could we add a test for this new behavior? one that takes in tag with more than one object name parts

eliaperantoni: Added :)

alamb:
👍

PR #1893, opened 2025-06-18
Fix `impl Ord for Ident`

eliaperantoni:
At the moment, the `Span` of an `Ident` is used influences its ordering, but this doesn't seem consistent with the implementations of `PartialEq`, `Hash`, and the general guidelines for developing with `Span`: that they should be opaque information that doesn't influence the ordering or equality of the AST nodes. Which is the reason why `TokenWithSpan` was born.

This PR change the implementation of `Ord` to ensure that it doesn't depend on the `Span`.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -33,6 +33,7 @@ use core::{
     fmt::{self, Display},
     hash,
 };
+use std::cmp::Ordering;
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


iffyio:
Thanks @eliaperantoni! Could you add some test cases to demonstrate the behavior being fixed? (Also there's some CI failures on the branch that need adressing it seems)

eliaperantoni:
@iffyio Thank you for your feedback! I added a test and fixed the CI issues :)

iffyio (approved):
LGTM! Thanks @eliaperantoni!
cc @alamb

alamb:
🚀

PR #1894, opened 2025-06-19
fix: parse snowflake fetch clause

Vedin:
Snowflake treat `FETCH` in a different way. https://docs.snowflake.com/en/sql-reference/constructs/limit
Most keywords like `FIRST/NEXT`, `ROW/ROWS` and `ONLY` can be skipped.
`PERSENT` and `WITH TIES` should cause an error.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15017,6 +15017,9 @@ impl<'a> Parser<'a> {
 
     /// Parse a FETCH clause
     pub fn parse_fetch(&mut self) -> Result<Fetch, ParserError> {
+        if dialect_of!(self is SnowflakeDialect) {
+            return self.parse_snowflake_fetch();
+        }
         self.expect_one_of_keywords(&[Keyword::FIRST, Keyword::NEXT])?;
         let (quantity, percent) = if self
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -15018,6 +15018,9 @@ impl<'a> Parser<'a> {
 
     /// Parse a FETCH clause
     pub fn parse_fetch(&mut self) -> Result<Fetch, ParserError> {
+        if dialect_of!(self is SnowflakeDialect) {
+            return self.parse_snowflake_fetch();
+        }
```
iffyio: It looks like the snowflake behavior is mostly a subset of the other dialects? If that's the case I think a reasonable approach would be to make the parser logic for the existing `parse_fetch` more permissive and accept the optional components regardless of dialect

Vedin: I made `parse_fetch` more permissive. As I understood, the parser will allow wire syntax and skip some errors in other dialects. Is it a common way to fix such issues? Just want to know it for the future.

iffyio: Yeah in some cases like this the parser may accept syntax that is more permissive than the configured dialect allows, and downstream crates are expected to validate the returned AST if necessary before consuming

PR #1896, opened 2025-06-20
Support for Map values in ClickHouse settings

solontsev:
https://clickhouse.com/docs/operations/settings/settings#additional_table_filters

Current implementation only parses simple settings with key = value format

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2771,7 +2771,7 @@ impl<'a> Parser<'a> {
 
         if self.dialect.supports_dictionary_syntax() {
             self.prev_token(); // Put back the '{'
-            return self.parse_duckdb_struct_literal();
+            return self.parse_duckdb_and_clickhouse_struct_literal();
         }
 
         self.expected("an expression", token)
@@ -3147,7 +3147,7 @@ impl<'a> Parser<'a> {
         Ok(fields)
     }
 
-    /// DuckDB specific: Parse a duckdb [dictionary]
+    /// DuckDB and ClickHouse specific: Parse a duckdb [dictionary] or a clickhouse [map] setting
     ///
     /// Syntax:
     ///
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -3156,18 +3156,21 @@ impl<'a> Parser<'a> {
     /// ```
     ///
     /// [dictionary]: https://duckdb.org/docs/sql/data_types/struct#creating-structs
-    fn parse_duckdb_struct_literal(&mut self) -> Result<Expr, ParserError> {
+    /// [map]: https://clickhouse.com/docs/operations/settings/settings#additional_table_filters
+    fn parse_duckdb_and_clickhouse_struct_literal(&mut self) -> Result<Expr, ParserError> {
```
iffyio: Can we maybe rename this method to `parse_dictionary` and `parse_dictionary_field` respectively? so that the name doesn't change as a result of dialects being extended to support it

src/parser/mod.rs
```diff
@@ -11190,7 +11194,12 @@ impl<'a> Parser<'a> {
             let key_values = self.parse_comma_separated(|p| {
                 let key = p.parse_identifier()?;
                 p.expect_token(&Token::Eq)?;
-                let value = p.parse_value()?.value;
+
+                let value = if p.peek_token_ref().token == Token::LBrace {
+                    p.parse_duckdb_and_clickhouse_struct_literal()?
+                } else {
+                    Expr::Value(p.parse_value()?)
+                };
```
iffyio: I think here it would be fine to parse an arbitrary expression instead, in order to simplify the logic, and likely to have comprehensive support covering other scenarios
```rust
let value = self.parse_expr()
```

src/parser/mod.rs
solontsev: Thanks! I overthought the name for this one

src/test_utils.rs
```diff
@@ -366,6 +366,11 @@ pub fn number(n: &str) -> Value {
     Value::Number(n.parse().unwrap(), false)
 }
 
+/// Creates a `Value::SingleQuotedString`
+pub fn single_quoted_string(s: &str) -> Value {
+    Value::SingleQuotedString(s.to_string())
+}
```
iffyio:
```suggestion
/// Creates a [Value::SingleQuotedString]
pub fn single_quoted_string(s: impl Into<String>) -> Value {
    Value::SingleQuotedString(s.into())
}
```
Thinking we can make this accept the various string types?

iffyio:
@solontsev when you have some time could you merge in the latest from main in order to fix the ci lint failures?

solontsev:
> @solontsev when you have some time could you merge in the latest from main in order to fix the ci lint failures?

Done. Should be OK now

PR #1897, opened 2025-06-20
Fix parsing error when having fields after nested struct in BigQuery

git-hulk:
Before applying this patch, the following SQL cannot be parsed:

```SQL
CREATE TABLE my_table (
  f0 STRING,
  f1 STRUCT<a STRING, b STRUCT<c INT64, d STRING>>,
  f2 STRING,
)
```

But it's a valid SQL in BigQuery. The root cause is having comma after the trailing bracket will be reconigized as a mismatched closing.


git-hulk:
@iffyio Could you please take a look at this fix? The root cause is that it might have more field definitions after the trailing bracket, but now always regarded as an unmatched bracket condition.

The code block is somewhat complex to fix due to its recursive behavior, but I have added more cases for it, and it works well.

tests/sqlparser_bigquery.rs
```diff
@@ -2472,3 +2472,52 @@ fn test_struct_field_options() {
         ")",
     ));
 }
+
+#[test]
+fn test_struct_trailing_bracket() {
```
iffyio: Could we merge this test with [this one](https://github.com/apache/datafusion-sqlparser-rs/blob/b66a3779bb2300384965eaa14a6fdaaf10007555/tests/sqlparser_bigquery.rs#L639-L665)? since they're covering the same behavior. I imagine we could just move the test cases in the latter function into this new function and should be good?

git-hulk: @iffyio Thanks for your suggestion. I didn't notice that test case, and have moved it into a dedicated one.

iffyio (approved):
LGTM! Thanks @git-hulk!
cc @alamb

PR #1900, opened 2025-06-20
Use `IndexColumn` in all index definitions

mvzink:
Index column lists generally allow for more flexibility than just column names: i.e. `ASC`/`DESC` modifiers, Postgres opclasses, MySQL column prefix length, and generic expressions/"functional key parts".

This change uses the existing support for these constructs added in #1707 for `CREATE INDEX` and changes the AST for `ALTER TABLE <table> ADD KEY` (and associated variants), and constraints present in `CREATE TABLE` statements to support these constructs in those location.

Note that, as was already the case with existing `CREATE INDEX` support, there is no special representation for MySQL column prefix length (`INDEX (textcol(10))`), nor do we require the enclosing parentheses for MySQL functional key parts (`INDEX ((col1 + col2))`; for Postgres, this is optional if it doesn't introduce ambiguity). Instead, these are parsed generically as expressions, so the former is parsed as a function call and the latter is wrapped in `Expr::Nested`.

Also note that, as far as I can tell, no dialect supports these more general column expressions in the case of `FOREIGN KEY` constraints, so the parsing and AST for foreign keys are left unchanged.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6868,9 +6868,7 @@ impl<'a> Parser<'a> {
             None
         };
 
-        self.expect_token(&Token::LParen)?;
-        let columns = self.parse_comma_separated(Parser::parse_create_index_expr)?;
-        self.expect_token(&Token::RParen)?;
+        let columns = self.parse_parenthesized_index_column_list()?;
 
         let include = if self.parse_keyword(Keyword::INCLUDE) {
             self.expect_token(&Token::LParen)?;
@@ -8070,7 +8068,7 @@ impl<'a> Parser<'a> {
                 let index_name = self.parse_optional_ident()?;
                 let index_type = self.parse_optional_using_then_index_type()?;
 
-                let columns = self.parse_parenthesized_column_list(Mandatory, false)?;
+                let columns = self.parse_parenthesized_index_column_list()?;
                 let index_options = self.parse_index_options()?;
                 let characteristics = self.parse_constraint_characteristics()?;
                 Ok(Some(TableConstraint::Unique {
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -10595,6 +10593,16 @@ impl<'a> Parser<'a> {
         self.parse_parenthesized_column_list_inner(optional, allow_empty, |p| p.parse_identifier())
     }
 
+    /// Parses a parenthesized comma-separated list of index columns, which can be arbitrary
+    /// expressions with ordering information (and an opclass in some dialects).
+    pub fn parse_parenthesized_index_column_list(
```
```suggestion
    fn parse_parenthesized_index_column_list(
```
We can have this within the crate unless it would be required/useful to export?

src/parser/mod.rs
```diff
@@ -16476,6 +16484,22 @@ mod tests {
...
+                        options: OrderByOptions {
+                            asc: None,
+                            nulls_first: None,
+                        },
+                        with_fill: None,
+                    },
+                    operator_class: None,
+                }
+            };
+        }
```
iffyio: Would it make more sense to use a function for this instead? since macros aren't the best for readability

src/test_utils.rs
```diff
@@ -448,3 +448,47 @@ pub fn call(function: &str, args: impl IntoIterator<Item = Expr>) -> Expr {
         within_group: vec![],
     })
 }
+
+/// Gets the first index column (mysql calls it a key part) of the first index a CREATE INDEX,
+/// CREATE TABLE, or ALTER TABLE statement.
```
iffyio:
```suggestion
/// Gets the first index column (mysql calls it a key part) of a [Statement::CreateIndex], [Statement::CreateTable] or [Statement::AlterTAble]
```

mvzink:
Having started using this, I noticed that it's actually fairly inconvenient to have the `TableConstraint::ForeignKey` variant have a different type for columns (e.g. if you want to implement `get_columns(constraint: &TableConstraint) -> &[???]`). Thoughts on using `IndexColumn` for `TableConstraint::ForeignKey` as well, even though it should aways only be plain column names?

iffyio:
Ah yeah that's not ideal when consuming the AST, but I think we could leave the foreign key representation as `Ident` since it maps to what's expected at that level - it would be one less thing for downstream crates to validate in the AST

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1901, opened 2025-06-22
Support procedure argmode

ZacJW:
Currently there's no support for argmodes (`IN`, `OUT`, `INOUT`) in `CREATE PROCEDURE` statements despite the standard (I'm reading a [BNF for ISO/IEC 9075-2:2003](https://ronsavage.github.io/SQL/sql-2003-2.bnf.html#parameter%20mode) since I don't have anything newer) and 'RDBMS's like MySQL and PostgreSQL supporting them.

This PR adds support for the standard modes (PostgreSQL's `VARIADIC` mode is not supported).

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -30,7 +30,11 @@ use sqlparser_derive::{Visit, VisitMut};
 
 use crate::ast::value::escape_single_quote_string;
 use crate::ast::{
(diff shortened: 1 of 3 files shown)


ZacJW:
Just found the link to <https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html> in the README, and this implementation is still valid.

Issue #1904, opened 2025-06-24
Wrong join precedence parsing for non-Snowflake dialects (nested joins parsed incorrectly)


Dimchikkk:
```toml
[package]
name = "sqlparsertest"
version = "0.1.0"
edition = "2024"

[dependencies]
sqlparser055 = { package = "sqlparser", version = "0.55.0" }
sqlparser056 = { package = "sqlparser", version = "0.56.0" }
```
```rs
fn main() {
    let sql = r#"SELECT *
FROM t1
NATURAL JOIN t5
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0
WHERE t0.v1 = t1.v0;"#;

    println!("--- SQL ---\n{}\n", sql);

    {
        use sqlparser055 as sqlparser;
        use sqlparser::dialect::GenericDialect;
        use sqlparser::parser::Parser;

        println!("--- sqlparser 0.55 ---");
        match Parser::parse_sql(&GenericDialect {}, sql) {
            Ok(statements) => {
                for stmt in &statements {
                    println!("Full statement (SQL): {}", stmt); // whole query as SQL string
                }
                for stmt in statements {
                    if let sqlparser::ast::Statement::Query(query) = stmt {
                        if let sqlparser::ast::SetExpr::Select(select) = &*query.body {
                            for tbl in &select.from {
                                println!("joins (debug): {:?}", tbl.joins);
                                println!("joins (SQL): {}", tbl); // print TableWithJoins as SQL string
                            }
                        }
                    }
                }
            }
            Err(err) => eprintln!("Error: {}", err),
        }
    }

    {
        use sqlparser056 as sqlparser;
        use sqlparser::dialect::GenericDialect;
        use sqlparser::parser::Parser;

        println!("\n--- sqlparser 0.56 ---");
        match Parser::parse_sql(&GenericDialect {}, sql) {
            Ok(statements) => {
                for stmt in &statements {
                    println!("Full statement (SQL): {}", stmt);
                }
                for stmt in statements {
                    if let sqlparser::ast::Statement::Query(query) = stmt {
                        if let sqlparser::ast::SetExpr::Select(select) = &*query.body {
                            for tbl in &select.from {
                                println!("joins (debug): {:?}", tbl.joins);
                                println!("joins (SQL): {}", tbl);
                            }
                        }
                    }
                }
            }
            Err(err) => eprintln!("Error: {}", err),
        }
    }
}
```

[...]

PR #1905, opened 2025-06-24
Fix join precedence for non-snowflake queries

Dimchikkk:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1904.

The bug was introduced in https://github.com/apache/datafusion-sqlparser-rs/pull/1799.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -12467,7 +12467,7 @@ impl<'a> Parser<'a> {
                 };
                 let mut relation = self.parse_table_factor()?;
 
-                if self.peek_parens_less_nested_join() {
+                if dialect_of!(self is SnowflakeDialect) && self.peek_parens_less_nested_join() {
                     let joins = self.parse_joins()?;
                     relation = TableFactor::NestedJoin {
                         table_with_joins: Box::new(TableWithJoins { relation, joins }),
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -12486,7 +12486,7 @@ impl<'a> Parser<'a> {
                 };
                 let mut relation = self.parse_table_factor()?;
 
-                if self.peek_parens_less_nested_join() {
+                if dialect_of!(self is SnowflakeDialect) && self.peek_parens_less_nested_join() {
```
iffyio: Can we change this to a dialect method that is only set to true for snowflake? e.g. `if self.dialect.supports_parens_less_nested_join() && self.peek_parens_less_nested_join()`

tests/sqlparser_common.rs
```diff
@@ -15357,6 +15357,28 @@ fn check_enforced() {
     );
 }
 
+#[test]
+fn join_precedence() {
+    all_dialects_except(|d| d.is::<SnowflakeDialect>()).verified_query_with_canonical(
```
iffyio: By switching to the dialect method, we would be able to change these to use `all_dialects_where(|d|d.supports_parens_less_nested_join())`

iffyio (commented):
Thanks for the fix @Dimchikkk! Left a couple comments

Dimchikkk:
Thanks for the review, @iffyio - I've addressed your feedback and it’s ready for another round when you are.

src/dialect/mod.rs
```diff
@@ -278,6 +278,34 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Indicates whether the dialect supports left-associative join parsing
```
iffyio, line 281: Very nice thanks for the description!

iffyio (approved):
LGTM! Thanks again @Dimchikkk!
cc @alamb

PR #1908, opened 2025-06-25
Postgres constraint NOT VALID and VALIDATE CONSTRAINT

### LINKED ISSUE #1907: Postgres NOT VALID and VALIDATE CONSTRAINT not parsed for ALTER TABLE
Doesn't recognize statements creating an unvalidated constraint (`ALTER TABLE "xyz" ADD "constraint_name" NOT VALID`)

And also doesn't recognize statements validating constraints (`ALTER TABLE xyz VALIDATE CONSTRAINT "constraint_name"`)

Docs for both are under ALTER TABLE: https://www.postgresql.org/docs/current/sql-altertable.html

achristmascarl:
closes #1907 

Handles creating constraints with `NOT VALID`, as well as `ALTER TABLE table VALIDATE CONSTRAINT constraint`

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8477,7 +8477,11 @@ impl<'a> Parser<'a> {
     pub fn parse_alter_table_operation(&mut self) -> Result<AlterTableOperation, ParserError> {
         let operation = if self.parse_keyword(Keyword::ADD) {
             if let Some(constraint) = self.parse_optional_table_constraint()? {
-                AlterTableOperation::AddConstraint(constraint)
+                let not_valid = self.parse_keywords(&[Keyword::NOT, Keyword::VALID]);
+                AlterTableOperation::AddConstraint {
+                    constraint,
+                    not_valid,
+                }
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -8477,7 +8477,14 @@ impl<'a> Parser<'a> {
     pub fn parse_alter_table_operation(&mut self) -> Result<AlterTableOperation, ParserError> {
         let operation = if self.parse_keyword(Keyword::ADD) {
             if let Some(constraint) = self.parse_optional_table_constraint()? {
-                AlterTableOperation::AddConstraint(constraint)
+                let mut not_valid = false;
```
iffyio: For both this flag and the new validate constraint clause, can we can skip the dialect methods and have the parser always accept the syntax if it shows up in a statement? since the syntax doesn't look like it conflicts with other dialects

achristmascarl: @iffyio sounds good, just pushed the update

PR #1911, opened 2025-06-29
Make `GenericDialect` support from-first syntax

simonvandel:
The docs for `GenericDialect` says that is can be permissive, so I thought it could support from-first syntax.

This would help a bit on the "friendly sql" issue https://github.com/apache/datafusion/issues/14514


alamb:
The code keeps on rolling

PR #1912, opened 2025-06-29
Redshift alter column type no set

yoavcloud:
Expanded the support for `ALTER TABLE tbl ALTER COLUMN col TYPE <type>` to the Redshift dialect in addition to PostgreSQL. Generalized this option using the `supports_xxx` functions in the Dialect trait to maintain backwards compatibility with expected error messages in non-supporting dialects.

Improve SQL output from Ast by introducing an option to determine if `SET DATA` appears before the `TYPE` keyword.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -893,7 +893,10 @@ pub enum AlterColumnOperation {
         data_type: DataType,
         /// PostgreSQL specific
         using: Option<Expr>,
+        /// Whether the statement included the optional `SET DATA` keywords
+        had_set: bool,
     },
+
     /// `ADD GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ]`
     ///
     /// Note: this is a PostgreSQL-specific operation.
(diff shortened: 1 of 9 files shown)


src/dialect/mod.rs
```diff
@@ -1060,6 +1060,22 @@ pub trait Dialect: Debug + Any {
...
+    /// Returns true if the dialect supports `ALTER TABLE tbl ALTER COLUMN col TYPE <type>`
+    /// without specifying `SET DATA` before `TYPE`.
+    ///
+    /// - [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html#r_ALTER_TABLE-synopsis)
+    /// - [PostgreSQL](https://www.postgresql.org/docs/current/sql-altertable.html)
+    fn supports_alter_column_type_without_set(&self) -> bool {
+        false
+    }
+
+    /// Returns true if the dialect supports `ALTER TABLE tbl ALTER COLUMN col SET DATA TYPE <type> USING <exp>`
```
```suggestion
    /// Returns true if the dialect supports the `USING` clause in an `ALTER COLUMN` statement.
    /// Example:
    ///  ```sql
    ///  ALTER TABLE tbl ALTER COLUMN col SET DATA TYPE <type> USING <exp>`
    /// ```
```
thinking something like this so that its a bit clearer which part of the statement the flag refers to?

src/dialect/mod.rs
```diff
@@ -1060,6 +1060,22 @@ pub trait Dialect: Debug + Any {
     fn supports_space_separated_column_options(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports `ALTER TABLE tbl ALTER COLUMN col TYPE <type>`
+    /// without specifying `SET DATA` before `TYPE`.
+    ///
+    /// - [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html#r_ALTER_TABLE-synopsis)
+    /// - [PostgreSQL](https://www.postgresql.org/docs/current/sql-altertable.html)
+    fn supports_alter_column_type_without_set(&self) -> bool {
```
iffyio: not sure I followed that this flag is necessary, it looks like we would be able to accept either variant with or without the `SET DATA` prefix? The error message of existing tests if I understood the PR description correctly we can change to match the new behavior I think

src/ast/ddl.rs
```diff
@@ -893,7 +893,10 @@ pub enum AlterColumnOperation {
         data_type: DataType,
         /// PostgreSQL specific
         using: Option<Expr>,
+        /// Whether the statement included the optional `SET DATA` keywords
+        had_set: bool,
```
```suggestion
        /// Set to true if the statement includes the `SET DATA TYPE` keywords
        set_data_type: bool,
```

src/dialect/mod.rs
yoavcloud: Yes that is an option. I hesitated a bit because saw specific negative tests, and did not want to break anything.

PR #1913, opened 2025-06-29
Add support for NULL escape char in pattern match searches

yoavcloud:
It turns out that both in Redshift and Snowflake, you can pass a `NULL` escape char for `ILIKE`, `LIKE` and `SIMILAR TO`. This PR expands the Ast to support that.


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1914, opened 2025-06-30
Support for Postgres `CREATE SERVER`

### LINKED ISSUE #1814: Postgres CREATE SERVER can't be parsed
Postgres's `CREATE SERVER` fails to be parsed with the following error:
```rust
ParserError("Expected: an object type after CREATE, found: SERVER at Line: 1, Column: 8")
```

Example code that I'm running:
```sql
CREATE SERVER foo
FOREIGN DATA WRAPPER postgres_fdw
OPTIONS (
  host 'db.bar.com'
  dbname 'postgres'
  port '5432'
)
```

docs: https://www.postgresql.org/docs/current/sql-createserver.html

solontsev:
Closes #1814

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -996,7 +996,7 @@ pub enum Expr {
         data_type: DataType,
         /// The value of the constant.
         /// Hint: you can unwrap the string value using `value.into_string()`.
-        value: Value,
+        value: ValueWithSpan,
     },
     /// Scalar function call e.g. `LEFT(foo, 5)`
     Function(Function),
@@ -1124,6 +1124,8 @@ pub enum Expr {
     /// [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-lambda-functions.html)
     /// [DuckDb](https://duckdb.org/docs/sql/functions/lambda.html)
     Lambda(LambdaFunction),
+    /// Checks membership of a value in a JSON array
+    MemberOf(MemberOf),
 }
 
 impl Expr {
@@ -1912,6 +1914,7 @@ impl fmt::Display for Expr {
             }
             Expr::Prior(expr) => write!(f, "PRIOR {expr}"),
             Expr::Lambda(lambda) => write!(f, "{lambda}"),
+            Expr::MemberOf(member_of) => write!(f, "{member_of}"),
         }
     }
 }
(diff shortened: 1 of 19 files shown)


src/parser/mod.rs
```diff
@@ -4662,6 +4662,10 @@ impl<'a> Parser<'a> {
             self.parse_create_procedure(or_alter)
         } else if self.parse_keyword(Keyword::CONNECTOR) {
             self.parse_create_connector()
+        } else if self.parse_keyword(Keyword::SERVER)
+            && dialect_of!(self is PostgreSqlDialect | GenericDialect)
```
```suggestion
```
I think it should be fine to let the parser be permissive and accept this statement whenever it shows up

src/ast/mod.rs
```diff
@@ -7973,6 +8015,20 @@ impl fmt::Display for SecretOption {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ServerOption {
```
```suggestion
pub struct CreateServerOption {
```

src/ast/mod.rs
```diff
@@ -3316,6 +3316,18 @@ pub enum Statement {
         options: Vec<SecretOption>,
     },
     /// ```sql
+    /// CREATE SERVER
+    /// ```
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-createserver.html)
+    CreateServer {
```
iffyio: For new statements [we're moving away from using anonymous structs](https://github.com/apache/datafusion-sqlparser-rs/issues/1204), here can we introduce a dedicated struct for this statement?
```rust
CreateServer(CreateServerStatement)
```
See [Statement::Raise](https://github.com/apache/datafusion-sqlparser-rs/blob/9d68a5663dcdb0357ee52bc351dd99a5905fbec5/src/ast/mod.rs#L3145-L3146) as an example

src/ast/mod.rs
```diff
@@ -3316,6 +3316,18 @@ pub enum Statement {
...
+        name: ObjectName,
+        if_not_exists: bool,
+        server_type: Option<Ident>,
+        version: Option<Ident>,
+        fdw_name: ObjectName,
```
iffyio: Can we rename this to something more descriptive e.g. explicitly `foreign_data_wrapper`?

src/parser/mod.rs
```diff
@@ -15806,6 +15810,53 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// ```sql
+    ///     CREATE SERVER [ IF NOT EXISTS ] server_name [ TYPE 'server_type' ] [ VERSION 'server_version' ]
+    ///     FOREIGN DATA WRAPPER fdw_name
+    ///     [ OPTIONS ( option 'value' [, ... ] ) ]
+    /// ```
+    ///
+    /// [PostgreSQL Documentation](https://www.postgresql.org/docs/current/sql-createserver.html)
+    pub fn parse_pg_create_server(&mut self) -> Result<Statement, ParserError> {
```
```suggestion
    ///   Parse a `CREATE SERVER` statement.
    ///
    ///  See [Statement::CreateServer]
    pub fn parse_pg_create_server(&mut self) -> Result<Statement, ParserError> {
```
I think we can do something like this to avoid duplicating the comments. The syntax we can skip entirely since that should be covered by the doc link if the reader requires further context

src/ast/mod.rs
solontsev: Thanks for the example! All done, ready to review

solontsev:
Thanks @iffyio! Done

iffyio (approved):
LGTM! Thanks @solontsev!
cc @alamb

alamb:
THank you @iffyio  for being a coding machine -- it is pretty amazing to see all this code go in -- its like we don't really have any idea how many crazy variants of SQL there are until you see the steady stream of PRs

PR #1916, opened 2025-06-30
fix: DuckDB accepts 2nd characters argument to TRIM

ryanschneider:
Like BigQuery and Snowflake, DuckDB also supports the 2nd `characters` argument to `TRIM`: https://duckdb.org/docs/stable/sql/functions/text.html#trimstring-characters

- DuckDB was added to the list of dialects potentially expecting a comma
- `test_duckdb_trim` is basically a copy of `test_bigquery_trim` using the proper dialect
- `parse_trim` test was updated to exclude DuckDB alongside BigQuery and Snowflake.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -7762,7 +7762,7 @@ fn parse_trim() {
         Box::new(MySqlDialect {}),
         //Box::new(BigQueryDialect {}),
         Box::new(SQLiteDialect {}),
-        Box::new(DuckDbDialect {}),
+        //Box::new(DuckDbDialect {}),
     ]);
 
     assert_eq!(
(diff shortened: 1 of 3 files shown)


tests/sqlparser_common.rs
```diff
@@ -7762,7 +7762,7 @@ fn parse_trim() {
         Box::new(MySqlDialect {}),
         //Box::new(BigQueryDialect {}),
         Box::new(SQLiteDialect {}),
-        Box::new(DuckDbDialect {}),
+        //Box::new(DuckDbDialect {}),
```
```suggestion
```
I think it should be fine to remove the dialect from the list

ryanschneider: I commented it out to match the existing code style where BigQuery and Snowflake were commented out as well.

PR #1917, opened 2025-07-01
Add support for MySQL MEMBER OF

yoavcloud:
https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#operator_member-of


src/ast/mod.rs
```diff
@@ -1124,6 +1124,14 @@ pub enum Expr {
...
     /// [DuckDb](https://duckdb.org/docs/sql/functions/lambda.html)
     Lambda(LambdaFunction),
+    /// Checks membership of a value in a JSON array
+    ///
+    /// Syntax:
+    /// ```sql
+    /// <value> MEMBER OF(<array>)
+    /// ```
+    /// [MySQL](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#operator_member-of)
+    MemberOf(Box<Expr>, Box<Expr>),
```
iffyio: Can we introduce a struct for the expression e.g. `MemberOf(MemberOf{ value, json })`? That way it would be less breaking if spans or other options are added to the expression later on

tests/sqlparser_mysql.rs
```diff
@@ -4109,3 +4109,9 @@ fn parse_alter_table_drop_index() {
         AlterTableOperation::DropIndex { name } if name.value == "idx_index"
     );
 }
+
+#[test]
+fn parse_json_member_of() {
+    mysql().verified_stmt(r#"SELECT 17 MEMBER OF('[23, "abc", 17, "ab", 10]')"#);
```
iffyio, line 4115: For one of the tests can we assert the reurned expr that it contains the expected values in the expected fields in the AST?

src/parser/mod.rs
```diff
@@ -3609,6 +3609,16 @@ impl<'a> Parser<'a> {
                         self.expected("IN or BETWEEN after NOT", self.peek_token())
                     }
                 }
+                Keyword::MEMBER => {
+                    if self.parse_keyword(Keyword::OF) {
+                        let _ = self.expect_token(&Token::LParen);
+                        let expr2 = self.parse_expr()?;
+                        let _ = self.expect_token(&Token::RParen);
```
iffyio: It looks like we're dropping potential error returned by the expect_token calls (we should probably add a test case for this behavior)?

yoavcloud: Updated, I'm just a bit rusty :-) ...

PR #1919, opened 2025-07-02
Add span for `Expr::TypedString`

feral-dot-io:
I have need to get the location info of an `Expr::TypedString` but this currently returns a `Value` rather than a `ValueWithSpan`. I saw that the parser already has access to this information. So this is a small PR changes Expr::TypedString to use a ValueWithSpan.

I have also updated the test cases and docs.

One issue with this PR is that the I've used `Span::empty()` in the test cases and it still passed. Looking at a PG test more closely the span was being filled with non-zero locations. Checking another PG test it seemed that span tests were skipped by using pattern matching; I didn't want to do this because I wanted to avoid accidentally change the tests. Or it might be fine as it is. Let me know how you'd like to deal with this.

### DIFF
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -1410,7 +1410,6 @@ impl Spanned for AssignmentTarget {
 /// f.e. `IS NULL <expr>` reports as `<expr>::span`.
 ///
 /// Missing spans:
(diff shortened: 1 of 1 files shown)


iffyio (approved):
LGTM! Thanks @feral-dot-io!
cc @alamb

PR #1921, opened 2025-07-02
Make `GenericDialect` support trailing commas in projections

simonvandel:
Similar to https://github.com/apache/datafusion-sqlparser-rs/pull/1911.

The docs for GenericDialect says that is can be permissive, so I thought it could support trailing commas in projections.

This would help a bit on the "friendly sql" issue
https://github.com/apache/datafusion/issues/14514

I removed a test-case that tested that ´GenericDialect` did not support trailing commas in projections. Positive tests for trailing commas still exists in the `parse_projection_trailing_comma` test

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -11125,8 +11125,15 @@ fn parse_trailing_comma() {
     );
     trailing_commas.verified_stmt(r#"SELECT "from" FROM "from""#);
 
-    // doesn't allow all trailing commas
-    let trailing_commas = TestedDialects::new(vec![Box::new(GenericDialect {})]);
+    // doesn't allow any trailing commas
+    let trailing_commas = TestedDialects::new(vec![Box::new(PostgreSqlDialect {})]);
+
+    assert_eq!(
+        trailing_commas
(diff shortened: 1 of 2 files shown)


tests/sqlparser_common.rs
```diff
@@ -11125,16 +11125,9 @@ fn parse_trailing_comma() {
     );
     trailing_commas.verified_stmt(r#"SELECT "from" FROM "from""#);
 
-    // doesn't allow any trailing commas
+    // doesn't allow all trailing commas
     let trailing_commas = TestedDialects::new(vec![Box::new(GenericDialect {})]);
 
-    assert_eq!(
```
iffyio, line 11131: could we instead change this to use a dialect like postgres that doesnt support trailing commas? thinking that would let us keep the negative test around.

simonvandel: Of course, good idea. Implemented in https://github.com/apache/datafusion-sqlparser-rs/pull/1921/commits/83812914bd4f35c6d98b8d9af2ba5d24693492f3

alamb:
Thank you -- I think this will be really nice improvement to SQL in DataFusion by default

PR #1924, opened 2025-07-04
Align Snowflake dialect to new test of reserved keywords

yoavcloud:
We've encountered a problem parsing statements like `SELECT 1, sort FROM tbl` which are valid in Snowflake.

The reason is that in the Snowflake dialect, supports_projection_trailing_commas is `true`. The flow then checks the keyword after the comma to see if it's reserved (`FROM` for example), and if not, it stops parsing the projection. However, the check relies on `Dialect::is_column_alias` which wasn't implemented in the Snowflake dialog. 

The solution is to align the behavior of the Snowflake dialect to other dialects by overriding `Dialect::is_column_alias` instead of `Dialect::is_select_item_alias`.

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -301,9 +301,8 @@ impl Dialect for SnowflakeDialect {
         true
     }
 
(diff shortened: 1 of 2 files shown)


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1927, opened 2025-07-05
Add support for `NOT NULL` and `NOTNULL` expressions

### LINKED ISSUE #1920: Parser: Cannot parse `COUNT(CASE WHEN x NOT NULL THEN 1 END)`
I'm not sure if this is generally accepted SQL syntax but duckdb supports it:

```
> duckdb
DuckDB v1.3.0 (Ossivalis) 71c5c07cdd
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
> WITH t AS (SELECT NULL as x) SELECT COUNT(CASE WHEN x NOT NULL THEN 1 END) AS x FROM t;
┌───────┐
│   x   │
│ int64 │
├───────┤
│   0   │
└───────┘
>
```

However the parser rejects it:

```rust
#[test]
fn test_duckdb_case() {
    let real_sql = r#"WITH t AS (SELECT NULL as x) SELECT COUNT(CASE WHEN x NOT NULL THEN 1 END) AS x FROM t"#;
    assert_eq!(duckdb().verified_stmt(real_sql).to_string(), real_sql);
}
```

This fails w/ `WITH t AS (SELECT NULL as x) SELECT COUNT(CASE WHEN x NOT NULL THEN 1 END) AS x FROM t: ParserError("Expected: ), found: WHEN")`

I've tried stepping through myself and got as far as looking at `parse_case_expr` and it's failing here:

```rust
        llet mut conditions = vec![];
        loop {
            let condition = self.parse_expr()?;
            self.expect_keyword_is(Keyword::THEN)?;   <-- fails here
            let result = self.parse_expr()?;
            conditions.push(CaseWhen { condition, result });
            if !self.parse_keyword(Keyword::WHEN) {
                break;
            }
        }
```

At that point, condition is just the Expr `x` and conditions is still empty, it still needs to parse `NOT NULL` before expecting the `THEN` keyword, but I'm not sure what the fix is.

ryanschneider:
Fixes #1920 by adding `NOT NULL` generally as an alias for `IS NOT NULL`, (only in `ParserState::Normal`) and adds the non-standard `NOTNULL` keyword supported by DuckDB, SQLite, and Postgres.  It also adds a new `ParserState::ColumnDefinition` to avoid incorrectly parsing `NOT NULL` as `IS NOT NULL` in column definitions.

Since this is my first non-trivial PR please provide any feedback!  I confirmed all tests pass w/ `cargo test` and `cargo test --all-features` but if there's anything else I need to do please let me know, thanks!

### DIFF
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::dialect::Dialect;
+use crate::dialect::{Dialect, IsNotNullAlias};
 
 /// A [`Dialect`] for [DuckDB](https://duckdb.org/)
 #[derive(Debug, Default)]
@@ -95,11 +95,12 @@ impl Dialect for DuckDbDialect {
         true
     }
 
-    fn supports_not_null(&self) -> bool {
-        true
-    }
-
-    fn supports_notnull(&self) -> bool {
-        true
+    /// DuckDB supports `NOT NULL` and `NOTNULL` as aliases
+    /// for `IS NOT NULL`, see DuckDB Comparisons <https://duckdb.org/docs/stable/sql/expressions/comparison_operators#between-and-is-not-null>
+    fn supports_is_not_null_alias(&self, alias: IsNotNullAlias) -> bool {
+        match alias {
+            IsNotNullAlias::NotNull => true,
+            IsNotNullAlias::NotSpaceNull => true,
+        }
     }
 }
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -94,4 +94,12 @@ impl Dialect for DuckDbDialect {
     fn supports_order_by_all(&self) -> bool {
         true
     }
+
+    fn supports_not_null(&self) -> bool {
+        true
+    }
+
+    fn supports_notnull(&self) -> bool {
+        true
+    }
 }
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::dialect::{Dialect, IsNotNullAlias};
+use crate::dialect::Dialect;
 
 /// A [`Dialect`] for [DuckDB](https://duckdb.org/)
 #[derive(Debug, Default)]
@@ -95,12 +95,9 @@ impl Dialect for DuckDbDialect {
         true
     }
 
-    /// DuckDB supports `NOT NULL` and `NOTNULL` as aliases
-    /// for `IS NOT NULL`, see DuckDB Comparisons <https://duckdb.org/docs/stable/sql/expressions/comparison_operators#between-and-is-not-null>
-    fn supports_is_not_null_alias(&self, alias: IsNotNullAlias) -> bool {
-        match alias {
-            IsNotNullAlias::NotNull => true,
-            IsNotNullAlias::NotSpaceNull => true,
-        }
+    /// DuckDB supports `NOTNULL` as an alias for `IS NOT NULL`,
+    /// see DuckDB Comparisons <https://duckdb.org/docs/stable/sql/expressions/comparison_operators#between-and-is-not-null>
+    fn supports_notnull_operator(&self) -> bool {
+        true
     }
 }
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -98,7 +98,7 @@ impl Dialect for DuckDbDialect {
     fn supports_select_wildcard_exclude(&self) -> bool {
         true
     }
-  
+
     /// DuckDB supports `NOTNULL` as an alias for `IS NOT NULL`,
     /// see DuckDB Comparisons <https://duckdb.org/docs/stable/sql/expressions/comparison_operators#between-and-is-not-null>
     fn supports_notnull_operator(&self) -> bool {
--- a/src/dialect/duckdb.rs
+++ b/src/dialect/duckdb.rs
@@ -95,6 +95,10 @@ impl Dialect for DuckDbDialect {
         true
     }
 
+    fn supports_select_wildcard_exclude(&self) -> bool {
+        true
+    }
+  
     /// DuckDB supports `NOTNULL` as an alias for `IS NOT NULL`,
     /// see DuckDB Comparisons <https://duckdb.org/docs/stable/sql/expressions/comparison_operators#between-and-is-not-null>
     fn supports_notnull_operator(&self) -> bool {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -590,6 +590,13 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returne true if the dialect supports specifying multiple options
+    /// in a `CREATE TABLE` statement for the structure of the new table. For example:
+    /// `CREATE TABLE t (a INT, b INT) AS SELECT 1 AS b, 2 AS a`
+    fn supports_create_table_multi_schema_info_sources(&self) -> bool {
+        false
+    }
+
     /// Dialect-specific infix parser override
     ///
     /// This method is called to parse the next infix expression.
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -55,6 +55,7 @@ use crate::keywords::Keyword;
 use crate::parser::{Parser, ParserError};
 use crate::tokenizer::Token;
 
+use crate::dialect::IsNotNullAlias::{NotNull, NotSpaceNull};
 #[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
 
@@ -650,12 +651,17 @@ pub trait Dialect: Debug + Any {
                 Token::Word(w) if w.keyword == Keyword::MATCH => Ok(p!(Like)),
                 Token::Word(w) if w.keyword == Keyword::SIMILAR => Ok(p!(Like)),
                 Token::Word(w) if w.keyword == Keyword::MEMBER => Ok(p!(Like)),
-                Token::Word(w) if w.keyword == Keyword::NULL && self.supports_not_null() => {
+                Token::Word(w)
+                    if w.keyword == Keyword::NULL
+                        && self.supports_is_not_null_alias(NotSpaceNull) =>
+                {
                     Ok(p!(Is))
                 }
                 _ => Ok(self.prec_unknown()),
             },
-            Token::Word(w) if w.keyword == Keyword::NOTNULL && self.supports_notnull() => {
+            Token::Word(w)
+                if w.keyword == Keyword::NOTNULL && self.supports_is_not_null_alias(NotNull) =>
+            {
                 Ok(p!(Is))
             }
             Token::Word(w) if w.keyword == Keyword::IS => Ok(p!(Is)),
@@ -1096,14 +1102,13 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if the dialect supports `NOTNULL` in expressions.
-    fn supports_notnull(&self) -> bool {
-        false
-    }
-
-    /// Returns true if the dialect supports `NOT NULL` in expressions.
-    fn supports_not_null(&self) -> bool {
-        false
+    /// Returns true if the dialect supports the passed in alias.
+    /// See [IsNotNullAlias].
+    fn supports_is_not_null_alias(&self, alias: IsNotNullAlias) -> bool {
+        match alias {
+            NotNull => false,
+            NotSpaceNull => false,
+        }
     }
 }
 
@@ -1129,6 +1134,13 @@ pub enum Precedence {
     UnaryNot,
     And,
     Or,
+}
+
+/// Possible aliases for `IS NOT NULL` supported
+/// by some non-standard dialects.
+pub enum IsNotNullAlias {
+    NotNull,
+    NotSpaceNull,
 }
 
 impl dyn Dialect {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -55,7 +55,6 @@ use crate::keywords::Keyword;
 use crate::parser::{Parser, ParserError};
 use crate::tokenizer::Token;
 
-use crate::dialect::IsNotNullAlias::{NotNull, NotSpaceNull};
 #[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
 
@@ -651,17 +650,9 @@ pub trait Dialect: Debug + Any {
                 Token::Word(w) if w.keyword == Keyword::MATCH => Ok(p!(Like)),
                 Token::Word(w) if w.keyword == Keyword::SIMILAR => Ok(p!(Like)),
                 Token::Word(w) if w.keyword == Keyword::MEMBER => Ok(p!(Like)),
-                Token::Word(w)
-                    if w.keyword == Keyword::NULL
-                        && self.supports_is_not_null_alias(NotSpaceNull) =>
-                {
-                    Ok(p!(Is))
-                }
                 _ => Ok(self.prec_unknown()),
             },
-            Token::Word(w)
-                if w.keyword == Keyword::NOTNULL && self.supports_is_not_null_alias(NotNull) =>
-            {
+            Token::Word(w) if w.keyword == Keyword::NOTNULL && self.supports_notnull_operator() => {
                 Ok(p!(Is))
             }
             Token::Word(w) if w.keyword == Keyword::IS => Ok(p!(Is)),
@@ -1102,13 +1093,10 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if the dialect supports the passed in alias.
-    /// See [IsNotNullAlias].
-    fn supports_is_not_null_alias(&self, alias: IsNotNullAlias) -> bool {
-        match alias {
-            NotNull => false,
-            NotSpaceNull => false,
-        }
+    /// Returns true if the dialect supports the `x NOTNULL`
+    /// operator expression.
+    fn supports_notnull_operator(&self) -> bool {
+        false
     }
 }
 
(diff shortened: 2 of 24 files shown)


src/dialect/duckdb.rs
```diff
@@ -94,4 +94,12 @@ impl Dialect for DuckDbDialect {
...
         true
     }
+
+    fn supports_not_null(&self) -> bool {
+        true
+    }
+
+    fn supports_notnull(&self) -> bool {
+        true
+    }
```
ryanschneider: I also considered makng these one function:

```rust
fn supports_not_null(&self, with_space: bool) -> bool { 
  false
}
```

Let me know if you prefer that.

src/parser/mod.rs
```diff
@@ -3571,6 +3572,11 @@ impl<'a> Parser<'a> {
                             ),
                             regexp,
                         })
+                    } else if dialect.supports_not_null() && negated && null {
+                        Ok(Expr::NotNull {
+                            expr: Box::new(expr),
+                            one_word: false,
```
ryanschneider: On second thought I do think I prefer inverting this to `with_space` instead of `one_word`, thoughts?

src/dialect/mod.rs
```diff
@@ -1076,6 +1088,15 @@ pub trait Dialect: Debug + Any {
     fn supports_comma_separated_drop_column_list(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports the passed in alias.
+    /// See [IsNotNullAlias].
+    fn supports_is_not_null_alias(&self, alias: IsNotNullAlias) -> bool {
```
iffyio: would it make sense to have this behavior dialect agnostic and let the parser always accept any variant that shows up? it would let us skip this dialect method as a result

src/dialect/mod.rs
```diff
@@ -650,8 +651,19 @@ pub trait Dialect: Debug + Any {
...
+                {
+                    Ok(p!(Is))
+                }
                 _ => Ok(self.prec_unknown()),
             },
+            Token::Word(w)
+                if w.keyword == Keyword::NOTNULL && self.supports_is_not_null_alias(NotNull) =>
+            {
+                Ok(p!(Is))
+            }
```
iffyio: for the new operators can we add a test case that cover their precedence behavior? see [here](https://github.com/apache/datafusion-sqlparser-rs/blob/2bf08db93d994b056820f747f3a6335e3046738a/tests/sqlparser_common.rs#L1893) for example. It would be good if we don't already have coverage to also ensure that we aren't also inadvertently changing the precedence of the `IS NOT NULL` operator as well, so we could also include a test case demonstrating that if lacking

tests/sqlparser_common.rs
```diff
@@ -15974,3 +15974,133 @@ fn parse_create_procedure_with_parameter_modes() {
...
+    }
+}
+
+#[test]
+fn parse_notnull_unsupported() {
+    // Only Postgres, DuckDB, and SQLite support `x NOTNULL` as an expression
+    // All other dialects consider `x NOTNULL` like `x AS NOTNULL` and thus
+    // consider `NOTNULL` an alias for x.
+    let sql = r#"WITH t AS (SELECT NULL AS x) SELECT x NOTNULL FROM t"#;
+    let canonical = r#"WITH t AS (SELECT NULL AS x) SELECT x AS NOTNULL FROM t"#;
```
iffyio: I think we can simplify this test, since we're only interested in testing the expression, doing so via a CTE makes the test verbose and harder to maintain going forward without providing additional coverage (note we also have `verified_expr()` helper functions to make expression testing easier)

src/dialect/mod.rs
ryanschneider: Hmm, the main issue I see w/ that approach is that technically `x NOTNULL` is shorthand for `x AS NOTNULL` for dialects that don't support `NOTNULL` but do support aliases w/o `AS`, so this could be considered a breaking change if a query happened to by using `NOTNULL` as a column alias.

src/dialect/mod.rs
ryanschneider: Added in 637fb69dd2666b0c1999fbbb0ba9b3

tests/sqlparser_common.rs
ryanschneider: Thanks!  I simplified the tests in 431fca3682cfc78ee61a2493f85f23, thanks for the pointers they are much more readable now!

src/dialect/mod.rs
iffyio: Ah makes sense! That behavior would only be applicable to the `x NOTNULL` scenario or? thinking if so, for `x NOT NULL` we should be able to parse that unambiguously as most other operators? Thinking if so then we could get away with only introducing a dialect method `supports_notnull_operator()` and parse the latter unconditionally?

ryanschneider: > Thinking if so then we could get away with only introducing a dialect method supports_notnull_operator() and parse the latter unconditionally?

Sounds good, I like that change, it gets rid of the new `enum` as well, I'll re-request review when it's ready, thanks!

ryanschneider: Hmm, I made the change but now some of the `CREATE TABLE` tests are failing around the `IsNotNull` column options, so I think I broke something w/ precedence afterall so need to take a closer look at the parsing.

ryanschneider: Ok, I was able to make it work, but it required an ugly special case in `parse_subexpr`.  The good news is with this special case I didn't need to do any precedence changes for `NOT NULL`, just `NOTNULL`.

If you have any suggestions on how to better handle this special case, please let me know!  The reason I need to handle it there is because `get_next_precedence()` doesn't know about the current `expr`, so one fix would be to pass `expr` to `get_next_precedence(expr)` then the precedence of `NOT NULL` could be dynamic if the current expr is an `Expr::Identifier` or `Value::Null`, then the logic could be moved back into `parse_infix`.

Alternaitvely, I could move the special case into a helper function to compartmentalize the changes, but figured I'd get your feedback before going to far.  Thanks!

ryanschneider: ^ @iffyio (sorry forget to mention you in previous message)

ryanschneider:
Rebased against main to resolve conflicts.

src/dialect/mod.rs
ryanschneider: I forgot to explain what I'm trying to work around: in `CREATE TABLE` statements the `NOT NULL` should never be considered an `Expr::IsNotNull` but it's getting parsed as such by the `parse_expr` calls in `parse_optional_column_option` on column definitions like `foo DEFAULT true NOT NULL` etc.

So an alternative to the above special case would be to add a "mode" to `parse_expr` so it knows when it's parsing column definitions vs. other expressions, or maybe I could replace the calls to `parse_expr` in `parse_optional_column_option` with calls to `parse_subexpr(some_high_precedence)` where `some_high_precedence` is high enough to prevent it from considering the `NOT NULL`.

ryanschneider: Hmm, what about adding a `ParserState::ColumnDefinition` and using the state in `get_next_precedence_default()` to decide if `NOT NULL` should have `Is` precedence?  If you're ok with this approach I think it's the cleanest way to handle it.

ryanschneider:
@iffyio I went with the new ParserState::ColumnDefinition idea mentioned here: https://github.com/apache/datafusion-sqlparser-rs/pull/1927#discussion_r2200796876 I think it leads to the cleanest code changes, let me know what you think!  Also merged in main and resolved conflicts.

src/dialect/mod.rs
iffyio: Right yeah seems its a bit tricky with the `NOT NULL`scenario in create statements, the parser state looks clean but not sure if it can correctly handle the different scenarios? thinking for example, the following if I understand the syntax correctly should be equivalent:
```rust
verified_stmt("CREATE TABLE foo (abc INT DEFAULT (42 IS NOT NULL) NOT NULL)");
verified_stmt("CREATE TABLE foo (abc INT DEFAULT (42 NOT NULL) NOT NULL)");
```
so that only setting the state once on the outer expression/clause wouldn't be correct. not sure if there's a clean way that doesn't require an invasive change here :thinking: it might indeed be that we need to explicitly configure the parse_expr function depending on the context

ryanschneider: Hmm good point, I'll experiment with some possibilities and see if I find a clean solution.

ryanschneider: Ok @iffyio, in bf25f42 I followed the pattern of a couple other spots in that fn and in the `Keyword::DEFAULT` handler I peek the next token, and if it's a LParen then switch back to `ParserState::Normal` while parsing that.

I see a couple other spots where this function calls `parse_expr`:

- `ColumnOption::Materialized`
- `ColumnOption::Alias`
- `ColumnOption::Ephemeral`
- `ColumnOption::Check`
- `ColumnOption::Srid`

I'll be honest I'm not familiar with most of these, so will need to look to see if any of them need the same treatment.

ryanschneider: Ok in bc454c6 I added support for `NOT NULL` inside `MATERIALIZED` and `CHECK` options as well.  Let me know what you think!

tests/sqlparser_clickhouse.rs
```diff
@@ -1705,6 +1705,15 @@ fn parse_table_sample() {
     clickhouse().verified_stmt("SELECT * FROM tbl SAMPLE 1 / 10 OFFSET 1 / 2");
 }
 
+#[test]
+fn test_parse_not_null_in_column_options() {
+    // In addition to DEFAULT and CHECK ClickHouse also supports MATERIALIZED, all of which
+    // can contain `IS NOT NULL` and thus `NOT NULL` as an alias.
+    let canonical = "CREATE TABLE foo (abc INT DEFAULT (42 IS NOT NULL) NOT NULL, not_null BOOL MATERIALIZED (abc IS NOT NULL), CHECK (abc IS NOT NULL))";
+    clickhouse().verified_stmt(canonical);
+    clickhouse().one_statement_parses_to("CREATE TABLE foo (abc INT DEFAULT (42 NOT NULL) NOT NULL, not_null BOOL MATERIALIZED (abc IS NOT NULL), CHECK (abc NOT NULL) )", canonical);
```
iffyio:
```suggestion
    clickhouse().one_statement_parses_to("CREATE TABLE foo (abc INT DEFAULT (42 NOT NULL) NOT NULL, not_null BOOL MATERIALIZED (abc NOT NULL), CHECK (abc NOT NULL) )", canonical);
```
I think this would be the intention?

tests/sqlparser_clickhouse.rs
```diff
@@ -1705,6 +1705,15 @@ fn parse_table_sample() {
     clickhouse().verified_stmt("SELECT * FROM tbl SAMPLE 1 / 10 OFFSET 1 / 2");
 }
 
+#[test]
+fn test_parse_not_null_in_column_options() {
+    // In addition to DEFAULT and CHECK ClickHouse also supports MATERIALIZED, all of which
+    // can contain `IS NOT NULL` and thus `NOT NULL` as an alias.
+    let canonical = "CREATE TABLE foo (abc INT DEFAULT (42 IS NOT NULL) NOT NULL, not_null BOOL MATERIALIZED (abc IS NOT NULL), CHECK (abc IS NOT NULL))";
```
iffyio: For the longer strings, in order to break them up into lines could we use this [concat pattern](https://github.com/apache/datafusion-sqlparser-rs/blob/b7d4e1e7fcf0abcbf2692485e90796b70297b08b/tests/sqlparser_clickhouse.rs#L545-L553)?

src/parser/mod.rs
```diff
@@ -7724,6 +7737,27 @@ impl<'a> Parser<'a> {
...
+            |parser| -> Result<Option<ColumnOption>, ParserError> {
+                parser.parse_optional_column_option_inner()
+            },
+        )
+    }
+
+    fn parse_optional_column_option_inner(&mut self) -> Result<Option<ColumnOption>, ParserError> {
+        /// In some cases, we need to revert to [ParserState::Normal] when parsing nested expressions
+        /// In those cases we use the following macro to parse instead of calling [parse_expr] directly.
+        macro_rules! parse_expr_normal {
```
iffyio: Can we use a regular function for this vs a macro?

In terms of fn signature, would this work just well `fn parse_column_option_expr(&mut self) -> Result<Expr, ParserError>`?
then the caller can explicitly do e.g. `Ok(Some(ColumnOption::Default(self.parse_column_option()?)))`

iffyio: Documentation wise, I'm thinking it could be helpful to illustrate why we need this function and with an example like below. And that we also callout that any column option that parses an arbitrary expression likely wants to call this variant instead of `parse_expr()`, for folks that wander around this area of the code in the future looking to add/update column options support.

```rust
CREATE TABLE foo (abc (42 NOT NULL) NOT NULL);
vs
CREATE TABLE foo (abc 42 NOT NULL);
```

src/parser/mod.rs
```diff
@@ -7724,6 +7737,27 @@ impl<'a> Parser<'a> {
...
+            ($option:expr) => {
+                if matches!(self.peek_token().token, Token::LParen) {
```
```suggestion
                if self.peek_token_ref().token == Token::LParen {
```

src/parser/mod.rs
```diff
@@ -16514,6 +16549,10 @@ impl<'a> Parser<'a> {
             Ok(None)
         }
     }
+
+    pub fn in_normal_state(&self) -> bool {
+        matches!(self.state, ParserState::Normal)
```
iffyio: I'm wondering should this check be `self.state != ParserState::ColumnDefinition`? Thinking that covers the enum being extended to other interactions later on (feels like the not null logic only cares that its not in this state in order to be correct)?

src/parser/mod.rs
```diff
@@ -275,6 +275,9 @@ enum ParserState {
     /// PRIOR expressions while still allowing prior as an identifier name
     /// in other contexts.
     ConnectBy,
+    /// The state when parsing column definitions.  This state prohibits
+    /// NOT NULL as an alias for IS NOT NULL.
+    ColumnDefinition,
```
iffyio, line 284: Could we include a sql example to illustrate this? I'm wondering that folks won't be able to fully get the picture without reading the code otherwise

src/parser/mod.rs
```diff
@@ -28,16 +28,16 @@ use helpers::attached_token::AttachedToken;
...
 use crate::ast::helpers::stmt_create_table::{CreateTableBuilder, CreateTableConfiguration};
 use crate::ast::Statement::CreatePolicy;
 use crate::ast::*;
 use crate::dialect::*;
 use crate::keywords::{Keyword, ALL_KEYWORDS};
 use crate::tokenizer::*;
+use recursion::RecursionCounter;
+use sqlparser::parser::ParserState::ColumnDefinition;
+use IsLateral::*;
+use IsOptional::*;
```
iffyio: was the change intentional or autofmt it looks like only one import is being added, if the latter could we revert to a minimal diff?

src/parser/mod.rs
```diff
@@ -16514,6 +16549,10 @@ impl<'a> Parser<'a> {
             Ok(None)
         }
     }
+
+    pub fn in_normal_state(&self) -> bool {
```
```suggestion
    fn in_normal_state(&self) -> bool {
```

src/dialect/mod.rs
iffyio: Yeah I think the approach look reasonable thanks! Having to explicitly configure each option that accepts arbitrary expression would be the downside but that feels like an acceptable trade-off. We would indeed need to do a pass through the existing options like Ephemeral and do similar like we have for `MATERIALIZED` and `CHECK`. We also have a couple like `OnUpdate` and some in `self.parse_optional_column_option_generated()`

ryanschneider:
Thanks for the latest round of feedback @iffyio, I've been busy with some unrelated work so haven't had time to look at it all yet or start on the changes but hopefully will have some space cycles the next couple days, I'll re-request review when it's ready!

src/parser/mod.rs
ryanschneider: This function is used in `Dialect::get_next_precedence_default` but I'll change it to `pub(crate)`.

ryanschneider:
Ok @iffyio I believe I addressed all your latest feedback in https://github.com/apache/datafusion-sqlparser-rs/pull/1927/commits/1466e2ab212744bc2270546369a04ad8ab2e893a, let me know!  Thanks!

tests/sqlparser_common.rs
```diff
@@ -16174,3 +16198,39 @@ fn test_identifier_unicode_start() {
...
+        supported_dialects.expr_parses_to("NOT NULL NOTNULL", "NOT NULL IS NOT NULL"),
+        Expr::UnaryOp {
+            op: UnaryOperator::Not,
+            ..
+        }
+    );
+
+    // for unsupported dialects, parsing should stop at `NOT NULL`
+    unsupported_dialects.expr_parses_to("NOT NULL NOTNULL", "NOT NULL");
+}
```
iffyio: Since they're the same feature, can we merge these scenarios into the same `fn parse_notnull(){}` test function?.

Also for the comments, ideally we skip mentioning the individual dialects (Postgres, duckdb etc), so that the comment doesnt' become stale as other/new dialects are extended with support for this feature.

tests/sqlparser_common.rs
```diff
@@ -16031,6 +16031,30 @@ fn parse_create_procedure_with_parameter_modes() {
...
+        }
+    );
+    assert_matches!(
+        all_dialects().expr_parses_to("NOT x NOT NULL", "NOT x IS NOT NULL"),
+        Expr::UnaryOp {
+            op: UnaryOperator::Not,
+            ..
+        }
+    );
+}
```
iffyio: Similar to the `notnull` scenarios, can we merge these into the same function?

iffyio (commented):
Thanks @ryanschneider! Could we do a pass through the options mentioned [here](https://github.com/apache/datafusion-sqlparser-rs/pull/1927#discussion_r2205650933) and [here](https://github.com/apache/datafusion-sqlparser-rs/pull/1927#discussion_r2210620051) to extend them with similar support for the `NOT NULL` clause? Beyond that I think the changes look good! (Also headsup that there's currently a docs failure in ci on the branch)

ryanschneider:
Sorry I should've mentioned that I did check the other ColumnOptions and confirmed that they don't _need_ to use `parse_column_option_expr` but after thinking about it more I'll go ahead and use it anyways for clarity.  Also fixing the doc failure and consolidating the tests.

ryanschneider:
Ok, I consolidated the tests, fixed the `cargo doc` error and use `parse_column_option_expr` in all spots for consistency.

PR #1928, opened 2025-07-06
Fix for Postgres regex and like binary operators

### LINKED ISSUE #1776: Postgres dialect fails to parse "~ any(...)"
The following query
```sql
select a from aa where a ~ any(array['x'])
```
yields 
```
Expected one of [=, >, <, =>, =<, !=] as comparison operator, found: ~ at Line: 1, Column: 26
```

solontsev:
Closes #1776

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -2181,19 +2181,35 @@ fn parse_pg_regex_match_ops() {
     ];
 
     for (str_op, op) in pg_regex_match_ops {
-        let select = pg().verified_only_select(&format!("SELECT 'abc' {} '^a'", &str_op));
+        // Basic binary operation usage
+        let select = pg().verified_only_select(&format!("SELECT 'abc' {str_op} '^a'"));
         assert_eq!(
             SelectItem::UnnamedExpr(Expr::BinaryOp {
-                left: Box::new(Expr::Value(
-                    (Value::SingleQuotedString("abc".into())).with_empty_span()
-                )),
+                left: Box::new(Expr::Value(single_quoted_string("abc").with_empty_span(),)),
                 op: op.clone(),
-                right: Box::new(Expr::Value(
-                    (Value::SingleQuotedString("^a".into())).with_empty_span()
-                )),
+                right: Box::new(Expr::Value(single_quoted_string("^a").with_empty_span(),)),
             }),
-            select.projection[0]
+            select.projection[0],
         );
+
+        // Binary operator with ANY operator
+        let select =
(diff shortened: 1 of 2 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -2207,19 +2223,31 @@ fn parse_pg_like_match_ops() {
...
-                right: Box::new(Expr::Value(
-                    (Value::SingleQuotedString("a_c%".into())).with_empty_span()
-                )),
+                right: Box::new(Expr::Value(single_quoted_string("a_c%").with_empty_span(),)),
             }),
             select.projection[0]
         );
+
+        // Binary operator with ALL operator
+        let select =
```
iffyio, line 2242: I'm not sure I fully understood the fix, is the issue specific to the `ALL` and `ANY` functions? the current test doesn't seem to make changes to the [list of operators](https://github.com/apache/datafusion-sqlparser-rs/blob/60acdc5bb67383ba0e4b3a6291ca6701ab2ec752/tests/sqlparser_postgres.rs#L2218-L2223) being tested so that its not clear to me why this test fails without the fix in this PR

solontsev: Currently, when parsing expressions with regex and like operators, only literals are supported, while syntax like `where a ~ any(array['x'])` is also valid (matching against an array of values using ALL or ANY). I've added additional more complex expressions to the tests (fails with the current version) + allowed 8 more operators before the ALL and ANY functions in the `parse_infix` function.

https://github.com/apache/datafusion-sqlparser-rs/blob/60acdc5bb67383ba0e4b3a6291ca6701ab2ec752/src/parser/mod.rs#L3481-L3488

iffyio: > + allowed 8 more operators before the ALL and ANY functions in the parse_infix function.

Are the added operators covered by the tests? if not could we add tests covering them?

solontsev: > Are the added operators covered by the tests? if not could we add tests covering them?

Yeah, I've added the tests as well. Maybe it was not the best to test 2 cases during each cycle iteration, I've modified it a bit: a previous match against a single value, and added one more to test against an array of values. Does it look better now?

Also, to wrap up: The problem now, is that before ANY, ALL or SOME functions, parser allows only 6 basic binary operators (gt, lt, gteq, lteq, eq, noteq). But those 8 additional are also valid. So the fix basically just adds them to the allowlist and tests for match and like operators are modified to make sure, parser doesn't return an error in a valid cases. I haven't added exhaustive test cases, just chose ANY function for match tests and ALL function for like tests.

iffyio (approved):
LGTM! Thanks @solontsev!
cc @alamb

PR #1929, opened 2025-07-06
Add support for Snowflake identifier function

yoavcloud:
Expand the support for the `IDENTIFIER` function in Snowflake, which is used to generate identifiers dynamically. See here: https://docs.snowflake.com/en/sql-reference/identifier-literal

The ObjectNamePart enum was extended with a new part type to support the use of the identifier function as an object name. Some work went into organizing `Parser:parse_object_name` a bit.

A few examples:
```sql
-- identifier for table name
create or replace temp table identifier('test')(c1 int, identifier int); 
-- identifier for column name
select identifier('identifier') from test;
-- identifier for table name
select c1 from identifier('test');
-- identifier for table alias
select identifier('alias1').* from test as alias1;
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10409,7 +10409,10 @@ impl<'a> Parser<'a> {
                     parts.push(ObjectNamePart::Identifier(Ident::new("")));
                 } else {
                     let ident = self.parse_identifier()?;
-                    let part = if self.dialect.is_identifier_generating_function_name(&ident) {
+                    let part = if self
+                        .dialect
+                        .is_identifier_generating_function_name(&ident, &parts)
+                    {
                         self.expect_token(&Token::LParen)?;
                         let args: Vec<FunctionArg> =
                             self.parse_comma_separated0(Self::parse_function_args, Token::RParen)?;
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -4350,4 +4350,20 @@ fn test_snowflake_identifier_function() {
         }
         _ => unreachable!(),
     }
+
+    // Cannot have more than one IDENTIFIER part in an object name
+    assert_eq!(
+        snowflake()
+            .parse_sql_statements(
+                "CREATE TABLE IDENTIFIER('db1').IDENTIFIER('sc1').IDENTIFIER('tbl') (id INT)"
+            )
+            .is_err(),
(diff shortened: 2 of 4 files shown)


src/parser/mod.rs
```diff
@@ -10353,70 +10353,84 @@ impl<'a> Parser<'a> {
...
+                if allow_wildcards && self.peek_token().token == Token::Mul {
                     let span = self.next_token().span;
-                    Ident {
+                    parts.push(ObjectNamePart::Identifier(Ident {
                         value: Token::Mul.to_string(),
                         quote_style: None,
                         span,
+                    }));
+                } else if let Some(func_part) =
+                    self.maybe_parse(|parser| parser.parse_object_name_function_part())?
```
iffyio: wondering can we instead extende the `else` clause to do something like this?
```rust
else {
    let ident = self.parse_identifier()?;
    let part = if self.dialect.is_identifier_generating_function_name(&ident) {
          let args = ...
         ObjectNamePartFunction { name:ident, args }
    } else {
          ObjectNamePart::Identifier(ident));
    }
}
```
thinking that way we skip this extra branch, and the extra error scenario in the function part parsing

tests/sqlparser_snowflake.rs
```diff
@@ -4232,3 +4232,122 @@ fn test_snowflake_create_view_with_composite_policy_name() {
         r#"CREATE VIEW X (COL WITH MASKING POLICY foo.bar.baz) AS SELECT * FROM Y"#;
     snowflake().verified_stmt(create_view_with_tag);
 }
+
+#[test]
```
iffyio, line 4236: Could we add a test scenario for how this interacts with the [double dot](https://github.com/apache/datafusion-sqlparser-rs/blob/018992b22a78586dc4f176e80c6f1bf171bbfd7d/tests/sqlparser_snowflake.rs#L3330) feature?

yoavcloud: In what way? I believe you cannot chain IDENTIFIER. For example, this is illegal: SELECT * FROM IDENTIFIER('db1')..IDENTIFIER('tbl1').

iffyio: Yeah was thinking a test as in your example to demonstrate if/how the parser handles it. If its unsupported then it'll be a negative test

PR #1931, opened 2025-07-08
Sf create table as

yoavcloud:
The code that parses `CREATE TABLE` in the Snowflake dialect assumed that if the `AS`, `LIKE` or `CLONE` options are used, then no other options can be specified. That is not true, so that restriction is now removed.

For example:
`CREATE TEMP TABLE dst AS (SELECT * FROM src) ON COMMIT PRESERVE ROWS`


src/ast/helpers/stmt_create_table.rs
```diff
@@ -383,6 +383,16 @@ impl CreateTableBuilder {
         self
     }
 
+    /// Returns true if information on the structure of the table
+    /// to be created was provided to the builder. If not, the
+    /// statement is invalid.
+    pub fn has_schema_info(&self) -> bool {
```
iffyio:
```suggestion
    pub(crate) fn has_schema_info(&self) -> bool {
```

src/ast/helpers/stmt_create_table.rs
```diff
@@ -383,6 +383,16 @@ impl CreateTableBuilder {
...
+        !self.columns.is_empty()
+            || self.query.is_some()
+            || self.like.is_some()
+            || self.clone.is_some()
```
iffyio: could we add tests covering this behavior? e.g a test that fails when only a query, columns, like etc is provided in a create statement

tests/sqlparser_snowflake.rs
```diff
@@ -995,6 +995,21 @@ fn test_snowflake_create_iceberg_table_without_location() {
...
+        .parse_sql_statements(
+            "CREATE TEMP TABLE dst AS (SELECT * FROM src) ON COMMIT PRESERVE ROWS",
+        )
+        .unwrap();
+    snowflake()
+        .parse_sql_statements("CREATE TEMP TABLE tbl LIKE customers ON COMMIT PRESERVE ROWS;")
+        .unwrap();
+    snowflake()
+        .parse_sql_statements("CREATE TEMP TABLE tbl CLONE customers ON COMMIT PRESERVE ROWS;")
+        .unwrap();
```
iffyio: can we use verified_stmt here or similar roundtrip assertion?

yoavcloud: Improved the tests, note Snowflake supports specifying the options either before or after the "source" definition while the SQL output assumes one. Added tests for both options.

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1934, opened 2025-07-09
Snowflake Reserved SQL Keywords as Implicit Table Alias

yoavcloud:
Following the same pattern of `Dialect::is_select_item_alias` for table aliases. Added handling for Snowflake-specific keywords when used as an implicit table alias (i.e. without `AS`).

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -318,8 +318,11 @@ impl Dialect for SnowflakeDialect {
             }
 
             // `FETCH` can be considered an alias as long as it's not followed by `FIRST`` or `NEXT`
-            // which would give it a different meanings, for example: `SELECT 1 FETCH FIRST 10 ROWS` - not an alias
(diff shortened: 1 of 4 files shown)


src/dialect/snowflake.rs
```diff
@@ -345,6 +345,85 @@ impl Dialect for SnowflakeDialect {
...
+                false
+            }
+
+            // `FETCH` can be considered an alias as long as it's not followed by `FIRST`` or `NEXT`
+            // which would give it a different meanings, for example:
+            // `SELECT * FROM tbl FETCH FIRST 10 ROWS` - not an alias
+            // `SELECT * FROM tbl FETCH 10` - not an alias
+            Keyword::FETCH
+                if parser.peek_keyword(Keyword::FIRST)
+                    || parser.peek_keyword(Keyword::NEXT)
```
```suggestion
                if parser.peek_one_of_keyword(&[Keyword::FIRST, Keyword::NEXT])
```
I think this would be equivalent and simplify?

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1936, opened 2025-07-10
Add support for Redshift `SELECT * EXCLUDE`

yoavcloud:
Redshift supports placing the `EXCLUDE` option at the end of the projection list, not necessarily after the wildcard. For example: `SELECT *, c1 EXCLUDE c2 FROM test` (exclude column c2 from the list of all columns including c1).

Added support for parsing this option by reusing the existing `ExcludeSelectItem` struct but placing it at the query level, after the projection.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -16058,4 +16058,48 @@ fn test_select_exclude() {
         select.exclude,
         Some(ExcludeSelectItem::Single(Ident::new("c1")))
     );
+
+    let dialects = all_dialects_where(|d| {
+        d.supports_select_wildcard_exclude() && !d.supports_select_exclude()
+    });
+    let select = dialects.verified_only_select("SELECT * EXCLUDE c1 FROM test");
+    match &select.projection[0] {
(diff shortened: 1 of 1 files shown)


tests/sqlparser_common.rs
```diff
@@ -15982,3 +15992,64 @@ fn parse_create_procedure_with_parameter_modes() {
...
+        _ => unreachable!(),
+    }
+    match &select.projection[1] {
+        SelectItem::UnnamedExpr(Expr::Identifier(ident)) => {
+            assert_eq!(*ident, Ident::new("c2"));
+        }
+        _ => unreachable!(),
+    }
+
+    let dialects = all_dialects_where(|d| d.supports_select_exclude());
```
iffyio, line 16049: Could we add a test with
```rust
let dialects = all_dialects_where(|d| d.supports_select_wildcard_exclude() && !d.supports_select_exclude());
```
like in snowflake's case to demonstrate the behavior when there is e.g. an exclude that is an ident `SELECT *, EXCLUDE c1 FROM test` or similar

PR #1937, opened 2025-07-11
Support optional semicolon between statements

yoavcloud:
T-SQL scripts do not require a semicolon as a statement delimiter. Added a new parser option to not require semicolons, as introducing this at the dialect level seems to be very intrusive. Also added a `tsql()` function to create an MS SQL dialect for testing with the new option.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -223,7 +223,7 @@ pub struct ParserOptions {
     /// [`Tokenizer::with_unescape`] for more details.
     pub unescape: bool,
     /// Controls if the parser expects a semi-colon token
-    /// between statements.
+    /// between statements. Default is `true`.
     pub require_semicolon_stmt_delimiter: bool,
 }
 
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -294,6 +294,11 @@ pub fn all_dialects() -> TestedDialects {
     ])
 }
 
+// Returns all available dialects with the specified parser options
+pub fn all_dialects_with_options(options: ParserOptions) -> TestedDialects {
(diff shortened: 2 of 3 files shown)


tests/sqlparser_mssql.rs
```diff
@@ -2480,3 +2492,15 @@ fn parse_mssql_grant() {
 fn parse_mssql_deny() {
     ms().verified_stmt("DENY SELECT ON my_table TO public, db_admin");
 }
+
+#[test]
```
iffyio, line 2499: Could we have a similar test or move this over to common.rs? thinking since this flag is being added to the parser itself, it feels a bit like it could be lost if only tested here within mssql

src/parser/mod.rs
```diff
@@ -222,13 +222,17 @@ pub struct ParserOptions {
     /// Controls how literal values are unescaped. See
     /// [`Tokenizer::with_unescape`] for more details.
     pub unescape: bool,
+    /// Controls if the parser expects a semi-colon token
+    /// between statements.
```
iffyio: Thinking it might be helpful if we include in the doc that the default is true?

aharpervc:
I went to rebase https://github.com/apache/datafusion-sqlparser-rs/pull/1843 and noticed this PR had been merged. Great! I see we had similar thoughts on how to approach it. I closed my PR.

However, my PR also introduced a lot more testing for the feature missing from this PR. I'll open a separate PR for more test coverage if it turns out to be needed.

alamb:
> However, my PR also introduced a lot more testing for the feature missing from this PR. I'll open a separate PR for more test coverage if it turns out to be needed.

Havign some more test coverage would be great -- if you have the time to make one please ping me and I will be happy to review it

PR #1939, opened 2025-07-11
Snowflake create database

### LINKED ISSUE #1938: Add support for Snowflake CREATE DATABASE
https://docs.snowflake.com/en/sql-reference/sql/create-database

osipovartem:
https://docs.snowflake.com/en/sql-reference/sql/create-database

Added support for
 ```sql
CREATE [ OR REPLACE ] [ TRANSIENT ] DATABASE [ IF NOT EXISTS ] <name>
    [ CLONE <source_schema>
        [ { AT | BEFORE } ( { TIMESTAMP => <timestamp> | OFFSET => <time_difference> | STATEMENT => <id> } ) ]
        [ IGNORE TABLES WITH INSUFFICIENT DATA RETENTION ]
        [ IGNORE HYBRID TABLES ] ]
    [ DATA_RETENTION_TIME_IN_DAYS = <integer> ]
    [ MAX_DATA_EXTENSION_TIME_IN_DAYS = <integer> ]
    [ EXTERNAL_VOLUME = <external_volume_name> ]
    [ CATALOG = <catalog_integration_name> ]
    [ REPLACE_INVALID_CHARACTERS = { TRUE | FALSE } ]
    [ DEFAULT_DDL_COLLATION = '<collation_specification>' ]
    [ STORAGE_SERIALIZATION_POLICY = { COMPATIBLE | OPTIMIZED } ]
    [ COMMENT = '<string_literal>' ]
    [ CATALOG_SYNC = '<snowflake_open_catalog_integration_name>' ]
    [ CATALOG_SYNC_NAMESPACE_MODE = { NEST | FLATTEN } ]
    [ CATALOG_SYNC_NAMESPACE_FLATTEN_DELIMITER = '<string_literal>' ]
    [ [ WITH ] TAG ( <tag_name> = '<tag_value>' [ , <tag_name> = '<tag_value>' , ... ] ) ]
    [ WITH CONTACT ( <purpose> = <contact_name> [ , <purpose> = <contact_name> ... ] ) ]
```

Closes #1938

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -598,29 +598,11 @@ pub fn parse_create_table(
                 }
                 Keyword::ENABLE_SCHEMA_EVOLUTION => {
                     parser.expect_token(&Token::Eq)?;
-                    let enable_schema_evolution =
-                        match parser.parse_one_of_keywords(&[Keyword::TRUE, Keyword::FALSE]) {
-                            Some(Keyword::TRUE) => true,
-                            Some(Keyword::FALSE) => false,
-                            _ => {
-                                return parser.expected("TRUE or FALSE", next_token);
-                            }
-                        };
-
-                    builder = builder.enable_schema_evolution(Some(enable_schema_evolution));
+                    builder = builder.enable_schema_evolution(Some(parser.parse_boolean_string()?));
                 }
                 Keyword::CHANGE_TRACKING => {
                     parser.expect_token(&Token::Eq)?;
-                    let change_tracking =
-                        match parser.parse_one_of_keywords(&[Keyword::TRUE, Keyword::FALSE]) {
-                            Some(Keyword::TRUE) => true,
-                            Some(Keyword::FALSE) => false,
-                            _ => {
-                                return parser.expected("TRUE or FALSE", next_token);
-                            }
-                        };
-
-                    builder = builder.change_tracking(Some(change_tracking));
+                    builder = builder.change_tracking(Some(parser.parse_boolean_string()?));
                 }
                 Keyword::DATA_RETENTION_TIME_IN_DAYS => {
                     parser.expect_token(&Token::Eq)?;
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9772,7 +9772,7 @@ impl<'a> Parser<'a> {
     }
 
     /// Parse a boolean string
-    pub fn parse_boolean_string(&mut self) -> Result<bool, ParserError> {
+    pub(crate) fn parse_boolean_string(&mut self) -> Result<bool, ParserError> {
         match self.parse_one_of_keywords(&[Keyword::TRUE, Keyword::FALSE]) {
             Some(Keyword::TRUE) => Ok(true),
             Some(Keyword::FALSE) => Ok(false),

### CODE REVIEW

File: src/ast/ddl.rs [hunk 1]
```diff
@@ -2524,3 +2524,104 @@ impl fmt::Display for CreateConnector {
         Ok(())
     }
 }
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct CreateSnowflakeDatabase {
```
iffyio: Can we reuse [this existing representation](https://github.com/apache/datafusion-sqlparser-rs/blob/7654a49b50aa0d05475901fc145bdf1f8c14dfff/src/ast/mod.rs#L3853-L3857) of the `CREATE DATABASE` statement? essentially adding the snowflake parameters to it. We try to avoid dialect specific representations in the AST in general.

Since we're looking to add quite a few  options, it might be that we would ideally need to first convert the current `Statement::CreateDatabase{...}` away from the [anonymous struct syntax](https://github.com/apache/datafusion-sqlparser-rs/issues/1204) towards an explicit struct e.g.
```rust
Statement::CreateDatabase(CreateDatabase{...})
```

src/ast/mod.rs
```diff
@@ -3857,6 +3857,31 @@ pub enum Statement {
...
+    ///     [ CATALOG = <catalog_integration_name> ]
+    ///     [ REPLACE_INVALID_CHARACTERS = { TRUE | FALSE } ]
+    ///     [ DEFAULT_DDL_COLLATION = '<collation_specification>' ]
+    ///     [ STORAGE_SERIALIZATION_POLICY = { COMPATIBLE | OPTIMIZED } ]
+    ///     [ COMMENT = '<string_literal>' ]
+    ///     [ CATALOG_SYNC = '<snowflake_open_catalog_integration_name>' ]
+    ///     [ CATALOG_SYNC_NAMESPACE_MODE = { NEST | FLATTEN } ]
+    ///     [ CATALOG_SYNC_NAMESPACE_FLATTEN_DELIMITER = '<string_literal>' ]
+    ///     [ [ WITH ] TAG ( <tag_name> = '<tag_value>' [ , <tag_name> = '<tag_value>' , ... ] ) ]
+    ///     [ WITH CONTACT ( <purpose> = <contact_name> [ , <purpose> = <contact_name> ... ] ) ]
```
iffyio: Re documentation, i think using a simple example or only mentioning that it represents the `CREATE DATABASE` statement would suffice, the actual spec we can delegate to the linked documentation for each dialect.

src/ast/mod.rs
```diff
@@ -9524,6 +9550,29 @@ impl Display for Tag {
...
+pub struct ContactEntry {
+    pub purpose: String,
+    pub contact: String,
+}
+
+impl ContactEntry {
+    pub fn new(purpose: String, contact: String) -> Self {
+        Self { purpose, contact }
+    }
+}
```
```suggestion
```
I think we can remove this in order to keep the public API surface smaller

src/dialect/snowflake.rs
```diff
@@ -731,6 +734,115 @@ pub fn parse_create_table(
...
+                            let contact = p.parse_identifier()?.value;
+                            Ok(ContactEntry::new(purpose, contact))
+                        })?;
+                        parser.expect_token(&Token::RParen)?;
+                        builder = builder.with_contacts(Some(contacts));
+                    } else {
+                        return parser.expected("TAG or CONTACT", next_token);
+                    }
+                }
+                _ => return parser.expected("end of statementrrr", next_token),
```
```suggestion
                _ => return parser.expected("end of statement", next_token),
```

tests/sqlparser_snowflake.rs
```diff
@@ -4438,3 +4438,63 @@ fn test_snowflake_identifier_function() {
...
+        DATA_RETENTION_TIME_IN_DAYS = 1
+        MAX_DATA_EXTENSION_TIME_IN_DAYS = 5
+        EXTERNAL_VOLUME = 'volume1'
+        CATALOG = 'my_catalog'
+        REPLACE_INVALID_CHARACTERS = TRUE
+        DEFAULT_DDL_COLLATION = 'en-ci'
+        STORAGE_SERIALIZATION_POLICY = COMPATIBLE
+        COMMENT = 'This is my database'
+        CATALOG_SYNC = 'sync_integration'
+        CATALOG_SYNC_NAMESPACE_FLATTEN_DELIMITER = '/'
```
iffyio, line 4531: Can we add an example for `CATALOG_SYNC_NAMESPACE_MODE`?

tests/sqlparser_snowflake.rs
```diff
@@ -4438,3 +4438,63 @@ fn test_snowflake_identifier_function() {
         true
     );
 }
+
+#[test]
```
iffyio, line 4508: Can we merge all the added tests into the same `fn test_create_database()` function? since they are part of the same feature

src/ast/mod.rs
osipovartem: What should we use instead?

osipovartem: it has specific display format write!(f, "{} = {}", self.purpose, self.contact)

iffyio: Ah sorry to clarify I meant that we should remove the `new` function, not the struct itself. So that callers can manually do `ContactEntry { purpose, "foo", contact: "bar" }` instead of `ContactEntry::new("foo", "bar")`

osipovartem:
@iffyio please take a look

src/parser/mod.rs
```diff
@@ -9756,6 +9771,15 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// Parse a boolean string
+    pub fn parse_boolean_string(&mut self) -> Result<bool, ParserError> {
+        match self.parse_one_of_keywords(&[Keyword::TRUE, Keyword::FALSE]) {
+            Some(Keyword::TRUE) => Ok(true),
+            Some(Keyword::FALSE) => Ok(false),
+            _ => self.expected("TRUE or FALSE", self.peek_token()),
+        }
+    }
```
iffyio: oh this looks like its only used once, can we inline it instead?

iffyio (commented):
Thanks @osipovartem! Left a minor comment otherwise the changes look good to me

src/parser/mod.rs
osipovartem: This will be used in the next PR
related to https://github.com/apache/datafusion-sqlparser-rs/issues/1982

iffyio: Ah I see, but I think we can inline it in this PR and if it ends up being used in the next PR we can break it out into a function. Also heads up for the function we would want to make it either private or `pub(crate)` visibility wise instead of `pub`

osipovartem: Replaced duplicated code by parse_boolean_string call

osipovartem: Please review again

iffyio (approved):
LGTM! Thanks @osipovartem!
cc @alamb

PR #1940, opened 2025-07-12
MSSQL: Add support for EXEC output and default keywords

yoavcloud:
Added support for parsing the OUTPUT and DEFAULT keywords in MSSQL when calling a stored procedure.


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1941, opened 2025-07-13
Snowflake: Improve accuracy of lookahead in implicit LIMIT alias

yoavcloud:
Some needed adjustments to the lookahead logic when considering the `LIMIT` keyword as an implicit table alias.


src/dialect/snowflake.rs
```diff
@@ -365,6 +364,18 @@ impl Dialect for SnowflakeDialect {
                 false
             }
 
+            // `LIMIT` can be considered an alias as long as it's not followed by a value. For example:
+            // `SELECT * FROM tbl LIMIT WHERE 1=1` - alias
+            // `SELECT * FROM tbl LIMIT 3` - not an alias
+            Keyword::LIMIT
+                if matches!(
+                    parser.peek_token().token,
+                    Token::Number(_, _) | Token::Placeholder(_)
```
iffyio: from their [docs they seem to mention](https://docs.snowflake.com/en/sql-reference/constructs/limit#parameters) that null and empty string, `$` values are possible, could we also detect for those here?

yoavcloud: Thanks for the great feedback! The latest commit introduces a more accurate lookahead logic.

alamb:
It is pretty epic that this code keeps rolling along

PR #1942, opened 2025-07-14
Snowflake: Improve support for reserved keywords for table factor

yoavcloud:
Aligned logic in the `Dialect` trait to other similar flows (`is_column_alias`, `is_table_alias`, etc) and added a list of reserved keywords in Snowflake.


src/dialect/snowflake.rs
```diff
@@ -427,6 +503,21 @@ impl Dialect for SnowflakeDialect {
         }
     }
 
+    fn is_table_factor(&self, kw: &Keyword, parser: &mut Parser) -> bool {
+        match kw {
+            Keyword::LIMIT
+                if matches!(
+                    parser.peek_token().token,
+                    Token::Number(_, _) | Token::Placeholder(_)
+                ) =>
```
iffyio: would it make sense to use the [peek_for_limit_options](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/snowflake.rs#L483C4-L491) helper method here?

PR #1945, opened 2025-07-15
MySQL: Support `EXPLAIN ANALYZE` format variants

yoavcloud:
Add support for the MySQL`FORMAT=<FORMAT TYPE>` syntax for `EXPLAIN ANALYZE`

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4126,7 +4126,7 @@ pub enum Statement {
         /// A SQL query that specifies what to explain
         statement: Box<Statement>,
         /// Optional output format of explain
-        format: Option<AnalyzeFormat>,
+        format: Option<AnalyzeFormatKind>,
         /// Postgres style utility options, `(analyze, verbose true)`
         options: Option<Vec<UtilityOption>>,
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -7641,13 +7641,32 @@ impl fmt::Display for DuplicateTreatment {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum AnalyzeFormatKind {
+    Keyword(AnalyzeFormat),
+    Assignment(AnalyzeFormat),
+}
```
iffyio: Could we add some description ideally with example syntax to show the difference between each variant? Thinking that might not be obvious otherwise to someone looking at the API

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1947, opened 2025-07-15
Postgres: ALTER TABLE SET ( storage_parameters )

### LINKED ISSUE #1946: Postgres ALTER TABLE SET ( storage_parameter [= value] [, ... ] ) fails to parse
The following example is valid syntax (https://www.postgresql.org/docs/current/sql-altertable.html) but fails to parse:
```sql
ALTER TABLE your_table
SET (
  autovacuum_vacuum_scale_factor = 0.001,
  autovacuum_vacuum_threshold = 500
);
```

achristmascarl:
Closes #1946


src/ast/ddl.rs
```diff
@@ -351,6 +351,10 @@ pub enum AlterTableOperation {
     ValidateConstraint {
         name: Ident,
     },
+    /// `SET ( storage_parameter [= value] [, ... ] )`
```
iffyio: Could we include a link to the docs where this syntax comes from?

Also I'm wondering whether storageparameter is right terminology here? From what I can tell with the syntax it looks possible that it accepts arbitrary sql option in parenthesis? If so thinking something generic like `SetOptionsParens { options: Vec<SqlOption> }` or similar naming as the enum variant

achristmascarl: so Postgres does expect one of a specific set of parameters there: https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-STORAGE-PARAMETERS

and will give `ERROR:  unrecognized parameter` for an unexpected parameter. The parameters also have been the same in the last 5 versions, so they seem relatively stable.

Happy to either change to the more generic form, or update the parsing to look for the specific storage parameters instead of arbitrary options. Whatever you think is more appropriate @iffyio

iffyio: Ah I see, indeed I think the generic form would be preferred beyond the postgres dialect in this case

src/ast/ddl.rs
```diff
@@ -351,6 +351,13 @@ pub enum AlterTableOperation {
...
+    ///
+    /// Note: this is a PostgreSQL-specific operation.
+    /// Please refer to [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-altertable.html)
```
```suggestion
    /// Arbitrary parenthesized `SET` options.
    ///
    /// Example:
    /// ```sql
    /// SET (scale_factor = 0.01, threshold = 500)`
    /// ```
    /// [PostgreSQL](https://www.postgresql.org/docs/current/sql-altertable.html)
```
Can we change to something more generic like the following?

iffyio (approved):
LGTM! Thanks @achristmascarl!
cc @alamb

PR #1950, opened 2025-07-16
Snowflake: CREATE USER

yoavcloud:
Added support for the `CREATE USER` statement in Snowflake. Enhanced the KeyValueOptions struct with:
1. A custom delimiter 
2. Optional parentheses
3. Optional keywords that indicate the end of the options section

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4339,6 +4339,11 @@ pub enum Statement {
     ///
     /// See [ReturnStatement]
     Return(ReturnStatement),
+    /// ```sql
+    /// CREATE [OR REPLACE] USER <user> [IF NOT EXISTS]
+    /// ```
+    /// [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-user)
+    CreateUser(CreateUser),
 }
 
 /// ```sql
@@ -6169,6 +6174,7 @@ impl fmt::Display for Statement {
             Statement::Return(r) => write!(f, "{r}"),
             Statement::List(command) => write!(f, "LIST {command}"),
             Statement::Remove(command) => write!(f, "REMOVE {command}"),
+            Statement::CreateUser(s) => write!(f, "{s}"),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/ast/mod.rs
```diff
@@ -10074,6 +10080,42 @@ impl fmt::Display for MemberOf {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct CreateUser {
```
iffyio, line 10145: Could we add a description and link to the snowflake docs here as well? I think its usually more visible on the struct than on the enum variant since the struct is usually referenced in different contexts

src/parser/mod.rs
```diff
@@ -4689,6 +4696,32 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_create_user(&mut self, or_replace: bool) -> Result<Statement, ParserError> {
```
```suggestion
    fn parse_create_user(&mut self, or_replace: bool) -> Result<Statement, ParserError> {
```

src/parser/mod.rs
```diff
@@ -16514,6 +16547,83 @@ impl<'a> Parser<'a> {
...
+                    self.prev_token();
+                    break;
+                }
+                _ => return self.expected("another option, EOF, Comma or ')'", self.peek_token()),
+            };
+        }
+        Ok(options)
+    }
+
+    // Parses a `KEY = VALUE` construct based on the specified key
```
```suggestion
    /// Parses a `KEY = VALUE` construct based on the specified key
```

src/parser/mod.rs
```diff
@@ -16514,6 +16547,83 @@ impl<'a> Parser<'a> {
...
+                option_type: KeyValueOptionType::BOOLEAN,
+                value: "TRUE".to_string(),
+            })
+        } else if self.parse_keyword(Keyword::FALSE) {
+            Ok(KeyValueOption {
+                option_name: key.value,
+                option_type: KeyValueOptionType::BOOLEAN,
+                value: "FALSE".to_string(),
+            })
+        } else {
```
iffyio: would something like this be equivalent and simplify the if/elseif branches?
```rust
if let Some(kw) = self.parse_one_of_keywords(&[Keyword::TRUE, Keyword::FALSE]) {
    Ok(KeyValueOption{ ..., kw.to_string() })
} else {
    // ...
}
```

tests/sqlparser_common.rs
```diff
@@ -16156,3 +16156,16 @@ fn test_identifier_unicode_start() {
     ]);
     let _ = dialects.verified_stmt(sql);
 }
+
+#[test]
+fn parse_create_user() {
+    verified_stmt("CREATE USER u1");
```
iffyio: Can we include in one of the scenarios an assertion on the returned AST?

PR #1953, opened 2025-07-17
Add ODBC escape syntax support for time expressions

etgarperets:
…timestamp literals. For this I modified TypedString by adding uses_odbc_syntax flag.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -16317,20 +16317,28 @@ fn parse_notnull() {
 }
 
 #[test]
-fn test_odbc_time_date_timestamp_support() {
+fn parse_odbc_time_date_timestamp() {
+    // Supported statements
     let sql_d = "SELECT {d '2025-07-17'}, category_name FROM categories";
     let _ = all_dialects().verified_stmt(sql_d);
     let sql_t = "SELECT {t '14:12:01'}, category_name FROM categories";
     let _ = all_dialects().verified_stmt(sql_t);
     let sql_ts = "SELECT {ts '2025-07-17 14:12:01'}, category_name FROM categories";
     let _ = all_dialects().verified_stmt(sql_ts);
-}
-
-#[test]
-#[should_panic]
-fn test_invalid_odbc_literal_fails() {
+    // Unsupported statement
+    let supports_dictionary = all_dialects_where(|d| d.supports_dictionary_syntax());
+    let dictionary_unsupported = all_dialects_where(|d| !d.supports_dictionary_syntax());
     let sql = "SELECT {tt '14:12:01'} FROM foo";
-    let _ = all_dialects().verified_stmt(sql);
+    let res = supports_dictionary.parse_sql_statements(sql);
+    let res_dict = dictionary_unsupported.parse_sql_statements(sql);
+    assert_eq!(
+        ParserError::ParserError("Expected: :, found: '14:12:01'".to_string()),
+        res.unwrap_err()
+    );
+    assert_eq!(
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -2016,6 +2018,46 @@ impl<'a> Parser<'a> {
         })
     }
 
+    // Tries to parse the body of an [ODBC escaping sequence]
```
```suggestion
    /// Tries to parse the body of an [ODBC escaping sequence]
```

iffyio: > [ODBC escaping sequence]

Can we include a doc link referencing this part of the comment?

tests/sqlparser_common.rs
```diff
@@ -16174,3 +16199,13 @@ fn test_identifier_unicode_start() {
     ]);
     let _ = dialects.verified_stmt(sql);
 }
+
+#[test]
+fn test_odbc_time_date_timestamp_support() {
+    let sql_d = "SELECT {d '2025-07-17'}, category_name FROM categories";
```
iffyio, line 16322: Could we include a negative test for an expression that doesn't use the expected character? e.g. `SELECT {tt '14:12:01'} FROM foo`

iffyio (commented):
Thanks @etgarperets! The changes look good to me overall, left some minor comments

src/ast/mod.rs
```diff
@@ -7397,6 +7389,49 @@ pub struct DropDomain {
...
+            false => {
+                write!(f, "{data_type}")?;
+                write!(f, " {value}")
+            }
+            true => match data_type {
+                DataType::Date => write!(f, "{{d {value}}}"),
+                DataType::Time(..) => write!(f, "{{t {value}}}"),
+                DataType::Timestamp(..) => write!(f, "{{ts {value}}}"),
+                _ => write!(f, "{{? {value}}}"),
+            },
```
```suggestion
            true => {
                let dt = match data_type {
                    DataType::Date => "d",
                    DataType::Time(..) => "t"
                    DataType::Timestamp(..) => "ts",
                };
                write!("{{ {dt} }}")
            },
```
I think this branch could be simplified to something like above?

tests/sqlparser_common.rs
```diff
@@ -16289,6 +16314,23 @@ fn parse_notnull() {
...
+    let sql_ts = "SELECT {ts '2025-07-17 14:12:01'}, category_name FROM categories";
+    let _ = all_dialects().verified_stmt(sql_ts);
+}
+
+#[test]
+#[should_panic]
+fn test_invalid_odbc_literal_fails() {
+    let sql = "SELECT {tt '14:12:01'} FROM foo";
+    let _ = all_dialects().verified_stmt(sql);
+}
```
iffyio: here's [one example](https://github.com/apache/datafusion-sqlparser-rs/blob/3b08353f56830c9fee914a55d222d231f7be67ae/tests/sqlparser_common.rs#L420-L423) of how we do negative tests, can we rewrite to use that pattern instead of the should_panic macro? Also we can inline the scenario into the `test_odbc_time_date_timestamp_support` function since they belong to the same feature

tests/sqlparser_common.rs
```diff
@@ -16289,6 +16314,23 @@ fn parse_notnull() {
     notnull_unsupported_dialects.expr_parses_to("NOT NULL NOTNULL", "NOT NULL");
 }
 
+#[test]
+fn test_odbc_time_date_timestamp_support() {
```
```suggestion
fn parse_odbc_time_date_timestamp() {
```

iffyio:
@etgarperets could you take a look at resolving the conflicts on the branch when you get the chance?

Issue #1952, opened 2025-07-17
Release sqlparser 0.56.1  (backport/fix release)


crepererum:
Since DataFusion has fallen a bit behind, having a stable intermediate step which we could use before jumping to 0.57 and beyond would be nice. Sadly, we cannot use 0.56 due to #1898. So we would like to have 0.56.0 + #1899 = 0.56.1.

- Also see https://github.com/apache/datafusion/pull/16456 .

alamb:
I will plan to make a release candidate later today or tomorrow.

FYI @iffyio  (No action required on your part, just FYI that I plan to make a patch release)

- Next schedule release is in August: #1886

alamb:
It turns out backporting https://github.com/apache/datafusion-sqlparser-rs/issues/1898 would break other users (it changes the AST and this is not SemVer compatible with current 0.56.0). 

- See https://github.com/apache/datafusion-sqlparser-rs/pull/1954 for details

Thus I can not complete this task as specified and am closing this issue as I don't think it is possible to complete as specified

PR #1899, opened 2025-06-20
Fix `limit` in subqueries

Dimchikkk:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/1898.

The bug was introduced in https://github.com/apache/datafusion-sqlparser-rs/pull/1793 I think.

PR #1954, opened 2025-07-18
[branch-0.56] Fix `limit` in subqueries (#1899)

alamb:
- Part of https://github.com/apache/datafusion-sqlparser-rs/issues/1952

- Backport of https://github.com/apache/datafusion-sqlparser-rs/pull/1899 to 0.56.0 branch

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -729,7 +729,7 @@ pub enum Expr {
     /// `[ NOT ] IN (SELECT ...)`
     InSubquery {
         expr: Box<Expr>,
-        subquery: Box<SetExpr>,
+        subquery: Box<Query>,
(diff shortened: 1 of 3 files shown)


src/ast/mod.rs
```diff
@@ -729,7 +729,7 @@ pub enum Expr {
     /// `[ NOT ] IN (SELECT ...)`
     InSubquery {
         expr: Box<Expr>,
-        subquery: Box<SetExpr>,
+        subquery: Box<Query>,
```
alamb, line 732: 🤔  double checking,  this change it is not semver-compatible as this AST structure is part of the public API

 I don't think we can release this as 0.56.1, we will have to release it in a normal release instead

PR #1957, opened 2025-07-18
Snowflake: Support IDENTIFIER for GRANT ROLE

yoavcloud:


src/ast/mod.rs
```diff
@@ -6623,7 +6623,7 @@ pub enum Action {
     Replicate,
     ResolveAll,
     Role {
-        role: Ident,
+        role: ObjectName,
```
iffyio, line 6626: Could we add a test that uses a qualified name when mentioning the role? e.g. `GRANT ROLE a.b.c to role d.e.f`

iffyio (commented):
@yoavcloud double checking the intent is to add `ObjectName` support rather than `Identifier` as the MR description mentions?

src/ast/mod.rs
yoavcloud: I added it in tests/sqlparser_snowflake.rs using the `IDENTIFIER` function. Qualified role names are not allowed in Snowflake (only database roles can be qualified).

iffyio: Ah right, I confused `Ident` for the `Objectname` enum

iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1960, opened 2025-07-21
Snowflake: Add support for `CREATE DYNAMIC TABLE`

yoavcloud:
Added support for `CREATE DYNAMIC TABLE` in Snowflake:
1. Extended `CreateTableBuilder` with new options
2. Removed the `CreateTableBuilder::validate_schema_info` function and related tests because, in my opinion, it introduces semantic validation, which is out of scope for the parser.
3. As so many options exist for creating tables, we should evaluate whether specifying them explicitly is robust enough, or a generic parsing of options is a better alternative. Food for thought.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -30,7 +30,14 @@ use sqlparser_derive::{Visit, VisitMut};
 
 use crate::ast::value::escape_single_quote_string;
 use crate::ast::{
-    display_comma_separated, display_separated, ArgMode, CommentDef, CreateFunctionBody, CreateFunctionUsing, CreateTableLikeKind, CreateTableOptions, DataType, Expr, FileFormat, FunctionBehavior, FunctionCalledOnNull, FunctionDeterminismSpecifier, FunctionParallel, HiveDistributionStyle, HiveFormat, HiveIOFormat, HiveRowFormat, Ident, InitializeKind, MySQLColumnPosition, ObjectName, OnCommit, OneOrManyWithParens, OperateFunctionArg, OrderByExpr, ProjectionSelect, Query, RefreshModeKind, RowAccessPolicy, SequenceOptions, Spanned, SqlOption, StorageSerializationPolicy, TableVersion, Tag, Value, ValueWithSpan, WrappedCollection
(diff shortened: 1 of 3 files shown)


tests/sqlparser_snowflake.rs
```diff
@@ -1104,6 +1066,56 @@ fn parse_sf_create_table_or_view_with_dollar_quoted_comment() {
     );
 }
 
+#[test]
+fn parse_create_dynamic_table() {
+    snowflake().verified_stmt(r#"CREATE OR REPLACE DYNAMIC TABLE my_dynamic_table TARGET_LAG='20 minutes' WAREHOUSE=mywh AS SELECT product_id, product_name FROM staging_table"#);
+    snowflake()
+        .parse_sql_statements(
```
iffyio: For the usages of `parse_sql_statements` can we instead use `snowflake().one_statement_parses_to(sql, canonical)` in the cases that `verified_stmt` isnt applicable?

tests/sqlparser_snowflake.rs
```diff
@@ -1104,6 +1066,56 @@ fn parse_sf_create_table_or_view_with_dollar_quoted_comment() {
     );
 }
 
+#[test]
+fn parse_create_dynamic_table() {
```
iffyio, line 1070: Can we include scenarios for the `BEFORE` and `RefreshMode` variants? those don't seem to be covered by the existing tests?

tests/sqlparser_snowflake.rs
```diff
@@ -1104,6 +1066,59 @@ fn parse_sf_create_table_or_view_with_dollar_quoted_comment() {
...
+        " INITIALIZE=ON_SCHEDULE",
+        " REQUIRE USER",
+        " AS SELECT product_id, product_name FROM staging_table"
+    ));
+
+    snowflake().verified_stmt(concat!(
+        "CREATE DYNAMIC TABLE my_dynamic_table",
+        " TARGET_LAG='DOWNSTREAM'",
+        " WAREHOUSE=mywh",
+        " REFRESH_MODE=AUTO",
```
iffyio, line 1115: Can we add similar coverage for `REFRESH_MODE=FULL` and `REFRESH_MODE=INCREMENTAL`?

PR #1961, opened 2025-07-21
Snowflake: Support `CREATE VIEW myview IF NOT EXISTS`

etgarperets:
…, also added ability to write view name before if not exists in snowflake as it is implemented, replaced dialect of with trait functions

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3263,7 +3263,16 @@ pub enum Statement {
         materialized: bool,
         /// View name
         name: ObjectName,
-        // Name IF NOT EXIST instead of IF NOT EXIST name
+        /// If `if_not_exists` is true, this flag is set to true if the view name comes before the `IF NOT EXISTS` clause.
+        /// Example:
+        /// ```sql
+        /// CREATE VIEW myview IF NOT EXISTS AS SELECT 1`
+        ///  ```
+        /// Otherwise, the flag is set to false if the view name comes after the clause
+        /// Example:
+        /// ```sql
+        /// CREATE VIEW IF NOT EXISTS myview AS SELECT 1`
+        ///  ```
         name_before_not_exists: bool,
         columns: Vec<ViewColumnDef>,
         query: Box<Query>,
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5769,22 +5769,14 @@ impl<'a> Parser<'a> {
         let materialized = self.parse_keyword(Keyword::MATERIALIZED);
         self.expect_keyword_is(Keyword::VIEW)?;
         let allow_unquoted_hyphen = dialect_of!(self is BigQueryDialect);
-        let mut if_not_exists = false;
-        let name: ObjectName;
-        let mut name_before_not_exists = false;
-        if self.peek_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]) {
-            // Possible syntax -> ... IF NOT EXISTS <name>
-            if_not_exists = self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
-            name = self.parse_object_name(allow_unquoted_hyphen)?;
-        } else {
-            // Possible syntax -> ... <name> IF NOT EXISTS
-            // Supported by snowflake but is undocumented
-            name = self.parse_object_name(allow_unquoted_hyphen)?;
-            if self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]) {
-                if_not_exists = true;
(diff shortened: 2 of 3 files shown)


tests/sqlparser_common.rs
```diff
@@ -16183,3 +16190,21 @@ fn test_identifier_unicode_start() {
...
+fn parse_create_view_if_not_exists() {
+    let sql = "CREATE VIEW IF NOT EXISTS v AS SELECT 1";
+    let dialects = TestedDialects::new(vec![
+        Box::new(SnowflakeDialect {}),
+        Box::new(GenericDialect {}),
+        Box::new(SQLiteDialect {}),
+        Box::new(BigQueryDialect {}),
+    ]);
+    let _ = dialects.verified_stmt(sql);
+    let sql = "CREATE VIEW v IF NOT EXISTS AS SELECT 1";
```
iffyio, line 16394: From a quick look I couldn't find which dialects in practice support this syntax? Do you have a link to documentation for one such dialect?

related the MR not super sure it would be necessary to have this behavior covered by a flag and the parser can always look to accept `CREATE VIEW IF NOT EXISTS v`, as well as  the other syntax if its relevant)

etgarperets: You're right, I couldn't find official documentation for this syntax, though we have confirmed it works in practice on Snowflake.

I'm happy to remove the dialect flag and update the parser to support both forms for all dialects. Let me know if you'd like me to proceed with that change.

iffyio: Ah yeah in that case I think we can remove the dialect flags, not sure that we need to support both forms if `CREATE VIEW v IF NOT EXISTS` is not supported by any dialect

etgarperets: It's supported in snowflake, but not officially documented.

iffyio: Ah alright, in that case yeah I think we could skip the dialect methods entirely and let the parser accept either format if they show up? Also we can add a comment flagging that the second format is supported by snowflake but is undocumented

src/parser/mod.rs
```diff
@@ -5768,12 +5768,25 @@ impl<'a> Parser<'a> {
...
+            name = self.parse_object_name(allow_unquoted_hyphen)?;
+            if self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]) {
+                if_not_exists = true;
+                name_before_not_exists = true;
+            }
+        }
         // Many dialects support `OR ALTER` right after `CREATE`, but we don't (yet).
         // ANSI SQL and Postgres support RECURSIVE here, but we don't support it either.
-        let allow_unquoted_hyphen = dialect_of!(self is BigQueryDialect);
-        let name = self.parse_object_name(allow_unquoted_hyphen)?;
```
iffyio, line 5776: Would it work to only put something like this line after we've parsed? `name`
```rust
let name_before_not_exists = !if_not_exists && self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
let if_not_exists = if_not_exists || name_before_not_exists;
```
thinking if so that would let us avoid the if/else and `mut`

tests/sqlparser_common.rs
```diff
@@ -16377,3 +16384,11 @@ fn parse_drop_stream() {
     }
     verified_stmt("DROP STREAM IF EXISTS s1");
 }
+
+#[test]
+fn parse_create_view_if_not_exists() {
+    let sql: &'static str = "CREATE VIEW IF NOT EXISTS v AS SELECT 1";
+    let _ = all_dialects().verified_stmt(sql);
+    let sql = "CREATE VIEW v IF NOT EXISTS AS SELECT 1";
+    let _ = all_dialects().verified_stmt(sql);
```
iffyio, line 16395: Can we add a test case for `CREATE VIEW IF NOT EXISTS AS SELECT 1` to verify how the parser behaves in that scenario?

src/ast/mod.rs
```diff
@@ -3263,6 +3263,8 @@ pub enum Statement {
         materialized: bool,
         /// View name
         name: ObjectName,
+        // Name IF NOT EXIST instead of IF NOT EXIST name
```
```suggestion
        /// If `if_not_exists` is true, this flag is set to true if the view name comes before the `IF NOT EXISTS` clause.
        /// Example:
        /// ```sql
        /// CREATE VIEW myview IF NOT EXISTS AS SELECT 1`
        ///  ```
        /// Otherwise, the flag is set to false if the view name comes after the clause
        /// Example:
        /// ```sql
        /// CREATE VIEW IF NOT EXISTS myview AS SELECT 1`
        ///  ```
```
we could say something descriptive like this?

PR #1965, opened 2025-07-22
MySQL: Support `ALTER TABLE RENAME AS`

altmannmarcelo:
Add support for RENAME AS in ALTER TABLE. This is a MySQL extension.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -33,11 +33,11 @@ use crate::ast::{
     display_comma_separated, display_separated, ArgMode, CommentDef, CreateFunctionBody,
     CreateFunctionUsing, DataType, Expr, FunctionBehavior, FunctionCalledOnNull,
     FunctionDeterminismSpecifier, FunctionParallel, Ident, IndexColumn, MySQLColumnPosition,
-    ObjectName, OperateFunctionArg, OrderByExpr, ProjectionSelect, SequenceOptions, SqlOption, Tag,
-    Value, ValueWithSpan,
+    ObjectName, OperateFunctionArg, OrderByExpr, ProjectionSelect, SequenceOptions, Spanned,
+    SqlOption, Tag, Value, ValueWithSpan,
 };
 use crate::keywords::Keyword;
-use crate::tokenizer::Token;
+use crate::tokenizer::{Span, Token};
 
 /// ALTER TABLE operation REPLICA IDENTITY values
 /// See [Postgres ALTER TABLE docs](https://www.postgresql.org/docs/current/sql-altertable.html)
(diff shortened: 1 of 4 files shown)


src/ast/ddl.rs
```diff
@@ -264,6 +264,7 @@ pub enum AlterTableOperation {
     },
     /// `RENAME TO <table_name>`
     RenameTable {
+        to_keyword: bool,
         table_name: ObjectName,
```
iffyio: Could we introduce a wrapper enum to encode this new alternative behavior? thinking something like
```rust
enum RenameTableNameKind {
    As(ObjectName),
    To(ObjectName)
}
RenameTable { table_name: RenameTableNameKind }
```

altmannmarcelo: Done.

src/ast/ddl.rs
```diff
@@ -2524,3 +2524,31 @@ impl fmt::Display for CreateConnector {
...
 }
+
+/// `RenameTableNameKind` is the kind used in `ALTER TABLE _ RENAME TO/AS _`.
+/// Used in MySQL.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum RenameTableNameKind {
+    As(ObjectName),
+    To(ObjectName),
```
iffyio, line 2538:
```suggestion
    /// `ALTER TABLE foo RENAME as bar`
    As(ObjectName),
    /// `ALTER TABLE foo RENAME to bar`
    To(ObjectName),
```

src/ast/ddl.rs
```diff
@@ -2524,3 +2524,31 @@ impl fmt::Display for CreateConnector {
         Ok(())
     }
 }
+
+/// `RenameTableNameKind` is the kind used in `ALTER TABLE _ RENAME TO/AS _`.
+/// Used in MySQL.
```
```suggestion
/// `RenameTableNameKind` is the kind used in an `ALTER TABLE _ RENAME` statement.
```

src/ast/ddl.rs
```diff
@@ -2524,3 +2524,31 @@ impl fmt::Display for CreateConnector {
...
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum RenameTableNameKind {
+    As(ObjectName),
```
iffyio, line 2537: Can we add a link to the mysql docs, flagging that its the one that supports the `AS` syntax?

src/ast/ddl.rs
altmannmarcelo: Done

src/ast/ddl.rs
altmannmarcelo: Done

iffyio (approved):
LGTM! Thanks @altmannmarcelo!
cc @alamb

PR #1967, opened 2025-07-23
Redshift: CREATE TABLE ... (LIKE ..)

yoavcloud:
Add support for Redshift `CREATE TABLE new_table (LIKE parent_table [ { INCLUDING | EXCLUDING } DEFAULTS ])`


src/ast/mod.rs
```diff
@@ -10125,6 +10125,63 @@ impl fmt::Display for MemberOf {
...
+/// '''sql
+/// CREATE TABLE new (LIKE old ...)
+/// '''
+/// [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CreateTableLikeKind {
+    Parenthesized(CreateTableLike),
+    NotParenthesized(CreateTableLike),
```
```suggestion
    Plain(CreateTableLike),
```
Thinking we can do similar to e.g. `CreateTableOptions`, so that `NotParenthesized` doesn't become ambiguous if there happens to be a third variant in the future

```suggestion
/// Specifies how to create a new table based on an existing table's schema.
///
/// '''sql
/// CREATE TABLE new LIKE old ...
/// '''
#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
pub enum CreateTableLikeKind {
    /// '''sql
    /// CREATE TABLE new (LIKE old ...)
    /// '''
    /// [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html)
    Parenthesized(CreateTableLike),
    /// '''sql
    /// CREATE TABLE new LIKE old ...
    /// '''
    /// [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-table#label-create-table-like)
    /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_like)
    NotParenthesized(CreateTableLike),
```
Thinking something like this to document each variant with its example?

src/dialect/mod.rs
```diff
@@ -1136,6 +1136,25 @@ pub trait Dialect: Debug + Any {
...
+    /// '''
+    /// [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-table#label-create-table-like)
+    /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_like)
+    ///
+    /// Parenthesized:
+    /// '''sql
+    /// CREATE TABLE new (LIKE old ...)
+    /// '''
+    /// [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html)
+    fn supports_create_table_like_in_parens(&self) -> bool {
```
```suggestion
    fn supports_create_table_like_parenthesized(&self) -> bool {
```

src/parser/mod.rs
```diff
@@ -7220,8 +7220,35 @@ impl<'a> Parser<'a> {
         // Clickhouse has `ON CLUSTER 'cluster'` syntax for DDLs
         let on_cluster = self.parse_optional_on_cluster()?;
 
-        let like = if self.parse_keyword(Keyword::LIKE) || self.parse_keyword(Keyword::ILIKE) {
-            self.parse_object_name(allow_unquoted_hyphen).ok()
+        // Try to parse `CREATE TABLE new (LIKE old [{INCLUDING | EXCLUDING} DEFAULTS])` or `CREATE TABLE new LIKE old`
+        let like = if self.dialect.supports_create_table_like_in_parens()
```
iffyio: Can we pull the logic out to a function like `maybe_parse_create_table_like()`? since it now adds a bit of code to the existing function

PR #1969, opened 2025-07-23
feat: support multi value columns and aliases in unpivot

chenkovsky:
for example, spark supports 

```
SELECT * FROM sales_quarterly
    UNPIVOT EXCLUDE NULLS (
        (first_quarter, second_quarter)
        FOR half_of_the_year IN (
            (q1, q2) AS H1,
            (q3, q4) AS H2
        )
    );
```

https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-unpivot.html


src/ast/query.rs
```diff
@@ -1351,9 +1392,9 @@ pub enum TableFactor {
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/unpivot>.
     Unpivot {
         table: Box<TableFactor>,
-        value: Ident,
+        value: Vec<Ident>,
```
chenkovsky: should we use expr::Identifier ? otherwise we cannot visit this ident by vistor.

tests/sqlparser_common.rs
```diff
@@ -11022,6 +11022,67 @@ fn parse_unpivot_table() {
         verified_stmt(sql_unpivot_include_nulls).to_string(),
         sql_unpivot_include_nulls
     );
+
+    let sql_unpivot_with_alias = concat!(
```
iffyio, line 11020: @chenkovsky do you know or would you be able to check if this feature is supported by any of the dialects that the parser currently supports?

chenkovsky: This comes from spark.

iffyio: Ah I see, spark dialect isn't a supported by the parser currently so I'm not sure that we can merge this changes without repercussions later on e.g. if we hit scenarios with ambiguous syntax

chenkovsky: Spark adds some features based on Hive. We can regard it as Hive dialect.

chenkovsky: also databricks supports https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-select-unpivot

iffyio: Ah cool yeah makes sense to regard it as hive or databricks in that case

tests/sqlparser_common.rs
```diff
@@ -11022,6 +11022,67 @@ fn parse_unpivot_table() {
...
+                "SELECT * FROM sales AS s ",
+                "UNPIVOT INCLUDE NULLS (quantity FOR quarter IN (Q1 AS Quater1, Q2 AS Quater2, Q3 AS Quater3, Q4 AS Quater4)) AS u (product, quarter, quantity)"
+            );
```
iffyio: it look like the formatting is a bit off here with cargo fmt, could you manually unindent these lines?

tests/sqlparser_common.rs
```diff
@@ -11022,6 +11022,67 @@ fn parse_unpivot_table() {
...
+        );
+        assert_eq!(*value, vec![Ident::new("quantity")]);
+    }
+
+    assert_eq!(
+        verified_stmt(sql_unpivot_with_alias).to_string(),
+        sql_unpivot_with_alias
+    );
+
+    let sql_unpivot_with_alias = concat!(
```
iffyio: this has the same name as the preceeding `sql_unpivot_with_alias` scenario, but testing different aspects, do we need to rename this accordingly?

src/ast/query.rs
iffyio: Hmm yeah I think repr wise we could represent this as an arbitrary `Expr`, it would cover both the single_value and multi_value scenarios. Then similarly, we change to `columns: Vec<ExprWithAlias>`, which would additionally cover the [fully qualified `column_name` variant](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-select-unpivot#parameters)

src/ast/query.rs
```diff
@@ -1351,9 +1392,9 @@ pub enum TableFactor {
     /// See <https://docs.snowflake.com/en/sql-reference/constructs/unpivot>.
```
iffyio, line 1351: Can we include the link to the databricks documentation here as well?

src/parser/mod.rs
```diff
@@ -13882,11 +13892,25 @@ impl<'a> Parser<'a> {
...
+                    self.parse_parenthesized_column_list(Mandatory, false)?
+                        .into_iter()
+                        .map(Expr::Identifier)
+                        .collect(),
+                )
+            }
+            _ => {
+                // single value column unpivot
+                Expr::Identifier(self.parse_identifier()?)
+            }
```
iffyio: Can we call `self.parse_expr()` and `self.parse_expr_with_alias()` transparently instead?

chenkovsky: parse_expr may accept some invalid sql.

iffyio: Ah yes I think it would be ok for the parser to be permissive in this case, downstream crates are expected to validate the AST further if needed.
Related can we add a test case that uses a [fully qualified column name](https://github.com/apache/datafusion-sqlparser-rs/pull/1969#discussion_r2239416734)? e.g. `FOR half_of_the_year IN ((foo.bar, bar.baz) AS x)` - if I understand the intent correctly I think the parser is supposed to successfully parse that whereas with the current impl it is unable to

src/parser/mod.rs
```diff
@@ -10810,6 +10810,16 @@ impl<'a> Parser<'a> {
...
     }
 
+    pub fn parse_parenthesized_columns_with_alias_list(
+        &mut self,
+        optional: IsOptional,
+        allow_empty: bool,
+    ) -> Result<Vec<ExprWithAlias>, ParserError> {
+        self.parse_parenthesized_column_list_inner(optional, allow_empty, |p| {
+            p.parse_expr_with_alias()
+        })
```
iffyio: Can we inline this function since its only a couple lines and used once?

iffyio (commented):
Thanks @chenkovsky! Left a minor comment otherwise the changes look good to me

iffyio (approved):
LGTM! Thanks @chenkovsky!
cc @alamb

PR #1970, opened 2025-07-23
feat: support multiple value for pivot

chenkovsky:
https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-pivot.html

currently this sql is not supported.
```
SELECT * FROM person
    PIVOT (
        SUM(age) AS a, AVG(class) AS c
        FOR (name, age) IN (('John', 30) AS c1, ('Mike', 40) AS c2)
    );
```

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -1336,7 +1336,7 @@ pub enum TableFactor {
     Pivot {
         table: Box<TableFactor>,
         aggregate_functions: Vec<ExprWithAlias>, // Function expression
-        value_column: Vec<Expr>, // Expr is a identifier or a compound identifier
+        value_column: Vec<Expr>,                 // Expr is a identifier or a compound identifier
         value_source: PivotValueSource,
         default_on_null: Option<Expr>,
         alias: Option<TableAlias>,
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -1336,7 +1336,7 @@ pub enum TableFactor {
     Pivot {
         table: Box<TableFactor>,
         aggregate_functions: Vec<ExprWithAlias>, // Function expression
-        value_column: Vec<Ident>,
+        value_column: Vec<Expr>, // Expr is a identifier or a compound identifier
         value_source: PivotValueSource,
         default_on_null: Option<Expr>,
         alias: Option<TableAlias>,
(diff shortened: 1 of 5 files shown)


src/ast/query.rs
```diff
@@ -1336,7 +1336,7 @@ pub enum TableFactor {
     Pivot {
         table: Box<TableFactor>,
         aggregate_functions: Vec<ExprWithAlias>, // Function expression
-        value_column: Vec<Ident>,
+        value_column: Vec<Expr>,                 // Expr is a identifier or a compound identifier
```
```suggestion
        value_column: Vec<Expr>,
```

tests/sqlparser_common.rs
```diff
@@ -10922,6 +10925,15 @@ fn parse_pivot_table() {
...
     );
+
+    let sql_with_multiple_value_column = concat!(
+        "SELECT * FROM person ",
+        "PIVOT(SUM(age) AS a, AVG(class) AS c FOR (name, age) IN (('John', 30) AS c1, ('Mike', 40) AS c2))"
+    );
+    assert_eq!(
+        verified_stmt(sql_with_multiple_value_column).to_string(),
+        sql_with_multiple_value_column
+    );
```
iffyio: hmm this assertion is probably not needed, I think verified_stmt already does the same assertion

src/parser/mod.rs
```diff
@@ -13828,7 +13840,13 @@ impl<'a> Parser<'a> {
...
         let aggregate_functions = self.parse_comma_separated(Self::parse_aliased_function_call)?;
         self.expect_keyword_is(Keyword::FOR)?;
-        let value_column = self.parse_period_separated(|p| p.parse_identifier())?;
+        let value_column = if self.peek_token_ref().token == Token::LParen {
+            self.parse_parenthesized_compound_identifier_list(Mandatory, false)?
+        } else {
+            vec![Expr::CompoundIdentifier(
+                self.parse_period_separated(|p| p.parse_identifier())?,
+            )]
+        };
```
iffyio: similar to unpivot can we call parse_expr directly here?

chenkovsky: no, we cannot do it here. `(name, age) IN (('John', 30) AS c1, ('Mike', 40) AS c2)` will be parsed as an expr.

iffyio: Ah right that makes sense, I think we can do something this instead to reuse existing capabilities of the parser?
```rust
let value_column = if self.peek_token_ref().token == Token::LParen {
    self.parse_parenthesized_column_list_inner(Mandatory, false, |p| {
         p.parse_subexpr(self.dialect.prec_value(Precedence::Between))
     })?
} else {
    vec![self.parse_subexpr(self.dialect.prec_value(Precedence::Between))?]
};
```

tests/sqlparser_common.rs
```diff
@@ -11143,7 +11146,7 @@ fn parse_pivot_unpivot_table() {
                 expr: call("sum", [Expr::Identifier(Ident::new("population"))]),
                 alias: None
             }],
-            value_column: vec![Ident::new("year")],
+            value_column: vec![Expr::CompoundIdentifier(vec![Ident::new("year")])],
```
iffyio: hmm heads up that this representation wouldn't be ideal - a compound identifier would be expected to have more than one ident in it

tests/sqlparser_common.rs
```diff
@@ -10875,7 +10875,10 @@ fn parse_pivot_table() {
                 expected_function("b", Some("t")),
```
iffyio, line 10875: can we add tests to demonstrate the new behavior? it seems like those are currently lacking in the PR?

PR #1971, opened 2025-07-23
feat: support datetime_field as expr for bigquery

chenkovsky:
bigquery supports 
```
SELECT DATETIME_TRUNC(CURRENT_DATETIME, WEEK(MONDAY))
```
`WEEK(MONDAY)` should be parsed to datetime_field

### DIFF
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -2566,3 +2566,101 @@ fn test_struct_trailing_and_nested_bracket() {
         )
     );
 }
+
+#[test]
+fn test_datetime_granularity() {
+    let stmt = bigquery().verified_stmt(concat!(
+        "SELECT ",
+        "DATE_TRUNC(CURRENT_DATE, DAY), ",
+        "DATE_TRUNC(CURRENT_DATE, WEEK(MONDAY)) ",
+        "FROM my_table",
+    ));
+    match stmt {
+        Statement::Query(query) => {
+            let body = query.body.as_ref();
+            match body {
+                SetExpr::Select(select) => {
+                    let projection = &select.projection;
+                    assert_eq!(
+                        projection[0],
+                        SelectItem::UnnamedExpr(Expr::Function(Function {
+                            name: ObjectName(vec![ObjectNamePart::Identifier(Ident::new(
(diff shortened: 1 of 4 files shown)


tests/sqlparser_bigquery.rs
```diff
@@ -2566,3 +2566,101 @@ fn test_struct_trailing_and_nested_bracket() {
...
 }
+
+#[test]
+fn test_datetime_granularity() {
+    let stmt = bigquery().verified_stmt(concat!(
+        "SELECT ",
+        "DATE_TRUNC(CURRENT_DATE, DAY), ",
+        "DATE_TRUNC(CURRENT_DATE, WEEK(MONDAY)) ",
+        "FROM my_table",
+    ));
```
iffyio, line 2577: double checking the intent of the changes: my understanding is that currently given `SELECT DATE_TRUNC(CURRENT_DATE, WEEK(MONDAY)`, `WEEK(MONDAY)` should be parsed as a function with name `WEEK` and having a single identifier `MONDAY` as argument? Is that indeed the case (vs e.g the parser not being able to handle such a statement)

chenkovsky: currently,  DAY will be parsed as ident, week will be parsed as function. I think this is flawed.
1. when transpiling bigquery sql to another dialect, `DATE_TRUNC(CURRENT_DATE, DAY)` should be transpiled to `DATE_TRUNC(CURRENT_DATE,'day')`, but currently, it will be transpiled into `DATE_TRUNC(CURRENT_DATE, DAY)`.
2. when compiling sql to logical plan, planner will try to find column called DAY, MONDAY, and function called WEEK. we should tell planner this is not function or column. I think we should convert it into str when planning.

iffyio: Ah I see, I think the current behavior of having it parsed as a function is preferable since the construct is syntatically a function -  there's quite a lot of such special scenarios across dialects and it would likely be unwieldy to cover them consistently and I think out of scope for the parser as well

chenkovsky: thanks. if it's out of scope， i will close this pr. it can also be solved with a postprocessor.

PR #1973, opened 2025-07-24
Snowflake: DROP STREAM

yoavcloud:


iffyio (approved):
LGTM! Thanks @yoavcloud!
cc @alamb

PR #1974, opened 2025-07-24
Add support for `SHOW CHARSET`

etgarperets:


src/ast/mod.rs
```diff
@@ -3704,6 +3704,20 @@ pub enum Statement {
...
+    // ```
+    // where:
+    // like_or_where: {
+    //     LIKE 'pattern'
+    //   | WHERE expr
+    // }
+    // MySQL specific statement
+    // <https://dev.mysql.com/doc/refman/8.4/en/show.html#:~:text=SHOW%20%7BCHARACTER%20SET%20%7C%20CHARSET%7D%20%5Blike_or_where%5D>
+    ShowCharset {
+        is_shorthand: bool,
```
iffyio: Could we add a description what this flag does?

src/ast/mod.rs
```diff
@@ -3704,6 +3704,20 @@ pub enum Statement {
...
+        filter: Option<ShowStatementFilter>,
+    },
```
iffyio: Can we switch to using a named struct syntax for this statement? [See ref](https://github.com/apache/datafusion-sqlparser-rs/issues/1204) we're trying to move away from the anonymous struct syntax ideally.

We could have something like?
```rust
struct ShowCharSet { ... }
Statement::ShowCharSet(ShowCharSet{...})
```

src/ast/mod.rs
```diff
@@ -3704,6 +3704,20 @@ pub enum Statement {
...
+    // ```sql
+    // SHOW {CHARACTER SET | CHARSET} [like_or_where]
+    // ```
+    // where:
+    // like_or_where: {
+    //     LIKE 'pattern'
+    //   | WHERE expr
+    // }
+    // MySQL specific statement
+    // <https://dev.mysql.com/doc/refman/8.4/en/show.html#:~:text=SHOW%20%7BCHARACTER%20SET%20%7C%20CHARSET%7D%20%5Blike_or_where%5D>
```
```suggestion
    // 
 [MySQL]: https://dev.mysql.com/doc/refman/8.4/en/show.html#:~:text=SHOW%20%7BCHARACTER%20SET%20%7C%20CHARSET%7D%20%5Blike_or_where%5D
```

Or something similar, essentially dropping the mention of it being mysql specific, since that can change in the future or other existing dialects might also support it in some form

src/ast/mod.rs
```diff
@@ -3704,6 +3704,20 @@ pub enum Statement {
...
         show_options: ShowStatementOptions,
     },
+    // ```sql
+    // SHOW {CHARACTER SET | CHARSET} [like_or_where]
+    // ```
+    // where:
+    // like_or_where: {
+    //     LIKE 'pattern'
+    //   | WHERE expr
+    // }
```
```suggestion
    // SHOW {CHARACTER SET | CHARSET}
    // ```
```
Thinking a brief example would be enough. For detailed syntax readers can follow the documentation

tests/sqlparser_mysql.rs
```diff
@@ -4143,3 +4143,11 @@ fn parse_json_member_of() {
         _ => panic!("Unexpected statement {stmt}"),
     }
 }
+
+#[test]
+fn parse_show_charset() {
+    let _ = mysql().verified_stmt("SHOW CHARACTER SET");
```
iffyio: Since this adds a new node to the AST, can we assert one of the scenarios that the AST looks as expected? [For example](https://github.com/apache/datafusion-sqlparser-rs/blob/4a533132b58090512c03337111925f8bc7098e97/tests/sqlparser_mysql.rs#L88-L98)

src/ast/mod.rs
```diff
@@ -9792,6 +9799,32 @@ impl fmt::Display for ShowStatementIn {
     }
 }
 
+/// A Show Charset statement
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ShowCharset {
+    /// The statement can be written as SHOW CHARSET or SHOW CHARACTER SET
```
```suggestion
    /// The statement can be written as `SHOW CHARSET` or `SHOW CHARACTER SET`
```

iffyio (commented):
Thanks @etgarperets! Left a minor doc comment otherwise I think this looks good!

PR #1975, opened 2025-07-24
fix: begin statement for bigquery

chenkovsky:
big query supports,

```
BEGIN [TRANSACTION];
```

but currently parse_statement is overwritten for bigquery, and this statement is not supported.

### DIFF
--- a/src/dialect/bigquery.rs
+++ b/src/dialect/bigquery.rs
@@ -19,6 +19,7 @@ use crate::ast::Statement;
 use crate::dialect::Dialect;
 use crate::keywords::Keyword;
 use crate::parser::{Parser, ParserError};
+use crate::tokenizer::Token;
 
 /// These keywords are disallowed as column identifiers. Such that
 /// `SELECT 5 AS <col> FROM T` is rejected by BigQuery.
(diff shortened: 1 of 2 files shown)


tests/sqlparser_bigquery.rs
```diff
@@ -2566,3 +2566,14 @@ fn test_struct_trailing_and_nested_bracket() {
         )
     );
 }
+
+// https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#begin_transaction
```
iffyio:
```suggestion
```

src/dialect/bigquery.rs
```diff
@@ -47,6 +48,15 @@ pub struct BigQueryDialect;
...
         if parser.parse_keyword(Keyword::BEGIN) {
+            if parser.peek_keyword(Keyword::TRANSACTION) {
+                parser.prev_token();
+                return None;
+            }
+            let peek_token = parser.peek_token_ref();
+            if peek_token.token == Token::SemiColon || peek_token.token == Token::EOF {
+                parser.prev_token();
+                return None;
+            }
```
iffyio: Can we merge both blocks with something like `if parse.parse_keyword(BEGIN) || peek_token.token == SemiColon || peek_token.token == EOF`

PR #1976, opened 2025-07-24
feat: support export data for bigquery

chenkovsky:


src/ast/mod.rs
```diff
@@ -4355,6 +4355,14 @@ pub enum Statement {
...
     /// See [ReturnStatement]
     Return(ReturnStatement),
+    /// Export data statement
+    ///
+    /// Example:
+    /// ```sql
+    /// EXPORT DATA OPTIONS(uri='gs://bucket/folder/*', format='PARQUET', overwrite=true) AS
+    /// SELECT field1, field2 FROM mydataset.table1 ORDER BY field1 LIMIT 10
+    /// ```
+    ExportData(ExportData),
```
iffyio, line 4366: Could we add a link to the bigquery docs where the syntax comes from?

src/parser/mod.rs
```diff
@@ -645,6 +645,7 @@ impl<'a> Parser<'a> {
                 Keyword::COMMENT if self.dialect.supports_comment_on() => self.parse_comment(),
                 Keyword::PRINT => self.parse_print(),
                 Keyword::RETURN => self.parse_return(),
+                Keyword::EXPORT => self.parse_export(),
```
iffyio: Could we do [similar to here](https://github.com/apache/datafusion-sqlparser-rs/blob/1a7e8d969ae509d99f5f9aa1136092da002a77c6/src/parser/mod.rs#L607) where we call `prev_token()` so that the `parse-export()` function is stand alone and able to parse a complete statement?

src/parser/mod.rs
```diff
@@ -16510,6 +16511,27 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_export(&mut self) -> Result<Statement, ParserError> {
```
```suggestion
    fn parse_export_date(&mut self) -> Result<Statement, ParserError> {
```
Could we also add a short description mentioning that this parses a `Statement::ExportData` statement? [For example](https://github.com/apache/datafusion-sqlparser-rs/blob/1a7e8d969ae509d99f5f9aa1136092da002a77c6/src/parser/mod.rs#L659-L661)

tests/sqlparser_bigquery.rs
```diff
@@ -2566,3 +2566,24 @@ fn test_struct_trailing_and_nested_bracket() {
         )
     );
 }
+
+#[test]
+fn test_export() {
```
iffyio: Since this introduces a new statement, can we include an assertion on the AST for one of the statements, checking that the returned AST looks like what we expect? [For example](https://github.com/apache/datafusion-sqlparser-rs/blob/1a7e8d969ae509d99f5f9aa1136092da002a77c6/tests/sqlparser_bigquery.rs#L2203-L2232)


Also could we add coverage for some scenarios like the following?
```
"EXPORT DATA OPTIONS()";
"EXPORT DATA AS SELECT ...";
"EXPORT DATA WITH CONNECTION ... AS SELECT ..."
```

Issue #291, opened 2021-01-24
Parameterized queries


Jasperav:
This fails to parse, but it can be successfully executed by SQLite (after providing a value for the parameter):

```
#[test]
fn go() {
    use sqlparser::dialect::SQLiteDialect;
    use sqlparser::parser::Parser;

    let sql = "select * from test_table where test_column = ?;";

    let dialect = SQLiteDialect {}; // or AnsiDialect, or your own dialect ...

    let ast = Parser::parse_sql(&dialect, sql).unwrap();

    println!("AST: {:#?}", ast);
}
```

Any idea why I get this error message?:

_called `Result::unwrap()` on an `Err` value: ParserError("Expected an expression, found: ?")
thread 'dynamic_queries::go' panicked at 'called `Result::unwrap()` on an `Err` value: ParserError("Expected an expression, found: ?")',_

Dandandan:
Thanks @Jasperav .
In think the `?` is not something that's supported yet.

H2CO3:
I was just going to open an issue about the same thing. Support for other flavors of parameterized queries like `:named`, `@named`, and `$named` would be nice as well.

alex-dukhno:
see discussions in https://github.com/ballista-compute/sqlparser-rs/pull/265

mikoto2000:
I would also be happy to have this feature.
I would like to support the MyBatis's `#{named}` method.

H2CO3:
> see discussions in #265

That issue is closed. Of what use is it? Does that mean that we're decidedly not getting parameter support?

H2CO3:
@git-hulk Hi, and thanks for getting back to this!

The current implementation largely works indeed. Unfortunately, it still seems to have a bug that is blocking in my case. I slightly modified one of the cases in the test suite to reproduce it, and – at least in the case of the SQLite dialect – it appears that `$` placeholders are incorrectly recognized as an identifier. The following code:


```rust
fn main() {
    let ast = sqlparser::parser::Parser::parse_sql(
        &sqlparser::dialect::SQLiteDialect {},
        "SELECT * FROM student WHERE id = $foo",
    );
    dbg!(ast);
}
```

prints the following AST (irrelevant parts excluded for clarity):

```
                        selection: Some(
                            BinaryOp {
                                left: Identifier(
                                    Ident {
                                        value: "id",
                                        quote_style: None,
                                    },
                                ),
                                op: Eq,
                                right: Identifier(
                                    Ident {
[...]

PR #1979, opened 2025-07-27
Fix placeholder spans

xitep:
* Optional support for `:name` as a parameters in queries; initially requested through #291 
* The naming convention chosen here is based on the java persistence api spec (https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#a4401) and should fit in well
* The support here is already at the level of the tokenizer, to get the span of the whole placeholder right (i'm actually interested in replacing the parameter as a whole in the original query)
* in fact, sqlparser's [parser](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/parser/mod.rs#L9591) combines a colon token followed by a word token into a placeholder, but does not adjust the span of the resulting ast node :/ and allows for whitespace between the colon and the word. this is something i wanted to avoid, but i'm don't feel confident to change the parser without breaking already existing functionality (if any) based on this behaviour.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -841,12 +841,6 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if this dialect allow colon placeholders
-    /// e.g. `SELECT :var` (JPA named parameters)
-    fn supports_colon_placeholder(&self) -> bool {
-        false
-    }
-
     /// Does the dialect support with clause in create index statement?
     /// e.g. `CREATE INDEX idx ON t WITH (key = value, key2)`
     fn supports_create_index_with_clause(&self) -> bool {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -841,6 +841,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allow colon placeholders
+    /// e.g. `SELECT :var` (JPA named parameters)
+    fn supports_colon_placeholder(&self) -> bool {
+        false
+    }
+
     /// Does the dialect support with clause in create index statement?
     /// e.g. `CREATE INDEX idx ON t WITH (key = value, key2)`
     fn supports_create_index_with_clause(&self) -> bool {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9584,8 +9584,13 @@ impl<'a> Parser<'a> {
             Token::HexStringLiteral(ref s) => ok_value(Value::HexStringLiteral(s.to_string())),
             Token::Placeholder(ref s) => ok_value(Value::Placeholder(s.to_string())),
             tok @ Token::Colon | tok @ Token::AtSign => {
-                // Not calling self.parse_identifier(false)? because only in placeholder we want to check numbers as idfentifies
-                // This because snowflake allows numbers as placeholders
+                // 1. Not calling self.parse_identifier(false)?
+                //    because only in placeholder we want to check
+                //    numbers as idfentifies.  This because snowflake
+                //    allows numbers as placeholders
+                // 2. Not calling self.next_token() to enforce `tok`
+                //    be followed immediately by a word/number, ie.
+                //    without any whitespace in between
                 let next_token = self.next_token_no_skip().unwrap_or(&EOF_TOKEN).clone();
                 let ident = match next_token.token {
                     Token::Word(w) => Ok(w.into_ident(next_token.span)),
@@ -17493,5 +17498,13 @@ mod tests {
             ),
             canonical,
         );
+    }
+
+    #[test]
+    fn test_placeholder_invalid_whitespace() {
+        for w in ["  ", "/*invalid*/"] {
+            let sql = format!("\nSELECT\n  :{w}fooBar");
+            assert!(Parser::parse_sql(&GenericDialect, &sql).is_err());
+        }
     }
 }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9586,14 +9586,14 @@ impl<'a> Parser<'a> {
             tok @ Token::Colon | tok @ Token::AtSign => {
                 // Not calling self.parse_identifier(false)? because only in placeholder we want to check numbers as idfentifies
                 // This because snowflake allows numbers as placeholders
-                let next_token = self.next_token();
+                let next_token = self.next_token_no_skip().unwrap_or(&EOF_TOKEN).clone();
                 let ident = match next_token.token {
                     Token::Word(w) => Ok(w.into_ident(next_token.span)),
-                    Token::Number(w, false) => Ok(Ident::new(w)),
+                    Token::Number(w, false) => Ok(Ident::with_span(next_token.span, w)),
                     _ => self.expected("placeholder", next_token),
                 }?;
-                let placeholder = tok.to_string() + &ident.value;
-                ok_value(Value::Placeholder(placeholder))
+                Ok(Value::Placeholder(tok.to_string() + &ident.value)
+                    .with_span(Span::new(span.start, ident.span.end)))
             }
             unexpected => self.expected(
                 "a value",
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1521,11 +1521,6 @@ impl<'a> Tokenizer<'a> {
                     match chars.peek() {
                         Some(':') => self.consume_and_return(chars, Token::DoubleColon),
                         Some('=') => self.consume_and_return(chars, Token::Assignment),
-                        Some(c)
-                            if self.dialect.supports_colon_placeholder() && c.is_alphabetic() =>
-                        {
-                            self.tokenize_colon_preceeded_placeholder(chars).map(Some)
-                        }
                         _ => Ok(Some(Token::Colon)),
                     }
                 }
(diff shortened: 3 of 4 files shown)


src/dialect/mod.rs
```diff
@@ -841,6 +841,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allow colon placeholders
+    /// e.g. `SELECT :var` (JPA named parameters)
+    fn supports_colon_placeholder(&self) -> bool {
+        false
+    }
```
iffyio: > in fact, sqlparser's [parser](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/parser/mod.rs#L9591) combines a colon token followed by a word token into a placeholder, but does not adjust the span of the resulting ast node :/ and allows for whitespace between the colon and the word.

@xitep could you elaborate on this part? I'm not sure I followed the intent, is my understanding correct that currently the parser indeed handles e.g. `SELECT :var`? If so what's the requirement for the tokenizer updates?

xitep: hello @iffyio, yes the parser can deal with things like `:var` already, but i'm not sure this is a feature or rather just a side effect of the code pointed out above. here's an example:

```rust
    let q = sqlparser::parser::Parser::new(&sqlparser::dialect::GenericDialect)
        .try_with_sql("select :var")
        .unwrap()
        .parse_query()
        .unwrap();
    let _ = sqlparser::ast::visit_expressions(&q, |expr: &Expr| {
        if let Expr::Value(v) = expr
            && let Value::Placeholder(s) = &v.value
        {
            println!("value: {v:?}");
            println!("placeholder: {s:?})");
        }
        ControlFlow::<()>::Continue(())
    });
// OUTPUT:
// value: ValueWithSpan { value: Placeholder(":var"), span: Span(Location(1,8)..Location(1,9)) }
// placeholder: ":var")
```
note:
1. the span is incorrect, it covers only the colon
2. the placeholder represents `:<word>` (which in fact is what i'm after)

however, you'll get the same output when using `"select :    var"` or even `"select : /*foobar*/   var"` as input :-/ this is _not_ what i want.

so the intention is to make `:<word>` a first level token, a placeholder.

does this help?

src/tokenizer.rs
```diff
@@ -1756,6 +1761,30 @@ impl<'a> Tokenizer<'a> {
         }
     }
 
+    /// Tokenizes an identifier followed immediately after a colon,
+    /// aka named query parameter, e.g. `:name`. The next char of the
+    /// processed char stream is to be an alphabetic - panics otherwise.
+    fn tokenize_colon_preceeded_placeholder(
```
iffyio: I think that it would likely be ideal to use existing logic like `tokenize_identifier_or_keyword` or similar. Also heads up in general for the parser that returning an error when input is invalid etc is preferred over panicking

src/dialect/mod.rs
iffyio: I think that make sense, it seems that the handling of placeholders isn't very consistent by the parser as some are done in tokenizer while others later on. I think it could be good to do a more general hopefully improvement if we feel that ideally placeholders are managed by the tokenizer, the we can parse and represent it consistently/accordingly.

Did a quick pass through existing support, and I can imagine something like the following representation to merge the different representations explicitly, wdyt?
```rust
enum PlaceholderKind {
	/// Example:
	/// ```sql
	/// SELECT ?, ?foo
	/// ```
	SingleQuestionMark,
	/// Example:
	/// ```sql
	/// SELECT @foo
	/// ```
	SingleAt,
	/// Example:
	/// ```sql
	/// SELECT $1, $foo
	/// ```
	SingleDollar,
	/// Example:
	/// ```sql
	/// SELECT :42, :foo
	/// ```
	Colon,
	/// Example:
	/// ```sql
	/// SELECT $$, $$abc$$, $abc$, $$$abc$$$
	/// ```
	DollarEnclosed {
		count: usize
	}
}

Token::Placeholder {
	kind: PlaceholderKind,
	value: Option<String>
}
```
It would also remove the need for this code as well
https://github.com/apache/datafusion-sqlparser-rs/blob/15d8bfea62726464a3065765f5511af7fa1548f8/src/parser/mod.rs#L9637-L9646

xitep: ok, i can try it out and upgrade this PR accordingly. here's one thought to be clarified before i set off with it:

```rust
Token::Placeholder {
	kind: PlaceholderKind,
	value: Option<String>
}
```

this implies to me that "value" would be stripped off of the placeholder demarcation character(s); e.g. for the input `":foo"` -> `value` would read `Some("foo")`, for the input `"$$foo$$"` it would read `Some("foo")`, for the input `"$"` `value` would be `None`, correct?

But then, what would [`sqlparser::ast::Value::Placeholder(..)`](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Value.html#variant.Placeholder) look like? Clients surely might want to know exactly which placeholder style/kind was used; since `sqlparser` supports various formats, some clients might opt in to reject what they don't want their queries to contain.

all in all, it looks to me that in your proposal we want to end up with `value` being the original, full token as identified in the scanned input (which, naturally, is never `None` or empty), therefore:

```
Token::Placeholder {
	kind: PlaceholderKind,
	value: String,  // ~ preferrably &'a str
}
```

any feedback on this thought?

src/tokenizer.rs
xitep: * wrt `tokenize_identifier_or_keyword` ... i guess it would make it quite hard to support the `$` then as part of the identifier (if enabled by `Dialect::supports_dollar_placeholder` https://docs.rs/sqlparser/latest/sqlparser/dialect/trait.Dialect.html#method.supports_dollar_placeholder.) i see the placeholder identifiers distinct (ie. with different restrictions) from other SQL identifiers.
* thanks for the heads-up; the panic here would be a programming error, as that contract says that the caller must make sure ... but I'm fine with the established convention and would make it return an error instead.

src/dialect/mod.rs
iffyio: Yeah that's correct that the input would be stripped into two, the kind and the value as in your examples.

`Value::Placeholder` I imagine would have the same representation mirroring `Token::Placeholder` since the goal becomes that placeholder parsing is handled by the tokenizer  i.e.`Value::Placeholder(Token::Placeholder)`

> which, naturally, is never None or empty

I think we might not always have a variable name, hence the `Option` - sqlite for example uses anonymous `?` in some contexts for placeholders e.g. `WHERE city IN (?, ?, ?, ?)`. In that scenario the representation becomes `Token::Placeholder{kind: PlaceholderKind::SingleQuestionMark, value: None}`

> 	value: String,  // ~ preferrably &'a str

ah heads up I think a reference `&str` wouldn't be feasible here without introducing lifetimes to the Token/Value enums which would be quite an invasive change and likely not worth it.

xitep: wrt ...

```
	DollarEnclosed {
		count: usize
	}
```

i'm seeing "$$foo$$" being treated as a "double quoted string", not as a placeholder. if `dialect.supports_dollar_placeholder() == true)` then `$...` is considered a placeholder, but the dollars at the start and end don't have to be balanced. this make the `DollarEnclosed` variant essentially void.

xitep: alright, after a lot of struggle i came to the conclusion that the approach with first-class placeholder is not feasible in `sqlparser`. After having the tokenizer and the model change ready, [this test](https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_postgres.rs#L2150) (now failing) convinced me, that my concern cannot be resolved at the level of the tokenizer since `sqlparser` needs to be very flexible and determine the meaning only on the level of the parser. you can [have a look here](https://github.com/xitep/datafusion-sqlparser-rs/commit/c97440558c16dedb024ff6a332fc8259b10cec55) at how the change we discussed would have looked like. (this would be the [consequential change](https://github.com/xitep/datafusion-sqlparser-rs/commit/6426bef3bdba7a1257550c85188349e32610eda8) to the parser.)

fortunately, i have found a far smaller change that satisfies my needs and will update this PR.

xitep:
@iffyio would you like to have another look? i changed strategy to the whole problem.

src/parser/mod.rs
```diff
@@ -9586,14 +9586,14 @@ impl<'a> Parser<'a> {
             tok @ Token::Colon | tok @ Token::AtSign => {
                 // Not calling self.parse_identifier(false)? because only in placeholder we want to check numbers as idfentifies
                 // This because snowflake allows numbers as placeholders
-                let next_token = self.next_token();
+                let next_token = self.next_token_no_skip().unwrap_or(&EOF_TOKEN).clone();
```
iffyio, line 9594: If I understand the intent correctly is this saying that e.g. `SELECT :foo` is to be parsed as a placeholder but `SELECT : foo` should not? If so I think it would be good to mention that in a comment. Also can we add a test case to demonstrate this change?

src/parser/mod.rs
```diff
@@ -9586,14 +9586,14 @@ impl<'a> Parser<'a> {
...
                 let ident = match next_token.token {
                     Token::Word(w) => Ok(w.into_ident(next_token.span)),
-                    Token::Number(w, false) => Ok(Ident::new(w)),
+                    Token::Number(w, false) => Ok(Ident::with_span(next_token.span, w)),
                     _ => self.expected("placeholder", next_token),
                 }?;
-                let placeholder = tok.to_string() + &ident.value;
-                ok_value(Value::Placeholder(placeholder))
+                Ok(Value::Placeholder(tok.to_string() + &ident.value)
```
iffyio, line 9600: Can we add a test case for this to avoid future regressions? we have some [span test cases here](https://github.com/apache/datafusion-sqlparser-rs/blob/10cd9e2bb008f71b6fcac5d439b78588b593d210/src/ast/spans.rs#L2365) that we could add to

iffyio (approved):
LGTM! Thanks @xitep!
cc @alamb

xitep:
many thanks for your support @iffyio :bow:

alamb:
> many thanks for your support @iffyio 🙇

I agree @iffyio  is awesome 🙏

PR #1980, opened 2025-07-28
feat: Add `ALTER SCHEMA` support

chenkovsky:


src/ast/mod.rs
```diff
@@ -3381,6 +3381,17 @@ pub enum Statement {
...
+    /// ALTER SCHEMA
+    /// ```
+    /// See [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_schema_collate_statement)
+    AlterSchema {
+        /// Schema name
+        #[cfg_attr(feature = "visitor", visit(with = "visit_relation"))]
+        name: ObjectName,
+        if_exists: bool,
+        operations: Vec<AlterSchemaOperation>,
+    },
```
iffyio: Can we use a [named struct syntax](https://github.com/apache/datafusion-sqlparser-rs/blob/0aaa561287772341ef1bfbee729d7be75e2e359b/src/ast/mod.rs#L4050-L4053) here? e.g. `AlterSchema(AlterSchemaStatement)`

src/parser/mod.rs
```diff
@@ -9241,6 +9243,40 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_alter_schema(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 9358: We can call prev_token before invoking this method, so that it is self contained and able to parse a full `ALTER SCHEMA` statement. Also can we add some documentation for this function, ideally mentioning that it returns a `Statement::AlterSchema` variant

src/parser/mod.rs
```diff
@@ -9241,6 +9243,40 @@ impl<'a> Parser<'a> {
...
+            } else {
+                None
+            };
+            AlterSchemaOperation::AddReplica { replica, options }
+        } else if self.parse_keywords(&[Keyword::DROP, Keyword::REPLICA]) {
+            let replica = self.parse_identifier()?;
+            AlterSchemaOperation::DropReplica { replica }
+        } else {
+            return self.expected_ref(
+                "{SET OPTIONS | SET DEFAULT COLLATE | ADD REPLICA | DROP REPLICA}",
```
```suggestion
                "ALTER SCHEMA operation",
```
Thinking something generic, in order to avoid an maintaining a growing list that may potentially go out of sync

tests/sqlparser_bigquery.rs
```diff
@@ -2806,3 +2806,23 @@ fn test_begin_transaction() {
...
+
+#[test]
+fn test_alter_schema_drop_replica() {
+    bigquery_and_generic().verified_stmt("ALTER SCHEMA mydataset DROP REPLICA 'us'");
+}
+
+#[test]
+fn test_alter_schema_set_options() {
+    bigquery_and_generic().verified_stmt("ALTER SCHEMA mydataset SET OPTIONS (location = 'us')");
+}
```
iffyio: Can we merge these into a single `test_alter_schema()` function?

Also we can move this to the sqlparser_common.rs file since the parser covers the statement it for all dialects (I imagine other dialects have some variant of this statement as well).

Also can we add scenarios for e.g.
```sql
ALTER SCHEMA DROP REPLICA;
ALTER SCHEMA SET OPTIONS ();
ALTER SCHEMA  SET OPTIONS;
```

tests/sqlparser_bigquery.rs
```diff
@@ -2806,3 +2806,23 @@ fn test_begin_transaction() {
...
 }
+
+#[test]
+fn test_alter_schema_default_collate() {
+    bigquery_and_generic().verified_stmt("ALTER SCHEMA mydataset SET DEFAULT COLLATE 'und:ci'");
+}
+
+#[test]
+fn test_alter_schema_add_replica() {
+    bigquery_and_generic().verified_stmt("ALTER SCHEMA mydataset ADD REPLICA 'us'");
```
iffyio, line 2832: from the implementation, `ADD REPLICA` takes in options, can we add test coverage for that behavior?

tests/sqlparser_bigquery.rs
chenkovsky: postgres also has `alter schema`, but it's quite different. https://www.postgresql.org/docs/17/sql-alterschema.html

iffyio: Ah right, yeah we can leave the tests in bigquery file as is

iffyio (approved):
LGTM! Thanks @chenkovsky!
cc @alamb

PR #1984, opened 2025-07-30
Postgres: Support `INTERVAL` data type options

mvzink:
[Postgres] allows extra options for the `INTERVAL` data type; namely fields and subsecond precision. For example `'3 years 1 second'::interval year to month` casts the interval and strips the seconds, and `'1.3333 seconds'::interval(1)` returns `1.3` seconds.

This is supported by adding two optional fields to `DataType::Interval`, along with a new `enum` for the allowed fields.

Note that [MSSQL] also supports similar options, but with more complicated precision syntax, e.g. `INTERVAL HOUR(p) TO SECOND(q)`. This is not implemented in this commit because I don't have a way to test it.

[Postgres]: https://www.postgresql.org/docs/17/datatype-datetime.html
[MSSQL]: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/sql-data-types?view=sql-server-ver17

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -346,7 +346,16 @@ pub enum DataType {
     /// [1]: https://docs.databricks.com/aws/en/sql/language-manual/data-types/timestamp-ntz-type
     TimestampNtz,
     /// Interval type.
-    Interval,
+    Interval {
+        /// [PostgreSQL] fields specification like `INTERVAL YEAR TO MONTH`.
+        ///
+        /// [PostgreSQL]: https://www.postgresql.org/docs/17/datatype-datetime.html
+        fields: Option<IntervalFields>,
+        /// [PostgreSQL] subsecond precision like `INTERVAL HOUR TO SECOND(3)`
+        ///
+        /// [PostgreSQL]: https://www.postgresql.org/docs/17/datatype-datetime.html
+        precision: Option<u64>,
+    },
     /// JSON type.
     JSON,
     /// Binary JSON type.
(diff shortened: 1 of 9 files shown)


src/ast/data_type.rs
```diff
@@ -346,7 +346,11 @@ pub enum DataType {
     /// [1]: https://docs.databricks.com/aws/en/sql/language-manual/data-types/timestamp-ntz-type
     TimestampNtz,
     /// Interval type.
-    Interval,
+    ///
+    /// Optional [Postgres] field list and precision: `interval [ fields ] [ (p) ]`
+    ///
+    /// [Postgres]: https://www.postgresql.org/docs/17/datatype-datetime.html
+    Interval(Option<IntervalFields>, Option<u64>),
```
iffyio: Could we use a struct variant to contain the fields? thinking that would be clearer API wise and easier to document in the cases like this one where we have multiple entries e.g.
```
Interval {
    /// Postgres ...
    fields: Option<IntervalFields>,
    /// Postgres ...
    precision: Option<u64>
}
```

src/dialect/mod.rs
```diff
@@ -1136,6 +1136,13 @@ pub trait Dialect: Debug + Any {
     fn supports_notnull_operator(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports the `INTERVAL` data type with [Postgres]-style options.
```
iffyio, line 1152: Could we add an example expression here? Thinking it would clarify the behavior without having folks consult the pg docs if they don't want to dig deeper into the syntax

src/parser/mod.rs
```diff
@@ -10995,6 +11000,85 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_optional_interval_fields(
```
```suggestion
    fn maybe_parse_optional_interval_fields(
```

src/ast/data_type.rs
```diff
@@ -346,7 +346,14 @@ pub enum DataType {
...
     TimestampNtz,
     /// Interval type.
-    Interval,
+    ///
+    /// Optional [Postgres] field list and precision: `interval [ fields ] [ (p) ]`
+    ///
+    /// [Postgres]: https://www.postgresql.org/docs/17/datatype-datetime.html
+    Interval {
+        fields: Option<IntervalFields>,
+        precision: Option<u64>,
```
iffyio:
```suggestion
    Interval {
       /// `Example syntax`
       ///
       /// [Postgres]: https://www.postgresql.org/docs/17/datatype-datetime.html
        fields: Option<IntervalFields>,
        /// `Example syntax`
       ///
       /// [Postgres]: https://www.postgresql.org/docs/17/datatype-datetime.html
        precision: Option<u64>,
```
Could we move the documentation to the individual properties?

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1985, opened 2025-07-30
MySQL: Allow optional `SIGNED` suffix on integer data types

mvzink:
In MySQL, a data type like `INT(20) SIGNED` is equivalent to `INT(20)`. In other dialects, this may be interpreted differently; e.g. in Postgres, `SELECT 1::integer signed` indicates an alias of "signed". So we parse the optional `SIGNED` suffix only on dialects that allow it (I currently don't know of any other than MySQL).

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -1136,6 +1136,18 @@ pub trait Dialect: Debug + Any {
     fn supports_notnull_operator(&self) -> bool {
         false
     }
+
+    /// Returns true if this dialect allows an optional `SIGNED` suffix after integer data types.
+    ///
+    /// Example:
+    /// ```sql
+    /// CREATE TABLE t (i INT(20) SIGNED);
+    /// ```
+    ///
+    /// Note that this is canonicalized to `INT(20)`.
+    fn supports_data_type_signed_suffix(&self) -> bool {
+        false
+    }
 }
 
 /// This represents the operators for which precedence must be defined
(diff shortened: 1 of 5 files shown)


src/dialect/mod.rs
```diff
@@ -835,6 +835,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allows an optional `SIGNED` suffix after integer data types.
+    fn allow_optional_signed_suffix(&self) -> bool {
```
```suggestion
    fn supports_data_type_signed_suffix(&self) -> bool {
```

src/dialect/mod.rs
```diff
@@ -835,6 +835,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if this dialect allows an optional `SIGNED` suffix after integer data types.
```
```suggestion
    /// Returns true if this dialect allows an optional `SIGNED` suffix after integer data types.
    /// Example: `CREATE TABLE foo (bar INT(20) SIGNED)`
```
Thinking maybe with an example to clarify the syntax

src/parser/mod.rs
```diff
@@ -9852,6 +9852,8 @@ impl<'a> Parser<'a> {
                     if self.parse_keyword(Keyword::UNSIGNED) {
                         Ok(DataType::TinyIntUnsigned(optional_precision?))
                     } else {
+                        let _ = dialect.allow_optional_signed_suffix()
+                            && self.parse_keyword(Keyword::SIGNED);
```
```suggestion
                        if dialect.allow_optional_signed_suffix() {
                            let _ = self.parse_keyword(Keyword::SIGNED);
                        }
```
Thinking we can be explicit with an if statement for these?

tests/sqlparser_mysql.rs
```diff
@@ -1705,6 +1705,51 @@ fn parse_create_table_unsigned() {
     }
 }
 
+#[test]
+fn parse_create_table_signed() {
```
```suggestion
fn parse_signed_data_types() {
```

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

PR #1989, opened 2025-08-01
MySQL: Support comma-separated `CREATE TABLE` options

mvzink:
In [MySQL], options for `CREATE TABLE` following the body can be optionally separated by commas. I'm not aware of any cases where this affects parsing (e.g. eliminating ambiguity or anything), so we just optionally eat comma tokens after each option is parsed.

[MySQL]: https://dev.mysql.com/doc/refman/8.4/en/create-table.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7708,6 +7708,9 @@ impl<'a> Parser<'a> {
 
         while let Some(option) = self.parse_plain_option()? {
             options.push(option);
+            // Some dialects support comma-separated options; it shouldn't introduce ambiguity to
+            // consume it for all dialects.
+            let _ = self.consume_token(&Token::Comma);
         }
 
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -7708,6 +7708,9 @@ impl<'a> Parser<'a> {
 
         while let Some(option) = self.parse_plain_option()? {
             options.push(option);
+            if self.dialect.supports_comma_separated_create_table_options() {
+                let _ = self.consume_token(&Token::Comma);
+            }
```
iffyio: Thinking since the syntax wouldn't conflict with other dialects we can skip the dialect option and have the parser always consume the comma if it shows up?

mvzink: Sure, I see no problem with that.

PR #1990, opened 2025-08-03
Postgres: enhance NUMERIC/DECIMAL parsing to support negative scale

### LINKED ISSUE #1923: Postgres: Support negative scale for `NUMERIC`
When specifying explicit precision and scale for numerics, Postgres allows the ranges `0..=1000` for precision and `-1000..=1000` for scale. Currently, `ExactNumberInfo` uses `u64` to represent scale, and the parser fails on the `-` when trying to parse DDL with a negative scale:

```sql
create table t (x numeric(1000,-10));

Error during parsing: ParserError("Expected: literal int, found: - at Line: 1, Column: 32")
```

IndexSeek:
Closes #1923 

> Beginning in PostgreSQL 15, it is allowed to declare a numeric column with a negative scale.

**Source:** [PostgreSQL Documentation - Numeric Types](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL)

## Changes
- Modified `ExactNumberInfo::PrecisionAndScale` to use `i64` instead of `u64` for scale
- Added `parse_scale_value()` method to handle positive, negative, and explicitly positive scale values
- Updated tests to cover negative scale scenarios

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -11213,7 +11213,7 @@ impl<'a> Parser<'a> {
         if self.consume_token(&Token::LParen) {
             let precision = self.parse_literal_uint()?;
             let scale = if self.consume_token(&Token::Comma) {
-                Some(self.parse_literal_uint()?)
+                Some(self.parse_scale_value()?)
             } else {
                 None
             };
@@ -11226,6 +11226,38 @@ impl<'a> Parser<'a> {
             }
         } else {
             Ok(ExactNumberInfo::None)
+        }
+    }
+
+    /// Parse a scale value for NUMERIC/DECIMAL data types.
+    ///
+    /// Supports positive, negative, and explicitly positive (with `+`) scale values.
+    /// Negative scale values are particularly useful for PostgreSQL, where they indicate
+    /// rounding to the left of the decimal point. For example:
+    /// - `NUMERIC(5, 2)` stores up to 5 digits with 2 decimal places (e.g., 123.45)
+    /// - `NUMERIC(5, -2)` stores up to 5 digits rounded to hundreds (e.g., 12300)
+    fn parse_scale_value(&mut self) -> Result<i64, ParserError> {
+        let next_token = self.next_token();
+        match next_token.token {
+            Token::Number(s, _) => Self::parse::<i64>(s, next_token.span.start),
+            Token::Minus => {
+                let next_token = self.next_token();
+                match next_token.token {
+                    Token::Number(s, _) => {
+                        let positive_value = Self::parse::<i64>(s, next_token.span.start)?;
+                        Ok(-positive_value)
+                    }
+                    _ => self.expected("number after minus", next_token),
+                }
+            }
+            Token::Plus => {
+                let next_token = self.next_token();
+                match next_token.token {
+                    Token::Number(s, _) => Self::parse::<i64>(s, next_token.span.start),
+                    _ => self.expected("number after plus", next_token),
+                }
+            }
+            _ => self.expected("number", next_token),
         }
     }
 
@@ -17069,7 +17101,7 @@ mod tests {
         use crate::ast::{
             CharLengthUnits, CharacterLength, DataType, ExactNumberInfo, ObjectName, TimezoneInfo,
         };
-        use crate::dialect::{AnsiDialect, GenericDialect};
+        use crate::dialect::{AnsiDialect, GenericDialect, PostgreSqlDialect};
         use crate::test_utils::TestedDialects;
 
         macro_rules! test_parse_data_type {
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -11229,6 +11229,38 @@ impl<'a> Parser<'a> {
...
+            }
+            Token::Plus => {
+                let next_token = self.next_token();
+                match next_token.token {
+                    Token::Number(s, _) => Self::parse::<i64>(s, next_token.span.start),
+                    _ => self.expected("number after plus", next_token),
+                }
+            }
+            _ => self.expected("number", next_token),
+        }
```
iffyio: I think we might be able to simplify this to something like the following?
```rust
if !self.consume_token(Token::Minus) {
    return i64::try_from(self.parse_literal_uint()?)
}

let next_token = self.next_token_ref();
match &next_token.token {
  Token::Number(s, _) => Self::parse::<i64>(s, next_token.span.start),
  _ => self.expected_ref("literal int", next_token),
}
```

src/parser/mod.rs
```diff
@@ -11229,6 +11229,38 @@ impl<'a> Parser<'a> {
...
     }
 
+    /// Parse a scale value for NUMERIC/DECIMAL data types.
+    ///
+    /// Supports positive, negative, and explicitly positive (with `+`) scale values.
+    /// Negative scale values are particularly useful for PostgreSQL, where they indicate
+    /// rounding to the left of the decimal point. For example:
+    /// - `NUMERIC(5, 2)` stores up to 5 digits with 2 decimal places (e.g., 123.45)
+    /// - `NUMERIC(5, -2)` stores up to 5 digits rounded to hundreds (e.g., 12300)
+    fn parse_scale_value(&mut self) -> Result<i64, ParserError> {
```
```suggestion
    /// Parse an optionally signed integer literal.
    fn parse_signed_integer(&mut self) -> Result<i64, ParserError> {
```
Think we can make the description and function name specific since it should be a straight-forward use case.

src/parser/mod.rs
```diff
@@ -17319,6 +17351,75 @@ mod tests {
                 "DEC(2,10)",
                 DataType::Dec(ExactNumberInfo::PrecisionAndScale(2, 10))
             );
+
+            // Test negative scale values (PostgreSQL supports scale from -1000 to 1000)
```
```suggestion
            // Test negative scale values.
```
we can skip the postgres comment since the parser doesnt treat it specially

src/parser/mod.rs
```diff
@@ -17319,6 +17351,75 @@ mod tests {
...
+                    data_type
+                );
+                // Note: Explicit '+' sign is not preserved in output, which is correct
+                assert_eq!("NUMERIC(10,5)", data_type.to_string());
+            });
+        }
+
+        #[test]
+        fn test_numeric_negative_scale() {
+            let dialect = TestedDialects::new(vec![
```
iffyio: I think the negative test cases can be merged into `test_ansii_exact_numeric_types`? since they're the same feature. Also would let us avoid the duplicate tests between the two functions

src/parser/mod.rs
IndexSeek: Wow, yes, this is much simpler and a way better way to handle the signs. 

Thank you!

src/parser/mod.rs
IndexSeek: I agree, this is a bit redundant.

IndexSeek:
Thank you so much for the detailed review @iffyio! I think I have addressed your comments. Please let me know if there are any other changes I should make.

src/parser/mod.rs
```diff
@@ -11229,6 +11229,30 @@ impl<'a> Parser<'a> {
...
+            Token::Minus => {
+                let number_token = self.next_token();
+                (-1, number_token)
+            }
+            Token::Plus => {
+                let number_token = self.next_token();
+                (1, number_token)
+            }
+            _ => (1, next_token),
+        };
```
iffyio: was the pattern in the previous comment example insufficient? I think the current approach to optionally multiply by negative makes things slightly longer to follow what the code is trying to achieve.

fyi note ideally we use `next_token_ref` since that avoid cloning, especially given that `parse::<i64>` does not require an owned value

IndexSeek: When I tried to implement your suggested pattern, it failed on test cases with explicit `+` signs (like `NUMERIC(10,+5)`) because `parse_literal_uint()` expected a number token. I should have posted my issue earlier. 

I actually went back and tested that PostgreSQL doesn't even take that (ERROR: type modifiers must be simple constants or identifiers), so I was overcomplicating it and I will remove this test case.

> fyi note ideally we use next_token_ref since that avoid cloning, especially given that parse::<i64> does not require an owned value

I did not find `next_token_ref` but it looks that `advance_token()` + `get_current_token()` will give us this behavior (#1618)? 

Here's what I am working with right now after removing that explicit "+" test:

```rs
    fn parse_signed_integer(&mut self) -> Result<i64, ParserError> {
        if !self.consume_token(&Token::Minus) {
            return i64::try_from(self.parse_literal_uint()?)
                .map_err(|_| ParserError::ParserError("Integer overflow".to_string()));
        }

        self.advance_token();
        let next_token = self.get_current_token();
        match &next_token.token {
            Token::Number(s, _) => {
                let positive_value = Self::parse::<i64>(s.clone(), next_token.span.start)?;
                Ok(-positive_value)
            }
            _ => self.expected_ref("literal int", next_token),
        }
    }
```

I believe this should only clone the string data and not the tokens now.

iffyio: Ah yeah that makes sense, I confused `peek_token_ref` and was hoping to avoid the clone entirely but looking at `Self::parse` now I see its not currently set up to take in references so the current approach to clone the string sounds reasonable! Here's a couple minor updates to your example I think should do what we intended?

```rust
    fn parse_signed_integer(&mut self) -> Result<i64, ParserError> {
        if self.consume_token(&Token::Minus) {
            return i64::try_from(self.parse_literal_uint()?)
                .map(|v| -v)
                .or_else(|_| self.expected_ref("i64 literal", self.peek_token_ref()))
        }

        let _ = self.consume_token(&Token::Plus);
        self.advance_token();
        let next_token = self.get_current_token();
        match &next_token.token {
            Token::Number(s, _) => Self::parse::<i64>(s.clone(), next_token.span.start),
            _ => self.expected_ref("literal int", next_token),
        }
    }
```

IndexSeek: Thanks for the suggestion! It does read cleaner not starting with negation, looking back, opening with `if !self.consume_token(&Token::Minus)` felt a bit strange. 

I like how your solution still handles the + sign, but I noticed a few issues with the logic flow. I think that with an optional + token on `let _ = self.consume_token(&Token::Plus);` followed by `self.advance_token()` and then `self.get_current_token()` could advance twice and potentially skip the actual number token. 

I did take a look at `peek_token_ref` and was wondering if we could use it like so?

```rs
    fn parse_signed_integer(&mut self) -> Result<i64, ParserError> {
        let is_negative = self.consume_token(&Token::Minus);

        if !is_negative {
            let _ = self.consume_token(&Token::Plus);
        }

        let current_token = self.peek_token_ref();
        match &current_token.token {
            Token::Number(s, _) => {
                let s = s.clone();
                let span_start = current_token.span.start;
                self.advance_token();
                let value = Self::parse::<i64>(s, span_start)?;
                Ok(if is_negative { -value } else { value })
            }
            _ => self.expected_ref("number", current_token),
        }
    }
```

I think that should avoid us needing to consume the token until we know we have a number.

IndexSeek: Should we still parse the "+" character or error out? I know I originally added a test for it but I'm not sure any dialects support a "+" token in the scale definition for numeric types.

iffyio: Yeah the last snippet sounds reasonable to me! I think we should still parse the `+` character since the idea is that the api can be used in other contexts beyond the numeric type scale

IndexSeek: That sounds good to me! I also added back that test that was explicitly checking to ensure that "+" would work and be removed when converted back to a string, but we can remove those lines if it is redundant.

iffyio (approved):
LGTM! Thanks @IndexSeek!
cc @alamb

PR #1991, opened 2025-08-04
Add support for `UPDATE ... LIMIT ...`

xtuc:


tests/sqlparser_sqlite.rs
```diff
@@ -592,6 +593,41 @@ fn test_regexp_operator() {
     sqlite().verified_only_select(r#"SELECT count(*) FROM messages WHERE msg_text REGEXP '\d+'"#);
 }
 
+#[test]
+fn test_update_delete_limit() {
```
iffyio, line 597: hmm was there a requirement to use a test dialect here instead of `sqlite()` method?

src/parser/mod.rs
```diff
@@ -15116,13 +15116,20 @@ impl<'a> Parser<'a> {
         } else {
             None
         };
+        let limit =
+            if self.dialect.supports_update_delete_limit() && self.parse_keyword(Keyword::LIMIT) {
+                Some(self.parse_expr()?)
+            } else {
+                None
+            };
```
iffyio: Since the syntax doesn't seem to conflict with other dialects, can we have the parser accept the clause if it shows up, avoiding the dialect method. e.g.
```rust
let limit = self.maybe_parse(|p| {
  if p.parse_keyword(Keyword::LIMIT) {
    Some(self.parse_expr()?)
  }
  None
})?;
```

tests/sqlparser_sqlite.rs
xtuc: I need a new dialect that has supports_update_delete_limit.

Unfortunately if we add parameters to `SQLiteDialect {}` it will be a breaking change. I think it would be better though.

src/parser/mod.rs
xtuc: Yeah that would simplify things. It means that sqlparser can accept a statement and sqlite won't (since `update ... limit ...` is behind a compilation flag). I though it would be better for the user to configure it, but up to you.

iffyio: Yeah I think it would be fine to let the parser always accept the LIMIT clause if it shows up

xtuc: Ok, pushed the changes.

iffyio (approved):
LGTM! Thanks @xtuc!
cc @alamb

iffyio:
@xtuc could you merge in the latest from main in order to resolve the CI lint error?

PR #1993, opened 2025-08-05
Display `CREATE INDEX` column list separated by commas with spaces

mvzink:
The column list in `CREATE INDEX` now matches the style used elsewhere, e.g. in `TableConstraint`, which is to use spaces after commas.

```sql
-- before:
CREATE INDEX idx_name ON table_name (column1,column2,column3);
-- after:
CREATE INDEX idx_name ON table_name (column1, column2, column3);
```

When `CreateIndex` was added, there was no explanation for the lack of spaces, so I assume it was just author preference. But standard style in all documentation I've seen is to use spaces after commas (including [MSSQL]'s documentation of `INCLUDE`, which copied the no-spaces style when added).

[MSSQL]: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver17#i-create-an-index-with-included-non-key-columns

### DIFF
--- a/src/ast/dml.rs
+++ b/src/ast/dml.rs
@@ -106,9 +106,9 @@ impl Display for CreateIndex {
         if let Some(value) = &self.using {
             write!(f, " USING {value} ")?;
         }
-        write!(f, "({})", display_separated(&self.columns, ","))?;
+        write!(f, "({})", display_comma_separated(&self.columns))?;
         if !self.include.is_empty() {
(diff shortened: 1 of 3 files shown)


mvzink:
Actually, I'm making much more substantial changes to `CREATE INDEX` and this is kind of minor and not worth reviewing in isolation and holding up a different more important PR; so I'll close this one and open a new one shortly.

PR #1996, opened 2025-08-06
Snowflake - support table function in table factor (regression)

tomershaniii:


iffyio (approved):
LGTM! Thanks @tomershaniii!
cc @alamb

PR #1998, opened 2025-08-07
Improve MySQL `CREATE TRIGGER` parsing

mvzink:
MySQL uses a statement body similar to MSSQL (but without the `AS` keyword) instead of `EXECUTE` body style used in Postgres and standard SQL. But unlike MSSQL, MySQL puts the trigger period before the target table. We add some flags to indicate these differences and allow parsing and round tripping MySQL triggers. The main benefit is that we can now handle MySQL triggers which include `BEGIN; ... END;` compound statements.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3962,6 +3962,15 @@ pub enum Statement {
         /// EXECUTE FUNCTION trigger_function();
         /// ```
         period: TriggerPeriod,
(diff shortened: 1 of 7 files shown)


mvzink:
Had to fix some new lint errors in preexisting code because 1.89 came out; happy to pull that into a separate PR if desired.

iffyio (approved):
LGTM! Thanks @mvzink!
cc @alamb

Issue #1956, opened 2025-07-18
Release sqlparser-rs version `0.59.0` around 2025-09-15


alamb:
Follow on to 
- https://github.com/apache/datafusion-sqlparser-rs/issues/1886

This ticket tracks creating the next sqlparser release (mostly so others can follow along)

**Target Release Date**: Approximately 2025-9-15 (2 months after the last release)

Release instructions: https://github.com/apache/datafusion-sqlparser-rs/blob/main/dev/release/README.md

Steps:
- [ ] Tested with DataFusion: 
- [x] Make a PR to update version / `CHANGELOG.md`: https://github.com/apache/datafusion-sqlparser-rs/pull/2039
- [x] Create a Release Candidate (RC) and start vote
- [x] (When vote is approved)  release to crates.io
- [x] Make ticket for next release: https://github.com/apache/datafusion-sqlparser-rs/issues/2045

Required fixes: 
- [ ] None yet 🎉

alamb:
@iffyio  -- I am thinking of making a release next week -- is there anything in particular you think we should wait for?

iffyio:
@alamb sounds good to me! Nothing in particular I don't think needs waiting on

alamb:
I have created a PR with a new version and changelog:
- https://github.com/apache/datafusion-sqlparser-rs/pull/2039

I will now proceed to making a RC

alamb:
I have started the vote thread: https://lists.apache.org/thread/nd89vc32l9cls2mj983gy5dd019vdjkb

lovasoa:
Hi @alamb ! When do you think the release will be made ? Just checking to know whether I should wait for it before publishing the new version of SQLPage.

alamb:
> Hi [@alamb](https://github.com/alamb) ! When do you think the release will be made ? Just checking to know whether I should wait for it before publishing the new version of SQLPage.

I expect it to be out within the next day -- I am trying to get one more vote to finalize the release. I'll try and scare one up now

lovasoa:
I know I don't have a right to vote but for what it's worth I tested it with SQLPage and my full test suite passes without changes :)

alamb:
Thanks for testing @lovasoa ❤️ 

The release is available here:
  https://dist.apache.org/repos/dist/release/datafusion/datafusion-sqlparser-rs-0.59.0

I have also published it on [crates.io](http://crates.io/)
https://crates.io/crates/sqlparser/0.59.0

[...]

PR #1999, opened 2025-08-08
feat: Include end token in `ALTER TABLE` statement

### LINKED ISSUE #1858: Postgres: Incorrect ending line reported in `Span` for multi-line query
Using `0.56.0` and the following code:

```rust
use sqlparser::dialect::PostgreSqlDialect;
use sqlparser::parser::Parser as Parser;

fn f()
    let sql = r#"-- foo
ALTER TABLE users
  ADD COLUMN foo
  varchar; -- hi there

-- bar
-- baz
ALTER TABLE users ADD COLUMN bar varchar; SELECT 1;"#;

    for stmt in ast {
        let s = stmt.span();
        println!("stmt: {} | start: {} | end: {}", stmt, s.start, s.end);
    }
}
```

The output I get is:

```
stmt: ALTER TABLE users ADD COLUMN foo VARCHAR | start:  at Line: 2, Column: 13 | end:  at Line: 3, Column: 17
stmt: ALTER TABLE users ADD COLUMN bar VARCHAR | start:  at Line: 8, Column: 13 | end:  at Line: 8, Column: 33
stmt: SELECT 1 | start:  at Line: 8, Column: 43 | end:  at Line: 8, Column: 51
```

Everything seems correct to me, except of the ending line of the first query, which I expected it to be `4` instead of `3`.

IndexSeek:
Resolves #1858 in span calculation for `ALTER TABLE` statements. The span for `ALTER TABLE` statements was ending too early and not including the semicolon.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3382,6 +3382,9 @@ pub enum Statement {
         /// Snowflake "ICEBERG" clause for Iceberg tables
         /// <https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table>
         iceberg: bool,
+        /// Token that represents the end of the statement (semicolon or EOF)
+        /// Used for accurate span calculation
+        end_token: AttachedToken,
     },
     /// ```sql
     /// ALTER INDEX
@@ -5419,6 +5422,7 @@ impl fmt::Display for Statement {
                 location,
                 on_cluster,
                 iceberg,
+                end_token: _,
             } => {
                 if *iceberg {
                     write!(f, "ALTER ICEBERG TABLE ")?;
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3383,7 +3383,6 @@ pub enum Statement {
         /// <https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table>
         iceberg: bool,
         /// Token that represents the end of the statement (semicolon or EOF)
-        /// Used for accurate span calculation
         end_token: AttachedToken,
     },
     /// ```sql
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -4759,6 +4759,25 @@ fn parse_alter_table() {
 }
 
 #[test]
+fn alter_table_span_includes_semicolon() {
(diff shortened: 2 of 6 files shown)


IndexSeek:
I'm not sure about the linting violations, I addressed one, but it seemed unrelated to the change.

src/ast/mod.rs
```diff
@@ -3382,6 +3382,9 @@ pub enum Statement {
         /// Snowflake "ICEBERG" clause for Iceberg tables
         /// <https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table>
         iceberg: bool,
+        /// Token that represents the end of the statement (semicolon or EOF)
+        /// Used for accurate span calculation
```
```suggestion
```

tests/sqlparser_common.rs
```diff
@@ -4758,6 +4758,25 @@ fn parse_alter_table() {
...
+  varchar; -- hi there"#;
+    let dialect = PostgreSqlDialect {};
+    let ast = Parser::parse_sql(&dialect, sql).unwrap();
+    let stmt = &ast[0];
+    let span = stmt.span();
+
+    assert_eq!(span.end.line, 4);
+    assert_eq!(span.end.column, 11);
+}
+
```
iffyio: this seems like a similar test to the one in spans.rs, the latter felt like it should provide sufficient coverage so that we can likely skip this one?

iffyio (commented):
Thanks @IndexSeek! Changes look good to me overall, left some minor comments.

The lint failure has been fixed on the latest main so that merging in from main should resolve the failure in this branch

tests/sqlparser_common.rs
IndexSeek: Good catch, yes, that was redundant! I've removed this one.

agis:
Thanks! I can confirm this solved #1858. Are there any estimates on the next release?

alamb:
> Thanks! I can confirm this solved #1858. Are there any estimates on the next release?

This is the next scheduled release
- https://github.com/apache/datafusion-sqlparser-rs/issues/1956

We mostly do releases on a schedule, but feel free to discss on that ticket if you would like to request an earlier release

PR #2002, opened 2025-08-15
Add drop behavior to DROP PRIMARY/FOREIGN KEY

yoavcloud:


src/ast/ddl.rs
```diff
@@ -658,8 +659,31 @@ impl fmt::Display for AlterTableOperation {
...
             }
-            AlterTableOperation::DropPrimaryKey => write!(f, "DROP PRIMARY KEY"),
-            AlterTableOperation::DropForeignKey { name } => write!(f, "DROP FOREIGN KEY {name}"),
+            AlterTableOperation::DropPrimaryKey { drop_behavior } => {
+                write!(
+                    f,
+                    "DROP PRIMARY KEY{}",
+                    match drop_behavior {
+                        None => "",
+                        Some(DropBehavior::Restrict) => " RESTRICT",
```
mvzink: `DropBehavior` has `Display`, so you could use that.

alamb:
Woohoo -- we are past issue/PR 2000 -- a major milestone indeed

PR #2003, opened 2025-08-17
SGA-3801 Added support for concatenating string literals saperated by…

etgarperets:
… space in mySQL

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -476,6 +476,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    // Does the Dialect support concatenating of string literal
+    // Example: SELECT 'Hello ' "world" => SELECT 'Hello world'
+    fn supports_concat_quoted_identifiers(&self) -> bool {
+        false
+    }
+
     /// Does the dialect support trailing commas in the projection list?
     fn supports_projection_trailing_commas(&self) -> bool {
         self.supports_trailing_commas()
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -71,6 +71,11 @@ impl Dialect for MySqlDialect {
         true
     }
 
+    // see <https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#:~:text=mysql%3E%20SELECT%20%27My%27%20%27S%27%20%27QL%27%3B%0A%20%20%20%20%20%20%20%20%2D%3E%20%27MySQL%27>
+    fn supports_concat_quoted_identifiers(&self) -> bool {
+        true
+    }
+
     fn ignores_wildcard_escapes(&self) -> bool {
         true
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9684,8 +9684,18 @@ impl<'a> Parser<'a> {
             // bigdecimal feature is enabled, and is otherwise a no-op
             // (i.e., it returns the input string).
             Token::Number(n, l) => ok_value(Value::Number(Self::parse(n, span.start)?, l)),
-            Token::SingleQuotedString(ref s) => ok_value(Value::SingleQuotedString(s.to_string())),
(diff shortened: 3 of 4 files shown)


src/dialect/mysql.rs
```diff
@@ -71,6 +71,11 @@ impl Dialect for MySqlDialect {
         true
     }
 
+    // see <https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#:~:text=mysql%3E%20SELECT%20%27My%27%20%27S%27%20%27QL%27%3B%0A%20%20%20%20%20%20%20%20%2D%3E%20%27MySQL%27>
```
iffyio:
```suggestion
    /// see <https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_concat>
```

src/dialect/mod.rs
```diff
@@ -476,6 +476,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    // Does the Dialect support concatenating of string literal
+    // Example: SELECT 'Hello ' "world" => SELECT 'Hello world'
+    fn supports_concat_quoted_identifiers(&self) -> bool {
```
```suggestion
    /// Does the Dialect support concatenating of string literal
    /// Example: `SELECT 'Hello ' "world" => 'Hello world'`
    fn supports_string_literal_concatenation(&self) -> bool {
```

src/parser/mod.rs
```diff
@@ -9755,6 +9765,35 @@ impl<'a> Parser<'a> {
...
+            if !self.is_quoted_string(&next_token.token) {
+                self.prev_token();
+                break;
+            }
+            let s = self.get_quoted_string(&next_token.token);
+            combined_string.push_str(&s);
+        }
+        combined_string
+    }
+
```
iffyio: Can we simplify with something roughly like the following?

```rust
fn maybe_concat_string_literal<F: FnOnce(String) -> Value>(
	mut str: String,
	value_fn: F
) -> Result<Value> {
	if self.dialect.supports_concat_quoted_identifiers() {
		loop {
			match self.peek_token_ref() {
				Token::SingleQuotedString(s) | Token::DoubleQuotedString(s) => {
					self.advance_token();
					str.append(s);
				}
				_ => {
					break
				}
			}
		}
	}

	Ok(value_fn(str))
}
```
And the caller only does e.g.
```rust
Token::SingleQuotedString(ref s) => self.maybe_concat_string_literal(s.to_string(), Value::SingleQuotedString)
```

tests/sqlparser_mysql.rs
```diff
@@ -4245,3 +4245,12 @@ fn test_create_index_options() {
         "CREATE INDEX idx_name ON t(c1, c2) USING BTREE LOCK = EXCLUSIVE ALGORITHM = DEFAULT",
     );
 }
+
+#[test]
+fn parse_adjacent_string_literal_concatenation() {
+    let sql = r#"SELECT 'M' "y" 'S' "q" 'l'"#;
+    mysql().one_statement_parses_to(sql, r"SELECT 'MySql'");
```
iffyio: since this is gated by a dialect flag can we use `all_dialects_where(|d| d.supports...)` so that new dialects are covered by tests when enabled

etgarperets:
@iffyio All fixes are in and waiting on a review

src/dialect/mod.rs
```diff
@@ -476,6 +476,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    // Does the Dialect support concatenating of string literal
+    // Example: SELECT 'Hello ' "world" => SELECT 'Hello world'
```
```suggestion
    /// Returns true if the dialect supports concatenating of string literal
    /// Example: `SELECT 'Hello ' "world" => SELECT 'Hello world'`
```

iffyio (commented):
Thanks @etgarperets! Changes look good to me overall, left a minor comment

src/dialect/mod.rs
etgarperets: Done @iffyio

iffyio (approved):
LGTM! Thanks @etgarperets!
cc @alamb

PR #2004, opened 2025-08-18
Add SECURE keyword for views in Snowflake

Vedin:
Snowflake views support `SECURE` keyword on creation. 
https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax
This PR introduced this keyword and view-related logic.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4713,11 +4713,11 @@ impl<'a> Parser<'a> {
         let create_view_params = self.parse_create_view_params()?;
         if self.parse_keyword(Keyword::TABLE) {
             self.parse_create_table(or_replace, temporary, global, transient)
-        } else if self.parse_keyword(Keyword::MATERIALIZED)
-            || self.parse_keyword(Keyword::VIEW)
-            || (dialect_of!(self is SnowflakeDialect) && self.parse_keyword(Keyword::SECURE))
+        } else if self.peek_keyword(Keyword::MATERIALIZED)
+            || self.peek_keyword(Keyword::VIEW)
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -4713,7 +4713,10 @@ impl<'a> Parser<'a> {
         let create_view_params = self.parse_create_view_params()?;
         if self.parse_keyword(Keyword::TABLE) {
             self.parse_create_table(or_replace, temporary, global, transient)
-        } else if self.parse_keyword(Keyword::MATERIALIZED) || self.parse_keyword(Keyword::VIEW) {
+        } else if self.parse_keyword(Keyword::MATERIALIZED)
+            || self.parse_keyword(Keyword::VIEW)
+            || (dialect_of!(self is SnowflakeDialect) && self.parse_keyword(Keyword::SECURE))
+        {
             self.prev_token();
```
```suggestion
        } else if self.peek_keyword(Keyword::MATERIALIZED)
            || self.peek_keyword(Keyword::VIEW)
            || self.peek_keywords(&[Keyword::SECURE, Keyword::MATERIALIZED])
            || self.peek_keywords(&[Keyword::SECURE, Keyword::VIEW])

            || (dialect_of!(self is SnowflakeDialect) && self.parse_keyword(Keyword::SECURE))
        {
```
We could probably do something like this? also I imagine no need for the snowflakeDialect check since the syntax doesn't conflict with other dialects.

Vedin: Sure, done. snowflake dialect check was removed as well.

iffyio (approved):
LGTM! Thanks @Vedin!
cc @alamb

PR #2005, opened 2025-08-19
Add support for VACUUM in Redshift

yoavcloud:


src/ast/mod.rs
```diff
@@ -10597,6 +10605,50 @@ impl fmt::Display for InitializeKind {
     }
 }
 
+/// Re-sorts rows and reclaims space in either a specified table or all tables in the current database
+///
+/// '''sql
+/// VACUUM [ FULL | SORT ONLY | DELETE ONLY | REINDEX | RECLUSTER ] [ \[ table_name \] [ TO threshold PERCENT ] \[ BOOST \] ]
+/// '''
```
iffyio, line 10619:
```suggestion
/// ```sql
/// VACUUM [ FULL | SORT ONLY | DELETE ONLY | REINDEX | RECLUSTER ] [ \[ table_name \] [ TO threshold PERCENT ] \[ BOOST \] ]
/// ```
```

src/parser/mod.rs
```diff
@@ -649,6 +649,7 @@ impl<'a> Parser<'a> {
                     self.prev_token();
                     self.parse_export_data()
                 }
+                Keyword::VACUUM => self.parse_vacuum(),
```
```suggestion
                Keyword::VACUUM => {
                    self.prev_token();
                    self.parse_vacuum()
                }
```
Thinking so that we can make `parse_vacuum` standalone to be able to parse a full `VACUUM` statement

src/parser/mod.rs
```diff
@@ -16874,6 +16875,38 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    fn parse_vacuum(&mut self) -> Result<Statement, ParserError> {
+        let full = self.parse_keyword(Keyword::FULL);
+        let sort_only = self.parse_keywords(&[Keyword::SORT, Keyword::ONLY]);
+        let delete_only = self.parse_keywords(&[Keyword::DELETE, Keyword::ONLY]);
+        let reindex = self.parse_keyword(Keyword::REINDEX);
+        let recluster = self.parse_keyword(Keyword::RECLUSTER);
+        let (table_name, threshold, boost) = match self.parse_object_name(false) {
```
```suggestion
        let (table_name, threshold, boost) = match self.parse_object_name(false)? {
```

yoavcloud: Good catch, my intention was to optionally parse an object_name. See latest commit.

PR #2009, opened 2025-08-20
Add support for `SEMANTIC_VIEW` table factor

bombsimon:
Implements querying `SEMANTIC_VIEW` in Snowflake: https://docs.snowflake.com/en/sql-reference/constructs/semantic_view

```sql
SEMANTIC_VIEW(
  [<namespace>.]<semantic_view_name>
  [
    {
      METRICS <metric> [ , ... ] |
      FACTS <fact_expr>  [ , ... ]
    }
  ]
  [ DIMENSIONS <dimension_expr>  [ , ... ] ]
  [ WHERE <predicate> ]
)
```

---

This change was more or less completely made with AI using [Claude](https://claude.ai/). I thought that's interesting to share both to ensure there's no non-AI policy I might violate, but also to showcase how quick it can be to extend functionality of this parser!

Other than some missing links, assumptions about `METRICS` and the usage of the discouraged `dialect_of` macro it more or less worked with a single prompt with some example queries and constraints as input! 🎉


src/parser/mod.rs
```diff
@@ -13438,6 +13439,10 @@ impl<'a> Parser<'a> {
         } else if self.parse_keyword_with_tokens(Keyword::XMLTABLE, &[Token::LParen]) {
             self.prev_token();
             self.parse_xml_table_factor()
+        } else if self.dialect.supports_semantic_view()
+            && self.parse_keyword_with_tokens(Keyword::SEMANTIC_VIEW, &[Token::LParen])
+        {
+            self.parse_semantic_view_table_factor()
```
iffyio: It would be good to have `parse_semantic_view_table_factor()` as a standalone function that can parse a semantic view table - so that it can be used in other contexts. I think one way to do that would be to call `self.prev_token()` twice here but likely a better way would be to implement a similar `self.peek_keyword_with_tokens(Keyword::SEMANTIC_VIEW, &[Token::LParen])` that we can call here?

src/parser/mod.rs
```diff
@@ -13769,6 +13774,66 @@ impl<'a> Parser<'a> {
         Ok(XmlPassingClause { arguments })
     }
 
+    fn parse_semantic_view_table_factor(&mut self) -> Result<TableFactor, ParserError> {
```
iffyio, line 13866:
```suggestion
    /// Parse a [TableFactor::SemanticView]
    fn parse_semantic_view_table_factor(&mut self) -> Result<TableFactor, ParserError> {
```

tests/sqlparser_snowflake.rs
```diff
@@ -4581,3 +4581,129 @@ fn test_drop_constraints() {
...
+#[test]
+fn test_semantic_view_all_variants_should_pass() {
+    let test_cases = [
+        ("SELECT * FROM SEMANTIC_VIEW(model)", None),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model DIMENSIONS dim1, dim2)",
+            None,
+        ),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model METRICS met1, met2)",
```
```suggestion
            "SELECT * FROM SEMANTIC_VIEW(a.b METRICS c.d, met2)",
```
thinking we can include a test case that parses a qualified object name instead vs identifier

tests/sqlparser_snowflake.rs
```diff
@@ -4581,3 +4581,129 @@ fn test_drop_constraints() {
...
+            None,
+        ),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model FACTS fact1, fact2)",
+            None,
+        ),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model DIMENSIONS dim1 METRICS met1)",
+            None,
```
iffyio: Can we include test to demonstrate that facts can accept arbitrary expressions in addition to identifiers?

tests/sqlparser_snowflake.rs
```diff
@@ -4581,3 +4581,129 @@ fn test_drop_constraints() {
...
+        ),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model DIMENSIONS DATE_PART('year', col))",
+            None,
+        ),
+        (
+            "SELECT * FROM SEMANTIC_VIEW(model METRICS orders.col, orders.col2)",
+            None,
+        ),
+        // We can parse in any order bu will always produce a result in a fixed order.
```
```suggestion
        // We can parse in any order but will always produce a result in a fixed order.
```

tests/sqlparser_snowflake.rs
```diff
@@ -4581,3 +4581,129 @@ fn test_drop_constraints() {
...
+    ];
+
+    for sql in invalid_sqls {
+        let result = snowflake().parse_sql_statements(sql);
+        assert!(result.is_err(), "Expected error for invalid SQL: {}", sql);
+    }
+}
+
+#[test]
+fn test_semantic_view_ast_structure() {
```
iffyio: can we inline all three scenarios into the same `test_semantic_view_table_factor()` function? since they belong to the same feature

src/parser/mod.rs
bombsimon: I'm not sure if it would look the same in other contexts, the syntax above comes _specifically_ from the [`SELECT FROM`](https://docs.snowflake.com/en/user-guide/views-semantic/querying) docs which even states:

> Specifies the [semantic view](https://docs.snowflake.com/en/user-guide/views-semantic/overview) to [query](https://docs.snowflake.com/en/user-guide/views-semantic/querying). You specify SEMANTIC_VIEW(…) in a [FROM](https://docs.snowflake.com/en/sql-reference/constructs/from) clause in a [SELECT](https://docs.snowflake.com/en/sql-reference/sql/select) statement.

In the context of [creating a semantic view](https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view), the syntax is much different:

```sql
CREATE [ OR REPLACE ] SEMANTIC VIEW [ IF NOT EXISTS ] <name>
  TABLES ( logicalTable [ , ... ] )
  [ RELATIONSHIPS ( relationshipDef [ , ... ] ) ]
  [ FACTS ( semanticExpression [ , ... ] ) ]
  [ DIMENSIONS ( semanticExpression [ , ... ] ) ]
  [ METRICS ( semanticExpression [ , ... ] ) ]
  [ COMMENT = '<comment_about_semantic_view>' ]
  [ COPY GRANTS ]
```

In what other cases or context do you think we could make use of the current parser function?

tests/sqlparser_snowflake.rs
bombsimon: I can change this, but I'm not sure I agree. I think there are three different test scenarios and splitting them up makes the tests smaller and easier to read, we seem to do this elsewhere. In fact, we even have things like this 😅 

https://github.com/apache/datafusion-sqlparser-rs/blob/cb7a51e85f434d10d5b3ac79cfb09533cca004f7/tests/sqlparser_snowflake.rs#L4400-L4419

If this is something you want to change going forward I can merge them but my opinion is that keeping them separate is easier to read and maintain so my personal preference is to keep it as is.

**EDIT** Nvm, I merged them to one as suggested!

src/parser/mod.rs
iffyio: Ah so to clarify what I mean is basically that the function `self.parse_semantic_view_table_factor()` should be able to parse a full table factor `SEMANTIC_VIEW(...)`. With the current format that isn't the case because `self.parse_semantic_view_table_factor()` is set up to parse only a suffix of that table factor, which makes things harder to reuse.

bombsimon: Alright! I understod what you meant, just not why it was needed since I don't se a use-case for when it would be re-used. But easy enough to change so implemented this and now we have `peek_keyword_with_tokens` as well if ever needed 😄

src/parser/mod.rs
```diff
@@ -4241,17 +4241,33 @@ impl<'a> Parser<'a> {
...
     pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        self.keyword_with_tokens(expected, tokens, true)
+    }
+
+    /// Peeks to see if the current token is the `expected` keyword followed by specified tokens
+    /// without consuming them.
+    ///
+    /// Note that if the length of `tokens` is too long, this function will not be efficient as it
+    /// does a loop on the tokens with `peek_nth_token` each time.
+    pub fn peek_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
```
```suggestion
    pub(crate) fn peek_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
```

src/parser/mod.rs
```diff
@@ -4241,17 +4241,33 @@ impl<'a> Parser<'a> {
...
     /// each time.
     pub fn parse_keyword_with_tokens(&mut self, expected: Keyword, tokens: &[Token]) -> bool {
+        self.keyword_with_tokens(expected, tokens, true)
+    }
+
+    /// Peeks to see if the current token is the `expected` keyword followed by specified tokens
+    /// without consuming them.
+    ///
+    /// Note that if the length of `tokens` is too long, this function will not be efficient as it
+    /// does a loop on the tokens with `peek_nth_token` each time.
```
```suggestion
    /// See [Self::parse_keyword_with_tokens] for details.
```

src/dialect/mod.rs
```diff
@@ -1182,6 +1182,20 @@ pub trait Dialect: Debug + Any {
...
+    ///
+    /// ```sql
+    /// SELECT * FROM SEMANTIC_VIEW(
+    ///     model_name
+    ///     DIMENSIONS customer.name, customer.region
+    ///     METRICS orders.revenue, orders.count
+    ///     WHERE customer.active = true
+    /// )
+    /// ```
+    fn supports_semantic_view(&self) -> bool {
```
```suggestion
    fn supports_semantic_view_table_factor()(&self) -> bool {
```

src/parser/mod.rs
```diff
@@ -17869,4 +17954,18 @@ mod tests {
...
+        let mut parser = Parser::new(&GenericDialect {})
+            .try_with_sql(sql)
+            .expect("failed to create parser");
+
+        let ast = parser
+            .parse_semantic_view_table_factor()
+            .expect("should parse SEMANTIC_VIEW");
+
+        assert!(matches!(ast, TableFactor::SemanticView { .. }));
+    }
```
iffyio: Oh actually maybe we can move the existing test in snowflake over here instead, since we don't have any snowflake specific construct. Then here we can do `let dialect = all_dialects_where(|d| d.supports_semantic_view_table_factor());`

bombsimon: Moved the other tests to `sqlparser_common` and using the `TestedDialects` for the full queries.

I also removed this test after a discussion outside of GitHub. This means that we actually don't test whether or not `parse_semantic_view_table_factor` expects the first token to be `Keyword::SEMANTIC_VIEW` or if we just peeked the tokens before.

iffyio (approved):
LGTM! Thanks @bombsimon!
cc @alamb

PR #2013, opened 2025-08-23
Redshift: UNLOAD

yoavcloud:
Added support for the Redshift variant of UNLOAD


src/ast/mod.rs
```diff
@@ -8849,10 +8946,73 @@ impl fmt::Display for CopyLegacyOption {
...
 
+/// Specifies the partition keys for the unload operation
+///
+/// ```sql
+/// PARTITION BY ( column_name [, ... ] ) [ INCLUDE ]
+/// ```
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct PartitionBy {
```
```suggestion
pub struct UnloadPartitionBy {
```
Thinking since partitionby has meaning in other contexts?

src/ast/mod.rs
```diff
@@ -8796,16 +8831,46 @@ pub enum CopyLegacyOption {
...
+    /// JSON
+    Json,
+    /// MANIFEST \[ VERBOSE \]
+    Manifest { verbose: bool },
+    /// MAXFILESIZE \[ AS \] max-size \[ MB | GB \]
+    MaxFileSize(FileSize),
     /// NULL \[ AS \] 'null_string'
     Null(String),
+    /// PARALLEL
+    Parallel(Option<bool>),
```
iffyio, line 8872: can we add a description here to mention the behavior of on/off?

src/parser/mod.rs
```diff
@@ -9657,16 +9683,81 @@ impl<'a> Parser<'a> {
...
+            }
+            Some(Keyword::ROWGROUPSIZE) => {
+                let _ = self.parse_keyword(Keyword::AS);
+                let size = self.parse_number_value()?.value;
+                let unit = match self.parse_one_of_keywords(&[Keyword::MB, Keyword::GB]) {
+                    Some(Keyword::MB) => Some(FileSizeUnit::MB),
+                    Some(Keyword::GB) => Some(FileSizeUnit::GB),
+                    _ => None,
+                };
+                CopyLegacyOption::RowGroupSize(FileSize { size, unit })
```
iffyio: Can we pull this out to a function like `parse_file_size_unit()` so that it can be reused?

src/parser/mod.rs
```diff
@@ -16477,19 +16569,35 @@ impl<'a> Parser<'a> {
...
         let to = self.parse_identifier()?;
-
-        let with_options = self.parse_options(Keyword::WITH)?;
-
+        let auth = if self.parse_keyword(Keyword::IAM_ROLE) {
+            Some(self.parse_iam_role_kind()?)
+        } else {
+            None
+        };
+        let with = self.parse_options(Keyword::WITH)?;
```
iffyio, line 16629: Can we add a test case that uses the `WITH` keyword in the syntax?

yoavcloud: There's already one test case at the start of the unit test. This statement has two flavours, one for Athena (using WITH) and one for Redshift (not using WITH).

PR #2015, opened 2025-08-26
Add PostgreSQL `CREATE USER` and `ALTER USER` support

ramnes:
Currently, the library only supports `CREATE ROLE` and `ALTER ROLE` for
PostgreSQL. `CREATE USER` and `ALTER USER` fail to parse with errors
like `"Expected: an object type after CREATE, found: USER"`

But in PostgreSQL reference:
- `CREATE USER` is equivalent to `CREATE ROLE`, except that `LOGIN` is assumed by default
- `ALTER USER` is an alias to `ALTER ROLE`
- Both should support the same options as their ROLE counterparts

This commit extends the existing `CreateRole` and `AlterRole`
structures to distinct which keyword has been used: `USER` or
`ROLE`. It allows these expressions to be parsed and displayed back.


src/ast/mod.rs
```diff
@@ -3314,6 +3314,8 @@ pub enum Statement {
     CreateRole {
         names: Vec<ObjectName>,
         if_not_exists: bool,
+        /// Whether ROLE or USER keyword was used
+        keyword: RoleKeyword,
```
iffyio, line 3318: Could we use something like this representation instead? I think its a somewhat more invasive changes but clearer API wise.
```rust
struct CreateUserOrRole {
  names: Vec<ObjectName>,
  if_not_exists: bool,
  // Postgres
  login: Option<bool>,
  inherit: Option<bool>,
  // ...
}

Statement::CreateUser(CreateUserOrRole)
Statement::CreateRole(CreateUserOrRole)

struct AlterUserOrRole {
  name: Ident
}

Statement::AlterUser(AlterUserOrRole)
Statement::AlterRole(AlterUserOrRole)
```

src/parser/alter.rs
```diff
@@ -39,6 +39,15 @@ impl Parser<'_> {
         ))
     }
 
+    pub fn parse_alter_user(&mut self) -> Result<Statement, ParserError> {
+        if dialect_of!(self is PostgreSqlDialect) {
+            return self.parse_pg_alter_user();
+        }
```
iffyio: I think we can skip the dialect checks and leave the parser to always accept the statement if it shows up

src/parser/alter.rs
```diff
@@ -39,6 +39,15 @@ impl Parser<'_> {
         ))
     }
 
+    pub fn parse_alter_user(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 42:
```suggestion
    fn parse_alter_user(&mut self) -> Result<Statement, ParserError> {
```

src/ast/mod.rs
ramnes: We have a namespace problem here, because there's already a `CreateUser` structure for Snowflake semantics. In the grand scheme of things, `CreateUser` and `CreateRole` should probably be both compatible with both Snowflake and PostgreSQL? I'm not sure if I'm seasoned enough with the codebase to bring that in, and that would be a very different kind of work than this PR. :)

src/parser/alter.rs
ramnes: Merged it because it seemed right at first glance, but this one actually needs to stay public, so I'm removing that commit. :)

src/parser/alter.rs
ramnes: Alright!

src/ast/mod.rs
iffyio: Ah I'm not sure I followed, do we mean snowflake has CreateUser supported by the parser already?

ramnes: Yes, but as far as I understand, it's a very different syntax / semantics. Postgres "USER" is really just an alias for "ROLE", and Postgres semantics are currently handled by `CreateRole` and `AlterRole`. Which is why I just added the `keyword` here: it doesn't introduce any backwards compatibility. It's just more data to the AST for an already existing expression parsing.

ramnes: Again, this would probably need a refactor so that all dialects use the same structures for `ALTER / CREATE` `USER / ROLE`, but I guess it would make more sense in a second step – the inconsistency is already there. This PR fixes a currently broken expression parsing "for free".

iffyio: Sorry for the delay in getting back to this. I think ideally we would use separate representations, it would likely be confusing if a `CREATE USER` statement was being parsed as an `CREATE ROLE` statement or vice versa by the parser.

Also heads up there is a related PR adding `ALTER USER` [here](https://github.com/apache/datafusion-sqlparser-rs/pull/2035)

ramnes: From what I see of that PR, it is modeled on the current `CreateUser` structure, which only supports Snowflake.

So my understanding is that we'll have `CreateUser` and `AlterUser` that work for Snowflake, `CreateRole` and `AlterRole` that work for PostgreSQL, and no support for the opposite keyword / technology combo (`CREATE USER` support on PostgreSQL, or `CREATE ROLE` support on Snowflake).

I'd be happy to see the four structures unified into something that work for both technologies but I don't think I can do that at the moment, alas. :(

PR #2016, opened 2025-08-27
Support wildcard metrics for `SEMANTIC_VIEW`

bombsimon:
This adds support for the valid syntax of using wildcard metrics. It also improves the error message on unexpected tokens.


iffyio (approved):
LGTM! Thanks @bombsimon!
cc @alamb

PR #2017, opened 2025-08-28
Allow wilrdacrd for all `SEMANTIC_VIEW` types

bombsimon:
Followup on #2016, I didn't update all the types. `DIMENSIONS` also support wildcard, from [the docs](https://docs.snowflake.com/en/sql-reference/constructs/semantic_view)

> To specify all dimensions in a logical table, use an asterisk as a wildcard, qualified by the logical table name (for example, `my_logical_table.*`).

To avoid too restrictive parsing, although not currently valid according to documentation syntax, this also adds support for wildcard expression for `FACTS`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13951,7 +13951,7 @@ impl<'a> Parser<'a> {
                         "DIMENSIONS clause can only be specified once".to_string(),
                     ));
                 }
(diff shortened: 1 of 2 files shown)


iffyio (approved):
LGTM! Thanks @bombsimon!
cc @alamb

PR #2018, opened 2025-08-29
feat: add some DELTA keywords

roeap:
The LITE and INVENTORY keywords are used in the delta [VACUUM](https://docs.databricks.com/aws/en/sql/language-manual/delta-vacuum) command, would be great if we could add it.

The SHALLOW keyword is used in shallow clones of delta tables.

### DIFF
--- a/src/keywords.rs
+++ b/src/keywords.rs
@@ -491,6 +491,7 @@ define_keywords!(
     INTERSECTION,
     INTERVAL,
     INTO,
+    INVENTORY,
     INVOKER,
     IO,
     IS,


src/keywords.rs
```diff
@@ -491,6 +491,7 @@ define_keywords!(
     INTERSECTION,
     INTERVAL,
     INTO,
+    INVENTORY,
```
iffyio, line 494: Can we add test cases to demonstrate the added functionality?

roeap: Certainly @iffyio. I had been looking through the code for existing tests, but to no avail. So unsure if I should add some cases to the tokenizer tests or in the parser module?

Or is there a more obvious place I am missing?

iffyio: We can probably extend[ this test ](https://github.com/apache/datafusion-sqlparser-rs/blob/main/tests/sqlparser_redshift.rs#L412)to include new scenarios?

roeap: @iffyio - your hint got me to dive a bit deeper into the codebase. Unfortunately the Redhsift and Databricks VACUUM commands are quite different.

That said, would you accept a PR where I extend this a bit, add a custom statement parser option the databricks dialect and a `VacuumDatabricks` Statement variant?

iffyio: Ah yeah it should possible to support databricks' variant of the vacuum command. But we tend to not have dialect specific statements so that it'll likely need to be part of the existing `Statement::Vacuum` variant in this case, being extended to support the new options

Issue #713, opened 2022-11-18
Is there a way to parse `table_factor` only?


devgony:
I want to parse some table_factor with args like
```rs
"Table(3)"
```
but without parsing full sentense (`SELECT * FROM Table(3)`)

I found there is `fn parse_table_factor` but it seem like it can not be used stand-alone

AugustoFKL:
@devgony 

```rust
    #[test]
    fn parse_table_factor() {
        let dialect = GenericDialect {};
        let mut parser = Parser::new(
            Tokenizer::new(&dialect, "Table(3)").tokenize().unwrap(),
            &dialect,
        );

        let table_factor = parser.parse_table_factor().unwrap();
        println!("{:#?}", table_factor)
    }
```
This prints: 

```plaintext
TableFunction {
    expr: Value(
        Number(
            "3",
            false,
        ),
    ),
    alias: None,
}
```

I'm pretty sure you can do it externally from the crate, as everything I used is public.

AugustoFKL:
@devgony, can you please close this issue? I think it is solved, right...?

PR #2024, opened 2025-09-08
Add support for ClickHouse CSE.

### LINKED ISSUE #1514: Clickhouse `WITH <expression> AS <identifier>` syntax
Hey 👋 

The `WITH <expression> AS <identifier>` (doc [here](https://clickhouse.com/docs/en/sql-reference/statements/select/with)) does not seem to be supported. 

For instance, this command with the `ClickHouseDialect` does not work (nor with `GenericDialiect`):
```
WITH concat(prefix_addr, '/', prefix_len) AS 'prefix' SELECT prefix FROM table
```

```
sql parser error: Expected: (, found: 'prefix' at Line: 1, Column: 46
```

pravic:
https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions:

```sql
WITH <expression> AS <identifier>
```
fixes #1514.

* [x] syntax
* [x] docs
* [x] tests
* [ ] Unfortunately, this changes the public API a bit, so requires a version bump.

### DIFF
--- a/tests/sqlparser_clickhouse.rs
+++ b/tests/sqlparser_clickhouse.rs
@@ -1730,7 +1730,7 @@ fn test_parse_not_null_in_column_options() {
 }
 
 #[test]
-fn parse_cses() {
+fn parse_cse() {
     clickhouse().verified_stmt("WITH x AS (SELECT 1) UPDATE t SET bar = (SELECT * FROM x)");
 
     let with = concat!(
@@ -1755,6 +1755,37 @@ fn parse_cses() {
         "WHERE c < today"
     );
     clickhouse().verified_query(mixed);
+
+    // valid
+    clickhouse()
+        .parse_sql_statements("WITH foo() AS bar SELECT 1")
+        .unwrap();
+
+    // ClickHouse allows these, but not sqlparser
+    clickhouse()
+        .parse_sql_statements("WITH foo, bar SELECT 1")
+        .expect_err("Expected: AS, found: ,");
+
+    clickhouse()
+        .parse_sql_statements("WITH foo(), bar SELECT 1")
+        .expect_err("Expected: identifier, found: )");
+
+    // invalid
+    clickhouse()
+        .parse_sql_statements("WITH foo bar SELECT 1")
+        .expect_err("Expected: ");
+
+    clickhouse()
+        .parse_sql_statements("WITH foo() bar SELECT 1")
+        .expect_err("Expected: ");
+
+    clickhouse()
+        .parse_sql_statements("WITH foo() bar() SELECT 1")
+        .expect_err("Expected: ");
+
+    clickhouse()
+        .parse_sql_statements("WITH foo() AS bar() SELECT 1")
+        .expect_err("Expected: ");
 }
 
 fn clickhouse() -> TestedDialects {


src/parser/mod.rs
```diff
@@ -12260,6 +12260,27 @@ impl<'a> Parser<'a> {
...
+        } else {
+            CteOrCse::Cte(self.parse_cte()?)
+        })
+    }
+
+    /// Parse a CSE (`<expr> AS <ident>`).
+    pub fn parse_cse(&mut self) -> Result<Cse, ParserError> {
+        let expr = self.parse_expr()?;
+        let _after_as = self.parse_keyword(Keyword::AS);
+        let ident = self.parse_identifier()?;
```
pravic, line 12280: I guess, here we need to _expect_ this keyword.

src/ast/query.rs
```diff
@@ -641,6 +641,56 @@ impl fmt::Display for CteAsMaterialized {
...
+impl fmt::Display for CteOrCse {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            CteOrCse::Cte(cte) => cte.fmt(f),
+            CteOrCse::Cse(cse) => cse.fmt(f),
+        }
+    }
+}
+
+/// A single CSE (used after `WITH`): `<expr> AS <ident>`.
```
```suggestion
/// A common scalar expression.
/// Example:
/// ```sql
/// <expr> AS <ident>
/// ```
```
Can we add a link to the clickhouse docs here as well? in case the reader would like further context from here.

src/ast/query.rs
```diff
@@ -641,6 +641,56 @@ impl fmt::Display for CteAsMaterialized {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum CteOrCse {
+    Cte(Cte),
+    Cse(Cse),
+}
```
```suggestion
pub enum WithExpression {
    Cte(Cte),
    Cse(Cse),
}
```

src/ast/query.rs
```diff
@@ -641,6 +641,56 @@ impl fmt::Display for CteAsMaterialized {
...
+        }
+    }
+
+    pub fn cse(&self) -> Option<&Cse> {
+        match self {
+            CteOrCse::Cte(_) => None,
+            CteOrCse::Cse(cse) => Some(cse),
+        }
+    }
+}
```
```suggestion
```
I think we can skip this impl, in order to reduce the API surface of the library

src/parser/mod.rs
```diff
@@ -12260,6 +12260,27 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse a CTE or CSE.
+    pub fn parse_cte_or_cse(&mut self) -> Result<CteOrCse, ParserError> {
```
```suggestion
    /// Parse an expression in a `WITH` clause.
    fn parse_with_expression(&mut self) -> Result<WithExpression, ParserError> {
```

src/parser/mod.rs
```diff
@@ -12260,6 +12260,27 @@ impl<'a> Parser<'a> {
...
+        Ok(if dialect_of!(self is ClickHouseDialect) {
```
```suggestion
        Ok(if self.dialect.supports_common_scalar_expressions()) {
```
Can we use a dialect method instead? The `dialect_of` macro we try to avoid for new code.

src/parser/mod.rs
```diff
@@ -12260,6 +12260,27 @@ impl<'a> Parser<'a> {
...
+            } else {
+                CteOrCse::Cte(self.parse_cte()?)
+            }
+        } else {
+            CteOrCse::Cte(self.parse_cte()?)
+        })
+    }
+
+    /// Parse a CSE (`<expr> AS <ident>`).
+    pub fn parse_cse(&mut self) -> Result<Cse, ParserError> {
```
iffyio, line 12277:
```suggestion
    fn parse_cse(&mut self) -> Result<Cse, ParserError> {
```

src/parser/mod.rs
```diff
@@ -12260,6 +12260,27 @@ impl<'a> Parser<'a> {
...
+                CteOrCse::Cse(cse)
+            } else {
+                CteOrCse::Cte(self.parse_cte()?)
+            }
+        } else {
+            CteOrCse::Cte(self.parse_cte()?)
+        })
+    }
+
+    /// Parse a CSE (`<expr> AS <ident>`).
```
```suggestion
    /// Parse a [CSE].
```
Since we already mention the syntax in the struct description, to avoid them going out of sync in the future.

src/parser/mod.rs
iffyio: Yeah I think we'd want to use expect (we can also add a test case for this e.g. `WITH foo bar SELECT 1`)

tests/sqlparser_clickhouse.rs
```diff
@@ -1729,6 +1729,24 @@ fn test_parse_not_null_in_column_options() {
     );
 }
 
+#[test]
+fn parse_cses() {
+    clickhouse().verified_stmt("WITH x AS (SELECT 1) UPDATE t SET bar = (SELECT * FROM x)");
+
+    let with = concat!(
+        "WITH",
+        " toIntervalSecond(300) AS bucket_size,",
```
iffyio, line 1738: Can we include a CTE somewhere in this list to demonstrate the parser's behavior when CTE and CSE are mixed in the same WITH clause?

src/ast/query.rs
pravic: Here I just followed the doc style from the `CTE` below. But yeah, a better documentation is way more helpful.

src/ast/query.rs
pravic: But then we'd need way more changes in the existing code that relies on https://docs.rs/sqlparser/latest/sqlparser/ast/struct.With.html#structfield.cte_tables being `Cte`. I am on the edge here.

src/parser/mod.rs
pravic: Sure. I followed the existing style but this is a valid suggestion.

src/parser/mod.rs
pravic: Not `pub`?

Here there is a suggestion to use a parser method directly (as a solution/workaround for some problem): #713

So, I assumed all/most of `Parser` methods are/should be available to users.

pravic: For example, currently it's 195 private methods and 322 public methods.

src/parser/mod.rs
pravic: Done in f2ada38f6e and tests in a8cca3bc18.

tests/sqlparser_clickhouse.rs
pravic: Added in ec2abf60.

src/dialect/mod.rs
```diff
@@ -596,6 +596,20 @@ pub trait Dialect: Debug + Any {
...
+    /// ```sql
+    /// WITH
+    ///   toDate('2000-01-01') AS start_date
+    /// SELECT * from tbl WHERE col1 > start_date;
+    /// ```
+    ///
+    /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
+    fn supports_common_scalar_expressions(&self) -> bool {
+        false
+    }
```
pravic, line 611: Should we also add this to the generic dialect? I could not grasp the idea of what expressions should be supported in the generic dialect and what should be not.

For example, [`supports_select_expr_star`](https://github.com/apache/datafusion-sqlparser-rs/blob/4a307a937e071d5f315d347e931539af6ee644ad/src/dialect/mod.rs#L540-L542) isn't listed as supported. Neither is [`supports_select_exclude`](https://github.com/apache/datafusion-sqlparser-rs/blob/4a307a937e071d5f315d347e931539af6ee644ad/src/dialect/mod.rs#L595-L597).

src/parser/mod.rs
pravic: Same as above: https://github.com/apache/datafusion-sqlparser-rs/pull/2024#discussion_r2336307564

tests/sqlparser_clickhouse.rs
```diff
@@ -1729,6 +1729,65 @@ fn test_parse_not_null_in_column_options() {
...
+        .unwrap();
+
+    // ClickHouse allows these, but not sqlparser
+    clickhouse()
+        .parse_sql_statements("WITH foo, bar SELECT 1")
+        .expect_err("Expected: AS, found: ,");
+
+    clickhouse()
+        .parse_sql_statements("WITH foo(), bar SELECT 1")
+        .expect_err("Expected: identifier, found: )");
```
pravic, line 1771: Interesting: https://fiddle.clickhouse.com/1292850a-31e4-49c7-a309-79bc47a273a2

I am not sure whether it's a feature or a bug in the ClickHouse syntax parser. But since these examples don't make much sense, let's leave them as rejected.

src/parser/mod.rs
iffyio: I think it should be fine to leave them as public APIs in this case!

src/dialect/mod.rs
iffyio: In general [we try to add support for new syntax to the generic dialect](https://github.com/apache/datafusion-sqlparser-rs?tab=readme-ov-file#new-syntax) as long as the syntax doesn't conflict with existing syntax support on the dialect

src/ast/query.rs
iffyio: I think this is okay to skip, users can match directly on the enum when needed similar to other nodes in the AST

src/dialect/mod.rs
pravic: I think, CSE exist only in ClickHouse. Does it make sense to include this to the generic dialect? If yes, then:

> as long as the syntax doesn't conflict with existing syntax support on the dialect

It's difficult for me to answer to this question - I don't know whether there can be a conflict. But since it's only a ClickHouse feature - I'd not include this to the generic dialect. But I won't oppose either.

iffyio: Ah yeah if there's a conflict that would show up in the tests, so that it should be a matter of flagging it to true in the dialect and then changing the tests to use `all_dialects_where(|d| di.supports_common_scalar_expressions())` to test instead of `clickhouse()`

iffyio: @pravic just double checking the status of this PR, if you would have time to address this comment?

pravic: @iffyio No, haven't touched this bit so far. I you can, I would appreciate changing this.

PR #2025, opened 2025-09-09
MySQL:  Support `CROSS JOIN` constraint

rs-sac:
This is a MySQL-ism.

https://dev.mysql.com/doc/refman/8.4/en/join.html

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2333,7 +2333,11 @@ impl fmt::Display for Join {
                 self.relation,
                 suffix(constraint)
             )),
-            JoinOperator::CrossJoin => f.write_fmt(format_args!("CROSS JOIN {}", self.relation)),
+            JoinOperator::CrossJoin(constraint) => f.write_fmt(format_args!(
+                "CROSS JOIN {}{}",
+                self.relation,
+                suffix(constraint)
+            )),
             JoinOperator::Semi(constraint) => f.write_fmt(format_args!(
                 "{}SEMI JOIN {}{}",
                 prefix(constraint),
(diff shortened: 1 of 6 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -4132,6 +4132,40 @@ fn parse_straight_join() {
...
+                op: BinaryOperator::Eq,
+                right: Box::new(Expr::Identifier(Ident::new("b"))),
+            })),
+        },
+        only(only(select.from).joins),
+    );
+}
+
+#[test]
+fn parse_cross_join_using() {
```
iffyio: Can we merge this test case into the one above as a generic `fn parse_cross_join_constraint()` test case? since they are the same feature

tests/sqlparser_mysql.rs
```diff
@@ -4132,6 +4132,40 @@ fn parse_straight_join() {
         .verified_stmt("SELECT a.*, b.* FROM table_a STRAIGHT_JOIN table_b AS b ON a.b_id = b.id");
 }
 
+#[test]
+fn parse_cross_join_on() {
+    let sql = "SELECT * FROM t1 CROSS JOIN t2 ON a = b";
+    let select = mysql().verified_only_select(sql);
```
```suggestion
    let dialect = all_dialects_where(|d| d.supports_cross_join_constraint());
    let select = dialect.verified_only_select(sql);
```
can we do something like this instead, so that new dialects are automatically covered by this test case in the future (we could likely move this test to sqlparser_common.rs as a result too)

tests/sqlparser_mysql.rs
rs-sac: Done.

tests/sqlparser_mysql.rs
rs-sac: Done.

PR #2028, opened 2025-09-15
Implemented the `From` method for all clear variants in Statement

### LINKED ISSUE #2020: Implementing `From` for `sqlparser::ast::Statement` variants
The enum [`Statement`](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Statement.html) has several variants of the type `Variant(VariantStruct)`, such as `Set(Set)` or [`CreateIndex(CreateIndex)`](https://docs.rs/sqlparser/latest/sqlparser/ast/enum.Statement.html#variant.CreateIndex).

I would like to implement the `From<Variant> for Statement` trait for all such variants of Statement.

Do let me know whether that would be desirable.

LucaCappelletti94:
This PR introduces the `From` trait implementation for all variants of the `Statement` enum which are evidently mapped 1-1. There are a few variants which are not mappable in such a way, and therefore I refrained from creating one From impl for those. I have also added some `.into()` in the code where I have seen clear examples of statements conversion which may benefit from using the `From` trait instead of specifying the Statement variant. There are certainly more such cases, but I thought it was outside of the scope of this PR to replace such things.

This PR closes #2020


src/ast/statement_from_impls.rs
iffyio, line 1: For this I'm thinking it could make more sense to have the impls right next to the struct definitions (similar to how we do for display impls) vs a dedicated file for it?

LucaCappelletti94: I created the different module as the main one is ever more massive, so I thought it best not to add even more to it. That being said, if you prefer avoiding specialised sub modules but keep the monolitic one I can copy everything back into it.

iffyio: Yeah I think we can move the impls close to the associated struct definitions, indeed it would be nice to split the large files but I imagine splitting based on e.g. statement types or similar dimension grouping related structs in the same file

LucaCappelletti94: Done in commit [9ef033f](https://github.com/apache/datafusion-sqlparser-rs/pull/2028/commits/9ef033fc349cbfce0b94ae72570b6b7e30c79c72)

iffyio (approved):
cc @alamb

PR #2029, opened 2025-09-15
MySQL: `CREATE INDEX`: allow `USING` clause before `ON`

MohamedAbdeen21:
MySQL allows specifying the index type `USING index_type` before the `ON` clause in `CREATE INDEX` statements.

This PR allows the `CREATE INDEX` parser to accept both positions of the `USING` clause, regardless of the dialect.

docs: https://dev.mysql.com/doc/refman/8.4/en/create-index.html


src/parser/mod.rs
```diff
@@ -7061,19 +7061,27 @@ impl<'a> Parser<'a> {
     pub fn parse_create_index(&mut self, unique: bool) -> Result<Statement, ParserError> {
         let concurrently = self.parse_keyword(Keyword::CONCURRENTLY);
         let if_not_exists = self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
+
+        let mut using = None;
```
MohamedAbdeen21, line 7067: Not a fan of this `mut` but also not a fan of returning a tuple of (index_name, using) at line 7067.

src/parser/mod.rs
```diff
@@ -8588,6 +8596,14 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_using_index_type_clause(&mut self) -> Result<Option<IndexType>, ParserError> {
```
mvzink: `parse_optional_using_then_index_type` already exists

src/parser/mod.rs
```diff
@@ -7061,19 +7061,27 @@ impl<'a> Parser<'a> {
...
         let table_name = self.parse_object_name(false)?;
-        let using = if self.parse_keyword(Keyword::USING) {
-            Some(self.parse_index_type()?)
-        } else {
-            None
-        };
+
+        if let Some(second_using) = self.parse_using_index_type_clause()? {
+            if using.is_some() {
+                return Err(ParserError::ParserError("USING already specified".into()));
```
mvzink: MySQL actually allows specifying it in both positions and discards the first if both are present; I think we should do the same. I think then this entire block can just be `using = self.parse_optional_using_then_index_type()?.or(using);`

src/parser/mod.rs
mvzink: Seems okay IMO.

mvzink (changes requested):

src/parser/mod.rs
MohamedAbdeen21: I wasn't aware of that behaviour.

src/parser/mod.rs
MohamedAbdeen21: ack

src/parser/mod.rs
mvzink: Neither was I until I accidentally tried it...

mvzink (approved):
Great, thanks!

src/parser/mod.rs
```diff
@@ -7061,19 +7061,22 @@ impl<'a> Parser<'a> {
...
         };
+
         let table_name = self.parse_object_name(false)?;
-        let using = if self.parse_keyword(Keyword::USING) {
-            Some(self.parse_index_type()?)
-        } else {
-            None
-        };
+
+        using = self.parse_optional_using_then_index_type()?.or(using);
```
iffyio, line 7083: Should this be something like 
```rust
using = using.or_else(|| self.parse_optional_using_then_index_type()).transpose()?;
```
in that we should only look to parse again if we didn't parse a value already?

MohamedAbdeen21: As Michael pointed out earlier, MySQL does allow having two "USING" clauses. In that case, the second one overwrites the first.

mvzink: Might be worth a comment on this line, because at first glance (or if you haven't tried it on MySQL yourself) it does seem backwards 😛

iffyio: Ah I see, double checking do the tests include the scenario where two `USING` clauses are specified? If not we can probably add those

MohamedAbdeen21: Good point. Added a second test with two `USING` clauses. Force-pushed to resolve conflicts and squashed the commits.

tests/sqlparser_common.rs
```diff
@@ -17246,3 +17246,47 @@ fn parse_invisible_column() {
         _ => panic!("Unexpected statement {stmt}"),
     }
 }
+
+#[test]
+fn parse_create_index_using_before_on() {
+    let sql = "CREATE INDEX idx_name USING BTREE ON table_name (col1)";
+    // Can't use `verified_stmt` here as the USING will be placed after the `ON` clause
+    match all_dialects().parse_sql_statements(sql).unwrap()[0].clone() {
```
iffyio: for the tests, I think we can use `one_statement_parses_to()` instead so that the test covers the display impl as well?
Also can we merge both test functions into the same `parse_create_index_using()` or similar function - since they cover the same feature

src/parser/mod.rs
```diff
@@ -7118,6 +7123,13 @@ impl<'a> Parser<'a> {
         // parse it anyway (as we do inside `ALTER TABLE` and `CREATE TABLE` parsing).
         let index_options = self.parse_index_options()?;
 
+        if index_options
+            .iter()
+            .any(|opt| matches!(opt, IndexOption::Using(_)))
+        {
+            using = None;
+        };
+
```
MohamedAbdeen21: USING can be used in 3 different positions across the other dialects. This makes sure that only the latest USING is recognized.

tests/sqlparser_common.rs
MohamedAbdeen21: Updated tests and one other small change for the third position of USING.

tests/sqlparser_common.rs
```diff
@@ -17246,3 +17246,46 @@ fn parse_invisible_column() {
...
+            assert_eq!(table_name.to_string(), "table_name");
+            assert_eq!(using, Some(IndexType::BTree));
+            assert_eq!(columns.len(), 1);
+            assert!(!unique);
+        }
+        _ => unreachable!(),
+    }
+
+    let sql = "CREATE INDEX idx_name USING BTREE ON table_name (col1) USING HASH";
+    let expected = "CREATE INDEX idx_name ON table_name(col1) USING HASH";
```
iffyio: I'm not sure I followed the intent, why do we drop the first clause vs keeping both as in the input?

MohamedAbdeen21: We can, but it's just super confusing because there is the `USING` field and another IndexOption::IndexType in the `index_options` field.

The point here is to mimic mysql behaviour of only keeping the second, not preserve the input.

iffyio: hmm what's confusing do you mean with the first clause going into the `using`  field and the second clause going into the `index_options` field?
It feels like it'd be unusual for the parser to swallow some of its input, and in general we avoid mimicking server behavior in the parser, so its currently unclear to me what the rationale is

MohamedAbdeen21: I guess we can keep both then, but it will be confusing for users having two index types.

And while I don't like how we have two separate fields for that, I really don't want to hold this PR any longer

MohamedAbdeen21: updated

iffyio:
@MohamedAbdeen21 could you resolve the conflicts on the branch when you get the chance?

iffyio (approved):

PR #2031, opened 2025-09-16
MySQL: Add support for unsigned numeric types

MohamedAbdeen21:
Adds support for MySQL's deprecated UNSIGNED keyword syntax for some numeric types. This includes:
- DECIMAL UNSIGNED, DEC UNSIGNED
- FLOAT UNSIGNED
- DOUBLE UNSIGNED
- REAL UNSIGNED, DOUBLE PRECISION UNSIGNED

Also, allow FLOAT to accept precision and scale, not just precision.

docs: https://dev.mysql.com/doc/refman/8.4/en/numeric-type-syntax.html


mvzink (approved):
Seems good to me; only thing I might suggest is checking how these types behave (or don't) in the context of `CAST`.

iffyio (approved):

PR #2032, opened 2025-09-16
Add AUTO JOIN support

justhaug:


CHANGELOG.md
```diff
@@ -27,7 +27,9 @@ Given that the parser produces a typed AST, any changes to the AST will
 technically be breaking and thus will result in a `0.(N+1)` version. 
 
 
-- Unreleased: Check https://github.com/sqlparser-rs/sqlparser-rs/commits/main for undocumented changes.
+- Unreleased:
```
iffyio, line 30: I think we can revert this diff, the changelog is handled separately upon release.

src/ast/query.rs
```diff
@@ -2267,6 +2267,7 @@ impl fmt::Display for Join {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         fn prefix(constraint: &JoinConstraint) -> &'static str {
             match constraint {
+                JoinConstraint::Auto => "AUTO ",
```
iffyio, line 2270: Hi @justhaug could you add some context around this syntax and a link to the documentation that supports it?

justhaug:
Sorry about that, I accidentally let codex-cli off its leash and it made this PR.

This is for our own database, not standard SQL syntax I was planning on maintaining a branch for this for our own use.

Like natural join, but it matches on primary key - foreign key automatically regardless of name e.g.:

SELECT * FROM customers AUTO JOIN orders

is same as:

SELECT * FROM customers JOIN orders ON customers.id = orders.customer_id

If other databases implement this I'd be happy to make a proper PR then :)

PR #2033, opened 2025-09-17
Add support for INVISIBLE columns in MySQL

altmannmarcelo:
This commit adds support for INVISIBLE columns in MySQL dialect.

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -4247,33 +4247,3 @@ fn test_create_index_options() {
         "CREATE INDEX idx_name ON t(c1, c2) USING BTREE LOCK = EXCLUSIVE ALGORITHM = DEFAULT",
     );
 }
-
-#[test]
-fn parse_invisible_column() {
-    mysql().verified_stmt("CREATE TABLE t (foo INT, bar INT INVISIBLE)");
-    mysql().verified_stmt("ALTER TABLE t ADD COLUMN bar INT INVISIBLE");
-    let stmt = mysql().verified_stmt("CREATE TABLE t (foo INT, bar INT INVISIBLE)");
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -8144,6 +8144,11 @@ impl<'a> Parser<'a> {
                     Keyword::REPLACE,
                 ])?,
             )))
+        } else if dialect_of!(self is MySqlDialect | GenericDialect)
+            && self.parse_keyword(Keyword::INVISIBLE)
```
```suggestion
        } else if self.parse_keyword(Keyword::INVISIBLE)
```
would it make sense to let the parser accept this syntax if it shows up regardless of dialect? we're trying to move away from the `dialect_of` macro and if the parser can be more permissive in this case I think it would be ideal

altmannmarcelo: The only downside I see is that if you try this syntax on (let's say) PostgreSQL it will fail, but if you try it on sqlparser acting as the PG parser it will accept the invalid syntax. If that is ok I can remove the dialect check.

altmannmarcelo: Done.

tests/sqlparser_mysql.rs
```diff
@@ -4247,3 +4247,33 @@ fn test_create_index_options() {
         "CREATE INDEX idx_name ON t(c1, c2) USING BTREE LOCK = EXCLUSIVE ALGORITHM = DEFAULT",
     );
 }
+
+#[test]
+fn parse_invisible_column() {
```
iffyio: Since it has now become dialect agnostic I think we can move the test to common.rs and use `verified_stmt()` there so that it runs on all dialects

altmannmarcelo: Done.

PR #2034, opened 2025-09-17
Add support for C-style comments

altmannmarcelo:
This commit adds support for C-style comments supported by MySQL. It parses and consumes the optional version number after the `!` character.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -1086,6 +1086,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports optimizer hints in multiline comments
+    /// e.g. `/*!50110 KEY_BLOCK_SIZE = 1024*/`
+    fn supports_multiline_comment_hints(&self) -> bool {
+        false
+    }
+
     /// Returns true if this dialect supports treating the equals operator `=` within a `SelectItem`
     /// as an alias assignment operator, rather than a boolean expression.
     /// For example: the following statements are equivalent for such a dialect:
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -945,10 +945,65 @@ impl<'a> Tokenizer<'a> {
         while let Some(token) = self.next_token(&mut state, buf.last().map(|t| &t.token))? {
             let span = location.span_to(state.location());
 
-            buf.push(TokenWithSpan { token, span });
+            // Check if this is a multiline comment hint that should be expanded
+            match &token {
+                Token::Whitespace(Whitespace::MultiLineComment(comment))
+                    if self.dialect.supports_multiline_comment_hints()
+                        && comment.starts_with('!') =>
+                {
+                    // Re-tokenize the hints and add them to the buffer
+                    self.tokenize_comment_hints(comment, span, buf)?;
+                }
+                _ => {
+                    buf.push(TokenWithSpan { token, span });
+                }
+            }
 
             location = state.location();
         }
+        Ok(())
+    }
+
+    /// Re-tokenize optimizer hints from a multiline comment and add them to the buffer.
+    /// For example, `/*!50110 KEY_BLOCK_SIZE = 1024*/` becomes tokens for `KEY_BLOCK_SIZE = 1024`
+    fn tokenize_comment_hints(
+        &self,
+        comment: &str,
+        span: Span,
+        buf: &mut Vec<TokenWithSpan>,
+    ) -> Result<(), TokenizerError> {
+        // Strip the leading '!' and any version digits (e.g., "50110")
+        let hint_content = comment
+            .strip_prefix('!')
+            .unwrap_or(comment)
+            .trim_start_matches(|c: char| c.is_ascii_digit());
+
+        // If there's no content after stripping, nothing to tokenize
+        if hint_content.is_empty() {
+            return Ok(());
+        }
+
+        // Create a new tokenizer for the hint content
+        let inner = Tokenizer::new(self.dialect, hint_content).with_unescape(self.unescape);
+
+        // Create a state for tracking position within the hint
+        let mut state = State {
+            peekable: hint_content.chars().peekable(),
+            line: span.start.line,
+            col: span.start.column,
+        };
+
+        // Tokenize the hint content and add tokens to the buffer
+        let mut location = state.location();
+        while let Some(token) = inner.next_token(&mut state, buf.last().map(|t| &t.token))? {
+            let token_span = location.span_to(state.location());
+            buf.push(TokenWithSpan {
+                token,
+                span: token_span,
+            });
+            location = state.location();
+        }
+
         Ok(())
     }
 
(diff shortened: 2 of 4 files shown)


src/dialect/mod.rs
```diff
@@ -898,6 +898,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports hint and C-style comments
```
iffyio: Can we add an example here? I think it won't be clear otherwise what this means by only reading the description

src/tokenizer.rs
```diff
@@ -4070,4 +4101,39 @@ mod tests {
             panic!("Tokenizer should have failed on {sql}, but it succeeded with {tokens:?}");
         }
     }
+    #[test]
+    fn tokenize_multiline_comment_with_c_style_comment() {
+        let sql = String::from("0/*! word */1");
```
iffyio, line 4280: Looking at their docs, I'm wondering if/how we support these examples?
```sql
SELECT /*! STRAIGHT_JOIN */ col1 FROM table1,table2
/*!50110 KEY_BLOCK_SIZE=1024 */
SELECT /*! BKA(t1) */ FROM T
```

altmannmarcelo: @iffyio - Parsing the c_style comment unblocks sqlparser to not discard those as if they were a normal comment. Support for each hint will have to be added in a case by case bases. For example #2033 - MySQL adds a c-style comment if you run SHOW CREATE TABLE:

```
mysql> SHOW CREATE TABLE b;
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                    |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| b     | CREATE TABLE `b` (
  `ID` int DEFAULT NULL,
  `b` char(1) DEFAULT NULL /*!80023 INVISIBLE */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0,008 sec)
```

Without the current patch, the invisible keyword will be discarded.

src/dialect/mod.rs
```diff
@@ -898,6 +898,12 @@ pub trait Dialect: Debug + Any {
...
+    /// e.g. `/*! hint */`
+    fn supports_c_style_comments(&self) -> bool {
```
```suggestion
    fn supports_c_style_comment_hints(&self) -> bool {
```
Since c_style comment I imagine is referring to `/**/` which other dialects support?

src/tokenizer.rs
iffyio: Ah so to clarify I'm rather wondering regarding the parser behavior for hints that aren't singe words e.g. `/*!50110 KEY_BLOCK_SIZE=1024 */` - can we demonstrate the behavior with test cases for such scenarios?

altmannmarcelo: @iffyio thanks flagging this. I have fixed the issue and now we properly return individual tokens inside a C-style hint comment.

src/dialect/mod.rs
```diff
@@ -898,6 +898,12 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports hint and C-style comments
+    /// e.g. `/*! hint */`
+    fn supports_c_style_hints(&self) -> bool {
```
```suggestion
    /// Returns true if the dialect supports optimizer hints in multiline comments
    /// e.g. `/*!50110 KEY_BLOCK_SIZE = 1024*/`
    fn supports_multiline_comment_hints(&self) -> bool {
```

src/tokenizer.rs
```diff
@@ -936,10 +939,16 @@ impl<'a> Tokenizer<'a> {
 
     /// Get the next token or return None
     fn next_token(
-        &self,
+        &mut self,
         chars: &mut State,
         prev_token: Option<&Token>,
     ) -> Result<Option<Token>, TokenizerError> {
+        // Return any previously injected tokens first
```
iffyio: impl wise I'm thinking we can do something like the following that would be less invasive to the tokenizer:

If we take as an example `/*!50110 KEY_BLOCK_SIZE = 1024*/` - today that will get parse as a `Token::MultilineComment("!50110 KEY_BLOCK_SIZE = 1024")` - which we receive [here in the tokenizer loop](https://github.com/altmannmarcelo/datafusion-sqlparser-rs/blob/f96249cc1d52d34edcfb0ca246bdd78f62552b91/src/tokenizer.rs#L907).

Then the idea would be to essentially re-tokenize that string and add to the tokens buffer.

So something like this:

```rust
// we check if this is a token containing optimizer hints
match token {
	Token::Whitespace(Whitespace::MultiLineComment(comment)) if self.dialect.supports_multiline_comment_hints() && s.starts_with("!") => {
    // re-tokenize the hints
	buf.extend(tokenize_comment_hints(comment, span)?);
	token => {
		buf.push(TokenWithSpan{ token, span })
	}
}

// here we reuse existing tokenizer machinery to re-tokenize the hints.
fn tokenize_comment_hints(&self, hints: String, span: Span, tokens: &mut Vec<TokenWithSpan>) -> Result<()> {
	let mut state = State {
		peekable: hints.chars().peekable(),
		line: span.start.line,
		col: span.start.column,
	};

	while let Some(token) = self.next_token(&mut state, None)? {
		tokens.push(token);
	}

	Ok(())
}
```

src/dialect/mod.rs
iffyio: Ah to clarify the suggested change, c-style comment I think is rather understood as  `/* */` - i.e. multiline comment in general, the feature here I think is the optimizer hints within comments as mysql calls them

altmannmarcelo:
I will pick this up in a few days.

altmannmarcelo:
@iffyio sorry for my delay on getting back - I have implemented the re-tokenize function.

src/tokenizer.rs
```diff
@@ -980,7 +1036,7 @@ impl<'a> Tokenizer<'a> {
 
     /// Get the next token or return None
     fn next_token(
-        &self,
+        &mut self,
```
```suggestion
        &self,
```
the `mut` ref doesn't seem like it's being used?

src/tokenizer.rs
```diff
@@ -2227,13 +2283,12 @@ impl<'a> Tokenizer<'a> {
     }
 
     fn tokenize_multiline_comment(
-        &self,
+        &mut self,
```
```suggestion
        &self,
```
similarly, looks unused?

src/tokenizer.rs
```diff
@@ -4218,6 +4273,47 @@ mod tests {
...
+                keyword: Keyword::NoKeyword,
+            }),
+            Token::Number("1".to_string(), false),
+        ];
+        compare(expected, tokens);
+    }
+
+    #[test]
+    fn tokenize_multiline_comment_with_comment_hint_and_version() {
+        let sql_multi = String::from("0 /*!50110 KEY_BLOCK_SIZE = 1024*/ 1");
```
iffyio, line 4300: can we add a similar test cases where the comment has an empty hint? e.g.
```
0 /*!50110 */
0 /*!*/
0 /*!   */
```

PR #2035, opened 2025-09-17
Snowflake: ALTER USER and KeyValueOptions Refactoring

yoavcloud:
This PR adds support for the `ALTER USER` statement with specific support for the Snowflake options. In addition, the `KeyValueOptions` struct was refactored:
1. Used the `Value` type instead of its custom types for option values
2. Added support for multi-value options, i.e. `KEY=(1,2,3)`
3. Added support for nested key-value options, i.e. `KEY=(K1=1, K2=2)`


alamb:
Perhaps this would be a good PR for @yoavcloud to try and merge to ensure the credentials are setup correctly

src/ast/helpers/key_value_options.rs
```diff
@@ -41,6 +39,13 @@ pub struct KeyValueOptions {
     pub delimiter: KeyValueOptionsDelimiter,
 }
 
+impl KeyValueOptions {
+    /// Returns true iff the options list is empty
+    pub fn is_empty(&self) -> bool {
+        self.options.is_empty()
+    }
+}
+
```
iffyio: I'm thinking we could skip the added API and have the caller call `value.options.is_empty()` instead? since this is only a thin wrapper might not be worth it

src/parser/mod.rs
```diff
@@ -17463,40 +17471,67 @@ impl<'a> Parser<'a> {
...
+                    option_value: KeyValueOptionKind::Single(Value::Placeholder(
+                        word.value.clone(),
+                    )),
+                })
+            }
+            Token::LParen => {
+                // Can be a list of values or a list of key value properties.
+                // Try to parse a list of values and if that fails, try to parse
+                // a list of key-value properties.
+                match self.try_parse(|parser| {
```
iffyio: should this use `maybe_parse` instead of `try_parse` (noticed we're ignoring the error from which on first glance looks unintentional)?

yoavcloud:
@iffyio please take a look, and if OK I'll try to merge using my creds

src/ast/mod.rs
```diff
@@ -10558,6 +10564,199 @@ impl fmt::Display for CreateUser {
...
+    pub enroll_mfa: bool,
+    pub set_default_mfa_method: Option<MfaMethodKind>,
+    pub remove_mfa_method: Option<MfaMethodKind>,
+    pub modify_mfa_method: Option<AlterUserModifyMfaMethod>,
+    pub set_policy: Option<AlterUserSetPolicy>,
+    pub unset_policy: Option<UserPolicyKind>,
+    pub set_tag: KeyValueOptions,
+    pub unset_tag: Vec<String>,
+    pub set_props: KeyValueOptions,
+    pub unset_props: Vec<String>,
```
iffyio, line 10597: @yoavcloud one thing I realised now with these props, could we add a link tho the snowflake docs individually for them? Thinking so that when new options are inevitably added for other dialects it would be clear where each propery comes from

yoavcloud: Unfortunately Snowflake docs doesn't have a link to each, but we can group them all together. I added a comment about that.

yoavcloud:
@alamb I don't see the merge option, even though the PR has been approved. Who should I check this with?

alamb:
> @alamb I don't see the merge option, even though the PR has been approved. Who should I check this with?

@yoavcloud  -- perhaps you need to link your github and apache account. To do so, follow the instructions at [gitbox.apache.org](https://gitbox.apache.org/)

alamb:
Screenshot 2025-09-25 at 4 00 27 PM

It's working!

yoavcloud:
That is soooo cool! Thank you @alamb and @iffyio!

alamb:
> That is soooo cool! Thank you @alamb and @iffyio!

Welcome aboard!

Issue #2023, opened 2025-09-07
CREATE TRIGGER support for SQLite syntax


LucaCappelletti94:
At the time of writing, as described earlier in issues such as #1388, the [CREATE TRIGGER](https://sqlite.org/lang_createtrigger.html) syntax for SQLite is not supported.

For instance, once this feature is fully supported, I expect the following snippet to execute, while at this time it crashes with `ParserError("Expected: an object type after CREATE, found: TRIGGER at Line: 1, Column: 8")`.

I will proceed with a PR attempt after determining what to do with #2020 and #2022.

```rust
let to_parse = r#"CREATE TRIGGER trg_inherit_asset_models
AFTER INSERT ON parent_procedure_templates
FOR EACH ROW
BEGIN
    INSERT INTO procedure_template_asset_models (
        name,
        procedure_template,
        based_on,
        asset_model,
        created_by,
        created_at
    )
    SELECT
        pam.name,
        NEW.parent,
        pam.id,
        pam.asset_model,
        NEW.created_by,
        NEW.created_at
    FROM procedure_template_asset_models pam
    WHERE pam.procedure_template = NEW.child;
END;
"#;
let stmt =
  sqlparser::parser::Parser::parse_sql(&sqlparser::dialect::SQLiteDialect {}, to_parse)
      .unwrap();
```

LucaCappelletti94:
Closing as solved in PR #2037

PR #2037, opened 2025-09-18
Added support for SQLite triggers

LucaCappelletti94:
This PR:

- Enables`CREATE TRIGGER` for SQLite dialect, which was early blocked by a dialect guard
- Adds support for TEMPORARY triggers - at this time, I have not added a dialect guard on the use of the `TEMPORARY` keyword, which may be desirable if we want to raise errors in other dialect if `TEMP` triggers are provided.
- Adds support for optional `FOR EACH ROW` in triggers, which is optional in SQLite
- Adds test suite for several `CREATE TRIGGER` cases in SQLite.
- Enables `DROP TRIGGER` for SQLite dialect, which was early blocked by a dialect guard
- Adds test suite for a `DROP TRIGGER` case in SQLite.

Closes issue #2023


src/ast/ddl.rs
```diff
@@ -3262,7 +3280,9 @@ pub struct CreateTrigger {
     pub referencing: Vec<TriggerReferencing>,
     /// This specifies whether the trigger function should be fired once for
     /// every row affected by the trigger event, or just once per SQL statement.
-    pub trigger_object: TriggerObject,
+    /// This is optional in some SQL dialects, such as SQLite, and if not specified, in
+    /// those cases, the implied default is `FOR EACH ROW`.
```
iffyio, line 3049:
```suggestion
```
Since the optionality is clear from the type and we can avoid documenting semantic implications

src/ast/ddl.rs
```diff
@@ -3243,14 +3259,16 @@ pub struct CreateTrigger {
...
     /// ```sql
-    /// -- period_before_table == true: Postgres, MySQL, and standard SQL
+    /// -- period_specified_before_table == true: Postgres, MySQL, and standard SQL
     /// CREATE TRIGGER t BEFORE INSERT ON table_name ...;
-    /// -- period_before_table == false: MSSQL
+    /// -- period_specified_before_table == false: MSSQL
     /// CREATE TRIGGER t ON table_name BEFORE INSERT ...;
     /// ```
-    pub period_before_table: bool,
+    pub period_specified_before_table: bool,
```
iffyio: was there a requirement to rename this variable?

src/ast/ddl.rs
```diff
@@ -3199,6 +3199,22 @@ pub struct CreateTrigger {
     ///
     /// [MsSql](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql?view=sql-server-ver16#arguments)
     pub or_alter: bool,
+    /// True if this is a temporary trigger, which is supported in SQLite.
```
```suggestion
    /// True if this is a temporary trigger.
```

src/ast/ddl.rs
```diff
@@ -3199,6 +3199,22 @@ pub struct CreateTrigger {
...
+    /// CREATE TEMP TRIGGER trigger_name
+    /// ```
+    ///
+    /// or
+    ///
+    /// ```sql
+    /// CREATE TEMPORARY TRIGGER trigger_name
+    /// ```
+    ///
+    /// [Temporary Triggers in SQLite](https://sqlite.org/lang_createtrigger.html#temp_triggers_on_non_temp_tables)
```
```suggestion
    /// Examples:
    ///
    /// ```sql
    /// CREATE TEMP TRIGGER trigger_name
    /// ```
    ///
    /// or
    ///
    /// ```sql
    /// CREATE TEMPORARY TRIGGER trigger_name;
    /// CREATE TEMP TRIGGER trigger_name;
    /// ```
    ///
    /// [SQLite](https://sqlite.org/lang_createtrigger.html#temp_triggers_on_non_temp_tables)
```

src/parser/mod.rs
```diff
@@ -5574,11 +5575,14 @@ impl<'a> Parser<'a> {
...
     pub fn parse_create_trigger(
         &mut self,
+        temporary: bool,
         or_alter: bool,
         or_replace: bool,
         is_constraint: bool,
     ) -> Result<Statement, ParserError> {
-        if !dialect_of!(self is PostgreSqlDialect | GenericDialect | MySqlDialect | MsSqlDialect) {
+        if !dialect_of!(self is PostgreSqlDialect | SQLiteDialect | GenericDialect | MySqlDialect | MsSqlDialect)
+            || dialect_of!(self is SQLiteDialect) && (or_alter || or_replace || is_constraint)
```
iffyio: What is the intent of the `&& (or_alter ..)` part? not sure I fully followed

src/parser/mod.rs
```diff
@@ -5605,14 +5609,24 @@ impl<'a> Parser<'a> {
...
+                        Keyword::ROW => TriggerObject::Row,
+                        Keyword::STATEMENT => TriggerObject::Statement,
+                        _ => unreachable!(),
+                    },
+                ),
+            )
+        } else {
+            if !dialect_of!(self is SQLiteDialect ) {
+                self.expect_keyword_is(Keyword::FOR)?;
+            }
```
```suggestion
            self.expect_keyword_is(Keyword::FOR)?;
```
Can we skip the dialect guard here?

src/ast/ddl.rs
LucaCappelletti94: I thought it was important to specify why it is optional, as otherwise other developers which are not familiar with the peculiarities of SQLite might in the future attempt to remove the `Option`. I have added a test for this case, sure, but I believe it best to put this information immediately next to the type.

src/ast/ddl.rs
LucaCappelletti94: I was finding myself mistaking the meaning of the attribute for: "This boolean specified whether the TRIGGER is triggered BEFORE the insertion of the TABLE" instead of the more simple "the period when this trigger is activated is specified before the TABLE in the SQL", hence the name change to improve clarity.

src/ast/ddl.rs
LucaCappelletti94: Since it is solely supported in SQLite (that I know of), I thought it important to document it.

src/parser/mod.rs
LucaCappelletti94: SQLite does not support TRIGGERS with ALTER, REPLACE or constraints. In such cases, an error should be raised.

src/ast/ddl.rs
iffyio: The Sqlite bit is documented via the documentation reference, in the description header I think its easier to skip enumerating dialects so that the description doesn't change or go out of sync when new dialects are added

src/ast/ddl.rs
iffyio: Ah I see, I think we can revert the diff since it might not be worth breaking the API.

src/ast/ddl.rs
iffyio: > the implied default is `FOR EACH ROW`
If I'm not mistaken, this is rather describing the behavior of the sqlserver rather than the implied default of the parser? If so we try to avoid semantic descriptions in general since those vary vastly, if a reader is interested in such implications we could include a reference to the sqlite docs.

Also I think a clearer representation in that case actually would be something like the following?
```rust
enum TriggerObjectKind {
    For(TriggerObject),
    ForEach(TriggerObject),
}
pub trigger_object: Option<TriggerObjectKind>
```
That would let us skip the extra boolean `include_each` it looks like?

src/parser/mod.rs
iffyio: I think we can skip the flags and let the parser be permissive, downstream crates maybe validate the AST further when necessary (it would be too much of a large effort for the parser to treat each dialect individually in such a granular manner otherwise).

LucaCappelletti94:
Regarding commit [c893bda](https://github.com/apache/datafusion-sqlparser-rs/pull/2037/commits/c893bda671a2dcf90cb0b7c22270782bcaaaa5fa), it should be noted that the `MsSQL` parser for triggers is incorrect, as it does not include [much of their syntax](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql?view=sql-server-ver17), and currently always adds in a `FOR STATEMENT`, which is incorrect.  This is clearly way outside the current PR objectives though. I have patched the parse trigger so that it would be compatible with the `MsSQL` tests, but I am completely unfamiliar with that dialect so I would leave a more refined extension to somebody more familiar with it.

src/parser/mod.rs
iffyio: Seems this comment is unresolved? ideally we want to skip dialect_of checks where possible

LucaCappelletti94:
Is there anything else needed to be done for this PR? @iffyio

iffyio:
@LucaCappelletti94 could you take a look at [this comment](https://github.com/apache/datafusion-sqlparser-rs/pull/2037#discussion_r2362178757)?

LucaCappelletti94:
Hi @iffyio, I replied to it here: https://github.com/apache/datafusion-sqlparser-rs/pull/2037#discussion_r2379300179

iffyio:
@LucaCappelletti94 [the link](https://github.com/apache/datafusion-sqlparser-rs/pull/2037#discussion_r2379300179) takes me to the top of the page not sure why but I'm unable to find the comment in this case

LucaCappelletti94:
> @LucaCappelletti94 [the link](https://github.com/apache/datafusion-sqlparser-rs/pull/2037#discussion_r2379300179) takes me to the top of the page not sure why but I'm unable to find the comment in this case

@iffyio Here follows the comment:

> So, if you mean to refactor this into:
> 
> ```rust
> } else {
>     None
> };
> ```
> 
> would simply make it more permissive and allow for parsing stuff that does not make sense in some dialects.
> 
> Conversely, the following:
> 
> ```rust
> } else {
>     self.expect_keyword_is(Keyword::FOR)?;
> 
>     None
> };
> ```
> 
> would crash for valid SQLite statements.

src/parser/mod.rs
iffyio: @LucaCappelletti94 ah re the [comment](https://github.com/apache/datafusion-sqlparser-rs/pull/2037#issuecomment-3381061328) I think its to essentially do something like the following where the parser is more permissive
```rust
} else {
    self.parse_keyword(Keyword::FOR)?;

    None
};
```

iffyio (approved):

PR #2041, opened 2025-09-19
Add support for procedure parameter default values

aharpervc:
For example, this now parses appropriately:

```sql
create procedure test_proc (a int = 42)
...
```

Syntax documentation reference: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-procedure-transact-sql?view=sql-server-ver17#e-use-a-procedure-with-wildcard-parameters. This is the syntax used by SQL Server, but I've implemented this as a "normal" Option property of the ProcedureParam struct.

I also added a corresponding common unit test

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -16535,10 +16535,8 @@ fn parse_create_procedure_with_parameter_modes() {
         }
         _ => unreachable!(),
     }
-}
 
-#[test]
-fn create_procedure_with_parameter_default_value() {
+    // parameters with default values
     let sql = r#"CREATE PROCEDURE test_proc (IN a INTEGER = 1, OUT b TEXT = '2', INOUT c TIMESTAMP = NULL, d BOOL = 0) AS BEGIN SELECT 1; END"#;
     match verified_stmt(sql) {
         Statement::CreateProcedure {
@@ -16549,29 +16547,17 @@ fn create_procedure_with_parameter_default_value() {
         } => {
             assert_eq!(or_alter, false);
             assert_eq!(name.to_string(), "test_proc");
-            let fake_span = Span {
-                start: Location { line: 0, column: 0 },
-                end: Location { line: 0, column: 0 },
-            };
             assert_eq!(
                 params,
                 Some(vec![
                     ProcedureParam {
-                        name: Ident {
-                            value: "a".into(),
-                            quote_style: None,
-                            span: fake_span,
-                        },
+                        name: Ident::new("a"),
                         data_type: DataType::Integer(None),
                         mode: Some(ArgMode::In),
                         default: Some(Expr::Value((number("1")).with_empty_span())),
                     },
                     ProcedureParam {
-                        name: Ident {
-                            value: "b".into(),
-                            quote_style: None,
-                            span: fake_span,
-                        },
+                        name: Ident::new("b"),
                         data_type: DataType::Text,
                         mode: Some(ArgMode::Out),
                         default: Some(Expr::Value(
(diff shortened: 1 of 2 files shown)


tests/sqlparser_common.rs
```diff
@@ -16533,6 +16537,40 @@ fn parse_create_procedure_with_parameter_modes() {
     }
 }
 
+#[test]
+fn create_procedure_with_parameter_default_value() {
```
iffyio: Can we merge the test into the existing `parse_create_procedure_with_parameter_modes`? since they cover the same feature

tests/sqlparser_common.rs
```diff
@@ -16533,6 +16537,40 @@ fn parse_create_procedure_with_parameter_modes() {
...
+    let sql = r#"CREATE PROCEDURE test_proc (a INT = 42) AS BEGIN SELECT 1; END"#;
```
iffyio: Can we add scenarios demonstrating
- multiple paramenters having defaults (i.e. `(a INT= 42, b INT=43)`)
- parameters that include a `mode`

tests/sqlparser_common.rs
aharpervc: Hm, I don't understand. A parameter can be in/out and still have/not have a default, right? Are you suggesting like

```diff
pub enum ArgMode {
  In,
  Out,
  InOut,
+ InWithDefault(Expr),
+ OutWithDefault(Expr),
+ InOutWithDefault(Expr),
}
```

Perhaps I'm not understanding

tests/sqlparser_common.rs
aharpervc: Done 👍. I've also added an additional unit test for SQL Server, since the parameter identifier syntax is different there.

tests/sqlparser_common.rs
iffyio: Ah no I meant to only inline this test case into that `parse_create_procedure_with_parameter_modes` function vs having one function per test case

tests/sqlparser_common.rs
```diff
@@ -16533,6 +16537,74 @@ fn parse_create_procedure_with_parameter_modes() {
...
+            };
+            assert_eq!(
+                params,
+                Some(vec![
+                    ProcedureParam {
+                        name: Ident {
+                            value: "a".into(),
+                            quote_style: None,
+                            span: fake_span,
+                        },
```
```suggestion
                        name: Ident::new("a"),
```
I think this can be simplified? that would let us skip introducing the fake span as well

tests/sqlparser_mssql.rs
```diff
@@ -198,6 +200,12 @@ fn parse_mssql_create_procedure() {
     let _ = ms().verified_stmt("CREATE PROCEDURE [foo] AS BEGIN UPDATE bar SET col = 'test'; SELECT [foo] FROM BAR WHERE [FOO] > 10; END");
 }
 
+#[test]
+fn parse_mssql_create_procedure_with_parameter_default_value() {
+    let sql = r#"CREATE PROCEDURE foo (IN @a INTEGER = 1, OUT @b TEXT = '2', INOUT @c DATETIME = NULL, @d BOOL = 0) AS BEGIN SELECT 1; END"#;
+    let _ = ms().verified_stmt(sql);
+}
```
iffyio: Similar comment here that we can inline this scenario into the `parse_mssql_create_procedure` function above

tests/sqlparser_common.rs
aharpervc: Yes, certainly. I did it this way because that's how the similar modes test did it, but I will simplify

tests/sqlparser_common.rs
aharpervc: Oh yes, I understand now 👍

tests/sqlparser_mssql.rs
aharpervc: Done 👍

PR #2043, opened 2025-09-23
Support `CREATE SEQUENCE` options in any order

stevenliebregt:
This changes the parsing of the `CREATE SEQUENCE` options to allow the individual options to appear in any order.

The reason I'm opening this pull request is that the Postgres `pg_dump` utility dumps `CREATE SEQUENCE` statements like this:

```sql
CREATE SEQUENCE name6
AS INTEGER
START WITH 1
INCREMENT BY 1
NO MINVALUE
NO MAXVALUE
CACHE 1
```

Where the `START WITH 1` appears before the `INCREMENT BY 1`, Postgres allows this but the parser did not.

There should not be any noticable performance impact I think. I did run the benchmarks and did not see any impact.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -17019,44 +17019,68 @@ impl<'a> Parser<'a> {
 
     fn parse_create_sequence_options(&mut self) -> Result<Vec<SequenceOptions>, ParserError> {
         let mut sequence_options = vec![];
-        //[ INCREMENT [ BY ] increment ]
-        if self.parse_keywords(&[Keyword::INCREMENT]) {
-            if self.parse_keywords(&[Keyword::BY]) {
-                sequence_options.push(SequenceOptions::IncrementBy(self.parse_number()?, true));
-            } else {
-                sequence_options.push(SequenceOptions::IncrementBy(self.parse_number()?, false));
-            }
-        }
-        //[ MINVALUE minvalue | NO MINVALUE ]
-        if self.parse_keyword(Keyword::MINVALUE) {
-            sequence_options.push(SequenceOptions::MinValue(Some(self.parse_number()?)));
-        } else if self.parse_keywords(&[Keyword::NO, Keyword::MINVALUE]) {
-            sequence_options.push(SequenceOptions::MinValue(None));
-        }
-        //[ MAXVALUE maxvalue | NO MAXVALUE ]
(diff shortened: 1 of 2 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -309,6 +309,15 @@ fn parse_create_sequence() {
...
     ));
+
+    let sql9 = "CREATE SEQUENCE name6
+    AS INTEGER
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1";
+    pg().one_statement_parses_to(sql9, "CREATE SEQUENCE name6 AS INTEGER START WITH 1 INCREMENT BY 1 NO MINVALUE NO MAXVALUE CACHE 1");
```
iffyio, line 320:
```suggestion

    pg().verified_stmt("CREATE SEQUENCE name6 AS INTEGER START WITH 1 INCREMENT BY 1 NO MINVALUE NO MAXVALUE CACHE 1");
```
it looks like this should be equivalent?

src/parser/mod.rs
```diff
@@ -17019,44 +17019,68 @@ impl<'a> Parser<'a> {
...
+        // To disallow duplicate options
+        let has_increment = false;
+        let has_minvalue = false;
+        let has_maxvalue = false;
+        let has_start = false;
+        let has_cache = false;
+        let has_cycle = false;
+
+        let mut should_continue_loop = true;
+        while should_continue_loop {
```
iffyio, line 17032:
```suggestion
        loop {
```
It looks like we can skip the bool var and instead use a combination of `continue` and `break`  to manage the loop?

tests/sqlparser_postgres.rs
```diff
@@ -309,6 +309,15 @@ fn parse_create_sequence() {
         pg().parse_sql_statements("CREATE SEQUENCE foo INCREMENT 1 NO MINVALUE NO"),
         Err(ParserError::ParserError(_))
     ));
+
+    let sql9 = "CREATE SEQUENCE name6
```
iffyio, line 313: Can we add a test case with duplicate options?

stevenliebregt:
@iffyio thanks for the review, I'll be unable to address the comments for the next 2 weeks, but after that I'll address them

PR #2047, opened 2025-09-25
[databricks] update dialect to support grouping by with modifier

n-young:
This syntax is supported by Databricks; see here: https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-select-groupby


src/dialect/databricks.rs
```diff
@@ -64,4 +64,9 @@ impl Dialect for DatabricksDialect {
     fn supports_struct_literal(&self) -> bool {
         true
     }
+
+    // See https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-groupby.html
```
iffyio:
```suggestion
    /// See https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-groupby.html
```

iffyio:
@n-young could you fix the ci failures when you get the chance?

Issue #2053, opened 2025-10-03
Converting `TableConstraint` struct enum variants into separated structs


LucaCappelletti94:
Hi,

I need to implement traits describing for instance `CheckConstraint` and `UniqueConstraint` and so on for various crates, including sqlparser.

These objects are currently represented in the `TableConstraint` enum. I believe it would be better to split them into their own structs. 

I will open up a PR addressing this, but do let me know whether you agree that it is desirable.

Best,
Luca

LucaCappelletti94:
Closed as completed by merging PR #2054

PR #2054, opened 2025-10-03
Moved constraint variant outside of `TableConstraint` enum

LucaCappelletti94:
As per title, this PR moved all of the struct variants out of the `TableConstraint` enum. This is done to allow for implementing traits in dependent crates which only apply to some of the variants and not all of them.

Furthermore it:

- Implemented `Display` and `Spanned` for each of the new structs
- Implemented `From<variant> for TableConstraint` for all variants
- Documented the variants and updated tests

This PR closes issue #2053

Ciao,
Luca


src/ast/ddl.rs
```diff
@@ -2065,7 +1902,7 @@ pub enum GeneratedExpressionMode {
 }
 
 #[must_use]
-fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
+pub fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
```
```suggestion
pub(crate) fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
```

src/ast/ddl.rs
```diff
@@ -2082,7 +1919,7 @@ fn display_constraint_name(name: &'_ Option<Ident>) -> impl fmt::Display + '_ {
 /// * `Some(inner)` => create display struct for `"{prefix}{inner}{postfix}"`
 /// * `_` => do nothing
 #[must_use]
-fn display_option<'a, T: fmt::Display>(
+pub fn display_option<'a, T: fmt::Display>(
```
```suggestion
pub(crate) fn display_option<'a, T: fmt::Display>(
```

src/ast/ddl.rs
```diff
@@ -2104,7 +1941,7 @@ fn display_option<'a, T: fmt::Display>(
 /// * `Some(inner)` => create display struct for `" {inner}"`
 /// * `_` => do nothing
 #[must_use]
-fn display_option_spaced<T: fmt::Display>(option: &Option<T>) -> impl fmt::Display + '_ {
+pub fn display_option_spaced<T: fmt::Display>(option: &Option<T>) -> impl fmt::Display + '_ {
```
```suggestion
pub(crate) fn display_option_spaced<T: fmt::Display>(option: &Option<T>) -> impl fmt::Display + '_ {
```

src/ast/table_constraints/check_constraint.rs
```diff
@@ -0,0 +1,67 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
iffyio: I think we can use a single file `table_constraints.rs` to contain the new structs, vs having one struct per file. So that we don't end up accumulating a lot of small files which would be the other extreme of where we are today

LucaCappelletti94: Ok - should the table_constraints.rs file contain the `TableConstraint` enum itself?

LucaCappelletti94: Done in commit [258cc8c](https://github.com/apache/datafusion-sqlparser-rs/pull/2054/commits/258cc8cb65603824c86a7e21089a426e021abcf8)

iffyio (approved):

PR #2057, opened 2025-10-06
Moving several struct variants out of `Statement` enum to allow for trait impls for specific sub-variants

LucaCappelletti94:
As per title, in this PR I moved several of the struct variants out of the `Statement` enum. This refactoring allows for implementing traits and methods which specifically expect one particular statement, and not all of them.

I have focused on the variants which I am familiar with, primarily those related to PostgreSQL. There are still other struct variant which we might want to move out, such as `Drop`, but that variant would need to be renamed into some other naming convention because of the collision with the [`Drop`](https://doc.rust-lang.org/std/ops/trait.Drop.html) trait.

This PR works towards solving issues such as #1204

Best,
Luca


iffyio (approved):

iffyio:
@LucaCappelletti94 could you take a look at resolving the conflicts when you get the chance?

LucaCappelletti94:
> @LucaCappelletti94 could you take a look at resolving the conflicts when you get the chance?

Yeah, let's first wrap up also the sqlite trigger and then I will handle all of the merge conflicts.

Issue #2045, opened 2025-09-24
Release sqlparser-rs version `0.60.0` around 2025-11-15


alamb:
Follow on to 
- https://github.com/apache/datafusion-sqlparser-rs/issues/1956

This ticket tracks creating the next sqlparser release (mostly so others can follow along)

**Target Release Date**: Approximately 2025-11-15 (2 months after the last release)

Release instructions: https://github.com/apache/datafusion-sqlparser-rs/blob/main/dev/release/README.md

Steps:
- [ ] Tested with DataFusion: 
- [ ] Make a PR to update version / `CHANGELOG.md`
- [ ] Create a Release Candidate (RC) and start vote
- [ ] (When vote is approved)  release to crates.io
- [ ] Make ticket for next release: 

Required fixes: 
- [ ] None yet 🎉

romanoff:
Hi! Do we know yet when the next release branch will be created? Just planning around the upcoming release.

alamb:
Hi @romanoff  -- Sorry I lost track of this one. 

I can make a release whenever we want

@iffyio are there any PRs I should wait for before creating a release candidate?

iffyio:
@alamb nothing specific being waited on, I think we should be good for a new release yeah!

alamb:
I have made a PR to prepare for the release:
- https://github.com/apache/datafusion-sqlparser-rs/pull/2116

alamb:
I have created a release candidate and started a vote thread: https://lists.apache.org/thread/1c0z1n6dpdoscgh5xjl0drj2d8mslfvo

Assuming the vote passes, we can publish 0.60.0 in 3 days (72 hours)

alamb:
The release was officially approved
- https://lists.apache.org/thread/24skp3zk0b6mz4orvmjb0pbk1lwmf1go

The release is available here:
  https://dist.apache.org/repos/dist/release/datafusion/datafusion-sqlparser-rs-0.60.0


It is also on crates.io
https://crates.io/crates/sqlparser/0.60.0

PR #2060, opened 2025-10-08
Increase version of sqlparser_derive from 0.3.0 to 0.4.0

jjbayer:
Bump the version of the internal dependency so that users of the library can leverage the recursion protection added in https://github.com/apache/datafusion-sqlparser-rs/pull/1522.


alamb:
I will release 0.4.0 as part of
- https://github.com/apache/datafusion-sqlparser-rs/issues/2045

Issue #2061, opened 2025-10-08
Support for `MATCH PARTIAL`, `MATCH SIMPLE` and `MATCH FULL`


LucaCappelletti94:
Hi,

At the time of writing, [`MATCH [PARTIAL | SIMPLE | FULL]` syntax](https://www.postgresql.org/docs/current/sql-createtable.html) is not supported.

I will make a PR adding support for that.

Best,
Luca

LucaCappelletti94:
Closing as solved

Issue #2059, opened 2025-10-07
Align more uniformely the table constraints and column options


LucaCappelletti94:
Hi,

As it stands, the variants used for representing several table constraints or column options are different entries in different enums.

Once PRs like #2054 are merged, it will be possible to refactor the column options so that the `sqlparser::ast::ColumnOption::ForeignKey` variant becomes a simple tuple variant using the newly created `ForeignKeyConstraint`.

Is this desirable? If so, once the pending PRs are merged I can refactor those two enums to be more tightly aligned.

Best,
Luca

PR #2062, opened 2025-10-08
Added support for MATCH syntax and unified column option ForeignKey

LucaCappelletti94:
This PR adds support for the [SQL standard `MATCH [FULL | PARTIAL | SIMPLE]` syntax](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK) for foreign key constraints.

### Changes

- Extended keywords to include `PARTIAL` and `SIMPLE`
- New `MatchKind` enum with `Full`, `Partial`, and `Simple` variants
- Parser support for `MATCH` syntax in both column-level and table-level foreign key constraints
- Refactored `ColumnOption::ForeignKey` to use `ForeignKeyConstraint` struct, eliminating code duplication
- Added opportune tests
- Added opportune documentation for newly supported syntax

### Examples
```sql
-- Column-level constraint
CREATE TABLE orders (
    id INT REFERENCES customers(id) MATCH FULL ON DELETE CASCADE
);

-- Table-level constraint  
CREATE TABLE orders (
    customer_id INT,
    FOREIGN KEY (customer_id) REFERENCES customers(id) MATCH SIMPLE
);
```

Closes issue #2061 and works towards closing issue #2059


src/ast/mod.rs
```diff
@@ -655,6 +655,40 @@ pub enum CastKind {
...
+///
+/// Specifies how to match composite foreign keys against the referenced table.
+/// A value inserted into the referencing column(s) is matched against the values
+/// of the referenced table and referenced columns using the given match type.
+///
+/// See: <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES>
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum MatchKind {
```
```suggestion
pub enum ConstraintReferenceMatchKind {
```
Thinking `MATCH` might be too generic and likely to conflict naming wise in other contexts as a result

src/ast/mod.rs
```diff
@@ -655,6 +655,40 @@ pub enum CastKind {
     DoubleColon,
 }
 
+/// The MATCH option for foreign key constraints.
+///
+/// Specifies how to match composite foreign keys against the referenced table.
+/// A value inserted into the referencing column(s) is matched against the values
+/// of the referenced table and referenced columns using the given match type.
```
```suggestion
/// `MATCH` type for constraint references
```
We can avoid defining semantics and leave the description generic so that it can be reused correctly in other context (like table constraint references)

src/ast/mod.rs
```diff
@@ -655,6 +655,40 @@ pub enum CastKind {
...
+    /// `MATCH FULL` - Will not allow one column of a multicolumn foreign key to be null
+    /// unless all foreign key columns are null; if they are all null, the row is not
+    /// required to have a match in the referenced table.
+    Full,
+    /// `MATCH PARTIAL` - Not yet implemented by most databases (part of SQL standard).
+    /// Would allow partial matches in multicolumn foreign keys.
+    Partial,
+    /// `MATCH SIMPLE` - The default behavior. Allows any of the foreign key columns
+    /// to be null; if any of them are null, the row is not required to have a match
+    /// in the referenced table.
```
```suggestion
    /// `MATCH FULL`
    Full,
    /// `MATCH PARTIAL`
    Partial,
    /// `MATCH SIMPLE`
```

src/parser/mod.rs
```diff
@@ -7929,22 +7929,22 @@ impl<'a> Parser<'a> {
         loop {
             if self.parse_keyword(Keyword::CONSTRAINT) {
                 let name = Some(self.parse_identifier()?);
-                if let Some(option) = self.parse_optional_column_option()? {
+                if let Some(option) = self.parse_optional_column_option(&col_name)? {
```
iffyio: not sure what this change implies, it looks like we'd be storing the column name twice in the AST node which doesn't seem ideal?

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
         _ => unreachable!(),
     }
 }
+
+#[test]
+/// Test to verify whether `MATCH FULL` syntax for foreign keys is parsed correctly.
```
```suggestion
```
The comment read superfluous since the fact should be clear from the test name

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+                }
+                _ => panic!("Expected TableConstraint::ForeignKey"),
+            }
+        }
+        _ => unreachable!("{:?} should parse to Statement::CreateTable", sql),
+    }
+}
+
+#[test]
+/// Test to verify whether `MATCH SIMPLE` syntax for foreign keys is parsed correctly.
```
```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+                }
+                _ => panic!("Expected TableConstraint::ForeignKey"),
+            }
+        }
+        _ => unreachable!("{:?} should parse to Statement::CreateTable", sql),
+    }
+}
+
+#[test]
+/// Test to verify whether `MATCH PARTIAL` syntax for foreign keys is parsed correctly.
```
```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+fn parse_foreign_key_match_partial() {
+    let sql = "CREATE TABLE orders (order_id INT PRIMARY KEY REFERENCES another_table (id) MATCH PARTIAL, customer_id INT, FOREIGN KEY (customer_id) REFERENCES customers(customer_id) MATCH PARTIAL)";
+    let statement = pg_and_generic().verified_stmt(sql);
+    match statement {
+        Statement::CreateTable(CreateTable {
+            columns,
+            constraints,
+            ..
+        }) => {
+            // Check column-level foreign key with MATCH PARTIAL
```
```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+            assert_eq!(columns[0].name.value, "order_id");
+            match &columns[0].options[1].option {
+                ColumnOption::ForeignKey(constraint) => {
+                    assert_eq!(constraint.foreign_table.to_string(), "another_table");
+                    assert_eq!(constraint.match_kind, Some(MatchKind::Partial));
+                }
+                _ => panic!("Expected ColumnOption::ForeignKey"),
+            }
+
+            // Check table-level foreign key constraint with MATCH PARTIAL
```
```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+                }
+                _ => panic!("Expected TableConstraint::ForeignKey"),
+            }
+        }
+        _ => unreachable!("{:?} should parse to Statement::CreateTable", sql),
+    }
+}
+
+#[test]
+/// Test to verify foreign key MATCH syntax combined with ON DELETE/ON UPDATE actions
```
```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
...
+                    assert_eq!(constraint.foreign_table.to_string(), "customers");
+                    assert_eq!(constraint.match_kind, Some(MatchKind::Simple));
+                    assert_eq!(constraint.on_delete, Some(ReferentialAction::SetNull));
+                    assert_eq!(constraint.on_update, Some(ReferentialAction::Cascade));
+                }
+                _ => panic!("Expected TableConstraint::ForeignKey"),
+            }
+        }
+        _ => unreachable!("{:?} should parse to Statement::CreateTable", sql),
+    }
```
```suggestion
    pg_and_generic().verified_stmt(sql);
```
I think in scenarios like these where we want to verify more complicated combinations (or different variant of the same component) and each component has already been asserted previously we can rely solely on the  verified_stmt assertions to keep the tests leaner. Similarly, given that we've asserted the returned AST on say `MATCH FULL` test scenario, we can rely on verified_stmt only for the othe variants like `MATCH SIMPLE/PARTIAL`. Leads to significantly less test code to maintain going forward

tests/sqlparser_postgres.rs
```diff
@@ -6646,3 +6647,144 @@ fn parse_alter_schema() {
         _ => unreachable!(),
     }
 }
+
+#[test]
```
iffyio, line 6608: Can we merge the tests into a single function given they're testing the same feature?

src/parser/mod.rs
LucaCappelletti94: It is needed to standardize `ForeignKeyConstraint` and avoid having a duplicated struct to represent the same type of concept. I could possibly leave the `columns` vector of the `ForeignKeyConstraint` empty, but when I use that struct in code that works on top of the AST it is quite useful to have the column ident defined there, and not have to handle the case of a special `ForeignKeyConstraint`.

tests/sqlparser_postgres.rs
LucaCappelletti94: Merged and simplified.

src/parser/mod.rs
LucaCappelletti94: You will note that it is the same change needed for PR https://github.com/apache/datafusion-sqlparser-rs/pull/2064, which I did separately but the goal is identical - to standardize the structs used to represent constraints in the columns and table.

iffyio: I see, the goal of the AST is to represent the syntax as closely as possible so that I don't think that storing the column name in another node in the tree than where it was found seems reasonable. haven't looked closely at the different foreign key constraint variants but I would assume that if both variants show up in different parts of the syntax then its not unreasonable that they are represented differently

LucaCappelletti94: I agree with that, but this philosophy of using the most strictly compatible AST nodes in the different parts of the tree has already been broken several times to reduce code duplication.

Here I am quite unsure whether the most appropriate solution would be to:

1) use the table constraint struct and duplicate the column ident
2) use the table constraint struct and leave its columns attribute empty
3) use a new struct to strictly represent the column option for this constraint variant (for the same reasons of having structs instead of struct variants)

I would personally be inclined with 1 or 2, but I understand the duplication concerns of 1. Btw, is there any particular reason for the current clone-based approach for idents instead of a copy and lifetime based one?

Lmk your opinion regarding how do you feel best proceeding.

LucaCappelletti94: I went for solution number (2), let me know whether you agree with it.

iffyio (approved):

PR #2068, opened 2025-10-15
chore: add stack overflow warning for Visitor and VisitorMut

niebayes:
In our scenario, we set `Break` a custom `Error` type, with its size being around 100 bytes. This caused our stack space to be rapidly consumed during visitor recursion, and for more complex SQL queries, it even led to stack overflow.
We solved this problem by boxing the Error.

### DIFF
--- a/src/ast/visitor.rs
+++ b/src/ast/visitor.rs
@@ -124,10 +124,6 @@ visit_noop!(bigdecimal::BigDecimal);
 /// node and `post_visit_` methods are invoked after visiting all
 /// children of the node.
 ///
-/// Important note: The `Break` type should be kept as small as possible to prevent
-/// stack overflow during recursion. If you need to return an error, consider
-/// boxing it with `Box` to minimize stack usage.
-///
 /// # See also
 ///
 /// These methods provide a more concise way of visiting nodes of a certain type:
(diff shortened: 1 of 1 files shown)


src/ast/visitor.rs
```diff
@@ -124,6 +124,10 @@ visit_noop!(bigdecimal::BigDecimal);
 /// node and `post_visit_` methods are invoked after visiting all
 /// children of the node.
 ///
+/// Important note: The `Break` type should be kept as small as possible to prevent
```
iffyio: could we place the note on the documentation of `Break` instead? [here for example](https://github.com/niebayes/datafusion-sqlparser-rs/blob/6419cbd073d6fda4f8b43e7b320d9642ae1bba66/src/ast/visitor.rs#L300)

iffyio: could we place the note on the documentation of `Break` instead? [here for example](https://github.com/niebayes/datafusion-sqlparser-rs/blob/6419cbd073d6fda4f8b43e7b320d9642ae1bba66/src/ast/visitor.rs#L300)

niebayes: Sure, moved the note.

niebayes: moved

Issue #2065, opened 2025-10-09
Limited Inline Comment Support in AST


RPG-Alex:
### Context

Currently, `datafusion-sqlparser-rs` strips out all inline comments (`-- …` and `/* … */`). There is existing support for `COMMENT ON …` statements, and it would be good to have support for inline comments as well, given they often used  for documenting sql. However, these comments are currently lost while parsing.

There is an open issue on [lossless AST](https://github.com/apache/datafusion-sqlparser-rs/issues/175), but it has been open for about 5 years. Moreover this topic is referenced in this [PR](https://github.com/apache/datafusion-sqlparser-rs/pull/189). So there is a precedent to move forward with inline comment support. 

### Objective/Motivation

Capturing inline comments in the AST would be helpful for: 
1) **Documentation Generation**: This is the primary purpose of this issue, to extend the existing ability to use `COMMENT ON` support, for downstream documentation generation based on the parsed AST. 
2) **Linting Coverage**: Going along with 1, being able to track inline comments would be useful for checking for completeness of documentation, and especially helpful in places where documentation in the form of comments is necessary as inline comments and not `COMMENT ON` comments.

##### Inline Comment Associated with a table:
```sql
-- Table for storing values for users
CREATE TABLE IF NOT EXISTS users (
  id BIGINT PRIMARY KEY
  -- Stores the username as text, required value for creating a user account.
  name TEXT NOT NULL
);
```
Expectation: The AST node for the `users` table would include a `comment: Option<String>` that would then be populated with the comment located on the previous line: `Table for storing values for users`.  Additionally the column `name` would have an associated `comment: Option<String>` that would contain `Stores the username as text, required value for creating a user account.`. Therefore all tables and columns would carry a `comment: Option<String>` derived from the trailing `--` based comment preceding the field. 

**NOTE:** For now the assumption is parsing the comments immediately preceding the line the field is on, though comments can be arranged as such as well:
```sql
CREATE TABLE IF NOT EXISTS users ( -- Table for storing values for users
  id BIGINT PRIMARY KEY
  name TEXT NOT NULL  -- Stores the username as text, required value for creating a user account.
);
```
It might prove fruitful to allow parsing both types of comments, and perhaps specifying when parsing which are relevant and which are not. 

##### Inline Comments out of Scope
```sql
-- migration script v2025-01

CREATE TABLE IF NOT EXISTS users (
  id BIGINT PRIMARY KEY
[...]

Issue #175, opened 2020-05-29
Comment handling


maxcountryman:
I've been working on [a SQL formatter](https://github.com/maxcountryman/forma) that leverages this library extensively. One problem with using `sqlparser-rs` right now is that comments are treated like whitespace and removed from the AST. I'm curious if there would be any interest in bringing comments into the AST? Perhaps this could be optional, since I imagine it's not terribly useful for most cases.

nickolay:
I mentioned [before](https://github.com/andygrove/sqlparser-rs/issues/161#issuecomment-617230324) that I am interested in experimenting with an alternative design, in which the parser's output can be concatenated back together to form exactly the input string. I should probably share the prototype I have, despite it being very incomplete as I seldom work on it...

I don't see another way for the comments to be retained.

maxcountryman:
@nickolay I'd be very curious to see your prototype--if there's any way I can contribute, I'd be happy to. 🙂

mullr:
This is /exactly/ the use case I had in mind for #161.

nickolay:
Sorry for the delay, I posted the prototype in #189

olalonde:
I also would like to access comments in the AST :( Any progress on this?

LucaCappelletti94:
Reasonably completed by #2107 and #2354

PR #2069, opened 2025-10-16
Leading comment support added for `AlterTable`, `CreateTable`, and `ColumnDef` 

RPG-Alex:
This PR refactors the internal representation of comments in the tokenizer and parser and adds leading comment support for `AlterTable`, `CreateTable`, and `ColumnDef`.

The `SingleLine` and `MultiLine` comment variants previously defined in the `Whitespace` enum are now represented by a new, dedicated `Comment` enum. The `Comment` is used for both whitespace comments (interstitial comments) and leading comments. Leading comment support propagated for `AlterTable`, `CreateTable`, and `ColumnDef`, to parse leading comments alongside the associatied struct (see below for example). 

### Rationale

This change improves the semantic clarity of comment handling in the SQL tokenizer and parser.
As discussed in **#2065**, comments preceding a table or column definition may serve as **inline documentation**, and should be distinguishable from interstitial (whitespace) comments.
For example:

```sql
-- Leading comment for table users
CREATE TABLE IF NOT EXISTS users (
  id BIGINT PRIMARY KEY,
  -- Leading comment for table field name
  name TEXT NOT NULL -- interstitial comment ignored by parser
);
```
#### Term Definitions
**Intersitial Comment**: a comment preceded by something (if nothing then defined as leading comment) that is not a comma or semicolon:
```sql
CREATE -- comment trailing the create keyword 
TABLE -- comment trailing the table keyword 
my_table
```

**Leading Comment**: a comment that is preceded by either nothing, a comma, or a semicolon. 
currently the variants covered include single line comments:
```sql
-- a comment preceding the create table statement 
CREATE TABLE IF NOT EXISTS users (...
```
and multi-line comments:
```sql
/* a multi-line comment 
Preceding this table */
CREATE TABLE IF NOT EXISTS users (...
```

By separating comment handling from generic whitespace, the parser can now support more context-aware interpretations and contribute to a **lossless syntax tree**, addressing **#175** and complementing **PR #189**.

### Summary of Changes

* **Added:** `Comment` enum encapsulating `SingleLine` and `MultiLine` comment variants.
* **Refactored:**
  * `Whitespace` to include `InterstitialComment(Comment)` variant.
  * `Tokenizer.rs` to emit `Comment` values instead of `Whitespace::SingleLineComment` / `Whitespace::MultiLineComment`.
  * `parser/mod.rs` implemented `LeadingComment` for `CreateTable`, `ColumnDef`, `AlterTable` parsing.
  * added `leading_comment: Option<Comment>` to the  `CreateTable`, `ColumnDef`, `AlterTable`:
```rust
pub struct ColumnDef {
    pub name: Ident,
    pub data_type: DataType,
    pub options: Vec<ColumnOptionDef>,
    /// Leading comment for the column.
    pub leading_comment: Option<Comment>,
}

```
* **Added logic** for differentiating `InterstitialComment` and `LeadingComment`.
* **Propagated refactors** to all dependent parser components.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -1209,7 +1209,7 @@ pub struct ColumnDef {
 impl fmt::Display for ColumnDef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if let Some(leading_comment) = &self.leading_comment {
-            write!(f,"{leading_comment}")?;
+            write!(f, "{leading_comment}")?;
         }
         if self.data_type == DataType::Unspecified {
             write!(f, "{}", self.name)?;
@@ -2305,7 +2305,7 @@ impl fmt::Display for CreateTable {
         // Columns provided for CREATE TABLE AS:
         //   `CREATE TABLE t (a INT) AS SELECT a from t2`
         if let Some(leading_comment) = &self.leading_comment {
-            write!(f,"{leading_comment}")?;
+            write!(f, "{leading_comment}")?;
         }
         write!(
             f,
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -553,7 +553,7 @@ impl Spanned for CreateTable {
             refresh_mode: _,
             initialize: _,
             require_user: _,
-            leading_comment: _,                 // Option<Comment>
+            leading_comment: _, // Option<Comment>
         } = self;
 
         union_spans(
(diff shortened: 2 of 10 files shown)


Viicos (commented):
Hi @RPG-Alex,

Coming from https://github.com/apache/datafusion-sqlparser-rs/pull/2077#discussion_r2515603106, I'm wondering if it is necessary to distinguish between the two types of comments. In theory, by analyzing the token coming right after the comment one?

Maybe this makes sense as a user convenience, but it feels weird to have two separate tokens for this, I'm wondering if we could have a concept of flags, where this information would be stored.

That way, users that don't care about which type of comment it is can simply match against a single `Comment` token kind. But this would require a larger refactor, and is not only relevant for comments (e.g. for strings, if you don't care about the quote style and number).

PR #2071, opened 2025-10-18
SQLite: make period optional for CREATE TRIGGER

takluyver:
I just tried out the new CREATE TRIGGER support for SQLite, but ran across a limitation. SQLite's [CREATE TRIGGER](https://sqlite.org/lang_createtrigger.html) statement makes the period (`BEFORE`, `AFTER`, `INSTEAD OF`) optional. Leaving it out is equivalent to BEFORE. This allows that in sqlparser, making the field an `Option<TriggerPeriod>`.

This may well be clumsily implemented - I don't write all that much Rust, so suggestions are welcome. I've enabled this only for SQLite for now - I haven't researched if the same syntax is legal in any other dialects.

cc @LucaCappelletti94 - and thanks for adding the overall CREATE TRIGGER support :slightly_smiling_face:

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5592,20 +5592,7 @@ impl<'a> Parser<'a> {
         }
 
         let name = self.parse_object_name(false)?;
-        let period = if dialect_of!(self is SQLiteDialect)
-            && self
-                .peek_one_of_keywords(&[
-                    Keyword::INSERT,
-                    Keyword::UPDATE,
-                    Keyword::DELETE,
-                    Keyword::TRUNCATE,
-                ])
-                .is_some()
-        {
-            None // SQLite: period can be skipped (equivalent to BEFORE)
-        } else {
-            Some(self.parse_trigger_period()?)
-        };
+        let period = self.maybe_parse(|parser| parser.parse_trigger_period())?;
 
         let events = self.parse_keyword_separated(Keyword::OR, Parser::parse_trigger_event)?;
         self.expect_keyword_is(Keyword::ON)?;
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -5830,7 +5830,7 @@ fn parse_create_trigger_invalid_cases() {
         ),
         (
             "CREATE TRIGGER check_update TOMORROW UPDATE ON accounts EXECUTE FUNCTION check_account_update",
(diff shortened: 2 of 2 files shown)


src/ast/ddl.rs
```diff
@@ -3029,7 +3029,9 @@ pub struct CreateTrigger {
     /// FOR EACH ROW
     /// EXECUTE FUNCTION trigger_function();
     /// ```
-    pub period: TriggerPeriod,
+    ///
+    /// This may be omitted in SQLite, the effect is equivalent to BEFORE.
```
```suggestion
```
we can skip the description since the optionality is clear from the representation (and the conditions can change over time)

src/parser/mod.rs
```diff
@@ -5592,7 +5592,20 @@ impl<'a> Parser<'a> {
...
+                    Keyword::UPDATE,
+                    Keyword::DELETE,
+                    Keyword::TRUNCATE,
+                ])
+                .is_some()
+        {
+            None // SQLite: period can be skipped (equivalent to BEFORE)
+        } else {
+            Some(self.parse_trigger_period()?)
+        };
```
iffyio: I think we can do something like this to support optionality
```
let period = self.maybe_parse(|parser| parser.parse_trigger_period());
```
It would imply the same behavior for other dialects than sqlite but that's okay for the parser to be more permissive

tests/sqlparser_sqlite.rs
```diff
@@ -863,6 +863,50 @@ fn test_create_trigger() {
...
+            assert_eq!(
+                trigger_object,
+                Some(TriggerObjectKind::ForEach(TriggerObject::Row))
+            );
+            assert!(condition.is_none());
+            assert!(!statements_as);
+            assert!(characteristics.is_none());
+        }
+        _ => unreachable!("Expected CREATE TRIGGER statement"),
+    }
```
iffyio:
```suggestion
    sqlite().verified_stmt(statement7);
```
To cut down on the lines of test code I think the following would suffice, since the other test scenarios already inspect the returned AST

src/parser/mod.rs
takluyver: Thanks!

Allowing this for all  dialects changes an error message that is tested for the Postgres dialect. For now I'll change that test so that everything passes, but if you'd prefer a different direction, like making it conditional on the dialect again, I can easily cut that commit off.

iffyio (approved):
LGTM! Thanks @takluyver!

takluyver:
Thank you!

PR #2072, opened 2025-10-20
Redshift: more copy options

yoavcloud:
Added support for REMOVEQUOTES, COMPUPDATE and STATUPDATE.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -8287,18 +8287,18 @@ impl fmt::Display for CopyLegacyOption {
             CompUpdate { preset, enabled } => {
                 write!(f, "COMPUPDATE")?;
                 if *preset {
-                    write!(f, " PRESET")
-                } else {
+                    write!(f, " PRESET")?;
+                } else if let Some(enabled) = enabled {
                     write!(
                         f,
                         "{}",
-                        match *enabled {
-                            Some(true) => " TRUE",
-                            Some(false) => " FALSE",
-                            _ => "",
+                        match enabled {
+                            true => " TRUE",
+                            false => " FALSE",
                         }
-                    )
+                    )?;
                 }
+                Ok(())
             }
             Csv(opts) => {
                 write!(f, "CSV")?;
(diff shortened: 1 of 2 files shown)


LucaCappelletti94:
Do consider adding some tests for the newly added code.

yoavcloud:
Have you looked at tests/sqlparser_common.rs? The changes there represent all cases for the newly added options.

src/ast/mod.rs
```diff
@@ -8278,6 +8710,22 @@ impl fmt::Display for CopyLegacyOption {
...
+                } else {
+                    write!(
+                        f,
+                        "{}",
+                        match *enabled {
+                            Some(true) => " TRUE",
+                            Some(false) => " FALSE",
+                            _ => "",
+                        }
+                    )
```
iffyio: this looks like it can be simplified as?
```rust
if let Some(enabled) = enabled {
    write!()
}
```

src/parser/mod.rs
```diff
@@ -9766,6 +9676,21 @@ impl<'a> Parser<'a> {
...
             Some(Keyword::CLEANPATH) => CopyLegacyOption::CleanPath,
+            Some(Keyword::COMPUPDATE) => {
+                let preset = self.parse_keyword(Keyword::PRESET);
+                let enabled = match self.parse_one_of_keywords(&[
+                    Keyword::TRUE,
+                    Keyword::FALSE,
+                    Keyword::ON,
+                    Keyword::OFF,
+                ]) {
+                    Some(Keyword::PRESET) => None,
```
iffyio: is this branch valid, we don't seem to expect it in the keywords?

LucaCappelletti94:
As far as I can understand, the changes made in `tests/sqlparser_common.rs` only check that the parse happens without failure, but not that it parses to the expected structs.

yoavcloud:
> As far as I can understand, the changes made in `tests/sqlparser_common.rs` only check that the parse happens without failure, but not that it parses to the expected structs.

It tests the full SQL to AST and AST to SQL flow. The new options are added to an existing flow, not introducing new AST structs. The existing structs are covered in existing tests.

Issue #2036, opened 2025-09-17
Improve performance by reducing string copies


alamb:
@Omega359 [mentioned in discord](https://discord.com/channels/885562378132000778/1344430130252218439/1415763431516278794) a blog post by the Databend folks about how they wrote a replacement of sqlparser-rs for their system

https://www.databend.com/blog/category-engineering/2025-09-10-query-parser/

<img width="464" height="338" alt="Image" src="https://github.com/user-attachments/assets/3e3bfc61-bb1d-4ef9-b214-6bde36cce649" />

One thing they cite is the fact that sqlparser-rs often copies strings, even when this could be avoided

If anyone cares, we can probably avoid copying strings, someone just has to care enough to thread the lifetime information through all the tokens (or maybe use an Arc<str> w/ an offset)

@AdamGS mentioned he might be interested in trying this

This could possibly be related to another ticket here:
- https://github.com/apache/datafusion-sqlparser-rs/issues/1558

alamb:
FYI for your amusement @iffyio and @yoavcloud

AdamGS:
I'm going to give it a try, mostly to satisfy my own curiosity. My intuition is more the `Arc` path because lifetimes are painful, but its too early to tell if I'm on the right path there, I might try both.

If it works, it might be a significant breaking change externally, but I guess we'll worry about that if we get there.

eyalsatori:
Some thoughts on this topic:  


I've been thinking about the Arc suggestion from the previous discussion, and I see some potential issues:
The tokenizer currently holds only a reference to the original query. If we use Arc, the library would need to create an owned copy of the query, which seems to miss the point of zero-copy tokenization. Additionally, since each copy would need to handle a different slice of the string, we'd need to implement something similar to the [Bytes](https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L88) library.
Alternative approach using lifetimes


Another option is to introduce lifetimes directly into the Token enum:

```
pub enum Token<'a> {
    /// An end-of-file marker, not a real token
    EOF,
    /// A keyword (like SELECT) or an optionally quoted SQL identifier
    Word(Word<'a>),
    /// An unsigned numeric literal
    Number(&'a str, bool),
    // ...
}
```


And the tokenize method would become:
`pub fn tokenize(&'a mut self) -> Result<Vec<Token<'a>>, TokenizerError>` 


[...]

PR #2073, opened 2025-10-23
Prepare tokenizer for using borrowed strings instead of allocations.

eyalleshem:
Key points for this commit:
- The peekable trait isn't sufficient for using string slices, as we need the byte indexes (start/end) to create string slices, so added the current byte position to the State struct (Note: in the long term we could potentially remove peekable and use only the current position as an iterator)
- Created internal functions that create slices from the original query instead of allocating strings, then converted these functions to return String to maintain compatibility (the idea is to make a small, reviewable commit without changing the Token struct or the parser)

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -743,8 +743,12 @@ impl std::error::Error for TokenizerError {}
 
 struct State<'a> {
     peekable: Peekable<Chars<'a>>,
-    pub line: u64,
-    pub col: u64,
+    /// Reference to the original source string being tokenized
+    source: &'a str,
+    line: u64,
+    col: u64,
+    /// Byte position in the source string
+    byte_pos: usize,
 }
 
 impl State<'_> {
@@ -759,6 +763,8 @@ impl State<'_> {
                 } else {
                     self.col += 1;
                 }
+                // Update byte position (characters can be multi-byte in UTF-8)
+                self.byte_pos += s.len_utf8();
                 Some(s)
             }
         }
@@ -767,6 +773,16 @@ impl State<'_> {
     /// return the next character but do not advance the stream
     pub fn peek(&mut self) -> Option<&char> {
         self.peekable.peek()
+    }
+
+    /// Return the character `n` positions ahead without advancing the stream.
+    /// For example, `peek_nth(0)` returns the current character (same as peek),
+    /// and `peek_nth(1)` returns the next character.
+    pub fn peek_nth(&self, n: usize) -> Option<char> {
+        if self.byte_pos >= self.source.len() {
+            return None;
+        }
+        self.source[self.byte_pos..].chars().nth(n)
     }
 
     pub fn location(&self) -> Location {
@@ -893,8 +909,10 @@ impl<'a> Tokenizer<'a> {
     ) -> Result<(), TokenizerError> {
         let mut state = State {
             peekable: self.query.chars().peekable(),
+            source: self.query,
             line: 1,
             col: 1,
+            byte_pos: 0,
         };
 
         let mut location = state.location();
@@ -908,22 +926,24 @@ impl<'a> Tokenizer<'a> {
         Ok(())
     }
 
-    // Tokenize the identifier or keywords in `ch`
+    /// Tokenize an identifier or keyword after consuming the first character(s).
+    /// `consumed_byte_len` is the total byte length of the character(s) already consumed.
     fn tokenize_identifier_or_keyword(
         &self,
-        ch: impl IntoIterator<Item = char>,
-        chars: &mut State,
+        consumed_byte_len: usize,
+        chars: &mut State<'a>,
     ) -> Result<Option<Token>, TokenizerError> {
         chars.next(); // consume the first char
-        let ch: String = ch.into_iter().collect();
-        let word = self.tokenize_word(ch, chars);
+        let word = self.tokenize_word(consumed_byte_len, chars);
 
         // TODO: implement parsing of exponent here
         if word.chars().all(|x| x.is_ascii_digit() || x == '.') {
             let mut inner_state = State {
                 peekable: word.chars().peekable(),
+                source: &word,
                 line: 0,
                 col: 0,
+                byte_pos: 0,
             };
             let mut s = peeking_take_while(&mut inner_state, |ch| matches!(ch, '0'..='9' | '.'));
             let s2 = peeking_take_while(chars, |ch| matches!(ch, '0'..='9' | '.'));
@@ -937,7 +957,7 @@ impl<'a> Tokenizer<'a> {
     /// Get the next token or return None
     fn next_token(
         &self,
-        chars: &mut State,
+        chars: &mut State<'a>,
         prev_token: Option<&Token>,
     ) -> Result<Option<Token>, TokenizerError> {
         match chars.peek() {
@@ -988,7 +1008,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "b" or "B"
-                            let s = self.tokenize_word(b, chars);
+                            let s = self.tokenize_word(b.len_utf8(), chars);
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1015,7 +1035,7 @@ impl<'a> Tokenizer<'a> {
                             ),
                         _ => {
                             // regular identifier starting with an "r" or "R"
-                            let s = self.tokenize_word(b, chars);
+                            let s = self.tokenize_word(b.len_utf8(), chars);
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1034,7 +1054,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "N"
-                            let s = self.tokenize_word(n, chars);
+                            let s = self.tokenize_word(n.len_utf8(), chars);
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1051,7 +1071,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "E" or "e"
-                            let s = self.tokenize_word(x, chars);
+                            let s = self.tokenize_word(x.len_utf8(), chars);
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1070,7 +1090,7 @@ impl<'a> Tokenizer<'a> {
                         }
                     }
                     // regular identifier starting with an "U" or "u"
-                    let s = self.tokenize_word(x, chars);
+                    let s = self.tokenize_word(x.len_utf8(), chars);
                     Ok(Some(Token::make_word(&s, None)))
                 }
                 // The spec only allows an uppercase 'X' to introduce a hex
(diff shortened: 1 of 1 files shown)


eyalleshem:
This commit is a first step toward addressing issue #2036.

It doesn't contain any behavioral changes yet, and the scope is intentionally limited because it touches some core elements of the tokenizer. I wanted to:

1. Keep the changes small and easy to review
2. Get early feedback before proceeding with the full implementation

I'd appreciate any thoughts on this approach before I continue!

src/tokenizer.rs
```diff
@@ -769,6 +775,12 @@ impl State<'_> {
         self.peekable.peek()
     }
 
+    /// return the character after the next character (lookahead by 2) without advancing the stream
+    pub fn peek_next(&self) -> Option<char> {
```
```suggestion
    /// Return the `nth` next character without advancing the stream.
    pub fn peek_nth(&self, n: usize) -> Option<char> {
```
Thinking we can have this more generic so that it can be reused in other context? Similar to the `peek_nth*` parser methods

src/tokenizer.rs
```diff
@@ -769,6 +775,12 @@ impl State<'_> {
...
+        // Use the source and byte_pos instead of cloning the peekable iterator
```
iffyio: Can we add a guard here that the index is safe? So that we don't panic if we hit a bug or the API is being misused. e.g.
```rust
if self.byte_pos >= self.source.len() {
    return None
}
```

src/tokenizer.rs
```diff
@@ -912,18 +926,21 @@ impl<'a> Tokenizer<'a> {
     fn tokenize_identifier_or_keyword(
         &self,
         ch: impl IntoIterator<Item = char>,
-        chars: &mut State,
+        chars: &mut State<'a>,
     ) -> Result<Option<Token>, TokenizerError> {
         chars.next(); // consume the first char
-        let ch: String = ch.into_iter().collect();
-        let word = self.tokenize_word(ch, chars);
+                      // Calculate total byte length without allocating a String
```
iffyio: The indentation looks a bit off on this line?

src/tokenizer.rs
```diff
@@ -912,18 +926,21 @@ impl<'a> Tokenizer<'a> {
...
+        let consumed_byte_len: usize = ch.into_iter().map(|c| c.len_utf8()).sum();
```
iffyio: I wonder if we can instead replace the `ch` parameter to this `tokenize_identifier_or_keyword` function with the actual `consumed_byte_len: usize`?

If I'm understanding the intent/requirement of this change correctly the caller of this function should have that value on hand given that we seem to require that `ch` contains the preceeding characters in the stream?

The current flow was initially unclear to me that the caller passes in an iterator of items, whose contents we did not use here, and it required digging further into `tokenize_word` to realised why that was the case.

src/tokenizer.rs
```diff
@@ -1876,13 +1893,26 @@ impl<'a> Tokenizer<'a> {
...
-        let mut s = first_chars.into();
-        s.push_str(&peeking_take_while(chars, |ch| {
-            self.dialect.is_identifier_part(ch)
-        }));
-        s
+    /// Tokenize an identifier or keyword, after the first char(s) have already been consumed.
+    /// `consumed_byte_len` is the byte length of the consumed character(s).
+    fn tokenize_word(&self, consumed_byte_len: usize, chars: &mut State<'a>) -> String {
+        // Calculate where the first character started
+        let first_char_byte_pos = chars.byte_pos - consumed_byte_len;
```
iffyio, line 1914: Can we add a check that the operation doesn't overflow? e.g.
```rust
if consumed_byte_len >= chars.byte_pos {
    return "".to_string()
}
```

src/tokenizer.rs
```diff
@@ -1876,13 +1893,26 @@ impl<'a> Tokenizer<'a> {
...
+
+    /// Tokenize an identifier or keyword, returning a borrowed slice when possible.
+    /// The first character position must be provided (before it was consumed).
+    /// Returns a slice with the same lifetime as the State's source.
+    fn tokenize_word_borrowed(&self, first_char_byte_pos: usize, chars: &mut State<'a>) -> &'a str {
+        // Consume the rest of the word
+        borrow_slice_until(chars, |ch| self.dialect.is_identifier_part(ch));
+
+        // Return a slice from the first char to the current position
+        &chars.source[first_char_byte_pos..chars.byte_pos]
```
iffyio, line 1934: Similarly here we can check that the indexing is safe and maybe return a literal `""` if not?

src/tokenizer.rs
```diff
@@ -2176,35 +2206,82 @@ impl<'a> Tokenizer<'a> {
...
+
+/// Borrow a slice from the original string until `predicate` returns `false` or EOF is hit.
+///
+/// # Arguments
+/// * `chars` - The character iterator state (contains reference to original source)
+/// * `predicate` - Function that returns true while we should continue taking characters
+///
+/// # Returns
+/// A borrowed slice of the source string containing the matched characters
+fn borrow_slice_until<'a>(
```
```suggestion
fn peeking_take_while_ref<'a>(
```
Thinking so that its clear that the function is related to `peeking_take_while`

src/tokenizer.rs
```diff
@@ -2176,35 +2206,82 @@ impl<'a> Tokenizer<'a> {
...
+}
+
+/// Borrow a slice from the original string until `predicate` returns `false` or EOF is hit.
+///
+/// # Arguments
+/// * `chars` - The character iterator state (contains reference to original source)
+/// * `predicate` - Function that returns true while we should continue taking characters
+///
+/// # Returns
+/// A borrowed slice of the source string containing the matched characters
```
iffyio: Doc wise I think its easier to only reference the `peeking_take_while` method instead, and we mention only the difference to this function. That way we only describe the functionality once. Also, the project doesn't use the `# Arguments, # Returns etc` documentation format so that we can skip that for consistency I think.

src/tokenizer.rs
```diff
@@ -2176,35 +2206,82 @@ impl<'a> Tokenizer<'a> {
...
+fn borrow_slice_until<'a>(
+    chars: &mut State<'a>,
+    mut predicate: impl FnMut(char) -> bool,
+) -> &'a str {
+    // Record the starting byte position
```
iffyio, line 2239: We can sanity check the index before  using it here as well?

src/tokenizer.rs
```diff
@@ -743,8 +743,12 @@ impl std::error::Error for TokenizerError {}
 
 struct State<'a> {
     peekable: Peekable<Chars<'a>>,
+    /// Reference to the original source string being tokenized
+    source: &'a str,
     pub line: u64,
     pub col: u64,
+    /// Byte position in the source string
+    pub byte_pos: usize,
```
```suggestion
    byte_pos: usize,
```

src/tokenizer.rs
```diff
@@ -2176,35 +2206,82 @@ impl<'a> Tokenizer<'a> {
...
+/// This is a zero-copy version of `peeking_next_take_while`.
+///
+/// # Arguments
+/// * `chars` - The character iterator state (contains reference to original source)
+/// * `predicate` - Function that returns true while we should continue taking characters.
+///   Takes current char and optional next char for lookahead.
+///
+/// # Returns
+/// A borrowed slice of the source string containing the matched characters
+fn borrow_slice_until_next<'a>(
```
iffyio: wondering, given this function is quite similar, would it make sense to implement `borrowed_slice_until` as following instead? To reuse the same definition/impl

```rust
fn borrow_slice_until(chars) {
    borrow_slice_until_next(chars, |ch, _|{
        predicate(ch)
    })
}
```

src/tokenizer.rs
```diff
@@ -2176,35 +2206,82 @@ impl<'a> Tokenizer<'a> {
...
+    let end_pos = chars.byte_pos;
+
+    // Return the slice from the original source
+    &chars.source[start_pos..end_pos]
 }
 
-/// Same as peeking_take_while, but also passes the next character to the predicate.
-fn peeking_next_take_while(
-    chars: &mut State,
+/// Borrow a slice from the original string until `predicate` returns `false` or EOF is hit.
```
iffyio, line 2263: just flagging similar comments for `borrow_while_until` apply to this function

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: code comment deleted (as result of the next suggestion)

src/tokenizer.rs
eyalleshem: agree , move it out . 
The downside is that callers now need to calculate the UTF-8 byte length of their characters.

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: s a sanity check needed here? The start_pos and end_pos is taken from the iterator, and the iterator is incremented according to the characters in the buffer

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: Maybe, but I'm not sure what happens if EOF is reached on the next character. I don't think I want to include that as part of this commit.

src/tokenizer.rs
eyalleshem: I think its ok now, let me know if not

src/tokenizer.rs
iffyio: > and the iterator is incremented according to the characters in the buffer

Yeah I think this is the part that we don't necessarily have a guarantee on, hence the sanity check part if we have a bug somewhere or make wrong assumption etc

eyalleshem: Ok, added the check here as well. I think this case is slightly different since the position is derived directly from the char iterator, but I don't object to adding the sanity check for extra safety.

eyalleshem:
@iffyio - I think all comments have been addressed. Let me know if there's anything else.
Another thing - could you open a feature branch for this task? Then I can change the PR destination to that branch. (I don't think I have permission to create the branch myself.)

src/tokenizer.rs
eyalleshem: Ok, added the check here as well. I think this case is slightly different since the position is derived directly from the char iterator, but I don't object to adding the sanity check for extra safety."

iffyio:
Thanks @eyalleshem! I've created a new [branch here](https://github.com/apache/datafusion-sqlparser-rs/tree/reduce-string-copying) that we can use

eyalleshem:
Thanks @iffyio! I've changed the PR to target this branch. Do we want to keep the branch protected to enforce reviews?

iffyio:
> Do we want to keep the branch protected to enforce reviews?

I unfortunately don't have permissions to do this, but I think we can manually enforce it should be fine

PR #2075, opened 2025-10-27
Reduce string copies cow

eyalleshem:
This continues the preparation for a borrowed Tokenizer (#2036) and adds more internal functions that borrow strings during tokenization.
This commit also handles places where the Tokenizer could modify the original strings. In such cases, the strategy is to create functions that return `Cow<'a, str>` and create an owned version of the string when modification is needed.
This commit is rebased on PR #2073 and will remain in draft until #2073 is reviewed.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -23,12 +23,15 @@
 
 #[cfg(not(feature = "std"))]
 use alloc::{
-    borrow::ToOwned,
+    borrow::{Cow, ToOwned},
     format,
     string::{String, ToString},
     vec,
     vec::Vec,
 };
+#[cfg(feature = "std")]
+use std::borrow::Cow;
+
 use core::iter::Peekable;
 use core::num::NonZeroU8;
 use core::str::Chars;
@@ -934,7 +937,7 @@ impl<'a> Tokenizer<'a> {
         chars: &mut State<'a>,
     ) -> Result<Option<Token>, TokenizerError> {
         chars.next(); // consume the first char
-        let word = self.tokenize_word(consumed_byte_len, chars);
+        let word = self.tokenize_word(consumed_byte_len, chars)?;
 
         // TODO: implement parsing of exponent here
         if word.chars().all(|x| x.is_ascii_digit() || x == '.') {
@@ -1008,7 +1011,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "b" or "B"
-                            let s = self.tokenize_word(b.len_utf8(), chars);
+                            let s = self.tokenize_word(b.len_utf8(), chars)?;
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1035,7 +1038,7 @@ impl<'a> Tokenizer<'a> {
                             ),
                         _ => {
                             // regular identifier starting with an "r" or "R"
-                            let s = self.tokenize_word(b.len_utf8(), chars);
+                            let s = self.tokenize_word(b.len_utf8(), chars)?;
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1054,7 +1057,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "N"
-                            let s = self.tokenize_word(n.len_utf8(), chars);
+                            let s = self.tokenize_word(n.len_utf8(), chars)?;
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1071,7 +1074,7 @@ impl<'a> Tokenizer<'a> {
                         }
                         _ => {
                             // regular identifier starting with an "E" or "e"
-                            let s = self.tokenize_word(x.len_utf8(), chars);
+                            let s = self.tokenize_word(x.len_utf8(), chars)?;
                             Ok(Some(Token::make_word(&s, None)))
                         }
                     }
@@ -1090,7 +1093,7 @@ impl<'a> Tokenizer<'a> {
                         }
                     }
                     // regular identifier starting with an "U" or "u"
-                    let s = self.tokenize_word(x.len_utf8(), chars);
+                    let s = self.tokenize_word(x.len_utf8(), chars)?;
                     Ok(Some(Token::make_word(&s, None)))
                 }
                 // The spec only allows an uppercase 'X' to introduce a hex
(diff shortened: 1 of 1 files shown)


alamb:
FYI @iffyio and @yoavcloud

eyalsatori:
Hi @iffyio  ,  could you mark this PR as ready for review? (This option is only available to those with write access to the repo)

iffyio:
@eyalleshem I updated the target branch for the PR could you update the PR to only include the new diff? I think its still picking up some of the earlier tokenizer changes after all

eyalleshem:
rebased :)

src/tokenizer.rs
```diff
@@ -1783,96 +1786,115 @@ impl<'a> Tokenizer<'a> {
...
-                        s.push(ch);
+    /// Tokenize a dollar-quoted string ($$text$$ or $tag$text$tag$), returning borrowed slices.
+    /// tag_prefix: None for $$, Some("tag") for $tag$
+    /// Returns (value: Cow<'a, str>, tag: Option<Cow<'a, str>>)
+    fn tokenize_dollar_quoted_string_borrowed(
+        &self,
+        chars: &mut State<'a>,
+        tag_prefix: Option<&'a str>,
+    ) -> Result<(Cow<'a, str>, Option<Cow<'a, str>>), TokenizerError> {
+        chars.next(); // consume $ after tag (or second $ for $$)
```
iffyio, line 1851: Can we explicitly check here that we have a `$` token being consumed?i.e. something like
```
if chars.peek() != Some('$') {
  return Err(...)
}
chars.next(); // Consume '$'
```
It looks incorrect otherwise from the method's pov that its potentially blindly skipping a token

src/tokenizer.rs
```diff
@@ -1783,96 +1786,115 @@ impl<'a> Tokenizer<'a> {
...
 
     /// Tokenize dollar preceded value (i.e: a string/placeholder)
-    fn tokenize_dollar_preceded_value(&self, chars: &mut State) -> Result<Token, TokenizerError> {
-        let mut s = String::new();
-        let mut value = String::new();
+    fn tokenize_dollar_preceded_value(
+        &self,
+        chars: &mut State<'a>,
+    ) -> Result<Token, TokenizerError> {
+        chars.next(); // consume first $
```
iffyio, line 1799: Similar comment as below that we can explicitly check what the token is before consuming it

src/tokenizer.rs
```diff
@@ -1783,96 +1786,115 @@ impl<'a> Tokenizer<'a> {
...
+            None => {
+                // Case: $$text$$
+                let mut prev: Option<char> = None;
+
+                while let Some(&ch) = chars.peek() {
+                    if prev == Some('$') && ch == '$' {
+                        chars.next(); // consume final $
+                                      // content_end is before the first $ of $$
+                        let content_end = chars.byte_pos - 2;
+                        let value = &chars.source[content_start..content_end];
```
iffyio: can we do a pass through the places where we index into the buffer, to ensure that the indexing is safe and we return an error otherwise. Maybe we could have a helper function for it for reuse

src/tokenizer.rs
```diff
@@ -2304,7 +2386,69 @@ fn peeking_next_take_while(
...
 
 fn unescape_single_quoted_string(chars: &mut State<'_>) -> Option<String> {
-    Unescape::new(chars).unescape()
+    borrow_or_unescape_single_quoted_string(chars, true).map(|cow| cow.into_owned())
+}
+
+/// Scans a single-quoted string and returns either a borrowed slice or an unescaped owned string.
+///
+/// Strategy: Scan once to find the end and detect escape sequences.
+/// - If no escapes exist (or unescape=false), return Cow::Borrowed
```
```suggestion
/// - If no escapes exist (or unescape=false), return [Cow::Borrowed]
```

src/tokenizer.rs
```diff
@@ -2304,7 +2386,69 @@ fn peeking_next_take_while(
...
+/// - If escapes exist and unescape=true, reprocess using existing Unescape logic
```
```suggestion
/// - If escapes exist and unescape=true, reprocess using existing [Unescape] logic
```

src/tokenizer.rs
```diff
@@ -2304,7 +2386,69 @@ fn peeking_next_take_while(
...
+                    chars.next(); // consume the second '
+                } else {
+                    // End of string found (including closing ')
+                    let content_end = chars.byte_pos;
+                    let full_content = &chars.source[content_start..content_end];
+
+                    // If no unescaping needed, return borrowed (without quotes)
+                    if !unescape || !has_escapes {
+                        // Strip opening and closing quotes
+                        return Some(Cow::Borrowed(&full_content[1..full_content.len() - 1]));
```
iffyio, line 2494: Similar comment here re ensureing that the indexing is safe

src/tokenizer.rs
eyalleshem: Done

src/tokenizer.rs
eyalleshem: Done

src/tokenizer.rs
eyalleshem: Added a safe_slice method that returns an error when the index is out of bounds, and replaced otherquery slicing operations to use it

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
eyalleshem: done

src/tokenizer.rs
iffyio: Great thanks!

iffyio (approved):
LGTM! Thanks @eyalleshem!

src/tokenizer.rs
```diff
@@ -2188,27 +2289,71 @@ impl<'a> Tokenizer<'a> {
...
+        &self,
+        chars: &mut State<'a>,
+        quote_end: char,
+    ) -> Result<(String, Option<char>), TokenizerError> {
+        let (cow, last_char) = self.parse_quoted_ident_borrowed(chars, quote_end)?;
+        Ok((cow.into_owned(), last_char))
+    }
+
+    /// Parse quoted identifier, returning borrowed slice when possible.
+    /// Returns `(Cow<'a, str>, Option<char>)` where the `Option<char>` is the closing quote.
```
alamb, line 2302: 🎉

PR #2077, opened 2025-10-29
Remove Whitespace Tokens from Parser

### LINKED ISSUE #2076: Eliminating whitespace from the parser logic
Hi,

At this time, whitespace tokens are stored in the parser, and are then filtered out in several distinct points in the parser logic, such as:

* https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4032-L4049
* https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4055-L4069
* https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4077-L4094
* https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4149-L4160
* https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4183-L4202

and many more.

SQL, as far as I know, is not a language that cares about spaces like Python - it should be safe to remove all concepts of whitespaces after [the tokenization process](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/tokenizer.rs#L937-L942) is complete, and this should:

* Reduce memory requirements, as whitespace tokens would not be stored anymore
* Significantly simplify parser logic by removing all of the whitespace-related logic from the parser
* Move the parser closer to a streaming logic, but that will require many more PRs

Since such a PR would require quite a bit of effort on my part, I would appreciate some feedback on it before moving forward with it.

@iffyio do you happen to have any opinion regarding such a refactoring?

Ciao,
Luca

LucaCappelletti94:
This PR implements a significant architectural refactoring by moving whitespace filtering from the parser to the tokenizer. Instead of emitting whitespace tokens (spaces, tabs, newlines, comments) and filtering them throughout the parser logic, the tokenizer now consumes whitespace during tokenization and never emits these tokens.

While some duplicated logic still remains in the parser (to be addressed in future PRs), this change eliminates a substantial amount of looping overhead. This PR sets the groundwork for a cleaner streaming version, where the tokens are parsed simultaneously as the statements, with no parser memory and only local context passed between parser function calls.

Fixes #2076

## Motivation

As discussed in #2076, whitespace tokens were being filtered at numerous points throughout the parser. This approach had several drawbacks:

- **Poor separation of concerns**: Whitespace handling was scattered across both tokenizer and parser
- **Memory overhead**: Whitespace tokens were stored in memory unnecessarily
- **Code duplication**: Multiple loops throughout the parser to skip whitespace tokens, looking ahead or backwards for non-whitespace tokens
- **Performance**: Each token access required checking and skipping whitespace tokens

The parser had extensive whitespace-handling logic scattered throughout:

**Functions with whitespace-skipping loops:**
- [`peek_tokens_with_location`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4028-L4049) - loops to skip whitespace
- [`peek_tokens_ref`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4055-L4069) - loops to skip whitespace
- [`peek_nth_token_ref`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4077-L4094) - loops to skip whitespace
- [`advance_token`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4149-L4160) - loops to skip whitespace
- [`prev_token`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4183-L4202) - loops backward to skip whitespace

**Special variant functions that are now obsolete:**
- [`peek_token_no_skip`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4096-L4100) - **removed entirely** (no longer needed)
- [`peek_nth_token_no_skip`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4102-L4111) - **removed entirely** (no longer needed)
- [`next_token_no_skip`](https://github.com/apache/datafusion-sqlparser-rs/blob/67684c84d4c2589356c411ea4917dcf1defcd77c/src/parser/mod.rs#L4140-L4144) - **removed entirely** (no longer needed)

Since SQL is not a whitespace-sensitive language (unlike Python), so it *should be* safe to remove whitespace tokens entirely after tokenization.

## Handling Edge Cases

While SQL is generally not whitespace-sensitive, there are specific edge cases that require careful consideration:

### 1. PostgreSQL COPY FROM STDIN

The `COPY FROM STDIN` statement requires preserving the actual data content, which may include meaningful whitespace and newlines. The data section is treated as raw input that should be parsed according to the specified format (tab-delimited, CSV, etc.).

**Solution**: The tokenizer now properly handles this by consuming the data as a single token. The parser then actually parses the body of the CSV-like string, which was not actually done correctly before this refactoring. I have extended the associated tests appropriately.

### 2. Hyphenated and path identifiers

The tokenizer now includes enhanced logic for hyphenated identifier parsing with proper validation:

- Detects when hyphens/paths/tildes are part of identifiers vs. operators
- Validates that identifiers don't start with digits after hyphens
- Ensures identifiers don't end with trailing hyphens
- Handles the whitespace-dependent context correctly

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -59,23 +59,23 @@ pub use self::dcl::{
     AlterRoleOperation, CreateRole, ResetConfig, RoleOption, SecondaryRoles, SetConfigValue, Use,
 };
 pub use self::ddl::{
-    Alignment, AlterColumnOperation, AlterConnectorOwner, AlterIndexOperation,
-    AlterPolicyOperation, AlterSchema, AlterSchemaOperation, AlterTable, AlterTableAlgorithm,
-    AlterTableLock, AlterTableOperation, AlterTableType, AlterType, AlterTypeAddValue,
-    AlterTypeAddValuePosition, AlterTypeOperation, AlterTypeRename, AlterTypeRenameValue,
-    ClusteredBy, ColumnDef, ColumnOption, ColumnOptionDef, ColumnOptions, ColumnPolicy,
-    ColumnPolicyProperty, ConstraintCharacteristics, CreateConnector, CreateDomain,
+    Alignment, AlterColumnOperation, AlterConnectorOwner, AlterIndexOperation, AlterOperator,
+    AlterOperatorOperation, AlterPolicyOperation, AlterSchema, AlterSchemaOperation, AlterTable,
+    AlterTableAlgorithm, AlterTableLock, AlterTableOperation, AlterTableType, AlterType,
+    AlterTypeAddValue, AlterTypeAddValuePosition, AlterTypeOperation, AlterTypeRename,
+    AlterTypeRenameValue, ClusteredBy, ColumnDef, ColumnOption, ColumnOptionDef, ColumnOptions,
+    ColumnPolicy, ColumnPolicyProperty, ConstraintCharacteristics, CreateConnector, CreateDomain,
     CreateExtension, CreateFunction, CreateIndex, CreateOperator, CreateOperatorClass,
     CreateOperatorFamily, CreateTable, CreateTrigger, CreateView, Deduplicate, DeferrableInitial,
     DropBehavior, DropExtension, DropFunction, DropOperator, DropOperatorClass, DropOperatorFamily,
     DropOperatorSignature, DropTrigger, GeneratedAs, GeneratedExpressionMode, IdentityParameters,
     IdentityProperty, IdentityPropertyFormatKind, IdentityPropertyKind, IdentityPropertyOrder,
     IndexColumn, IndexOption, IndexType, KeyOrIndexDisplay, Msck, NullsDistinctOption,
-    OperatorArgTypes, OperatorClassItem, OperatorPurpose, Owner, Partition, ProcedureParam,
-    ReferentialAction, RenameTableNameKind, ReplicaIdentity, TagsColumnOption, TriggerObjectKind,
-    Truncate, UserDefinedTypeCompositeAttributeDef, UserDefinedTypeInternalLength,
-    UserDefinedTypeRangeOption, UserDefinedTypeRepresentation, UserDefinedTypeSqlDefinitionOption,
-    UserDefinedTypeStorage, ViewColumnDef,
+    OperatorArgTypes, OperatorClassItem, OperatorOption, OperatorPurpose, Owner, Partition,
+    ProcedureParam, ReferentialAction, RenameTableNameKind, ReplicaIdentity, TagsColumnOption,
+    TriggerObjectKind, Truncate, UserDefinedTypeCompositeAttributeDef,
+    UserDefinedTypeInternalLength, UserDefinedTypeRangeOption, UserDefinedTypeRepresentation,
+    UserDefinedTypeSqlDefinitionOption, UserDefinedTypeStorage, ViewColumnDef,
 };
 pub use self::dml::{Copy, Delete, Insert, Update};
 pub use self::operator::{BinaryOperator, UnaryOperator};
@@ -3383,6 +3383,11 @@ pub enum Statement {
     /// ```
     AlterType(AlterType),
     /// ```sql
+    /// ALTER OPERATOR
+    /// ```
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-alteroperator.html)
+    AlterOperator(AlterOperator),
+    /// ```sql
     /// ALTER ROLE
     /// ```
     AlterRole {
@@ -4921,6 +4926,7 @@ impl fmt::Display for Statement {
             Statement::AlterType(AlterType { name, operation }) => {
                 write!(f, "ALTER TYPE {name} {operation}")
             }
+            Statement::AlterOperator(alter_operator) => write!(f, "{alter_operator}"),
             Statement::AlterRole { name, operation } => {
                 write!(f, "ALTER ROLE {name} {operation}")
             }
@@ -9764,8 +9770,8 @@ impl fmt::Display for ShowCharset {
         } else {
             write!(f, " CHARACTER SET")?;
         }
-        if self.filter.is_some() {
-            write!(f, " {}", self.filter.as_ref().unwrap())?;
+        if let Some(filter) = &self.filter {
+            write!(f, " {filter}")?;
         }
         Ok(())
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1194,7 +1194,11 @@ impl<'a> Parser<'a> {
                     let mut id_parts: Vec<Ident> = vec![match t {
                         Token::Word(w) => w.into_ident(next_token.span),
                         Token::SingleQuotedString(s) => Ident::with_quote('\'', s),
-                        _ => unreachable!(), // We matched above
+                        _ => {
+                            return Err(ParserError::ParserError(
+                                "Internal parser error: unexpected token type".to_string(),
+                            ))
+                        }
                     }];
 
                     while self.consume_token(&Token::Period) {
@@ -1641,7 +1645,11 @@ impl<'a> Parser<'a> {
                     Token::PGSquareRoot => UnaryOperator::PGSquareRoot,
                     Token::PGCubeRoot => UnaryOperator::PGCubeRoot,
                     Token::AtSign => UnaryOperator::PGAbs,
-                    _ => unreachable!(),
+                    _ => {
+                        return Err(ParserError::ParserError(
+                            "Internal parser error: unexpected unary operator token".to_string(),
+                        ))
+                    }
                 };
                 Ok(Expr::UnaryOp {
                     op,
@@ -1709,18 +1717,22 @@ impl<'a> Parser<'a> {
                 Ok(Expr::Value(self.parse_value()?))
             }
             Token::LParen => {
-                let expr = if let Some(expr) = self.try_parse_expr_sub_query()? {
-                    expr
-                } else if let Some(lambda) = self.try_parse_lambda()? {
-                    return Ok(lambda);
-                } else {
-                    let exprs = self.parse_comma_separated(Parser::parse_expr)?;
-                    match exprs.len() {
-                        0 => unreachable!(), // parse_comma_separated ensures 1 or more
-                        1 => Expr::Nested(Box::new(exprs.into_iter().next().unwrap())),
-                        _ => Expr::Tuple(exprs),
-                    }
-                };
+                let expr =
+                    if let Some(expr) = self.try_parse_expr_sub_query()? {
+                        expr
+                    } else if let Some(lambda) = self.try_parse_lambda()? {
+                        return Ok(lambda);
+                    } else {
+                        let exprs = self.parse_comma_separated(Parser::parse_expr)?;
+                        match exprs.len() {
+                            0 => return Err(ParserError::ParserError(
+                                "Internal parser error: parse_comma_separated returned empty list"
+                                    .to_string(),
+                            )),
+                            1 => Expr::Nested(Box::new(exprs.into_iter().next().unwrap())),
+                            _ => Expr::Tuple(exprs),
+                        }
+                    };
                 self.expect_token(&Token::RParen)?;
                 Ok(expr)
             }
@@ -3591,7 +3603,9 @@ impl<'a> Parser<'a> {
                         right: Box::new(right),
                         is_some: keyword == Keyword::SOME,
                     },
-                    _ => unreachable!(),
+                    unexpected_keyword => return Err(ParserError::ParserError(
+                        format!("Internal parser error: expected any of {{ALL, ANY, SOME}}, got {unexpected_keyword:?}"),
+                    )),
                 })
             } else {
                 Ok(Expr::BinaryOp {
@@ -5526,13 +5540,14 @@ impl<'a> Parser<'a> {
         } else {
             None
         };
-        let option = self
-            .parse_one_of_keywords(&[Keyword::CASCADE, Keyword::RESTRICT])
-            .map(|keyword| match keyword {
-                Keyword::CASCADE => ReferentialAction::Cascade,
-                Keyword::RESTRICT => ReferentialAction::Restrict,
-                _ => unreachable!(),
-            });
+        let option = match self.parse_one_of_keywords(&[Keyword::CASCADE, Keyword::RESTRICT]) {
+            Some(Keyword::CASCADE) => Some(ReferentialAction::Cascade),
+            Some(Keyword::RESTRICT) => Some(ReferentialAction::Restrict),
+            Some(unexpected_keyword) => return Err(ParserError::ParserError(
+                format!("Internal parser error: expected any of {{CASCADE, RESTRICT}}, got {unexpected_keyword:?}"),
+            )),
+            None => None,
+        };
         Ok(Statement::DropTrigger(DropTrigger {
             if_exists,
             trigger_name,
@@ -5582,7 +5597,9 @@ impl<'a> Parser<'a> {
                 match self.expect_one_of_keywords(&[Keyword::ROW, Keyword::STATEMENT])? {
                     Keyword::ROW => TriggerObject::Row,
                     Keyword::STATEMENT => TriggerObject::Statement,
-                    _ => unreachable!(),
+                    unexpected_keyword => return Err(ParserError::ParserError(
+                        format!("Internal parser error: unexpected keyword `{unexpected_keyword}` in ROW/STATEMENT"),
+                    )),
                 };
 
             Some(if include_each {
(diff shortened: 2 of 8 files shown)


LucaCappelletti94:
The newly added dependency `csv` does not have an analogous `no-std` version, so the alloc-only test is currently failing. I will see whether a PR to `csv` adding support for an `alloc-only` mode is possible. I have used `csv` because I do not want to re-implement CSV-like document parsing in `sqlparser`.

LucaCappelletti94:
I have decided to replace the [`csv`](https://github.com/BurntSushi/rust-csv) crate with a custom solution as right now that crate does not provide an alloc-only feature, and it seems rather complex to make a PR making it available.

src/tokenizer.rs
```diff
@@ -449,29 +449,6 @@ impl Word {
...
-#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub enum Whitespace {
-    Space,
-    Newline,
-    Tab,
-    SingleLineComment { comment: String, prefix: String },
-    MultiLineComment(String),
-}
```
Viicos, line 461: While it would make sense for the parser to not have to deal with whitespace _including_ comments, it will be useful to preserve comment tokens. This is really useful when trying to provide IDE completions, where you don't want anything to be suggested in inside a comment.

LucaCappelletti94: Hi @Viicos, I have nothing against comments - in fact, a PhD student in my team is working on the #2069 PR for handling them in a more structured manner in the AST. They just should not exist in the form of whitespace. Could you kindly expand on your IDE completions note? I am not sure I understood it.

Viicos: > They just should not exist in the form of whitespace.

Agree, maybe they could be made their own token.

> Could you kindly expand on your IDE completions note? I am not sure I understood it.

I'm currently experimenting with the tokenizer and parser to use it in a SQL query input we have for our project at work. From a `(line, col)` position, I'm writing some logic to provide useful completions at this position (by completions, I mean [this](https://microsoft.github.io/language-server-protocol/img/vscode-css-code-complete.png)).

If you are currently writing a comment, the token immediately before the position is going to be a comment token, and in this case I can short-circuit and returns an empty list of competions.

As an example, this is how the Ruff project (they are working on a Python language server) [is doing it](https://github.com/astral-sh/ruff/blob/bd8812127daa556bd86fa81c9a79f5f49a2feaa8/crates/ty_ide/src/completion.rs#L930-L933).

Viicos: Actually having a comment token kind would defeat the purpose of this PR, because the logic in the parser to skip those comment tokens would be the same.

The Ruff parser solves this by having a [`TokenSource`](https://github.com/astral-sh/ruff/blob/bd8812127daa556bd86fa81c9a79f5f49a2feaa8/crates/ruff_python_parser/src/token_source.rs#L11) struct, acting as a bridge between the lexer/tokenizer and parser. It has [a couple methods](https://github.com/astral-sh/ruff/blob/bd8812127daa556bd86fa81c9a79f5f49a2feaa8/crates/ruff_python_parser/src/token_source.rs#L138-L161) to bump the tokens, ignoring the trivia tokens (in our case, that would only be the comment tokens). Maybe we could take inspiration from this pattern?

LucaCappelletti94: Somewhat - the idea in the aforementioned PR was to have the concept of leading and interstitial comment, with the formed getting into the ast and the latter being dropped as white space

Viicos: I see; I still think it would be important for any kind of comment to be preserved as tokens (my use case is related to completions, but it can be also useful if you want to implement a formatting mechanism that preserves comments, or if a comment can be used to e.g. specify a linter directive -- presumably would be relevant for interstitial comments as well).

I commented some thoughts on this PR ;)

src/ast/mod.rs
```diff
@@ -4579,19 +4574,97 @@ impl fmt::Display for Statement {
...
+                        CopyOption::Quote(c) => {
+                            quote = *c;
+                        }
+                        CopyOption::Escape(c) => {
+                            escape = *c;
+                        }
+                        CopyOption::Null(null) => {
+                            null_symbol = null;
+                        }
+                        _ => {}
```
iffyio: Can we make the match exhaustive instead of using wildcard to catch-all?
i.e.
```rust
CopyOption::Encoding(_) | CopyOption::ForceNull(_) ... => {}
```
so that we're guided to revisit this code if the CopyOption enum is extended in the future

src/ast/mod.rs
```diff
@@ -4579,19 +4574,97 @@ impl fmt::Display for Statement {
...
+                                        quote = *c;
+                                    }
+                                    CopyLegacyCsvOption::Escape(c) => {
+                                        escape = *c;
+                                    }
+                                    _ => {}
+                                }
+                            }
+                        }
+                        _ => {}
```
iffyio: Similarly, in both wildcards we can use an exhaustive match in place.

src/dialect/mod.rs
```diff
@@ -178,6 +178,16 @@ pub trait Dialect: Debug + Any {
     /// Determine if a character is a valid unquoted identifier character
     fn is_identifier_part(&self, ch: char) -> bool;
 
+    /// Returns whether the dialect supports hyphenated identifiers
```
iffyio: Can we elaborate here what an hyphenated identifier means? (preferrably with an example)

src/dialect/mod.rs
```diff
@@ -178,6 +178,16 @@ pub trait Dialect: Debug + Any {
...
+    fn supports_hyphenated_identifiers(&self) -> bool {
+        false
+    }
+
+    /// Returns whether the dialect supports path-like identifiers
```
iffyio: Similarly, I think we would need to explain what path-like identifier means

src/parser/mod.rs
```diff
@@ -9597,6 +9533,178 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn parse_csv_body(
```
```suggestion
    fn parse_csv_body(
```

src/parser/mod.rs
```diff
@@ -9597,6 +9533,178 @@ impl<'a> Parser<'a> {
...
+                            CopyLegacyCsvOption::Escape(c) => {
+                                escape = *c;
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
```
iffyio: this looks like a duplicate of the logic on the display path, can we pull it into a function and reuse in both places?

tests/sqlparser_common.rs
```diff
@@ -3589,6 +3589,7 @@ fn test_double_value() {
 
     for (input, expected) in test_cases {
         for (i, expr) in input.iter().enumerate() {
+            println!("Testing expression: {}", expr);
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -9597,6 +9533,178 @@ impl<'a> Parser<'a> {
...
+
+        // Simple CSV parser
```
iffyio: thinking instead of having a full csv parser it would probably be better/simpler to come up with a way that lets us just grab everything  up until the terminal tokens `\.` - looks like that was how it was handled previously

src/tokenizer.rs
```diff
@@ -896,14 +929,37 @@ impl<'a> Tokenizer<'a> {
...
+                }
+            }
 
+            let span = location.span_to(state.location());
+            cs_handler.update(&token);
             buf.push(TokenWithSpan { token, span });
-
             location = state.location();
+
+            if cs_handler.is_in_copy_from_stdin() {
```
iffyio, line 954: I'm thinking this solution with the `CopyStdinHandler` isn't ideal because we're effectively introducing a parser within the tokenizer.

One thought related to the comment on the 'simple csv parser' - if it would work to have the parser when looking to parse a csv string blindly consume all tokens until the terminal (new lines can be detected by inspecting the token's span), I imagine the tokenizer might not require any supporting logic

src/ast/mod.rs
LucaCappelletti94: Ok

src/ast/mod.rs
LucaCappelletti94: Ok

src/parser/mod.rs
LucaCappelletti94: It is nearly a duplicate, yes, and that is why I wanted to handle both with the `csv` crate, but there are some significant differences which I found a bit hard to normalize.

src/parser/mod.rs
LucaCappelletti94: Yes, but as it was handled previously it did not provide an error message when the CSV was not a CSV. I am not sure what is preferable.

src/tokenizer.rs
LucaCappelletti94: Yes, I agree it is sub-optimal, but the syntax in question is not valid SQL and it relies on whitespaces (e.g. when the input are TSVs), so it is necessary for the tokenizer to know which is the current parsing context. It really isn't ideal, but I don't see any alternative.

src/ast/mod.rs
LucaCappelletti94: Moved all of the logic into a new struct `CsvFormatOptions` and added exaustive match.

src/ast/mod.rs
LucaCappelletti94: Done

src/dialect/mod.rs
LucaCappelletti94: Done, added doc tests as well.

src/parser/mod.rs
LucaCappelletti94: Simplified.

src/dialect/mod.rs
LucaCappelletti94: Done, also added doctest.

src/parser/mod.rs
iffyio: I think we can parse it as a raw string, and the actual csv validation would be left to any downstream crates in this case

LucaCappelletti94: Ok

LucaCappelletti94: I have now removed CSV validation. I have refactored the relative tests to actually check that invalid CSV does not raise an error.

src/tokenizer.rs
iffyio: I was thinking on the parser end, we have the delimeter context to know if whitespace is in play, if so the parsing code could manually inject the whitespace character between tokens as it consumes them?
i.e. given two arbitrary tokens, by comparing their location we can tell if they're right next to each other, separated by one or more whitespaces or by one or more newlines

LucaCappelletti94: I am not sure I understood - at this time, the idea is to treat the csv/tsv as a string with no parsing, right? It just needs to remain untouched by the tokeniser.

In order to remain untouched, we need to detect we are in a csv context, and to do so we need a small part of the parser logic in the tokeniser (i.e the helper struct)

iffyio: Yeah so the thinking was that the tokenizer is as usual agnostic of the context.

So using this as an example input

```sql
COPY actor (actor_id) FROM STDIN;
1	PENELOPE
2	NICK
\.
```

The tokenizer produces the token stream

```rust
[COPY, actor, (, actor_id, ), FROM, STDIN, ;, 1, PENELOPE, 2, NICK, \, .]
```

Then the parser, after parsing the `STDIN;`, knows that the delimeter is whitespace implicitly. Then it consumes the tokens `[1, PENELOPE ...]` interspersed with whitespace in-between tokens, to build the actual csv string.

LucaCappelletti94: It would work for correctly formatted TSVs for which the separator and number of header columns is provided, but would have weird behaviours with more cursed string inputs, for instance an empty row would be just a sequence of tabs.

I think it may be okay to add this logic duplication for the tokeniser in these specific cases because we are dealing with something that is not SQL, and breaks otherwise valid assumptions.


Ideally, in the future, we might be able to make the tokeniser and parser co-inform each other of the current context while parsing so to avoid any such code duplication and allow for streaming parsing.

iffyio: > for instance an empty row would be just a sequence of tabs

how do you mean in this case (what's producing the sequence of tabs)?

LucaCappelletti94: Basically, if you were to insert one or more None values (or empty strings), they would be represented as empty strings (if I am not mistaken) and therefore those values would be represented as `\t{empty string}\t`, and the tokenizer would strip the subsequent tabs, making that row of the CSV not reconstructible with the approach you described.

Suppose you are parsing:

```SQL
OPY public.actor (actor_id, first_name, last_name, last_update, value) FROM stdin;
1	PENELOPE	GUINESS	2006-02-15 09:34:33 0.11111
2	NICK			2006-02-15 09:34:33 0.22222
3	ED	CHASE	2006-02-15 09:34:33 0.312323
4	JENNIFER	DAVIS	2006-02-15 09:34:33 0.3232
\.
```

At the entry with ID 2, you notice that a value is left blank to represent an optional entry, or analogously, an empty string which would still be a valid value. The tokenizer should leave these tabs to avoid loosing that information, as once it is stripped it cannot be reconstructed.

That being said, I personally find this syntax cursed. It is light years from proper use of SQL lol.

src/tokenizer.rs
xitep: hello, I thought i can contribute some insight from my side:

> I still think it would be important for any kind of comment to be preserved as tokens

i agree; one real world use-case are surely the (not yet supported) oracle optimiser hints: https://docs.oracle.com/cd/B13789_01/server.101/b10752/hintsref.htm

> They [comments] just should not exist in the form of whitespace.

:+1: yes, they should not! (in fact i had to introduce another type to expose them as "comments" :confused:  #2107 )

> ... can be also useful if you want to implement a formatting mechanism that preserves comments ...

that's precisely my challenge

src/tokenizer.rs
```diff
@@ -1877,12 +1962,66 @@ impl<'a> Tokenizer<'a> {
...
+        chars: &mut State,
+        prev_keyword: Option<Keyword>,
+    ) -> Result<String, TokenizerError> {
         let mut s = first_chars.into();
         s.push_str(&peeking_take_while(chars, |ch| {
             self.dialect.is_identifier_part(ch)
         }));
-        s
+
+        while !matches!(prev_keyword, Some(Keyword::SELECT))
```
iffyio, line 1976: not sure I followed if/why the previous token being a `SELECT` is sufficient to look to parse the hyphenated identifiers 🤔

LucaCappelletti94: At this time it passes both the pre-existing tests and the tests I added to try and break it.

LucaCappelletti94:
> Thanks @LucaCappelletti94 and sorry for the delay in getting back to this. I haven't looked at the PR in depth but I think so far I'm not comfortable with the approach as I mentioned previously of mixing in parsing logic into the tokenizer. I think the complexity add is too significant going forward to warrant the pay off

I think that it should not be looked as the added complexity of the tokenizer but how much it simplifies the parser. I agree that the duplicated parser context logic is extremely undesirable here in this form.

In my opinion the cleanest solution should be to execute tokenization and parsing side-by-side, with the tokenizer refactored so that it can always use the information of what is the current context to optimally execute the tokenization. Of course, such thing is solely needed due to some pattern-breaking non-sql features. Would you prefer such an approach?

LucaCappelletti94:
This refactor is mostly motivated by three reasons: improved code cleanliness, and ideally, reduced memory footprint and time requirements. Regarding code cleanliness, it removes certain code complexity from the parser but increases complexity in the tokenizer, so the benefit is unclear. I proceeded to measure the time requirements in the largest SQL documents I have, and there is no significant change in performance between main and this branch. I have not measured changes in memory footprint yet, but given the lack of clear benefit in the aforementioned points, I would proceed to close the PR.

PR #2079, opened 2025-10-29
feat: Add RESET to the base dialect #2078

### LINKED ISSUE #2078: Add support for RESET statement
The RESET keyword is available, but the statement is not:
```sql
RESET configuration_parameter;
RESET ALL;
```

watford-ep:
This adds support for the RESET statement to the base dialect.

- Closes #2078

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2787,10 +2787,11 @@ impl fmt::Display for Declare {
 }
 
 /// Sql options of a `CREATE TABLE` statement.
-#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum CreateTableOptions {
+    #[default]
     None,
     /// Options specified using the `WITH` keyword.
     /// e.g. `WITH (description = "123")`
@@ -2817,12 +2818,6 @@ pub enum CreateTableOptions {
     Plain(Vec<SqlOption>),
 
     TableProperties(Vec<SqlOption>),
-}
-
-impl Default for CreateTableOptions {
-    fn default() -> Self {
-        Self::None
-    }
 }
 
 impl fmt::Display for CreateTableOptions {
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -17636,22 +17636,18 @@ fn parse_generic_unary_ops() {
 #[test]
 fn parse_reset_statement() {
     match verified_stmt("RESET some_parameter") {
-        Statement::Reset(ResetStatement { reset }) => match reset {
-            Reset::ConfigurationParameter(o) => {
-                assert_eq!(o, ObjectName::from(vec!["some_parameter".into()]))
-            }
-            _ => unreachable!(),
-        },
+        Statement::Reset(ResetStatement {
+            reset: Reset::ConfigurationParameter(o),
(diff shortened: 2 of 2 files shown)


src/dialect/mod.rs
```diff
@@ -1207,6 +1207,17 @@ pub trait Dialect: Debug + Any {
...
+    /// Returns true if the dialect supports the `RESET` statement
+    /// for resetting session variables.
+    ///
+    /// ```sql
+    /// RESET configuration_parameter;
+    /// RESET ALL;
+    /// ```
+    fn supports_reset(&self) -> bool {
+        false
+    }
```
```suggestion
```
I think we can let the parser always accept the statement if it shows up, without considering the dialect

src/parser/mod.rs
```diff
@@ -17723,6 +17724,18 @@ impl<'a> Parser<'a> {
             _ => self.expected("expected option value", self.peek_token()),
         }
     }
+
+    fn parse_reset(&mut self) -> Result<Statement, ParserError> {
+        // RESET { ALL | <configuration_parameter> }
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -17723,6 +17724,18 @@ impl<'a> Parser<'a> {
             _ => self.expected("expected option value", self.peek_token()),
         }
     }
+
+    fn parse_reset(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 17733:
```suggestion
    /// Parse a [Statement::Reset]
    fn parse_reset(&mut self) -> Result<Statement, ParserError> {
```

src/parser/mod.rs
```diff
@@ -18529,4 +18542,29 @@ mod tests {
...
+        let sql = "RESET ALL";
+        let ast = Parser::parse_sql(&PostgreSqlDialect {}, sql).unwrap();
+        assert_eq!(
+            ast,
+            vec![Statement::Reset(ResetStatement { reset: Reset::ALL })]
+        );
+    }
+
+    #[test]
+    fn test_reset_parameter() {
```
iffyio: Can we merge the two tests into the same `parse_reset()` function? since they belong to the same feature

src/parser/mod.rs
```diff
@@ -18529,4 +18542,29 @@ mod tests {
             assert!(Parser::parse_sql(&GenericDialect, &sql).is_err());
         }
     }
+
+    #[test]
+    fn test_reset_all() {
+        let sql = "RESET ALL";
+        let ast = Parser::parse_sql(&PostgreSqlDialect {}, sql).unwrap();
```
```suggestion
        let ast = verified_stmt(sql);
```
we should use the built in functions for the tests - those additionally asserts roundtrip behavior

tests/sqlparser_postgres.rs
```diff
@@ -6652,3 +6652,30 @@ fn parse_foreign_key_match_with_actions() {
 
     pg_and_generic().verified_stmt(sql);
 }
+
+#[test]
```
iffyio: I think we can move and merge this into the test function in sqlparser_common

src/dialect/mod.rs
watford-ep: Can do. ~I think out of the list of dialects only ANSI, sqlite, and Hive lack support (Hive SQL doesn't take a parameter).~

src/parser/mod.rs
watford-ep: I've removed this test as it duplicated the coverage in sqlparser_common.

src/dialect/mod.rs
watford-ep: I updated the ticket to include exactly which dialects support RESET, and its a bit touch and go. Right now this seems reasonable (in that it accepts a very reasonable use, though not strictly correct for all dialects).

iffyio:
@watford-ep could you take a look at resolving the ci failures?

watford-ep:
@iffyio resolved.

PR #2080, opened 2025-10-30
Added TIMESTAMP_NTZ type support with precision to Snowflake dialect

romanoff:
Adding support for optional precision for `TIMESTAMP_NTZ` type in snowflake. Example: `TIMESTAMP_NTZ(9)`
It's supported in spec: https://docs.snowflake.com/en/sql-reference/data-types-datetime
Specifically
```
All timestamp variations, as well as the TIMESTAMP alias, support an optional precision parameter for
fractional seconds (for example, TIMESTAMP(3)). Timestamp precision can range from 0 (seconds) 
to 9 (nanoseconds). The default precision is 9.
```
Example of sql:
```sql
SELECT CAST(TO_CHAR(CURRENT_DATE, 'YYYY-MM-DD')||' 01:00:00' AS TIMESTAMP_NTZ(9))
```

### DIFF
--- a/src/ast/data_type.rs
+++ b/src/ast/data_type.rs
@@ -375,7 +375,7 @@ pub enum DataType {
     /// Databricks timestamp without time zone. See [1].
     ///
     /// [1]: https://docs.databricks.com/aws/en/sql/language-manual/data-types/timestamp-ntz-type
-    TimestampNtz,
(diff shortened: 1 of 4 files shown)


iffyio (approved):
LGTM! Thanks @romanoff!

PR #2081, opened 2025-10-30
Make `BitwiseNot` ("~") available for all dialects, not just PostgreSQL

alexander-beedie:
The bitwise "NOT" operator `"~"` is currently gated behind the PostgreSQL dialect, and is given the `UnaryOperator` enum value `PGBitwiseNot`.

However,  this operator (with the same meaning) is actually _extremely_ common, and is available across a wide range of SQL dialects; for example:

* Databricks [[docs](https://docs.databricks.com/aws/en/sql/language-manual/functions/tildesign)]
* Spark SQL [[docs](https://spark.apache.org/docs/latest/sql-ref-operators.html#operators)]
* DuckDB [[docs](https://duckdb.org/docs/stable/sql/functions/numeric#numeric-operators)]
* BigQuery [[docs](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/operators)]
* SQLite [[docs](https://www.sqlite.org/lang_expr.html)]
* MariaDB [[docs](https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/bit-functions-and-operators/bitwise-not)]
* MySQL [[docs](https://dev.mysql.com/doc/refman/en/bit-functions.html)]
* SQL Server [[docs](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-not-transact-sql)]
* RisingWave [[docs](https://docs.risingwave.com/sql/functions/mathematical#mathematical-operators)]
* Vertica [[docs](https://docs.vertica.com/25.3.x/en/sql-reference/language-elements/operators/bitwise-operators/)]
* _etc..._

Given the extent of backend implementation, it seems worth bringing it out from behind the PostgreSQL dialect-gate and renaming the enum value to the more generic `BitwiseNot` instead of `PGBitwiseNot`.

### DIFF
--- a/src/ast/operator.rs
+++ b/src/ast/operator.rs
@@ -33,35 +33,35 @@ use super::display_separated;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum UnaryOperator {
+    /// `@-@` Length or circumference (PostgreSQL/Redshift geometric operator)
+    /// see <https://www.postgresql.org/docs/9.5/functions-geometry.html>
+    AtDashAt,
+    /// Unary logical not operator: e.g. `! false` (Hive-specific)
+    BangNot,
+    /// Bitwise Not, e.g. `~9`
+    BitwiseNot,
+    /// `@@` Center (PostgreSQL/Redshift geometric operator)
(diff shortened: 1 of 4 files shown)


iffyio (approved):

alexander-beedie:
> LGTM! Thanks @alexander-beedie!

Awesome; thanks for the super-quick review/merge :)

PR #2083, opened 2025-10-31
Add snowflake dynamic table parsing

romanoff:
Add snowflake dynamic table parsing. Reference to spec: 
https://docs.snowflake.com/en/sql-reference/sql/alter-dynamic-table

Example of sql:
```sql
ALTER DYNAMIC TABLE MY_DYNAMIC_TABLE REFRESH;
ALTER DYNAMIC TABLE MY_DYNAMIC_TABLE SUSPEND;
ALTER DYNAMIC TABLE MY_DYNAMIC_TABLE RESUME;
```

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -3553,6 +3553,20 @@ impl Spanned for DropExtension {
     }
 }
 
+/// Table type for ALTER TABLE statements.
+/// Used to distinguish between regular tables, Iceberg tables, and Dynamic tables.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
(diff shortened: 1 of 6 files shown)


src/ast/ddl.rs
```diff
@@ -3559,6 +3583,8 @@ impl fmt::Display for AlterTable {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if self.iceberg {
             write!(f, "ALTER ICEBERG TABLE ")?;
+        } else if self.dynamic {
+            write!(f, "ALTER DYNAMIC TABLE ")?;
```
iffyio: Can we replace the `iceberg:bool` field with an enum instead? thinking so that we don't accumulate booleans over time (e.g. its not clear or enforceable that iceberg and dynamic are mutually exclusive).

e.g.
```rust
enum AlterTableType {
    Iceberg
    Dynamic
}
Statement::AlterTable { table_type: AlterTableType }
```

romanoff: Thank you. Addressed your comment.

PR #2085, opened 2025-11-03
Add support for `INSERT INTO VALUE`

etgarperets:


src/ast/query.rs
```diff
@@ -3135,12 +3135,32 @@ pub struct Values {
     /// Was there an explicit ROWs keyword (MySQL)?
     /// <https://dev.mysql.com/doc/refman/8.0/en/values.html>
     pub explicit_row: bool,
+    // MySql supports both VALUES and VALUE keywords.
+    // <https://dev.mysql.com/doc/refman/9.2/en/insert.html>
```
iffyio, line 3139: Can we add this documentation to the `ValuesKeyword` enum? (its ok if we want to duplicate it here as well, primarily thinking the comment is most visible on the enum itself)

Also note it should be a doc comment rather than a regular code comment

iffyio: Actually given that we only support two values here, it would probably be simpler with a bool.
```rust
value_keyword: bool
```

tests/sqlparser_mysql.rs
```diff
@@ -4353,3 +4362,41 @@ fn test_create_index_options() {
...
+                    for_clause: None,
+                    settings: None,
+                    format_clause: None,
+                    pipe_operators: vec![],
+                })),
+                source
+            );
+        }
+        _ => unreachable!(),
+    }
```
```suggestion
    mysql_and_generic().verified_stmt(sql);
```
I think this format should suffice since we cover the AST assertions in other tests

iffyio: Also instead of adding a oneoff test here, can we add a test case to the existing test function [here](https://github.com/SatoriCyber/datafusion-sqlparser-rs/blob/0ebcaae646d4f52c23462b9715fd92f9fd335b7b/tests/sqlparser_common.rs#L99)?

etgarperets:
@iffyio I added your suggestions, ready for review.

Issue #1548, opened 2024-11-25
[EPIC] Complete Span (source location) information / feature


alamb:
This ticket tracks the work remaining to complete adding source location information into sqlparser

## Background
- @Nyrox and @iffyio introduced the foundations for storing source location information in the AST nodes in https://github.com/apache/datafusion-sqlparser-rs/pull/1435. This information can be used to provide more specific error messages, and potentially syntax highlighting among other great things.
- In order to 1. minimize the disruption to downstream projects that use `sqlparser-rs` and 2. avoid a single massive PR and 3. work together as a community, we are implementing this feature incrementally over several releases.

Let's use this ticket to organize needed / remaining work. If you find additional features are needed / issues, please leave a comment on this ticket

## Tasks
- [ ] Look into reducing AST size with smaller offset sizes (e.g. `u32` rather than `usize`)

alamb:
I think the initial thing we should do to kick this project is get a test pattern setup:
- https://github.com/apache/datafusion-sqlparser-rs/issues/1563

Then we should be good to start cranking out location information

eliaperantoni:
I noticed that `CASE .. END` expressions don't have correct spans because the locations of the heading and trailing keywords are not considered. ie. the following test fails:

```rust
#[test]
fn test_case_expr_span() {
    let sql = "CASE 1 WHEN 2 THEN 3 ELSE 4 END";
    let mut parser = Parser::new(&GenericDialect {}).try_with_sql(sql).unwrap();
    assert_eq!(
        parser.parse_expr().unwrap().span(),
        Span::new(Location::new(1, 1), Location::new(1, sql.len() as u64 + 1))
    );
}
```

I'm about to open a PR that fixes the issue following the footsteps of what was done for `Select`: including the `AttachedToken`s for the keywords in the AST node and using them in the `Spanned` impl.

Should I make a subticket for this? @alamb

agis:
https://github.com/apache/datafusion-sqlparser-rs/issues/1858 is also related to this

alamb:
> Should I make a subticket for this? @alamb

Sure! I always think tickets help the process (and you never know who might pick them up!)

Issue #2050, opened 2025-09-29
Inconsistent spans for function calls


ramnes:
When using functions spans, it appears that sometimes the closing parenthesis is included, sometimes not.

Consider the following test:

```rust
    #[test]
    fn test_function_span() {
        let sql = "SELECT database(), left(user(),instr(concat(user(),'@'),'@')-1);";
        let r = Parser::parse_sql(&crate::dialect::MySqlDialect {}, sql).unwrap();

        let query = match &r[0] {
            crate::ast::Statement::Query(q) => q,
            _ => panic!("Expected query"),
        };

        let select = match query.body.as_ref() {
            crate::ast::SetExpr::Select(s) => s,
            _ => panic!("Expected select"),
        };

        let database_func = match &select.projection[0] {
            crate::ast::SelectItem::UnnamedExpr(crate::ast::Expr::Function(func)) => func,
            _ => panic!("Expected function expression"),
        };
        let span = database_func.span();
        assert_eq!(span.start, (1, 8).into());
        assert_eq!(span.end, (1, 17).into()); // fails here

        let left_func = match &select.projection[1] {
            crate::ast::SelectItem::UnnamedExpr(crate::ast::Expr::Function(func)) => func,
            _ => panic!("Expected function expression"),
        };
        let span = left_func.span();
        assert_eq!(span.start, (1, 20).into());
        assert_eq!(span.end, (1, 63).into());
    }
```

In both cases, we're testing if the closing parenthesis is included in the span. The first span test for `database_func` fails:

```diff
Diff < left / right > :
-Location(1,16)
+Location(1,17)
```

If I comment that part, the second pair of span assertions passes correctly.

In other words, it seems that we only include closing parentheses for function calls with arguments, but not if there's no argument.

Related: #1548 #1563 #1676

xitep:
i have the same issue for `insert into table values (...)`; the trouble appears to be that a `Values` expression tracks only its nested children (ie. the value expressions) from which it then derives its whole span, and this omits the closing paren (and/or any intermediate whitespace.)

PR #2090, opened 2025-11-08
Include DML keyword in statement span

xitep:
* Includes the "update" keyword itself in the corresponding statement, making the statement's span cover its parsed source code
* Motivation: party a (SQL) re-formatter and party the need to extract possible comments preceding the statements (and/or right after the keyword)

* Related: #1548

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -39,8 +39,8 @@ use sqlparser::dialect::{
 };
 use sqlparser::keywords::{Keyword, ALL_KEYWORDS};
 use sqlparser::parser::{Parser, ParserError, ParserOptions};
-use sqlparser::tokenizer::{Location, Span};
 use sqlparser::tokenizer::Tokenizer;
+use sqlparser::tokenizer::{Location, Span};
 use test_utils::{
     all_dialects, all_dialects_where, all_dialects_with_options, alter_table_op, assert_eq_vec,
     call, expr_from_projection, join, number, only, table, table_alias, table_from_name,
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -26,8 +26,8 @@ use sqlparser::ast::MysqlInsertPriority::{Delayed, HighPriority, LowPriority};
 use sqlparser::ast::*;
 use sqlparser::dialect::{GenericDialect, MySqlDialect};
 use sqlparser::parser::{ParserError, ParserOptions};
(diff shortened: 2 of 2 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -5140,6 +5140,10 @@ fn test_simple_postgres_insert_with_alias() {
     assert_eq!(
         statement,
         Statement::Insert(Insert {
+            insert_token: AttachedToken(TokenWithSpan {
+                token: Token::make_keyword("INSERT"),
+                span: Span::new((1, 1).into(), (1, 7).into()),
+            }),
```
```suggestion
            insert_token: AttachedToken::empty(),
```
I think for these we can use an empty token - since token comparisons are ignored in tests. Same for the other files

src/ast/spans.rs
```diff
@@ -2535,4 +2542,22 @@ ALTER TABLE users
         assert_eq!(stmt_span.start, (2, 13).into());
         assert_eq!(stmt_span.end, (4, 11).into());
     }
+
+    #[test]
+    fn test_update_statement_span() {
```
iffyio, line 2552: Can we add similar test for `DELETE`, `INSERT`, `REPLACE` statements?

tests/sqlparser_postgres.rs
xitep: https://docs.rs/sqlparser/0.59.0/src/sqlparser/ast/helpers/attached_token.rs.html#100 and https://docs.rs/sqlparser/0.59.0/src/sqlparser/ast/helpers/attached_token.rs.html#106 ... ah, alright!

src/ast/spans.rs
xitep: i fear i'm affected by #2050 wrt to asserting the end of the insert/replace statements. would you like me to look at a solution for that? (most likely i'd propose [`Values`](https://docs.rs/sqlparser/latest/sqlparser/ast/struct.Values.html) to receive a `span: Span` attribute.) on the other hand, this feels rather ad-hoc and not conceptually thought through :/

iffyio: Oh I imagine for the scope of this PR, we can either assert the incomplete span as-is in the insert statement test, or we could change the test sql to e.g. `INSERT INTO foo SELECT ...` given the select query has implemented the Spanned trait?

xitep: oh, yeah ... a good workaround! thanks :+1:

xitep:
thanks @iffyio

PR #2091, opened 2025-11-08
fix: parse error on unnamed arg with default syntax

r1b:
Using the `DEFAULT` variant of the `CREATE FUNCTION` default argument syntax causes a parse error with unnamed arguments. The proposed fix skips the logic to disambiguate `argname argtype` from `argtype` when the `DEFAULT` keyword is detected. IIUC, this is safe because `argname` and `argtype` can only take on the name `DEFAULT` when quoted.

e.g, in postgres this is allowed:

```sql
CREATE DOMAIN "DEFAULT" AS int;
CREATE FUNCTION add(INTEGER, "DEFAULT" "DEFAULT" DEFAULT 1) RETURNS INTEGER LANGUAGE SQL AS 'select $1 + $2;';
```

But "DEFAULT" (the argname) and "DEFAULT" (the argtype) cannot appear unquoted

Ref: https://github.com/apache/datafusion/pull/18450#discussion_r2501243160

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5524,9 +5524,7 @@ impl<'a> Parser<'a> {
         let data_type_idx = self.get_current_index();
 
         // DEFAULT will be parsed as `DataType::Custom`, which is undesirable in this context
-        fn parse_data_type_no_default(
-            parser: &mut Parser,
-        ) -> Result<DataType, ParserError> {
+        fn parse_data_type_no_default(parser: &mut Parser) -> Result<DataType, ParserError> {
             if parser.peek_keyword(Keyword::DEFAULT) {
                 // This dummy error is ignored in `maybe_parse`
                 parser_err!(


r1b:
Please let me know if there is a better place for a more general fix. It seems like adding `DEFAULT` to `RESERVED_FOR_IDENTIFIER` might ~do the trick~ be necessary as well, but not sure if this is appropriate in all supported dialects.

src/parser/mod.rs
```diff
@@ -5521,16 +5521,19 @@ impl<'a> Parser<'a> {
...
+        if !self.peek_keyword(Keyword::DEFAULT) {
+            if let Some(next_data_type) = self.maybe_parse(|parser| parser.parse_data_type())? {
+                let token = self.token_at(data_type_idx);
 
-            // We ensure that the token is a `Word` token, and not other special tokens.
-            if !matches!(token.token, Token::Word(_)) {
-                return self.expected("a name or type", token.clone());
-            }
+                // We ensure that the token is a `Word` token, and not other special tokens.
+                if !matches!(token.token, Token::Word(_)) {
```
```suggestion
                if self.peek_keyword(DEFAULT) || !matches!(token.token, Token::Word(_)) {
```
thinking we can simplify the introduced if statement by doing the check here instead?

tests/sqlparser_postgres.rs
```diff
@@ -4476,6 +4476,46 @@ fn parse_create_function_detailed() {
...
+            function_body: Some(CreateFunctionBody::AsBeforeOptions(Expr::Value(
+                (Value::SingleQuotedString("select $1 + $2;".into())).with_empty_span()
+            ))),
+            if_not_exists: false,
+            using: None,
+            determinism_specifier: None,
+            options: None,
+            remote_connection: None,
+        })
+    );
```
iffyio, line 4481:
```suggestion
pg_and_generic().verified_stmt(r#"CREATE OR REPLACE FUNCTION return_table(i INTEGER) RETURNS TABLE(id UUID, is_active BOOLEAN) LANGUAGE plpgsql AS $$ BEGIN RETURN QUERY SELECT NULL::UUID, NULL::BOOLEAN; END; $$"#);
pg_and_generic().one_statement_parses_to(
    "CREATE FUNCTION add(INTEGER, INTEGER DEFAULT 1) RETURNS INTEGER AS 'select $1 + $2;'",
    "CREATE FUNCTION add(INTEGER, INTEGER = 1) RETURNS INTEGER AS 'select $1 + $2;'"
);
```
Can we do something like this instead, merge with existing test, and no need to assert the full AST in this scenario

src/parser/mod.rs
r1b: Won't we return a `Result<_, ParserError>` from `parse_function_arg` in this case? I was thinking we do still need to continue on to parse the `default_expr`. I'll take a closer look.

I agree that the nesting is not ideal in any case, maybe another option would be to push this check into `maybe_parse` above?

iffyio: Oh yeah indeed, I misread that block as happening inside of the maybe_parse. Yes the intent would be to ideally do the check in the maybe_parse I think

tests/sqlparser_postgres.rs
r1b: thanks, makes sense - updated in https://github.com/apache/datafusion-sqlparser-rs/pull/2091/commits/624c0f7149b3f2cebca884f43772903085644db8

src/parser/mod.rs
r1b: I took a stab at this in https://github.com/apache/datafusion-sqlparser-rs/pull/2091/commits/3fc3094d05b8a3e46cdc37438ae62436a014ff61

src/parser/mod.rs
```diff
@@ -5522,7 +5522,23 @@ impl<'a> Parser<'a> {
         // peek the next token, which if it is another type keyword, then the
         // first token is a name and not a type in itself.
         let data_type_idx = self.get_current_index();
-        if let Some(next_data_type) = self.maybe_parse(|parser| parser.parse_data_type())? {
+
+        // FIXME: DEFAULT will be parsed as `DataType::Custom`, which is undesirable in this context
```
iffyio: hmm how come the comment tagged with a 'FIXME', there's no action required right or?

r1b: Ah yeah I can remove this - no action required here

iffyio:
@r1b could you take a look at the failure on the branch?

r1b:
@iffyio done

iffyio (approved):
LGTM! Thanks @r1b!

Issue #2092, opened 2025-11-09
`CREATE TYPE` without `AS` fails parsing.


LucaCappelletti94:
At the time of writing, the following [valid](https://www.postgresql.org/docs/current/sql-createtype.html) syntax fails:

```sql
CREATE TYPE CAS;
```

```rust
let create_type_without_as = "CREATE TYPE CAS;";
    match verified_stmt(create_type_without_as) {
        Statement::CreateType {
            name,
            representation,
        } => {
            assert_eq_vec(&["CAS"], &name.0);
        }
        _ => unreachable!(),
    }
```

```
CREATE TYPE CAS;: ParserError("Expected: AS, found: ;")
```

I will make a PR shortly to fix this. I am starting to work more with [`pgrx`](https://github.com/pgcentralfoundation/pgrx) type definitions, so I may find myself subsequently extending the `CREATE TYPE` syntax a few times.

Best,
Luca

PR #2094, opened 2025-11-09
Complete PostgreSQL `CREATE TYPE` Support

LucaCappelletti94:
Adds full parsing for all PostgreSQL `CREATE TYPE` statement variants. Previously only composite and enum types were supported.

## What's New

**Range Types** - Parse range type definitions with 6 configuration options

```sql
CREATE TYPE int4range AS RANGE (SUBTYPE = int4, CANONICAL = fn);
```

**Base Types** - Parse low-level type definitions with 19 options (INPUT/OUTPUT functions, ALIGNMENT, STORAGE, etc.)

```sql
CREATE TYPE complex (INPUT = in_fn, OUTPUT = out_fn, INTERNALLENGTH = 16);
```

**Simple Declarations** - Parse type shells without representation

```sql
CREATE TYPE typename;
```

The AST extends `UserDefinedTypeRepresentation` with three new variants (`Range`, `SqlDefinition`, `None`) and adds five supporting enums (`UserDefinedTypeRangeOption`, `UserDefinedTypeSqlDefinitionOption`, `Alignment`, `UserDefinedTypeStorage`, `UserDefinedTypeInternalLength`) with comprehensive rustdoc. The parser refactors `parse_create_type()` into modular functions (`parse_create_type_composite()`, `parse_create_type_range()`, and base type option parsers) that handle all PostgreSQL type configuration syntax. Tests provide comprehensive coverage with explicit AST verification for all variants and edge cases. Added 17 PostgreSQL-specific keywords, necessary to parse the keywords used in the type options (ALIGNMENT, CANONICAL, INTERNALLENGTH, PASSEDBYVALUE, RECEIVE, SEND, STORAGE, SUBTYPE, TYPMOD_IN, etc.).

Closes issue #2092


src/ast/ddl.rs
```diff
@@ -2002,22 +2002,45 @@ impl fmt::Display for DropBehavior {
...
+    /// Range type: `CREATE TYPE name AS RANGE (options)`
+    Range {
+        options: Vec<UserDefinedTypeRangeOption>,
+    },
+    /// Base type (SQL definition): `CREATE TYPE name (options)`
+    ///
+    /// Note the lack of `AS` keyword
+    SqlDefinition {
+        options: Vec<UserDefinedTypeSqlDefinitionOption>,
+    },
```
iffyio, line 2047: Could we include a link to the postgres docs for the new variants?

src/ast/ddl.rs
```diff
@@ -2002,22 +2002,45 @@ impl fmt::Display for DropBehavior {
...
+    /// When the representation of the type is not specified.
+    /// This is used in `CREATE TYPE <name>;` statements.
+    None,
```
iffyio: Instead of having a special `None` variant, ideally the usage becomes an optional field, would that work here? i.e `representation: Option<UserDefinedTypeRepresentation>`

src/parser/mod.rs
```diff
@@ -17331,20 +17331,60 @@ impl<'a> Parser<'a> {
...
+            self.expected("ENUM, RANGE, or '(' after AS", self.peek_token())
         }
+    }
 
-        let mut attributes = vec![];
-        if !self.consume_token(&Token::LParen) || self.consume_token(&Token::RParen) {
+    /// Parse remainder of `CREATE TYPE AS (attributes)` statement (composite type)
+    ///
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-createtype.html)
+    pub fn parse_create_type_composite(
```
```suggestion
    fn parse_create_type_composite(
```

src/parser/mod.rs
```diff
@@ -17387,6 +17425,254 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse remainder of `CREATE TYPE AS RANGE` statement
+    ///
+    /// See [PostgreSQL](https://www.postgresql.org/docs/current/sql-createtype.html)
+    pub fn parse_create_type_range(&mut self, name: ObjectName) -> Result<Statement, ParserError> {
```
```suggestion
    fn parse_create_type_range(&mut self, name: ObjectName) -> Result<Statement, ParserError> {
```

src/parser/mod.rs
```diff
@@ -17387,6 +17425,254 @@ impl<'a> Parser<'a> {
...
+            name,
+            representation: UserDefinedTypeRepresentation::Range { options },
+        })
+    }
+
+    /// Parse a single range option for CREATE TYPE AS RANGE
+    fn parse_range_option(&mut self) -> Result<UserDefinedTypeRangeOption, ParserError> {
+        let keyword = self.parse_one_of_keywords(&[
+            Keyword::SUBTYPE,
+            Keyword::SUBTYPE_OPCLASS,
```
iffyio, line 17473: can we add tests for `SUBTYPE_OPCLASS`,`SUBTYPE_DIFF` and any other variants that are missing coverage?

src/parser/mod.rs
```diff
@@ -17387,6 +17425,254 @@ impl<'a> Parser<'a> {
...
+        let options = self.parse_comma_separated0(|p| p.parse_range_option(), Token::RParen)?;
+        self.expect_token(&Token::RParen)?;
+
+        Ok(Statement::CreateType {
+            name,
+            representation: UserDefinedTypeRepresentation::Range { options },
+        })
+    }
+
+    /// Parse a single range option for CREATE TYPE AS RANGE
```
```suggestion
    /// Parse a single range option for a `CREATE TYPE AS RANGE` statement
```

iffyio (commented):
Thanks @LucaCappelletti94! Changes look good to me overall, left some comments

src/ast/ddl.rs
LucaCappelletti94: I wanted to, but they do not have a linkable subsection so the link would be to the same `CREATE TYPE` page. The sub-headers do not have the anchor link unfortunately.

src/ast/ddl.rs
LucaCappelletti94: Sure I can switch to that case - should we normalize other analogous enumerations which also have the `None` field, maybe in another PR?

src/parser/mod.rs
LucaCappelletti94: Oke

src/ast/ddl.rs
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: Tentatively done, I think all are covered now.

src/ast/ddl.rs
iffyio: ah I think in that case linking to the postgres create type page for each variant would suffice, most important that its clear where each variant comes from especially if the overlying statement/enum is shared between multiple dialects. Basically similar to what the `Enum { ... }` variant above documents?

LucaCappelletti94: Ok, doing that.

LucaCappelletti94: Added

iffyio (approved):

PR #2095, opened 2025-11-09
Support PostgreSQL C Functions with Multiple AS Parameters

### LINKED ISSUE #2093: Support PostgreSQL `CREATE FUNCTION ... AS 'lib', 'symbol'` syntax
`sqlparser-rs` fails to parse valid PostgreSQL `CREATE FUNCTION` statements that specify both a library path and symbol name, e.g.:

```sql
CREATE FUNCTION cas_in(input cstring)
RETURNS cas
IMMUTABLE PARALLEL SAFE
LANGUAGE c
AS 'MODULE_PATHNAME', 'cas_in_wrapper';
```

```
ParserError("Expected: end of statement, found: ,")
```

I will make a PR adding support for this later after I am finished with #2092 

Luca

LucaCappelletti94:
Enables parsing of PostgreSQL C language functions that specify both object file and link symbol in the `AS` clause.

## Example

```sql
CREATE FUNCTION cas_in(input cstring) RETURNS cas
LANGUAGE c
AS 'MODULE_PATHNAME', 'cas_in_wrapper';
```

Fixes #2093


src/parser/mod.rs
```diff
@@ -10224,17 +10224,32 @@ impl<'a> Parser<'a> {
...
+        };
+
+        let first_expr = parse_string_expr(self)?;
+
+        // Check if there's a comma, indicating multiple strings (e.g., AS 'obj_file', 'link_symbol')
+        // This is used for C language functions: AS 'MODULE_PATHNAME', 'link_symbol'
+        if self.consume_token(&Token::Comma) {
+            let mut exprs = vec![first_expr];
+            exprs.extend(self.parse_comma_separated(parse_string_expr)?);
+            Ok(Expr::Tuple(exprs))
```
```suggestion
            Ok(Expr::Tuple(vec![
                first_expr,
                self.parse_comma_separated(parse_string_expr)?
            ]))
```
looks like this can be simplified?

iffyio: oh, does wrapping this in a Tuple not include a parenthesis `(first_expr, second_expr)` when displaying the AST? wondering if/how we worked around that somehow in this case

LucaCappelletti94: Ah simply the following happens:

```rust
if let Some(CreateFunctionBody::AsBeforeOptions(function_body)) = &self.function_body {
        write!(f, " AS ")?;
        // Special handling for tuple expressions to format without parentheses
        // PostgreSQL C functions use: AS 'obj_file', 'link_symbol'
        // rather than: AS ('obj_file', 'link_symbol')
        if let Expr::Tuple(exprs) = function_body {
            write!(f, "{}", display_comma_separated(exprs))?;
        } else {
            write!(f, "{function_body}")?;
        }
    }
```

iffyio: right I see, is there a way we can maybe introduce a new field to store the link_symbol? I think having this special case would potentially break in a surprising manner if a tuple was stored in `CreateFunctionBody::AsBeforeOptions` in another context

LucaCappelletti94: You mean a `link_symbol` field in the `AsBeforeOptions` struct?

iffyio: Yeah, maybe something like `CreateFunctionBody::AsBeforeOptions{ body: Expr, link_symbol: Option<String> }`

src/ast/mod.rs
```diff
@@ -9099,8 +9099,19 @@ pub enum CreateFunctionBody {
...
+    /// AS 'MODULE_PATHNAME', 'cas_in_wrapper'
+    /// ```
+    ///
     /// [BigQuery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_11
-    AsBeforeOptions(Expr),
+    AsBeforeOptions {
+        /// The primary expression (object file path for C functions, or regular expression)
+        body: Expr,
+        /// Optional link symbol for C language functions
+        link_symbol: Option<Expr>,
```
```suggestion
   /// [BigQuery]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_11
    /// [PostgreSQL]: https://www.postgresql.org/docs/current/sql-createfunction.html
    AsBeforeOptions {
        /// The primary expression.
        body: Expr,
        /// Link symbol if the primary expression contains the name of shared library file.
        ///
        /// Example:
        /// ```sql
        /// CREATE FUNCTION cas_in(input cstring) RETURNS cas
        /// AS 'MODULE_PATHNAME', 'cas_in_wrapper'
        /// ```
        /// [PostgreSQL]: https://www.postgresql.org/docs/current/sql-createfunction.html
        link_symbol: Option<Expr>,
```

src/parser/mod.rs
```diff
@@ -10521,19 +10522,33 @@ impl<'a> Parser<'a> {
...
+                )),
             }
-            _ => Ok(Expr::Value(
-                Value::SingleQuotedString(self.parse_literal_string()?).with_span(span),
-            )),
-        }
+        };
+
+        // Check if there's a comma, indicating multiple strings (e.g., AS 'obj_file', 'link_symbol')
+        // This is used for C language functions: AS 'MODULE_PATHNAME', 'link_symbol'
```
```suggestion
```
I think we can remove this comment since its mostly describing what the code is doing, (the link_symbol feature is already described in the enum variant

src/parser/mod.rs
```diff
@@ -10521,19 +10522,33 @@ impl<'a> Parser<'a> {
...
-    fn parse_create_function_body_string(&mut self) -> Result<Expr, ParserError> {
-        let peek_token = self.peek_token();
-        let span = peek_token.span;
-        match peek_token.token {
-            Token::DollarQuotedString(s) if dialect_of!(self is PostgreSqlDialect | GenericDialect) =>
-            {
-                self.next_token();
-                Ok(Expr::Value(Value::DollarQuotedString(s).with_span(span)))
+    fn parse_create_function_body_string(&mut self) -> Result<CreateFunctionBody, ParserError> {
+        // Helper closure to parse a single string value (quoted or dollar-quoted)
```
```suggestion
```

src/ast/mod.rs
LucaCappelletti94: Done, for some reason I could not apply the change from here.

iffyio (approved):

PR #2096, opened 2025-11-09
Add PostgreSQL Operator DDL Support

LucaCappelletti94:
Adds support for `CREATE OPERATOR`, `CREATE OPERATOR FAMILY`, and `CREATE OPERATOR CLASS` statements.

## Examples

```sql
-- CREATE OPERATOR with all parameters
CREATE OPERATOR < (
    PROCEDURE = cas_lt,
    LEFTARG = CAS,
    RIGHTARG = CAS,
    COMMUTATOR = >,
    NEGATOR = >=,
    RESTRICT = scalarltsel,
    JOIN = scalarltjoinsel
);

-- CREATE OPERATOR FAMILY
CREATE OPERATOR FAMILY integer_ops USING btree;

-- CREATE OPERATOR CLASS
CREATE OPERATOR CLASS int4_ops
    DEFAULT FOR TYPE int4 USING btree FAMILY integer_ops AS
    OPERATOR 1 <,
    OPERATOR 2 <=,
    FUNCTION 1 btint4cmp(int4, int4),
    STORAGE int4;
```


src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    /// Helper function to parse an operator name (which can contain special characters)
+    /// Operator names can be schema-qualified (e.g., schema.operator)
+    fn parse_operator_name(&mut self) -> Result<ObjectName, ParserError> {
```
iffyio, line 6453: does this differ from `parse_object_name`?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+            name_parts.push(part);
+
+            if !self.consume_token(&Token::Period) {
+                break;
+            }
+        }
+        Ok(ObjectName(name_parts))
+    }
+
+    /// Parse a CREATE OPERATOR statement
```
```suggestion
    /// Parse a `CREATE OPERATOR` statement
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                    self.expect_token(&Token::Eq)?;
+                    let func_name = self.parse_object_name(false)?;
+                    restrict = Some(func_name);
+                }
+                Keyword::JOIN => {
+                    self.expect_token(&Token::Eq)?;
+                    let func_name = self.parse_object_name(false)?;
+                    join = Some(func_name);
+                }
+                _ => unreachable!("unexpected keyword in CREATE OPERATOR"),
```
iffyio: Can we return an error here instead of crashing in the case of a bug?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+            commutator,
+            negator,
+            restrict,
+            join,
+            hashes,
+            merges,
+        }))
+    }
+
+    /// Parse a CREATE OPERATOR FAMILY statement
```
```suggestion
    /// Parse a `CREATE OPERATOR FAMILY` statement
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+        let default = self.parse_keyword(Keyword::DEFAULT);
+        self.expect_keywords(&[Keyword::FOR, Keyword::TYPE])?;
+        let for_type = self.parse_data_type()?;
+        self.expect_keyword(Keyword::USING)?;
+        let using = self.parse_identifier()?;
+
+        let family = if self.parse_keyword(Keyword::FAMILY) {
+            Some(self.parse_object_name(false)?)
+        } else {
+            None
```
iffyio, line 6610: Can we add a test case for a statement without a family?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                let strategy_number = self.parse_literal_uint()? as u32;
+                let operator_name = self.parse_operator_name()?;
+
+                // Optional operator argument types
+                let op_types = if self.consume_token(&Token::LParen) {
+                    let left = self.parse_data_type()?;
+                    self.expect_token(&Token::Comma)?;
+                    let right = self.parse_data_type()?;
+                    self.expect_token(&Token::RParen)?;
+                    Some(OperatorArgTypes { left, right })
```
iffyio, line 6627: Are we missing test coverage for this block?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                // Optional purpose
+                let purpose = if self.parse_keyword(Keyword::FOR) {
+                    if self.parse_keyword(Keyword::SEARCH) {
+                        Some(OperatorPurpose::ForSearch)
+                    } else if self.parse_keywords(&[Keyword::ORDER, Keyword::BY]) {
+                        let sort_family = self.parse_object_name(false)?;
+                        Some(OperatorPurpose::ForOrderBy { sort_family })
+                    } else {
+                        return self.expected("SEARCH or ORDER BY after FOR", self.peek_token());
+                    }
```
iffyio, line 6641: Are we missing test coverage for this block?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                    if self.consume_token(&Token::LParen) && self.peek_token() != Token::RParen {
+                        let mut types = vec![];
+                        loop {
+                            types.push(self.parse_data_type()?);
+                            if !self.consume_token(&Token::Comma) {
+                                break;
+                            }
+                        }
+                        self.expect_token(&Token::RParen)?;
+                        Some(types)
```
iffyio, line 6666: can we add test coverage for this block?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                    } else if self.consume_token(&Token::LParen) {
+                        self.expect_token(&Token::RParen)?;
+                        Some(vec![])
```
iffyio, line 6669: similarly, can we add test coverage for this block?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                    operator_name,
+                    op_types,
+                    purpose,
+                });
+            } else if self.parse_keyword(Keyword::FUNCTION) {
+                let support_number = self.parse_literal_uint()? as u32;
+
+                // Optional operator types
+                let op_types =
+                    if self.consume_token(&Token::LParen) && self.peek_token() != Token::RParen {
```
iffyio, line 6657: I think there might be a bug in the conditions due to the use of `consume_token()` here - note that the `else if` branch is also trying to consume `LParen`, so that unless `(()` is valid syntax, the else block looks ineffectual. We probably want to instead do something like
```rust
if self.parse_keywords([LParen, RParen]) { ... }
else if self.parse_keyword(LParen) { ... }
else { None }
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                        }
+                        types.push(self.parse_data_type()?);
+                        if !self.consume_token(&Token::Comma) {
+                            break;
+                        }
+                    }
+                    self.expect_token(&Token::RParen)?;
+                    types
+                } else {
+                    vec![]
```
iffyio, line 6691: Can we add test coverage for this block?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+
+                items.push(OperatorClassItem::Function {
+                    support_number,
+                    op_types,
+                    function_name,
+                    argument_types,
+                });
+            } else if self.parse_keyword(Keyword::STORAGE) {
+                let storage_type = self.parse_data_type()?;
+                items.push(OperatorClassItem::Storage { storage_type });
```
iffyio, line 6702: do we have tests for this part of the syntax?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+    }
+
+    /// Parse a CREATE OPERATOR statement
+    ///
+    /// [PostgreSQL Documentation](https://www.postgresql.org/docs/current/sql-createoperator.html)
+    pub fn parse_create_operator(&mut self) -> Result<Statement, ParserError> {
+        // Parse the operator name (can be schema-qualified)
+        // Operators can contain special characters like +, -, *, /, <, >, =, ~, !, @, #, %, ^, &, |, `, ?
+        // See https://www.postgresql.org/docs/current/sql-createoperator.html
+        let name = self.parse_operator_name()?;
```
iffyio, line 6470: can we add test cases that use qualified operator names where applicable?

src/parser/mod.rs
```diff
@@ -6422,6 +6431,291 @@ impl<'a> Parser<'a> {
...
+                Keyword::NEGATOR,
+                Keyword::RESTRICT,
+                Keyword::JOIN,
+                Keyword::HASHES,
+                Keyword::MERGES,
+            ])?;
+
+            // Check if this is a flag (HASHES or MERGES) - no '=' expected
+            match keyword {
+                Keyword::HASHES => {
```
iffyio: I think since we're only expecting one of each clause, we might need to add guards to the match clauses - e.g. `Keyword::Hashes if !hashes => {...} Keyword::LEFTARG if left_arg.is_none() => {...}`

it would be good to add negative tests that exercise such behavior.

iffyio: Can we add test coverage for hashes, merges etc? we can do a pass through the clauses to ensure that they're properly covered

iffyio (commented):
Left some comments, a bit worried that its a lot of code and seems to be lacking test coverage, I tried to point out a few of them but would be good that we do a full pass to ensure code coverage in the PR

src/parser/mod.rs
LucaCappelletti94: Basically, the other function does not support correctly the cases where the operator name is `+` or `schema_name.@`  etc etc

src/parser/mod.rs
LucaCappelletti94: Done, I think I might have tab-autocompleted with copilot this one. Sorry.

src/parser/mod.rs
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: Made sure we do cover this case.

src/parser/mod.rs
LucaCappelletti94: Covered, I think

src/parser/mod.rs

src/parser/mod.rs
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: And done

src/parser/mod.rs
LucaCappelletti94: Done

src/parser/mod.rs
```diff
@@ -6422,6 +6431,314 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    /// Parse an operator name, which can contain special characters like +, -, <, >, =
+    /// that are tokenized as operator tokens rather than identifiers.
+    /// This is used for PostgreSQL CREATE OPERATOR statements.
+    ///
+    /// Examples: `+`, `myschema.+`, `pg_catalog.<=`
+    pub fn parse_operator_name(&mut self) -> Result<ObjectName, ParserError> {
```
```suggestion
    fn parse_operator_name(&mut self) -> Result<ObjectName, ParserError> {
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,314 @@ impl<'a> Parser<'a> {
...
+
+    /// Parse a `CREATE OPERATOR` statement
+    ///
+    /// [PostgreSQL Documentation](https://www.postgresql.org/docs/current/sql-createoperator.html)
+    pub fn parse_create_operator(&mut self) -> Result<Statement, ParserError> {
+        macro_rules! dup_err {
+            ($name:expr) => {
+                ParserError::ParserError(format!("Duplicate {} clause in CREATE OPERATOR", $name))
+            };
+        }
```
iffyio: we can convert this to either a function or inline the usages (we try to avoid macros where possible since they're not maintenance friendly).

But actually I think a simpler way of writing the loop logic would be to rely on rust's match conditions
i.e. `Keyword::HASHES if !hashes => { hashes = true }`

Then any duplicates are implicitly handled once by the match all clause
`_ => { return Err('unexpected keyword {} ...') }`

src/parser/mod.rs
```diff
@@ -6422,6 +6431,314 @@ impl<'a> Parser<'a> {
...
+        self.expect_keyword(Keyword::USING)?;
+        let using = self.parse_identifier()?;
+
+        Ok(Statement::CreateOperatorFamily(CreateOperatorFamily {
+            name,
+            using,
+        }))
+    }
+
+    /// Parse a CREATE OPERATOR CLASS statement
```
```suggestion
    /// Parse a [Statement::CreateOperatorClass]
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,314 @@ impl<'a> Parser<'a> {
...
+            commutator,
+            negator,
+            restrict,
+            join,
+            hashes,
+            merges,
+        }))
+    }
+
+    /// Parse a `CREATE OPERATOR FAMILY` statement
```
```suggestion
    /// Parse a [Statement::CreateOperatorFamily]
```

src/parser/mod.rs
```diff
@@ -6422,6 +6431,314 @@ impl<'a> Parser<'a> {
...
+                self.next_token().to_string(),
+            )));
+            if !self.consume_token(&Token::Period) {
+                break;
+            }
+        }
+        Ok(ObjectName(parts))
+    }
+
+    /// Parse a `CREATE OPERATOR` statement
```
```suggestion
    /// Parse a [Statement::CreateOperator]
```
thinking this style would be preferrable to link to what the function returns

iffyio (approved):
Thanks @LucaCappelletti94!

PR #2098, opened 2025-11-10
Immutable parser (Tokenize borrow strings cont)

eyalleshem:
refactor: Make Parser methods immutable using interior mutability

Changed all parsing methods to take `&self` instead of `&mut self`. 
Mutable parser state (token index and parser state) now uses `Cell` 
for interior mutability.

This refactoring is preparation for the borrowed tokenizer work. When 
holding borrowed tokens from the parser (with lifetime tied to `&self`), 
we cannot call methods requiring `&mut self` due to Rust's borrowing 
rules. Using interior mutability resolves this conflict by allowing 
state mutations through shared references.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -43,6 +43,7 @@ use crate::ast::*;
 use crate::dialect::*;
 use crate::keywords::{Keyword, ALL_KEYWORDS};
 use crate::tokenizer::*;
+use core::cell::Cell;
 use sqlparser::parser::ParserState::ColumnDefinition;
 
 mod alter;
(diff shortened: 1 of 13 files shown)


eyalleshem:
@iffyio @alamb - Heads up that I've moved this PR to ready for review! Would appreciate your feedback when you get a chance.

src/parser/mod.rs
```diff
@@ -331,9 +332,9 @@ pub struct Parser<'a> {
     /// The tokens
     tokens: Vec<TokenWithSpan>,
     /// The index of the first unprocessed token in [`Parser::tokens`].
-    index: usize,
+    index: Cell<usize>,
```
alamb, line 335: This is a clever way to go about reducing the need for mutable borrows. Thank you @eyalleshem 

One thing that might be worth doing (perhaps as a follow on PR) is documenting how this works (notable that Parser has interior mutability so functions like `next_token()` actually do change state, even though they are immutable

PR #2100, opened 2025-11-18
impl `Spanned` for MERGE statements

xitep:
* Provides the span for MERGE statements
* Part of #1548

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -17239,10 +17239,12 @@ impl<'a> Parser<'a> {
                         clause_kind,
                         MergeClauseKind::NotMatched | MergeClauseKind::NotMatchedByTarget
                     ) {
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -17225,66 +17229,84 @@ impl<'a> Parser<'a> {
 
             self.expect_keyword_is(Keyword::THEN)?;
 
+            macro_rules! not_allowed {
```
iffyio: maybe we can keep this inlined as before, alternative would be to pull it out to a function but that might not be worth it. But in generally we'd ideally avoid custom macros where possible since those are harder to maintain

xitep: ok, will do.

iffyio (approved):
LGTM! Thanks @xitep!

PR #2106, opened 2025-11-26
Added support for `DROP OPERATOR FAMILY`

LucaCappelletti94:
As per title, this PR adds support for the [`DROP OPERATOR FAMILY`](https://www.postgresql.org/docs/17//sql-dropopfamily.html) syntax.


iffyio (approved):

PR #2107, opened 2025-11-26
Extract source comments

### LINKED ISSUE #2065: Limited Inline Comment Support in AST

xitep:
* Should (partially) resolve #2065

### DIFF
--- a/src/ast/comments.rs
+++ b/src/ast/comments.rs
@@ -29,14 +29,19 @@ use crate::tokenizer::{Location, Span};
 pub struct Comments(Vec<CommentWithSpan>);
 
 impl Comments {
-    pub(crate) fn push(&mut self, comment: CommentWithSpan) {
-        debug_assert!(
-            self.0
-                .last()
-                .map(|last| last.span < comment.span)
-                .unwrap_or(true)
-        );
-        self.0.push(comment);
+    /// Accepts `comment` if its the first or is located strictly after the
+    /// last accepted comment.  In other words, this method will skip the
+    /// comment if its comming out of order (as encountered in the parsed
+    /// source code.)
+    pub(crate) fn offer(&mut self, comment: CommentWithSpan) {
+        if self
+            .0
+            .last()
+            .map(|last| last.span < comment.span)
+            .unwrap_or(true)
+        {
+            self.0.push(comment);
+        }
     }
 
     /// Finds comments starting within the given location range. The order of
@@ -221,25 +226,25 @@ mod tests {
             // */
             // ```
             let mut c = Comments(Vec::new());
-            c.push(CommentWithSpan {
+            c.offer(CommentWithSpan {
                 comment: Comment::SingleLine {
                     content: " abc".into(),
                     prefix: "--".into(),
                 },
                 span: Span::new((1, 1).into(), (1, 7).into()),
             });
-            c.push(CommentWithSpan {
+            c.offer(CommentWithSpan {
                 comment: Comment::MultiLine(" hello ".into()),
                 span: Span::new((2, 3).into(), (2, 14).into()),
             });
-            c.push(CommentWithSpan {
+            c.offer(CommentWithSpan {
                 comment: Comment::SingleLine {
                     content: ", world".into(),
                     prefix: "--".into(),
                 },
                 span: Span::new((2, 14).into(), (2, 21).into()),
             });
-            c.push(CommentWithSpan {
+            c.offer(CommentWithSpan {
                 comment: Comment::MultiLine(" def\n ghi\n jkl\n".into()),
                 span: Span::new((3, 3).into(), (7, 1).into()),
             });
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -551,7 +551,7 @@ impl<'a> Parser<'a> {
         for t in self.tokens.into_iter() {
             match t.token {
                 Token::Whitespace(Whitespace::SingleLineComment { comment, prefix }) => {
-                    comments.push(comments::CommentWithSpan {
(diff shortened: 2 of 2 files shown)


src/ast/comments.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// Finds comments starting within the given location range. The order of
+    /// iterator reflects the order of the comments as encountered in the parsed
+    /// source code.
+    pub fn find<R: RangeBounds<Location>>(&self, range: R) -> Iter<'_> {
+        let (start, end) = (
+            self.start_index(range.start_bound()),
+            self.end_index(range.end_bound()),
+        );
+        // ~ in case the user specified a rever range
+        Iter(if start <= end {
```
iffyio, line 97:
```suggestion
        Iter(if start <= end && end <= self.0.len() {
```
maybe this condition as well to ensure that the indexing is always safe?

src/ast/comments.rs
```diff
@@ -0,0 +1,280 @@
...
+
+use crate::tokenizer::{Location, Span};
+
+/// An opaque container for comments from a parse SQL source code.
+#[derive(Default, Debug)]
+pub struct Comments(Vec<CommentWithSpan>);
+
+impl Comments {
+    pub(crate) fn push(&mut self, comment: CommentWithSpan) {
+        debug_assert!(self
```
iffyio: Since we're requiring the underlying vec be sorted, can we instead skip pushing if we see that the invariant is going to be violated by the input?

src/ast/comments.rs
```diff
@@ -0,0 +1,280 @@
...
+            .0
+            .last()
+            .map(|last| last.span < comment.span)
+            .unwrap_or(true));
+        self.0.push(comment);
+    }
+
+    /// Finds comments starting within the given location range. The order of
+    /// iterator reflects the order of the comments as encountered in the parsed
+    /// source code.
```
iffyio, line 49: Would it be possible to provide an example in the doc comment here? Like a sample sql containing comments, and sample requests that illustrate what the function returns in certain scenarios. I think that would help folks figure out how to use the feature

src/ast/comments.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn find<R: RangeBounds<Location>>(&self, range: R) -> Iter<'_> {
+        let (start, end) = (
+            self.start_index(range.start_bound()),
+            self.end_index(range.end_bound()),
+        );
+        // ~ in case the user specified a rever range
```
iffyio:
```suggestion
        // in case the user specified a reverse range
```

src/ast/comments.rs
xitep: actually, it would be a programming error; the idea was that `start_index` and `end_index` return something in the range of `0..=self.0.len()`. i'll add appropriate debug assertions and document the helper methods.

PR #2108, opened 2025-11-29
Fixing location for extrenal tables

romanoff:
Location isn't required for external tables for hive. Example of query:
```
CREATE EXTERNAL TABLE my_table (
    c INT
)
```
Spec:
https://hive.apache.org/docs/latest/language/languagemanual-ddl/?utm_source=chatgpt.com#external-tables

It says: `The EXTERNAL keyword lets you create a table and provide a LOCATION so that Hive does not use a default location`. It's not very clear, but if location is not specified, it will provide default location.


tests/sqlparser_hive.rs
```diff
@@ -566,3 +566,8 @@ fn hive() -> TestedDialects {
 fn hive_and_generic() -> TestedDialects {
     TestedDialects::new(vec![Box::new(HiveDialect {}), Box::new(GenericDialect {})])
 }
+
+#[test]
+fn parse_create_external_table_without_location() {
+    hive().verified_stmt("CREATE EXTERNAL TABLE t (c INT)");
+}
```
iffyio: Can we add this entry to [this](https://github.com/romanoff/datafusion-sqlparser-rs/blob/37581a62d3b969ac320f1dfbe6563cf441f46f30/tests/sqlparser_hive.rs#L33) existing test case instead?

romanoff: @iffyio Updated. Thank you

PR #2109, opened 2025-11-29
Added support for `DROP OPERATOR CLASS` syntax

LucaCappelletti94:
As per title, this PR adds support for the [`DROP OPERATOR CLASS`](https://www.postgresql.org/docs/current/sql-dropopclass.html) syntax.


iffyio (approved):

PR #2110, opened 2025-11-30
Oracle dialect

xitep:
As discussed in #2101


iffyio:
@xitep could we split the PR into smaller more focused ones? its easier for me to make the time to review it that way. I think we can have this one just introduce the dialect (basically only implementing the required trait methods), then follow up PRs can introduce any precedence changes, and another for the string support

xitep:
sure, will do

PR #2112, opened 2025-12-01
Add lifetime to token

eyalleshem:
his PR adds a lifetime parameter to the Token enum to enable zero-copy tokenization in the future.

  Backward Compatibility:
  To maintain backward compatibility, we renamed Token to BorrowedToken<'a> and introduced Token as a type alias for BorrowedToken<'static>. This allows existing consumers of the library to continue using Token without needing to handle
  lifetimes throughout their code.

  New API:
  This PR also adds a tokenized_owned() method for use cases where consumers prefer to pay the cost of copying in exchange for owned tokens.

  Current State:
  This commit does not yet change the tokenizer's behavior—all string allocations remain in place. The goal of the following commits is to replace String with Cow<'a, str> in as many places as possible, leveraging the Borrowed variant to
  achieve zero-copy tokenization where feasible.


eyalleshem:
@iffyio @alamb - Marking this one ready for review too :)

alamb:
I'll try and review this one over the next few days

PR #2114, opened 2025-12-01
Added support for `ALTER OPERATOR` syntax

LucaCappelletti94:
As per title, this PR adds support for the [`ALTER OPERATOR` syntax](https://www.postgresql.org/docs/current/sql-alteroperator.html), and simplifies a bit the existing `CREATE OPERATOR` struct to reduce the code duplication between the two statement representations in the AST.


src/parser/mod.rs
```diff
@@ -9931,6 +9939,114 @@ impl<'a> Parser<'a> {
...
+                        let op_name = self.parse_operator_name()?;
+                        options.push(OperatorOption::Negator(op_name));
+                    }
+                    Keyword::HASHES => {
+                        options.push(OperatorOption::Hashes);
+                    }
+                    Keyword::MERGES => {
+                        options.push(OperatorOption::Merges);
+                    }
+                    _ => unreachable!(),
```
iffyio: Can we return an error here instead?

LucaCappelletti94: To future proof this, I have added:

```toml
[lints.clippy]
unreachable = "forbid"
```

To the `Cargo.toml`, so it will be clearer. There are a few `unreachable` in the lib, I will go through them and change them into errors.

LucaCappelletti94: Done, now unreachable patterns in the lib will trigger clippy. I have replaced all unreachable patterns with errors, or simplified them out when reasonable in their context.

iffyio: Awesome! Thank you!

iffyio (approved):

PR #2115, opened 2025-12-02
[Oracle] Lower StringConcat precedence

xitep:


tests/sqlparser_oracle.rs
```diff
@@ -0,0 +1,81 @@
...
+            *expr = Expr::Nested(Box::new(inner));
+        }
+    }
+    match item {
+        SelectItem::UnnamedExpr(expr) => nest(expr),
+        SelectItem::ExprWithAlias { expr, alias: _ } => nest(expr),
+        SelectItem::QualifiedWildcard(_, _) => {}
+        SelectItem::Wildcard(_) => {}
+    }
+}
```
iffyio: can we move these to the top of the file instead? folks tend to add new tests to the bottom of the file, so that these don't end up awkwardly in the middle over time.

tests/sqlparser_oracle.rs
```diff
@@ -0,0 +1,81 @@
...
+    let sql = "SELECT 3 + 5 || '.3' || 7 - 9 FROM dual";
+    let mut query = oracle_dialect().verified_query(sql);
+    nest_binary_ops(&mut query.body.as_select_mut().expect("not a SELECT").projection[0]);
+    assert_eq!(
+        &format!("{query}"),
+        "SELECT ((((3 + 5) || '.3') || 7) - 9) FROM dual"
+    );
+}
+
+fn oracle_dialect() -> TestedDialects {
```
```suggestion
fn oracle() -> TestedDialects {
```

tests/sqlparser_oracle.rs
```diff
@@ -0,0 +1,81 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#![warn(clippy::all)]
+//! Test SQL syntax, specific to [sqlparser::dialect::OracleDialect].
+
+extern crate core;
```
iffyio: what's this used for?

tests/sqlparser_oracle.rs
```diff
@@ -0,0 +1,81 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#![warn(clippy::all)]
+//! Test SQL syntax, specific to [sqlparser::dialect::OracleDialect].
```
iffyio, line 18:
```suggestion
```
is the clippy attribute necessary here?

src/ast/query.rs
```diff
@@ -174,6 +174,15 @@ impl SetExpr {
...
     }
+
+    /// If this `SetExpr` is a `SELECT`, returns a mutable [`Select`].
+    pub fn as_select_mut(&mut self) -> Option<&mut Select> {
+        if let Self::Select(select) = self {
+            Some(&mut **select)
+        } else {
+            None
+        }
+    }
```
```suggestion
```
I think we would want to avoid a public mut API that is only used for internal tests, if needed it would likely make more sense as a helper function in the test module

tests/sqlparser_oracle.rs
```diff
@@ -0,0 +1,81 @@
...
+        "SELECT (((3 / 5) || 'asdf') || (7 * 9)) FROM dual"
+    );
+}
+
+#[test]
+fn plusminus_have_same_precedence_as_strconcat() {
+    // ~ oracle: `+`, `-`, and `||` have the same precedence and parse from left-to-right
+    let sql = "SELECT 3 + 5 || '.3' || 7 - 9 FROM dual";
+    let mut query = oracle_dialect().verified_query(sql);
+    nest_binary_ops(&mut query.body.as_select_mut().expect("not a SELECT").projection[0]);
```
iffyio: I think instead of introducing the nest_binary_ops tfn hat mutates the ast, can we instead explicitly assert what we expect the ast to look like?

src/dialect/oracle.rs
```diff
@@ -75,6 +82,35 @@ impl Dialect for OracleDialect {
...
+            Precedence::Period => 100,
+            Precedence::DoubleColon => 50,
+            Precedence::AtTz => 41,
+            Precedence::MulDivModOp => 40,
+            Precedence::PlusMinus => 30,
+            Precedence::Xor => 24,
+            Precedence::Ampersand => 23,
+            Precedence::Caret => 22,
+            Precedence::Pipe => 21,
+            Precedence::Between | Precedence::Eq | Precedence::Like | Precedence::Is => 20,
```
iffyio: This looks like the only changes compared to the default dialect precedence, can we add tests demonstrating what they imply?

Also since we're manually overriding all the precedence values, it would be good to have test coverage for each variant here for oracle as well

xitep: there's probably no point in overriding this. i'll remove it. oracle [lists the precedence for these the same](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/About-SQL-Conditions.html#GUID-65B103FE-C00C-46A3-8173-A731DBF62C80), but i can't come up with a statement that would involve some of these intertwined.

iffyio (approved):
LGTM! Thanks @xitep!

PR #2122, opened 2025-12-03
Added alter external table support for snowflake

romanoff:
Added snowflake alter external table
Spec: https://docs.snowflake.com/en/sql-reference/sql/alter-external-table


tests/sqlparser_snowflake.rs
```diff
@@ -4635,3 +4635,9 @@ fn test_alter_dynamic_table() {
     snowflake().verified_stmt("ALTER DYNAMIC TABLE my_dyn_table SUSPEND");
     snowflake().verified_stmt("ALTER DYNAMIC TABLE my_dyn_table RESUME");
 }
+
+#[test]
+fn test_alter_external_table() {
+    snowflake().verified_stmt("ALTER EXTERNAL TABLE some_table REFRESH");
```
iffyio, line 4641: can we add a test scenario that includes the 'IF EXISTS' clause?

romanoff: @iffyio Updated. Thank you

PR #2123, opened 2025-12-03
Support parsing parenthesized wildcard `(*)`

romanoff:
Support parsing parenthesized wildcard `(*)`

Example of query:
```
SELECT DISTINCT (*) FROM table1
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1276,8 +1276,6 @@ impl<'a> Parser<'a> {
                     self.next_token(); // consume RParen
                     return Ok(Expr::Wildcard(AttachedToken(mul_token)));
                 }
-                // Not a (*), fall through to reset index and call parse_expr
-                self.prev_token();
             }
             _ => (),
         };


tests/sqlparser_common.rs
```diff
@@ -17905,3 +17905,22 @@ fn test_parse_set_session_authorization() {
         }))
     );
 }
+
+#[test]
+fn parse_select_distinct_parenthesized_wildcard() {
```
```suggestion
fn parse_select_parenthesized_wildcard() {
```
thinking since the syntax isn't implemented solely on the distinct keyword?

src/parser/mod.rs
```diff
@@ -1222,6 +1222,15 @@ impl<'a> Parser<'a> {
...
                 return Ok(Expr::Wildcard(AttachedToken(next_token)));
             }
+            // Handle parenthesized wildcard: (*)
+            Token::LParen => {
+                let inner_token = self.next_token();
+                if inner_token.token == Token::Mul && self.peek_token().token == Token::RParen {
+                    self.next_token(); // consume RParen
+                    return Ok(Expr::Wildcard(AttachedToken(inner_token)));
+                }
+                // Not a (*), reset and fall through to parse_expr
```
iffyio: this looks incorrect per the comment, by reset should we have called `prev_token()` or similar to undo the consumed inner token `*`?

src/parser/mod.rs
```diff
@@ -1222,6 +1222,15 @@ impl<'a> Parser<'a> {
             Token::Mul => {
                 return Ok(Expr::Wildcard(AttachedToken(next_token)));
             }
+            // Handle parenthesized wildcard: (*)
+            Token::LParen => {
+                let inner_token = self.next_token();
+                if inner_token.token == Token::Mul && self.peek_token().token == Token::RParen {
```
iffyio: can this be simplified by calling `self.peek_tokens_ref(Token::MUL, Token::RParen)`?

romanoff:
@iffyio Updated. Thank you for reviewing

src/parser/mod.rs
```diff
@@ -1222,6 +1222,17 @@ impl<'a> Parser<'a> {
...
+            // Handle parenthesized wildcard: (*)
+            Token::LParen => {
+                let [maybe_mul, maybe_rparen] = self.peek_tokens_ref();
+                if maybe_mul.token == Token::Mul && maybe_rparen.token == Token::RParen {
+                    let mul_token = self.next_token(); // consume Mul
+                    self.next_token(); // consume RParen
+                    return Ok(Expr::Wildcard(AttachedToken(mul_token)));
+                }
+                // Not a (*), fall through to reset index and call parse_expr
+                self.prev_token();
```
iffyio: calling `self.prev_token()` doesn't seem needed? similar to the `_` branch, the fall-through paths call `self.index = index;` to reset at the end it looks like

romanoff: Updated

iffyio (approved):
LGTM! Thanks @romanoff!

PR #2124, opened 2025-12-04
GenericDialect: support colon operator for JsonAccess

Samyak2:
- Port JsonAccess colon operator from Snowflake to Generic dialect
- This will be used in variant data type support in Datafusion
- see discussion in https://github.com/datafusion-contrib/datafusion-variant/issues/2

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3921,7 +3921,7 @@ impl<'a> Parser<'a> {
                 expr: Box::new(expr),
             })
         } else if Token::LBracket == *tok && self.dialect.supports_partiql()
-            || (dialect_of!(self is SnowflakeDialect | GenericDialect) && Token::Colon == *tok)
+            || (Token::Colon == *tok)
         {
             self.prev_token();
             self.parse_json_access(expr)
@@ -3957,7 +3957,8 @@ impl<'a> Parser<'a> {
         let lower_bound = if self.consume_token(&Token::Colon) {
             None
         } else {
-            Some(self.parse_expr()?)
+            // parse expr until we hit a colon (or any token with lower precedence)
+            Some(self.parse_subexpr(self.dialect.prec_value(Precedence::Colon))?)
         };
 
         // check for end
@@ -3985,7 +3986,8 @@ impl<'a> Parser<'a> {
                 stride: None,
             });
         } else {
-            Some(self.parse_expr()?)
+            // parse expr until we hit a colon (or any token with lower precedence)
+            Some(self.parse_subexpr(self.dialect.prec_value(Precedence::Colon))?)
         };
 
         // check for end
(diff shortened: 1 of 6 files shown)


Samyak2:
I will fix the CI in a bit - sorry did not notice that!

Samyak2:
CI should pass now. There were some changes in main that I had not rebased on.

tests/sqlparser_common.rs
```diff
@@ -17972,3 +17972,126 @@ fn parse_select_parenthesized_wildcard() {
     assert_eq!(select2.projection.len(), 1);
     assert!(matches!(select2.projection[0], SelectItem::Wildcard(_)));
 }
+
+// https://docs.snowflake.com/en/user-guide/querying-semistructured
```
```suggestion
```

tests/sqlparser_common.rs
```diff
@@ -17972,3 +17972,126 @@ fn parse_select_parenthesized_wildcard() {
...
+#[test]
+fn parse_semi_structured_data_traversal() {
+    let dialects = TestedDialects::new(vec![
+        Box::new(GenericDialect {}),
+        Box::new(SnowflakeDialect {}),
+    ]);
```
iffyio, line 18077: should this be all dialects instead? we don't seem to have special handling for generic and snowflake

src/parser/mod.rs
```diff
@@ -3847,7 +3847,8 @@ impl<'a> Parser<'a> {
         let lower_bound = if self.consume_token(&Token::Colon) {
             None
         } else {
-            Some(self.parse_expr()?)
+            // parse expr until we hit a colon (or any token with lower precedence)
+            Some(self.parse_subexpr(self.dialect.prec_value(Precedence::Colon))?)
```
iffyio, line 3961: for the changes to the subscript behavior, we don't seem to have any new tests to accompany them?

tests/sqlparser_common.rs
Samyak2: We actually do have special handling for `SnowflakeDialect` and `GenericDialect`: https://github.com/apache/datafusion-sqlparser-rs/blob/ce74e7fe217ac5d2cb5392237a03c7d252a3ea6a/src/parser/mod.rs#L3814

(this is in main currently, not a change in this PR)

From what I know, only Snowflake and Databricks support this syntax. So I can expand the test to include Databricks as well. But I don't think it would make sense to run this test on dialects that don't support this syntax. What do you think?

src/parser/mod.rs
Samyak2: These changes actually came from a failing test, but I will add some more tests to explicitly look for this behavior.

alamb:
@Samyak2  any chance you can resolve the conflicts in this PR and address @iffyio 's comments? Then we can merge it in

tests/sqlparser_common.rs
Samyak2: I have added Databricks to the list of dialects to test for.

src/parser/mod.rs
Samyak2: I have added some `parse_array_subscript` tests. I have verified that these catch the fixes made by these changes in subscript behavior.

Samyak2:
> @Samyak2 any chance you can resolve the conflicts in this PR and address @iffyio 's comments? Then we can merge it in

Done!

PR #2125, opened 2025-12-05
Added support for `ALTER OPERATOR FAMILY` syntax

LucaCappelletti94:
This PR adds support for the [`ALTER OPERATOR FAMILY`](https://www.postgresql.org/docs/current/sql-alteropfamily.html) syntax.

It was honestly much more complex than I expected, it is a very rich syntax.


iffyio (approved):

PR #2130, opened 2025-12-10
Oracle: Support for quote delimited strings

xitep:


src/tokenizer.rs
```diff
@@ -1994,6 +2023,70 @@ impl<'a> Tokenizer<'a> {
...
+    ) -> Result<Token, TokenizerError> {
+        match chars.peek() {
+            Some('\'') => {
+                chars.next();
+                // ~ determine the "quote character(s)"
+                let error_loc = chars.location();
+                let (start_quote_char, end_quote_char) = match chars.next() {
+                    // ~ "newline" is not allowed by Oracle's SQL Reference,
+                    // but works with sql*plus nevertheless
+                    None | Some(' ') | Some('\t') | Some('\r') | Some('\n') => {
```
iffyio: can we ensure that these cases are covered by the tests?

src/tokenizer.rs
```diff
@@ -1994,6 +2023,70 @@ impl<'a> Tokenizer<'a> {
...
+                        if let Some('\'') = chars.peek() {
+                            chars.next(); // ~ consume the quote
+                            return Ok(as_literal(start_quote_char, s, end_quote_char));
+                        }
+                    }
+                    s.push(ch);
+                }
+                self.tokenizer_error(error_loc, "Unterminated string literal")
+            }
+            // ~ not a literal introduced with _token_prefix_, assm
```
```suggestion
```

src/tokenizer.rs
```diff
@@ -1994,6 +2023,70 @@ impl<'a> Tokenizer<'a> {
...
+                while let Some(ch) = chars.next() {
+                    if ch == end_quote_char {
+                        if let Some('\'') = chars.peek() {
+                            chars.next(); // ~ consume the quote
+                            return Ok(as_literal(start_quote_char, s, end_quote_char));
+                        }
+                    }
+                    s.push(ch);
+                }
+                self.tokenizer_error(error_loc, "Unterminated string literal")
```
iffyio: can we add a test case for this?

src/tokenizer.rs
```diff
@@ -1994,6 +2023,70 @@ impl<'a> Tokenizer<'a> {
...
+        word_prefix: &[char],
+        // turns an identified quote string literal,
+        // ie. `(start-quote-char, string-literal, end-quote-char)`
+        // into a token
+        as_literal: fn(char, String, char) -> Token,
+    ) -> Result<Token, TokenizerError> {
+        match chars.peek() {
+            Some('\'') => {
+                chars.next();
+                // ~ determine the "quote character(s)"
```
```suggestion
```

src/tokenizer.rs
```diff
@@ -1994,6 +2023,70 @@ impl<'a> Tokenizer<'a> {
...
+    fn tokenize_word_or_quote_delimited_string(
+        &self,
+        chars: &mut State,
+        // the prefix that introduced the possible literal or word,
+        // e.g. "Q" or "nq"
+        word_prefix: &[char],
+        // turns an identified quote string literal,
+        // ie. `(start-quote-char, string-literal, end-quote-char)`
+        // into a token
+        as_literal: fn(char, String, char) -> Token,
```
iffyio: since the function is only used twice, I think it would be reasonable to have the caller do something like

```rust
chars.next(); // consume 'q'
if chars.peek() != Some('\'') {
    return 'q' + tokenize_word()
}
let QuoteString(quote_delimited_string())
```
i.e. that this function should only care about parsing the string, in order to simplify its implementation

src/ast/value.rs
```diff
@@ -167,6 +167,12 @@ pub enum Value {
     TripleDoubleQuotedRawStringLiteral(String),
     /// N'string value'
     NationalStringLiteral(String),
+    /// Quote delimited literal. Examples `Q'{ab'c}'`, `Q'|ab'c|'`, `Q'|ab|c|'`
+    /// [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Literals.html)
+    QuoteDelimitedStringLiteral(char, String, char),
```
iffyio: We can add something like 
```rust
struct QuoteDelimitedString {
    start_quote: char,
    value: String,
    end_quote: char
}
```
that would better clarify the args to the enum variants

src/tokenizer.rs
```diff
@@ -98,6 +98,12 @@ pub enum Token {
     TripleDoubleQuotedRawStringLiteral(String),
     /// "National" string literal: i.e: N'string'
     NationalStringLiteral(String),
+    /// Quote delimited literal. Examples `Q'{ab'c}'`, `Q'|ab'c|'`, `Q'|ab|c|'`
+    /// [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Literals.html)
```
iffyio: for the docs, can we point to this link instead, that take the user directly to text literals?
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Literals.html#GUID-1824CBAA-6E16-4921-B2A6-112FB02248DA

src/tokenizer.rs
```diff
@@ -1032,13 +1042,32 @@ impl<'a> Tokenizer<'a> {
                                 self.tokenize_single_quoted_string(chars, '\'', backslash_escape)?;
                             Ok(Some(Token::NationalStringLiteral(s)))
                         }
+                        Some(&q @ 'q') | Some(&q @ 'Q') if dialect_of!(self is OracleDialect | GenericDialect) =>
```
iffyio: We can add a dialect method `dialect.supports_quote_delimited_string()` instead of hardcoding the supported dialect in the tokenizer

tests/sqlparser_oracle.rs
```diff
@@ -103,3 +113,192 @@ fn plusminus_have_same_precedence_as_strconcat() {
...
+                      Q'|abc'''|', \
+                      Q'{abc}d}', \
+                      Q'[]abc[]', \
+                      Q'<a'bc>', \
+                      Q'<<<a'bc>', \
+                      Q'('abc'('abc)', \
+                      Q'(abc'def))', \
+                      Q'(abc'def)))' \
+                 FROM dual";
+    let select = oracle().verified_only_select(sql);
```
iffyio: can we change the dialect being tested to use the dialect method?

```rust
let dialect = all_dialects_where(|d| d.supports_quote_delimited_string());
dialect.verified_only_select(sql);
```

tests/sqlparser_oracle.rs
```diff
@@ -33,6 +34,15 @@ fn oracle() -> TestedDialects {
     TestedDialects::new(vec![Box::new(OracleDialect)])
 }
 
+/// Convenience constructor for [QuoteDelimitedstring].
+fn qds(start_quote: char, value: &'static str, end_quote: char) -> QuoteDelimitedString {
```
```suggestion
fn quote_delimited_string(start_quote: char, value: &'static str, end_quote: char) -> QuoteDelimitedString {
```

iffyio (approved):

PR #2131, opened 2025-12-12
PostgreSQL: Support schema-qualified operator classes in CREATE INDEX

dddenis:
This PR adds support for schema-qualified operator classes in PostgreSQL `CREATE INDEX` statements. Previously, the parser only accepted simple identifiers for operator classes (e.g., `vector_cosine_ops`). With this change, schema-qualified names are now supported (e.g., `public.vector_cosine_ops`).

The following SQL is now correctly parsed:
```sql
CREATE INDEX my_index ON my_table USING HNSW (embedding public.vector_cosine_ops);
```


iffyio (approved):
LGTM! Thanks @dddenis!

Issue #2006, opened 2025-08-19
Time travel support


westonpace:
Several databases and data lakes support time travel which allows historical versions of a table to be queried.  Regrettably, syntax and capabilities do not seem to be terribly consistent across the various implementations.

## Querying Historical Times

This option lets a database be queried as it existed at a certain point in time.  (*Question:* Is this "last version prior to the given time"?  or is it an error if no snapshot exists with that exact time?)

### [Databricks](https://docs.databricks.com/gcp/en/delta/history#delta-time-travel-syntax)

```
SELECT * FROM people10m TIMESTAMP AS OF '2018-10-18T22:15:12.013Z';
```

### [Snowflake](https://docs.snowflake.com/en/sql-reference/constructs/at-before)

```
SELECT * FROM my_table AT(TIMESTAMP => 'Wed, 26 Jun 2024 09:20:00 -0700'::TIMESTAMP_LTZ);
```

### [Cockroach](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/)

```
SELECT * FROM t AS OF SYSTEM TIME '2016-06-15 12:45:00'
```

### [DuckLake](https://ducklake.select/docs/stable/duckdb/usage/time_travel.html)

```
SELECT * FROM tbl AT (TIMESTAMP => now() - INTERVAL '1 week');
```

## Querying Specific versions

This options lets a specific "version" of a database to be queried.  What exactly this means seems to depend on the database implementation in question.

### Databricks

```
SELECT * FROM people10m VERSION AS OF 123;
```

### Snowflake

Note: unclear if this is the same thing (specific versions) or different

```
SELECT * FROM my_table BEFORE(STATEMENT => '8e5d0ca9-005e-44e6-b858-a8f5b37c5726');
```

### Cockroach

Unsupported

### DuckLake

```
SELECT * FROM tbl AT (VERSION => 3);
```

JamesVorder:
I've started taking a look at this, and adding support for these seems doable. The functionality can be bolted onto `maybe_parse_table_version`. I can start by adding support for Databricks time travel, and get a PR together for that on its own to keep it small.

PR #2134, opened 2025-12-17
Databricks: Support Timetravel With "TIMESTAMP AS OF"

JamesVorder:
This PR adds support for Databricks' timstamp-based [time travel](https://docs.databricks.com/aws/en/delta/history#delta-time-travel-syntax). This is a first step toward resolvign https://github.com/apache/datafusion-sqlparser-rs/issues/2006.

The following is now supported:

```sql
SELECT 1 FROM t1 TIMESTAMP AS OF '2018-10-18T22:15:12.013Z';
SELECT 1 FROM t1 TIMESTAMP AS OF CURRENT_TIMESTAMP() - INTERVAL 12 HOURS;
-- etc.
```


src/dialect/databricks.rs
```diff
@@ -47,6 +47,11 @@ impl Dialect for DatabricksDialect {
         true
     }
 
+    // https://docs.databricks.com/gcp/en/delta/history#delta-time-travel-syntax
```
iffyio:
```suggestion
    /// https://docs.databricks.com/gcp/en/delta/history#delta-time-travel-syntax
```

src/parser/mod.rs
```diff
@@ -15188,6 +15137,11 @@ impl<'a> Parser<'a> {
                 let func_name = self.parse_object_name(true)?;
                 let func = self.parse_function(func_name)?;
                 return Ok(Some(TableVersion::Function(func)));
+            } else if dialect_of!(self is DatabricksDialect)
```
iffyio: I think this should use the defined `supports_timestamp_versioning`?

tests/sqlparser_bigquery.rs
```diff
@@ -1753,6 +1753,10 @@ fn parse_table_time_travel() {
 
     let sql = "SELECT 1 FROM t1 FOR SYSTEM TIME AS OF 'some_timestamp'".to_string();
     assert!(bigquery().parse_sql_statements(&sql).is_err());
+
+    // The following time travel syntax(es) are invalid in BigQuery dialect
```
iffyio:
```suggestion
```
since the comment is already implied by the test

tests/sqlparser_databricks.rs
```diff
@@ -366,3 +366,87 @@ fn data_type_timestamp_ntz() {
         s => panic!("Unexpected statement: {s:?}"),
     }
 }
+
+#[test]
+fn parse_table_time_travel() {
+    let version = "2018-10-18T22:15:12.013Z".to_string();
+    let sql = format!("SELECT 1 FROM t1 TIMESTAMP AS OF '{version}'");
+    let select = databricks().verified_only_select(&sql);
```
iffyio: for the tests we can use
```rust
let dialect = all_dialecs_where(|d| d.supports_timestamp_versioning());
```

tests/sqlparser_bigquery.rs
JamesVorder: Addressed in https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/324ade9afc7c88778cb530528bec3d335de88f71

src/parser/mod.rs
JamesVorder: I think relying on `supports_timestamp_versioning` alone might be insufficient. Please help me understand the desired behavior, I'm pretty new to the project.

The reason that I think that `supports_timestamp_versioning` may be insufficient on its own is that if we rely exclusively on that function to decide whether we parse `TIMESTAMP AS OF` syntax, then it will become available in the BigQuery dialect (and Snowflake, and MSSQL) - incorrectly. This is the reason I added an assertion to the BigQuery time travel test, looking for an error when `TIMESTAMP AS OF` is used.

In fact, I think what I have in this PR is also insufficient. If I use valid Snowflake time travel syntax (`SELECT * FROM tbl AT(TIMESTAMP => '2024-12-15 00:00:00')`), it's being accepted by the Databricks dialect. Instead, I would expect the Databricks dialect to throw an error (`AT...` is not valid time travel syntax in Databricks.)

Please help me understand the desired behavior here. Am I right to expect that an error should be thrown in the cases mentioned?

JamesVorder: To demonstrate my thinking here, I've added the following 2 commits, which result in two failing tests:
* https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/b3534992a082e3c5d13129862843c60b8c42627a -- `sqlparse_bigquery::parse_table_timetravel` fails after this commit
* https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/759406fc535ca722543d04b545194b477c3c3c98 -- `sqlparse_databricks::parse_table_timetravel` fails after this commit

Are these failures valid, or is my expectation incorrect?

iffyio: Yeah its fine for the parser to be permissive and accept sql statements for a dialect that wouldn't necessarily be supported by the dialect (downstream crates are expected to validate the returned AST if the behavior is undesirable for a specific use case), so that e.g. having bigquery accept the 'TIMESTAMP AS OF' would be fine.

tests/sqlparser_databricks.rs
JamesVorder: Addressed in https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/270c41b9ac6c46b396dc47ec435f1a37afe6e883

src/parser/mod.rs
JamesVorder: Ah! That makes sense, thank you for explaining. I believe this is addressed by the following 2 commits:
* https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/21783cc67360cc4e0a53f91cd04349b668fcbc33
* https://github.com/apache/datafusion-sqlparser-rs/pull/2134/commits/270c41b9ac6c46b396dc47ec435f1a37afe6e883

iffyio:
@JamesVorder could you take a look at the ci failure?

JamesVorder:
@iffyio Can you see if the build failure is resolved? Looks like it was just a silly mistake with a docstring.

PR #2135, opened 2025-12-18
Added support for `ALTER OPERATOR CLASS` syntax

LucaCappelletti94:
Added support for [`ALTER OPERATOR CLASS`](https://www.postgresql.org/docs/17//sql-alteropclass.html) syntax.


src/parser/mod.rs
```diff
@@ -10300,6 +10302,34 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    pub fn parse_alter_operator_class(&mut self) -> Result<Statement, ParserError> {
```
iffyio, line 10567: Can we add documentation for the function?

LucaCappelletti94: Can I add the deny missing doc to clippy?

LucaCappelletti94: I have added the deny missing doc, but there is an impressive amount of missing doc. I am trying to add the missing documentation so to make it impossible for missing documentation in future development.

LucaCappelletti94: Ok, it was quite an effort but now `#![forbid(missing_docs)]` is in place and everything public is documented

iffyio (approved):

iffyio:
@LucaCappelletti94 could you look to resolve the conflicts on this branch when you have some time?

LucaCappelletti94:
I fixed stuff rapidly in the github editor, so likely there will be some mess. Will fix in a sec.

LucaCappelletti94:
I have documented the code that had no documentation in the main.

Issue #1994, opened 2025-08-06
[Postgres] Parsing sequence of `SHOW` and `SET` is a bit off


vrmiguel:
```rust
Parser::parse_sql(&PostgreSqlDialect {}, "SHOW search_path; SHOW timezone; SET search_path TO public;").unwrap()
```

The above results in:

```rust
[
    ShowVariable {
        variable: [
            Ident {
                value: "search_path",
                quote_style: None,
                span: Span(Location(1,6)..Location(1,17)),
            },
            Ident {
                value: "SHOW",
                quote_style: None,
                span: Span(Location(1,19)..Location(1,23)),
            },
            Ident {
                value: "timezone",
                quote_style: None,
                span: Span(Location(1,24)..Location(1,32)),
            },
            Ident {
                value: "SET",
                quote_style: None,
                span: Span(Location(1,34)..Location(1,37)),
            },
            Ident {
                value: "search_path",
                quote_style: None,
                span: Span(Location(1,38)..Location(1,49)),
            },
            Ident {
                value: "TO",
                quote_style: None,
                span: Span(Location(1,50)..Location(1,52)),
            },
            Ident {
                value: "public",
                quote_style: None,
                span: Span(Location(1,53)..Location(1,59)),
            },
        ],
    },
]
```

Its `Display` impl results in `"SHOW search_path SHOW timezone SET search_path TO public"`

LucaCappelletti94:
Closing as resolved in https://github.com/apache/datafusion-sqlparser-rs/pull/2137

PR #2137, opened 2025-12-19
Fix parse_identifiers not taking semicolons into account

jnlt3:
Related to: https://github.com/apache/datafusion-sqlparser-rs/issues/1994

The PR fixes the issue with semicolons disappearing when `SHOW` statements are parsed. The test prior to the fix would fail as follows:
```
assertion `left == right` failed
  left: "SHOW search_path; SELECT 1"
 right: "SHOW search_path SELECT"
```

The fix adds a `Semicolon` alongside `EOF` and `Eq` which were the only two tokens that triggered a break in `parse_identifiers`. This should not cause any unintended side effects as semicolons are exclusively used for statement ends.

Additional considerations:
- The `parse_show` test requires `SHOW ALL ALL` to be parsed. This is not valid PostgreSQL syntax according to [postgres 18 docs](https://www.postgresql.org/docs/18/sql-show.html) and was not valid since version 7.1 at least. 
- `parse_identifiers` is seemingly supposed to be very relaxed (as indicated by the above test not failing) and also is used only twice in the code base. The first instance is `SHOW`, the other is the `DROP` that comes after a pipe operator for which the current tests do not need the extra flexibility provided by `parse_identifiers`.

I am willing to make a separate PR for modifying the `parse_show` test to use valid PostgreSQL syntax, I am not familiar enough with the code base to do anything with the latter.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -12024,7 +12024,7 @@ impl<'a> Parser<'a> {
                 Token::Word(w) => {
                     idents.push(w.clone().into_ident(self.peek_token_ref().span));
                 }
-                Token::EOF | Token::Eq => break,
+                Token::EOF | Token::Eq | Token::SemiColon => break,
                 _ => {}
             }
             self.advance_token();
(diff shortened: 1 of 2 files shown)


yoavcloud:
LGTM, nicely done @jnlt3!

FYI @iffyio

PR #2139, opened 2025-12-21
Make benchmark statement valid

xitep:


sqlparser_bench/benches/sqlparser_bench.rs
```diff
@@ -51,7 +51,7 @@ fn basic_queries(c: &mut Criterion) {
         let tables = (0..1000)
             .map(|n| format!("TABLE_{n}"))
             .collect::<Vec<_>>()
-            .join(" JOIN ");
+            .join(" CROSS JOIN ");
```
alamb, line 54: The issue here is that there is no ON clause, right?

xitep: yep, exactly

PR #2140, opened 2025-12-25
Key Value Options: add support for trailing semicolon

yoavcloud:


src/parser/mod.rs
```diff
@@ -19536,4 +19311,11 @@ mod tests {
             assert!(Parser::parse_sql(&GenericDialect, &sql).is_err());
         }
     }
+
+    #[test]
+    fn test_key_value_options_trailing_semicolon() {
+        let sql = "CREATE USER u1 option1 = 'value1' option2 = 'value2';";
+        let ast = Parser::parse_sql(&GenericDialect, sql);
+        assert!(ast.is_ok());
+    }
```
iffyio: Can we have it as a test case in sqlparser_common.rs? also if the test case can cover all applicable dialects

PR #2141, opened 2025-12-26
Redshift: Add support for optional JSON format in copy option

yoavcloud:


iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2142, opened 2025-12-26
PostgreSQL: ALTER USER password option

yoavcloud:


src/ast/mod.rs
```diff
@@ -10046,12 +10046,15 @@ impl fmt::Display for CreateUser {
 
 /// Modifies the properties of a user
 ///
-/// Syntax:
+/// [Snowflake Syntax]:
 /// ```sql
 /// ALTER USER [ IF EXISTS ] [ <name> ] [ OPTIONS ]
 /// ```
 ///
-/// [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/alter-user)
```
iffyio, line 10116: It seems we dropped this doc reference, could we add it back together with the new pg reference?

tests/sqlparser_common.rs
```diff
@@ -17885,6 +17885,12 @@ fn test_parse_alter_user() {
         _ => unreachable!(),
     }
     verified_stmt("ALTER USER u1 SET DEFAULT_SECONDARY_ROLES=('ALL'), PASSWORD='secret', WORKLOAD_IDENTITY=(TYPE=AWS, ARN='arn:aws:iam::123456789:r1/')");
+
+    verified_stmt("ALTER USER u1 PASSWORD 'AAA'");
```
iffyio, line 17954: can we add test cases for `ENCRYPTED` and `PASSWORD NULL`?

src/parser/alter.rs
```diff
@@ -292,6 +293,24 @@ impl Parser<'_> {
...
+                    encrypted,
+                    password: None,
+                })
+            } else {
+                let password = self.parse_literal_string()?;
+                Some(AlterUserPassword {
+                    encrypted,
+                    password: Some(password),
+                })
+            }
```
iffyio: minor probably to clarify the code, we could do e.g.

```rust
let password = if self.parse_keyword(Keyword::NULL) {
    None
} else {
    Some(self.parse_literal_string()?)
};
Some(AlterUserPassword {
    encrypted,
    password,
})
```

PR #2144, opened 2025-12-29
MySQL: Add support for && as boolean AND

yoavcloud:


src/dialect/mod.rs
```diff
@@ -1216,6 +1216,11 @@ pub trait Dialect: Debug + Any {
     fn supports_quote_delimited_string(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect considers the `&&` operator as a boolean AND operator.
+    fn supports_overlap_as_and_operator(&self) -> bool {
```
```suggestion
    fn supports_double_ampersand_operator(&self) -> bool {
```
we could probably call it something like this? (Its not clear what overlap means in this context)

PR #2146, opened 2025-12-30
MySQL: Add support for casting using the BINARY keyword

yoavcloud:
MySQL allows casting using the `BINARY` keyword, for example:
```
SELECT BINARY 1+1
```

Adding support for parsing this as `SELECT CAST(1+1 AS BINARY)`. Ideally, the parser would maintain the original query form, however, since this is deprecated in MySQL, I preferred a less instrusive solution (versus introducing a `TypedExpr` struct silimar to, or to replace `TypedString`).

### DIFF
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -176,7 +176,7 @@ impl Dialect for MySqlDialect {
     fn supports_double_ampersand_operator(&self) -> bool {
         true
     }
-    
+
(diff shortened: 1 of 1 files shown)


iffyio (approved):
LGTM! Thanks @yoavcloud!

iffyio:
FYI @yoavcloud there's some conflicts in this branch when you have some time

PR #2147, opened 2025-12-30
MySQL: Add support for the PUBLIC privilege

yoavcloud:


yoavcloud:
The fix suggested here is incorrect, MySQL doesn't have a PUBLIC privilege, I need to find the right solution.

PR #2148, opened 2026-01-04
Snowflake: Add multi table insert support

finchxxia:
To solve issue #2150 

1. I found that datafusion-sqlparser-rs cannot support [multi-table insert](https://docs.snowflake.com/en/sql-reference/sql/insert-multi-table) currently. 
```
-- Unconditional multi-table insert
INSERT [ OVERWRITE ] ALL
  intoClause [ ... ]
<subquery>

-- Conditional multi-table insert
INSERT [ OVERWRITE ] { FIRST | ALL }
  { WHEN <condition> THEN intoClause [ ... ] }
  [ ... ]
  [ ELSE intoClause ]
<subquery>
```

2. Another case is when I using **SAMPLE** in subquery, it cannot work normally. 
```
SELECT * FROM (SELECT * FROM (select report_from from my_table) SAMPLE (10000 ROWS)) AS anon_1
```

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -36,11 +36,11 @@ use crate::ast::{
     RefreshModeKind, RowAccessPolicy, ShowObjects, SqlOption, Statement,
     StorageSerializationPolicy, TableObject, TagsColumnOption, Value, WrappedCollection,
 };
-use crate::tokenizer::TokenWithSpan;
 use crate::dialect::{Dialect, Precedence};
 use crate::keywords::Keyword;
 use crate::parser::{IsOptional, Parser, ParserError};
 use crate::tokenizer::Token;
+use crate::tokenizer::TokenWithSpan;
 #[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
 #[cfg(not(feature = "std"))]
@@ -1824,7 +1824,9 @@ fn parse_multi_table_insert_into_clause(
 }
 
 /// Parse a single value in a multi-table INSERT VALUES clause.
-fn parse_multi_table_insert_value(parser: &mut Parser) -> Result<MultiTableInsertValue, ParserError> {
+fn parse_multi_table_insert_value(
+    parser: &mut Parser,
+) -> Result<MultiTableInsertValue, ParserError> {
     if parser.parse_keyword(Keyword::DEFAULT) {
         Ok(MultiTableInsertValue::Default)
     } else {
@@ -1835,8 +1837,13 @@ fn parse_multi_table_insert_value(parser: &mut Parser) -> Result<MultiTableInser
 /// Parse WHEN clauses for conditional multi-table INSERT.
 fn parse_multi_table_insert_when_clauses(
     parser: &mut Parser,
-) -> Result<(Vec<MultiTableInsertWhenClause>, Option<Vec<MultiTableInsertIntoClause>>), ParserError>
-{
+) -> Result<
+    (
+        Vec<MultiTableInsertWhenClause>,
+        Option<Vec<MultiTableInsertIntoClause>>,
+    ),
+    ParserError,
+> {
     let mut when_clauses = vec![];
     let mut else_clause = None;
 
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -84,7 +84,8 @@ pub use self::ddl::{
 pub use self::dml::{
     Delete, Insert, Merge, MergeAction, MergeClause, MergeClauseKind, MergeInsertExpr,
     MergeInsertKind, MergeUpdateExpr, MultiTableInsertIntoClause, MultiTableInsertType,
-    MultiTableInsertValue, MultiTableInsertValues, MultiTableInsertWhenClause, OutputClause, Update,
+    MultiTableInsertValue, MultiTableInsertValues, MultiTableInsertWhenClause, OutputClause,
+    Update,
 };
 pub use self::operator::{BinaryOperator, UnaryOperator};
 pub use self::query::{
--- a/src/ast/spans.rs
+++ b/src/ast/spans.rs
@@ -1312,8 +1312,8 @@ impl Spanned for Insert {
             priority: _,     // todo, mysql specific
             insert_alias: _, // todo, mysql specific
             assignments,
-            settings: _,      // todo, clickhouse specific
-            format_clause: _, // todo, clickhouse specific
+            settings: _,                 // todo, clickhouse specific
+            format_clause: _,            // todo, clickhouse specific
             multi_table_insert_type: _,  // snowflake multi-table insert
             multi_table_into_clauses: _, // snowflake multi-table insert
             multi_table_when_clauses: _, // snowflake multi-table insert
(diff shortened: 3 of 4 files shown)


finchxxia:
Hi @iffyio,

 I've added the test about multi-table insert and remove all changes about sample data in subquery to another issue & PR. 
Plz take a look again:)

src/ast/dml.rs
```diff
@@ -90,10 +90,80 @@ pub struct Insert {
...
+    pub multi_table_else_clause: Option<Vec<MultiTableInsertIntoClause>>,
 }
 
 impl Display for Insert {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Check if this is a Snowflake multi-table insert
+        let is_multi_table = !self.multi_table_into_clauses.is_empty()
+            || !self.multi_table_when_clauses.is_empty();
+
+        if is_multi_table {
```
iffyio: Can we would need to merge this logic with the rest of the function, instead vs having a clause that is dedicated to the multi_table mode currently? with the latter is unclear how much is covered by the clause vs the rest of the statement and whether we could end up with incorrect formatting as a result. essentially that we update the current function at the points where we have new entries to display

src/dialect/snowflake.rs
```diff
@@ -318,6 +320,30 @@ impl Dialect for SnowflakeDialect {
             parser.prev_token();
         }
 
+        // Check for multi-table INSERT (Snowflake specific)
+        // INSERT [OVERWRITE] ALL ... or INSERT [OVERWRITE] FIRST ...
```
```suggestion
        // Check for multi-table INSERT
        // `INSERT [OVERWRITE] ALL ... or INSERT [OVERWRITE] FIRST ...`
```

src/dialect/snowflake.rs
```diff
@@ -1628,3 +1654,173 @@ fn parse_show_objects(terse: bool, parser: &mut Parser) -> Result<Statement, Par
...
+/// Syntax: `INTO <table> [ ( <columns> ) ] [ VALUES ( <values> ) ]`
+fn parse_multi_table_insert_into_clause(
+    parser: &mut Parser,
+) -> Result<MultiTableInsertIntoClause, ParserError> {
+    let table_name = parser.parse_object_name(false)?;
+
+    // Parse optional column list
+    let columns = if parser.peek_token() == Token::LParen && !parser.peek_keyword(Keyword::VALUES) {
+        // Check if this is a column list (not VALUES)
+        let peeked = parser.peek_tokens::<2>();
```
iffyio: can we use pattern matching instead of array indexing the result? since the latter is error prone. just in case, if we only need to peek one item we can call `peek_token_ref` instead

src/dialect/snowflake.rs
```diff
@@ -1628,3 +1654,173 @@ fn parse_show_objects(terse: bool, parser: &mut Parser) -> Result<Statement, Par
...
+/// Parse a single INTO clause for multi-table INSERT.
+///
+/// Syntax: `INTO <table> [ ( <columns> ) ] [ VALUES ( <values> ) ]`
+fn parse_multi_table_insert_into_clause(
+    parser: &mut Parser,
+) -> Result<MultiTableInsertIntoClause, ParserError> {
+    let table_name = parser.parse_object_name(false)?;
+
+    // Parse optional column list
+    let columns = if parser.peek_token() == Token::LParen && !parser.peek_keyword(Keyword::VALUES) {
```
iffyio: is the negated peek values condition always true?

iffyio: It looks like this and the nested if statement can be collapsed to `if self.consume_token(LParen)`?

src/ast/dml.rs
finchxxia: that makes sense. I'll merge this logic to standard flow.

src/dialect/snowflake.rs
finchxxia: > It looks like this and the nested if statement can be collapsed to `if self.consume_token(LParen)`?

I found another function `parse_parenthesized_column_list` which can be simplier to solve it.

src/ast/dml.rs
```diff
@@ -90,92 +90,174 @@ pub struct Insert {
...
 impl Display for Insert {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let table_name = if let Some(alias) = &self.table_alias {
-            format!("{0} AS {alias}", self.table)
-        } else {
-            self.table.to_string()
-        };
+        // Check if this is a Snowflake multi-table insert
+        let is_multi_table = !self.multi_table_into_clauses.is_empty()
+            || !self.multi_table_when_clauses.is_empty();
```
finchxxia: @iffyio Hi, do you think this is a better way to parse multi-table insert statement? I'm not sure if I should refactor the whole insert statement because there are too much difference between standard insert and multi-table insert

iffyio: I think the new diff looks reasonable, but the `is_multi_table` I don't think should be needed. i.e. if a value is present in the struct we should display it (otherwise there's a bug in the parser), the display shouldn't infer meaning from combinations of fields in the struct

finchxxia: > I think the new diff looks reasonable, but the `is_multi_table` I don't think should be needed. i.e. if a value is present in the struct we should display it (otherwise there's a bug in the parser), the display shouldn't infer meaning from combinations of fields in the struct

@iffyio Hi, I add insert type in the struct and display it directly. Could you please take a look again?

iffyio: Oh so here's an example to clarify what I meant in the previous comment. If we take this block in the display logic
```rust
if !is_multi_table {
	let table_name = if let Some(alias) = &self.table_alias {
    	format!("{0} AS {alias}", self.table)
    } else {
    	self.table.to_string()
    };
    write!(f, " {table_name} ")?;
}
```

if we change it to the following - removing the `if` condition:

```rust
let table_name = if let Some(alias) = &self.table_alias {
	format!("{0} AS {alias}", self.table)
} else {
	self.table.to_string()
};
write!(f, " {table_name} ")?;
```

what would be the effect (is the function display still correct)? Its unclear to me if/why the `is_multi_table` flag/condition is needed.

finchxxia: > Oh so here's an example to clarify what I meant in the previous comment. If we take this block in the display logic

@iffyio Oh, sorry for misinterpreting your meaning.  That could be different between these two code blocks. For example, input query can be:

`INSERT ALL INTO t1 INTO t2 SELECT n1 FROM src`

and if we don't add ` if !is_multi_table`, the output would be:
```
INSERT ALL  INTO t1 INTO t2 SELECT n1 FROM src
           ^^extra space here
```

If you don't mind, I can revert my code and add another test for that.

iffyio:
```
INSERT ALL  INTO t1 INTO t2 SELECT n1 FROM src
```
oh what's inserting the extra space in that example? I think ideally we're able to match the input query without the bool flag

finchxxia: @iffyio I see. I can remove this extra space more simply rather than using `is_multi_table`. Could you please take a look again?

iffyio:
@finchxxia could you resolve the conflicts on the branch when you have some time?

finchxxia:
> @finchxxia could you resolve the conflicts on the branch when you have some time?

@iffyio done

iffyio (approved):
Thanks @finchxxia!

PR #2151, opened 2026-01-07
MySQL: Support `CAST(... AS ... ARRAY)` syntax

mvzink:
MySQL has a special case in `CAST` parsing where the type can be followed by `ARRAY`. This is only used for creating multi-valued indexes in InnoDB, so is only allowed in `CREATE TABLE` and other DDL statements when specifying keys. See the [docs].

Given those restrictions, we could be significantly more restrictive in parsing this, e.g. not parsing `ARRAY` unless we are in a key specification. Or, if there was such a thing as a suffix array type definition, we could parse it as a type. But as far as I know, that doesn't exist in any supported SQL dialects, and encountering `ARRAY` here is unambiguous. So it seemed simplest to be permissive and always parse it. The only downside I can see is we are now adding a field to `Expr::Cast`. 

[docs]: https://dev.mysql.com/doc/refman/8.4/en/cast-functions.html#function_cast

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1033,6 +1033,12 @@ pub enum Expr {
         expr: Box<Expr>,
         /// Target data type.
         data_type: DataType,
+        /// [MySQL] allows CAST(... AS type ARRAY) in functional index definitions for InnoDB
+        /// multi-valued indices. It's not really a datatype, and is only allowed in `CAST` in key
+        /// specifications, so it's a flag here.
+        ///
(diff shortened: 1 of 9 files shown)


iffyio:
@mvzink could you take a look at the ci failures when you have some time?

mvzink:
> @mvzink could you take a look at the ci failures when you have some time?

Done. Looks like another instance of `Cast` got added since my last push and needed to be updated with the new field. Thanks!

PR #2152, opened 2026-01-07
MySQL: Parse bitwise shift left/right operators

mvzink:
According to the [docs], MySQL supports `<<` and `>>` operators. They are currently named `PGBitwiseShiftLeft` and `PGBitwiseShiftRight`, respectively, so the name is a bit misleading, but doesn't see worth changing at the moment.

[docs]: https://dev.mysql.com/doc/refman/8.4/en/bit-functions.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3485,10 +3485,10 @@ impl<'a> Parser<'a> {
             Token::DuckIntDiv if dialect_is!(dialect is DuckDbDialect | GenericDialect) => {
                 Some(BinaryOperator::DuckIntegerDivide)
             }
-            Token::ShiftLeft if dialect_is!(dialect is PostgreSqlDialect | DuckDbDialect | GenericDialect | RedshiftSqlDialect | MySqlDialect) => {
+            Token::ShiftLeft if dialect.supports_bitwise_shift_operators() => {
                 Some(BinaryOperator::PGBitwiseShiftLeft)
             }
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -3485,10 +3485,10 @@ impl<'a> Parser<'a> {
             Token::DuckIntDiv if dialect_is!(dialect is DuckDbDialect | GenericDialect) => {
                 Some(BinaryOperator::DuckIntegerDivide)
             }
-            Token::ShiftLeft if dialect_is!(dialect is PostgreSqlDialect | DuckDbDialect | GenericDialect | RedshiftSqlDialect) => {
+            Token::ShiftLeft if dialect_is!(dialect is PostgreSqlDialect | DuckDbDialect | GenericDialect | RedshiftSqlDialect | MySqlDialect) => {
```
yoavcloud: I would add a dialect function such as `supports_shift_left_operator` to adhere to the new coding standard

yoavcloud (approved):
LGTM, nicely done @mvzink!

FYI @iffyio

alexander-beedie:
FYI: bit late now, but the `PGBitwiseNot` operator was also made available to more backends (ref: https://github.com/apache/datafusion-sqlparser-rs/pull/2081), and we did remove the "PG" prefix there (it's now just `BitwiseNot`). Should probably do the same to the shift ops for consistency (follow-up PR?) 👍

PR #2155, opened 2026-01-08
Databricks: Support Timetravel With "VERSION AS OF"

JamesVorder:
This PR adds support for Databricks' version-based [time travel](https://docs.databricks.com/aws/en/delta/history#delta-time-travel-syntax). This is a next step toward resolving https://github.com/apache/datafusion-sqlparser-rs/issues/2006, following https://github.com/apache/datafusion-sqlparser-rs/pull/2134.

The following is now supported:

```sql
SELECT 1 FROM t1 VERSION AS OF 10;
```

**Note:** This PR makes a couple opinionated changes:
* When parsing a version to time-travel to, only scalar values are allowed. This is consistent with the limitations of Databricks. None of the other target query engines support time travel by version.
* I've made the assumption that we want to parse `VERSION AS OF` syntax for all dialects that support time travel, regardless of downstream support for that syntax. (Consistent with guidance given [here](https://github.com/apache/datafusion-sqlparser-rs/pull/2134#discussion_r2631929918).)
* I've refactored `supports_timestamp_versioning` --> `supports_table_versioning`, since we now support table versioning by something other than a timestamp.

### DIFF
--- a/src/ast/helpers/stmt_create_table.rs
+++ b/src/ast/helpers/stmt_create_table.rs
@@ -504,8 +504,8 @@ impl CreateTableBuilder {
         self.require_user = require_user;
         self
     }
-    /// Consume the builder and produce a `Statement::CreateTable`.
-    pub fn build(self) -> Statement {
+    /// Consume the builder and produce a `CreateTable`.
+    pub fn build(self) -> CreateTable {
         CreateTable {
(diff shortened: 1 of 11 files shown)


JamesVorder:
@iffyio I would love a review if you have time.

PR #2156, opened 2026-01-08
Fix identifier parsing not breaking on the `|>` pipe operator

alexander-beedie:
There's a bug in Pipe SQL parsing with identifiers followed by the `|>` operator.

## Fix

`parse_identifiers` needs to break on  `Token::VerticalBarRightAngleBracket` as well, otherwise it continues to consume the following tokens as `Ident` (until one of the other break conditions is encountered).

## Example
```sql
FROM t |> DROP c |> RENAME a AS x
```
Without the fix the DROP section consumes the RENAME tokens as identifiers:
```rust
Drop { columns: [
    Ident { value: "c" },
    Ident { value: "RENAME" },
    Ident { value: "a" },
    Ident { value: "AS" },
    Ident { value: "x" }]
}
```
After the fix the identifier boundary is respected, and the operations parse correctly:
```rust
Drop { columns: [Ident { value: "c" }] }
Rename { mappings: [
    IdentWithAlias { ident: Ident { value: "a" }, alias: Ident { value: "x" } }]
}
```

## Also

* Split the monolithic test into several more manageable tests for clarity 
  (one per pipe-operator).

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -16220,702 +16220,289 @@ fn parse_set_names() {
     dialects.verified_stmt("SET NAMES UTF8 COLLATE bogus");
 }
 
-/// Returns dialect configurations for pipe operator tests:
-/// 1. Dialects supporting FROM-first syntax (e.g., "FROM users |> ...")
-/// 2. Dialects requiring SELECT-first syntax (e.g., "SELECT * FROM users |> ...")
-fn pipe_dialects() -> [(TestedDialects, bool); 2] {
-    let from_first =
-        all_dialects_where(|d| d.supports_pipe_operator() && d.supports_from_first_select());
-    let select_first =
-        all_dialects_where(|d| d.supports_pipe_operator() && !d.supports_from_first_select());
-    [(from_first, true), (select_first, false)]
-}
-
-/// Macro to test pipe operator parsing with explicit input and canonical output.
-/// Usage: `test_pipe!(ctx, input = "...", canonical = "...")`
-macro_rules! test_pipe {
-    ($ctx:expr, input = $input:expr, canonical = $canonical:expr $(,)?) => {{
-        let (ref dialects, from_first) = $ctx;
-        let prefix = if from_first {
-            "FROM tbl"
-        } else {
-            "SELECT * FROM tbl"
-        };
-        dialects.verified_query_with_canonical(
-            &format!("{prefix} |> {}", $input),
-            &format!("{prefix} |> {}", $canonical),
-        );
-    }};
-}
-
 #[test]
 fn parse_pipe_operator_as() {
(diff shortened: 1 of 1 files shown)


tests/sqlparser_common.rs
```diff
@@ -16157,302 +16157,411 @@ fn parse_set_names() {
...
-
-    // union pipe operator with multiple queries
-    dialects.verified_stmt(
-        "SELECT * FROM users |> UNION ALL (SELECT * FROM admins), (SELECT * FROM guests)",
-    );
-    dialects.verified_stmt("SELECT * FROM users |> UNION DISTINCT (SELECT * FROM admins), (SELECT * FROM guests), (SELECT * FROM employees)");
-    dialects.verified_stmt(
-        "SELECT * FROM users |> UNION (SELECT * FROM admins), (SELECT * FROM guests)",
-    );
+        // 'select' pipe operator
```
iffyio: if its possible can we use the previous format of explicitly writing out the test cases and expected results? I imagine that might be more verbose but its useful readability wise to know the cases that are being covered and what the tests do at a glance

alexander-beedie: No problem - done 😎👍 Still streamlined (and testing multiple forms), but now shows the query input and the expected canonical form explicitly.

alexander-beedie:
Rebased / addressed test format👌

tests/sqlparser_common.rs
```diff
@@ -16200,304 +16198,702 @@ fn parse_set_names() {
...
+    }};
+}
 
-    // pivot pipe operator mixed with other pipe operators
-    dialects.verified_stmt("SELECT * FROM sales_data |> WHERE year = 2023 |> PIVOT(SUM(revenue) FOR quarter IN ('Q1', 'Q2', 'Q3', 'Q4'))");
+#[test]
+fn parse_pipe_operator_as() {
+    for dialect in pipe_dialects() {
+        test_pipe!(dialect, input = "AS new_users", canonical = "AS new_users");
+    }
```
iffyio: can we drop the test_pipe macro entirely and write the test cases explicitly like `dialects.verified_stmt("SELECT * FROM users |> WHERE id = 1");`. And likely we should drop the `pipe_dialects()` method and construct the dialects explicitly too. For similar reason  mentioned previously in that its not obvious without reading a bit of custom code to figure out what the tests are doing and I don't think its worth it in this case.

alexander-beedie: Fair enough, doesn't offer much now after the previous cleanup :) Dropped; kept the improved clarity from breaking out the monolithic test block into clearer sections (is now one set of tests per pipe op), but otherwise it's back to the standard `verified_stmt` (and a `verified_query_with_canonical` or two), plus the extra test coverage for the bug 👍

PR #2157, opened 2026-01-08
fix: qualified column names with SQL keywords parse as identifiers

bombsimon:
Since v0.55.0, qualified column names using SQL keywords failed to parse. For example, `T.interval` in `PARTITION BY T.key, T.interval ORDER BY ...` was incorrectly interpreted as an INTERVAL expression instead of a column.

In `3e90a18` changed `parse_compound_expr` to use `parse_subexpr()` for tokens after `.`, causing keywords like INTERVAL, CASE, CAST, etc. to be treated as expression starters.

Explicitly handle `Token::Word` in `parse_compound_expr` by treating it as an identifier. If followed by `(` (excluding the `(+)` outer join operator), parse as a method call. This restores the original behavior where words after `.` were always converted to identifiers.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1856,41 +1856,53 @@ impl<'a> Parser<'a> {
                         chain.push(AccessExpr::Dot(expr));
                         self.advance_token(); // The consumed string
                     }
-                    // Handle words (including keywords like INTERVAL) as identifiers
-                    // when they appear after a period. This ensures `T.interval` is
-                    // parsed as a compound identifier, not as an interval expression.
-                    // If followed by `(`, parse as a method call (but not for `(+)`
-                    // which is the outer join operator in some dialects).
-                    Token::Word(w) => {
-                        let ident = w.clone().into_ident(next_token.span);
-                        self.advance_token();
-                        if self.peek_token() == Token::LParen && !self.peek_outer_join_operator() {
-                            let expr = self.parse_function(ObjectName::from(vec![ident]))?;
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -1856,6 +1856,21 @@ impl<'a> Parser<'a> {
                         chain.push(AccessExpr::Dot(expr));
                         self.advance_token(); // The consumed string
                     }
+                    // Handle words (including keywords like INTERVAL) as identifiers
```
iffyio: Can we instead of having a new clause (which duplicates some edge cases), we can extend the catch all branch to be more strict around what we're expecting. I think that makes it easier to detect if we were supposed to parse the expression as an identifier. e.g.
```rust
// Fallback ...
_  => {
	// Parse any of the suffix expressions we expect to follow the dot operator.
	let expr = self.maybe_parser(|parser| {
		let expr = parser.parse_subexpr()?;
		if matches!(
			expr,
			CompoundFieldAccess {...} |
			CompoundIdentifier(_) |
			Identifier(_) |
			Value |
			Function {..}
		) {
			Some(expr)
		}
	});

	match expr {
		// As usual
		Some(CompoundFieldAccess{..}) => {..}
		Some(CompoundIdentifier(_)) => {...}
		Some(expr) => {...}
		// If its not a suffix we expect, accept only if its an identifier.
		None => {
			chain.push(AccessExpr::Dot(Expr::Identifier(self.parse_identifier()?)));
		}
	}
}
```

PR #2158, opened 2026-01-10
Added missing `Copy` derives

LucaCappelletti94:
There were a few structs such as `TriggerObjectKind` who were missing the `Copy` derive. I have added the derive were it was missing. I have also added a couple of `Hash` derive for structs deriving `Eq` but that were missing `Hash`.

Best,
Luca


iffyio (approved):

PR #2160, opened 2026-01-10
Added missing `From` impls for `Statement` variants

LucaCappelletti94:
Since my last PR adding the missing `From` impls for `Statement` enum variants to convert any given variant `X` into `Statement::X(X)` using the `From` trait, several such `X` variants have been added without the associated `From` impl.

This PR adds several (I hope all) missing `From` impls, and reasonably cleans up some of the code which was using the more verbose conversion.

Luca

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4872,6 +4872,7 @@ impl<'a> Parser<'a> {
         let create_view_params = self.parse_create_view_params()?;
         if self.parse_keyword(Keyword::TABLE) {
             self.parse_create_table(or_replace, temporary, global, transient)
+                .map(Into::into)
         } else if self.peek_keyword(Keyword::MATERIALIZED)
             || self.peek_keyword(Keyword::VIEW)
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -5455,14 +5460,14 @@ impl<'a> Parser<'a> {
         &mut self,
         or_replace: bool,
         temporary: bool,
-    ) -> Result<Statement, ParserError> {
+    ) -> Result<CreateFunction, ParserError> {
```
iffyio, line 5463: Not sure I followed the intent of this diff - we're changing the public APIs to return not an actual statement anymore but its not clear to me why that's a desirable change?

LucaCappelletti94: I made it so all of these type of function return the strictest concrete variant, which is the thing they actually return, so that in practice one can write:

```rust
let my_specific_variant = parse_my_specific_variant(...)?
```

instead of:

```rust
let Statement::MySpecificVariant(my_specific_variant) = parse_my_specific_variant(...)? else {
    unreachable!();
};
```

which seems much cleaner.

iffyio (approved):

Issue #2099, opened 2025-11-17
Track span of DISTINCT


xitep:
Adding span information to the AST node of a `DISTINCT` will require changing its structure. As per #1548 I'm opening a ticket for this.

### Motivation

As described in #1995, or in [oracle's docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/tgsql/influencing-the-optimizer.html#GUID-B198573A-C64B-4071-9CD8-738355A20DD7) optimiser hints can be placed right after the `SELECT` (and before a possible `DISTINCT`.) (Today, parsing the comment out of the original SQL source can be done manually by consuming the characters right after the SELECT since its AST node has already a `select_token` telling the keywords ends. Ideally we'd also have the next token's start.)

xitep:
i think a more convenient approach would be to expose hints in the AST as first class citizens (just as requested in #1995). i'm closing this issue, as it seems rather non-systematic.

PR #2162, opened 2026-01-14
[MySQL, Oracle] Parse optimizer hints

### LINKED ISSUE #1995: Associate multi-line comments with their query blocks for implementing Optimizer Hints
Our database system is currently implementing an ​Optimizer Hints​ feature similar to MySQL's implementation, using the following syntax:

```sql
SELECT /*+ SET_VAR(target_partitions=1) */ * FROM 
    (SELECT /*+ SET_VAR(target_partitions=8) */ * FROM t1 LIMIT 1) AS dt;
```

In this example, we can identify two distinct query blocks:
- The inner subquery block
- The outer main query block

However, we're facing a technical challenge that the current sqlparser library:
- Automatically ignores all multi-line comments (including our hint syntax)
- Does not preserve or expose these comments in the parsed statement structure

These limitations makes us cannot reliably associate optimizer hints with their corresponding query blocks, which makes it impossible to configure query parameters for each query block solely.

xitep:
this should resolve #1995

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -14077,11 +14077,16 @@ impl<'a> Parser<'a> {
                                 Some(text) => {
                                     let hint = OptimizerHint {
                                         text: text.into(),
-                                        style: if let Whitespace::SingleLineComment { prefix, .. } = ws {
-                                            OptimizerHintStyle::SingleLine { prefix: prefix.clone() }
+                                        style: if let Whitespace::SingleLineComment {
+                                            prefix, ..
+                                        } = ws
+                                        {
+                                            OptimizerHintStyle::SingleLine {
+                                                prefix: prefix.clone(),
+                                            }
                                         } else {
                                             OptimizerHintStyle::MultiLine
-                                        }
+                                        },
                                     };
                                     // Consume the comment token
                                     self.next_token_no_skip();


iffyio:
@xitep wondering would it not make more sense to support features like these via the new `parse_sql_with_comments` API?

xitep:
surely it would be one way. (the only premise for that would to be have the next non-comment token after the corresponding keywords covered with a span, like suggested in #2099.) however, when processing such statements, these hints are something you generally don't want to leave out, even when parsing _without_ comments. hence this proposal to embed it directly into the AST.

src/parser/mod.rs
```diff
@@ -14005,6 +14015,59 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parses an optional optimizer hint at the current token position
+    ///
+    /// [MySQL](https://dev.mysql.com/doc/refman/8.4/en/optimizer-hints.html#optimizer-hints-overview)
+    /// [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Comments.html#GUID-D316D545-89E2-4D54-977F-FC97815CD62E)
+    fn parse_optional_optimizer_hint(&mut self) -> Result<Option<OptimizerHint>, ParserError> {
```
```suggestion
    fn maybe_parse_optimizer_hint(&mut self) -> Result<Option<OptimizerHint>, ParserError> {
```

src/parser/mod.rs
```diff
@@ -14005,6 +14015,59 @@ impl<'a> Parser<'a> {
...
+        let supports_multiline = dialect_of!(self is MySqlDialect | OracleDialect | GenericDialect);
+        let supports_singleline = dialect_of!(self is OracleDialect | GenericDialect);
```
iffyio: can we change this to use dialect methods? e.g. `if self.supports_comment_optimizer_hint()`

Also I think we can use the same flag and let the parser always accept either the single or multiline cases if they show up

src/parser/mod.rs
```diff
@@ -14005,6 +14015,59 @@ impl<'a> Parser<'a> {
...
+        if !supports_multiline && !supports_singleline {
+            return Ok(None);
+        }
+        loop {
+            let t = self.peek_nth_token_no_skip_ref(0);
+            match &t.token {
+                // ~ only the very first comment
```
iffyio: hmm not sure I followed, what does this comment mean by 'very first comment'?

src/parser/mod.rs
```diff
@@ -14005,6 +14015,59 @@ impl<'a> Parser<'a> {
...
+                Token::Whitespace(ws) => {
+                    match ws {
+                        Whitespace::SingleLineComment { comment, prefix } => {
+                            return Ok(if supports_singleline && comment.starts_with("+") {
+                                let text = comment.split_at(1).1.into();
+                                let prefix = prefix.clone();
+                                self.next_token_no_skip(); // ~ consume the token
```
```suggestion
                                self.next_token_no_skip(); // Consume the comment token
```

src/parser/mod.rs
```diff
@@ -14005,6 +14015,59 @@ impl<'a> Parser<'a> {
...
+                                    text,
+                                    style: OptimizerHintStyle::MultiLine,
+                                })
+                            } else {
+                                None
+                            });
+                        }
+                        // ~ but skip (pure) whitespace
+                        Whitespace::Space | Whitespace::Tab | Whitespace::Newline => {
+                            // ~ consume the token and try with the next whitespace (if any)
```
```suggestion
                        // but skip (pure) whitespace
                        Whitespace::Space | Whitespace::Tab | Whitespace::Newline => {
                            // consume the token and try with the next whitespace (if any)
```
let's avoid the `~` before the comments, since that's not a convention used in the codebase

src/parser/mod.rs
xitep: i'll revise the comment; it's that only the first comment appearing right after the SELECT, INSERT... keywords can be a valid hint.

src/parser/mod.rs
xitep: i'm not sure `-- + foo` is a valid hint for mysql (i believe it's not.) that's why i did the separation in the first place. for oracle, though, a `--+ quux` is as valid as `/*+ quux */`.

PR #2165, opened 2026-01-16
Fixed select dollar column from stage for snowflake

romanoff:
Fixed this query for snowflake:
```sql
SELECT t.$1, t.$2 FROM @mystage1(file_format => 'myformat', pattern => '.*data.*[.]csv.gz') t
```
Documentation:
https://docs.snowflake.com/en/user-guide/querying-stage#example-1-querying-columns-in-a-csv-file

Specifically `t.$1` part didn't work before

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15447,10 +15447,8 @@ impl<'a> Parser<'a> {
             && self.peek_keyword_with_tokens(Keyword::SEMANTIC_VIEW, &[Token::LParen])
         {
             self.parse_semantic_view_table_factor()
-        } else if dialect_of!(self is SnowflakeDialect)
-            && self.peek_token_ref().token == Token::AtSign
-        {
-            // Snowflake stage reference: @mystage or @namespace.stage
+        } else if self.peek_token_ref().token == Token::AtSign {
+            // Stage reference: @mystage or @namespace.stage (e.g. Snowflake)
             self.parse_snowflake_stage_table_factor()
         } else {
             let name = self.parse_object_name(true)?;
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -15199,6 +15216,35 @@ impl<'a> Parser<'a> {
...
+    /// Handles syntax like: `@mystage1 (file_format => 'myformat', pattern => '...')`
+    fn parse_snowflake_stage_table_factor(&mut self) -> Result<TableFactor, ParserError> {
+        // Parse the stage name starting with @
+        let name = crate::dialect::parse_snowflake_stage_name(self)?;
+
+        // Parse optional stage options like (file_format => 'myformat', pattern => '...')
+        let args = if self.consume_token(&Token::LParen) {
+            Some(self.parse_table_function_args()?)
+        } else {
+            None
```
iffyio, line 15561: can we add a test case that doesn't include table function args?

src/parser/mod.rs
```diff
@@ -15199,6 +15216,35 @@ impl<'a> Parser<'a> {
...
+        };
+
+        let alias = self.maybe_parse_table_alias()?;
```
iffyio, line 15564: can we add a test case that doesn't include an alias?

src/parser/mod.rs
```diff
@@ -15199,6 +15216,35 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a Snowflake stage reference as a table factor.
+    /// Handles syntax like: `@mystage1 (file_format => 'myformat', pattern => '...')`
```
iffyio, line 15550: Can we link to the documentation that supports this syntax?

src/parser/mod.rs
```diff
@@ -15103,6 +15115,11 @@ impl<'a> Parser<'a> {
             && self.peek_keyword_with_tokens(Keyword::SEMANTIC_VIEW, &[Token::LParen])
         {
             self.parse_semantic_view_table_factor()
+        } else if dialect_of!(self is SnowflakeDialect)
```
iffyio: If we need a guard for this I think we would instead use a method on the dialect trait (we're moving away from the dialect_of macro for new code). But I think the guard shouldn't be needed in the first place? so that if we see a `SELECT * FROM @foo T` we can always accept it regardless of dialect?

romanoff:
@iffyio Thank you. Updated

iffyio (approved):
LGTM! Thanks @romanoff!

PR #2168, opened 2026-01-21
Fix parsing cast operator after parenthesized `DEFAULT` expression

isaacparker0:
Fix a regression introduced in v0.59.0 by https://github.com/apache/datafusion-sqlparser-rs/pull/1927:

```
CREATE TABLE t (c text DEFAULT (foo())::text);
```

fails to parse with
```
ParserError("Expected: ',' or ')' after column definition, found: ::")
```

To fix, continue parsing infix operators after the parenthesized prefix in `parse_column_option_expr`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1818,11 +1818,16 @@ impl<'a> Parser<'a> {
                     } else if let Some(lambda) = self.try_parse_lambda()? {
                         return Ok(lambda);
                     } else {
-                        // Parentheses create a normal expression context.
-                        // This ensures that e.g. `NOT NULL` inside parens is parsed
-                        // as `IS NOT NULL` (for dialects that support it), while
-                        // `NOT NULL` outside parens in a column definition context
-                        // remains a column constraint.
+                        // Parentheses in expressions switch to "normal" parsing state.
+                        // This matters for dialects (SQLite, DuckDB) where `NOT NULL` can
+                        // be an alias for `IS NOT NULL`. In column definitions like:
+                        //
+                        //   CREATE TABLE t (c INT DEFAULT (42 NOT NULL) NOT NULL)
+                        //
+                        // The `(42 NOT NULL)` is an expression with parens, so it parses
+                        // as `IsNotNull(42)`. The trailing `NOT NULL` is outside those
+                        // expression parens (the outer parens are CREATE TABLE syntax),
+                        // so it remains a column constraint.
                         let exprs = self.with_state(ParserState::Normal, |p| {
                             p.parse_comma_separated(Parser::parse_expr)
                         })?;
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -17305,6 +17305,7 @@ fn test_parse_not_null_in_column_options() {
 #[test]
 fn test_parse_default_expr_with_operators() {
     all_dialects().verified_stmt("CREATE TABLE t (c INT DEFAULT (1 + 2) + 3)");
+    all_dialects().verified_stmt("CREATE TABLE t (c INT DEFAULT (1 + 2) + 3 NOT NULL)");
 }
 
 #[test]


isaacparker0:
@iffyio looks like you reviewed https://github.com/apache/datafusion-sqlparser-rs/pull/1927, perhaps you have the most context to review this PR?

src/parser/mod.rs
```diff
@@ -9002,7 +9002,15 @@ impl<'a> Parser<'a> {
     /// [ColumnOption::NotNull].
     fn parse_column_option_expr(&mut self) -> Result<Expr, ParserError> {
         if self.peek_token_ref().token == Token::LParen {
-            let expr: Expr = self.with_state(ParserState::Normal, |p| p.parse_prefix())?;
+            let mut expr = self.with_state(ParserState::Normal, |p| p.parse_prefix())?;
```
```suggestion
            let mut expr = self.with_state(ParserState::Normal, |p| p.parse_subexpr())?;
```
not sure I followed the intent of the added code, but I wonder would the issue be fixed by changing this line instead (thinking since `(foo())::INT` should be parsed as an expression), or are there other considerations?

isaacparker0: That change would fix the bug here, but break the behavior added in https://github.com/apache/datafusion-sqlparser-rs/pull/1927. In Normal state, `NOT NULL` is treated as an expression (i.e., `IS NOT NULL`), but in ColumnDefinition state it's not, allowing the trailing `NOT NULL` to be parsed as a column constraint instead.

My proposed change is basically just duplicating part of the existing logic used in `parse_subexpr`[^1]. If we run `parse_prefix` in normal state but the infix loop in ColumnDefinition state, we maintain the NOT NULL parsing behavior added in the previous PR but still properly handle infix operators like `::TEXT`.

I'm still familiarizing myself with this repo, so there may be a better way to do this, but this is what I've come up with so far and why it seems like the right approach to me.

Also happy to do a bit of refactoring to cleanly extract a single shared logic for the part of `parse_subexpr` that we are duplicating if that seems preferable.

[^1]: https://github.com/apache/datafusion-sqlparser-rs/blob/62cf16f3ece6f3d5985e35893407c8db359ffd3f/src/parser/mod.rs#L1337-L1363

iffyio: could we do something like this instead?
```rust
let expr = self.with_state(ParserState::Normal, |p| p.parse_subexpr(self.dialect.prec_value(Precedence::DoubleColon)))?;
if self.consume_token(&Token::DoubleColon) {
	Ok(Expr::Cast {
    	kind: CastKind::DoubleColon,
        expr: Box::new(expr),
        data_type: self.parse_data_type()?,
        format: None,
    })
} else {
	Ok(expr)
}
```

tests/sqlparser_postgres.rs
```diff
@@ -512,6 +512,12 @@ fn parse_create_table_with_defaults() {
     }
 }
 
+#[test]
+fn parse_default_expr_parenthesized_then_cast() {
```
```suggestion
fn parse_cast_operator_in_column_option() {
```

tests/sqlparser_postgres.rs
```diff
@@ -512,6 +512,12 @@ fn parse_create_table_with_defaults() {
...
+    // Infix operators like :: after parenthesized expression in DEFAULT
+    pg().verified_stmt("CREATE TABLE t (c TEXT DEFAULT (foo())::TEXT)");
```
iffyio, line 517:
```suggestion
    pg().verified_stmt("CREATE TABLE t (c TEXT DEFAULT (foo())::TEXT NOT NULL)");
```
to test the behavior with the not null clause

isaacparker0: Added a separate case for this.

src/parser/mod.rs
isaacparker0: This approach would only handle the `::` case. There could be other infix operators or chained casts, which my current approach handles properly by leveraging the full existing infix loop.

I added additional test cases to document this behavior.

tests/sqlparser_postgres.rs
isaacparker0: Renamed test function to be more accurate after adding new test cases to it.

src/parser/mod.rs
iffyio: to support the chained cast, we would consume the double colon in a loop? I think the current approach duplicates parts of the expr parsing code most of which aren't exactly related to the problem being solved for, which makes it not ideal

isaacparker0: What about a non `::` operator, such as `CREATE TABLE t (c INT DEFAULT (foo()) + 1)`? (I just added a test case for this)

Admittedly, we are getting into the territory of obscure but technically legal postgres syntax, but it still seems like we should solve this in a more general way than hardcoding one specific operator.

iffyio: > What about a non :: operator, such as CREATE TABLE t (c INT DEFAULT (foo()) + 1)?

the precedence operator should handle this case?

iffyio: Im thinking a generic version would need to use sub_expr somehow, ideally we find a way to reuse the core expr parsing loop

isaacparker0: Thanks for pushing back on this. I thought about this some more, and I think I came up with a more principled approach that also simplifies the solution from https://github.com/apache/datafusion-sqlparser-rs/pull/1927: if we just change `parse_prefix()` for LParen to switch to `ParserState::Normal`, it allows us to remove the special-cased `parse_column_option_expr` added in https://github.com/apache/datafusion-sqlparser-rs/pull/1927, while also fixing the regression with `:: cast`.

Let me know what you think of this approach.

tests/sqlparser_common.rs
```diff
@@ -17376,6 +17302,11 @@ fn test_parse_not_null_in_column_options() {
     );
 }
 
+#[test]
+fn test_parse_default_expr_with_operators() {
+    all_dialects().verified_stmt("CREATE TABLE t (c INT DEFAULT (1 + 2) + 3)");
+}
```
isaacparker0, line 17309: Also added a common test case here, as I realized this is the more fundamental regression fixed by my proposed change (not only the postgres-specific `::` cast).

src/parser/mod.rs
iffyio: That looks good to me thanks!

tests/sqlparser_common.rs
iffyio: Can we add or extend this to include the `NOT NULL` behavior? e.g.
```sql
CREATE TABLE t (c INT DEFAULT (1 + 2) + 3 NOT NULL)
```

src/parser/mod.rs
```diff
@@ -1822,7 +1818,14 @@ impl<'a> Parser<'a> {
                     } else if let Some(lambda) = self.try_parse_lambda()? {
                         return Ok(lambda);
                     } else {
-                        let exprs = self.parse_comma_separated(Parser::parse_expr)?;
+                        // Parentheses create a normal expression context.
```
iffyio: Can we incorporate this documentation from the previous function here? I think that would be useful context to readers later on since this part of the code might be a bit too abstract otherwise

```text
    /// When parsing some column option expressions we need to revert to [ParserState::Normal] since
    /// `NOT NULL` is allowed as an alias for `IS NOT NULL`.
    /// In those cases we use this helper instead of calling [Parser::parse_expr] directly.
    ///
    /// For example, consider these `CREATE TABLE` statements:
    /// ```sql
    /// CREATE TABLE foo (abc BOOL DEFAULT (42 NOT NULL) NOT NULL);
    /// ```
    /// vs
    /// ```sql
    /// CREATE TABLE foo (abc BOOL NOT NULL);
    /// ```
    ///
    /// In the first we should parse the inner portion of `(42 NOT NULL)` as [Expr::IsNotNull],
    /// whereas is both statements that trailing `NOT NULL` should only be parsed as a
    /// [ColumnOption::NotNull].
```

tests/sqlparser_common.rs
isaacparker0: Added

src/parser/mod.rs
isaacparker0: Rewrote the comment to include the context from the deleted function's comment.

PR #2169, opened 2026-01-21
PostgreSQL: Support force row level security

isaacparker0:
Support parsing `FORCE ROW LEVEL SECURITY` and `NO FORCE ROW LEVEL SECURITY` statements in postgres. Reference for these statements can be seen here: https://www.postgresql.org/docs/current/sql-altertable.html.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -192,6 +192,7 @@ pub enum AlterTableOperation {
     /// `DISABLE ROW LEVEL SECURITY`
     ///
     /// Note: this is a PostgreSQL-specific operation.
+    /// Please refer to [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-altertable.html)
     DisableRowLevelSecurity,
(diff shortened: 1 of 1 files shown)


src/ast/ddl.rs
```diff
@@ -319,6 +319,14 @@ pub enum AlterTableOperation {
...
     /// Note: this is a PostgreSQL-specific operation.
     EnableRowLevelSecurity,
+    /// `FORCE ROW LEVEL SECURITY`
+    ///
+    /// Note: this is a PostgreSQL-specific operation.
+    ForceRowLevelSecurity,
+    /// `NO FORCE ROW LEVEL SECURITY`
+    ///
+    /// Note: this is a PostgreSQL-specific operation.
+    NoForceRowLevelSecurity,
```
iffyio, line 333: could we add a link to the postgres documentation for the new options?

isaacparker0: Done, added comments identical to the existing one for `REPLICA IDENTITY`.

PR #2172, opened 2026-01-23
MySQL: Add support for `SELECT` modifiers

mvzink:
Adds support for MySQL-specific `SELECT` modifiers that appear after the `SELECT` keyword. Grammar from the [docs]:

```sql
SELECT
    [ALL | DISTINCT | DISTINCTROW ]
    [HIGH_PRIORITY]
    [STRAIGHT_JOIN]
    [SQL_SMALL_RESULT] [SQL_BIG_RESULT] [SQL_BUFFER_RESULT]
    [SQL_NO_CACHE] [SQL_CALC_FOUND_ROWS]
    select_expr [, select_expr] ...
```

Manual testing shows that these options can appear in any order relative to each other, so for the sake of fidelity, we parse this separately from how we parse distinct and other options for other dialects, in a new `Parser::parse_select_modifiers` method.

`DISTINCTROW` is a legacy (but not deprecated) synonym for `DISTINCT`, so it just gets canonicalized as `DISTINCT`.

[docs]: https://dev.mysql.com/doc/refman/8.4/en/select.html


tests/sqlparser_common.rs
```diff
@@ -1040,18 +1041,18 @@ fn parse_outer_join_operator() {
 #[test]
 fn parse_select_distinct_on() {
     let sql = "SELECT DISTINCT ON (album_id) name FROM track ORDER BY album_id, milliseconds";
-    let select = verified_only_select(sql);
+    let select = all_dialects_but_mysql().verified_only_select(sql);
```
```suggestion
    let select = all_dialects_except(|d| d.is::<MySqlDialect>()).verified_only_select(sql);
```
thinking instead of introducing a dedicated function we can reuse this helper?

src/parser/mod.rs
```diff
@@ -14032,6 +14050,79 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parses SELECT modifiers and DISTINCT/ALL in any order. Allows HIGH_PRIORITY, STRAIGHT_JOIN,
+    /// SQL_SMALL_RESULT, SQL_BIG_RESULT, SQL_BUFFER_RESULT, SQL_NO_CACHE, SQL_CALC_FOUND_ROWS and
+    /// DISTINCT/DISTINCTROW/ALL to appear in any order.
```
```suggestion
    /// Parses `SELECT` modifiers and `DISTINCT/ALL` in any order. 
```
Thinking we can avoid enumerating the variants so that they don't go out of sync with the impl

src/parser/mod.rs
```diff
@@ -4906,14 +4906,17 @@ impl<'a> Parser<'a> {
...
         let all = self.parse_keyword(Keyword::ALL);
         let distinct = self.parse_keyword(Keyword::DISTINCT);
         if !distinct {
             return Ok(None);
         }
         if all {
-            return parser_err!("Cannot specify both ALL and DISTINCT".to_string(), loc);
+            self.prev_token();
+            return self.expected(
+                "ALL alone without DISTINCT or DISTINCTROW",
```
iffyio: this diff seems incorrect to mention `DISTINCTROW`?

src/parser/mod.rs
```diff
@@ -14032,6 +14050,79 @@ impl<'a> Parser<'a> {
...
 
+    /// Parses SELECT modifiers and DISTINCT/ALL in any order. Allows HIGH_PRIORITY, STRAIGHT_JOIN,
+    /// SQL_SMALL_RESULT, SQL_BIG_RESULT, SQL_BUFFER_RESULT, SQL_NO_CACHE, SQL_CALC_FOUND_ROWS and
+    /// DISTINCT/DISTINCTROW/ALL to appear in any order.
+    fn parse_select_modifiers(
+        &mut self,
+    ) -> Result<(SelectModifiers, Option<Distinct>), ParserError> {
+        let mut modifiers = SelectModifiers::default();
+        let mut distinct: Option<Distinct> = None;
+        let mut has_all = false;
```
iffyio: Can we instead introduce a proper enum to represent the three states? the current impl I think is harder to follow which combinations are valid and if/when the specified option is ignored in the output.

src/parser/mod.rs
```diff
@@ -14032,6 +14050,79 @@ impl<'a> Parser<'a> {
...
+                    if has_all {
+                        self.prev_token();
+                        return self.expected(
+                            "ALL alone without DISTINCT or DISTINCTROW",
+                            self.peek_token(),
+                        );
+                    }
+                    distinct = Some(Distinct::Distinct);
+                }
+                Keyword::HIGH_PRIORITY => modifiers.high_priority = true,
```
iffyio, line 14184:
```suggestion
                Keyword::HIGH_PRIORITY if !modifiers.high_priority => modifiers.high_priority = true,
```
similar for the others is something like this needed if it's invalid that they occur multiple times (can we add test cases for the behavior)?

src/ast/query.rs
```diff
@@ -345,6 +409,10 @@ pub struct Select {
     pub select_token: AttachedToken,
     /// `SELECT [DISTINCT] ...`
     pub distinct: Option<Distinct>,
+    /// MySQL-specific SELECT modifiers.
+    ///
+    /// See [MySQL SELECT](https://dev.mysql.com/doc/refman/8.4/en/select.html).
+    pub select_modifiers: SelectModifiers,
```
```suggestion
    pub select_modifiers: Option<SelectModifiers>,
```

src/ast/query.rs
```diff
@@ -334,6 +334,70 @@ pub enum SelectFlavor {
...
+///     [SQL_NO_CACHE] [SQL_CALC_FOUND_ROWS]
+///     select_expr [, select_expr] ...
+/// ```
+///
+/// See [MySQL SELECT](https://dev.mysql.com/doc/refman/8.4/en/select.html).
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct SelectModifiers {
+    /// `HIGH_PRIORITY` gives the SELECT higher priority than statements that update a table.
```
iffyio, line 358: could we include the mysql links individually on the fields as well? that way if other dialects add their own variants in the future the doc would be consistent

src/parser/mod.rs
mvzink: Not sure what you mean. Are you saying I should split out adding DISTINCTROW as an alias for DISTINCT into a separate PR?

src/parser/mod.rs
mvzink: Yeah, sure. Is this what you have in mind? 

```rust
enum Distinct {
    All,
    Distinct,
    On(Vec<Expr),
}
```

I'll go ahead and implement that, but let me know if you had something else in mind that I'll be happy to update it..

src/parser/mod.rs
mvzink: They are all allowed multiple times, but that took manual testing to find out. I'll add a test and a comment.

src/parser/mod.rs
mvzink: lmk what you think. I also reworked it a bit so that `DISTINCTROW` aliasing only happens here and not in the main `parse_all_or_distinct` which I believe addresses your other comment.

src/parser/mod.rs
mvzink: Per [other comment](https://github.com/apache/datafusion-sqlparser-rs/pull/2172#discussion_r2766466843): this aliasing now only happens in the MySQL specific subparser.

iffyio (approved):
LGTM! Thanks @mvzink!

PR #2171, opened 2026-01-23
Refactor: replace some `dialect_of!` checks with `Dialect` trait methods

andygrove:
## Summary

This PR replaces 6 `dialect_of!` macro invocations in the parser with dedicated `Dialect` trait methods. This improves the separation between generic parsing logic and dialect-specific behavior, making it easier for custom dialects to override behavior without modifying the parser.

## Motivation

The codebase currently has ~134 `dialect_of!` macro calls scattered throughout the parser. The [dialect module documentation](https://github.com/apache/datafusion-sqlparser-rs/blob/845e2130e908eb4e8e8ec7fab5c9f163cd2fff2c/src/dialect/mod.rs#L64-L71) already recommends using trait methods instead:

> Note: when possible please the new style, adding a method to the `Dialect` trait rather than using this macro.
>
> The benefits of adding a method on `Dialect` over this macro are:
> 1. user defined `Dialect`s can customize the parsing behavior
> 2. The differences between dialects can be clearly documented in the trait

This PR is a proof-of-concept to gauge community interest before doing more of these refactorings.

## Changes

New trait methods added to `Dialect`:

| Method | Dialects | Description |
|--------|----------|-------------|
| `supports_extract_comma_syntax()` | Snowflake, Generic | `EXTRACT(YEAR, d)` comma syntax |
| `supports_subquery_as_function_arg()` | Snowflake | `FLATTEN(SELECT ...)` without extra parens |
| `supports_create_view_comment_syntax()` | Snowflake, Generic | `CREATE VIEW ... COMMENT = '...'` |
| `supports_array_typedef_without_element_type()` | Snowflake | `ARRAY` type without element specification |
| `supports_parens_around_table_factor()` | Snowflake, Generic | `FROM (mytable)` extra parens |
| `supports_values_as_table_factor()` | Snowflake, Databricks, Generic | `FROM VALUES (1,'a')` without parens |

Each method includes documentation with examples and links to dialect documentation.

## Test plan

- [x] All existing tests pass
- [x] `cargo fmt` and `cargo clippy` pass
- No new tests needed as this is a refactoring of existing behavior

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #2174, opened 2026-01-23
Streamlined derivation of new `Dialect` objects

alexander-beedie:
`derive_dialect!(new, base, overrides = {…}, preserve_type_id = …)`

## Preamble

There have been a number of requests https://github.com/apache/datafusion-sqlparser-rs/issues/1430 to make `Dialect` derivation easier but there have remained some practical issues around implementation.

This PR represents an attempt at a streamlined option that covers the majority of what people want to do with _no_ impact on existing code, while also remaining largely future-proofed. It is complementary to, and compatible with, longer-term efforts to move away from internal calls to `dialect_of!` (such as https://github.com/apache/datafusion-sqlparser-rs/pull/2171).

## Implementation

Introduces a `derive_dialect!` proc macro in `sqlparser_derive`, gated behind the new `"derive-dialect"` feature. If you don't enable the feature, nothing changes - the _sole_ impact on the rest of the codebase would be that each defined `Dialect` is marked with `#[derive(Debug, Default)]` instead of just `#[derive(Debug)]`, which should be zero-cost for unit/empty structs.

In use, the macro discovers all the boolean traits of the base `Dialect` at compile time (plus the `identifier_quote_style` char definition) and allows them to be selectively overridden; new boolean traits added in future releases would be automatically discovered/included, with non-overridden traits deferring to the base.

The code in `sqlparser_derive` was split into "visit.rs" (existing visitor macros) and "dialect.rs" (the new dialect derivation macro/helpers) for clarity.

## In use

As well as the desired boolean overrides, you can set `preserve_type_id=true` to have the derived dialect be recognised as equivalent to the base, allowing the derived dialect to pass equivalence-checks for the remaining internal `dialect_of!` calls. This makes it easy to support use-cases like that of @samuelcolvin in https://github.com/apache/datafusion-sqlparser-rs/issues/1430, which was _"I want to use 'from first' syntax with otherwise postgres syntax"_ (example below).

If/when internal use of `dialect_of!` is phased out, this macro would remain a nice streamlined way of deriving new `Dialect` objects (so if "preserve_type_id" loses meaning, the utility of the macro remains, and we could either phase out that param or leave as-is).

#### Example:
```rust
use sqlparser::derive_dialect;
use sqlparser::dialect::{Dialect, PostgreSqlDialect};

derive_dialect!(EnhancedPostgreSqlDialect, PostgreSqlDialect,
    overrides = {
        supports_order_by_all = true,
        supports_from_first_select = true,
        supports_triple_quoted_string = true,
    },
    preserve_type_id = true,
);
```
The derived dialect has the requested trait values...
```rust
let dialect = EnhancedPostgreSqlDialect::new();

assert!(dialect.supports_order_by_all());          // → true
assert!(dialect.supports_from_first_select());     // → true
assert!(dialect.supports_triple_quoted_string());  // → true
```
...but still identifies internally as a valid `PostgreSqlDialect`:
```rust
let d: &dyn Dialect = &dialect;
assert!(d.is::<PostgreSqlDialect>());  // → true
```

## References

There are a few others (closed/open), but these cover the key points:

* https://github.com/apache/datafusion-sqlparser-rs/pull/1448/
* https://github.com/apache/datafusion-sqlparser-rs/issues/1430
* https://github.com/apache/datafusion-sqlparser-rs/pull/2171

### DIFF
--- a/src/ast/dcl.rs
+++ b/src/ast/dcl.rs
@@ -29,7 +29,10 @@ use serde::{Deserialize, Serialize};
 use sqlparser_derive::{Visit, VisitMut};
 
 use super::{display_comma_separated, Expr, Ident, Password, Spanned};
-use crate::ast::{display_separated, ObjectName};
+use crate::ast::{
+    display_separated, CascadeOption, CurrentGrantsKind, GrantObjects, Grantee, ObjectName,
+    Privileges,
+};
 use crate::tokenizer::Span;
 
 /// An option in `ROLE` statement.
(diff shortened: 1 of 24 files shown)


alexander-beedie:
> LGTM! Thanks @alexander-beedie!

Great, thanks :)

PR #2182, opened 2026-01-23
MySQL: Add support for DEFAULT CHARACTER SET in CREATE DATABASE

mvzink:
Parse MySQL-style [DEFAULT] CHARACTER SET and [DEFAULT] COLLATE options in CREATE DATABASE statements. This adds two new fields to CreateDatabase: default_charset and default_collation.

Supports the following syntax variants:
- DEFAULT CHARACTER SET [=] charset_name
- CHARACTER SET [=] charset_name  
- DEFAULT CHARSET [=] charset_name
- CHARSET [=] charset_name
- DEFAULT COLLATE [=] collation_name
- COLLATE [=] collation_name

### DIFF
--- a/src/ast/helpers/stmt_create_database.rs
+++ b/src/ast/helpers/stmt_create_database.rs
@@ -85,6 +85,14 @@ pub struct CreateDatabaseBuilder {
     pub storage_serialization_policy: Option<StorageSerializationPolicy>,
     /// Optional comment attached to the database.
     pub comment: Option<String>,
+    /// Optional default character set (MySQL).
+    ///
+    /// <https://dev.mysql.com/doc/refman/8.4/en/create-database.html>
+    pub default_charset: Option<String>,
+    /// Optional default collation (MySQL).
+    ///
+    /// <https://dev.mysql.com/doc/refman/8.4/en/create-database.html>
+    pub default_collation: Option<String>,
     /// Optional catalog sync configuration.
     pub catalog_sync: Option<String>,
     /// Optional catalog sync namespace mode.
@@ -120,6 +128,8 @@ impl CreateDatabaseBuilder {
             default_ddl_collation: None,
             storage_serialization_policy: None,
             comment: None,
+            default_charset: None,
+            default_collation: None,
             catalog_sync: None,
             catalog_sync_namespace_mode: None,
             catalog_sync_namespace_flatten_delimiter: None,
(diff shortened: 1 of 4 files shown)


src/ast/helpers/stmt_create_database.rs
```diff
@@ -85,6 +85,10 @@ pub struct CreateDatabaseBuilder {
     pub storage_serialization_policy: Option<StorageSerializationPolicy>,
     /// Optional comment attached to the database.
     pub comment: Option<String>,
+    /// Optional default character set (MySQL).
```
iffyio, line 88: can we include a link to the mysql documentation for the doc comments?

src/parser/mod.rs
```diff
@@ -5309,6 +5311,26 @@ impl<'a> Parser<'a> {
...
+        // Parse MySQL-style [DEFAULT] CHARACTER SET and [DEFAULT] COLLATE options
+        loop {
+            let has_default = self.parse_keyword(Keyword::DEFAULT);
+            if self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET])
+                || self.parse_keyword(Keyword::CHARSET)
+            {
+                self.expect_token(&Token::Eq).ok();
+                default_charset = Some(self.parse_identifier()?.value);
+            } else if self.parse_keyword(Keyword::COLLATE) {
+                self.expect_token(&Token::Eq).ok();
```
iffyio: can we handle the errors in the calls to `expect_token`?

src/parser/mod.rs
```diff
@@ -5309,6 +5311,26 @@ impl<'a> Parser<'a> {
             None
         };
 
+        // Parse MySQL-style [DEFAULT] CHARACTER SET and [DEFAULT] COLLATE options
```
iffyio, line 5333: Can we move the `default_charset` and `default_collation` closer here to the loop? (misread for a second wondering if the values were being previously set elsewhere before the loop)

src/parser/mod.rs
```diff
@@ -5309,6 +5311,26 @@ impl<'a> Parser<'a> {
...
+        loop {
+            let has_default = self.parse_keyword(Keyword::DEFAULT);
+            if self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET])
```
```suggestion
            if default_charset.is_none() && self.parse_keywords(&[Keyword::CHARACTER, Keyword::SET])
```

src/parser/mod.rs
```diff
@@ -5309,6 +5311,26 @@ impl<'a> Parser<'a> {
...
+                || self.parse_keyword(Keyword::CHARSET)
+            {
+                self.expect_token(&Token::Eq).ok();
+                default_charset = Some(self.parse_identifier()?.value);
+            } else if self.parse_keyword(Keyword::COLLATE) {
```
iffyio, line 5349:
```suggestion
            } else if default_collation.is_none() && self.parse_keyword(Keyword::COLLATE) {
```

src/parser/mod.rs
mvzink: I don't know why I used expect_token instead of just consume_token...

src/parser/mod.rs
mvzink: Actually, MySQL accepts multiple collations and uses the last one. I added a comment reflecting that, but fix the char set condition.

PR #2183, opened 2026-01-23
MySQL: Allow optional constraint name after CONSTRAINT keyword

mvzink:
MySQL allows `CONSTRAINT CHECK (expr)` without a name - the database will auto-generate a constraint name. This is a MySQL extension; the SQL standard requires a name after the CONSTRAINT keyword. See [docs].

Previously, we required an identifier after CONSTRAINT, causing us to interpret CONSTRAINT as a column name in cases like:

```sql
CREATE TABLE t (x INT, CONSTRAINT CHECK (x > 1))
```

Now we check if the token after CONSTRAINT is a constraint type keyword (CHECK, PRIMARY, UNIQUE, FOREIGN) and treat the name as optional for dialects that support this. This is just MySQL as far as I know (and Generic), but we introduce a new `Dialect` flag even though it's a minor feature. We could remove the flag and allow the more permissive optional name syntax across dialects if desired.

[docs]: https://dev.mysql.com/doc/refman/8.4/en/create-table.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9243,7 +9243,20 @@ impl<'a> Parser<'a> {
         &mut self,
     ) -> Result<Option<TableConstraint>, ParserError> {
         let name = if self.parse_keyword(Keyword::CONSTRAINT) {
-            Some(self.parse_identifier()?)
+            if self.dialect.supports_constraint_keyword_without_name()
+                && self
+                    .peek_one_of_keywords(&[
+                        Keyword::CHECK,
+                        Keyword::PRIMARY,
+                        Keyword::UNIQUE,
+                        Keyword::FOREIGN,
+                    ])
+                    .is_some()
+            {
+                None
+            } else {
+                Some(self.parse_identifier()?)
+            }
         } else {
             None
         };
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -3463,6 +3463,27 @@ fn parse_create_table_unallow_constraint_then_index() {
 }
 
 #[test]
+fn parse_create_table_constraint_check_without_name() {
+    let dialects = all_dialects_where(|d| d.supports_constraint_keyword_without_name());
+    dialects.one_statement_parses_to(
+        "CREATE TABLE t (x INT, CONSTRAINT PRIMARY KEY (x))",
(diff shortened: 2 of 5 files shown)


src/parser/mod.rs
```diff
@@ -9252,7 +9252,16 @@ impl<'a> Parser<'a> {
         &mut self,
     ) -> Result<Option<TableConstraint>, ParserError> {
         let name = if self.parse_keyword(Keyword::CONSTRAINT) {
-            Some(self.parse_identifier()?)
+            if self.dialect.supports_constraint_keyword_without_name()
+                && (self.peek_keyword(Keyword::CHECK)
+                    || self.peek_keyword(Keyword::PRIMARY)
+                    || self.peek_keyword(Keyword::UNIQUE)
+                    || self.peek_keyword(Keyword::FOREIGN))
```
iffyio: can we use `self.peek_one_of_keywords()` here?

tests/sqlparser_mysql.rs
```diff
@@ -3457,6 +3457,13 @@ fn parse_create_table_unallow_constraint_then_index() {
     assert!(mysql_and_generic().parse_sql_statements(sql).is_ok());
 }
 
+#[test]
+fn parse_create_table_constraint_check_without_name() {
+    let sql = "CREATE TABLE t (x INT, CONSTRAINT CHECK (x > 1))";
+    let normalized = "CREATE TABLE t (x INT, CHECK (x > 1))";
+    mysql_and_generic().one_statement_parses_to(sql, normalized);
```
iffyio: can we use `all_dialects_where()` to select the dialects?

tests/sqlparser_mysql.rs
```diff
@@ -3457,6 +3457,13 @@ fn parse_create_table_unallow_constraint_then_index() {
     assert!(mysql_and_generic().parse_sql_statements(sql).is_ok());
 }
 
+#[test]
+fn parse_create_table_constraint_check_without_name() {
+    let sql = "CREATE TABLE t (x INT, CONSTRAINT CHECK (x > 1))";
```
iffyio: can we also add cases for primary/unique/foreign ?

iffyio (approved):
LGTM! Thanks @mvzink!

PR #2184, opened 2026-01-24
Add Tokenizer custom token mapper support

askalt:
This patch adds a method to map tokens with provided mapper during
tokenization. This way tokens could be replaced without an additional
pass.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -935,6 +935,16 @@ impl<'a> Tokenizer<'a> {
         &mut self,
         buf: &mut Vec<TokenWithSpan>,
     ) -> Result<(), TokenizerError> {
+        self.tokenize_with_location_into_buf_with_mapper(buf, |token| token)
+    }
+
+    /// Tokenize the statement and produce a vector of tokens, mapping each token
+    /// with provided `mapper`
+    pub fn tokenize_with_location_into_buf_with_mapper(
+        &mut self,
+        buf: &mut Vec<TokenWithSpan>,
+        mut mapper: impl FnMut(TokenWithSpan) -> TokenWithSpan,
+    ) -> Result<(), TokenizerError> {
         let mut state = State {
             peekable: self.query.chars().peekable(),
             line: 1,
@@ -952,10 +962,10 @@ impl<'a> Tokenizer<'a> {
                         && comment.starts_with('!') =>
                 {
                     // Re-tokenize the hints and add them to the buffer
-                    self.tokenize_comment_hints(comment, span, buf)?;
+                    self.tokenize_comment_hints(comment, span, buf, &mut mapper)?;
                 }
                 _ => {
-                    buf.push(TokenWithSpan { token, span });
+                    buf.push(mapper(TokenWithSpan { token, span }));
                 }
             }
 
@@ -971,6 +981,7 @@ impl<'a> Tokenizer<'a> {
         comment: &str,
         span: Span,
         buf: &mut Vec<TokenWithSpan>,
+        mut mapper: impl FnMut(TokenWithSpan) -> TokenWithSpan,
     ) -> Result<(), TokenizerError> {
         // Strip the leading '!' and any version digits (e.g., "50110")
         let hint_content = comment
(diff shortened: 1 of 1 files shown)


src/tokenizer.rs
```diff
@@ -917,6 +937,23 @@ impl<'a> Tokenizer<'a> {
         self
     }
 
+    /// Return an iterator over tokens
+    pub fn iter(&mut self) -> TokenWithLocationIter<'a, '_> {
```
novartole: 1. `iter` doesn't reflect what the method does and sounds too general.
2. `TokenWithLocationIter` returns `TokenWithSpan` rather instead something with `Location`.

How about giving them more specific names?

```suggestion
    fn token_with_span_iter(&mut self) -> TokenWithSpanIter<'a, '_> {
```

src/tokenizer.rs
```diff
@@ -2299,6 +2325,34 @@ impl<'a> Tokenizer<'a> {
...
+    type Item = Result<TokenWithSpan, TokenizerError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let token = match self
+            .tokenizer
+            .next_token(&mut self.state, self.prev_token_kind)
+        {
+            Err(err) => return Some(Err(err)),
+            Ok(None) => return None,
+            Ok(Some(token)) => token,
```
novartole: BTW, `transpose` seems to play well here:

```suggestion
            .next_token(&mut self.state, self.prev_token_kind)
            .transpose()?
        {
            Err(err) => return Some(Err(err)),
            Ok(token) => token,
```

src/tokenizer.rs
```diff
@@ -2577,6 +2631,39 @@ mod tests {
         compare(expected, tokens);
     }
 
+    #[test]
+    fn tokenize_iterator_map() {
+        let sql = String::from("SELECT $1");
```
novartole: I guess you mean the following:

```suggestion
        let sql = String::from("SELECT ?");
```

src/tokenizer.rs
askalt: 1. General name is used because I cannot imagine why an other iterator method could be needed for tokenizer. So when I call `tokenizer.iter()` I expect that it will iterate over tokens. So, IMO the API should be named as simple as possible if it reflects semantic good enough. If someone need raw Token iterator, it could `.map(|t| t.token)` an existing one.
2. Renamed to `TokenWithSpanIter`.

src/tokenizer.rs
askalt: Applied.

src/tokenizer.rs
askalt: yes, sure, applied.

src/tokenizer.rs
novartole: Got your idea. Thanks!

novartole (approved):
LGTM

askalt:
@iffyio Could you review please?

src/tokenizer.rs
```diff
@@ -856,6 +856,26 @@ pub struct Tokenizer<'a> {
     unescape: bool,
 }
 
+/// Passed into [`Tokenizer::next_token`] as in some situations tokenization
+/// is context dependent. The separate enum is used to be able to not clone
+/// the previous token during [`TokenWithLocationIter`] iteration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PrevTokenKind {
```
iffyio: I'm not so sure about introducing this enum if its necessary? feels like we would be stuck in the tokenizer if some code later required access to the actual previous token (as opposed to its 'type')

askalt: Did it to avoid token cloning, as it may be noticeable when a long SQL like `insert into values ... many strings` is parsed.

askalt: It seems that the idea of "context-depending tokenizer without token cloning" is not compatible with the possibility to have an iterator over owned tokens. I'll try to consider another solution to the problem patch solves.

askalt: Removed an iterator, added a simple method allowing to pass custom `mapper` for tokens.

iffyio:
@askalt could you take a look at the conflicts on the branch when you get some time?

askalt:
> @askalt could you take a look at the conflicts on the branch when you get some time?

Yep, done, thank you for review!

PR #2185, opened 2026-01-30
[Oracle] Support hierarchical queries

xitep:
* this is merely building on top of already existing "connect by" functionality for snowflake :)
* and making the `start with` clause optional [as per docs](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Hierarchical-Queries.html) and encountered in the wild

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -12926,7 +12926,8 @@ fn parse_connect_by() {
         "SELECT 0 FROM t WHERE 1 = 1 START WITH 'a' = 'a' CONNECT BY 2 = 2 GROUP BY 3",
     );
     dialects.verified_only_select(
-        "SELECT 0 FROM t WHERE 1 = 1 CONNECT BY 2 = 2 START WITH 'a' = 'a' GROUP BY 3");
+        "SELECT 0 FROM t WHERE 1 = 1 CONNECT BY 2 = 2 START WITH 'a' = 'a' GROUP BY 3",
+    );
 }
 
 #[test]


src/ast/query.rs
```diff
@@ -374,6 +374,8 @@ pub struct Select {
     pub prewhere: Option<Expr>,
     /// WHERE
     pub selection: Option<Expr>,
+    /// STARTING WITH .. CONNECT BY
```
xitep:
```suggestion
    /// START WITH .. CONNECT BY
```

src/ast/query.rs
```diff
@@ -475,6 +475,10 @@ impl fmt::Display for Select {
             SpaceOrNewline.fmt(f)?;
             Indent(selection).fmt(f)?;
         }
+        if let Some(ref connect_by) = self.connect_by {
+            SpaceOrNewline.fmt(f)?;
+            connect_by.fmt(f)?;
+        }
```
xitep, line 591: i can't verify this change for snowflake, but [this doc](https://docs.snowflake.com/en/sql-reference/constructs) indicates to me that it's ok

src/ast/query.rs
```diff
@@ -374,6 +374,8 @@ pub struct Select {
     pub prewhere: Option<Expr>,
     /// WHERE
     pub selection: Option<Expr>,
+    /// START WITH .. CONNECT BY
```
xitep:
```suggestion
    /// [START WITH ..] CONNECT BY ..
```

src/parser/mod.rs
```diff
@@ -14000,6 +14000,17 @@ impl<'a> Parser<'a> {
             None
         };
 
+        let connect_by = if self.dialect.supports_connect_by()
+            && self
+                .parse_one_of_keywords(&[Keyword::START, Keyword::CONNECT])
+                .is_some()
+        {
+            self.prev_token();
```
```suggestion
            && self
                .peek_one_of_keywords(&[Keyword::START, Keyword::CONNECT])
                .is_some()
        {
```
would this be equivalent?

src/ast/query.rs
```diff
@@ -374,6 +374,8 @@ pub struct Select {
     pub prewhere: Option<Expr>,
     /// WHERE
     pub selection: Option<Expr>,
+    /// [START WITH ..] CONNECT BY ..
+    pub connect_by: Option<ConnectBy>,
```
iffyio: Can we add a link to the docs where this syntax comes from?

src/ast/query.rs
```diff
@@ -1094,23 +1094,34 @@ impl fmt::Display for TableWithJoins {
...
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct ConnectBy {
     /// START WITH
-    pub condition: Expr,
+    pub condition: Option<Expr>,
     /// CONNECT BY
     pub relationships: Vec<Expr>,
+    /// [CONNECT BY] NOCYCLE
+    pub nocycle: bool,
```
iffyio: Similarly, can we add a link to the docs for this syntax?

tests/sqlparser_common.rs
```diff
@@ -12785,6 +12785,61 @@ fn parse_connect_by() {
...
+                    op: BinaryOperator::Eq,
+                    right: Expr::Prior(Expr::Identifier(Ident::new("child")).into()).into(),
+                }],
+                nocycle: true,
+            }),
+            flavor: SelectFlavor::Standard,
+        }
+    );
+
+    // ~ CONNECT BY after WHERE and before GROUP BY
```
```suggestion
    // CONNECT BY after WHERE and before GROUP BY
```

src/parser/mod.rs
```diff
@@ -14185,25 +14185,35 @@ impl<'a> Parser<'a> {
 
     /// Parse a `CONNECT BY` clause (Oracle-style hierarchical query support).
     pub fn parse_connect_by(&mut self) -> Result<ConnectBy, ParserError> {
-        let (condition, relationships) = if self.parse_keywords(&[Keyword::CONNECT, Keyword::BY]) {
-            let relationships = self.with_state(ParserState::ConnectBy, |parser| {
-                parser.parse_comma_separated(Parser::parse_expr)
+        let (condition, relationships, nocycle) = if self
+            .parse_keywords(&[Keyword::CONNECT, Keyword::BY])
+        {
+            let (relationships, nocycle) = self.with_state(ParserState::ConnectBy, |parser| {
```
iffyio: it rather looks like `CONNECT BY` and `START WITH` are two different clauses, it looks a bit confusing they're being parsed by the same method?

iffyio: wondering a couple things around the current function impl

The current impl essentially duplicates the code in the if and else blocks. Since the goal is solely to parse in arbitrary order, can we rewrite it to use a loop instead? e.g.
```rust
loop {
    if self.parse_keywords(STARTS, WITH) {
    } else if self.parse_keywords(CONNECT, BY) {
    } else {
       break
    }
}
```
I think the function would be much easier to follow as a result.

Also, can we instead of having three separate but fields `relationships` and `nocycle` and `condition`, can we use a representation like e.g.

```rust
enum ConnectByKind {
    ConnectBy{
        relationships: Vec<Expr>,
        nocycle: bool,
    },
    StartsWith(Expr)
}

connect_by: Vec<ConnectByKind>
```
and the order in which the clauses show up in the query will be preserved as a result too (and we skip the consideration of whether snowflake supports specific ordering)

src/parser/mod.rs
xitep: :ok_hand: yes! thanks

tests/sqlparser_common.rs
xitep: :D old habits die hard, obviously! (i'm sorry for the troubles.)

src/ast/query.rs
xitep: i've added the links, but it kind of feels like duplication; after all the two relevant links are just on top of the struct.

src/parser/mod.rs
xitep: yes, that's good idea! and i'll go for it, since `sqlparser-rs` wants to be generic and cover many dialects.

however, there's a subtle difference between the two branches:

at least for `Oracle`, an initial `START WITH ...` _must be followed_ by a `CONNECT BY ...`; whereas an initial `CONNECT BY` _may have_ a `START WITH ...` follower.

with the `Vec<ConnectByKind>`, one can validate this afterwards; similarly the `loop` implies that a `CONNECT BY ... START WITH ... CONNECT BY ... CONNECT BY` would be parsed successfully, which the client would have to validate (and very likely reject.)

i guess it boils down to what `sqlparser-rs` wants be, either "strictly-typed" or "duck-typed", figuratively speaking.  personally i'm rather inclined towards the former, since it takes away the burden from the
clients and centralises know-how into the (shared) library. (just like most of the time, it's simply a trade-off.)

xitep: fyi; i just realised that for validation and error reporting we'll need at least to have the start tokens of the corresponding "kind" clauses; i'll extend `ConnectByKind` to this:

```rust
enum ConnectByKind {
    ConnectBy{
        connect_token: AttachedToken,
        nocycle: bool,
        relationships: Vec<Expr>,
    },
    StartsWith {
        start_token: AttachedToken,
        condition: Expr
   }
}
```

src/parser/mod.rs
```diff
@@ -14184,27 +14188,27 @@ impl<'a> Parser<'a> {
...
-            (condition, relationships)
-        };
-        Ok(ConnectBy {
-            condition,
-            relationships,
-        })
+    pub fn maybe_parse_connect_by(&mut self) -> Result<Vec<ConnectByKind>, ParserError> {
+        let mut clauses = Vec::with_capacity(2);
+        loop {
+            if let Some(idx) = self.parse_keywords_(&[Keyword::START, Keyword::WITH]) {
```
iffyio: can we instead save `let idx = self.index` in the loop before calling the parse method? looks like that would let us avoid introducing the `parse_keywords_` method?

xitep: unfortunately that won't work. (i tried.)

the underlying method calls `parse_keyword -> peek_keyword -> peek_token_ref -> peek_nth_token_ref` will skip leading whitespace. therefore, an index remembered before invoking this call chain might point to a whitespace instead of the identified keyword.

src/parser/mod.rs
```diff
@@ -4521,16 +4521,29 @@ impl<'a> Parser<'a> {
...
+                        self.index = start_index;
+                        first_keyword_index = None;
+                        break;
+                    }
+                }
             }
+            _ => {}
         }
-        true
+        first_keyword_index
```
iffyio: can we rewrite to something like this to follow the previous pattern? (with the slice matching and calling parse_keyword multiple times, makes the new version slightly harder to follow)
```rust
let start_index = self.index;
let mut first_keyword_index = None;
for keyword in keywords {
	if !self.parse_keyword(*keyword) {
		self.index = start_index;
		return None
	}
	if first_keyword_index.is_none() {
		first_keyword_index = Some(self.index.saturating_sub(1));
	}
}

first_keyword_index
```

src/parser/mod.rs
```diff
@@ -4521,16 +4521,29 @@ impl<'a> Parser<'a> {
...
-                // reset index and return immediately
-                self.index = index;
-                return false;
+        self.parse_keywords_(keywords).is_some()
+    }
+
+    /// Just like [Self::parse_keywords], but - upon success - returns the
+    /// token index of the first keyword.
+    #[must_use]
+    fn parse_keywords_(&mut self, keywords: &[Keyword]) -> Option<usize> {
```
```suggestion
    fn parse_keywords_indexed(&mut self, keywords: &[Keyword]) -> Option<usize> {
```

src/parser/mod.rs
iffyio: I see, that makes sense!

src/parser/mod.rs
xitep: i struggled with a good name; thanks for the suggestion

iffyio:
@xitep could you take a look at the conflicts when you have some time?

xitep:
should be good now, i rebased. please squash merge the whole pile of small commits. last but not least, many thanks for your support @iffyio

PR #2186, opened 2026-01-30
MSSQL: Support standalone BEGIN...END blocks

guan404ming:
## Why

I'm currently building sql-parser npm pkg for js/ts named [sqlparser-ts](https://github.com/guan404ming/sqlparser-ts), which powered by this amazing repo via WebAssembly! When reconstruct all dialects test in typescript, I find some issue and decide to contribute back to upstream~

The MSSQL dialect previously had no way to parse standalone BEGIN...END blocks. 
`BEGIN SELECT 1; END` would fail because parse_begin() only handles BEGIN TRANSACTION.

## How

wires MSSQL into the same path like BigQuery and Snowflake

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -17925,8 +17925,8 @@ impl<'a> Parser<'a> {
         })
     }
 
-    /// Parse a transaction modifier keyword that can follow a BEGIN statement.
-    pub fn parse_transaction_modifier(&mut self) -> Option<TransactionModifier> {
+    /// Parse a transaction modifier keyword that can follow a `BEGIN` statement.
+    pub(crate) fn parse_transaction_modifier(&mut self) -> Option<TransactionModifier> {
         if !self.dialect.supports_start_transaction_modifier() {
             None
         } else if self.parse_keyword(Keyword::DEFERRED) {


src/dialect/mssql.rs
```diff
@@ -145,7 +145,22 @@ impl Dialect for MsSqlDialect {
...
+                || parser.peek_keyword(Keyword::DEFERRED)
+                || parser.peek_keyword(Keyword::IMMEDIATE)
+                || parser.peek_keyword(Keyword::EXCLUSIVE)
+                || parser.peek_token_ref().token == Token::SemiColon
+                || parser.peek_token_ref().token == Token::EOF
+            {
+                parser.prev_token();
+                return None;
+            }
+            Some(parser.parse_begin_exception_end())
```
iffyio: would a condition like this do what we want?

```rust
if parse.peek_keywords(BEGIN, TRANSACTION) {
    None
} else if parse_keyword(BEGIN) {
    Some(parser.parse_begin_exception_end())
}
```
its not super clear to me why the current logic looks at WORK, TRY etc keywords?

guan404ming: Thanks for the review. These keywords are checked because `parse_begin()` handles more than just BEGIN TRANSACTION. It also handles WORK, TRY, CATCH, DEFERRED, IMMEDIATE, EXCLUSIVE (since MsSql's supports_start_transaction_modifier() returns true), as well as bare BEGIN;. If any of these are missed, they'd be incorrectly routed to parse_begin_exception_end() and fail to parse.

For example, `BEGIN TRY ... END TRY` is valid MSSQL syntax. If we only check for BEGIN TRANSACTION, then BEGIN TRY would fall through to `parse_begin_exception_end()`, which would try to parse TRY as a SQL statement and fail. because TRY is supposed to be handled as a https://github.com/apache/datafusion-sqlparser-rs/blob/c8b7f7cf4281cdfff3e08e9827928662b8be8095/src/parser/mod.rs#L17938-L17939

guan404ming: Feel free to let me know if there are any not clear, thanks!

iffyio: Oh I see, that makes sense! Can we do something like this to clarify the code?

we pull out the [linked fallback logic](https://github.com/apache/datafusion-sqlparser-rs/blob/c8b7f7cf4281cdfff3e08e9827928662b8be8095/src/parser/mod.rs#L17930-L17944) into a function  `parse_transaction_modifier()`. So that it can be reused.

Then here we can do e.g.

```rust
if without_modifier = self.maybe_parse(|parser| {
    if parser.parse_transaction_modifier()?.is_some() {
        parser_error!()
    } else {
        Some(())
    }
})?.is_some();
```
such that with that bool we know whether to defer to the main parser logic or continue with the begin/exception logic

guan404ming: Thanks for the really nice suggestion. I just updated it does more clean and work as well.

src/parser/mod.rs
```diff
@@ -17925,9 +17925,9 @@ impl<'a> Parser<'a> {
         })
     }
 
-    /// Parse a 'BEGIN' statement
-    pub fn parse_begin(&mut self) -> Result<Statement, ParserError> {
-        let modifier = if !self.dialect.supports_start_transaction_modifier() {
+    /// Parse a transaction modifier keyword that can follow a BEGIN statement.
+    pub fn parse_transaction_modifier(&mut self) -> Option<TransactionModifier> {
```
```suggestion
    pub(crate) fn parse_transaction_modifier(&mut self) -> Option<TransactionModifier> {
```

src/parser/mod.rs
```diff
@@ -17925,9 +17925,9 @@ impl<'a> Parser<'a> {
         })
     }
 
-    /// Parse a 'BEGIN' statement
-    pub fn parse_begin(&mut self) -> Result<Statement, ParserError> {
-        let modifier = if !self.dialect.supports_start_transaction_modifier() {
+    /// Parse a transaction modifier keyword that can follow a BEGIN statement.
```
```suggestion
    /// Parse a transaction modifier keyword that can follow a `BEGIN` statement.
```

guan404ming:
Just updated, thanks for the suggestion!

guan404ming:
Thanks @iffyio for your detailed reviews!

PR #2187, opened 2026-01-31
PostgreSQL: Support PostgreSQL ANALYZE with optional table and column

guan404ming:
## Why

ANALYZE, ANALYZE t, and ANALYZE t (col1, col2) all failed because parse_analyze() unconditionally required a table name and didn't handle the PostgreSQL parenthesized column syntax.

## How 

Made table_name optional (Option<ObjectName>) in the Analyze struct and added column list parsing after the table name. Updated Display and Spanned to match.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -3329,7 +3329,7 @@ impl Display for ExceptionWhen {
 ///
 /// Supported syntax varies by dialect:
 /// - Hive: `ANALYZE TABLE t [PARTITION (...)] COMPUTE STATISTICS [NOSCAN] [FOR COLUMNS [col1, ...]] [CACHE METADATA]`
-/// - PostgreSQL: `ANALYZE [VERBOSE] [t [(col1, ...)]]`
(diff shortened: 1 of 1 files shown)


src/ast/mod.rs
```diff
@@ -3325,19 +3325,24 @@ impl Display for ExceptionWhen {
     }
 }
 
-/// ANALYZE TABLE statement (Hive-specific)
+/// ANALYZE statement
+///
+/// Supported syntax varies by dialect:
+/// - Hive: `ANALYZE TABLE t [PARTITION (...)] COMPUTE STATISTICS [NOSCAN] [FOR COLUMNS [col1, ...]] [CACHE METADATA]`
+/// - PostgreSQL: `ANALYZE [VERBOSE] [t [(col1, ...)]]`
+/// - General: `ANALYZE [TABLE] t`
```
iffyio, line 3333: Can we add a link to the postgres docs describing the syntax? maybe this one https://www.postgresql.org/docs/current/sql-analyze.html

guan404ming: Nice suggestion, just updated. Thanks!

guan404ming:
Thanks!

PR #2189, opened 2026-02-01
Implement `core::error::Error` for `ParserError` and `TokenizerError`

LucaCappelletti94:
This PR updates `ParserError` and `TokenizerError` to implement `core::error::Error` instead of `std::error::Error`.

This change enables the Error trait implementation for these types in no_std environments unconditionally, removing the need for `#[cfg(feature = "std")]` guards.


iffyio (approved):

PR #2192, opened 2026-02-04
Snowflake: Lambda functions

yoavcloud:
This PR introduces the following changes:
1. Enabled Lambda function support for the Snowflake dialect
2. Expanded the Lambda struct to include an optional data type
3. Added support for parsing the Snowflake `TRANSFORM`, `FILTER` and `REDUCE` functions which take lambda as an argument.

See: https://docs.snowflake.com/en/user-guide/querying-semistructured#label-higher-order-functions


src/parser/mod.rs
```diff
@@ -1549,6 +1549,43 @@ impl<'a> Parser<'a> {
             Keyword::LAMBDA if self.dialect.supports_lambda_functions() => {
                 Ok(Some(self.parse_lambda_expr()?))
             }
+            Keyword::FILTER
+            | Keyword::TRANSFORM
+            | Keyword::REDUCE if self.dialect.supports_lambda_functions() => {
```
iffyio: the condition looks incompatible with non-snowflake dialects that support lambda functions, since those I'm imagining treat filter/transform/reduce as identifiers instead?

src/parser/mod.rs
```diff
@@ -1549,6 +1549,43 @@ impl<'a> Parser<'a> {
...
+                self.expect_token(&Token::LParen)?;
```
iffyio: can we move the logic to a standalone function? since its a lot of code to inline

src/parser/mod.rs
```diff
@@ -1549,6 +1549,43 @@ impl<'a> Parser<'a> {
...
+                let array = self.parse_expr()?;
```
iffyio: would it make sense to merge this and the existing `parse_lambda_expr`? that we accept arbitrary expressions as the parameters of the lambda instead of the latter only accepting identifiers? Similarly with the `typed` bool since the impl already uses `maybe_parse(data_type)`, thinking we can always accept a datatype if present in the syntax

src/ast/mod.rs
```diff
@@ -1445,6 +1445,26 @@ impl fmt::Display for LambdaFunction {
...
 
+/// A parameter to a lambda function, optionally with a data type.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct LambdaFunctionParameter {
+    /// The name of the parameter
+    pub name: Ident,
+    /// The optional data type of the parameter
+    pub data_type: Option<DataType>,
```
iffyio, line 1460: Can we add a link to the docs containing the data_type syntax?

tests/sqlparser_snowflake.rs
```diff
@@ -4559,3 +4559,11 @@ fn test_truncate_table_if_exists() {
     snowflake().verified_stmt("TRUNCATE TABLE my_table");
     snowflake().verified_stmt("TRUNCATE IF EXISTS my_table");
 }
+
+#[test]
+fn test_snowflake_lambda() {
+    snowflake().verified_expr("TRANSFORM([1, 2, 3], a -> a * 2)");
+    snowflake().verified_expr("TRANSFORM([1, 2, 3], a INT -> a * 2)");
+    snowflake().verified_expr("TRANSFORM([1, 2, 3], (x INT, y INT) -> (x + y))");
+    snowflake().verified_expr("REDUCE([1, 2, 3], 0, (acc, val) -> acc + val)");
```
iffyio: can we add test cases for `FILTER` as well?

yoavcloud:
@iffyio thanks for the feedback. I took a diff approach which I believe eliminates the concerns you raised. It's a bit of a shame that lambda expression parsing is spread across 3-4 diff flows but perhaps we can do something about it in the future.

src/parser/mod.rs
```diff
@@ -1603,10 +1603,34 @@ impl<'a> Parser<'a> {
...
+                    body: Box::new(self.parse_expr()?),
+                    syntax: LambdaSyntax::Arrow,
+                }))
+            }
+            // An unreserved word (likely an identifier) that is followed by another word (likley a data type)
+            // which is then followed by an arrow, which indicates a lambda function with a single, typed parameter.
+            // For example: `a INT -> a * 2`.
+            Token::Word(_)
+                if self.peek_nth_token_ref(1).token == Token::Arrow
+                    && self.dialect.supports_lambda_functions() =>
```
```suggestion
                if self.dialect.supports_lambda_functions() && self.peek_nth_token_ref(1).token == Token::Arrow  =>
```
thinking probably cleaner/quicker to first check the dialect before looking to parse any further?

src/keywords.rs
```diff
@@ -1051,6 +1052,7 @@ define_keywords!(
     TRACE,
     TRAILING,
     TRANSACTION,
+    TRANSFORM,
```
iffyio: it looks like both no longer need to be keywords in the diff?

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2197, opened 2026-02-05
Fix panic on incomplete REGEXP/MATCH expressions in SQLite dialect

LucaCappelletti94:
Parsing incomplete `REGEXP` or `MATCH` expressions (e.g. `SELECT 1 REGEXP`) would panic due to an `.unwrap()` on `parse_expr()` in the SQLite dialect's `parse_infix` method.

## Fix

Replaced the `.unwrap()` with proper error propagation so the parser returns a `ParserError` instead of panicking.

## Tests

Added assertions that `SELECT 1 REGEXP` and `SELECT 1 MATCH` produce errors.


iffyio (approved):

PR #2201, opened 2026-02-06
Fix panic in `SET AUTHORIZATION` parsing when scope modifier is missing

### LINKED ISSUE #2200: Reachable expect panic
With the following input, an expect is triggered which should really be an error:

```
"\tSET\t\t\t\t\t\t\t\t\t\tAUTHORIZATION\tTIME\t\t\t\t\t\tTIME��v\u{1}\0\0\t74843EUTI>\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0USER_RNCLUDE_NULL\0\0\0\0\0\0\0\0\0\t\t\t\t^^^^^^^^\tWHI\t\tIN"
```

From here: https://github.com/apache/datafusion-sqlparser-rs/blob/d7f56e8942f5e5ad2949b1a9be5f5b1c9bf1c3cd/src/parser/mod.rs#L14519

```
SET ... AUTHORIZATION must have a scope
```

LucaCappelletti94:
Parsing `SET AUTHORIZATION <value>` without a scope modifier (e.g. `SESSION`, `LOCAL`) causes a panic due to an `.expect()` call on a `None` value in `parse_set`. This was discovered via a fuzz-like input string but is trivially reproducible with any `SET AUTHORIZATION ...` statement that omits the scope keyword.

The panic occurs at mod.rs:
```rust
scope: scope.expect("SET ... AUTHORIZATION must have a scope"),
```

Valid SQL requires a scope modifier, e.g. `SET SESSION AUTHORIZATION 'user'`, so the missing scope is indeed an error — but it should be a graceful `ParserError`, not a panic.

Closes #2200


src/parser/mod.rs
```diff
@@ -14515,8 +14515,13 @@ impl<'a> Parser<'a> {
                 let value = self.parse_identifier()?;
                 SetSessionAuthorizationParamKind::User(value)
             };
+            let scope = scope.ok_or_else(|| {
+                ParserError::ParserError(
+                    "Expected a scope modifier (e.g. SESSION) before AUTHORIZATION".to_string(),
+                )
```
iffyio: can we use one of the helper functions like `self.expected_at`?

tests/sqlparser_common.rs
```diff
@@ -18273,6 +18273,18 @@ fn test_parse_set_session_authorization() {
     );
 }
 
+#[test]
+fn test_set_authorization_without_scope_errors() {
+    // This should return a parser error, not panic.
+    let res = parse_sql_statements(
+        "\tSET\t\t\t\t\t\t\t\t\t\tAUTHORIZATION\tTIME\t\t\t\t\t\tTIME\u{fffd}\u{fffd}v\u{1}\0\0\t74843EUTI>\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0USER_RNCLUDE_NULL\0\0\0\0\0\0\0\0\0\t\t\t\t^^^^^^^^\tWHI\t\tIN"
```
iffyio: is there a simpler test case that reproduces the error? I'm assuming the escape sequences and null characters shouldnt be necessary based on the diff

LucaCappelletti94: Could be, but I thought it was small enough of a crash case

src/parser/mod.rs
LucaCappelletti94: Ok

iffyio (approved):

PR #2202, opened 2026-02-06
MSSQL: Support `THROW` statement

guan404ming:
## Why

The MSSQL THROW statement was missing from the sqlparser, preventing parsing of T-SQL error handling code.

## How

- Add THROW keyword and Statement::Throw AST variant with optional error_number, message, state
- Add parse_throw() handling both THROW (re-throw) and THROW num, msg, state forms
- Add tests for literal args, variable args, and argument-less re-throw

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2816,6 +2816,41 @@ impl fmt::Display for RaiseStatementValue {
     }
 }
 
+/// A MSSQL `THROW` statement.
+///
+/// ```sql
+/// THROW [ error_number, message, state ]
+/// ```
+///
+/// [MSSQL](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/throw-transact-sql)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ThrowStatement {
+    /// Error number expression.
+    pub error_number: Option<Box<Expr>>,
+    /// Error message expression.
+    pub message: Option<Box<Expr>>,
+    /// State expression.
+    pub state: Option<Box<Expr>>,
+}
+
+impl fmt::Display for ThrowStatement {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let ThrowStatement {
+            error_number,
+            message,
+            state,
+        } = self;
+
+        write!(f, "THROW")?;
+        if let (Some(error_number), Some(message), Some(state)) = (error_number, message, state) {
+            write!(f, " {error_number}, {message}, {state}")?;
+        }
+        Ok(())
+    }
+}
+
 /// Represents an expression assignment within a variable `DECLARE` statement.
 ///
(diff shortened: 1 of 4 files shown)


guan404ming:
Hi @iffyio I would like modify the PR title to `MSSQL: support THROW statement` to better align the convention here but I found there is no edit button here for this pr. Could you help modify it, thanks!

iffyio:
Done!

guan404ming:
Thanks! Please let me know if there is anything could be better in this pr, thanks!

src/ast/mod.rs
```diff
@@ -4676,6 +4676,19 @@ pub enum Statement {
...
+    /// ```
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/language-elements/throw-transact-sql>
+    Throw {
+        /// Error number expression.
+        error_number: Option<Box<Expr>>,
+        /// Error message expression.
+        message: Option<Box<Expr>>,
+        /// State expression.
+        state: Option<Box<Expr>>,
+    },
```
iffyio: we're slowly [moving away](https://github.com/apache/datafusion-sqlparser-rs/issues/1204) from this representation over to having all statements wrapped in a named struct. As a result, can we do something like this instead?
```rust
struct Throw { ... }
Statement::Throw(Throw)
```
See the [RAISE](https://github.com/guan404ming/datafusion-sqlparser-rs/blob/379434abc35038bc5e839c2b6571a02b2fe130d2/src/ast/mod.rs#L3459-L3460) statement for an example

src/parser/mod.rs
```diff
@@ -670,6 +670,7 @@ impl<'a> Parser<'a> {
                 Keyword::RELEASE => self.parse_release(),
                 Keyword::COMMIT => self.parse_commit(),
                 Keyword::RAISERROR => Ok(self.parse_raiserror()?),
+                Keyword::THROW => Ok(self.parse_throw()?),
```
```suggestion
                Keyword::THROW => {
                    self.prev_token();
                    self.parse_throw().map(Into::into)
                },
```

see [RAISE](https://github.com/guan404ming/datafusion-sqlparser-rs/blob/379434abc35038bc5e839c2b6571a02b2fe130d2/src/parser/mod.rs#L618-L621) for an example. we're moving to have the statement parsing functions return the actual struct instead of a `Statement` enum variant. Also it would be ideal that the `parse_throw` is a standalone function (able to parse a `THROW` statement) so we rewind the token before invoking it

src/parser/mod.rs
```diff
@@ -18260,6 +18261,31 @@ impl<'a> Parser<'a> {
...
+        // THROW with no arguments is a re-throw inside a CATCH block
+        if self.peek_token_ref().token == Token::SemiColon
+            || self.peek_token_ref().token == Token::EOF
+        {
+            return Ok(Statement::Throw {
+                error_number: None,
+                message: None,
+                state: None,
+            });
+        }
```
```suggestion
```
I think we can skip this logic, it would be expected for the function to return an error if the input is empty (per the previous comment about making this function stand-alone)

src/ast/mod.rs
guan404ming: Thanks for your information. I just update with wrapped named struct.

src/parser/mod.rs
guan404ming: Ok, that makes sense to me. Just moved~

PR #2203, opened 2026-02-06
Updated Fuzzer harness to address all dialects

RPG-Alex:
The fuzzer harness was formerly only fuzzing `GenericDialect` and none of the other `Dialect` types. I modified the harness to account for all dialects and it should now be fuzzing all dialects.


LucaCappelletti94:
Looks good to me but I do not have merge rights - @iffyio it is a pretty small PR if you can check it out when you have time.

PR #2209, opened 2026-02-08
[PIVOT] Optional AS keyword for aliases

xitep:


xitep:
I'm sorry, but I seem not to be able to change this PR's title; it should read _"Optional AS keyword for PIVOT value aliases"_; it's _not_ affecting the `UNPIVOT` clause.

iffyio (approved):
LGTM! Thanks @xitep!

PR #2212, opened 2026-02-09
MSSQL: Add support for TRAN shorthand

guan404ming:
## Why

MSSQL supports TRAN as a shorthand for TRANSACTION in BEGIN/COMMIT/ROLLBACK statements, but the
parser did not recognize it.

## How

- Add TRAN variant to BeginTransactionKind enum and the keywords list
- Accept TRAN alongside TRANSACTION and WORK in BEGIN, COMMIT, and ROLLBACK parsing
- Add tests for BEGIN TRAN, COMMIT TRAN, and ROLLBACK TRAN

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -6369,7 +6369,7 @@ impl Display for CascadeOption {
     }
 }
 
-/// Transaction started with [ TRANSACTION | WORK ]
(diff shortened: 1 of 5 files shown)


guan404ming:
Thanks!

PR #2213, opened 2026-02-09
PostgreSQL: Support PRIMARY KEY/UNIQUE USING INDEX

guan404ming:
## Why

PostgreSQL allows promoting an existing unique index to a PRIMARY KEY or UNIQUE table constraint via ALTER TABLE, which was
not supported by the parser.

## How

- Add ConstraintUsingIndex variant to TableConstraint with display, span, and serialization support
- Parse USING INDEX index_name after PRIMARY KEY or UNIQUE keywords in constraint parsing
- Add tests for named/unnamed constraints and deferrable characteristics

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9276,6 +9276,21 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse `index_name [ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]`
+    /// after `{ PRIMARY KEY | UNIQUE } USING INDEX`.
+    fn parse_constraint_using_index(
+        &mut self,
+        name: Option<Ident>,
+    ) -> Result<ConstraintUsingIndex, ParserError> {
+        let index_name = self.parse_identifier()?;
+        let characteristics = self.parse_constraint_characteristics()?;
+        Ok(ConstraintUsingIndex {
+            name,
+            index_name,
+            characteristics,
+        })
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -9375,6 +9391,22 @@ impl<'a> Parser<'a> {
...
+                    return Ok(Some(
+                        ConstraintUsingIndex {
+                            name,
+                            is_primary_key: true,
+                            index_name,
+                            characteristics,
+                        }
+                        .into(),
+                    ));
+                }
```
iffyio, line 9407: can we pull this out into a e.g. `parse_constraint_using_index()` function so that it can be reused across both cases?

Also repr wise, I think it would be clearer to have both cases as unique variants on the `TableConstraint` enum vs using the `is_primary_key` bool to differentiate. i.e.
```rust
TableConstraint::UniqueUsingIndex(ConstraintUsingIndex)
TableConstraint::PrimaryKeyUsingIndex(ConstraintUsingIndex)
```

guan404ming: Yes, thanks for the nice suggestions. I just updated with 

1. Split enum variant: `TableConstraint::ConstraintUsingIndex` into `PrimaryKeyUsingIndex` and `UniqueUsingIndex`, removing the `is_primary_key bool` from the struct.                                                                                                                                                      
2. Extracted helper: Created `parse_constraint_using_index()` to deduplicate the parsing logic used by both `UNIQUE USING INDEX` and `PRIMARY KEY USING INDEX`.

PR #2214, opened 2026-02-11
[Oracle] Table alias for INSERTed table

xitep:
This supports an "AS"-less alias for inserts, as in  `INSERT table t ...` ([Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/INSERT.html#GUID-903F8043-0254-4EE9-ACC1-CB8AC0AF3423__GUID-AC239E95-0DD5-4F4B-A849-55C87840D0E6) doesn't like the keyword)

### DIFF
--- a/tests/sqlparser_oracle.rs
+++ b/tests/sqlparser_oracle.rs
@@ -517,7 +517,6 @@ fn test_insert_without_alias() {
     // check VALUES
     let sql = "INSERT INTO t VALUES (1)";
     let stmt = oracle_dialect.verified_stmt(sql);
-    dbg!(&stmt);
     assert!(matches!(
         stmt,
         Statement::Insert(Insert {


src/parser/mod.rs
```diff
@@ -4543,7 +4543,13 @@ impl<'a> Parser<'a> {
...
     /// Returns true if the current token matches the expected keyword.
     pub fn peek_keyword(&self, expected: Keyword) -> bool {
-        matches!(&self.peek_token_ref().token, Token::Word(w) if expected == w.keyword)
+        self.peek_keyword_one_of(&[expected])
+    }
+
+    #[must_use]
+    /// Checks whether the current token is one of the expected keywords without consuming it.
+    fn peek_keyword_one_of(&self, expected: &[Keyword]) -> bool {
+        matches!(&self.peek_token_ref().token, Token::Word(w) if expected.contains(&w.keyword))
```
iffyio: is this identical to the existing `peek_one_of_keywords` function?

src/parser/mod.rs
```diff
@@ -16289,13 +16281,40 @@ impl<'a> Parser<'a> {
         Ok(ExprWithAlias { expr, alias })
     }
 
+    /// Parse an expression followed by an optional alias; Unlike
```
iffyio, line 16374: can we fix up the commits with the diffs from https://github.com/apache/datafusion-sqlparser-rs/pull/2209 contained in this PR?

src/parser/mod.rs
xitep: yes, `self.peek_one_of_keywords().is_some()` would do instead of this new method. except this method avoids the clone of the token. i don't have a strong preference, but i'd find the clone wasteful. wdyt?

src/parser/mod.rs
xitep: looks like a rebase did the trick :)

src/parser/mod.rs
iffyio: oh but the keyword is copy? so that no cloning should be involved - unless I'm misunderstanding

src/parser/mod.rs
```diff
@@ -17133,12 +17139,26 @@ impl<'a> Parser<'a> {
             let table = self.parse_keyword(Keyword::TABLE);
             let table_object = self.parse_table_object()?;
 
-            let table_alias =
-                if dialect_of!(self is PostgreSqlDialect) && self.parse_keyword(Keyword::AS) {
-                    Some(self.parse_identifier()?)
+            let table_alias = if dialect_of!(self is OracleDialect) {
+                if !self.peek_sub_query()
+                    && !self.peek_keyword_one_of(&[Keyword::DEFAULT, Keyword::VALUES])
```
iffyio: did we have existing test cases covering both conditions (subquey and default values)?

src/ast/dml.rs
```diff
@@ -56,8 +56,9 @@ pub struct Insert {
     pub into: bool,
     /// TABLE
     pub table: TableObject,
-    /// table_name as foo (for PostgreSQL)
-    pub table_alias: Option<Ident>,
+    /// `table_name as foo` (for PostgreSQL)
+    /// `table_name foo` (for Oracle)
+    pub table_alias: Option<InsertTableAlias>,
```
iffyio:
```suggestion
    pub table_alias: Option<TableAlias>,
```
can we reuse `TableAlias` here?

src/parser/mod.rs
```diff
@@ -17133,12 +17139,26 @@ impl<'a> Parser<'a> {
             let table = self.parse_keyword(Keyword::TABLE);
             let table_object = self.parse_table_object()?;
 
-            let table_alias =
-                if dialect_of!(self is PostgreSqlDialect) && self.parse_keyword(Keyword::AS) {
-                    Some(self.parse_identifier()?)
+            let table_alias = if dialect_of!(self is OracleDialect) {
```
iffyio: maybe we can use a dialect method for this since the macro is deprecated.

src/ast/mod.rs
```diff
@@ -6451,6 +6451,17 @@ pub struct InsertAliases {
     pub col_aliases: Option<Vec<Ident>>,
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+/// Optional alias for an `INSERT` table; i.e. the table to be inserted into
+pub struct InsertTableAlias {
```
```suggestion
pub struct TableAliasWithoutColumns {
```
or something similar that isn't specific to the insert statement, in case   the struct needs to be reused in other contexts?

src/parser/mod.rs
xitep: of course! you are right! (for some reason i assumed a string underlying `Keyword` :facepalm:) with that, of course, the new method makes no sense.

src/ast/mod.rs
xitep: i could also live with using [TableAlias](https://docs.rs/sqlparser/latest/sqlparser/ast/struct.TableAlias.html) directly. (`columns` would then simply be left empty.) it would make the `Insert` struct needlessly larger by 24 bytes, though.

src/parser/mod.rs
xitep: since this branch is oracle specific, i'll need to add specific tests for this dialect

src/parser/mod.rs
```diff
@@ -17166,12 +17166,27 @@ impl<'a> Parser<'a> {
             let table = self.parse_keyword(Keyword::TABLE);
             let table_object = self.parse_table_object()?;
 
-            let table_alias =
-                if dialect_of!(self is PostgreSqlDialect) && self.parse_keyword(Keyword::AS) {
-                    Some(self.parse_identifier()?)
-                } else {
-                    None
-                };
+            let table_alias = if self.dialect.supports_insert_table_implicit_alias()
```
iffyio: thinking the double dialect methods introduced and the logic here could be simplified with something like:
```rust
let table_alias = if self.dialect.supports_insert_table_alias() && !self.peek_sub_query() && self.peek_one_of_keywords(AS, DEFAULT, VALUES).is_none() {
    let explicit = self.parse_keyword(AS);
    let alias = self.maybe_parse(parse_identifier); // required if `explicit`
    Some(TableAliasWithoutColumns {
        explicit: true,
        alias: self.parse_identifier()?,
    })
} else { None }
```

xitep: i understand having method less in the dialect is desirable. but i fear it would be too simplistic. the difference between postgres and oracle is that the former requires the "AS" keyword for the alias, while the latter does not accept it and will lead to an error. so, while both dialects want to allow the parsing of the "insert table aliases", they both need to tell the parser in which kind/style the alias is accepted.

i can offer to introduce an `enum AliasStyle { ExplicitOnly, ImplicitOnly, Both }` together with a `Dialect::supports_insert_table_alias(&self) -> Option<AliasStyle>`. nevertheless, it will still leave the control flow in the parser's "insert table" method rather complicated :/

iffyio: for oracle I imagine `AS` is a keyword - so that the following should be invalid query?

```sql
INSERT INTO T as SELECT 1
```

I don't think it would be necessarily problematic that the parser accepts the following

```sql
INSERT INTO T AS alias SELECT 1; // oracle
INSERT INTO T alias SELECT 1; // postgres
```

since downstream crates can further validate if desired.

xitep: * yes, it is a keyword and `insert into t as ...` is invalid for oracle.
* ok, let's do it as you suggested then.

i'll update this pr shortly.

tests/sqlparser_oracle.rs
```diff
@@ -421,3 +424,107 @@ fn test_connect_by() {
...
+            table_alias: None,
+            source: Some(source),
+            ..
+        })
+        if matches!(&**source, Query { body, .. } if matches!(&**body, SetExpr::Select(_)))));
+
+    // check VALUES
+    let sql = "INSERT INTO t VALUES (1)";
+    let stmt = oracle_dialect.verified_stmt(sql);
+    dbg!(&stmt);
```
```suggestion
```

xitep: uf! (i'm sorry.)

PR #2216, opened 2026-02-12
Allow custom OptimizerHints

altmannmarcelo:
Two orthogonal improvements to optimizer hint parsing:

1. `Option<OptimizerHint>` -> `Vec<OptimizerHint>`: the old Option silently dropped all but the first hint-style comment. Vec preserves all hint comments the parser encounters, letting consumers decide which to use. This is backwards compatible: `optimizer_hint: None` becomes `optimizer_hints: vec![]`, and `optimizer_hint.unwrap()` becomes `optimizer_hints[0]`.

2. Generic prefix extraction: the `/*+...*/` pattern is an established convention. Various systems extend it with `/*prefix+...*/` where the prefix is opaque alphanumeric text before `+`. Rather than adding a new dialect flag or struct for each system, the parser now captures any `[a-zA-Z0-9]*` run before `+` as a `prefix` field. Standard hints have `prefix: ""`. No new dialect surface -- same `supports_comment_optimizer_hint()` gate. This makes OptimizerHint a generic extension point: downstream consumers can define their own prefixed hint conventions and filter hints by prefix, without requiring any changes to the parser or dialect configuration.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13134,7 +13134,7 @@ impl<'a> Parser<'a> {
 
     /// Parse a `DELETE` statement and return `Statement::Delete`.
     pub fn parse_delete(&mut self, delete_token: TokenWithSpan) -> Result<Statement, ParserError> {
-        let optimizer_hint = self.maybe_parse_optimizer_hint()?;
+        let optimizer_hints = self.maybe_parse_optimizer_hints()?;
         let (tables, with_from_keyword) = if !self.parse_keyword(Keyword::FROM) {
             // `FROM` keyword is optional in BigQuery SQL.
             // https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#delete_statement
@@ -13178,7 +13178,7 @@ impl<'a> Parser<'a> {
 
         Ok(Statement::Delete(Delete {
             delete_token: delete_token.into(),
-            optimizer_hint,
+            optimizer_hints,
             tables,
             from: if with_from_keyword {
                 FromTable::WithFromKeyword(from)
@@ -13950,7 +13950,7 @@ impl<'a> Parser<'a> {
             if !self.peek_keyword(Keyword::SELECT) {
                 return Ok(Select {
                     select_token: AttachedToken(from_token),
-                    optimizer_hint: None,
+                    optimizer_hints: vec![],
                     distinct: None,
                     select_modifiers: None,
                     top: None,
@@ -13979,7 +13979,7 @@ impl<'a> Parser<'a> {
         }
 
         let select_token = self.expect_keyword(Keyword::SELECT)?;
-        let optimizer_hint = self.maybe_parse_optimizer_hint()?;
+        let optimizer_hints = self.maybe_parse_optimizer_hints()?;
         let value_table_mode = self.parse_value_table_mode()?;
 
         let (select_modifiers, distinct_select_modifier) =
@@ -14138,7 +14138,7 @@ impl<'a> Parser<'a> {
 
         Ok(Select {
             select_token: AttachedToken(select_token),
-            optimizer_hint,
+            optimizer_hints,
             distinct,
             select_modifiers,
             top,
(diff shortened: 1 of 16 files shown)


src/parser/mod.rs
```diff
@@ -14168,53 +14168,75 @@ impl<'a> Parser<'a> {
...
+    /// and returns `(prefix, text_after_plus)` if so.
+    fn extract_hint_prefix_and_text(comment: &str) -> Option<(String, String)> {
+        let plus_pos = comment.find('+')?;
+        let before_plus = &comment[..plus_pos];
+        if before_plus.chars().all(|c| c.is_ascii_alphanumeric()) {
+            let text = &comment[plus_pos + 1..];
+            Some((before_plus.to_string(), text.to_string()))
+        } else {
+            None
+        }
```
xitep, line 14229: i think using `str::split_once` would make this shorter and leaner (possibly slightly more efficient), e.g. https://gist.github.com/rust-play/146d81960095525d6384f34d84ac7419

tests/sqlparser_oracle.rs
```diff
@@ -338,36 +338,32 @@ fn parse_national_quote_delimited_string_but_is_a_word() {
...
-            .optimizer_hint
-            .as_ref()
-            .map(|hint| hint.text.as_str()),
-        Some("one two three")
+        "SELECT /*+one two three*/ /*+not a hint!*/ 1 FROM dual",
     );
+    assert_eq!(select.optimizer_hints.len(), 2);
+    assert_eq!(select.optimizer_hints[0].text, "one two three");
+    assert_eq!(select.optimizer_hints[0].prefix, "");
+    assert_eq!(select.optimizer_hints[1].text, "not a hint!");
```
xitep: well, this test was to assert that "not a hint!" is in fact "not an optimizer hint!" :)

xitep (commented):
hello @altmannmarcelo,

i quite like your generalisation if it makes the feature useful for dialects supporting multiple hints! :+1: so i'm in for the change.

however, i wish you would introduce dialect flags to guide the parser  to 1) allow the prefixes, and 2) allow accepting multiple hints (the AST can nicely present that with your suggestion of `Vec<OptimizerHint>`.)

yes, you wrote that downstream programs can do the validation themselves. and indeed they can. but, when you start writing your 3rd sql processor based on `sqlparser`, it become tiresome to repeat those validations in all of them (or to start maintaining a separate crate for these validations.) the great thing about `sqlparser` is that it has the concept of "dialects" and provides a common AST for all of them, yet is able to distinguish between the dialects' "idiosyncrasies." (having said that, i'm no authority and don't have a say in how `sqlparser-rs` wants to evolve.)

altmannmarcelo:
Thanks for the review @xitep, and glad you like the generalization!

  I understand the appeal of dialect flags — they're great when the parser can definitively say "this syntax is valid/invalid for dialect X." However, I think optimizer hints are a different category. The `/*+ ...
  */`  pattern is an established convention precisely because it's a standard way to extend functionality without requiring grammar changes. Different systems (and even different versions of the same system)
  recognize different hint keywords and prefixes, and the parser shouldn't need to know about all of them.

  The current approach keeps the parser's responsibility focused: if the dialect supports optimizer hints (`supports_comment_optimizer_hint()`), collect them all and let the downstream consumer decide which ones are
  relevant. This is intentional — adding dialect flags for which prefixes are allowed or how many hints are accepted would couple the parser to specific hint conventions that are really consumer-level concerns.

  For downstream projects, the filtering is straightforward — match on prefix and ignore what you don't recognize. This is simpler and more flexible than maintaining dialect flags that would need updating as hint
  conventions evolve. It also means new prefix conventions can be adopted by consumers without any changes to sqlparser-rs.

xitep:
i agree it will make the maintenance of `sqlparser` simpler by pushing the responsibilities downstream. and yes, in this case, with the optimizer hints, it is straightforward.

src/parser/mod.rs
```diff
@@ -14168,53 +14168,75 @@ impl<'a> Parser<'a> {
...
-                                            }
-                                        } else {
-                                            OptimizerHintStyle::MultiLine
-                                        },
-                                    };
-                                    // Consume the comment token
-                                    self.next_token_no_skip();
-                                    Some(hint)
-                                }
+                Token::Whitespace(ws) => match ws {
```
iffyio: since the match now only has one item, it looks like we can simplify the flow to remove one layer of indentation with e.g.?
```rust
let Token::Whitespace(ws) = &t.token else {
    break
};
match ws {
}
```

src/parser/mod.rs
iffyio: @altmannmarcelo could we use the `split_once` proposed to simplify the function body?

tests/sqlparser_mysql.rs
```diff
@@ -4641,6 +4641,25 @@ fn test_optimizer_hints() {
         "\
        DELETE /*+ foobar */ FROM table_name",
     );
+
+    // ~ prefixed hints: any alphanumeric prefix before `+` is captured
```
```suggestion
    // prefixed hints: any alphanumeric prefix before `+` is captured
```

tests/sqlparser_mysql.rs
```diff
@@ -4641,6 +4641,25 @@ fn test_optimizer_hints() {
...
+    let select = mysql_dialect.verified_only_select("SELECT /*abc+ text */ 1");
+    assert_eq!(select.optimizer_hints.len(), 1);
+    assert_eq!(select.optimizer_hints[0].prefix, "abc");
+    assert_eq!(select.optimizer_hints[0].text, " text ");
+
+    // ~ multiple hints with different prefixes
```
```suggestion
    // multiple hints with different prefixes
```

tests/sqlparser_mysql.rs
```diff
@@ -4641,6 +4641,25 @@ fn test_optimizer_hints() {
...
+    let select = mysql_dialect.verified_only_select("SELECT /*+ A */ /*x2+ B */ 1");
+    assert_eq!(select.optimizer_hints.len(), 2);
+    assert_eq!(select.optimizer_hints[0].prefix, "");
+    assert_eq!(select.optimizer_hints[0].text, " A ");
+    assert_eq!(select.optimizer_hints[1].prefix, "x2");
+    assert_eq!(select.optimizer_hints[1].text, " B ");
+
+    // ~ prefixed hints in INSERT/UPDATE/DELETE
```
```suggestion
    // prefixed hints in INSERT/UPDATE/DELETE
```

tests/sqlparser_mysql.rs
```diff
@@ -4641,6 +4641,25 @@ fn test_optimizer_hints() {
...
     );
+
+    // ~ prefixed hints: any alphanumeric prefix before `+` is captured
+    let select = mysql_dialect.verified_only_select("SELECT /*abc+ text */ 1");
+    assert_eq!(select.optimizer_hints.len(), 1);
+    assert_eq!(select.optimizer_hints[0].prefix, "abc");
+    assert_eq!(select.optimizer_hints[0].text, " text ");
+
+    // ~ multiple hints with different prefixes
+    let select = mysql_dialect.verified_only_select("SELECT /*+ A */ /*x2+ B */ 1");
```
iffyio, line 4652: can we add a test case like the following mixing hint with regular comments?
`SELECT /*+ A */ /* Regular comment */ /*x2+ B */ 1`

tests/sqlparser_oracle.rs
```diff
@@ -338,36 +338,32 @@ fn parse_national_quote_delimited_string_but_is_a_word() {
 fn test_optimizer_hints() {
     let oracle_dialect = oracle();
 
-    // selects
+    // ~ selects: all `/*+...*/` comments are collected as hints
```
```suggestion
    // selects: all `/*+...*/` comments are collected as hints
```

tests/sqlparser_oracle.rs
iffyio: @altmannmarcelo maybe we can change the `not a hint!` value to avoid confusion in the test case going forward?

tests/sqlparser_oracle.rs
```diff
@@ -387,6 +383,15 @@ fn test_optimizer_hints() {
                (pt.person_id, pt.first_name, pt.last_name, pt.title) \
                VALUES (ps.person_id, ps.first_name, ps.last_name, ps.title)",
     );
+
+    // ~ single-line prefixed hint (Oracle supports `--` without trailing whitespace)
```
```suggestion
    // single-line prefixed hint (Oracle supports `--` without trailing whitespace)
```

iffyio (approved):
LGTM! Thanks @altmannmarcelo!

PR #2217, opened 2026-02-13
Add SETOF support for PostgreSQL function return types

fmguerreiro:
PostgreSQL functions can return sets via `RETURNS SETOF <type>`. This is currently not supported by the parser, which fails when encountering `SETOF` after `RETURNS`.

### Changes

- Add `SETOF` keyword
- Add `DataType::SetOf(Box<DataType>)` variant
- Handle `SETOF` in `parse_data_type_helper()` so any `DataType::SetOf` round-trips through Display → parse
- Tests for SETOF with TEXT, INTEGER, RECORD, and schema-qualified custom types

### Example

```sql
CREATE FUNCTION get_users() RETURNS SETOF TEXT LANGUAGE sql AS 'SELECT name FROM users';
CREATE FUNCTION get_all() RETURNS SETOF my_schema."MyType" LANGUAGE sql AS 'SELECT * FROM t';
```

### Reference

- [PostgreSQL CREATE FUNCTION docs](https://www.postgresql.org/docs/current/sql-createfunction.html)


src/ast/data_type.rs
```diff
@@ -61,6 +61,11 @@ pub enum DataType {
         /// Table columns.
         columns: Vec<ColumnDef>,
     },
+    /// SETOF type modifier for [PostgreSQL] function return types,
+    /// e.g. `CREATE FUNCTION ... RETURNS SETOF text`.
+    ///
+    /// [PostgreSQL]: https://www.postgresql.org/docs/current/sql-createfunction.html
+    SetOf(Box<DataType>),
```
iffyio: would it make sense to represent this as a qualifier around the return type of a function instead? e.g.

```rust
enum FunctionReturnType {
    DataType(DataType),
    Setof(DataType)
}
CreateFunction { ... return_type: FunctionReturnType ... }
```
Thinking that would probably be better because it would be a bit confusing having a `SETOF` `DataType` that isn't a valid data type in any dialect

fmguerreiro: yep, makes sense! 62385ea

src/ast/ddl.rs
```diff
@@ -3466,6 +3466,28 @@ impl fmt::Display for CreateDomain {
...
+/// The return type of a `CREATE FUNCTION` statement.
+///
+/// [PostgreSQL](https://www.postgresql.org/docs/current/sql-createfunction.html)
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum FunctionReturnType {
+    /// `RETURNS <type>`
+    DataType(DataType),
+    /// `RETURNS SETOF <type>`
```
iffyio, line 3532: Can we instead move the link to the postgres docs here? since the enum itself applies to all dialects, whereas only this variant is pg specific

src/ast/spans.rs
```diff
@@ -632,33 +657,6 @@ impl Spanned for TableConstraint {
     }
 }
 
-impl Spanned for PartitionBoundValue {
```
iffyio, line 638: is there any changes in this diff (if not can we revert it)?

src/parser/mod.rs
```diff
@@ -5883,6 +5883,17 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse a [`FunctionReturnType`] after the `RETURNS` keyword.
+    ///
+    /// Handles `RETURNS SETOF <type>` and plain `RETURNS <type>`.
```
```suggestion
```

src/ast/ddl.rs
fmguerreiro: e9a35be

src/ast/spans.rs
fmguerreiro: 20d842f

src/parser/mod.rs
fmguerreiro: e9a35be

fmguerreiro:
@iffyio sorry for the delay — addressed all comments and merged upstream. would be great if you could take another look when you get a chance!

PR #2219, opened 2026-02-15
Fix `derive_dialect!` proc macro for use from external crates

alexander-beedie:
Unfortunately I made a classic proc-macro blunder in #2174, forgetting that `$CARGO_MANIFEST_DIR` points to the crate _being compiled_, which is not the `sqlparser` crate. So all the tests worked (as the crate being compiled and the `sqlparser` crate are the same), but actual use from an external crate fails (without exotic workarounds) 😅

This PR fixes and industrialises the source lookup; now it does the following (documented/explained in the code too):

1. `$CARGO_MANIFEST_DIR/src/dialect/mod.rs`: works when macro invoked from within the `sqlparser` crate (e.g. tests).
2. `<sqlparser_derive dir>/../src/dialect/mod.rs`: works when `sqlparser_derive` lives in a workspace alongside the main crate (standard `derive/` layout).
3. Sibling directories of compiled `sqlparser_derive` crate in the Cargo registry: works when an external crate uses `derive_dialect!` via registry dependency.

Confirmed this addresses the issue, and dialects can now be properly derived in/from external crates 👍 

**Also:** minor tweak to use slightly more idiomatic `map_err!` inside `derive_dialect`.

### DIFF
--- a/derive/src/dialect.rs
+++ b/derive/src/dialect.rs
@@ -120,24 +120,20 @@ impl Parse for DeriveDialectInput {
 
 /// Entry point for the `derive_dialect!` macro
 pub(crate) fn derive_dialect(input: DeriveDialectInput) -> proc_macro::TokenStream {
-    let err = |msg: String| {
-        Error::new(proc_macro2::Span::call_site(), msg)
-            .to_compile_error()
-            .into()
-    };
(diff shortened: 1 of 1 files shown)


iffyio (approved):

PR #2220, opened 2026-02-15
PostgreSQL: Support more COMMENT ON object types

guan404ming:
## Why                                                                                                               
  
PostgreSQL supports COMMENT ON for many object types, but the parser only handled a subset.                       
                                                                                       
## How

- Add Domain, Function, Index, MaterializedView, Procedure, Sequence, Type, and View to CommentObject
- Extend parser to handle the new object type keywords
- Expand test coverage to verify all supported object types


guan404ming:
Thanks!

PR #2223, opened 2026-02-19
Snowflake `COPY INTO`: support nested semi-structured path 

romanoff:
Extended the Snowflake COPY INTO data-load select item parser to support nested semi-structured path traversal (e.g. $1:a:b:c) by looping over consecutive :element segments instead of consuming only one.

Example of query:
```sql
COPY INTO tbl (col) FROM (SELECT $1:a:b AS col FROM @stage)
```

### DIFF
--- a/src/ast/helpers/stmt_data_loading.rs
+++ b/src/ast/helpers/stmt_data_loading.rs
@@ -21,6 +21,8 @@
 
 #[cfg(not(feature = "std"))]
 use alloc::string::String;
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
 use core::fmt;
 
 #[cfg(feature = "serde")]
--- a/src/ast/helpers/stmt_data_loading.rs
+++ b/src/ast/helpers/stmt_data_loading.rs
@@ -78,8 +78,9 @@ pub struct StageLoadSelectItem {
     pub alias: Option<Ident>,
     /// Column number within the staged file (1-based).
     pub file_col_num: i32,
-    /// Optional element identifier following the column reference.
-    pub element: Option<Ident>,
+    /// Optional semi-structured element path following the column reference
+    /// (e.g. `$1:UsageMetrics:hh` produces `["UsageMetrics", "hh"]`).
+    pub element: Option<Vec<Ident>>,
     /// Optional alias for the item (AS clause).
     pub item_as: Option<Ident>,
 }
@@ -116,8 +117,10 @@ impl fmt::Display for StageLoadSelectItem {
             write!(f, "{alias}.")?;
         }
         write!(f, "${}", self.file_col_num)?;
-        if let Some(element) = &self.element {
-            write!(f, ":{element}")?;
+        if let Some(elements) = &self.element {
+            for element in elements {
+                write!(f, ":{element}")?;
+            }
         }
         if let Some(item_as) = &self.item_as {
             write!(f, " AS {item_as}")?;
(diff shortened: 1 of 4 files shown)


src/dialect/snowflake.rs
```diff
@@ -1493,27 +1493,30 @@ fn parse_select_item_for_data_load(
...
 
-    // as
+    // optional alias: `AS alias` or just `alias` (implicit)
     if parser.parse_keyword(Keyword::AS) {
         item_as = Some(match parser.next_token().token {
             Token::Word(w) => Ok(Ident::new(w.value)),
             _ => parser.expected("column item alias", parser.peek_token()),
         }?);
+    } else if let Token::Word(w) = parser.peek_token().token {
+        if !RESERVED_FOR_COLUMN_ALIAS.contains(&w.keyword) {
```
iffyio: would it make sense to somehow reuse the [is_column_alias](https://github.com/romanoff/datafusion-sqlparser-rs/blob/ca092cff70baace15791c79c7e11ff3a80ee97a3/src/dialect/snowflake.rs#L464) method for this dialect instead?

Also can we add test cases that cover this behavior?

romanoff: @iffyio Updated PR

src/parser/mod.rs
```diff
@@ -12309,7 +12309,7 @@ impl<'a> Parser<'a> {
     }
 
     /// Optionally parses an alias for a select list item
-    fn maybe_parse_select_item_alias(&mut self) -> Result<Option<Ident>, ParserError> {
+    pub fn maybe_parse_select_item_alias(&mut self) -> Result<Option<Ident>, ParserError> {
```
```suggestion
    pub(crate) fn maybe_parse_select_item_alias(&mut self) -> Result<Option<Ident>, ParserError> {
```

src/dialect/snowflake.rs
```diff
@@ -1503,29 +1502,21 @@ fn parse_select_item_for_data_load(
...
         }
     }
-
-    // as
-    if parser.parse_keyword(Keyword::AS) {
-        item_as = Some(match parser.next_token().token {
-            Token::Word(w) => Ok(Ident::new(w.value)),
-            _ => parser.expected_ref("column item alias", parser.peek_token_ref()),
-        }?);
+    parser.prev_token();
```
iffyio: would it be possible to rewrite the code so that it `peeks` the expected `Token::Colon` before entering the loop? if that would let us avoid this call to `prev_token()`

romanoff:
@iffyio Updated. Thank you

src/ast/helpers/stmt_data_loading.rs
```diff
@@ -78,8 +80,9 @@ pub struct StageLoadSelectItem {
     pub alias: Option<Ident>,
     /// Column number within the staged file (1-based).
     pub file_col_num: i32,
-    /// Optional element identifier following the column reference.
-    pub element: Option<Ident>,
+    /// Optional semi-structured element path following the column reference
+    /// (e.g. `$1:UsageMetrics:hh` produces `["UsageMetrics", "hh"]`).
+    pub element: Option<Vec<Ident>>,
```
iffyio, line 85: I wonder, can we change this to `element: Vec<Ident>` to drop the option? feels like in this case it would be simpler if the Option no longer covers the previous intent

iffyio (commented):
Thanks @romanoff just a minor comment, (also fyi there's some conflicts on the branch), otherwise the changes look good to me

PR #2149, opened 2026-01-05
Add support for DuckDB `LAMBDA` keyword syntax

lovasoa:
## Summary

This PR adds support for parsing DuckDB's `LAMBDA` keyword syntax for lambda functions.

DuckDB [supports two syntaxes for lambda functions](https://duckdb.org/docs/stable/sql/functions/lambda):
- Arrow syntax: `x -> x + 1` (already supported, but deprecated in duckdb)
- Lambda keyword syntax: `lambda x : x + 1` (preferred syntax in duckdb, added in this PR)

Initially reported in https://github.com/sqlpage/SQLPage/issues/1171

### Changes

- Added `LAMBDA` keyword to keywords list
- Added `LambdaSyntax` enum to track which syntax style was used (for correct round-tripping)
- Added `parse_lambda_expr()` to parse `lambda <params> : <expr>` syntax
- Updated `LambdaFunction` struct to include `syntax` field
- Added tests for DuckDB lambda functions

### Examples

```sql
-- Single parameter
SELECT list_filter([1, 2, 3], lambda x : x > 1)

-- Multiple parameters (with index)
SELECT list_filter([1, 3, 1, 5], lambda x, i : x > i)

-- Method call syntax
SELECT [3, 4, 5, 6].list_filter(lambda x : x > 4)
```

## Test plan

- [x] Added `test_duckdb_lambda_function` test covering single param, multi-param, and method call syntax
- [x] All existing lambda tests pass (Databricks, ClickHouse, common)
- [x] All DuckDB tests pass

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) and and reviewed by me, @lovasoa, a human.

PR #2224, opened 2026-02-19
Resolve breaking change that set "supports_lambda_functions" on `GenericDialect`

alexander-beedie:
#2149 set "supports_lambda_functions" true for `DuckDbDialect` (which is fine), but also (accidentally?) set it for `GenericDialect` as well, breaking the existing use of `->` as a binary operator. 

This was previously identified in https://github.com/apache/datafusion-sqlparser-rs/pull/1686#discussion_r1934222955 as something that shouldn't be done; the `->` operator should remain a binary operator in `GenericDialect`. 

This PR keeps the `DuckDbDialect` addition, reverts the change to `GenericDialect`, and adds a line of additional coverage to try and prevent any future regression.

### DIFF
--- a/src/dialect/generic.rs
+++ b/src/dialect/generic.rs
@@ -225,10 +225,6 @@ impl Dialect for GenericDialect {
         true
     }
 
-    fn supports_lambda_functions(&self) -> bool {
-        true
-    }
-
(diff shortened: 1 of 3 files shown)


iffyio (approved):

PR #2227, opened 2026-02-21
Optimise prefix/identifier parsing

alexander-beedie:
## Description

* Tweaked `parse_expr_prefix_by_unreserved_word` to use `into_ident()` (move) instead of `to_ident()` (clone).
* Added a `NoKeyword` fast path in `parse_prefix` (most column references are likely plain identifiers).

## Results

Other `sqlparser_bench` tests just show noise (not really impacted by this change):
```
┌──────────────────────────────┬───────────┬───────────┬────────┐
│ benchmark                    │ baseline  │ this pr   │ change │
├──────────────────────────────┼───────────┼───────────┼────────┤
│ select_100_columns           │  67.93 µs │  64.32 µs │  −5.1% │
├──────────────────────────────┼───────────┼───────────┼────────┤
│ select_100_qualified_columns │ 140.92 µs │ 136.64 µs │  −3.0% │
├──────────────────────────────┼───────────┼───────────┼────────┤
│ parse_large_statement        │   4.50 ms │   4.47 ms │  −0.8% │
└──────────────────────────────┴───────────┴───────────┴────────┘
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1615,13 +1615,13 @@ impl<'a> Parser<'a> {
     /// Tries to parse an expression by a word that is not known to have a special meaning in the dialect.
     fn parse_expr_prefix_by_unreserved_word(
         &mut self,
-        w: &Word,
+        w: Word,
         w_span: Span,
     ) -> Result<Expr, ParserError> {
         let is_outer_join = self.peek_outer_join_operator();
         match &self.peek_token_ref().token {
             Token::LParen if !is_outer_join => {
-                let id_parts = vec![w.to_ident(w_span)];
+                let id_parts = vec![w.into_ident(w_span)];
                 self.parse_function(ObjectName::from(id_parts))
             }
             // string introducer https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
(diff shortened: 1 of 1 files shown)


alexander-beedie:
Rebased / conflicts resolved 👍

src/parser/mod.rs
```diff
@@ -1756,31 +1756,60 @@ impl<'a> Parser<'a> {
...
+                    let is_typed_lambda = matches!(peek, Token::Word(_))
+                        && self.dialect.supports_lambda_functions()
+                        && self.peek_nth_token_ref(1).token == Token::Arrow;
+                    if !is_special && !is_typed_lambda {
+                        Ok(Expr::Identifier(w.to_ident(span)))
+                    } else {
+                        // Non-keyword followed by special token (e.g. function call)
+                        let w = w.clone();
+                        Ok(self.parse_expr_prefix_by_unreserved_word(w, span)?)
+                    }
```
iffyio, line 1781: I think this block isnt ideal. its not clear what the special flag entails and why we're specifically looking ahead for a lambda function and I'm afraid that we'll likely accumulate similar special cases going forward

alexander-beedie: Sure, no problem. I can drop this part of the patch for now and see if there's a simpler way to get the benefits of the fast path 🤔

alexander-beedie: Actually, it looks like the `NoKeyword` fast-path was responsible for the vast majority of the performance gains here; the reduced patch would just save us a single String clone, which isn't measurable in benchmarks.

I'll close this one out for simplicity as I have several other more meaningful optimisations brewing 👍

PR #2228, opened 2026-02-21
MSSQL: Add support for OUTPUT clause on INSERT/UPDATE/DELETE

guan404ming:
## Why

MSSQL's OUTPUT clause allows returning inserted/deleted column values from DML statements, but was only supported on MERGE.

## How

- Add output: Option<OutputClause> to Insert, Delete, and Update AST structs with correct mid-statement positioning
- Reserve `OUTPUT` and `VALUES` keywords for table/column alias contexts to prevent ambiguous parsing
- Add tests covering `OUTPUT` with and without `INTO` on all three statement types

### DIFF
--- a/src/ast/dml.rs
+++ b/src/ast/dml.rs
@@ -80,6 +80,7 @@ pub struct Insert {
     /// RETURNING
     pub returning: Option<Vec<SelectItem>>,
     /// OUTPUT (MSSQL)
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql>
     pub output: Option<OutputClause>,
     /// Only for mysql
     pub replace_into: bool,
@@ -297,6 +298,7 @@ pub struct Delete {
     /// RETURNING
     pub returning: Option<Vec<SelectItem>>,
     /// OUTPUT (MSSQL)
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql>
     pub output: Option<OutputClause>,
     /// ORDER BY (MySQL)
     pub order_by: Vec<OrderByExpr>,
@@ -381,6 +383,7 @@ pub struct Update {
     /// RETURNING
     pub returning: Option<Vec<SelectItem>>,
     /// OUTPUT (MSSQL)
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql>
     pub output: Option<OutputClause>,
     /// SQLite-specific conflict resolution clause
     pub or: Option<SqliteOnConflict>,
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13310,13 +13310,7 @@ impl<'a> Parser<'a> {
 
         let from = self.parse_comma_separated(Parser::parse_table_and_joins)?;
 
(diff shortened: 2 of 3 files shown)


src/ast/dml.rs
```diff
@@ -77,6 +77,8 @@ pub struct Insert {
     pub on: Option<OnInsert>,
     /// RETURNING
     pub returning: Option<Vec<SelectItem>>,
+    /// OUTPUT (MSSQL)
```
iffyio, line 82: for these comments, could we include the link to the mssql docs describing the syntax? it would help with nagivation when folks look at the struct

src/parser/mod.rs
```diff
@@ -13288,6 +13288,15 @@ impl<'a> Parser<'a> {
         };
 
         let from = self.parse_comma_separated(Parser::parse_table_and_joins)?;
+
+        // MSSQL OUTPUT clause appears after FROM table, before USING/WHERE
+        // https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -17258,6 +17268,14 @@ impl<'a> Parser<'a> {
                     Default::default()
                 };
 
+                // MSSQL OUTPUT clause appears between columns and source
+                // https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -17475,6 +17501,15 @@ impl<'a> Parser<'a> {
         };
         self.expect_keyword(Keyword::SET)?;
         let assignments = self.parse_comma_separated(Parser::parse_assignment)?;
+
+        // MSSQL OUTPUT clause appears after SET, before FROM/WHERE
+        // https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql
```
```suggestion
```

maybe we can introduce a helper function `maybe_parse_output_clause()` that can be called from the relevant statement parsing functions?

tests/sqlparser_mssql.rs
```diff
@@ -2730,3 +2730,46 @@ fn parse_mssql_tran_shorthand() {
     // ROLLBACK TRAN normalizes to ROLLBACK (same as ROLLBACK TRANSACTION)
     ms().one_statement_parses_to("ROLLBACK TRAN", "ROLLBACK");
 }
+
+// MSSQL OUTPUT clause on INSERT/UPDATE/DELETE
+// https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql
+
```
iffyio:
```suggestion
```

src/parser/mod.rs
guan404ming: Yes, that would be better. Thanks, just updated!

src/ast/dml.rs
guan404ming: Sure, that makes sense and just added.

iffyio:
@guan404ming could you resolve the conflicts on the branch when you get the chance?

guan404ming:
> @guan404ming could you resolve the conflicts on the branch when you get the chance?

Yeah sure, just updated. Thanks!

guan404ming:
Thanks for your review!

PR #2239, opened 2026-02-25
Support `${placeholder}` syntax in tokenizer

cetra3:
Add support for dollar-brace placeholders (`${name}`, `${1}`, etc.) in the tokenizer, storing the full `${...}` string in the existing `Token::Placeholder` / `Value::Placeholder` variants with zero breaking changes. Unterminated `${name` produces a clear error.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1920,6 +1920,21 @@ impl<'a> Tokenizer<'a> {
 
         chars.next();
 
+        // Handle ${placeholder} syntax
+        if matches!(chars.peek(), Some('{')) {
+            chars.next(); // consume '{'
+            let placeholder = peeking_take_while(chars, |ch| ch != '}');
+            if matches!(chars.peek(), Some('}')) {
+                chars.next(); // consume '}'
+                return Ok(Token::Placeholder(format!("${{{placeholder}}}")));
+            } else {
+                return self.tokenizer_error(
+                    chars.location(),
+                    "Unterminated dollar-brace placeholder, expected '}'",
+                );
+            }
+        }
+
         // If the dialect does not support dollar-quoted strings, then `$$` is rather a placeholder.
         if matches!(chars.peek(), Some('$')) && !self.dialect.supports_dollar_placeholder() {
             chars.next();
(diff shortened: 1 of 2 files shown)


adriangb:
@iffyio could we bother your to take a look at this PR please? Thanks!

src/tokenizer.rs
```diff
@@ -1920,6 +1920,21 @@ impl<'a> Tokenizer<'a> {
 
         chars.next();
 
+        // Handle ${placeholder} syntax
+        if matches!(chars.peek(), Some('{')) {
```
iffyio, line 1924: wondering how this works with the `self.dialect.supports_dollar_placeholder()` dialects, is it possible that e.g. `${abc}$` is part of a dollar string or other potentially conflicting syntax?

src/tokenizer.rs
```diff
@@ -1920,6 +1920,21 @@ impl<'a> Tokenizer<'a> {
 
         chars.next();
 
+        // Handle ${placeholder} syntax
```
iffyio, line 1923: could we include a reference to the dialect that defines this syntax maybe [here](https://github.com/pydantic/datafusion-sqlparser-rs/blob/f16743ed2b41b299b748dfbef10c8262f308ab38/src/tokenizer.rs#L221)?

src/tokenizer.rs
cetra3: I could possibly add this as another flag to dialect if that makes sense?  I just thought it would be yet another flag, and if we could include it in this flag, but you're right it means that it will parse some dialects as valid even if it's not for that dialect

iffyio: I think if the syntax potentially conflicts we might need to ensure that it doesn't, because in this case its rather that `${abc}$` in the example will be parsed incorrectly as something that it is not (a placeholder vs a string)

cetra3: So there is some existing behaviour, whereby `$placeholder` will work currently with postgres and other dialects even though `supports_dollar_placeholder()` is `false` — it falls through to the placeholder fallback when the next character after the identifier isn't `$`.

The `${placeholder}` branch is consistent with that existing behavior, and there's no conflict with dollar-quoted strings since `{` is not a valid tag character

dollar-quoted string tags only allow alphanumeric characters and underscores (e.g., `$tag_1$...$tag_1$`), so `${abc}$` could never be interpreted as a dollar-quoted string.

src/tokenizer.rs
cetra3: This is a tough one, as it's not defined by a specific dialect but used in some of our other integrations around parametrised queries.  Namely we're using this with [perses](https://github.com/perses/perses)

PR #2240, opened 2026-02-25
Support two-argument `TRIM(string, characters)` in PostgreSQL

LucaCappelletti94:
PostgreSQL supports `TRIM(string, characters)` as a function form, but `sqlparser-rs` rejected it under `PostgreSqlDialect` with: `ParserError("Expected: ), found: ,")`.

This PR adds PostgreSQL support for the comma-style `TRIM` form and adds regression coverage.


src/parser/mod.rs
```diff
@@ -2962,7 +2962,13 @@ impl<'a> Parser<'a> {
                 trim_characters: None,
             })
         } else if self.consume_token(&Token::Comma)
-            && dialect_of!(self is DuckDbDialect | SnowflakeDialect | BigQueryDialect | GenericDialect)
+            && dialect_of!(
```
yoavcloud: Can you turn this into a function of the Dialect trait? For example: `self.dialect.supports_comma_separated_trim`.

tests/sqlparser_postgres.rs
```diff
@@ -8596,3 +8596,44 @@ fn parse_pg_analyze() {
         _ => panic!("Expected Analyze, got: {stmt:?}"),
     }
 }
+
+#[test]
+fn parse_postgres_two_argument_trim() {
```
yoavcloud: Perhaps this unit test can be moved to `sqlparser_common.rs` so it will cover all supporting dialects?

yoavcloud (changes requested):

src/parser/mod.rs
LucaCappelletti94: Ok, I will try to run a review of which dialects supports it, I hope it won't become a nightmare of little differences.

yoavcloud: I didn't mean to send you on a wild goose chase, just use whatever is already supported in code: DuckDbDialect, SnowflakeDialect, BigQueryDialect, GenericDialect and now PostgreSQL

LucaCappelletti94: I am currently just double checking the dialects I actively work with, just in case there are small differences. I found this missing feature via differential fuzzing and not with direct use, so I am not THAT familiar with it.

tests/sqlparser_postgres.rs
LucaCappelletti94: Ok done

src/parser/mod.rs
LucaCappelletti94: And done, a couple of dialects were still missing, it was a good thing to spend a bit more time to check these.

yoavcloud (approved):

yoavcloud:
Thank you @LucaCappelletti94!

Issue #2236, opened 2026-02-25
HiveQL, CTEs and FROM first SELECT statements


Viicos:
https://github.com/apache/datafusion-sqlparser-rs/pull/235 introduced support for HiveQL, and modified how CTEs are parsed to [parse an additional `FROM` keyword](https://github.com/apache/datafusion-sqlparser-rs/commit/9d9d681cbabf31d1d07ad166da7a1fd87d07d960#diff-4a04259da480a6b794a2e947e4cc03eff4d1aa9330836f5b91cac68c5398193fR2182-L2173).

A user [reported some questions](https://github.com/apache/datafusion-sqlparser-rs/pull/235#issuecomment-1189199817) on the PR, and looking at the documentation links provided, it seems like HiveQL has the ability to use `FROM` directly after a CTE, but it is unclear what for.
[This link](https://docs-archive.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_create_a_table_using_a_cte.html) shows an example to insert from a CTE, and [this one](https://cwiki.apache.org/confluence/display/Hive/Common+Table+Expression) shows a `SELECT` statement, using the _FROM_ first variant (but it also seems like the Hive dialect doesn't has [`supports_from_first_select()`](https://github.com/apache/datafusion-sqlparser-rs/blob/d9b53a0cdb369124d9b6ce6237959e66bad859af/src/dialect/mod.rs#L640-L649)?).

The issue is that when using the generic dialect (or dialects supporting from first), the parsing of the FROM keyword breaks, e.g.:

```sql
WITH test AS (FROM t SELECT a) FROM test SELECT a
```

The AST looks like (reduced for visibility):

```rs
Query(
    Query {
        with: Some(
            With {
                with_token: TokenWithSpan {
                    token: Word(
                        Word {
                            value: "WITH",
                            quote_style: None,
                            keyword: WITH,
                        },
                    ),
                    span: Span(Location(1,1)..Location(1,5)),
                },
                recursive: false,
                cte_tables: [
                    Cte {
                        alias: TableAlias {
                            name: Ident {
                                value: "test",
                                quote_style: None,
                                span: Span(Location(1,6)..Location(1,10)),
                            },
                            columns: [],
                        },
                        query: Query {
                            with: None,
                            body: Select(
                                Select {
                                    select_token: Some(
                                        TokenWithSpan {
                                            token: Word(
                                                Word {
                                                    value: "SELECT",
[...]

PR #2241, opened 2026-02-25
Only parse FROM identifier in CTE if using Hive

Viicos:
Fixes https://github.com/apache/datafusion-sqlparser-rs/issues/2236.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -13958,7 +13958,7 @@ impl<'a> Parser<'a> {
                 closing_paren_token: closing_paren_token.into(),
             }
         };
-        if dialect_of!(self is HiveDialect) && self.parse_keyword(Keyword::FROM) {
+        if self.dialect.supports_from_first_insert() && self.parse_keyword(Keyword::FROM) {
             cte.from = Some(self.parse_identifier()?);
         }
         Ok(cte)
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -16069,7 +16069,6 @@ fn test_select_from_first() {
             pipe_operators: vec![],
         };
         assert_eq!(expected, ast);
-        assert_eq!(ast.to_string(), q);
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/parser/mod.rs
```diff
@@ -13982,7 +13958,7 @@ impl<'a> Parser<'a> {
                 closing_paren_token: closing_paren_token.into(),
             }
         };
-        if self.parse_keyword(Keyword::FROM) {
+        if dialect_of!(self is HiveDialect) && self.parse_keyword(Keyword::FROM) {
```
iffyio: not sure I follow the intent of the PR, is it rather that hive should support `supports_from_first_select` (i.e. what's the difference in behavior from hive vs other dialects)?

Viicos: I gave more context in the issue I created; I'm not knowledgeable enough about Hive to understand if it should support FROM first selects, and it is unclear (see also https://github.com/apache/datafusion-sqlparser-rs/pull/235#issuecomment-1189199817) why this `FROM` token parsing was introduced here. Maybe the intent was indeed to support FROM first selects, but I'd rather not break things here so that's why I only gated this `FROM` parsing for Hive.

Viicos:
Hi @iffyio,

Just a gentle ping on this one, as it prevents properly parsing FROM first queries when used with a CTE, which can be fairly common considering FROM first queries are enabled on the generic dialect. Thanks in advance!

src/parser/mod.rs
iffyio: I see, oh yeah afaict it seems to be the case that from [the link](https://docs-archive.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_create_a_table_using_a_cte.html) the syntax was misinterpreted, the `FROM` doesn't belong to the CTE, rather its a variant of from_first (at least for insert statements).

I think ideally we would introduce a `self.dialect.supports_from_first_insert` method that hive flags and the `from` being set here moves from the cte to the insert statement. The latter might be out of scope for the MR however, so maybe we can just add a comment in the hive dialect method on the former mentioning that its incomplete/incorrect?

tests/sqlparser_common.rs
```diff
@@ -16118,6 +16073,129 @@ fn test_select_from_first() {
...
+        limit_clause: None,
+        fetch: None,
+        locks: vec![],
+        for_clause: None,
+        settings: None,
+        format_clause: None,
+        pipe_operators: vec![],
+    };
+    assert_eq!(expected, ast);
+    assert_eq!(ast.to_string(), q);
```
iffyio: I don't think this should be needed given `verified_query()` does this check?

tests/sqlparser_common.rs
```diff
@@ -16118,6 +16073,129 @@ fn test_select_from_first() {
...
 }
 
+#[test]
+fn test_select_from_first_with_cte() {
+    let dialects = all_dialects_where(|d| d.supports_from_first_select());
+    let q = "WITH test AS (FROM t SELECT a) FROM test SELECT 1";
+
+    let ast = dialects.verified_query(q);
+
+    let expected = Query {
```
iffyio: I wonder are we asserting something specific about the AST? It would be good to skip the assertion since the struct is quite verbose - thinking if its sufficient to only assert that the query parses and round trip. Otherwise if we can rewrite/split the assertion to only assert the components we're interested in

tests/sqlparser_common.rs
Viicos: I followed the existing [`test_select_from_first()`](https://github.com/apache/datafusion-sqlparser-rs/blob/67d684b1a0cf0a9da149fb7733fbdeecf70cf7f1/tests/sqlparser_common.rs#L16162) test for this one, but this makes sense. I removed the assertion from both tests.

tests/sqlparser_common.rs
Viicos: I simplified to only assert the query body's projection and from clauses.

src/parser/mod.rs
Viicos: I'm ok with this option, I can also create a follow up issue along with the comment (or just the comment if you are happy with it).

iffyio: An issue sounds great as well thanks!

PR #2245, opened 2026-02-26
Add support for parsing COPY statements from STDIN without a semicolon

LucaCappelletti94:
This PR fixes PostgreSQL `COPY ... FROM STDIN` parsing when the statement ends at EOF without a trailing semicolon.

This is one of several failure families I have found in the course of the correctness benchmarks I am running based on resources such as the PostgreSQL regression SQL suite.

Before this change, sqlparser rejected valid inputs such as:

```sql
COPY bitwise_test FROM STDIN NULL 'null'
```

with:

```text
ParserError("Expected: ;, found: EOF")
```

## Root Cause
`parse_copy` unconditionally required `;` for `CopyTarget::Stdin` before reading optional inline payload rows.

## Fix
For `COPY ... FROM STDIN`:
- accept immediate `EOF` as a valid terminator when no inline payload is present
- keep existing behavior for semicolon + payload parsing unchanged

## Tests
Added/updated PostgreSQL regression coverage:
- `parse_copy_from_stdin_without_semicolon`
- `parse_copy_from_stdin_without_semicolon_variants`

The variants cover multiple `FROM stdin` shapes without trailing semicolon (columns, legacy options, mixed casing), and assert they parse as `Statement::Copy` with `target: Stdin` and empty payload.


tests/sqlparser_postgres.rs
```diff
@@ -1084,6 +1084,61 @@ PHP	₱ USD $
...
+        "copy tab_progress_reporting from stdin",
+        "copy oversized_column_default from stdin",
+        "COPY x (a, b, c, d, e) from stdin",
+        "copy header_copytest (c, a) from stdin",
+        "COPY atest5 (two) FROM stdin",
+        "COPY main_table (a, b) FROM stdin",
+    ];
+
+    for sql in cases {
+        match pg().one_statement_parses_to(sql, "") {
```
iffyio: can this test use `verified_stmt` instead? Also, what's the difference between this test and `parse_copy_from_stdin_without_semicolon`?

src/parser/mod.rs
```diff
@@ -11028,8 +11028,12 @@ impl<'a> Parser<'a> {
             legacy_options.push(opt);
         }
         let values = if let CopyTarget::Stdin = target {
-            self.expect_token(&Token::SemiColon)?;
-            self.parse_tsv()
+            if self.peek_token_ref().token == Token::EOF {
```
iffyio: can this be collapsed to `if let copy && self.peek_token() != EOF`?

tests/sqlparser_postgres.rs
LucaCappelletti94: Done, and these are just a handful of distinct cases from the postgres regression SQL statements.

src/parser/mod.rs
LucaCappelletti94: With matches, yes. I cannot use chained if let && because we are still using 2021 and not 2024

iffyio (approved):

PR #2246, opened 2026-02-26
Add support for PostgreSQL's ORDER BY ... USING <operator> clause

LucaCappelletti94:
This PR adds parser and AST support for PostgreSQL `USING` operators in `ORDER BY` expressions, including:

- top-level `ORDER BY`
- aggregate argument list `ORDER BY` (e.g. inside `aggfns(...)` / `string_agg(...)` style clauses)

## Problem
`sqlparser-rs` previously rejected valid PostgreSQL statements containing:

- `ORDER BY <expr> USING <operator>`

with errors like:

- `Expected: ), found: using`

This blocked parsing of several PostgreSQL regression-style aggregate statements.

## Root Cause
`parse_order_by_expr_inner` only supported:

- `ASC | DESC`
- `NULLS FIRST | NULLS LAST`

and had no branch to parse `USING <operator>`.

## Dialect Scope Decision
`ORDER BY ... USING <operator>` is intentionally enabled for PostgreSQL only.

Rationale:
- PostgreSQL explicitly documents `ORDER BY expression [ USING operator ] ...` in `SELECT`.
- Other dialects supported by this repository generally document `ORDER BY` with expression + `ASC|DESC` (+ null ordering), but not `USING <operator>`.

References:
- PostgreSQL: https://www.postgresql.org/docs/current/sql-select.html
- Redshift: https://docs.aws.amazon.com/redshift/latest/dg/r_ORDER_BY_clause.html
- Snowflake: https://docs.snowflake.com/en/sql-reference/constructs/order-by
- BigQuery: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/query-expression.html
- SQLite: https://www.sqlite.org/lang_select.html
- DuckDB: https://duckdb.org/docs/stable/sql/query_syntax/orderby.html
- Oracle: https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/SELECT.html
- SQL Server: https://learn.microsoft.com/sql/t-sql/queries/select-order-by-clause-transact-sql

### PostgreSQL: accepted
```sql
SELECT a FROM t ORDER BY a USING <;

SELECT a FROM t ORDER BY a USING OPERATOR(pg_catalog.<) NULLS LAST;

SELECT aggfns(DISTINCT a, a, c ORDER BY c USING ~<~, a)
FROM (VALUES (1,3,'foo'),(0,NULL,NULL),(2,2,'bar'),(3,1,'baz')) v(a,b,c),
     generate_series(1,2) i;

CREATE OR REPLACE VIEW agg_view1 AS
SELECT aggfns(a,b,c ORDER BY c USING ~<~)
FROM (VALUES (1,3,'foo'),(0,NULL,NULL),(2,2,'bar'),(3,1,'baz')) v(a,b,c);
```

### PostgreSQL: rejected (invalid syntax)
```sql
SELECT a FROM t ORDER BY a USING ;
-- Error: Expected an ordering operator after USING

SELECT a FROM t ORDER BY a USING OPERATOR();
-- Error: Expected an operator name

SELECT a FROM t ORDER BY a USING < DESC;
-- Error: ASC/DESC cannot be used together with USING in ORDER BY
```


src/parser/mod.rs
```diff
@@ -18096,23 +18110,76 @@ impl<'a> Parser<'a> {
...
+        let token = self.next_token();
+        let operator = token.token.to_string();
+        if Self::is_valid_order_by_using_operator_symbol(&operator) {
+            Ok(ObjectName::from(vec![Ident::new(operator)]))
+        } else {
+            self.expected_ref("an ordering operator after USING", &token)
+        }
+    }
+
+    fn is_valid_order_by_using_operator_symbol(symbol: &str) -> bool {
```
iffyio: is this code necessary?

src/parser/mod.rs
```diff
@@ -18083,7 +18072,32 @@ impl<'a> Parser<'a> {
             None
         };
 
-        let options = self.parse_order_by_options()?;
+        let using_operator = if !with_operator_class
+            && dialect_of!(self is PostgreSqlDialect)
```
iffyio: if we need to dialect guard this, I think we can introduce a dialect method instead of the macro

src/parser/mod.rs
```diff
@@ -4990,10 +4989,10 @@ impl<'a> Parser<'a> {
...
             match &self.peek_nth_token_ref(0).token {
                 Token::EOF => break,
-                Token::Word(w) => {
-                    if w.quote_style.is_none() && terminal_keywords.contains(&w.keyword) {
-                        break;
-                    }
+                Token::Word(w)
+                    if w.quote_style.is_none() && terminal_keywords.contains(&w.keyword) =>
+                {
+                    break;
```
iffyio, line 5036: are these diffs required for this PR?

tests/sqlparser_postgres.rs
```diff
@@ -5685,6 +5687,85 @@ fn parse_array_agg() {
...
+
+    let err = pg()
+        .parse_sql_statements("SELECT a FROM t ORDER BY a USING < DESC;")
+        .unwrap_err();
+    assert!(
+        matches!(err, ParserError::ParserError(msg) if msg.contains("ASC/DESC cannot be used together with USING in ORDER BY"))
+    );
+
+    // `USING` in ORDER BY is PostgreSQL-specific and should not parse in GenericDialect.
+    let generic = TestedDialects::new(vec![Box::new(GenericDialect {})]);
```
iffyio: note, per the dialect comment, we'd want to avoid hardcoding dialects in tests

LucaCappelletti94:
I’ve applied the requested changes and force-pushed the branch, so to remove irrelevant clippy changes from the commits.

- Replaced the hardcoded parser guard `(dialect_of!(...PostgreSqlDialect))` with a dialect capability method: `Dialect::supports_order_by_using_operator()`, enabled in `PostgreSqlDialect` only. I have also spent a bit of time checking yet again if any other dialects had it, but so far no other dialects support it.
- Removed the dedicated `is_valid_order_by_using_operator_symbol` helper and now validate USING operators through existing dialect behavior `(dialect.is_custom_operator_part(...))`, so we don’t duplicate operator-character rules in parser code.
- Trimmed the PR scope/history to a single focused commit and removed unrelated nightly clippy code smell fixes - I hadn't realized it was a nightly only rant.

src/parser/mod.rs
LucaCappelletti94: Removed, `parse_order_by_using_operator` now uses `dialect.is_custom_operator_part(...)` directly. No separate symbol-check helper anymore. I wasn't aware there was a dialect helper method.

src/parser/mod.rs
LucaCappelletti94: Added `supports_order_by_using_operator()` on Dialect (default false), enabled in `PostgreSqlDialect`, and parser now checks that method instead of `dialect_of!`. I double checked that no other dialect supports it at this time.

src/parser/mod.rs
LucaCappelletti94: No, it was nightly clippy fixes. Removed from git history.

tests/sqlparser_postgres.rs
LucaCappelletti94: Replaced the `GenericDialect` negative case with a local wrapper dialect that behaves like Postgres but returns false for `supports_order_by_using_operator()`.

src/parser/mod.rs
```diff
@@ -18156,7 +18156,32 @@ impl<'a> Parser<'a> {
...
+            {
+                return parser_err!(
+                    "ASC/DESC cannot be used together with USING in ORDER BY".to_string(),
+                    self.peek_token_ref().span.start
+                );
+            }
+            OrderByOptions {
+                asc: None,
+                nulls_first: self.parse_order_by_nulls_first_last(),
+            }
```
iffyio: would it be possible to use something like this representation instead?

```rust
pub enum OrderByExpr {
    Asc,
    Desc,
    Using(...),
    UsingOperator(...),
}

pub struct OrderByOptions {
    pub expr: Option<OrderByExpr>,
    pub nulls_first: Option<bool>,
}
```

tests/sqlparser_postgres.rs
```diff
@@ -5727,6 +5729,101 @@ fn parse_array_agg() {
...
+        }
+
+        fn supports_order_by_using_operator(&self) -> bool {
+            false
+        }
+    }
+
+    let without_order_by_using = TestedDialects::new(vec![Box::new(OrderByUsingDisabledDialect)]);
+    assert!(without_order_by_using
+        .parse_sql_statements("SELECT a FROM t ORDER BY a USING <;")
```
iffyio: I think we can have the tests in common.rs instead to cover dialects via the `all_dialects_where(|d| d.support*)` pattern. Since the impl is guarded by the dialect method vs explicit to postgres

src/ast/query.rs
```diff
@@ -2860,6 +2860,8 @@ impl fmt::Display for OrderBy {
 pub struct OrderByExpr {
     /// The expression to order by.
     pub expr: Expr,
+    /// Optional PostgreSQL `USING <operator>` clause.
```
iffyio: can we include reference doc to maybe this?
https://www.postgresql.org/docs/current/sql-select.html

so folks know where to find this feature easily

src/parser/mod.rs
```diff
@@ -18169,23 +18194,61 @@ impl<'a> Parser<'a> {
...
+                return self.expected_ref("an operator name", self.peek_token_ref());
+            };
+            let operator = last_part.to_string();
+            if operator.is_empty()
+                || !operator
+                    .chars()
+                    .all(|ch| dialect.is_custom_operator_part(ch))
+            {
+                return self.expected_ref("an operator name", self.peek_token_ref());
+            }
```
iffyio: the `to_string()` cloning doesn't seem to be required here? But in general this logic iiuc is validating some property of the allowed naming? if so I think it would make sense to skip the validation here in the parser (same for the similar logic below, so that if the code is to be kept then we likely want to pull it out to a reusable function)

src/parser/mod.rs
```diff
@@ -18156,7 +18156,32 @@ impl<'a> Parser<'a> {
...
+                .is_some()
+            {
+                return parser_err!(
+                    "ASC/DESC cannot be used together with USING in ORDER BY".to_string(),
+                    self.peek_token_ref().span.start
+                );
+            }
+            OrderByOptions {
+                asc: None,
+                nulls_first: self.parse_order_by_nulls_first_last(),
```
iffyio: I think we can skip this validation

src/parser/mod.rs
```diff
@@ -18169,23 +18194,61 @@ impl<'a> Parser<'a> {
...
+                .chars()
+                .all(|ch| dialect.is_custom_operator_part(ch))
+        {
+            Ok(ObjectName::from(vec![Ident::new(operator)]))
+        } else {
+            self.expected_ref("an ordering operator after USING", &token)
+        }
+    }
+
+    fn parse_order_by_nulls_first_last(&mut self) -> Option<bool> {
```
```suggestion
    fn parse_null_ordering_modifier(&mut self) -> Option<bool> {
```

src/ast/query.rs
iffyio: fyi: re the comment would be nice to include the link

tests/sqlparser_postgres.rs
```diff
@@ -5727,6 +5729,101 @@ fn parse_array_agg() {
     pg().verified_stmt(sql4);
 }
 
+#[test]
+fn parse_pg_aggregate_order_by_using_operator() {
+    let sql = "SELECT aggfns(DISTINCT a, a, c ORDER BY c USING ~<~, a) FROM t";
```
iffyio: can we move the tests to common and use the `all_dialects_where` helper method for the test cases?

iffyio (approved):

LucaCappelletti94:
Ok merged main and fixed the collisions

PR #2251, opened 2026-02-27
PostgreSQL `UNLOGGED` Table Support and `ALTER TABLE ... SET LOGGED|UNLOGGED`

LucaCappelletti94:
This change adds parser and AST support for PostgreSQL table logging controls:

1. `CREATE UNLOGGED TABLE ...`
2. `ALTER TABLE ... SET LOGGED`
3. `ALTER TABLE ... SET UNLOGGED`

It also adds regression tests and updates cross-dialect test fixtures impacted by the AST shape change.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -5231,6 +5231,10 @@ impl<'a> Parser<'a> {
             .parse_one_of_keywords(&[Keyword::TEMP, Keyword::TEMPORARY])
             .is_some();
         let volatile = self.parse_keyword(Keyword::VOLATILE);
+        let unlogged = self.peek_keywords(&[Keyword::UNLOGGED, Keyword::TABLE]);
+        if unlogged {
+            self.expect_keyword(Keyword::UNLOGGED)?;
+        }
         let persistent = dialect_of!(self is DuckDbDialect)
             && self.parse_one_of_keywords(&[Keyword::PERSISTENT]).is_some();
         let create_view_params = self.parse_create_view_params()?;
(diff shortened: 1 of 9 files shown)


src/parser/mod.rs
```diff
@@ -508,10 +508,10 @@ impl<'a> Parser<'a> {
                 Token::EOF => break,
 
                 // end of statement
-                Token::Word(word) => {
-                    if expecting_statement_delimiter && word.keyword == Keyword::END {
-                        break;
-                    }
+                Token::Word(word)
```
iffyio, line 548: @LucaCappelletti94 I think this PR has the same clippy issues?

LucaCappelletti94: Yeah, likely several of them, I just used clippy --fix.

LucaCappelletti94: I will do a pass tomorrow and remove the clippy everywhere

src/parser/mod.rs
```diff
@@ -5108,14 +5108,18 @@ impl<'a> Parser<'a> {
...
             && self.parse_one_of_keywords(&[Keyword::PERSISTENT]).is_some();
         let create_view_params = self.parse_create_view_params()?;
         if self.peek_keywords(&[Keyword::SNAPSHOT, Keyword::TABLE]) {
             self.parse_create_snapshot_table().map(Into::into)
         } else if self.parse_keyword(Keyword::TABLE) {
-            self.parse_create_table(or_replace, temporary, global, transient)
+            self.parse_create_table(or_replace, temporary, unlogged, global, transient)
                 .map(Into::into)
+        } else if unlogged {
+            self.expected_ref("TABLE after UNLOGGED", self.peek_token_ref())
```
iffyio: I think this shouldn't be needed, the else branch covers the scenario where there is no match?

src/parser/mod.rs
```diff
@@ -5108,14 +5108,18 @@ impl<'a> Parser<'a> {
         let temporary = self
             .parse_one_of_keywords(&[Keyword::TEMP, Keyword::TEMPORARY])
             .is_some();
+        let unlogged = dialect_of!(self is PostgreSqlDialect | GenericDialect)
+            && self.parse_keyword(Keyword::UNLOGGED);
```
iffyio: is there a requirement to use a dialect guard? if not we can always accept the syntax

LucaCappelletti94:
@iffyio any further comments?

src/parser/mod.rs
```diff
@@ -5238,6 +5238,12 @@ impl<'a> Parser<'a> {
             self.parse_create_snapshot_table().map(Into::into)
         } else if self.peek_keywords(&[Keyword::TEXT, Keyword::SEARCH]) {
             self.parse_create_text_search().map(Into::into)
+        } else if self.peek_keywords(&[Keyword::UNLOGGED, Keyword::TABLE]) {
+            self.expect_keywords(&[Keyword::UNLOGGED, Keyword::TABLE])?;
+            let mut create_table = self
+                .parse_create_table(or_replace, temporary, global, transient, volatile, multiset)?;
```
iffyio: can this instead use the same pattern as temporary/volatile etc, that we flag the bool beforehand, so that we reuse the same call to parse_create_table?

LucaCappelletti94: Ok now `unlogged` is flagged up front and reuses the single `parse_create_table` call.

I kept a `peek_keywords([UNLOGGED, TABLE])` guard instead of an unconditional `parse_keyword` like `volatile`. Unconditional consume means every other `CREATE` branch swallows a stray `UNLOGGED` and drops it:

```
CREATE UNLOGGED VIEW v AS SELECT 1   => CREATE VIEW v AS SELECT 1
CREATE UNLOGGED INDEX idx ON t (a)   => CREATE INDEX idx ON t(a)
CREATE UNLOGGED DATABASE d           => CREATE DATABASE d
CREATE UNLOGGED SEQUENCE seq         => CREATE SEQUENCE seq   -- valid PG, modifier silently lost
```

The peek keeps these erroring via the existing final `else`.

iffyio (approved):

PR #2252, opened 2026-02-27
Adding support for PostgreSQL's XML functions

LucaCappelletti94:
This PR adds PostgreSQL XML expression support to `sqlparser-rs` by introducing dedicated AST variants (`Expr::Xml*`) and parser paths for `XMLCONCAT`, `XMLELEMENT`, `XMLFOREST`, `XMLPARSE`, `XMLPI`, `XMLROOT`, and `XMLSERIALIZE`, aligned with PostgreSQL XML function syntax ([PostgreSQL XML functions](https://www.postgresql.org/docs/current/functions-xml.html))

Implementation is intentionally enabled for both `PostgreSqlDialect` and `GenericDialect`, while non-target dialects keep treating `xml*` names as ordinary function identifiers and reject XML-special grammar forms.

The test updates explicitly lock these dialect boundaries, add AST-shape coverage for all new XML nodes, and prefer structural assertions over formatting-sensitive string checks.

This PR DOES NOT add support for XML parsing of course, which remains something that the user should validate, since PostgreSQL itself will accept (and fail at runtime) queries with invalid XML. Whether we should raise errors there is debatable.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -508,10 +508,10 @@ impl<'a> Parser<'a> {
                 Token::EOF => break,
 
                 // end of statement
-                Token::Word(word) => {
-                    if expecting_statement_delimiter && word.keyword == Keyword::END {
-                        break;
-                    }
+                Token::Word(word)
+                    if expecting_statement_delimiter && word.keyword == Keyword::END =>
+                {
+                    break;
                 }
                 _ => {}
             }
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -1716,6 +1715,17 @@ impl<'a> Parser<'a> {
                 // name, resulting in `NOT 'a'` being recognized as a `TypedString` instead of
                 // an unary negation `NOT ('a' LIKE 'b')`. To solve this, we don't accept the
                 // `type 'string'` syntax for the custom data types at all.
+                DataType::Custom(type_name, modifiers)
+                    if dialect_of!(self is PostgreSqlDialect | GenericDialect)
```
yoavcloud, line 1719: Can we use a dialect method for this?

yoavcloud (changes requested):
Hey @LucaCappelletti94. There are many more changes in this PR than are required to address the missing functionality, which makes code review harder.

I'm all for using coding agents to speed up the development of the project, but the bottleneck now becomes reviewing the PRs.

We should come up with guidelines for the proper usage of coding agents and I will do my best to advance this with the project members. In the meantime, can you please prepare a lighter version of this change?

LucaCappelletti94:
TLDR: I was trying to add these missing features as fast as the postgres differential fuzzer I had was finding them, and I ended up rushing bad code. Then the fuzzer found some goldmine of hundreds of missing things (which I decided to not look into it for the time) and I gave up with the PR series. **I will explode this PR in tiny focused ones.**

I will publish that fuzzer in another repo so that it can be used to gauge completes. It is nothing fancy, just honggfuzzer seeded with Postgres corpus of statements.

LucaCappelletti94:
Here is the first exploded PR btw: https://github.com/apache/datafusion-sqlparser-rs/pull/2299

LucaCappelletti94:
Step 2: https://github.com/apache/datafusion-sqlparser-rs/pull/2301

LucaCappelletti94:
Step 3: https://github.com/apache/datafusion-sqlparser-rs/pull/2368

LucaCappelletti94:

PR #2253, opened 2026-02-27
Adding support for TABLESPACE

LucaCappelletti94:
This PR adds cross-dialect `TABLESPACE` support by implementing PostgreSQL forms (`CREATE/ALTER/DROP TABLESPACE`) and MySQL forms (`CREATE/ALTER/DROP [UNDO] TABLESPACE`) with Generic dialect handling both syntaxes as a union parser, using explicit AST modeling instead of loose keyword parsing: `CreateTablespace` now carries dialect-specific definitions, MySQL `ALTER TABLESPACE` is represented as an ordered list of options to preserve multi-clause statements, and `DROP TABLESPACE` has a dedicated statement type to avoid over-accepting generic `DROP` behaviors and to keep dialect rules explicit (for example, PostgreSQL `IF EXISTS` vs MySQL-specific clauses)

Syntax and behavior were aligned against the official documentation for [PostgreSQL CREATE TABLESPACE](https://www.postgresql.org/docs/current/sql-createtablespace.html), [PostgreSQL ALTER TABLESPACE](https://www.postgresql.org/docs/current/sql-altertablespace.html), [PostgreSQL DROP TABLESPACE](https://www.postgresql.org/docs/current/sql-droptablespace.html), [MySQL CREATE TABLESPACE](https://dev.mysql.com/doc/refman/8.4/en/create-tablespace.html), [MySQL ALTER TABLESPACE](https://dev.mysql.com/doc/refman/8.4/en/alter-tablespace.html), and [MySQL DROP TABLESPACE](https://dev.mysql.com/doc/refman/8.4/en/drop-tablespace.html), with coverage added in both MySQL and PostgreSQL parser tests.

**If anybody's wondering where all of these PRs are coming from, I am running differential fuzzing against sqlparser vs postgres 18. It's a bit of a blood bath.**

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -50,7 +50,7 @@ use crate::ast::{
     OrderByExpr, ProjectionSelect, Query, RefreshModeKind, ResetConfig, RowAccessPolicy,
     SequenceOptions, Spanned, SqlOption, StorageLifecyclePolicy, StorageSerializationPolicy,
     TableVersion, Tag, TriggerEvent, TriggerExecBody, TriggerObject, TriggerPeriod,
-    TriggerReferencing, Value, ValueWithSpan, WrappedCollection,
+    TriggerReferencing, UtilityOption, Value, ValueWithSpan, WrappedCollection,
 };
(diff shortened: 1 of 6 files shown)


LucaCappelletti94:

PR #2254, opened 2026-02-27
Adding support for `PostgreSQL`'s `SUBSCRIPTION` (`CREATE`, `ALTER`, `DROP`, `COMMENT`)

LucaCappelletti94:
This PR adds PostgreSQL parsing support for subscription statements (`CREATE SUBSCRIPTION`, `ALTER SUBSCRIPTION`, `DROP SUBSCRIPTION`, and `COMMENT ON SUBSCRIPTION`) according to the upstream syntax documented in [`CREATE SUBSCRIPTION`](https://www.postgresql.org/docs/current/sql-createsubscription.html), [`ALTER SUBSCRIPTION`](https://www.postgresql.org/docs/current/sql-altersubscription.html), [`DROP SUBSCRIPTION`](https://www.postgresql.org/docs/current/sql-dropsubscription.html), and [`COMMENT`](https://www.postgresql.org/docs/current/sql-comment.html), and it also aligns privilege parsing with PostgreSQL role-style grant/revoke usage from [`GRANT`](https://www.postgresql.org/docs/current/sql-grant.html) and [`REVOKE`](https://www.postgresql.org/docs/current/sql-revoke.html): implementation-wise, the parser now uses dedicated subscription AST nodes/operations, enforces subscription names as single identifiers (matching PostgreSQL object naming behavior for subscriptions), handles predefined PostgreSQL roles in `GRANT`/`REVOKE` via a PostgreSQL-only `pg_*` path instead of a broad privilege fallback (to preserve typo detection for regular privilege keywords), and gates `RESET SESSION AUTHORIZATION` to PostgreSQL semantics as described in [`RESET`](https://www.postgresql.org/docs/current/sql-reset.html).

**If anybody's wondering where all of these PRs are coming from, I am running differential fuzzing against sqlparser vs postgres 18. It's a bit of a blood bath. Maybe we should have at least a unit test running against postgres's regression statements?**


LucaCappelletti94:
Closing this PR to fraction it into fragment PRs. Ideally, I would love to use stacked PRs, but alas they are not available for out-of-repo branches.

PR #2255, opened 2026-02-27
PostgreSQL SQL/JSON Parser Support

LucaCappelletti94:
This PR adds PostgreSQL SQL/JSON parsing coverage for `JSON_EXISTS`, `JSON_VALUE`, `JSON_QUERY`, and table-level `JSON_TABLE ... ON ERROR`, following the PostgreSQL SQL/JSON query-function and JSON_TABLE syntax documented in the official manuals ([SQL/JSON query functions](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-SQLJSON-QUERY), [JSON_TABLE](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-SQLJSON-TABLE)). Implementation-wise, the parser now uses function-aware clause parsing (instead of globally accepting these clauses in all function calls), introduces explicit AST clause variants for `PASSING`, `ON EMPTY/ON ERROR`, wrapper/quote behavior, and `FORMAT JSON`, and extends `TableFactor::JsonTable` with an optional table-level error-handling field.

This keeps round-trip formatting deterministic, and avoids broad dialect side effects. I left `IS JSON` exprs outside of this PR for now.

**If anybody's wondering where all of these PRs are coming from, I am running differential fuzzing against sqlparser vs postgres 18. It's a bit of a blood bath.**


LucaCappelletti94:

PR #2256, opened 2026-02-27
Adding support for `IS JSON` & `IS NOT JSON`

LucaCappelletti94:
This PR adds support for the SQL/JSON predicate syntax `expression IS [NOT] JSON [VALUE|SCALAR|ARRAY|OBJECT] [WITH|WITHOUT UNIQUE [KEYS]]`, aligning with the SQL/JSON testing syntax documented by PostgreSQL ([JSON Functions and Operators, Table 9.50](https://www.postgresql.org/docs/current/functions-json.html)) and Oracle ([SQL/JSON Conditions IS JSON and IS NOT JSON](https://docs.oracle.com/en/database/oracle/oracle-database/23/adjsn/sql-json-conditions-is-json-and-is-not-json.html)).

Implementation-wise, it introduces `Expr::IsJson` plus `JsonPredicateType` and `JsonKeyUniqueness` in the AST, parses the full optional modifier chain, and gates support behind a new dialect capability (`supports_is_json_predicate()`) enabled for Generic, ANSI, PostgreSQL, and Oracle so unsupported dialect behavior and hints remain explicit. I have included ANSI as this method in theory was introduced there in 2016.

We also accept optional `KEYS` (canonicalizing to include it on display) and added coverage for both `IS JSON` and `IS NOT JSON`, malformed/junk inputs, and duplicate uniqueness clauses.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -10996,45 +10996,83 @@ fn parse_is_boolean() {
     verified_stmt("SELECT f FROM foo WHERE field IS UNKNOWN");
     verified_stmt("SELECT f FROM foo WHERE field IS NOT UNKNOWN");
 
-    let sql = "SELECT f from foo where field is 0";
-    let res = parse_sql_statements(sql);
+    for sql in [
+        "SELECT f from foo where field is 0",
+        "SELECT s, s IS XYZ NORMALIZED FROM foo",
+        "SELECT s, s IS NFKC FROM foo",
+        "SELECT s, s IS TRIM(' NFKC ') FROM foo",
+    ] {
+        assert!(
+            parse_sql_statements(sql).is_err(),
+            "expected a parse failure for `{sql}`"
+        );
+    }
+}
+
+#[test]
+fn parse_is_json_predicate() {
+    use self::Expr::*;
+
+    // Assert the full AST once for a case that exercises every field.
+    let sql = "a IS NOT JSON OBJECT WITHOUT UNIQUE KEYS";
     assert_eq!(
-        ParserError::ParserError(
-            "Expected: [NOT] NULL | TRUE | FALSE | DISTINCT | [form] NORMALIZED FROM after IS, found: 0"
-                .to_string()
-        ),
-        res.unwrap_err()
+        IsJson {
+            expr: Box::new(Identifier(Ident::new("a"))),
+            kind: Some(JsonPredicateType::Object),
+            unique_keys: Some(JsonKeyUniqueness::WithoutUniqueKeys),
+            negated: true,
+        },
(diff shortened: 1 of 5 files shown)


LucaCappelletti94:
@iffyio I have refactored this PR also in line of the permissive approach you have suggested for #2356

tests/sqlparser_common.rs
```diff
@@ -11030,13 +11030,201 @@ fn parse_is_boolean() {
...
+}
+
+#[test]
+fn parse_is_json_predicate_negative() {
+    let dialects = all_dialects();
+
+    let cases = [
+        (
+            "SELECT * FROM t WHERE a IS JSON WITH FROM",
+            &["Expected: UNIQUE", "found: FROM"][..],
```
iffyio: can we drop the exact match on the error message (they tend to change as new features are added or code is refactored) and only assert that we get a parse failure?

tests/sqlparser_common.rs
```diff
@@ -11030,13 +11030,201 @@ fn parse_is_boolean() {
...
+        verified_expr("NOT a IS JSON"),
+        Expr::UnaryOp {
+            op: UnaryOperator::Not,
+            expr
+        } if matches!(&*expr, Expr::IsJson { .. })
+    );
+}
+
+#[test]
+fn parse_is_json_predicate_negative() {
```
```suggestion
fn parse_is_json_predicate_invalid() {
```

tests/sqlparser_common.rs
```diff
@@ -11030,13 +11030,201 @@ fn parse_is_boolean() {
...
         ),
         res.unwrap_err()
     );
 }
 
+#[test]
+fn parse_is_json_predicate() {
+    use self::Expr::*;
+
+    let sql = "a IS JSON";
```
iffyio: can we simplify the tests a bit since the struct assertions have a lot of overlap we can have most of them only rely on the roundtrip `verified_expr` - I imagine it would suffice to only have one of the scenarios assert the returned struct

tests/sqlparser_common.rs
LucaCappelletti94: Done

tests/sqlparser_common.rs
LucaCappelletti94: Done

tests/sqlparser_common.rs
LucaCappelletti94: Trimmed down

tests/sqlparser_common.rs
LucaCappelletti94: I have also simplified the other analogous error matching that pre-existed and overlapped with the current edit.

iffyio (approved):

PR #2257, opened 2026-02-27
Adding support for PostgreSQL's `STATISTICS` `CREATE`, `ALTER`, `DROP` and `COMMENT ON`.

LucaCappelletti94:
This PR adds PostgreSQL-style support for `CREATE STATISTICS`, `ALTER STATISTICS`, `DROP STATISTICS`, and `COMMENT ON STATISTICS` (see docs: [CREATE STATISTICS](https://www.postgresql.org/docs/current/sql-createstatistics.html), [ALTER STATISTICS](https://www.postgresql.org/docs/current/sql-alterstatistics.html), [DROP STATISTICS](https://www.postgresql.org/docs/current/sql-dropstatistics.html), and [COMMENT](https://www.postgresql.org/docs/current/sql-comment.html)), with dedicated AST nodes (`CreateStatistics`, `AlterStatistics`, and related enums) plus parser integration in both `PostgreSqlDialect` and `GenericDialect`.

Since I am re-using `parse_table_and_joins`, which accepts just about anything, I have added a validation function which verifies post hoc that it is actually correct syntax. I am not too happy with that solution, but it was either that or rewriting a variant of `parse_table_and_joins`. Happy to hear suggestions.

**This is yet another PR stemming from differential fuzzing with postgres 18 - I have about other 2500 crash cases, does anybody want some?**

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2509,6 +2509,8 @@ pub enum CommentObject {
     Schema,
     /// A sequence.
     Sequence,
+    /// Statistics object.
+    Statistics,
     /// A table.
     Table,
     /// A type.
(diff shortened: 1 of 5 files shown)


LucaCappelletti94:

PR #2261, opened 2026-03-04
Added SORTKEY keyword parsing for redshift queries

romanoff:
Added SORTKEY keyword parsing for redshift queries

Example of query:
```
CREATE TABLE users (
    user_id BIGINT,
    email VARCHAR(255),
    created_at TIMESTAMP
)
SORTKEY (created_at);
```

Spec:
https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html
https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_AS.html
https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -9988,6 +9988,18 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse Redshift `ALTER SORTKEY (column_list)`.
+    ///
+    /// See <https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html>
+    fn parse_alter_sort_key(&mut self) -> Result<AlterTableOperation, ParserError> {
+        self.expect_keyword_is(Keyword::ALTER)?;
+        self.expect_keyword_is(Keyword::SORTKEY)?;
+        self.expect_token(&Token::LParen)?;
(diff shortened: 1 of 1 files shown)


src/parser/mod.rs
```diff
@@ -10257,6 +10266,14 @@ impl<'a> Parser<'a> {
                 column_position,
             }
         } else if self.parse_keyword(Keyword::ALTER) {
+            // Redshift: ALTER SORTKEY (column_list)
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -10257,6 +10266,14 @@ impl<'a> Parser<'a> {
...
+            if self.parse_keyword(Keyword::SORTKEY) {
+                self.expect_token(&Token::LParen)?;
+                let columns = self.parse_comma_separated(|p| p.parse_expr())?;
+                self.expect_token(&Token::RParen)?;
+                return Ok(AlterTableOperation::AlterSortKey { columns });
+            }
```
iffyio: can we pull this out to its own function (the block is already quite  big)? we can do something like the following here to make the function self contained
```
if self.peek(TOKEN::SORT_KEY) {
    self.prev_token(); // Put back the `ALTER` keyword
    return self.parse_alter_sort_key() 
}

PR #2263, opened 2026-03-04
Fixed transaction handling for snowflake

romanoff:
Sample query:
```sql
BEGIN TRANSACTION;
  DROP TABLE IF EXISTS bla;
COMMIT;
```

Spec: https://docs.snowflake.com/en/sql-reference/sql/begin

### DIFF
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -250,11 +250,10 @@ impl Dialect for SnowflakeDialect {
             // Snowflake supports both `BEGIN TRANSACTION` and `BEGIN ... END` blocks.
             // If the next keyword indicates a transaction statement, let the
             // standard parse_begin() handle it.
-            if parser.peek_keyword(Keyword::TRANSACTION)
-                || parser.peek_keyword(Keyword::WORK)
-                || parser.peek_keyword(Keyword::NAME)
-                || parser.peek_token().token == Token::SemiColon
-                || parser.peek_token().token == Token::EOF
+            if parser
+                .peek_one_of_keywords(&[Keyword::TRANSACTION, Keyword::WORK, Keyword::NAME])
+                .is_some()
+                || matches!(parser.peek_token_ref().token, Token::SemiColon | Token::EOF)
             {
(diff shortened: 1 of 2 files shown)


src/dialect/snowflake.rs
```diff
@@ -247,6 +247,18 @@ impl Dialect for SnowflakeDialect {
...
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::BEGIN) {
+            // Snowflake supports both `BEGIN TRANSACTION` and `BEGIN ... END` blocks.
+            // If the next keyword indicates a transaction statement, let the
+            // standard parse_begin() handle it.
+            if parser.peek_keyword(Keyword::TRANSACTION)
+                || parser.peek_keyword(Keyword::WORK)
+                || parser.peek_keyword(Keyword::NAME)
+                || parser.peek_token().token == Token::SemiColon
+                || parser.peek_token().token == Token::EOF
```
iffyio: can we add a test case with only a begin keyword (for the EOF case)?

src/dialect/snowflake.rs
```diff
@@ -247,6 +247,18 @@ impl Dialect for SnowflakeDialect {
 
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::BEGIN) {
+            // Snowflake supports both `BEGIN TRANSACTION` and `BEGIN ... END` blocks.
+            // If the next keyword indicates a transaction statement, let the
+            // standard parse_begin() handle it.
+            if parser.peek_keyword(Keyword::TRANSACTION)
```
iffyio: can we use `parser.peek_one_of_keywords()` and `parser.peek_token_ref()` for the two cases to simplify or make the op less expensive?

romanoff:
@iffyio Updated. Thank you

iffyio (approved):
LGTM! Thanks @romanoff!

PR #2265, opened 2026-03-04
Fixed stage name parsing for snowflake

romanoff:
Fixed special characters in stage name parsing for snowflake
Example queries:
```sql
SELECT * FROM @stage/day=18/23.parquet;
```

```sql
SELECT * FROM @stage/0:18:23/23.parquet;
```

Spec: https://docs.snowflake.com/en/user-guide/querying-stage

### DIFF
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -2643,24 +2643,16 @@ fn test_snowflake_copy_into_stage_name_ends_with_parens() {
 #[test]
 fn test_snowflake_stage_name_with_special_chars() {
     // Stage path with '=' (Hive-style partitioning)
-    let sql = "SELECT * FROM @stage/day=18/23.parquet";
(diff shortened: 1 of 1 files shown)


tests/sqlparser_snowflake.rs
```diff
@@ -2640,6 +2640,29 @@ fn test_snowflake_copy_into_stage_name_ends_with_parens() {
     }
 }
 
+#[test]
+fn test_snowflake_stage_name_with_special_chars() {
+    // Stage path with '=' (Hive-style partitioning)
+    let sql = "SELECT * FROM @stage/day=18/23.parquet";
+    let stmt = snowflake().parse_sql_statements(sql).unwrap();
```
iffyio: can we use the `one_statement_parse_to` (or `verified_stmt` if that works) for the tests?

romanoff:
@iffyio Updated. Thank you

PR #2272, opened 2026-03-07
add support for databricks JSON accessors

whirlun:
Support databricks SQL JSON accessors
sample queries: 
```
SELECT c1:price
    FROM VALUES('{ "price": 5 }') AS T(c1);
 SELECT c1:['price']::decimal(5,2)
    FROM VALUES('{ "price": 5 }') AS T(c1);
SELECT c1:item[1].price::double
    FROM VALUES('{ "item": [ { "model" : "basic", "price" : 6.12 },
                             { "model" : "medium", "price" : 9.24 } ] }') AS T(c1);
```
spec: https://docs.databricks.com/en/sql/language-manual/functions/colonsign.html

the tests are written by Claude Code because it can write tests magnitudes better than me.

### DIFF
--- a/tests/sqlparser_databricks.rs
+++ b/tests/sqlparser_databricks.rs
@@ -601,7 +601,6 @@ fn parse_databricks_struct_type() {
     }
 }
 
-// https://docs.databricks.com/en/sql/language-manual/functions/colonsign.html
 #[test]
 fn parse_databricks_json_accessor() {
     // Basic colon accessor — unquoted field names are case-insensitive


src/parser/mod.rs
```diff
@@ -4223,13 +4233,26 @@ impl<'a> Parser<'a> {
...
+                    } else {
+                        path.push(self.parse_json_path_object_key()?);
+                    }
                 }
                 Token::Period if !path.is_empty() => {
                     path.push(self.parse_json_path_object_key()?);
                 }
                 Token::LBracket => {
-                    let key = self.parse_expr()?;
+                    let key = self.parse_json_path_bracket_key()?;
```
```suggestion
                    let key = self.parse_wildcard_expr()?;
```
would this work as well?

src/parser/mod.rs
```diff
@@ -4223,13 +4233,26 @@ impl<'a> Parser<'a> {
...
+                        // A bracket element directly after the colon, e.g. `raw:['field']`.
+                        // Push an empty Dot so the display re-emits the leading `:` for syntax roundtrip.
+                        path.push(JsonPathElem::Dot {
+                            key: String::new(),
+                            quoted: false,
+                        });
+                        self.next_token();
+                        let key = self.parse_json_path_bracket_key()?;
+                        self.expect_token(&Token::RBracket)?;
+                        path.push(JsonPathElem::Bracket { key });
```
iffyio: I think we can instead introduce a `JsonPathElem::ColonBracket` variant?

and in this match statement we can give it its own block with something like
```rust
Token::Colon if path.is_empty() && self.peek_token_ref() == Token::RBracket { ... }
```

tests/sqlparser_databricks.rs
```diff
@@ -600,3 +600,150 @@ fn parse_databricks_struct_type() {
         _ => unreachable!(),
     }
 }
+
+// https://docs.databricks.com/en/sql/language-manual/functions/colonsign.html
```
iffyio:
```suggestion
```
we can add a link to the docs in the introduced enum variant

tests/sqlparser_databricks.rs
```diff
@@ -600,3 +600,150 @@ fn parse_databricks_struct_type() {
...
+                        key: String::new(),
+                        quoted: false,
+                    },
+                    JsonPathElem::Bracket {
+                        key: Expr::value(Value::SingleQuotedString("OWNER".to_owned())),
+                    },
+                ],
+            },
+        }
+    );
```
iffyio: I think we can skip these assertions on the AST and rely on the ones above that use the `verified*` helper functions

src/parser/mod.rs
whirlun: Didn't know there's already a function parsing wildcard. It works well.

src/parser/mod.rs
whirlun: Yep it's more clear. done.

tests/sqlparser_databricks.rs
whirlun: Moved to `JsonPathElem::ColonBracket` docstring.

tests/sqlparser_databricks.rs
whirlun: Removed AST tests.

iffyio (approved):
LGTM! Thanks @whirlun!

PR #2273, opened 2026-03-07
Add support for PostgreSQL LOCK TABLE

mjbshaw:
PostgreSQL 14 added [LOCK](https://www.postgresql.org/docs/current/sql-lock.html). This implements parsing support for that statement. Beyond the added tests, the following minimum reproducible example demonstrate's sqlparser's current failure to parse this statement:

**Cargo.toml:**
```
[package]
name = "demo"
version = "0.1.0"
edition = "2024"

[workspace]

[dependencies]
sqlparser_patched = { package = "sqlparser", path = "path/to/mjbshaw/fork/datafusion-sqlparser-rs" }
sqlparser_published = { package = "sqlparser", version = "0.61.0" }
```

**src/main.rs:**
```
use sqlparser_patched::dialect::PostgreSqlDialect as PatchedPostgreSqlDialect;
use sqlparser_patched::parser::Parser as PatchedParser;
use sqlparser_published::dialect::PostgreSqlDialect as PublishedPostgreSqlDialect;
use sqlparser_published::parser::Parser as PublishedParser;

const SQL: &str = "LOCK TABLE public.widgets IN EXCLUSIVE MODE";

fn main() {
    match PublishedParser::parse_sql(&PublishedPostgreSqlDialect {}, SQL) {
        Ok(statements) => panic!(
            "expected published sqlparser 0.61.0 to reject `{SQL}`, but it parsed: {statements:?}"
        ),
        Err(error) => {
            println!("published sqlparser 0.61.0 failed as expected: {error}");
        }
    }

    let statements = PatchedParser::parse_sql(&PatchedPostgreSqlDialect {}, SQL)
        .expect("patched sqlparser should parse PostgreSQL LOCK TABLE");
    assert_eq!(statements.len(), 1, "expected exactly one statement");
    println!("patched sqlparser parsed successfully: {}", statements[0]);
}
```

Then run with `cargo run`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -6301,6 +6301,8 @@ impl<'a> Parser<'a> {
         let table_properties = self.parse_options(Keyword::TBLPROPERTIES)?;
         let table_options = if !table_properties.is_empty() {
             CreateTableOptions::TableProperties(table_properties)
+        } else if let Some(options) = self.maybe_parse_options(Keyword::OPTIONS)? {
+            CreateTableOptions::Options(options)
         } else {
             CreateTableOptions::None
         };
@@ -6379,6 +6381,7 @@ impl<'a> Parser<'a> {
         let name_before_not_exists = !if_not_exists_first
             && self.parse_keywords(&[Keyword::IF, Keyword::NOT, Keyword::EXISTS]);
         let if_not_exists = if_not_exists_first || name_before_not_exists;
+        let copy_grants = self.parse_keywords(&[Keyword::COPY, Keyword::GRANTS]);
         // Many dialects support `OR ALTER` right after `CREATE`, but we don't (yet).
         // ANSI SQL and Postgres support RECURSIVE here, but we don't support it either.
         let columns = self.parse_view_columns()?;
@@ -6446,6 +6449,7 @@ impl<'a> Parser<'a> {
             with_no_schema_binding,
             if_not_exists,
             temporary,
+            copy_grants,
             to,
             params: create_view_params,
             name_before_not_exists,
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -4293,6 +4293,9 @@ pub struct CreateView {
     pub if_not_exists: bool,
     /// if true, has SQLite `TEMP` or `TEMPORARY` clause <https://www.sqlite.org/lang_createview.html>
     pub temporary: bool,
+    /// Snowflake: `COPY GRANTS` clause
+    /// <https://docs.snowflake.com/en/sql-reference/sql/create-view>
+    pub copy_grants: bool,
     /// if not None, has Clickhouse `TO` clause, specify the table into which to insert results
     /// <https://clickhouse.com/docs/en/sql-reference/statements/create/view#materialized-view>
     pub to: Option<ObjectName>,
@@ -4336,6 +4339,9 @@ impl fmt::Display for CreateView {
                 .map(|to| format!(" TO {to}"))
                 .unwrap_or_default()
         )?;
+        if self.copy_grants {
+            write!(f, " COPY GRANTS")?;
+        }
         if !self.columns.is_empty() {
             write!(f, " ({})", display_comma_separated(&self.columns))?;
         }
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -247,6 +247,17 @@ impl Dialect for SnowflakeDialect {
 
     fn parse_statement(&self, parser: &mut Parser) -> Option<Result<Statement, ParserError>> {
         if parser.parse_keyword(Keyword::BEGIN) {
+            // Snowflake supports both `BEGIN TRANSACTION` and `BEGIN ... END` blocks.
+            // If the next keyword indicates a transaction statement, let the
+            // standard parse_begin() handle it.
+            if parser
+                .peek_one_of_keywords(&[Keyword::TRANSACTION, Keyword::WORK, Keyword::NAME])
+                .is_some()
+                || matches!(parser.peek_token_ref().token, Token::SemiColon | Token::EOF)
+            {
+                parser.prev_token();
+                return None;
+            }
             return Some(parser.parse_begin_exception_end());
         }
 
--- a/tests/sqlparser_bigquery.rs
+++ b/tests/sqlparser_bigquery.rs
@@ -592,6 +592,16 @@ fn parse_create_table_with_options() {
 }
 
 #[test]
+fn parse_create_external_table_with_options() {
+    bigquery().verified_stmt(
+        "CREATE EXTERNAL TABLE dataset_id.table1 (hvr_tx_seq STRING) OPTIONS(format = 'CSV')",
+    );
+    bigquery().verified_stmt(
+        "CREATE EXTERNAL TABLE dataset_id.table1 (hvr_tx_seq STRING) OPTIONS(format = 'CSV', allow_quoted_newlines = true, encoding = 'UTF8')",
+    );
+}
+
+#[test]
 fn parse_nested_data_types() {
     let sql = "CREATE TABLE table (x STRUCT<a ARRAY<INT64>, b BYTES(42)>, y ARRAY<STRUCT<INT64>>)";
     match bigquery_and_generic().one_statement_parses_to(sql, sql) {
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -8317,6 +8317,7 @@ fn parse_create_view() {
             params,
             name_before_not_exists: _,
             secure: _,
+            copy_grants: _,
         }) => {
             assert_eq!(or_alter, false);
             assert_eq!("myschema.myview", name.to_string());
@@ -8435,6 +8436,7 @@ fn parse_create_view_temporary() {
             params,
             name_before_not_exists: _,
             secure: _,
+            copy_grants: _,
         }) => {
             assert_eq!(or_alter, false);
             assert_eq!("myschema.myview", name.to_string());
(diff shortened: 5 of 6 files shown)


src/ast/mod.rs
```diff
@@ -4624,6 +4624,19 @@ pub enum Statement {
         is_eq: bool,
     },
     /// ```sql
+    /// LOCK [ TABLE ] [ ONLY ] name [ * ] [, ...] [ IN lockmode MODE ] [ NOWAIT ]
+    /// ```
+    ///
+    /// Note: this is a PostgreSQL-specific statement. See <https://www.postgresql.org/docs/current/sql-lock.html>
```
```suggestion
    /// See <https://www.postgresql.org/docs/current/sql-lock.html>
```

src/ast/mod.rs
```diff
@@ -4624,6 +4624,19 @@ pub enum Statement {
...
+    Lock {
```
iffyio: instead of using an anonymous struct we can introduce a `LockTable` struct (also I think the statement should be `LockTable` instead of `Lock`?

src/ast/mod.rs
```diff
@@ -4624,6 +4624,19 @@ pub enum Statement {
...
+        /// List of tables to lock.
+        tables: Vec<LockTableTarget>,
+        /// Optional lock mode. PostgreSQL defaults to `ACCESS EXCLUSIVE` when omitted.
```
```suggestion
        /// Lock mode.
```
we can avoid describing postgres semantics

src/parser/mod.rs
```diff
@@ -697,6 +697,9 @@ impl<'a> Parser<'a> {
                 // `INSTALL` is duckdb specific https://duckdb.org/docs/extensions/overview
                 Keyword::INSTALL if self.dialect.supports_install() => self.parse_install(),
                 Keyword::LOAD => self.parse_load(),
+                Keyword::LOCK if dialect_of!(self is PostgreSqlDialect | GenericDialect) => {
```
```suggestion
                Keyword::LOCK => {
```
I think we can skip the dialect check and let the parser always accept the statement

src/parser/mod.rs
```diff
@@ -697,6 +697,9 @@ impl<'a> Parser<'a> {
...
+                    self.parse_lock_table_statement()
```
iffyio: See the [CASE](https://github.com/mjbshaw/datafusion-sqlparser-rs/blob/69f04201df97c01238d8c5997358c590888db89d/src/parser/mod.rs#L608) statement for an example we can follow.  We can rewind the token befor calling the parse_lock function so that the function is standalone. Then the function should return a struct instead of the `Statement`

src/parser/mod.rs
```diff
@@ -18385,6 +18388,80 @@ impl<'a> Parser<'a> {
...
+            has_asterisk,
+        })
+    }
+
+    fn parse_lock_table_mode(&mut self) -> Result<LockTableMode, ParserError> {
+        if self.parse_keyword(Keyword::ACCESS) {
+            return match self.expect_one_of_keywords(&[Keyword::SHARE, Keyword::EXCLUSIVE])? {
+                Keyword::SHARE => Ok(LockTableMode::AccessShare),
+                Keyword::EXCLUSIVE => Ok(LockTableMode::AccessExclusive),
+                unexpected_keyword => Err(ParserError::ParserError(format!(
```
iffyio: can this use one of the error helper methods? (similar for the other places)

tests/sqlparser_postgres.rs
```diff
@@ -8703,3 +8703,39 @@ fn parse_pg_analyze() {
...
+        "LOCK TABLE ONLY public.widgets, analytics.events * IN SHARE ROW EXCLUSIVE MODE NOWAIT",
+    );
+    pg_and_generic().one_statement_parses_to(
+        "LOCK TABLE public.widgets NOWAIT",
+        "LOCK TABLE public.widgets NOWAIT",
+    );
+}
+
+#[test]
+fn parse_lock_table_ast() {
```
iffyio: can we merge this test case into the test function above?

tests/sqlparser_postgres.rs
```diff
@@ -8703,3 +8703,39 @@ fn parse_pg_analyze() {
...
+        "LOCK TABLE public.widgets IN EXCLUSIVE MODE",
+    );
+    pg_and_generic().one_statement_parses_to(
+        "LOCK TABLE ONLY public.widgets, analytics.events * IN SHARE ROW EXCLUSIVE MODE NOWAIT",
+        "LOCK TABLE ONLY public.widgets, analytics.events * IN SHARE ROW EXCLUSIVE MODE NOWAIT",
+    );
+    pg_and_generic().one_statement_parses_to(
+        "LOCK TABLE public.widgets NOWAIT",
+        "LOCK TABLE public.widgets NOWAIT",
+    );
```
iffyio: where applicable can we use `verified_stmt` instead?

tests/sqlparser_postgres.rs
```diff
@@ -8703,3 +8703,39 @@ fn parse_pg_analyze() {
...
     }
 }
+
+#[test]
+fn parse_lock_table() {
+    pg_and_generic().one_statement_parses_to(
+        "LOCK public.widgets IN EXCLUSIVE MODE",
+        "LOCK TABLE public.widgets IN EXCLUSIVE MODE",
+    );
+    pg_and_generic().one_statement_parses_to(
```
iffyio: can we ensure that the tests have coverage for all lock table modes?

src/parser/mod.rs
```diff
@@ -18385,6 +18388,80 @@ impl<'a> Parser<'a> {
...
+            tables,
+            lock_mode,
+            nowait,
+        })
+    }
+
+    fn parse_lock_table_target(&mut self) -> Result<LockTableTarget, ParserError> {
+        let only = self.parse_keyword(Keyword::ONLY);
+        let name = self.parse_object_name(false)?;
+        let has_asterisk = self.consume_token(&Token::Mul);
```
iffyio, line 18425: we seem to be lacking a test case where `has_asterisk` is false?

src/parser/mod.rs
```diff
@@ -18385,6 +18388,80 @@ impl<'a> Parser<'a> {
...
+
+        Ok(LockTableTarget {
+            name,
+            only,
+            has_asterisk,
+        })
+    }
+
+    fn parse_lock_table_mode(&mut self) -> Result<LockTableMode, ParserError> {
+        if self.parse_keyword(Keyword::ACCESS) {
```
iffyio: I think a simpler/clearer way to write this function would be something like the following?

```rust
if self.parse_keywords(ACCESS, SHARE) { ... }
else if self.parse_keywords(ACESS, EXCLUSIVE) { ... }
...
```
that way the accounted for cases are visible, and we avoid spreading out the error cases

src/ast/mod.rs
```diff
@@ -6387,6 +6414,74 @@ impl fmt::Display for TruncateTableTarget {
...
+        }
+        write!(f, "{}", self.name)?;
+        if self.has_asterisk {
+            write!(f, " *")?;
+        }
+        Ok(())
+    }
+}
+
+/// PostgreSQL lock modes for `LOCK TABLE`.
```
iffyio, line 6460: Can we add a link to the docs here?

src/ast/mod.rs
```diff
@@ -6387,6 +6414,74 @@ impl fmt::Display for TruncateTableTarget {
     }
 }
 
+/// Target of a PostgreSQL `LOCK TABLE` command
```
```suggestion
/// Target of a `LOCK TABLE` command
```

src/ast/mod.rs
```diff
@@ -6387,6 +6414,74 @@ impl fmt::Display for TruncateTableTarget {
...
+///
+/// Note this is its own struct because `visit_relation` requires an `ObjectName` (not a `Vec<ObjectName>`)
```
```suggestion
```
we can also include a reference to the docs here, so that its easy to find

src/ast/mod.rs
mjbshaw: I pulled this out to a separate struct, but I kept it named as `Lock` because Postgres defines the statement as `LOCK [ TABLE ] ...`, so `TABLE` is optional and the statement itself is `LOCK`. That seems to match existing naming here like `Analyze` and `Truncate`, where an optional `TABLE` keyword is represented in the data instead of the type name. I can change it if you really want but `Lock` seems the most consistent to me.

src/parser/mod.rs
mjbshaw: Done: refactored to follow `CASE`'s example.

src/parser/mod.rs
mjbshaw: Yes, done: the parser now uses the normal error helper path.

tests/sqlparser_postgres.rs
mjbshaw: Yes, done.

tests/sqlparser_postgres.rs
mjbshaw: Yes, done.

tests/sqlparser_postgres.rs
mjbshaw: Done: all 8 postgres lock modes are now tested.

src/parser/mod.rs
mjbshaw: Thanks for catching. The merged AST test now tests for this.

src/parser/mod.rs
mjbshaw: Oh nice, thank you. Done.

src/ast/mod.rs
mjbshaw: Yes, done.

src/ast/mod.rs
mjbshaw: Yes, I added a link. I'll also delete the last 2 lines with your suggested change.

PR #2277, opened 2026-03-13
Added encode, distkey, sortkey in attributes support for redshift

romanoff:
Sample query:
```sql
CREATE TABLE player_activity (date DATE ENCODE raw DISTKEY NOT NULL, userid INTEGER ENCODE az64 NOT NULL) DISTSTYLE KEY SORTKEY(date, userid)
```

Spec: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html


iffyio:
@romanoff fyi when you have some time if you can take a look at the conflicts in this branch

PR #2279, opened 2026-03-13
Fixed create table as with columns without types for redshift

romanoff:
Example query:
```sql
CREATE TEMP TABLE volt_tt_64c26ae9dfd43(userid, days_played_in_last_31) as select 1,2;
```

Spec: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_AS.html

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -8374,7 +8374,15 @@ impl<'a> Parser<'a> {
         };
 
         // parse optional column list (schema)
-        let (columns, constraints) = self.parse_columns()?;
+        // Redshift CTAS allows column names without types:
+        //   CREATE TABLE t (col1, col2) AS SELECT 1, 2
+        // Detect this by peeking for `( ident ,` or `( ident )` patterns.
+        let (columns, constraints) =
+            if dialect_of!(self is RedshiftSqlDialect) && self.peek_column_names_only() {
+                self.parse_columns_without_types()?
+            } else {
+                self.parse_columns()?
+            };
         let comment_after_column_def =
             if dialect_of!(self is HiveDialect) && self.parse_keyword(Keyword::COMMENT) {
                 let next_token = self.next_token();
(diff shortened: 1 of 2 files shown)


tests/sqlparser_redshift.rs
```diff
@@ -500,3 +500,15 @@ fn test_alter_table_alter_sortkey() {
...
+
+#[test]
+fn test_create_table_as_with_column_names() {
+    redshift().verified_stmt(
+        "CREATE TEMPORARY TABLE volt_tt (userid, days_played_in_last_31) AS SELECT 1, 2",
+    );
+    // TEMP is an alias for TEMPORARY
+    redshift().one_statement_parses_to(
+        "CREATE TEMP TABLE volt_tt(userid, days_played_in_last_31) AS SELECT 1, 2",
+        "CREATE TEMPORARY TABLE volt_tt (userid, days_played_in_last_31) AS SELECT 1, 2",
```
iffyio, line 512: this test/PR isn't related to temporary tables? if so it would probably be better to remove the temp qualifiers and we can use `verified_stmt` as usual?

src/parser/mod.rs
```diff
@@ -8374,7 +8374,15 @@ impl<'a> Parser<'a> {
         };
 
         // parse optional column list (schema)
-        let (columns, constraints) = self.parse_columns()?;
+        // Redshift CTAS allows column names without types:
+        //   CREATE TABLE t (col1, col2) AS SELECT 1, 2
+        // Detect this by peeking for `( ident ,` or `( ident )` patterns.
+        let (columns, constraints) =
+            if dialect_of!(self is RedshiftSqlDialect) && self.peek_column_names_only() {
```
iffyio, line 8381: Can we convert this into a dialect method?

Actually looking at the code this feature seems to be solving a similar problem to `is_column_type_sqlite_unspecified` - likely we can instead have that `parse_column_def` function optionally accept a datatype if this method is true (and can rename the method accordingly)

PR #2297, opened 2026-04-06
Snowflake: Add support for text data type modifiers

yoavcloud:


src/parser/mod.rs
```diff
@@ -12170,7 +12170,20 @@ impl<'a> Parser<'a> {
                     self.expect_token(&Token::RParen)?;
                     Ok(DataType::FixedString(character_length))
                 }
-                Keyword::TEXT => Ok(DataType::Text),
+                Keyword::TEXT => {
+                    if dialect.supports_text_type_modifiers() {
```
iffyio: can we always have the parser accept the modifier if it shows up in the statement? given that its optional, I think that would match the behavior of similar datatypes

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2299, opened 2026-04-09
Add xml '...' TypedString support for PostgreSQL

LucaCappelletti94:
This is the first PR in a chain adding PostgreSQL XML function support (split from #2252).

- Add xml `<foo/>` typed string support for PostgreSQL and Generic dialects, parsing it as `Expr::TypedString` (same mechanism used by DATE '...', TIME '...', etc.)
- Introduce `supports_xml_expressions()` dialect method (default false, true for PostgreSQL and Generic) per reviewer feedback to avoid `dialect_of!`
- Add `is_simple_unquoted_object_name` helper for matching single unquoted identifiers in `DataType::Custom`

### DIFF
--- a/src/ast/dml.rs
+++ b/src/ast/dml.rs
@@ -387,6 +387,9 @@ pub struct Update {
     pub output: Option<OutputClause>,
     /// SQLite-specific conflict resolution clause
     pub or: Option<SqliteOnConflict>,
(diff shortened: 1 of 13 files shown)


yoavcloud (approved):

yoavcloud:
Looks good, @LucaCappelletti94!

LucaCappelletti94:
Will proceed to step 2.

PR #2300, opened 2026-04-10
Redshift/PostgreSQL: Parse optional implicit aliases for wildcard select items

yoavcloud:
Add support for implicit aliases of wildcard select items, for example: `SELECT c.* cols FROM tbl`, which previously the parser would fail on, but are accepted by the dialects.

Semantically, however, it looks like the databases are ignoring this construct. For example, when running the following statement, PostgreSQL fails with the error `column "c1" does not exist`:
```sql
SELECT c.* AS c1
FROM customers AS c
ORDER BY c1;
```

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -1556,10 +1556,9 @@ pub trait Dialect: Debug + Any {
     ///
     /// Example:
     /// ```sql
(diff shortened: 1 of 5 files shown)


iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2301, opened 2026-04-10
Postgres regression 7b

LucaCappelletti94:
- Add `Expr::XmlConcat(Vec<Expr>)` variant with `Display` and `Spanned` implementations for PostgreSQL's [`XMLCONCAT(xml[, ...])`](https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-PRODUCING-XML-XMLCONCAT) function
- Introduce `maybe_parse_xml_function` dispatcher in the parser, hooked at the top of `parse_function`, which routes XML function names to dedicated AST nodes on dialects where `supports_xml_expressions()` is true
- Guards on a single unquoted identifier with case-insensitive matching, so only bare `xmlconcat(...)` on PostgreSQL and Generic dialects takes the new path — all other dialects and qualified/quoted names fall through to the normal function-call parser

This is the second PR in a chain splitting #2252 (XML function support) into reviewable units. Builds on #2299 which introduced the `supports_xml_expressions()` dialect method.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1332,6 +1332,12 @@ pub enum Expr {
     Lambda(LambdaFunction),
     /// Checks membership of a value in a JSON array
     MemberOf(MemberOf),
+    /// PostgreSQL `XMLCONCAT(xml[, ...])` — concatenates a list of
+    /// individual XML values to create a single value containing an
+    /// XML content fragment.
+    ///
+    /// [PostgreSQL](https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-PRODUCING-XML-XMLCONCAT)
+    XmlConcat(Vec<Expr>),
 }
 
 impl Expr {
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -1332,6 +1332,12 @@ pub enum Expr {
     Lambda(LambdaFunction),
     /// Checks membership of a value in a JSON array
     MemberOf(MemberOf),
+    /// PostgreSQL `XMLCONCAT(xml[, ...])` — concatenates a list of
+    /// individual XML values to create a single value containing an
+    /// XML content fragment.
+    ///
+    /// [PostgreSQL](https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-PRODUCING-XML-XMLCONCAT)
+    XmlConcat(Vec<Expr>),
```
iffyio, line 1340: xmlconcat looks like a regular function syntax wise, is the parser unable to parse it today?

LucaCappelletti94: It can be parsed as a function, I added it to the enum nevertheless for easy of use, for the same reason that other functions like CEIL for instance appear there. Of course, open to discussion whether it is worth it.

iffyio: ah I see, I think we should keep the parsing as is that it gets represented as a regular function vs making it a special case

LucaCappelletti94: Ok so, just to be clear, I should close this particular PR and leave this as-is?

iffyio: Yeah we can close this and have the current behavior as is

PR #2302, opened 2026-04-14
Add ASYNC keyword support for CREATE INDEX

amaksimo:
## Summary

Amazon Aurora DSQL uses `CREATE INDEX ASYNC` to create indexes asynchronously. This adds parsing support for the `ASYNC` keyword in `CREATE INDEX` statements, following the same pattern as the existing `CONCURRENTLY` keyword.

**Syntax**: `CREATE [UNIQUE] INDEX [ASYNC] [IF NOT EXISTS] <name> ON <table>(<columns>)`

## Changes

- Add `ASYNC` to the keyword list (`keywords.rs`)
- Add `pub r#async: bool` field to `CreateIndex` struct (`ast/ddl.rs`)
- Parse `ASYNC` keyword after `CONCURRENTLY` in `parse_create_index` (`parser/mod.rs`)
- Round-trip support via `Display` impl
- Update `Spanned` impl (`ast/spans.rs`)
- Tests for `CREATE INDEX ASYNC` and `CREATE UNIQUE INDEX ASYNC` (`sqlparser_postgres.rs`)
- Update all existing `CreateIndex` test patterns with the new field

Thanks for revewing!

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2819,6 +2819,7 @@ pub struct CreateIndex {
     /// whether the index is created concurrently
     pub concurrently: bool,
     /// whether the index is created asynchronously
+    /// [DSQL]: <https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-create-index-async.html>
     pub r#async: bool,
     /// IF NOT EXISTS clause
     pub if_not_exists: bool,
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -2818,6 +2818,8 @@ pub struct CreateIndex {
     pub unique: bool,
     /// whether the index is created concurrently
     pub concurrently: bool,
+    /// whether the index is created asynchronously
+    pub r#async: bool,
     /// IF NOT EXISTS clause
     pub if_not_exists: bool,
     /// INCLUDE clause: <https://www.postgresql.org/docs/current/sql-createindex.html>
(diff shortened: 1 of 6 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -3217,6 +3227,58 @@ fn parse_create_index_concurrently() {
...
+            assert!(r#async);
+            assert!(!if_not_exists);
+            assert_eq_vec(&["col1"], &columns);
+            assert!(include.is_empty());
+            assert!(with.is_empty());
+            assert!(index_options.is_empty());
+            assert!(alter_options.is_empty());
+        }
+        _ => unreachable!(),
+    }
```
```suggestion
    pg().verified_stmt("CREATE INDEX ASYNC my_index ON my_table(col1)"); 
```
same for the unique one. Since the flag is already asserted in other tests, a simple round trip test scenario should suffice.

Also can we move this test to common.rs - since its not guarded specifically to postgres

src/ast/ddl.rs
```diff
@@ -2818,6 +2818,8 @@ pub struct CreateIndex {
     pub unique: bool,
     /// whether the index is created concurrently
     pub concurrently: bool,
+    /// whether the index is created asynchronously
+    pub r#async: bool,
```
iffyio, line 2823: Can we add a link to the postgres docs describing the syntax?

amaksimo: Sure, did something similar to what is done for MySQL, given ASYNC is actually an Aurora DSQL extension, not standard PostgreSQL, so there isn't a PostgreSQL docs page for it.

tests/sqlparser_postgres.rs
amaksimo: Done, simplified both tests to round-trip verified_stmt() calls and moved them to sqlparser_common.rs.

Thank you

iffyio:
@amaksimo could you take a look at the ci failure when you get the chance?

amaksimo:
> @amaksimo could you take a look at the ci failure when you get the chance?

@iffyio fixed, thanks!

iffyio (approved):
LGTM! Thanks @amaksimo!

PR #2303, opened 2026-04-15
Redshift: PartiQL AT <index>

yoavcloud:
Add support for the `AT` keyword in PartiQL that is used during unnesting to retrieve the index of an element in an array.

For example:
```sql
SELECT index, val FROM my_table t, t.my_array val AT index;
```

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2533,6 +2533,8 @@ pub struct TableAlias {
     pub name: Ident,
     /// Optional column aliases declared in parentheses after the table alias.
     pub columns: Vec<TableAliasColumnDef>,
+    /// Optional PartiQL index alias declared with `AT`.
+    pub at: Option<Ident>,
 }
 
 impl fmt::Display for TableAlias {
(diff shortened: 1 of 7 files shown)


src/ast/query.rs
```diff
@@ -2527,6 +2527,8 @@ pub struct TableAlias {
     pub name: Ident,
     /// Optional column aliases declared in parentheses after the table alias.
     pub columns: Vec<TableAliasColumnDef>,
+    /// Optional PartiQL index alias declared with `AT`.
+    pub at: Option<Ident>,
```
iffyio, line 2531: Can we add a link to the doc defining the syntax? (maybe an example syntax would also be helpful like we have on the `explicit` field)

tests/sqlparser_redshift.rs
```diff
@@ -517,3 +517,27 @@ fn test_null_treatment_inside_and_outside_window_function() {
     redshift().verified_stmt("SELECT FIRST_VALUE(1 IGNORE NULLS) OVER () FROM (SELECT 1) t");
     redshift().verified_stmt("SELECT FIRST_VALUE(1) IGNORE NULLS OVER () FROM (SELECT 1) t");
 }
+
+#[test]
+fn test_partiql_from_alias_with_at_index() {
+    redshift().verified_stmt("SELECT * FROM lineitem AS l (a, b, c) AT idx");
```
iffyio, line 523: for the test can we switch to `all_dialect_where(|d| d.supports_partiql())` so that future dialects are covered without duplicating tests?

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2307, opened 2026-04-16
feat: add PostgreSQL EXCLUDE constraint parsing

fmguerreiro:
Adds PostgreSQL `EXCLUDE` table constraint parsing.

Reference: https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE

## Syntax
```sql
[ CONSTRAINT <name> ] EXCLUDE [ USING <index_method> ]
  ( { <col> | ( <expr> ) } [ COLLATE <c> ] [ <opclass> ] [ ASC | DESC ]
    [ NULLS FIRST | LAST ] WITH <operator> [, ...] )
  [ INCLUDE (<cols>) ]
  [ WHERE (<predicate>) ]
  [ DEFERRABLE | NOT DEFERRABLE ]
  [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
```

Operators may be a single token (`=`, `&&`, `<->`) or the Postgres
schema-qualified `OPERATOR(schema.op)` form.

## Changes
- `src/ast/table_constraints.rs`: `ExclusionConstraint`, `ExclusionElement`, `ExclusionOperator` (enum over token vs `OPERATOR(…)` form); `TableConstraint::Exclusion` with `Display`, `From`, `Spanned`
- `src/ast/mod.rs`, `src/ast/spans.rs`: re-exports + spans arm
- `src/parser/mod.rs`:
  - Gate `EXCLUDE` on `PostgreSqlDialect | GenericDialect`
  - `parse_exclusion_element` reuses `parse_order_by_expr_inner(true)` so the shared `index_elem` grammar stays in one place
  - `parse_pg_operator_ident_parts` extracted from the binary-op path; reused by `BinaryOperator::PGCustomBinaryOperator` and `ExclusionOperator::PgCustom`
- `tests/sqlparser_postgres.rs`: 20 tests covering basic/multi-element, INCLUDE, WHERE, default index method, DEFERRABLE/NOT DEFERRABLE × INITIALLY DEFERRED/IMMEDIATE, opclass, ASC/DESC, NULLS FIRST/LAST, function-expression elements, COLLATE, `OPERATOR(…)` form, cross-dialect rejection (and `exclude` remaining a legal column name in MySQL/SQLite), missing-WITH / missing-operator / empty-element-list errors


src/ast/table_constraints.rs
```diff
@@ -117,6 +117,12 @@ pub enum TableConstraint {
     ///
     /// [1]: https://www.postgresql.org/docs/current/sql-altertable.html
     UniqueUsingIndex(ConstraintUsingIndex),
+    /// PostgreSQL `EXCLUDE` constraint.
+    ///
+    /// `[ CONSTRAINT <name> ] EXCLUDE [ USING <index_method> ] ( <element> WITH <operator> [, ...] ) [ INCLUDE (<cols>) ] [ WHERE (<predicate>) ]`
+    ///
+    /// See <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE>
+    Exclusion(ExclusionConstraint),
```
iffyio:
```suggestion
    Exclude(ExcludeConstraint),
```
does this make more sense if the constraint is called 'EXCLUDE' rather than 'EXCLUSION'?

src/ast/table_constraints.rs
```diff
@@ -117,6 +117,12 @@ pub enum TableConstraint {
     ///
     /// [1]: https://www.postgresql.org/docs/current/sql-altertable.html
     UniqueUsingIndex(ConstraintUsingIndex),
+    /// PostgreSQL `EXCLUDE` constraint.
+    ///
+    /// `[ CONSTRAINT <name> ] EXCLUDE [ USING <index_method> ] ( <element> WITH <operator> [, ...] ) [ INCLUDE (<cols>) ] [ WHERE (<predicate>) ]`
+    ///
+    /// See <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE>
```
iffyio:
```suggestion
    /// `EXCLUDE` constraint.
    ///
    /// `[ CONSTRAINT <name> ] EXCLUDE [ USING <index_method> ] ( <element> WITH <operator> [, ...] ) [ INCLUDE (<cols>) ] [ WHERE (<predicate>) ]`
    ///
    /// [PostgreSQL](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE)
```

src/ast/table_constraints.rs
```diff
@@ -603,3 +616,123 @@ impl crate::ast::Spanned for ConstraintUsingIndex {
...
+            }
+        }
+    }
+}
+
+/// One element in an `EXCLUDE` constraint's element list.
+///
+/// `{ column_name | ( expression ) } [ opclass ] [ ASC | DESC ] [ NULLS { FIRST | LAST } ] WITH <operator>`
+///
+/// See <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE>
```
iffyio:
```suggestion
/// One element in an `EXCLUDE` constraint's element list.
///
/// [PostgreSQL](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE)
```

src/ast/table_constraints.rs
```diff
@@ -603,3 +616,123 @@ impl crate::ast::Spanned for ConstraintUsingIndex {
         start.union(&end)
     }
 }
+
+/// The operator that follows `WITH` in an `EXCLUDE` element.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum ExclusionOperator {
```
iffyio:
```suggestion
pub enum ExcludeConstraintOperator {
```

src/ast/table_constraints.rs
```diff
@@ -603,3 +616,123 @@ impl crate::ast::Spanned for ConstraintUsingIndex {
         start.union(&end)
     }
 }
+
+/// The operator that follows `WITH` in an `EXCLUDE` element.
```
iffyio: Can we add a link to the docs? Otherwise I think this struct wouldn't be very descriptive on its own

src/ast/table_constraints.rs
```diff
@@ -603,3 +616,123 @@ impl crate::ast::Spanned for ConstraintUsingIndex {
...
+
+/// One element in an `EXCLUDE` constraint's element list.
+///
+/// `{ column_name | ( expression ) } [ opclass ] [ ASC | DESC ] [ NULLS { FIRST | LAST } ] WITH <operator>`
+///
+/// See <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE>
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub struct ExclusionElement {
```
iffyio:
```suggestion
pub struct ExcludeConstraintElement {
```

src/ast/table_constraints.rs
```diff
@@ -603,3 +616,123 @@ impl crate::ast::Spanned for ConstraintUsingIndex {
...
+        }
+        span
+    }
+}
+
+/// A PostgreSQL `EXCLUDE` constraint.
+///
+/// `[ CONSTRAINT <name> ] EXCLUDE [ USING <index_method> ] ( <element> WITH <operator> [, ...] ) [ INCLUDE (<cols>) ] [ WHERE (<predicate>) ]`
+///
+/// See <https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE>
```
iffyio:
```suggestion
/// An `EXCLUDE` constraint.
///
/// [PostgreSql](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-EXCLUDE)
```

src/parser/mod.rs
```diff
@@ -508,10 +508,10 @@ impl<'a> Parser<'a> {
                 Token::EOF => break,
 
                 // end of statement
-                Token::Word(word) => {
-                    if expecting_statement_delimiter && word.keyword == Keyword::END {
-                        break;
-                    }
+                Token::Word(word)
+                    if expecting_statement_delimiter && word.keyword == Keyword::END =>
```
iffyio, line 512: to get rid of these diffs, I think we might need to fetch the latest main (if they're from clippy)

src/parser/mod.rs
```diff
@@ -9915,9 +9895,55 @@ impl<'a> Parser<'a> {
                     .into(),
                 ))
             }
+            Token::Word(w)
+                if w.keyword == Keyword::EXCLUDE
+                    && dialect_of!(self is PostgreSqlDialect | GenericDialect) =>
```
iffyio: can we introduce a dialect method for this instead of the macro

src/parser/mod.rs
```diff
@@ -9915,9 +9895,55 @@ impl<'a> Parser<'a> {
...
+            {
+                let index_method = if self.parse_keyword(Keyword::USING) {
```
iffyio: can we move the impl to a `parse_exclude_constraint()` function or similar?

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_exclusion_element(&mut self) -> Result<ExclusionElement, ParserError> {
+        // `index_elem` grammar: { col | (expr) } [ opclass ] [ ASC | DESC ] [ NULLS FIRST | LAST ].
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_exclusion_element(&mut self) -> Result<ExclusionElement, ParserError> {
```
```suggestion
    fn parse_exclude_constraint_element(&mut self) -> Result<ExclusionElement, ParserError> {
```

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
...
+        // `index_elem` grammar: { col | (expr) } [ opclass ] [ ASC | DESC ] [ NULLS FIRST | LAST ].
+        // Shared with `CREATE INDEX` columns.
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
...
+            operator_class,
+            order,
+            operator,
+        })
+    }
+
+    /// Parse the operator that follows `WITH` in an `EXCLUDE` element.
+    ///
+    /// Accepts either a single operator token (e.g. `=`, `&&`, `<->`) or the
+    /// Postgres `OPERATOR(schema.op)` form for schema-qualified operators.
```
```suggestion
```

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
...
+    fn parse_exclusion_element(&mut self) -> Result<ExclusionElement, ParserError> {
+        // `index_elem` grammar: { col | (expr) } [ opclass ] [ ASC | DESC ] [ NULLS FIRST | LAST ].
+        // Shared with `CREATE INDEX` columns.
+        let (
+            OrderByExpr {
+                expr,
+                options: order,
+                ..
+            },
+            operator_class,
```
iffyio: this pattern looks a bit unusual, does it make more sense to instead call parse_order_by_expr_inner directly and store the IndexColumn instead of destructuring it in the ExclusionElement struct?

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
...
+            order,
+            operator,
+        })
+    }
+
+    /// Parse the operator that follows `WITH` in an `EXCLUDE` element.
+    ///
+    /// Accepts either a single operator token (e.g. `=`, `&&`, `<->`) or the
+    /// Postgres `OPERATOR(schema.op)` form for schema-qualified operators.
+    fn parse_exclusion_operator(&mut self) -> Result<ExclusionOperator, ParserError> {
```
```suggestion
    fn parse_exclude_constraint_operator(&mut self) -> Result<ExclusionOperator, ParserError> {
```

src/parser/mod.rs
```diff
@@ -9926,6 +9952,71 @@ impl<'a> Parser<'a> {
...
+        if self.parse_keyword(Keyword::OPERATOR) {
+            return Ok(ExclusionOperator::PgCustom(
+                self.parse_pg_operator_ident_parts()?,
+            ));
+        }
+
+        let operator_token = self.next_token();
+        match &operator_token.token {
+            Token::EOF | Token::RParen | Token::Comma | Token::SemiColon => {
```
iffyio: hmm could you explain this condition e.g. how does an optional RParen work?

tests/sqlparser_postgres.rs
```diff
@@ -9193,3 +9243,310 @@ fn parse_lock_table() {
...
+            }
+            other => panic!("Expected AddConstraint(Exclusion), got {other:?}"),
+        },
+        _ => panic!("Expected AlterTable"),
+    }
+}
+
+#[test]
+fn roundtrip_exclude_constraint() {
+    let sql = "CREATE TABLE t (CONSTRAINT no_overlap EXCLUDE USING gist (room WITH =, during WITH &&) INCLUDE (id) WHERE (active = true))";
```
iffyio: a couple comments re the tests, for this scenario we already use verified_stmt in the other tests so that unless this is testing some specific syntax we can remove it (or rename it accordingly)?
Then can we merge the tests into the same function and just rely on verified statement for them (except maybe one simple scenario if we want to verify the ast)? that would keep the tests simpler

tests/sqlparser_postgres.rs
```diff
@@ -9193,3 +9243,310 @@ fn parse_lock_table() {
...
+            parser.parse_sql_statements(sql).is_err(),
+            "dialect unexpectedly accepted EXCLUDE: {sql}"
+        );
+    }
+}
+
+#[test]
+fn exclude_as_column_name_parses_in_mysql_and_sqlite() {
+    // `exclude` must remain usable as an identifier where it is not a
+    // reserved keyword; PG reserves it as a constraint keyword.
```
iffyio: for both tests, after introducing the dialect method I think we can move them to common.rs

src/ast/table_constraints.rs
fmguerreiro: renamed in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: updated in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: updated in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: renamed in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: doc link added in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: renamed in 60b5fd2.

src/ast/table_constraints.rs
fmguerreiro: updated in 60b5fd2.

src/parser/mod.rs
fmguerreiro: merged main in fc8b794, unrelated diff dropped in 60b5fd2.

src/parser/mod.rs
fmguerreiro: added `Dialect::supports_exclude_constraint()` in 60b5fd2 (default `false`, true for `PostgreSqlDialect` and `GenericDialect`). macro guard replaced with the new method.

src/parser/mod.rs
fmguerreiro: extracted into `parse_exclude_constraint` in 60b5fd2.

src/parser/mod.rs
fmguerreiro: done in 60b5fd2.

src/parser/mod.rs
fmguerreiro: renamed to `parse_exclude_constraint_element` in 60b5fd2. return type is now `ExcludeConstraintElement` to match the AST rename.

src/parser/mod.rs
fmguerreiro: done in 60b5fd2.

src/parser/mod.rs
fmguerreiro: done in 60b5fd2.

src/parser/mod.rs
fmguerreiro: switched to `parse_create_index_expr` (the public wrapper around `parse_order_by_expr_inner`) in 60b5fd2. `ExcludeConstraintElement` now holds the resulting `IndexColumn` directly instead of destructuring an `OrderByExpr`.

src/parser/mod.rs
fmguerreiro: renamed to `parse_exclude_constraint_operator` in 60b5fd2 (returning `ExcludeConstraintOperator`).

src/parser/mod.rs
fmguerreiro: not optional. it was guarding against `WITH` followed directly by a structural delimiter (`,`, `)`, `;`, EOF), i.e. a missing operator. rewrote as an explicit `if matches!` with a comment in 60b5fd2.

tests/sqlparser_postgres.rs
fmguerreiro: collapsed the per-form tests into a single `parse_exclude_constraint` driven by `verified_stmt` in 60b5fd2, with one explicit AST-asserting case to lock the structure.

tests/sqlparser_postgres.rs
fmguerreiro: moved to `tests/sqlparser_common.rs` in 60b5fd2, gated by `all_dialects_where(|d| d.supports_exclude_constraint())`.

src/parser/mod.rs
fmguerreiro: `Token::RParen` is in the rejection set of the operator-token allowlist to surface a clear parse error when the WITH operator is missing (e.g. `col WITH )`). The actual closing `)` of the element list is consumed by the caller (`parse_exclude_constraint`), not by `parse_exclude_constraint_operator`.

src/parser/mod.rs
```diff
@@ -10020,6 +10016,111 @@ impl<'a> Parser<'a> {
...
+            include,
+            where_clause,
+            characteristics,
+        })
+    }
+
+    fn parse_exclude_constraint_element(
+        &mut self,
+    ) -> Result<ExcludeConstraintElement, ParserError> {
+        // `index_elem` grammar: { col | (expr) } [ opclass ] [ ASC | DESC ] [ NULLS FIRST | LAST ].
```
```suggestion
```

src/parser/mod.rs
iffyio: it doesn't seem that this was addressed?

tests/sqlparser_common.rs
```diff
@@ -2888,6 +2888,22 @@ fn parse_order_by_using_operator_invalid_cases() {
...
+    let dialects = TestedDialects::new(vec![
+        Box::new(PostgreSqlDialect {}),
+        Box::new(GenericDialect {}),
+    ]);
+    let result = dialects.parse_sql_statements("SELECT a OPERATOR() b");
+    assert_eq!(
+        ParserError::ParserError("Expected: operator name, found: )".to_string()),
+        result.unwrap_err()
+    );
+}
```
iffyio: can we move this to postgres tests instead (using pg_and_generic())?

tests/sqlparser_common.rs
```diff
@@ -4458,6 +4474,94 @@ fn parse_create_table_with_multiple_on_delete_fails() {
     .expect_err("should have failed");
 }
 
+#[test]
+fn parse_exclude_constraint() {
+    let dialects = all_dialects_where(|d| d.supports_exclude_constraint());
+
+    // One AST-asserting case to lock the structure of the parsed constraint;
+    // every other case below relies on `verified_stmt` round-tripping.
```
```suggestion
```

tests/sqlparser_common.rs
```diff
@@ -4458,6 +4474,94 @@ fn parse_create_table_with_multiple_on_delete_fails() {
...
+            }
+            other => panic!("expected single Exclude constraint, got {other:?}"),
+        },
+        other => panic!("expected CreateTable, got {other:?}"),
+    }
+
+    // Round-trip a representative range of forms (`USING`/no `USING`,
+    // `INCLUDE`, `WHERE`, `DEFERRABLE`, multi-element, ordering options,
+    // operator class, function expression, schema-qualified `OPERATOR(...)`,
+    // collation, `ALTER TABLE ... ADD CONSTRAINT ... EXCLUDE`).
```
```suggestion
```

tests/sqlparser_common.rs
```diff
@@ -2888,6 +2888,22 @@ fn parse_order_by_using_operator_invalid_cases() {
     assert!(matches!(err, ParserError::ParserError(_)));
 }
 
+#[test]
+fn parse_operator_empty_parens_rejected() {
+    // The shared `parse_pg_operator_ident_parts` helper rejects an empty
+    // `OPERATOR()` in the binary-expression path, not only in EXCLUDE
+    // constraints. Lock in that behavior.
```
```suggestion
```

src/parser/mod.rs
fmguerreiro: removed in f03de0a.

tests/sqlparser_common.rs
fmguerreiro: moved to tests/sqlparser_postgres.rs using pg_and_generic() in f03de0a.

tests/sqlparser_common.rs
fmguerreiro: removed in f03de0a.

tests/sqlparser_common.rs
fmguerreiro: removed in f03de0a.

tests/sqlparser_common.rs
fmguerreiro: removed (comment dropped with the test move) in f03de0a.

src/parser/mod.rs
fmguerreiro: the destructuring was dropped in 60b5fd2: parse_exclude_constraint_element now calls parse_create_index_expr() and stores the IndexColumn directly as `column`, no OrderByExpr destructuring. the redundant inline grammar comment is removed in f03de0a. the remaining `///` on parse_exclude_constraint_operator is rustdoc, consistent with the other private parse_* helpers in this file. let me know if you want that dropped too.

fmguerreiro: trimmed the rustdoc to the summary line in 87fada6.

fmguerreiro:
thanks for taking the time again to look at it, just let me know if there's anything else!

src/parser/mod.rs
```diff
@@ -10020,6 +10029,107 @@ impl<'a> Parser<'a> {
...
+    fn parse_exclude_constraint_operator(
+        &mut self,
+    ) -> Result<ExcludeConstraintOperator, ParserError> {
+        if self.parse_keyword(Keyword::OPERATOR) {
+            return Ok(ExcludeConstraintOperator::PGOperator(
+                self.parse_pg_operator_ident_parts()?,
+            ));
+        }
+
+        // Without this guard a bare `WITH` at the end of an element would
```
iffyio: is this check necessary (if the parser already consumed a valid operator it should be able to return OK and leave the rest of the stream as is)? the comment seems to suggest otherwise but I'm not sure I see how it affects this function's behavior

fmguerreiro: yeah, you're right, it doesn't change correctness. tested removing it: `EXCLUDE (c WITH)` still errors because the trailing `expect RParen` catches it, just with a worse message (`Expected: ',' or ')' ... found: EOF` instead of `Expected: exclusion operator, found: )`). no valid operator is one of those delimiter tokens, so the guard never rejects anything real. kept it just for the error message and reworded the comment to say that. happy to drop it if you'd rather.

iffyio: Ah yeah I think we can drop it in that case

fmguerreiro: done in a80797c

iffyio (approved):
Thanks @fmguerreiro!

PR #2309, opened 2026-04-17
Add Teradata dialect

iffyio:
Adds base implementation for the Teradata SQL dialect.

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -4891,3 +4891,9 @@ fn parse_create_database_with_charset_option_ordering() {
         "CREATE DATABASE mydb DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci",
     );
 }
+
+#[test]
+fn parse_adjacent_string_literal_concatenation() {
+    let sql = r#"SELECT 'M' "y" 'S' "q" 'l'"#;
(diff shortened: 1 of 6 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -4891,3 +4891,9 @@ fn parse_create_database_with_charset_option_ordering() {
         "CREATE DATABASE mydb DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci",
     );
 }
+
```
yoavcloud, line 4894: Curious as to why you add this test explicitly for MySQL, given the test already exists in common?

yoavcloud (approved):
Exciting!

tests/sqlparser_mysql.rs
iffyio: ah right that wasnt clear in the PR, the issue was that this statement in the common test `SELECT 'M' "y" 'S' "q" 'l'"` only works on mysql, because it involves the string concat feature, but also that the dialect supports `"y"` as a string character instead of an identifier which only mysql does. So the statement was rewritten to single quotes to remain in common, but also moved here so that the double quote behavior is still covered for mysql

yoavcloud: Ahh... I see that now. All good.

PR #2313, opened 2026-04-18
feat: parse CREATE STATISTICS, CREATE ACCESS METHOD, CREATE EVENT TRIGGER, CREATE TRANSFORM

fmguerreiro:
## Summary

- **CREATE STATISTICS** (pgmold-103): parses `CREATE STATISTICS [IF NOT EXISTS] name [(ndistinct, dependencies, mcv)] ON expr, ... FROM table`. Kinds are optional; expression list accepts any expression, though in practice bare column references are the common case.
- **CREATE ACCESS METHOD** (pgmold-104): parses `CREATE ACCESS METHOD name TYPE INDEX | TABLE HANDLER handler_function`.
- **CREATE EVENT TRIGGER** (pgmold-105): parses `CREATE EVENT TRIGGER name ON event [WHEN TAG IN ('tag', ...)] EXECUTE FUNCTION | PROCEDURE name()`. Supports all four event names (`ddl_command_start`, `ddl_command_end`, `table_rewrite`, `sql_drop`) and the `PROCEDURE` alias.
- **CREATE TRANSFORM** (pgmold-106): parses `CREATE [OR REPLACE] TRANSFORM FOR type LANGUAGE lang (FROM SQL WITH FUNCTION f(...), TO SQL WITH FUNCTION f(...))`. `OR REPLACE` is supported; both FROM and TO elements are optional in the grammar.

## New keyword

`TRANSFORM` added to `keywords.rs` alphabetically.

## Notes

- Fork CI does not auto-run on this branch.
- `verified_stmt` round-trip tests included for each variant (2–3 per statement).


fmguerreiro:
Opened on wrong repository. This belongs on the fork.

PR #2316, opened 2026-04-18
PostgreSQL: add support for CREATE AGGREGATE

fmguerreiro:
Parses PostgreSQL `CREATE [OR REPLACE] AGGREGATE`, both the modern `name (arg_types) (options)` form and the legacy `BASETYPE = ...` form. Options cover the [full set from the docs](https://www.postgresql.org/docs/current/sql-createaggregate.html): state and final functions, the moving-aggregate variants, `PARALLEL`, `SORTOP`, `HYPOTHETICAL`, `FINALFUNC_MODIFY`.

Ordered-set aggregates (`ORDER BY` in the argument list) are out of scope.

Round-trip tests cover both syntaxes, star and variadic args, the moving-aggregate options, and an unknown-option error.


src/ast/mod.rs
```diff
@@ -10253,16 +10260,23 @@ pub enum FunctionParallel {
...
         }
     }
 }
 
+impl fmt::Display for FunctionParallel {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "PARALLEL {}", self.as_str())
+    }
+}
+
```
LucaCappelletti94, line 10279: What is the reason for this edit?

src/ast/spans.rs
```diff
@@ -2518,6 +2519,12 @@ impl Spanned for AlterTable {
     }
 }
 
+impl Spanned for CreateAggregate {
+    fn span(&self) -> Span {
+        Span::empty()
+    }
+}
+
```
LucaCappelletti94, line 2580: I believe this object contains spanned objects, therefore you can return a span for it.

LucaCappelletti94 (changes requested):
Just a couple of preliminary clarification requests.

src/ast/mod.rs
fmguerreiro: the aggregate option prints `PARALLEL = SAFE` while the function form prints `PARALLEL SAFE`, so as_str gives the bare keyword without duplicating the strings.

src/ast/spans.rs
fmguerreiro: fixed in c0d2a577.

src/parser/mod.rs
```diff
@@ -5317,13 +5317,15 @@ impl<'a> Parser<'a> {
...
+        } else if self.parse_keyword(Keyword::AGGREGATE) {
+            self.parse_create_aggregate(or_replace).map(Into::into)
         } else if self.parse_keyword(Keyword::SCHEMA) {
             self.parse_create_schema(or_replace)
         } else if self.parse_keyword(Keyword::WAREHOUSE) {
             self.parse_create_warehouse(or_replace).map(Into::into)
         } else if or_replace {
             self.expected_ref(
-                "[EXTERNAL] TABLE or [MATERIALIZED] VIEW or FUNCTION or SCHEMA or WAREHOUSE after CREATE OR REPLACE",
+                "[EXTERNAL] TABLE or [MATERIALIZED] VIEW or FUNCTION or AGGREGATE or SCHEMA or WAREHOUSE after CREATE OR REPLACE",
```
LucaCappelletti94: Please add a test to cover this case

src/parser/mod.rs
```diff
@@ -7460,6 +7453,218 @@ impl<'a> Parser<'a> {
...
+    fn parse_aggregate_modify_kind(&mut self) -> Result<AggregateModifyKind, ParserError> {
+        match self.expect_one_of_keywords(&[
+            Keyword::READ_ONLY,
+            Keyword::SHAREABLE,
+            Keyword::READ_WRITE,
+        ])? {
+            Keyword::READ_ONLY => Ok(AggregateModifyKind::ReadOnly),
+            Keyword::SHAREABLE => Ok(AggregateModifyKind::Shareable),
+            Keyword::READ_WRITE => Ok(AggregateModifyKind::ReadWrite),
+            _ => self.expected_ref(
```
LucaCappelletti94: Also this appears not to be covered by tests

src/ast/spans.rs
```diff
@@ -2528,6 +2529,65 @@ impl Spanned for AlterTable {
...
+                .chain(self.default_expr.iter().map(|expr| expr.span())),
+        )
+    }
+}
+
+impl Spanned for CreateAggregateOption {
+    fn span(&self) -> Span {
+        match self {
+            CreateAggregateOption::StateTransitionFunction(name)
+            | CreateAggregateOption::FinalFunction(name)
```
LucaCappelletti94: I am unsure whether these are covered (and llvm does not instrument the lines) or only a subset of these are tested

LucaCappelletti94 (changes requested):

PR #2317, opened 2026-04-18
PostgreSQL: Support CREATE FOREIGN DATA WRAPPER

fmguerreiro:
Parses `CREATE FOREIGN DATA WRAPPER name [HANDLER fn | NO HANDLER] [VALIDATOR fn | NO VALIDATOR] [OPTIONS (...)]` ([docs](https://www.postgresql.org/docs/current/sql-createforeigndatawrapper.html)).

Split out of this PR at @LucaCappelletti94's request. `CREATE FOREIGN TABLE` is now #2515.

`HANDLER` and `VALIDATOR` are accepted in either order, because the PostgreSQL grammar takes them as an unordered list even though the doc synopsis shows one order. A repeated clause errors. The `OPTIONS` list parser moves out of `parse_pg_create_server` so both statements share it; whichever of the two PRs lands second needs a rebase there.

`ALTER FOREIGN DATA WRAPPER` and `CREATE USER MAPPING` are out of scope.

Ported from the `pgmold-sqlparser` fork.


LucaCappelletti94:
I suggest to split the PR into two, once for each statement

fmguerreiro:
split. `CREATE FOREIGN TABLE` is now #2515.

src/ast/spans.rs
```diff
@@ -3163,4 +3164,19 @@ WHERE id = 1
...
+        let mut test = SpanTest::new(dialect, sql);
+
+        // Ends at the option key, not the statement: a quoted option value is an
+        // Ident with an empty span, so it contributes nothing to the union.
+        let stmt = test.0.parse_statement().unwrap();
+        assert_eq!(
+            test.get_source(stmt.span()),
+            "myfdw HANDLER myhandler OPTIONS (debug"
+        );
+    }
```
LucaCappelletti94: You should drop this test. It asserts the truncated span `... OPTIONS (debug`, which comes from `parse_identifier` returning `Span::empty()` for single-quoted strings (`src/parser/mod.rs:13877`). The test turns that defect into a contract, so a later fix of the value span would break it.

```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -9652,6 +9652,105 @@ fn parse_lock_table() {
...
+        "CREATE FOREIGN DATA WRAPPER myfdw VALIDATOR myvalidator HANDLER myhandler",
+        canonical,
+    );
+    pg_and_generic().one_statement_parses_to(
+        "CREATE FOREIGN DATA WRAPPER myfdw NO VALIDATOR NO HANDLER",
+        "CREATE FOREIGN DATA WRAPPER myfdw NO HANDLER NO VALIDATOR",
+    );
+
+    // Pinning the message keeps this from passing on the old positional parser,
+    // which also errored here but on the trailing tokens instead.
```
LucaCappelletti94: Differential comment with previous code.

```suggestion
```

tests/sqlparser_postgres.rs
```diff
@@ -9652,6 +9652,105 @@ fn parse_lock_table() {
...
+            key: "debug".into(),
+            value: Ident {
+                value: "true".to_string(),
+                quote_style: Some('\''),
+                span: Span::empty(),
+            },
+        }])
+    );
+
+    // A single hardcoded Display label would render this as `NO HANDLER NO HANDLER`.
```
LucaCappelletti94: Do not include comments describing the PR change, comments are for describing the code.

```suggestion
```

LucaCappelletti94 (changes requested):

PR #2318, opened 2026-04-19
feat: parse ALTER TABLE ATTACH/DETACH PARTITION for PostgreSQL

fmguerreiro:
## Summary

- Add `AlterTableOperation::AttachPartitionOf` for PostgreSQL `ALTER TABLE parent ATTACH PARTITION child { FOR VALUES <spec> | DEFAULT }`
- Add `AlterTableOperation::DetachPartitionOf` for PostgreSQL `ALTER TABLE parent DETACH PARTITION child [ CONCURRENTLY | FINALIZE ]`
- These are distinct from the existing ClickHouse `AttachPartition`/`DetachPartition` variants which use `PART|PARTITION <expr>` syntax
- Add `FINALIZE` keyword
- Add span impls for the new variants
- Bump version to 0.60.12

## Test plan

- Parser unit tests for all four ATTACH variants (range, list, hash, default)
- Parser unit tests for all three DETACH variants (plain, concurrently, finalize)
- Covers the exact pagila.sql statement shape: `ALTER TABLE ONLY public.payment ATTACH PARTITION public.payment_p2022_01 FOR VALUES FROM (...) TO (...)`


fmguerreiro:
Opened by mistake, not intended for upstream.

PR #2319, opened 2026-04-19
fix(postgres): treat FULLTEXT and SPATIAL as unreserved identifiers

fmguerreiro:
## Problem

PostgreSQL schemas that use `fulltext` as a column name (common with `tsvector`/trigram search) failed to parse:

```sql
CREATE TABLE public.film (
    fulltext tsvector NOT NULL
);
```

Error: `SQL parse error: sql parser error: Expected: a data type name, found: 'fulltext'`

## Root Cause

`parse_optional_table_constraint` consumed the `FULLTEXT` token before realising it was not a valid constraint keyword for the current dialect, then backtracked with `prev_token()`. The consume-then-backtrack pattern left the parser in a subtly broken state for non-MySQL/Generic dialects, causing the next `parse_column_def_inner` call to see the wrong current token.

## Fix

Add a peek-based early-return guard in `parse_optional_table_constraint` before any token is consumed. For dialects other than MySQL/Generic, `FULLTEXT` and `SPATIAL` are valid identifiers and must not be touched by the constraint parser:

```rust
if name.is_none()
    && self
        .peek_one_of_keywords(&[Keyword::FULLTEXT, Keyword::SPATIAL])
        .is_some()
    && !dialect_of!(self is GenericDialect | MySqlDialect)
{
    return Ok(None);
}
```

## Test

Added `parse_fulltext_column_and_index_in_postgres` in `tests/sqlparser_postgres.rs` covering:
- `CREATE TABLE film (fulltext TSVECTOR NOT NULL)` — column named `fulltext` with `tsvector` type
- `CREATE INDEX film_fulltext_idx ON film USING gist (fulltext)` — GiST index on a `fulltext` column

## Version

Bumped to `0.60.13`.


fmguerreiro:
closing this. the behavior it targets already works on main: in postgres, FULLTEXT and SPATIAL parse fine as column names through the existing catch-all in parse_optional_table_constraint, which calls prev_token() and returns Ok(None) for the unnamed case. ran this PR's four test cases against current main, all pass, so the guard added here is a no-op.

PR #2321, opened 2026-04-25
PostgreSQL: extend COMMENT ON for trigger/aggregate/policy and dollar-quoted bodies

fmguerreiro:
## What was not working

`Statement::Comment` could not faithfully represent several `COMMENT ON` forms from the PostgreSQL grammar:

```sql
COMMENT ON TRIGGER trig ON public.foo IS 'x';      -- parse error: no variant, no ON <tbl> tail
COMMENT ON POLICY pol ON public.foo IS 'x';        -- parse error: no variant
COMMENT ON AGGREGATE sum_distinct(int) IS 'x';     -- parse error: no variant, no (arg_types)
COMMENT ON FUNCTION add(int, int) IS 'x';          -- parse error at `(`: object_name stops there
COMMENT ON TABLE foo IS $$hello$$;                 -- parse error: parse_literal_string rejects DollarQuoted
```

## Changes

- `CommentObject` gains `Trigger`, `Aggregate`, and `Policy` variants with `Display` strings `TRIGGER`, `AGGREGATE`, `POLICY`.
- `Statement::Comment` gains two optional fields:
  - `arguments: Option<Vec<DataType>>` for `FUNCTION` / `PROCEDURE` / `AGGREGATE` overload signatures. `Some(vec![])` represents `()`; `None` means no parameter list was provided.
  - `relation: Option<ObjectName>` for the `ON <table>` tail of `TRIGGER` / `POLICY`.
- Parser consumes the optional arg list after the object name for function-shaped targets (following the existing pattern from `parse_grant_procedure_or_function`) and the mandatory `ON <ObjectName>` tail for trigger/policy.
- `parse_literal_string` now accepts `Token::DollarQuotedString` in the PostgreSQL and Generic dialects and returns its inner value.
- `Display` for `Statement::Comment` renders `(arg, arg)` after the object name and `ON <relation>` before `IS` where applicable.

Dollar-quoted bodies round-trip through `Display` as single-quoted strings (semantic value preserved), so the new test uses `one_statement_parses_to(dollar_input, single_quoted_canonical)`.

## Breaking change disclosure

Neither `CommentObject` nor `Statement::Comment` is `#[non_exhaustive]`. Adding variants and struct fields is a hard breaking change for downstream consumers that pattern-match exhaustively without `..`. I recommend cutting this as `0.61.0` rather than a `0.60.x` patch.

## Tests

New unit tests in `tests/sqlparser_postgres.rs`:

- `parse_comment_on_trigger`
- `parse_comment_on_policy`
- `parse_comment_on_aggregate`
- `parse_comment_on_function_with_arg_types`
- `parse_comment_dollar_quoted_body`

Existing `parse_comments` test in `tests/sqlparser_common.rs` updated to use `..` in its struct patterns.


fmguerreiro:
Sorry, opened against the wrong repository — closing.

PR #2323, opened 2026-04-26
PostgreSQL: cleanup CommentObject, AlterType, AlterDefaultPrivileges

fmguerreiro:
## Summary
- `CommentObject`: add `keyword_str` and `from_keyword` helpers, collapse the 18-arm `parse_comment` match into a single lookup. `MATERIALIZED VIEW` stays inline as the only multi-keyword case. `Display` now reuses `keyword_str`.
- `Statement::Comment` `Display`: drop a stray `;` after the `IF EXISTS` write.
- `parse_alter_type` (`ALTER ATTRIBUTE`): annotate `let _ = parse_keywords(&[SET, DATA])` so it reads as canonicalization rather than a swallowed result.
- `AlterDefaultPrivilegesAction` `Display`: factor a `write_privileges_body` helper for the `<privs> ON <obj_type> {TO|FROM} <grantees>` body shared by `Grant` and `Revoke`.
- `AlterDefaultPrivileges`: drop the redundant `Spanned` impl that returned `Span::empty()`; the `Statement::AlterDefaultPrivileges` arm in `spans.rs` already covers the same fall-back.
- `AlterTypeOperation` Attribute `Display` arms: use the existing `display_option_spaced` for the trailing `drop_behavior`, replacing four near-identical `if let Some(...)` blocks.
- tests: drop one duplicate empty-arg `verified_stmt` (`COMMENT ON FUNCTION public.noop()`); the `no_args()` line in `parse_comment_on_aggregate` already exercises the same `Some(vec![])` path.

No behavior change. `-23` LOC net.

## Test plan
- [ ] `cargo test` (CI)
- [x] `cargo fmt --all --check` (clean)


fmguerreiro:
Apologies — opened against the wrong repository. This is a downstream cleanup for the pgmold-sqlparser fork (fmguerreiro/datafusion-sqlparser-rs); will reopen there. Sorry for the noise.

PR #2325, opened 2026-04-27
Databricks: Add support for `UPDATE SET *` and `INSERT *` in MERGE statements

finchxxia:
Add parsing support for the Databricks star shorthand syntax in MERGE statements, allowing `UPDATE SET *` and `INSERT *` to reference all columns from the source.

## Syntax

```
[ common_table_expression ]
  MERGE [ WITH SCHEMA EVOLUTION ] INTO target_table_name [target_alias]
     USING source_table_reference [source_alias]
     ON merge_condition
     { WHEN MATCHED [ AND matched_condition ] THEN matched_action |
       WHEN NOT MATCHED [BY TARGET] [ AND not_matched_condition ] THEN not_matched_action |
       WHEN NOT MATCHED BY SOURCE [ AND not_matched_by_source_condition ] THEN not_matched_by_source_action } [...]

matched_action
 { DELETE |
   UPDATE SET * |
   UPDATE SET { column = { expr | DEFAULT } } [, ...] }

not_matched_action
 { INSERT * |
   INSERT (column1 [, ...] ) VALUES ( expr | DEFAULT ] [, ...] )

not_matched_by_source_action
 { DELETE |
   UPDATE SET { column = { expr | DEFAULT } } [, ...] }
```

## Changes
- Add `MergeUpdateKind` enum (`Set(Vec<Assignment>)` | `Star`) to replace the previous `assignments` field on `MergeUpdateExpr`, enforcing valid states at the type level
- Add `MergeInsertKind::Star` variant for `INSERT *` shorthand
- Add `supports_merge_star_syntax()` dialect method for Databricks and Generic


src/ast/dml.rs
```diff
@@ -649,6 +649,14 @@ pub enum MergeInsertKind {
...
     /// [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement)
     Row,
+    /// The insert expression uses the `*` shorthand to insert all columns.
+    ///
+    /// Example:
+    /// ```sql
+    /// INSERT *
+    /// ```
+    /// [Databricks](https://docs.databricks.com/en/sql/language-manual/delta-merge-into.html)
+    Star,
```
iffyio:
```suggestion
    Wildcard,
```
I think this would be more consistent naming wise?

src/parser/merge.rs
```diff
@@ -120,7 +121,13 @@ impl Parser<'_> {
 
                     let update_token = self.get_current_token().clone();
                     self.expect_keyword_is(Keyword::SET)?;
-                    let assignments = self.parse_comma_separated(Parser::parse_assignment)?;
+                    let kind = if self.dialect.supports_merge_star_syntax()
```
iffyio: I looks like we should be able to drop the dialect method and let the parser accept the `*` syntax anytime it shows up?

tests/sqlparser_databricks.rs
```diff
@@ -738,3 +738,61 @@ fn parse_cte_without_as() {
...
+                    assert!(insert_expr.columns.is_empty());
+                }
+                _ => panic!("Expected INSERT action"),
+            }
+        }
+        _ => panic!("Expected MERGE statement"),
+    }
+}
+
+#[test]
```
iffyio: the tests can be merged into a single function? (maybe into the existing MERGE tests we have in common)

src/parser/merge.rs
finchxxia: > I looks like we should be able to drop the dialect method and let the parser accept the `*` syntax anytime it shows up?

done

tests/sqlparser_databricks.rs
finchxxia: > the tests can be merged into a single function? (maybe into the existing MERGE tests we have in common)

done. Thanks for your suggestions.

iffyio (approved):
LGTM! Thanks @finchxxia!

PR #2327, opened 2026-04-28
Snowflake: Accept COPY GRANTS after CREATE VIEW column list

97nitt:
## Summary

Snowflake's [`CREATE VIEW` syntax][docs] documents `COPY GRANTS` as appearing
*after* the column list:

```sql
CREATE OR REPLACE VIEW v (a, b) COPY GRANTS AS SELECT a, b FROM t;
```

The parser already accepted `COPY GRANTS` *before* the column list, but
rejected the documented post-columns position. This PR accepts either
position (but not both).

[docs]: https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax

## Changes

- `parse_create_view`: after consuming the column list, attempt a second
  `COPY GRANTS` match if one wasn't already seen.
- `Display` continues to emit the pre-columns form, so the post-columns input
  round-trips to the existing canonical shape.
- Tests: cover `VIEW`, `SECURE VIEW`, and `MATERIALIZED VIEW` with
  post-columns `COPY GRANTS`, and a baseline asserting the flag is not
  spuriously set when `COPY GRANTS` is absent.

## Test plan

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`

### DIFF
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -4766,11 +4766,6 @@ fn test_snowflake_create_view_copy_grants() {
 
 #[test]
 fn test_snowflake_create_view_copy_grants_after_columns() {
-    // Snowflake's documented placement for `COPY GRANTS` on `CREATE VIEW` is
-    // *after* the column list. Display normalizes to the pre-columns form
-    // already supported, so use `one_statement_parses_to` to assert the
-    // post-columns input is accepted and the AST flag is set.
-    // <https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax>
     let cases = [
         (
             "CREATE OR REPLACE VIEW v (a, b) COPY GRANTS AS SELECT a, b FROM t",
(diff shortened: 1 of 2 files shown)


tests/sqlparser_snowflake.rs
```diff
@@ -4764,6 +4764,52 @@ fn test_snowflake_create_view_copy_grants() {
     );
 }
 
+#[test]
+fn test_snowflake_create_view_copy_grants_after_columns() {
+    // Snowflake's documented placement for `COPY GRANTS` on `CREATE VIEW` is
+    // *after* the column list. Display normalizes to the pre-columns form
+    // already supported, so use `one_statement_parses_to` to assert the
+    // post-columns input is accepted and the AST flag is set.
+    // <https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax>
```
```suggestion
```

tests/sqlparser_snowflake.rs
```diff
@@ -4764,6 +4764,52 @@ fn test_snowflake_create_view_copy_grants() {
...
+            "CREATE OR REPLACE SECURE VIEW v COPY GRANTS (a, b) AS SELECT a, b FROM t",
+        ),
+        (
+            "CREATE MATERIALIZED VIEW v (a) COPY GRANTS AS SELECT a FROM t",
+            "CREATE MATERIALIZED VIEW v COPY GRANTS (a) AS SELECT a FROM t",
+        ),
+    ];
+    for (sql, parsed) in cases {
+        match snowflake().one_statement_parses_to(sql, parsed) {
+            Statement::CreateView(CreateView {
```
iffyio: we can drop the ast assertions in the tests

tests/sqlparser_snowflake.rs
97nitt: Done.

tests/sqlparser_snowflake.rs
97nitt: Done.

iffyio (approved):
LGTM! Thanks @97nitt!

PR #2329, opened 2026-04-29
Teradata: Add `CREATE TABLE` options

iffyio:
Adds support for some options in a Teradata `CREATE TABLE` statement.

### DIFF
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -472,13 +472,6 @@ fn test_snowflake_create_invalid_temporal_table() {
     );
 
     assert_eq!(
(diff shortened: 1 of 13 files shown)


tests/sqlparser_snowflake.rs
```diff
@@ -471,13 +471,6 @@ fn test_snowflake_create_invalid_temporal_table() {
         ))
     );
 
-    assert_eq!(
-        snowflake().parse_sql_statements("CREATE TEMP VOLATILE TABLE my_table (a INT)"),
```
iffyio, line 475: this scenario no longer applies now that the parser accepts `VOLATILE` option in a create table statement

yoavcloud (approved):
Looks good @iffyio!

PR #2330, opened 2026-04-30
Fix panic in EscapeQuotedString and parse_flush, clean up a few unwraps

Mrmaxmeier:
Hi,
I tried generating SQL inputs with `sqlparser` and ran into a few crashes. In the process of fixing these, I've also refactored some of the code in `ast/` to avoid unwraps where applicable.
Let me know if I should add tests for the two commits that actually change behaviour.
Thanks!

### DIFF
--- a/src/ast/value.rs
+++ b/src/ast/value.rs
@@ -577,7 +577,8 @@ impl fmt::Display for EscapeQuotedString<'_> {
                             // The quote is not escaped.
                             // Including idx in the range, so the quote at idx will be printed twice:
                             // in this call to write_str() and in the next one.
-                            f.write_str(&self.string[start_idx..=idx])?;
+                            let end_idx = idx + ch.len_utf8();
+                            f.write_str(&self.string[start_idx..end_idx])?;
                             start_idx = idx;
                         }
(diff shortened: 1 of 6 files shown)


src/ast/value.rs
```diff
@@ -577,7 +577,8 @@ impl fmt::Display for EscapeQuotedString<'_> {
                             // The quote is not escaped.
                             // Including idx in the range, so the quote at idx will be printed twice:
                             // in this call to write_str() and in the next one.
-                            f.write_str(&self.string[start_idx..=idx])?;
+                            let end_idx = idx + ch.len_utf8();
```
iffyio, line 580: could we add a test case that fails without this fix?

Mrmaxmeier: Sure, I've added a test. Notably though, this crash is not reachable with normal `sqlparser` usage because `escape_quoted_string` is only called with ASCII quote chars. I've only encountered this due to the way I'm using `sqlparser` for my input generation experiments.

iffyio (approved):
LGTM! Thanks @Mrmaxmeier!

PR #2332, opened 2026-05-06
PostgreSQL: add support for `ABORT TRANSACTION`

sunng87:
In postgres, `abort transaction` is completely identical to `rollback transaction`. It's used in foreign data wrapper to fetch data from cursor. This patch adds support for it.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -674,7 +674,8 @@ impl<'a> Parser<'a> {
                     self.prev_token();
                     self.parse_throw().map(Into::into)
                 }
-                Keyword::ROLLBACK | Keyword::ABORT => self.parse_rollback(),
+                Keyword::ROLLBACK => self.parse_rollback(),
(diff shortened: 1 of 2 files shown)


waynexia:
checked postgresql's document https://www.postgresql.org/docs/current/sql-abort.html:

>ABORT rolls back the current transaction and causes all the updates made by the transaction to be discarded. This command is identical in behavior to the standard SQL command [ROLLBACK](https://www.postgresql.org/docs/current/sql-rollback.html), and is present only for historical reasons.

but `ROLLBACK` is also used with `TO SAVEPOINT`, which is not the same case as `ABORT` https://www.postgresql.org/docs/current/sql-rollback-to.html. Maybe we should parse them separately?

sunng87:
Updated. Now abort will not parse for savepoint.

waynexia (approved):
Looking good

PR #2333, opened 2026-05-07
ClickHouse: Support scalar expressions in WITH clause

### LINKED ISSUE #2221: ClickHouse: Parsing error for WITH statement using scalar expressions
### Describe the bug
The ClickHouseDialect parser fails to parse valid ClickHouse syntax when WITH clause is used declaring [scalar expressions](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions).

### To Reproduce
```
use sqlparser::dialect::ClickHouseDialect;

fn main() {
    let sql = r#"
        WITH 42 AS answer
        SELECT answer FROM table
    "#;
    
    let dialect = ClickHouseDialect {};
    let result = Parser::parse_sql(&dialect, sql);
    println!("{:?}", result);
}
```
**Error:**
Err(ParserError("Expected: identifier, found: 42 at Line: 2, Column: 14"))

### Expected behavior tested in ClickHouse 26.1.2.11
The parser using ClickHouseDialect should accept the clause as ClickHouse does support it:
`WITH 42 AS answer SELECT answer FROM table`

### Additional context
The syntax from the [ClickHouse documentation](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions-syntax) for this clause is: `WITH <expression> AS <identifier>`
sqlparser version: 0.61

alrevuelta:
Fixes #2221.

ClickHouse's `WITH` clause supports a reversed-order form `<expression> AS <identifier>` (alongside, and freely interleaved with, traditional CTEs). Currently the parser hard-errors on the leading non-identifier token:

```
WITH 42 AS answer SELECT answer FROM t
-- Expected: identifier, found: 42
```

## Changes

- New `WithItem` enum (`Cte(Cte)` | `Named { expr, alias }`); `With::cte_tables: Vec<Cte>` becomes `With::items: Vec<WithItem>`.
- New dialect flag `Dialect::supports_with_clause_scalar_expression` (default `false`, `true` for `ClickHouseDialect` only).
- New `parse_with_item` routes each WITH-list element: when the leading token can't begin a CTE name, it parses `<expr> AS <ident>` directly; otherwise it tries the CTE form via `maybe_parse` and falls back. `parse_cte` itself is unchanged.

## Tests

Added in `tests/sqlparser_clickhouse.rs`, covering: literal scalar (the issue's example), string literal, aggregate, scalar subquery, bare-identifier disambiguation, mixing scalar + CTE in one list, lambda, and a negative test confirming non-ClickHouse dialects still reject the form.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -754,8 +754,9 @@ pub struct With {
     pub with_token: AttachedToken,
     /// Whether the `WITH` is recursive (`WITH RECURSIVE`).
     pub recursive: bool,
-    /// The list of CTEs declared by this `WITH` clause.
-    pub cte_tables: Vec<Cte>,
+    /// The items declared by this `WITH` clause: traditional CTEs and,
+    /// for dialects that support it, named expressions.
+    pub items: Vec<WithItem>,
 }
 
 impl fmt::Display for With {
@@ -764,8 +765,38 @@ impl fmt::Display for With {
         if self.recursive {
             f.write_str("RECURSIVE ")?;
         }
-        display_comma_separated(&self.cte_tables).fmt(f)?;
+        display_comma_separated(&self.items).fmt(f)?;
         Ok(())
+    }
+}
+
+/// A single item in a `WITH` clause.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum WithItem {
+    /// A traditional common table expression: `name [(cols)] AS [MATERIALIZED] (query)`.
+    Cte(Cte),
+    /// `<expr> AS <alias>` — binds an expression (literal, scalar subquery,
+    /// lambda, …) to a name visible in the surrounding query.
+    ///
+    /// See ClickHouse's [common scalar expressions][1].
+    ///
+    /// [1]: https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions
+    Named {
+        /// The expression bound to the alias.
+        expr: Expr,
+        /// The name the expression is bound to.
+        alias: Ident,
+    },
+}
+
+impl fmt::Display for WithItem {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            WithItem::Cte(cte) => cte.fmt(f),
+            WithItem::Named { expr, alias } => write!(f, "{expr} AS {alias}"),
+        }
     }
 }
 
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -754,9 +754,8 @@ pub struct With {
     pub with_token: AttachedToken,
     /// Whether the `WITH` is recursive (`WITH RECURSIVE`).
     pub recursive: bool,
-    /// The items declared by this `WITH` clause: traditional CTEs and,
-    /// for dialects that support it, named expressions.
-    pub items: Vec<WithItem>,
+    /// The expressions declared by this `WITH` clause.
+    pub exprs: Vec<WithExpression>,
 }
 
 impl fmt::Display for With {
@@ -765,37 +764,29 @@ impl fmt::Display for With {
         if self.recursive {
             f.write_str("RECURSIVE ")?;
         }
-        display_comma_separated(&self.items).fmt(f)?;
+        display_comma_separated(&self.exprs).fmt(f)?;
         Ok(())
     }
 }
 
-/// A single item in a `WITH` clause.
+/// A single expression in a `WITH` clause.
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-pub enum WithItem {
-    /// A traditional common table expression: `name [(cols)] AS [MATERIALIZED] (query)`.
+pub enum WithExpression {
+    /// A common table expression.
     Cte(Cte),
-    /// `<expr> AS <alias>` — binds an expression (literal, scalar subquery,
-    /// lambda, …) to a name visible in the surrounding query.
+    /// A common scalar expression.
     ///
-    /// See ClickHouse's [common scalar expressions][1].
-    ///
-    /// [1]: https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions
-    Named {
-        /// The expression bound to the alias.
-        expr: Expr,
-        /// The name the expression is bound to.
-        alias: Ident,
-    },
+    /// [Clickhouse]: https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions
+    Cse(ExprWithAlias),
 }
 
-impl fmt::Display for WithItem {
+impl fmt::Display for WithExpression {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            WithItem::Cte(cte) => cte.fmt(f),
-            WithItem::Named { expr, alias } => write!(f, "{expr} AS {alias}"),
+            WithExpression::Cte(cte) => cte.fmt(f),
+            WithExpression::Cse(cse) => cse.fmt(f),
         }
     }
 }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -1745,18 +1745,15 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Returns true if the dialect allows a `WITH` clause item to bind a
-    /// scalar (or otherwise non-CTE) expression to a name, with the form
-    /// `<expression> AS <identifier>` — alongside or instead of the
-    /// traditional `<identifier> AS (<subquery>)` CTE form.
+    /// Returns true if the dialect supports Common Scalar Expressions in a `WITH` clause.
     ///
-    /// For example, in ClickHouse:
+    /// For example:
     /// ```sql
     /// WITH 42 AS answer SELECT answer FROM t
     /// ```
     ///
     /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
-    fn supports_with_clause_scalar_expression(&self) -> bool {
+    fn supports_common_scalar_expressions(&self) -> bool {
         false
     }
 
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -1745,6 +1745,21 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect allows a `WITH` clause item to bind a
+    /// scalar (or otherwise non-CTE) expression to a name, with the form
+    /// `<expression> AS <identifier>` — alongside or instead of the
+    /// traditional `<identifier> AS (<subquery>)` CTE form.
+    ///
+    /// For example, in ClickHouse:
+    /// ```sql
+    /// WITH 42 AS answer SELECT answer FROM t
+    /// ```
+    ///
+    /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
+    fn supports_with_clause_scalar_expression(&self) -> bool {
+        false
+    }
+
     /// Returns true if the dialect supports parenthesized multi-column
     /// aliases in SELECT items. For example:
     /// ```sql
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -14105,7 +14105,7 @@ impl<'a> Parser<'a> {
             Some(With {
                 with_token: with_token.clone().into(),
                 recursive: self.parse_keyword(Keyword::RECURSIVE),
-                items: self.parse_comma_separated(Parser::parse_with_item)?,
+                exprs: self.parse_comma_separated(Parser::parse_with_expression)?,
             })
         } else {
             None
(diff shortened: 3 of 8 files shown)


tests/sqlparser_clickhouse.rs
```diff
@@ -1845,6 +1845,70 @@ fn parse_inner_array_join() {
...
+#[test]
+fn parse_with_clause_named_expression_unsupported_in_other_dialects() {
+    // The named-expression form is only enabled for ClickHouse; other
+    // dialects should still reject `WITH 42 AS answer …`.
+    let res = sqlparser::parser::Parser::parse_sql(
+        &GenericDialect {},
+        "WITH 42 AS answer SELECT answer FROM t",
+    );
+    assert!(res.is_err(), "expected parse error, got {res:?}");
+}
```
iffyio: I think we can skip this, no need to test what the parser does for an invalid query for unsupported dialects

tests/sqlparser_clickhouse.rs
```diff
@@ -1845,6 +1845,70 @@ fn parse_inner_array_join() {
...
+
+    // Lambda as the bound expression (also taken from the docs).
+    clickhouse().verified_stmt(
+        "WITH '.txt' AS extension, (id, extension) -> concat(lower(id), extension) AS gen_name \
+         SELECT gen_name('test', '.sql') AS file_name",
+    );
+}
+
+#[test]
+fn parse_with_clause_named_expression_ast() {
```
iffyio: we can merge this with the test above since they cover the same feature

tests/sqlparser_clickhouse.rs
```diff
@@ -1845,6 +1845,70 @@ fn parse_inner_array_join() {
     }
 }
 
+#[test]
+fn parse_with_clause_named_expression() {
+    // Plain literal scalar.
+    clickhouse().verified_stmt("WITH 42 AS answer SELECT answer FROM t");
```
iffyio: for the tests, lets use `all_dialects_where` instead of hardcoding it to clickhouse

src/parser/mod.rs
```diff
@@ -14639,6 +14639,33 @@ impl<'a> Parser<'a> {
...
+        }
+
+        // CTE form must start with an identifier. If the leading token
+        // can't begin one (e.g. `42`, `(SELECT …)`, `(x, y) -> …`), this
+        // is unambiguously the named-expression form.
+        if matches!(self.peek_token().token, Token::Word(_)) {
+            if let Some(cte) = self.maybe_parse(|p| p.parse_cte())? {
+                return Ok(WithItem::Cte(cte));
+            }
+        }
```
iffyio: Can this be simplified to use `peek_subquery_or_cte_start`? e.g.

that we do `if self.dialect.supports() and !self.peek_subquery_or_cte_start { parse expr } else { parse cte }`

src/parser/mod.rs
```diff
@@ -14639,6 +14639,33 @@ impl<'a> Parser<'a> {
         Ok(cte)
     }
 
+    /// Parse a single item in a `WITH` clause.
+    ///
+    /// In standard SQL this is always a CTE (`name [(cols)] AS (query)`).
+    /// Dialects that enable [`Dialect::supports_with_clause_scalar_expression`]
+    /// — currently only ClickHouse — also accept the reversed form
+    /// `<expression> AS <identifier>`, which can be freely interleaved with
+    /// CTEs in the same comma-separated list.
```
```suggestion
    /// Parse a single item in a `WITH` clause.
```

src/dialect/mod.rs
```diff
@@ -1745,6 +1745,21 @@ pub trait Dialect: Debug + Any {
...
+    /// `<expression> AS <identifier>` — alongside or instead of the
+    /// traditional `<identifier> AS (<subquery>)` CTE form.
+    ///
+    /// For example, in ClickHouse:
+    /// ```sql
+    /// WITH 42 AS answer SELECT answer FROM t
+    /// ```
+    ///
+    /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
+    fn supports_with_clause_scalar_expression(&self) -> bool {
```
```suggestion
    fn supports_common_scalar_expressions(&self) -> bool {
```

src/dialect/mod.rs
```diff
@@ -1745,6 +1745,21 @@ pub trait Dialect: Debug + Any {
...
+    /// scalar (or otherwise non-CTE) expression to a name, with the form
+    /// `<expression> AS <identifier>` — alongside or instead of the
+    /// traditional `<identifier> AS (<subquery>)` CTE form.
+    ///
+    /// For example, in ClickHouse:
+    /// ```sql
+    /// WITH 42 AS answer SELECT answer FROM t
+    /// ```
+    ///
+    /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
```
```suggestion
    /// Returns true if the dialect supports Common Scalar Expressions in a `WITH` clause.
    ///
    /// For example:
    /// ```sql
    /// WITH 42 AS answer SELECT answer FROM t
    /// ```
    ///
    /// [ClickHouse](https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions)
```

src/ast/query.rs
```diff
@@ -764,11 +765,41 @@ impl fmt::Display for With {
...
+        display_comma_separated(&self.items).fmt(f)?;
         Ok(())
     }
 }
 
+/// A single item in a `WITH` clause.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum WithItem {
```
iffyio: repr wise can we do something like?
```rust
pub enum WithExpression {
    Cte(...),
    Cse(ExprWithAlias), // can we reuse exprwithalias?
}

File: src/ast/query.rs [hunk 9]
```diff
@@ -764,11 +765,41 @@ impl fmt::Display for With {
...
+    /// A traditional common table expression: `name [(cols)] AS [MATERIALIZED] (query)`.
```
```suggestion
    /// A common table expression.
```

src/ast/query.rs
```diff
@@ -764,11 +765,41 @@ impl fmt::Display for With {
...
+    Cte(Cte),
+    /// `<expr> AS <alias>` — binds an expression (literal, scalar subquery,
+    /// lambda, …) to a name visible in the surrounding query.
+    ///
+    /// See ClickHouse's [common scalar expressions][1].
+    ///
+    /// [1]: https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions
```
```suggestion
    /// A common scalar expression.
    ///
    /// [Clickhouse]: https://clickhouse.com/docs/sql-reference/statements/select/with#common-scalar-expressions
```

src/ast/query.rs
```diff
@@ -754,8 +754,9 @@ pub struct With {
     pub with_token: AttachedToken,
     /// Whether the `WITH` is recursive (`WITH RECURSIVE`).
     pub recursive: bool,
-    /// The list of CTEs declared by this `WITH` clause.
-    pub cte_tables: Vec<Cte>,
+    /// The items declared by this `WITH` clause: traditional CTEs and,
+    /// for dialects that support it, named expressions.
+    pub items: Vec<WithItem>,
```
```suggestion
    pub exprs: Vec<WithExpression>,
```

src/ast/query.rs
```diff
@@ -754,8 +754,9 @@ pub struct With {
     pub with_token: AttachedToken,
     /// Whether the `WITH` is recursive (`WITH RECURSIVE`).
     pub recursive: bool,
-    /// The list of CTEs declared by this `WITH` clause.
-    pub cte_tables: Vec<Cte>,
+    /// The items declared by this `WITH` clause: traditional CTEs and,
+    /// for dialects that support it, named expressions.
```
```suggestion
    /// The expressions declared by this `WITH` clause.
```

alrevuelta: fixed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-ff51937f34d0076928ac31baffb41b26e4cbc0ec6dd8496105a7f61188f92ed1R757-R758

src/ast/query.rs
alrevuelta: fixed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-ff51937f34d0076928ac31baffb41b26e4cbc0ec6dd8496105a7f61188f92ed1R757-R758

src/ast/query.rs
alrevuelta: fixed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-ff51937f34d0076928ac31baffb41b26e4cbc0ec6dd8496105a7f61188f92ed1R776-R782

src/ast/query.rs
alrevuelta: fixed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-ff51937f34d0076928ac31baffb41b26e4cbc0ec6dd8496105a7f61188f92ed1R776-R782

src/dialect/mod.rs
alrevuelta: fixed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-4565f732f2c32d704cf0866d0f2460a833c6a6ab8292a1d541ccc649f9930107R158

src/parser/mod.rs
alrevuelta: fixed and minor renaming
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-fc6a8d66b8cb6bd48a119a70e489cbcef82e7f551e7cf08e8e972ef4e774ef49R14642

src/parser/mod.rs
alrevuelta: tbh, not sure i fully understand this. with that, will WITH 42 AS answer SELECT answer FROM t parse? peek_subquery_or_cte_start returns false on 42 (not (SELECT / (WITH), so we'd go to parse_cte, which then errors on 42 not being an identifier?

tests/sqlparser_clickhouse.rs
alrevuelta: done
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-b9a99857d69356e86daf74627f3d5da1f187d71e35dde75b0d924ed668ab1a43R1850

tests/sqlparser_clickhouse.rs
alrevuelta: merged
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-b9a99857d69356e86daf74627f3d5da1f187d71e35dde75b0d924ed668ab1a43L1887

tests/sqlparser_clickhouse.rs
alrevuelta: removed
https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a#diff-b9a99857d69356e86daf74627f3d5da1f187d71e35dde75b0d924ed668ab1a43L1902

alrevuelta:
@iffyio thanks for the comments, fixed.

there is only [this](https://github.com/apache/datafusion-sqlparser-rs/pull/2333#discussion_r3218429458) that i dont fully understand if possible.

src/parser/mod.rs
iffyio: Yeah that would work right or? if we error on 42 its because the syntax was invalid and we expected a cte but got 42 instead?

tests/sqlparser_clickhouse.rs
```diff
@@ -1845,6 +1845,56 @@ fn parse_inner_array_join() {
...
+        WithExpression::Cse(ExprWithAlias { expr, alias }) => {
+            assert_eq!(alias.as_ref().unwrap().value, "answer");
+            assert!(matches!(expr, Expr::Value(_)));
+        }
+        other => panic!("expected a common scalar expression, got {other:?}"),
+    }
+
+    // String literal scalar (from the ClickHouse docs).
+    dialects.verified_stmt(
+        "WITH '2019-08-01 15:23:00' AS ts_upper_bound SELECT * FROM hits \
```
iffyio: for the line breaks can we use concat like e.g. [here](https://github.com/alrevuelta/datafusion-sqlparser-rs/blob/3b868426bd4f1cc15017efbc3648f1b7cfab4a2a/tests/sqlparser_clickhouse.rs#L241) to format the sql strings?

src/parser/mod.rs
alrevuelta: right, i was mixing things up. can you confirm this `WITH (SELECT max(x) FROM t) AS m SELECT m` will work with your approach? i tried and seems that the first tokens `(SELECT` make peek_subquery_or_cte_start return true, which routes to parse_cte, but parse_cte immediately fails on ( since it expects an identifier.

tests/sqlparser_clickhouse.rs
alrevuelta: fixed 4f693572363f6c74e34c47fcd705676f5c6db650

src/parser/mod.rs
iffyio: I think the main goal with the comment was to simplify the cases a bit, so that ideally we have one branch handling the cte case, and another handling the cse case. Another way (there might be others) to accomplish that would be to try to parse a cte and if that fails fallback to cse e.g.

```rust
if let Some(cte) = self.maybe_parse(|parser| parser...)? else {
	return cte
} else if self.supprots() {
	return self.parse_cse
} else {
	self.expected("CTE or CSE")
}
```
maybe above woul be straightforward in that the cases that are covered is clearer?

sgargan:
I am very interested in this feature, is there a technical blocker what's currently implemented here?

alrevuelta:
@sgargan the implementation was close to ready iirc, just some minor comments to address. i dont have bandwidth right now but if you give me a hand with the remaining comments, we can open it again.

PR #2336, opened 2026-05-08
Snowflake: Add support for CREATE FILE FORMAT

### LINKED ISSUE #2070: Snowflake File Format not supported.
The Snowflake dialect doesn't support DDL commands for [File Format](https://docs.snowflake.com/en/sql-reference/sql/create-file-format).

Running
```
    let sql = "CREATE FILE FORMAT my_file_format";
    let dialect = SnowflakeDialect {};
```
Results in
```
thread 'main' panicked at src/main.rs:7:48:
called `Result::unwrap()` on an `Err` value: ParserError("Expected: an object type after CREATE, found: FILE at Line: 1, Column: 8")
```

I could try adding it myself but I wanted to check first since there's already a [`FileFormat`](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/ast/mod.rs#L7341-L7353) in the ast module. It looks to be meant for something else in the other dialects so should I expand that or implement a snowflake specific one?

ramnes:
This PR adds a `Statement::CreateFileFormat` variant and parses the full grammar:

`CREATE [ OR REPLACE ] [ { TEMP | TEMPORARY | VOLATILE } ] FILE FORMAT [ IF NOT EXISTS ] <name> [formatTypeOptions ] [ COMMENT = '<string>' ]`

Format options are stored as `KeyValueOptions`, with `COMMENT` split out into its own `Option<String>` field to match `CREATE STAGE`. `IDENTIFIER(?)` style names are handled via `parse_object_name(true)`.

Closes #2070

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4487,6 +4487,28 @@ pub enum Statement {
         comment: Option<String>,
     },
     /// ```sql
+    /// CREATE [ OR REPLACE ] [ { TEMP | TEMPORARY | VOLATILE } ] FILE FORMAT [ IF NOT EXISTS ] <name>
+    ///   [ TYPE = { CSV | JSON | AVRO | ORC | PARQUET | XML } [ formatTypeOptions ] ]
+    ///   [ COMMENT = '<string_literal>' ]
+    /// ```
+    /// See <https://docs.snowflake.com/en/sql-reference/sql/create-file-format>
+    CreateFileFormat {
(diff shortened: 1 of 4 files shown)


yoavcloud (approved):

ramnes:
@yoavcloud Any chance to get this merged? Would like to avoid rebases if main drifts, my time for sqlparser is pretty limited!

ramnes:
Thanks!

PR #2337, opened 2026-05-08
Make clickhouse tuples work

### LINKED ISSUE #2338: Clickhouse tuples not supported
Clickhouse tuples are not supported. The following errors `ParserError("Expected: end of statement, found: .1 at Line: 1, Column: 22")`.

```
SELECT t.1 FROM x
```

This is a valid Clickhouse way of accessing tuple elements. See https://clickhouse.com/docs/sql-reference/functions/tuple-functions#tupleelement. It is some syntactic sugar to `tupleElement(t, 1)` which does work.

Request: Make the above query work, and similar ones.

```
SELECT t.1 FROM x
SELECT (1, 2, 3).2
SELECT arr[1].1 FROM t
SELECT arr[1].1.2 FROM t
```

alrevuelta:
closes #2338

### DIFF
--- a/src/dialect/clickhouse.rs
+++ b/src/dialect/clickhouse.rs
@@ -72,6 +72,11 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    /// See <https://clickhouse.com/docs/sql-reference/functions/tuple-functions#tupleelement>
+    fn supports_tuple_element_access(&self) -> bool {
+        true
+    }
+
     // ClickHouse uses this for some FORMAT expressions in `INSERT` context, e.g. when inserting
     // with FORMAT JSONEachRow a raw JSON key-value expression is valid and expected.
     //
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1381,10 +1381,18 @@ impl<'a> Tokenizer<'a> {
                         return Ok(Some(Token::HexStringLiteral(s2)));
                     }
 
-                    // match one period
-                    if let Some('.') = chars.peek() {
-                        s.push('.');
-                        chars.next();
+                    // match one period. if we've just consumed an integer
+                    // and the previous token is `.`, we're inside a ClickHouse
+                    // tuple element access chain, and the trailing dot belongs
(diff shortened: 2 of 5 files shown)


alrevuelta:
note. this PR does not address `tuple.-1` which seems to be valid clickhouse syntax.

> Extracts an element from a tuple by index or name.

> For access by index, an 1-based numeric index is expected. For access by name, the element name can be provided as a string (works only for named tuples).

> Negative indexes are supported. In this case, the corresponding element is selected, numbered from the end. For example, tuple.-1 is the last element in the tuple.

> An optional third argument specifies a default value which is returned instead of throwing an exception when the accessed element does not exist. All arguments must be constants.

> This function has zero runtime cost and implements the operators x.index and x.name.

source: https://clickhouse.com/docs/sql-reference/functions/tuple-functions#tupleelement

src/tokenizer.rs
```diff
@@ -4303,6 +4331,98 @@ mod tests {
...
     }
 
+    #[test]
+    fn tokenize_clickhouse_tuple_element_access() {
+        let dialects = all_dialects_where(|dialect| dialect.supports_tuple_element_access());
+
+        // After a Word, RBracket, or RParen, `.<digit>` is split into `Period`
+        // and a separate integer `Number`, so the parser can build a
+        // CompoundFieldAccess instead of seeing a single decimal literal.
+        dialects.tokenizes_to(
```
iffyio, line 4341: not sure I follow the patch/tests - are we saying that without this PR, a sql like `t.1` tokenizes to something other than what's being asserted (if so what does the previous behavior tokenized to)?

src/dialect/clickhouse.rs
```diff
@@ -72,6 +72,11 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    /// See <https://clickhouse.com/docs/sql-reference/functions/tuple-functions#tupleelement>
```
iffyio, line 75: does this behavior differ from `parse_compound_expr` handling (if so how, on a first glance they look identical syntax wise)?

PR #2340, opened 2026-05-09
Hive: Support DISTRIBUTE BY and SORT BY in window specs

### LINKED ISSUE #2339: `Hive` dialect does not support `distribute by` or `sort by` in window specs
In Hive 3.1.x, we found that the following query executes successfully:
```sql
select row_number() over(distribute by id sort by update_time desc) as row_num 
from (select 1 as id, current_timestamp as update_time) t
```

Image

Image

However, when generating the AST using **the latest code from the main branch**, the following error occurs:

Image

Reference documentation:

  https://github.com/apache/hive/blob/master/parser/src/java/org/apache/hadoop/hive/ql/parse/IdentifiersParser.g#L293-L297
  https://github.com/apache/hive/blob/master/parser/src/java/org/apache/hadoop/hive/ql/parse/SelectClauseParser.g#L135-L139
  https://github.com/apache/hive/blob/master/parser/src/java/org/apache/hadoop/hive/ql/parse/FromClauseParser.g#L271-L27

wugeer:
Add Hive-gated parsing for window specs that use DISTRIBUTE BY and SORT BY, and preserve the original clause kinds in WindowSpec display output.

fix: https://github.com/apache/datafusion-sqlparser-rs/issues/2339

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2300,12 +2300,11 @@ impl fmt::Display for WindowPartitionByKind {
 }
 
 /// The kind of ordering clause in a window specification.
-#[derive(Debug, Default, Clone, Copy, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub enum WindowOrderByKind {
     /// `ORDER BY`
-    #[default]
     Order,
     /// Hive `SORT BY`
     Sort,
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -2269,10 +2269,56 @@ pub struct WindowSpec {
     pub window_name: Option<Ident>,
     /// `OVER (PARTITION BY ...)`
     pub partition_by: Vec<Expr>,
+    /// The kind of partitioning clause used in the window specification.
+    pub partition_by_kind: WindowPartitionByKind,
     /// `OVER (ORDER BY ...)`
     pub order_by: Vec<OrderByExpr>,
+    /// The kind of ordering clause used in the window specification.
+    pub order_by_kind: WindowOrderByKind,
     /// `OVER (window frame)`
     pub window_frame: Option<WindowFrame>,
+}
+
+/// The kind of partitioning clause in a window specification.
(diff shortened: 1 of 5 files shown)


src/ast/mod.rs
```diff
@@ -2267,12 +2267,58 @@ pub struct WindowSpec {
     pub window_name: Option<Ident>,
     /// `OVER (PARTITION BY ...)`
     pub partition_by: Vec<Expr>,
+    /// The kind of partitioning clause used in the window specification.
+    pub partition_by_kind: WindowPartitionByKind,
```
iffyio, line 2273: for partition and order by kind, can we instead wrap the field with the enum instead of introducing a new field? i.e.
```rust
partition_by: WindowPartitionByKind

// where 
enum WindowPartitionByKind {
     PartitionBy(Vec<Expr>)
     DistributeBy(Vec<Expr>)
}
```

src/ast/mod.rs
```diff
@@ -2267,12 +2267,58 @@ pub struct WindowSpec {
...
     pub window_frame: Option<WindowFrame>,
 }
 
+/// The kind of partitioning clause in a window specification.
+#[derive(Debug, Default, Clone, Copy, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum WindowPartitionByKind {
+    /// `PARTITION BY`
+    #[default]
```
iffyio: is there aneed for these default attributes? (if not we can skip them I think)

src/dialect/mod.rs
```diff
@@ -341,6 +341,17 @@ pub trait Dialect: Debug + Any {
...
     }
 
+    /// Returns true if the dialect supports Hive-style `DISTRIBUTE BY` and `SORT BY`
+    /// clauses within a window specification.
+    ///
+    /// Example
+    /// ```sql
+    /// SELECT row_number() OVER (DISTRIBUTE BY a SORT BY b)
+    /// ```
+    fn supports_window_spec_distribute_sort(&self) -> bool {
```
iffyio: if this doesnt conflict with any existing syntax I think we can skip the dialect method and let the parser always accept the clauses when they show up

src/ast/mod.rs
wugeer: This is a good idea, but the change is very disruptive. Wouldn't it be better to open a separate PR for it?

iffyio: not sure it would make sense as a separate PR given this one, since such a PR would be undoing the changes made by this one

PR #2343, opened 2026-05-20
Make `visitor` feature `no_std`-compatible

LucaCappelletti94:
Enabling the `visitor` feature without `std` (e.g. `default-features = false, features = ["visitor"]` on `wasm32-unknown-unknown`) currently fails with ~3200 errors. The `Visit` / `VisitMut` derive in `sqlparser_derive` hard-codes `::std::ops::ControlFlow` in the generated impls, and `src/ast/visitor.rs` references `String` / `Vec` / `Box` without importing them from `alloc`, so neither path resolves once the crate root evaluates `#![no_std]`.

The existing `compile-no-std` CI job is extended with a `--features visitor` cargo check on `thumbv6m-none-eabi` so this regression cannot land silently again.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -82,6 +82,7 @@ jobs:
         with:
           targets: 'thumbv6m-none-eabi'
       - run: cargo check --no-default-features --target thumbv6m-none-eabi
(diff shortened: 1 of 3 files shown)


iffyio (approved):

PR #2344, opened 2026-05-20
Parser: fix exponential parse time on compound chains

LucaCappelletti94:
`parse_compound_expr` recursed into `parse_subexpr` after every `.`, which re-walks the rest of the chain inside a rollback boundary. On inputs like `IF a.b.c...x.#` the work doubled per chain element, giving 2^N parse times. Switching the inner call to `parse_prefix` removes the redundant traversal: the outer loop already walks the chain, so the resulting AST is identical on valid SQL.

Measured on `PostgreSqlDialect` with three inputs from a libFuzzer corpus, release build:

| Input                                    | Before  | After  |
|------------------------------------------|---------|--------|
| `iF i.D.i.:Fi....` (65 B)                | 419 ms  | 87 us  |
| `iF i.D.i.i. ... .*~` (58 B)             | 893 ms  | 69 us  |
| `if-stf-localtclocal33alt....` (306 B)   | >35 s   | 98 us  |

Regression test in `tests/sqlparser_postgres.rs` runs a 30-element chain with a 5 s timeout. Pre-fix it hits the timeout, post-fix it finishes in well under a millisecond.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2052,9 +2052,14 @@ impl<'a> Parser<'a> {
                         })?;
 
                         match expr {
-                            // `parse_prefix` does not itself follow compound chains, but a
-                            // dialect override could still return a compound expression, so
-                            // keep the flatten arms for safety.
+                            // If we get back a compound field access or identifier,
+                            // we flatten the nested expression.
+                            // For example if the current root is `foo`
+                            // and we get back a compound identifier expression `bar.baz`
+                            // The full expression should be `foo.bar.baz` (i.e.
+                            // a root with an access chain with 2 entries) and not
+                            // `foo.(bar.baz)` (i.e. a root with an access chain with
+                            // 1 entry`).
                             Some(Expr::CompoundFieldAccess { root, access_chain }) => {
                                 chain.push(AccessExpr::Dot(*root));
                                 chain.extend(access_chain);
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2028,14 +2028,16 @@ impl<'a> Parser<'a> {
                         chain.push(AccessExpr::Dot(expr));
                         self.advance_token(); // The consumed placeholder
                     }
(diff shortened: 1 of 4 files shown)


Dandandan (approved):

LucaCappelletti94:
Copilot does not seem to beat the AI bubble allegations lol

src/parser/mod.rs
```diff
@@ -2050,14 +2052,9 @@ impl<'a> Parser<'a> {
...
-                            // we flatten the nested expression.
-                            // For example if the current root is `foo`
-                            // and we get back a compound identifier expression `bar.baz`
-                            // The full expression should be `foo.bar.baz` (i.e.
-                            // a root with an access chain with 2 entries) and not
-                            // `foo.(bar.baz)` (i.e. a root with an access chain with
-                            // 1 entry`).
+                            // `parse_prefix` does not itself follow compound chains, but a
+                            // dialect override could still return a compound expression, so
+                            // keep the flatten arms for safety.
```
iffyio: is the previous comment no longer valid? the new comment reads like it would be confusing without context on the diff in this PR

tests/sqlparser_postgres.rs
```diff
@@ -9243,3 +9243,27 @@ fn parse_lock_table() {
         }
     }
 }
+
+/// `parse_compound_expr` used to do 2^N work on `IF a.b.c...x.#` because every
+/// `.` re-entered `parse_subexpr` over the rest of the chain.
+#[test]
+fn parse_compound_chain_no_exponential_blowup() {
```
iffyio: given the functionality isn't pg specific, can we maybe move this test to be a benchmark test?

LucaCappelletti94: I have moved it to the generic tests, and also added it as a criterion benchmark.

src/parser/mod.rs
LucaCappelletti94: The new comment should be more complete.

iffyio (approved):

PR #2346, opened 2026-05-21
Hive: Support MAP<K, V> column types

### LINKED ISSUE #2345: `Hive` dialect does not support `MAP<K, V>` column types
see: https://hive.apache.org/docs/latest/language/languagemanual-types/

Image

wugeer:
close: https://github.com/apache/datafusion-sqlparser-rs/issues/2345

### DIFF
--- a/tests/sqlparser_hive.rs
+++ b/tests/sqlparser_hive.rs
@@ -21,10 +21,10 @@
 //! is also tested (on the inputs it can handle).
 
 use sqlparser::ast::{
-    ClusteredBy, ColumnOption, CommentDef, CreateFunction, CreateFunctionBody, CreateFunctionUsing,
-    CreateTable, DataType, Expr, Function, FunctionArgumentList, FunctionArguments, Ident,
-    ObjectName, OrderByExpr, OrderByOptions, OrderBySort, SelectItem, Set, Statement, TableFactor,
-    UnaryOperator, Use, Value,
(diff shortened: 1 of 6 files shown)


tests/sqlparser_hive.rs
```diff
@@ -562,6 +562,38 @@ fn test_tample_sample() {
     hive().verified_stmt("SELECT * FROM source TABLESAMPLE (10 ROWS)");
 }
 
+#[test]
+fn parse_create_table_with_map_column_comment() {
+    let sql = "create table tmp.xxx (kv_map map<string, string> comment 'kv col comment');";
```
iffyio: can we have the test only do a round trip test via verified_stmt?

wugeer: 图片

In hive, the `map<string, string>` and `map(string, string)` are not exactly the same, is there a need to change the notation from `(` to `<`, for hive separately?

iffyio: I don't think we necessarily need to, the impl is ok if we're happy with the current roundtrip behavior, my comment was primarily around the test setup, if verified_stmt does not apply to the syntax then an alternative will be `one_statement_parse_to` where we pass the input sql and expected canonical sql to assert

wugeer: I've rethought it, converting `map<string, string>` to `map(string, string)` in the `hive` dialect will cause the parsed sql to fail compilation, which is a functionality breaker, it's better to keep the logic of the sql before and after consistent. (

iffyio (approved):
LGTM! Thanks @wugeer!

PR #2348, opened 2026-05-21
Improve accuracy of supports_string_literal_concatenation_with_newline

yoavcloud:


src/parser/mod.rs
```diff
@@ -12184,6 +12184,15 @@ impl<'a> Parser<'a> {
                         after_newline = true;
                         self.next_token_no_skip();
                     }
+                    // Tokenizer includes the newline in the single line comment
+                    // so we need to check for it specifically here, otherwise the newline will
+                    // not be consumed as a separate token.
+                    Token::Whitespace(Whitespace::SingleLineComment { comment, .. }) => {
```
iffyio: I'm wondering would it make sense to strip out the newline from the tokenizer instead? thinking that feels more like expected behavior for folks that rely on the content of a single line comment

yoavcloud: It would, I just thought it would be a disruptive change. However, I gave it a try and I think it makes sense. Take a look.

iffyio (approved):
LGTM! Thanks @yoavcloud!

PR #2349, opened 2026-05-22
Parser: fix exponential parse time on named-arg chains

LucaCappelletti94:
`parse_function_args` wrapped `parse_expr` in `maybe_parse` for the `<expr> <op> <expr>` named-arg path on PostgreSQL/MSSQL, so rollback re-walked nested function calls and cost compounded to ~2^N. Same shape as #2344 at a different parser site.

Fix: parse the leading expression once with `parse_wildcard_expr`, then `maybe_parse` only the `<op> <expr>` tail.

Bench on `PostgreSqlDialect`:

| n  | before  | after  |
|----|---------|--------|
| 5  | 78 us   | 15 us  |
| 10 | 2.4 ms  | 29 us  |
| 15 | 79 ms   | 42 us  |

Adds a timeout-based regression test and criterion bench mirroring #2344.

**If anyone is curious as to how I am finding these, the TLDR is that I am fuzzing libraries downhill which use sqlparser, and keep getting timeouts because of these exponential cases. Unfortunately, there are still several to be deal with, but there is only so much time in a day.**

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -16,7 +16,7 @@
 // under the License.
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use sqlparser::dialect::GenericDialect;
+use sqlparser::dialect::{GenericDialect, PostgreSqlDialect};
 use sqlparser::keywords::Keyword;
 use sqlparser::parser::Parser;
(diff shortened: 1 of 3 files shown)


LucaCappelletti94:
Closing as superseded by #2352. The new `parse_prefix` per-position failure cache in #2352 catches this PR's chain shape with the same effect: regression input parses in 550us on the amended branch vs >60s on apache main.

PR #2350, opened 2026-05-22
Parser: fix exponential parse time on compound keyword chains

LucaCappelletti94:
Same family as #2344 and #2349. A reserved keyword in field position (e.g. `NOT` in `.not-b.not-b...`) drives `parse_prefix` -> `parse_not` -> `parse_subexpr`, re-walking the chain at every segment and doubling the work. The result is always rejected (`UnaryOp`), so the speculation is wasted, and `parse_identifier` in the existing None branch produces the same `Identifier` directly. The solution I identified is to skip the speculative `parse_prefix` in `parse_compound_expr` when the next token is a `Word` not followed by `(`. Surfaced by subql fuzzing on a 527 B input that took 2.65 s and now takes 270 us.

| N | before | after |
| - | ------ | ----- |
| 5 | 63.3 us | 8.06 us |
| 10 | 1.97 ms | 15.1 us |
| 15 | 65.3 ms | 22.4 us |

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -177,11 +177,35 @@ fn parse_compound_chain(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark parsing pathological compound chains with a reserved keyword in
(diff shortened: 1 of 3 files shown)


iffyio (approved):
LGTM! Thanks for these fixes @LucaCappelletti94!

PR #2352, opened 2026-05-23
Parser: fix exponential parse time on speculative prefix parsing

LucaCappelletti94:
`parse_prefix` speculatively tries the next token as a reserved-word expression head and, on failure, falls back to treating it as an unreserved identifier. Both arms can independently recurse into the same downstream position. Two complementary caches now break the 2^N exploration: an outer per-position failure cache catches chains where both arms fail at every level, and a per-arm cache on the reserved-word `try_parse` catches chains where the arm fails but the fallback succeeds.

Supersedes #2349 (named-arg chain) and #2351 (NOT-prefix chain). Does NOT supersede #2350 (compound-keyword `.not-b` chain, which lives in `parse_compound_expr`).

Same machine, release build, apache main `182eae81` baseline:

| Input | Dialect | Size | Before | After |
|-------|---------|------|--------|-------|
| `if(current_time(...x`, n=30 | PG | 394 B | >60s | 591us |
| `.foo(t--,i)` chain, n=25 (was #2349) | PG | 308 B | >60s | 550us |
| `x-not-b.` chain, n=30 (was #2351) | Generic | 247 B | >60s | 1.45ms |
| `case\t-` chain, n=30 (new) | SQLite | 181 B | >60s | 700us |
| `cAse#f-#9-cAse#f...` fuzz artifact | PG | 142 B | 86ms | 486us |

Regression tests in `tests/sqlparser_common.rs` and benches under `sqlparser_bench` cover the `if(current_time` and `case-case` shapes.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -363,12 +363,29 @@ pub struct Parser<'a> {
     options: ParserOptions,
     /// Ensures the stack does not overflow by limiting recursion depth.
     recursion_counter: RecursionCounter,
-    /// Cached errors from `parse_prefix` calls that returned `Err`. See
+    /// Cached failures from `parse_prefix` calls that returned `Err`. See
     /// [`Parser::parse_prefix`] for the 2^N patterns this guards.
-    failed_prefix_positions: BTreeMap<usize, ParserError>,
-    /// Cached errors from the speculative reserved-word prefix arm. See
+    failed_prefix_positions: BTreeMap<usize, ExprPrefixError>,
+    /// Cached failures from the speculative reserved-word prefix arm. See
     /// [`Parser::parse_prefix`] for the 2^N patterns this guards.
-    failed_reserved_word_prefix_positions: BTreeMap<usize, ParserError>,
+    failed_reserved_word_prefix_positions: BTreeMap<usize, ExprPrefixError>,
+}
+
+/// Copy marker for a [`ParserError`] cached by the `parse_prefix` failure
+/// memoization, so the caches hold no strings.
+#[derive(Debug, Clone, Copy)]
+enum ExprPrefixError {
+    RecursionLimitExceeded,
+    Err,
+}
+
+impl From<&ParserError> for ExprPrefixError {
+    fn from(e: &ParserError) -> Self {
+        match e {
+            ParserError::RecursionLimitExceeded => Self::RecursionLimitExceeded,
+            _ => Self::Err,
+        }
+    }
 }
 
 impl<'a> Parser<'a> {
@@ -1737,14 +1754,22 @@ impl<'a> Parser<'a> {
         // chains where the reserved arm fails but the unreserved fallback
         // succeeds (e.g. `case-case-...c`).
         let start_index = self.index;
-        if let Some(cached) = self.failed_prefix_positions.get(&start_index) {
-            return Err(cached.clone());
+        if let Some(&cached) = self.failed_prefix_positions.get(&start_index) {
+            return Err(self.cached_prefix_error(cached, self.peek_token_ref()));
         }
         let result = self.parse_prefix_inner();
         if let Err(ref e) = result {
-            self.failed_prefix_positions.insert(start_index, e.clone());
+            self.failed_prefix_positions.insert(start_index, e.into());
         }
         result
+    }
+
+    /// Rebuild the error for a cached prefix failure at the `found` token.
+    fn cached_prefix_error(&self, cached: ExprPrefixError, found: &TokenWithSpan) -> ParserError {
+        match cached {
+            ExprPrefixError::RecursionLimitExceeded => ParserError::RecursionLimitExceeded,
+            ExprPrefixError::Err => self.expected_ref::<()>("an expression", found).unwrap_err(),
+        }
     }
 
     fn parse_prefix_inner(&mut self) -> Result<Expr, ParserError> {
@@ -1838,11 +1863,11 @@ impl<'a> Parser<'a> {
                 // succeeds, the overall `parse_prefix` returns `Ok` and the
                 // outer cache never fires. Chains like `case-case-...c`
                 // need this per-arm cache to break the doubling.
-                let try_parse_result = if let Some(cached) = self
+                let try_parse_result = if let Some(&cached) = self
                     .failed_reserved_word_prefix_positions
                     .get(&next_token_index)
                 {
-                    Err(cached.clone())
+                    Err(self.cached_prefix_error(cached, self.get_current_token()))
                 } else {
                     self.try_parse(|parser| parser.parse_expr_prefix_by_reserved_word(&w, span))
                 };
@@ -1861,7 +1886,7 @@ impl<'a> Parser<'a> {
                     // special expression (to maintain backwards compatibility of parsing errors).
                     Err(e) => {
                         self.failed_reserved_word_prefix_positions
-                            .insert(next_token_index, e.clone());
+                            .insert(next_token_index, (&e).into());
                         if !self.dialect.is_reserved_for_identifier(w.keyword) {
                             if let Ok(Some(expr)) = self.maybe_parse(|parser| {
                                 parser.parse_expr_prefix_by_unreserved_word(&w, span)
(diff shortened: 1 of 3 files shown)


LucaCappelletti94:
Me watching yet another fuzzer crash:

image

LucaCappelletti94:
I do not see how to turn this PR back into a draft, but I just found an unhandled corner case. Working on it now.

LucaCappelletti94:
No more exponential-class crashes for a day! Maybe we are done!

src/parser/mod.rs
```diff
@@ -1801,7 +1832,21 @@ impl<'a> Parser<'a> {
                 // We first try to parse the word and following tokens as a special expression, and if that fails,
                 // we rollback and try to parse it as an identifier.
                 let w = w.clone();
-                match self.try_parse(|parser| parser.parse_expr_prefix_by_reserved_word(&w, span)) {
+                // Memoize failed speculative reserved-word parses. When
+                // the reserved arm (CASE, CURRENT_TIME, etc.) does
+                // exponential work but the unreserved fallback ultimately
+                // succeeds, the overall `parse_prefix` returns `Ok` and the
+                // outer cache never fires. Chains like `case-case-...c`
+                // need this per-arm cache to break the doubling.
```
iffyio, line 1869: I wasn't able to follow this comment, what is 'outer cache' and 'break doubling' referring to?

src/parser/mod.rs
```diff
@@ -1717,6 +1731,23 @@ impl<'a> Parser<'a> {
...
+        // The per-arm cache in `parse_prefix_inner` complements this for
+        // chains where the reserved arm fails but the unreserved fallback
+        // succeeds (e.g. `case-case-...c`).
+        let start_index = self.index;
+        if let Some(cached) = self.failed_prefix_positions.get(&start_index) {
+            return Err(cached.clone());
+        }
+        let result = self.parse_prefix_inner();
+        if let Err(ref e) = result {
+            self.failed_prefix_positions.insert(start_index, e.clone());
```
iffyio: wondering about memory usage, how large do we expect the caches to get in the worse case scenarios?

LucaCappelletti94: I have a ~2GB limit on the fuzzer processes and after many billions of iterations it did not manage to OOM - it hit before the ASAN thread number limit than an OOM, so at the very least I believe it is not exponential. I can try and get a scaling though.

src/parser/mod.rs
LucaCappelletti94: Outer cache is the cache in the parent call before the recursive call, and doubling is that at each layer of depth it went in before, the operations doubled (hence the previous 2^layers complexity from before)

src/parser/mod.rs
iffyio: hmm yeah I'm mostly worried about the increased memory usage. A lot of deployments don't have e.g. 2GB to allocate to parsing a sql query. I think the main issue is that the map contains strings, maybe if its a much cheaper/copy object somehow we were looking to store, it might be feasible.

LucaCappelletti94: image

Got a scaling for the cache memory. The two caches are keyed by token index, so each entry is about 40 bytes (8 byte usize key plus a 32 byte ParserError). Panels plot entries, heap String bytes, and total memory vs chain length N for valid SQL, the nested if(current_time(...x chain, and the wide case<TAB>-...c chain.

- On valid SQL both maps stay empty (0 entries at a 27 KiB SELECT).
- Heap String bytes are bounded by recursion depth, not input: past the recursion limit the cached error is RecursionLimitExceeded with no String, so strings peak at ~1.8 KiB and never grow with input.
- Total memory is linear with a small constant: capped at ~4 KiB for deep nesting, at most ~13x input for the adversarial wide shape (159 KiB at a 12 KiB input). Reaching 2 GB would need a ~154 MB single statement.

So it is linear and never exponential, and valid SQL costs nothing.

iffyio: > valid SQL costs nothing

My understanding is that e.g. on the `parse_expr_prefix_by_reserved_word` path, each unique attempt will add an error string into the map, if so then we're potentially looking at one entry in that map per word roughly. is that the the case?

To be clear, its not about a particular amount of memory, main thing is that we're not increasing memory usage of the parser significantly - if the additional memory usage grows as a function of the sql string, already that is problematic, then to improve it we would like to have each entry as minimal as we can, or potentially consider other solutions

LucaCappelletti94: I am not sure I follow, currently the runtime is exponential, so the parser is just failing on these inputs. With the fix, the time requirement becomes linear and the the memory requirements remain linear with input size even for pathological cases

iffyio: essentially the cache is on the order of tokens - so that even a 2x memory increase is linear but doesn't make it ideal.
and it would be nice to improve the behavior if we can. One way I can think of boils down to the fact that the error message is already just a deterministic function of the index we're caching, so from what I can tell, including the actual error string in the cache isn't beneficial. Hence my thinking to switch to storing something light weight.

Here's a rough illustration of that thought in code (naming might be off).

```rust
// Copy type to track the failure type (recursion limit is the only non-expr related error)
#[Copy]
enum ExprPrefixError {
	RecursionLimitExceeded
	Err
}

impl From<&ParserError> for ExprPrefixError {
	fn from(e) -> Self {
		if matches!(e, RecursionLimitExceeded) {
			Self::RecursionLimitExceeded
		} else {
			Self::Err
		}
	}
}

// ...
// We change the cache to store metadata instead of the actual error.
failed_prefix_positions: BTreeMap<usize, ExprPrefixError>
failed_reserved_word_prefix_positions: BTreeMap<usize, ExprPrefixError>

// On a cache miss, we store the metadata
// ...
self.failed_reserved_word_prefix_positions.insert(idx, ExprPrefixError::from(&e));

// On a cache hit, we generate the error message on demand.
// ...
if let Some(cached) = self.failed_reserved_word_prefix_positions {
	return Err(self.generate_expr_prefix_error(cached, idx))
}

// where we can reconstruct the error message given the error context.
fn generate_expr_prefix_error(err, index) {
	match err {
    	ExprPrefixError::RecursionLimitExceeded => ParserError::RecursionLimitExceeded,
        ExprPrefixError::Err => self.expected_at::<()>("an expression", index)
    }
	
}
```

LucaCappelletti94:
image

I implemented the suggested approach: the caches now store a `Copy` marker (`RecursionLimitExceeded` vs generic error) instead of the full `ParserError`, and the message is regenerated on demand on a cache hit. Re-ran the same scaling measurement on both commits, entry counts are identical in every regime, so only the stored value changed.

| | old (cached `ParserError`) | new (`Copy` marker) |
|---|---|---|
| per-entry payload | 40 B (8 key + 32 `ParserError`) | 9 B (8 key + 1 marker) |
| heap strings | nested peaks 1863 B, wide plateaus 598 B | always 0 |
| nested cap (N >= 50) | 3880 B | 873 B (0.22x) |
| wide N=2000 (12 KB sql) | 159598 B, 13.30x input | 35775 B, 2.98x input |
| valid SQL | 0 | 0 |

The caches are now completely string-free and the worst-case overhead drops from ~13.3x to ~3x of the input size. One side effect: when a position is parsed twice, the regenerated error reports the expression start instead of the original deeper error, which changes the message for one test case (`SELECT MAX(interval) FROM tbl` on MsSql), I updated the test accordingly.

src/parser/mod.rs
```diff
@@ -1717,6 +1748,31 @@ impl<'a> Parser<'a> {
...
+        }
+        let result = self.parse_prefix_inner();
+        if let Err(ref e) = result {
+            self.failed_prefix_positions.insert(start_index, e.into());
+        }
+        result
+    }
+
+    /// Rebuild the error for a cached prefix failure at the `found` token.
+    fn cached_prefix_error(&self, cached: ExprPrefixError, found: &TokenWithSpan) -> ParserError {
```
```suggestion
    fn cached_prefix_error(<T>&self, cached: ExprPrefixError, found: &TokenWithSpan) -> Result<T, ParserError> {
```
can we make this return a result so that we avoid the `unwrap_err`?

iffyio (commented):
just a comment to remove an unwrap, otherwise I think this looks good

src/parser/mod.rs
LucaCappelletti94: Done!

iffyio (approved):

PR #2354, opened 2026-05-27
Expose Parser::into_comments

xitep:
* [Back then](https://github.com/apache/datafusion-sqlparser-rs/pull/2107) we didn't make the method public to guarantee the binary search over the comments
* Later in the process we skipped adding out-of-order comments: https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/ast/comments.rs#L40
* therefore exposing the method should be just fine now
* exposing the method allows to use the parser (e.g. extract the statements individually) and still obtain the `Comments` instances without calling `parse_sql_with_comments`


iffyio (approved):
LGTM! Thanks @xitep!

PR #2355, opened 2026-05-27
Fix issue #2353: Incorrect canonicalization of `NATURAL JOIN`

### LINKED ISSUE #2353: Incorrect canonicalization of NATURAL JOIN with INNER JOIN in Snowflake dialect produces non-equivalent query
**Test**: join_precedence
**File**: sqlparser_common.rs

**Description:**
The following query:

SELECT *
FROM t1
NATURAL JOIN t5
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0
WHERE t0.v1 = t1.v0;

is canonicalized for the Snowflake dialect as:

SELECT * FROM t1 NATURAL JOIN (t5 INNER JOIN t0 ON (t0.v1 + t5.v0) > 0) WHERE t0.v1 = t1.v0

However, this rewrite is not semantics-preserving.

NATURAL JOIN implicitly creates a join condition based on all columns
that have the same name in both inputs.

By moving t0 inside the right side of the NATURAL JOIN,
the set of columns visible to the NATURAL JOIN changes.
As a result, additional implicit join predicates may be introduced,
changing the query result.

This behavior violates the expectation that canonicalization should
not alter query semantics.

**Minimal reproducible example:**
CREATE OR REPLACE TEMP TABLE t1 (v0 INT, v1 INT);
CREATE OR REPLACE TEMP TABLE t5 (v0 INT);
CREATE OR REPLACE TEMP TABLE t0 (v1 INT);

INSERT INTO t1 VALUES (1, 100), (1, 200);
INSERT INTO t5 VALUES (1);
INSERT INTO t0 VALUES (1), (2);

-- Q1: original query
SELECT COUNT(*) AS cnt
FROM t1
NATURAL JOIN t5
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0
WHERE t0.v1 = t1.v0;

-- Q2: semantics-preserving canonical form
SELECT COUNT(*) AS cnt
FROM (t1 NATURAL JOIN t5)
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0
WHERE t0.v1 = t1.v0;

-- Q3: canonical form currently produced for Snowflake
SELECT COUNT(*) AS cnt
FROM t1
NATURAL JOIN (t5 INNER JOIN t0 ON (t0.v1 + t5.v0) > 0)
WHERE t0.v1 = t1.v0;

-- Expected: Q1 == Q2
-- Actual:   Q3 returns a different result

revitalkr:
…R JOIN in Snowflake dialect produces non-equivalent query
Closes #2353

The current canonicalization rewrites queries such as:

SELECT *
FROM t1
NATURAL JOIN t5
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0

into:

t1 NATURAL JOIN (t5 INNER JOIN t0 ...)

This transformation is not semantics-preserving.

NATURAL JOIN implicitly generates join predicates based on all shared column names.
Moving t0 into the right-hand side of the NATURAL JOIN changes the set of visible columns, which may introduce additional implicit join conditions and lead to different results.

Minimal example (see issue for full repro):

SELECT *
FROM t1
NATURAL JOIN t5
INNER JOIN t0 ON (t0.v1 + t5.v0) > 0

Expected behavior:
The canonicalization should preserve semantics, e.g.:

(t1 NATURAL JOIN t5) INNER JOIN t0 ...

Fix:
Avoid right-associative parsing when NATURAL JOIN is present, ensuring that NATURAL JOIN is evaluated left-to-right.

Tests:
- Fixed the `join_precedence` test, which previously assumed incorrect Snowflake behavior.
- Verified the actual behavior against Snowflake.
- Updated the test to reflect the behavior observed in Snowflake.

### DIFF
--- a/tests/sqlparser_common.rs
+++ b/tests/sqlparser_common.rs
@@ -17439,8 +17439,7 @@ fn column_check_enforced() {
 
 #[test]
 fn join_precedence() {
-    all_dialects_except(|d| !d.supports_left_associative_joins_without_parens())
-        .verified_query_with_canonical(
+    all_dialects().verified_query_with_canonical(
         "SELECT *
          FROM t1
          NATURAL JOIN t5
@@ -17448,15 +17447,6 @@ fn join_precedence() {
          WHERE t0.v1 = t1.v0",
         // canonical string without parentheses
         "SELECT * FROM t1 NATURAL JOIN t5 INNER JOIN t0 ON (t0.v1 + t5.v0) > 0 WHERE t0.v1 = t1.v0",
-    );
-    all_dialects_except(|d| d.supports_left_associative_joins_without_parens()).verified_query_with_canonical(
-        "SELECT *
-         FROM t1
-         NATURAL JOIN t5
-         INNER JOIN t0 ON (t0.v1 + t5.v0) > 0
-         WHERE t0.v1 = t1.v0",
-        // canonical string with parentheses
-        "SELECT * FROM t1 NATURAL JOIN (t5 INNER JOIN t0 ON (t0.v1 + t5.v0) > 0) WHERE t0.v1 = t1.v0",
     );
 }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15994,6 +15994,7 @@ impl<'a> Parser<'a> {
                 if !self
                     .dialect
                     .supports_left_associative_joins_without_parens()
+                    && !natural
                     && self.peek_parens_less_nested_join()
                 {
                     let joins = self.parse_joins()?;


tests/sqlparser_common.rs
```diff
@@ -17449,15 +17448,6 @@ fn join_precedence() {
         // canonical string without parentheses
         "SELECT * FROM t1 NATURAL JOIN t5 INNER JOIN t0 ON (t0.v1 + t5.v0) > 0 WHERE t0.v1 = t1.v0",
     );
-    all_dialects_except(|d| d.supports_left_associative_joins_without_parens()).verified_query_with_canonical(
```
iffyio, line 17452: we seem to be removing test coverage for the feature `supports_left_associative_joins_without_parens` - I think we might want to replace the removal with something else? otherwise its not clear to me if/why we need to keep the dialect flag around

revitalkr: It is used for queries with join which is not natural join, there's a snowflake dialect test for it, also I added a test for it in my next PR in the common file which is using this feature

iffyio: I guess I'm expecting us to have a test somewhere in the codebase that uses `all_dialects_except(|d| d.supports_left_associative_joins_without_parens())`? Otherwise it looks like we're not testing that feature?

revitalkr: yes, there should be. join_precedence is such a test, only instead of being defined for all_dialects_except(|d| d.supports_left_associative_joins_without_parens() it's in the snowflake only tests file, so it doesn't use it explicitly.
Changed it in the next PR, in which I've added more dialects to behave like snowflake.

iffyio (approved):
Thanks @revitalkr

PR #2359, opened 2026-05-31
Added support for parsing nested object in query for Snowflake

kfirSatori:
Queries parsing like:
`SELECT TRY_CAST(PARSE_JSON('{"obj_field":{"field":"value",}}') AS OBJECT(obj_field OBJECT(
        field VARCHAR)));`

in **Snowflake** are failing since nested object are not supported.

I have added support for it.


kfirSatori:
@iffyio Can you please help resolve the PR?

src/parser/mod.rs
```diff
@@ -13036,6 +13036,31 @@ impl<'a> Parser<'a> {
                     let fields = self.parse_union_type_def()?;
                     Ok(DataType::Union(fields))
                 }
+                Keyword::OBJECT if dialect_is!(dialect is SnowflakeDialect | GenericDialect) => {
```
iffyio, line 13144: can we change this to use a dialect method?

src/parser/mod.rs
```diff
@@ -13036,6 +13036,31 @@ impl<'a> Parser<'a> {
...
+                                let field_name = parser.parse_identifier()?;
+                                let field_type = parser.parse_data_type()?;
+                                Ok(StructField {
+                                    field_name: Some(field_name),
+                                    field_type,
+                                    options: None,
+                                })
+                            })?
+                        };
+                        self.expect_token(&Token::RParen)?;
```
iffyio: can we move this to its own parse_object_data_type() or similar helper method?

tests/sqlparser_snowflake.rs
```diff
@@ -4914,3 +4914,12 @@ fn test_select_dollar_column_from_stage() {
...
     snowflake().verified_stmt("SELECT $1, $2 FROM @mystage1(file_format => 'myformat')");
 }
+
+#[test]
+fn parse_nested_object() {
+    let sql = r#"SELECT TRY_CAST(PARSE_JSON('{"obj_field":{"field":"value",}}') AS OBJECT(obj_field OBJECT(
+        field VARCHAR
+)));"#;
+
+    snowflake().parse_sql_statements(sql).unwrap();
```
iffyio: can we use verified_stmt? also if we can include test cases for mulltiple struct fields, zero struct fields, `OBJECT()` and `OBJECT`

src/ast/data_type.rs
```diff
@@ -465,6 +465,10 @@ pub enum DataType {
     ///
     /// [DuckDB]: https://duckdb.org/docs/sql/data_types/union.html
     Union(Vec<UnionField>),
+    /// Object type, see [Snowflake].
+    ///
+    /// [Snowflake]: https://docs.snowflake.com/en/sql-reference/data-types-semistructured#object
+    Object(Vec<StructField>),
```
iffyio: Can we turn this into a `Object { xxx: Option<Vec<StructField>> }`? the named field syntax makes it possible to extend later on (like attaching spans) and the option iiuc is needed if we're to properly support `OBJECT` syntax (unclear since the behavior suggests it but the tests don't cover it)

src/parser/mod.rs
kfirSatori: Hi @iffyio, i dont understand what do you mean.
i did it the same way done as UNION and NULLABLE

tests/sqlparser_snowflake.rs
kfirSatori: @iffyio Done :)

src/parser/mod.rs
kfirSatori: @iffyio done :)

src/ast/data_type.rs
kfirSatori: @iffyio done

src/parser/mod.rs
iffyio: ah so I meant to use instead a `self.dialect.supportsxxx()` style method - that's [preferred](https://github.com/SatoriCyber/datafusion-sqlparser-rs/blob/e8e783abd2767871b743412333dc317221efd777/src/dialect/mod.rs#L146-L147) over the dialect_is macro

tests/sqlparser_snowflake.rs
```diff
@@ -4912,3 +4912,21 @@ fn test_select_dollar_column_from_stage() {
     // With table function args, without alias
     snowflake().verified_stmt("SELECT $1, $2 FROM @mystage1(file_format => 'myformat')");
 }
+
+#[test]
+fn parse_nested_object() {
+    // nested OBJECT with a single field
+    snowflake().verified_stmt("SELECT TRY_CAST(PARSE_JSON('{\"obj_field\":{\"field\":\"value\",}}') AS OBJECT(obj_field OBJECT(field VARCHAR)))");
```
iffyio, line 4919: heads up after switching to dialect method, these tests can use `all_dialects_where(|d| d.supportsxxx())` to select the dialect instead of hardcoding snowflake only

src/parser/mod.rs
```diff
@@ -3657,6 +3657,32 @@ impl<'a> Parser<'a> {
     /// ```
     ///
     /// [1]: https://duckdb.org/docs/sql/data_types/union.html
+    fn parse_object_data_type(&mut self) -> Result<DataType, ParserError> {
```
iffyio, line 3660: hmm this diff looks a bit off,does the comment above apply to this new function? it looks like it was rather meant for parse_union_type_def

src/ast/data_type.rs
```diff
@@ -465,6 +465,13 @@ pub enum DataType {
     ///
     /// [DuckDB]: https://duckdb.org/docs/sql/data_types/union.html
     Union(Vec<UnionField>),
+    /// Object type, see [Snowflake].
```
iffyio, line 468:
```suggestion
    /// Object type.
```

PR #2360, opened 2026-06-01
Support multiple targets for `SELECT INTO`

iffyio:
Dialects like mysql support

```sql
SELECT id, data INTO @x, @y FROM test.t1;
```
but the parser only expects a single table name in the `INTO` clause.

This adds support for multiple targets in the `INTO` clause.


yoavcloud (approved):

yoavcloud:
LGTM @iffyio!

PR #2361, opened 2026-06-01
Add xmlnamespace support mssql

kfirSatori:
Added support to parse xmlnamespace in mssql, for queris like:
`WITH XMLNAMESPACES ('urn:test' AS ns)
SELECT 1 AS [ns:Value]
FOR XML PATH('ns:Root');`


kfirSatori:
Hi @iffyio CAn you please help with this PR as well?

tests/sqlparser_mssql.rs
```diff
@@ -2923,3 +2923,25 @@ fn parse_mssql_money_constants() {
...
+fn parse_xmlnamespaces_with_cte() {
+    let sql = r#"
+WITH XMLNAMESPACES ('urn:example' AS ns), t AS (
+    SELECT 1 AS id
+)
+SELECT id FROM t
+"#;
+
+    ms().parse_sql_statements(sql).unwrap();
+}
```
iffyio: let's merge these and have them use verified_stmt

src/parser/mod.rs
```diff
@@ -14108,12 +14108,31 @@ impl<'a> Parser<'a> {
...
+                self.expect_token(&Token::RParen)?;
+
+                if self.consume_token(&Token::Comma) {
+                    Some(With {
+                        with_token: with_token.clone().into(),
+                        recursive: self.parse_keyword(Keyword::RECURSIVE),
+                        cte_tables: self.parse_comma_separated(Parser::parse_cte)?,
+                    })
+                } else {
+                    None
```
iffyio: the impl looks odd, is it parsing and dropping the content?

src/dialect/mssql.rs
```diff
@@ -248,6 +248,10 @@ impl Dialect for MsSqlDialect {
             _ => None,
         }
     }
+
+    fn supports_with_xmlnamespaces_clause(&self) -> bool {
```
iffyio, line 253: can we include a link to the documentation of this syntax?

kfirSatori: @iffyio Added documentation link

tests/sqlparser_mssql.rs
```diff
@@ -2926,22 +2926,7 @@ fn parse_mssql_money_constants() {
 
 #[test]
 fn parse_xmlnamespaces() {
-    let sql = r#"WITH XMLNAMESPACES ('urn:test' AS ns)
-SELECT 1 AS [ns:Value]
-FOR XML PATH('ns:Root');"#;
-
-    ms().parse_sql_statements(sql).unwrap();
 
+    ms().verified_stmt("WITH XMLNAMESPACES ('urn:test' AS ns) SELECT 1 AS [ns:Value] FOR XML PATH('ns:Root')");
```
kfirSatori: merged tests and used verifies_stmt() @iffyio

src/parser/mod.rs
kfirSatori: You are correct, I missed the addition to AST fixed now

tests/test_xmlnamespace_integration.rs
```diff
@@ -0,0 +1,133 @@
+/// Test to verify XMLNAMESPACES parsing and AST storage
```
iffyio: hmm I don't think we should introduce a new file for this test. Also let's merge the tests and used either verified_stmt or one_statement_parses_to in tests as I mentioned in the previous review. one simplification of the tests introduced here is to drop the AST assertions, the PR doesn't introduce a new node so its overkill to have each test expiclitly assert the full AST. please have the tests follow existing conventions (this is introducing some println and manual display assertions patterns and is unclear why that's needed)

src/ast/query.rs
```diff
@@ -756,11 +756,23 @@ pub struct With {
     pub recursive: bool,
     /// The list of CTEs declared by this `WITH` clause.
     pub cte_tables: Vec<Cte>,
+    /// Optional XML namespace definitions (`WITH XMLNAMESPACES (...)`).
+    pub xml_namespaces: Vec<XmlNamespaceDefinition>,
```
iffyio, line 760: this feature looks similar to what was [done here](https://github.com/apache/datafusion-sqlparser-rs/pull/2333/changes) for  clickhouse CSEs, such that I'm thinking we essentially want to introduce this feature in that style instead. is xml_namespaces looks like a regular expression (CSE) so that the enum changes might even be verbatim (similarly for the dialect method name)

kfirSatori: Indeed i use the WITH keyword as done with cte_tables, i hope i understood what you have meant.

kfirSatori:
@iffyio hi,
Can we aprove this please?

tests/test_xmlnamespace_integration.rs
kfirSatori: you are correct, i removed the file.

kfirSatori:
@iffyio Hi, kind reminder to check or approve :)

kfirSatori:
@iffyio a kind reminder

src/ast/query.rs
iffyio: hmm the latest diff isn't what I meant, to clarify I don't think we should have `pub xml_namespaces: Vec<XmlNamespaceDefinition>` or a new field at all. basically I meant that we implement effectively what's in the linked PR if that absorbs this feature (i.e. the assumption is that xml_namespace is simply a CSE?)

iffyio:
@kfirSatori just in case note there's an unadressed comment [here](https://github.com/apache/datafusion-sqlparser-rs/pull/2361#discussion_r3841446510) on this PR

kfirSatori:
@iffyio I think we can merege :)

iffyio:
setting this to draft since its no longer pending review

kfirSatori:
> setting this to draft since its no longer pending review

Is there an expectation for it to be merged?

iffyio:
> > setting this to draft since its no longer pending review
> 
> Is there an expectation for it to be merged?

not at the current diff I don't think it should be merged, please let me know if something's unclear about the unadressed comment I linked to earlier

Issue #742, opened 2022-12-01
Expand `TABLE` command support


sarahyurick:
#704 added support for the `CREATE TABLE x AS TABLE y` by adding basic support for the `TABLE` command within the context of a `CREATE` statement. However, as noted in the [Postgres docs](https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE):

"It can be used as a top-level command or as a space-saving syntax variant in parts of complex queries. Only the WITH, UNION, INTERSECT, EXCEPT, ORDER BY, LIMIT, OFFSET, FETCH and FOR locking clauses can be used with TABLE; the WHERE clause and any form of aggregation cannot be used."

These other cases have yet to be investigated.

PR #2364, opened 2026-06-04
Add support for TABLE command in Postgresql

yanghykevin:
Hi all. We're an AWS team using sqlparser-rs for MySQL/PG parsing and plan to contribute a series of dialect compatibility improvements. Happy to coordinate on approach.

### Change Description:
This closes Issue: https://github.com/apache/datafusion-sqlparser-rs/issues/742

Renamed Table struct → ExplicitTable. The original Table struct was created for only one specific use case: CREATE TABLE x AS TABLE y.
The new name matches is more precise and matches [SQL:2016 terminology](https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#explicit-table).

**Grammar**: https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE


tests/sqlparser_postgres.rs
```diff
@@ -9243,3 +9243,43 @@ fn parse_lock_table() {
...
+#[test]
+fn parse_table_command_rejects_invalid() {
+    // Inheritance modifiers are mutually exclusive.
+    assert!(
+        pg().parse_sql_statements("TABLE ONLY films *").is_err(),
+        "expected PG parser to reject: TABLE ONLY films *"
+    );
+
+    for sql in [
+        "TABLE films WHERE x > 0",
```
c2main, line 9574: might be interesting to test: `TABLE db.schema.table` is rejected.

tests/sqlparser_common.rs
```diff
@@ -4643,6 +4645,42 @@ fn parse_create_table_as_table() {
...
 
+#[test]
+fn parse_table_command() {
+    let dialects = all_dialects_where(|d| d.supports_table_command());
+
+    // Top-level.
+    dialects.verified_stmt("TABLE films");
+    // Schema-qualified.
+    dialects.verified_stmt("TABLE myschema.films");
+    // Database-qualified (three-part name).
```
c2main: this is for  MySQL, not PostgreSQL, right ? Maybe a subtle overlap with another PR to come for MySQL ?

c2main:
> Hi all. We're an AWS team using sqlparser-rs for MySQL/PG parsing and plan to contribute a series of dialect compatibility improvements. Happy to coordinate on approach.

Hi. FYI, we have similar plan at Data Bene for Oracle/PG.

tests/sqlparser_postgres.rs
yanghykevin: Currently it is accepted. Based on the existing design philosophy I thought the parser should be permissive on this and relies on downstream semantic analyzer to do verification. But I want to get your opinion on this and will be happy to make the change rejecting this with a trait method

tests/sqlparser_common.rs
yanghykevin: Yes, actually I will enable this command and add mysql testing in the next revision as well since this implementation works for both

docteurklein (approved):

tests/sqlparser_postgres.rs
docteurklein: so this is fixed by your second commit :+1:

docteurklein:
Eager to see this landing as I was about to fix it myself. Thanks!

yanghykevin:
Hi @iffyio 👋 — this PR has a community approval and is passing, but the CI workflow needs maintainer approval to run (and it needs a committer review to merge). Could you take a look when you have a chance? Thanks!

src/dialect/mod.rs
```diff
@@ -1425,6 +1425,29 @@ pub trait Dialect: Debug + Any {
...
+        false
+    }
+
+    /// Returns the maximum number of dot-separated parts allowed in a
+    /// table name for the `TABLE` command. For example, `2` means only
+    /// `schema.table` is accepted; `3` would allow `catalog.schema.table`.
+    ///
+    /// Returns `None` if the dialect does not restrict the number of parts.
+    fn table_command_max_name_parts(&self) -> Option<usize> {
+        None
```
iffyio, line 1448: I think we should be able to drop these dialect methods, the latter two are a bit too specific I think, and in general since we're introducing an entirely new statement it should be fine to have the parser always accept the statement without it conflicting with other dialects

src/ast/query.rs
```diff
@@ -293,28 +293,49 @@ impl fmt::Display for SetQuantifier {
...
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-/// A [`TABLE` command]( https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE)
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-/// A (possibly schema-qualified) table reference used in `FROM` clauses.
-pub struct Table {
-    /// Optional table name (absent for e.g. `TABLE` command without argument).
-    pub table_name: Option<String>,
-    /// Optional schema/catalog name qualifying the table.
-    pub schema_name: Option<String>,
+pub struct ExplicitTable {
```
iffyio, line 304: hmm not sure if there's a need to rename Table (its not clear to me what we gain by doing so)?

src/ast/query.rs
```diff
@@ -293,28 +293,49 @@ impl fmt::Display for SetQuantifier {
...
+
+/// Postgres inheritance-hierarchy modifier for table references.
+///
+/// Controls whether a query against a table includes rows from descendant
+/// tables (inheritance children or partitions). See
+/// <https://www.postgresql.org/docs/current/ddl-inherit.html>.
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+pub enum InheritanceModifier {
```
iffyio, line 319:
```suggestion
pub enum TableInheritanceModifier {
```

src/ast/query.rs
```diff
@@ -293,28 +293,49 @@ impl fmt::Display for SetQuantifier {
...
-pub struct Table {
-    /// Optional table name (absent for e.g. `TABLE` command without argument).
-    pub table_name: Option<String>,
-    /// Optional schema/catalog name qualifying the table.
-    pub schema_name: Option<String>,
+pub struct ExplicitTable {
+    /// The (possibly schema-qualified) table name.
+    pub name: ObjectName,
+    /// Postgres inheritance modifier (`ONLY` or trailing `*`), if present.
+    pub inheritance: InheritanceModifier,
```
iffyio, line 308: this should be an Option instead of introducing a custom None variant in the enum

src/ast/query.rs
```diff
@@ -293,28 +293,49 @@ impl fmt::Display for SetQuantifier {
...
-/// A (possibly schema-qualified) table reference used in `FROM` clauses.
-pub struct Table {
-    /// Optional table name (absent for e.g. `TABLE` command without argument).
-    pub table_name: Option<String>,
-    /// Optional schema/catalog name qualifying the table.
-    pub schema_name: Option<String>,
+pub struct ExplicitTable {
+    /// The (possibly schema-qualified) table name.
+    pub name: ObjectName,
+    /// Postgres inheritance modifier (`ONLY` or trailing `*`), if present.
```
iffyio, line 307:
```suggestion
    /// Inheritance modifier.
```
the rest of the comment is duplicated, the pg part we can drop since other dialects can support it either now or in the future

src/ast/query.rs
```diff
@@ -293,28 +293,49 @@ impl fmt::Display for SetQuantifier {
...
+    pub inheritance: InheritanceModifier,
+}
+
+/// Postgres inheritance-hierarchy modifier for table references.
+///
+/// Controls whether a query against a table includes rows from descendant
+/// tables (inheritance children or partitions). See
+/// <https://www.postgresql.org/docs/current/ddl-inherit.html>.
```
iffyio, line 315:
```suggestion
/// Inheritance-hierarchy modifier for table references.
///
/// Controls whether a query against a table includes rows from descendant
/// tables (inheritance children or partitions).
/// [Postgres]: https://www.postgresql.org/docs/current/ddl-inherit.html.
```

src/parser/mod.rs
```diff
@@ -660,6 +660,10 @@ impl<'a> Parser<'a> {
                     self.prev_token();
                     self.parse_query().map(Into::into)
                 }
+                Keyword::TABLE if self.dialect.supports_table_command() => {
+                    self.prev_token();
+                    self.parse_query().map(Into::into)
```
iffyio, line 665: which statement type is being returned?

PR #2365, opened 2026-06-05
Microsoft sql server: Escape quotes in national string display

lovasoa:
Valid T-SQL (mssql dialect) such as `N' '' '` would not round-trip and serialized to invalid SQL.

This bug is potentially a security vulnerability depending on how the library is used.


lovasoa:
Hi @iffyio ! I think this needs to be prioritized, this bug is potentially exploitable.

iffyio (approved):
LGTM! Thanks @lovasoa!

PR #2367, opened 2026-06-05
PostgreSQL: accept CREATE SEQUENCE options in any clause order

amaksimo:
## Summary

PostgreSQL allows the `CREATE SEQUENCE` option clauses (`INCREMENT`, `MINVALUE`, `MAXVALUE`, `START`, `CACHE`, `CYCLE`) to appear in any order, per the [docs][docs]. The previous parser was positional and rejected pg_dump output:

```sql
CREATE SEQUENCE public.t_id_seq
    AS integer
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1;
```

[docs]: https://www.postgresql.org/docs/current/sql-createsequence.html

## Changes

- Wrap the option-matching arms of `parse_create_sequence_options` in a loop that breaks only when no clause matches. The AST vector preserves source order.
- Doc-comment note on `Statement::CreateSequence` pointing at the PG docs.

Tests in `tests/sqlparser_common.rs` (PG-standard syntax also accepted by `GenericDialect`): pg_dump-shaped input parses to canonical form, two reorderings round-trip with order preserved, repeated-clause case pinned as characterization, and the original `INCREMENT`-first ordering still round-trips.

`cargo test --all-features`, `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings` pass locally.

## History

Re-pitch of #2043 (closed by stale-bot after the author was unavailable). Addresses the review feedback there: pure `else if … else break` loop without a bool sentinel, `verified_stmt` round-trip tests in common, and a duplicate-clause test case.


tests/sqlparser_common.rs
```diff
@@ -3883,6 +3883,96 @@ fn parse_negative_value() {
     );
 }
 
+#[test]
+fn parse_create_sequence_clause_order_independent() {
```
iffyio, line 3887: I think the tests can be simplified by
- using the canonical input and calling verified_stmt instead (no point in testing that `integer` input comes out as `INTEGER`)
- Dropping the AST assertions, the changes don't introduce any AST changes so I don't think the AST check provides much coverage to warrant it
- Merge the tests scenarios into the same function, better yet we can inline them into an existing `CREATE SEQUENCE` if present

LucaCappelletti94:
I have handled this problem in the more general case in #2414

PR #2368, opened 2026-06-06
Support aliased function arguments (expr AS name)

LucaCappelletti94:
PostgreSQL XML functions such as `XMLFOREST` and `XMLATTRIBUTES` take aliased arguments, e.g. `XMLFOREST(a AS x, b)`. This syntax currently fails to parse with "Expected: ), found: AS".

This PR adds a `supports_aliased_function_args()` dialect method (true for PostgreSQL and Generic) and makes the function-call parser wrap `expr AS name` arguments into the existing `Expr::Named` variant, so no new AST types are needed.

Part of the work to support PostgreSQL XML functions (split from #2252, follows #2299).

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1285,13 +1285,12 @@ pub enum Expr {
         /// Struct field definitions.
         fields: Vec<StructField>,
     },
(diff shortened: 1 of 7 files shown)


yoavcloud (approved):
LGTM @LucaCappelletti94 !

PR #2369, opened 2026-06-08
Forbid unsafe code

LucaCappelletti94:
Neither the parser nor the derive crate contains any unsafe code. This adds `unsafe_code = "forbid"` via the `[lints.rust]` table in both manifests so any future unsafe is rejected at compile time. The lint is set per crate because this repo is not a Cargo workspace, so there is no `[workspace.lints]` table to inherit from.


iffyio (approved):

PR #2371, opened 2026-06-08
Add XMLPARSE expression

LucaCappelletti94:
PostgreSQL supports `XMLPARSE(CONTENT expr)` and `XMLPARSE(DOCUMENT expr)`, which cannot be represented as a plain function call because of the leading `CONTENT` or `DOCUMENT` mode keyword. This follows how the parser already models other keyword-syntax functions such as `EXTRACT(field FROM expr)`, adding an `XmlParse` expression with an `XmlParseMode` of `CONTENT` or `DOCUMENT`. It is parsed for dialects that return true from `supports_xml_expressions()`, namely PostgreSQL and Generic, and on every other dialect `xmlparse` keeps parsing as an ordinary function.

Part of the work to support PostgreSQL XML functions (split from #2252, follows #2299 and #2368).


src/ast/mod.rs
```diff
@@ -1253,6 +1290,8 @@ pub enum Expr {
     /// This can represent ANSI SQL `DATE`, `TIME`, and `TIMESTAMP` literals (such as `DATE '2020-01-01'`),
     /// as well as constants of other types (a non-standard PostgreSQL extension).
     TypedString(TypedString),
+    /// XML parse expression: `XMLPARSE(CONTENT|DOCUMENT expr)`.
+    XmlParse(XmlParseExpr),
```
iffyio: I think repr wise, we can piggy back on the existing Function expression, by introducing a new `FunctionArgOperator`

LucaCappelletti94: Done, and simplified somewhat the tests also.

src/ast/mod.rs
```diff
@@ -8023,6 +8023,9 @@ pub enum FunctionArgOperator {
     Colon,
     /// function(arg1 VALUE value1)
     Value,
+    /// function(arg1 value1), with no operator between the name and the value,
+    /// as in PostgreSQL `XMLPARSE(DOCUMENT value)`
```
iffyio, line 8027: Can we add a link to the docs describing the syntax?

src/parser/mod.rs
```diff
@@ -2531,8 +2531,47 @@ impl<'a> Parser<'a> {
...
+        })
+    }
+
     /// Parse a function call expression named by `name` and return it as an `Expr`.
     pub fn parse_function(&mut self, name: ObjectName) -> Result<Expr, ParserError> {
+        if self.dialect.supports_xml_expressions()
+            && Self::is_simple_unquoted_object_name(&name, "xmlparse")
+        {
+            return self.parse_xmlparse_call(name).map(Expr::Function);
+        }
```
iffyio: I think we can simply by moving inside the parse_function_call - [here](https://github.com/LucaCappelletti94/sqlparser-rs/blob/78a68230eadb9298b5931461561d34f852bef4d2/src/parser/mod.rs#L2598) we could do 

```
let mut args = if dialect.supports_xml_expressions() && name = "xmlparse" {
      ...
} else {
    parse_function_argument_list
}
```

src/parser/mod.rs
```diff
@@ -2531,8 +2531,47 @@ impl<'a> Parser<'a> {
...
+    /// word is reserved, so both arrive as plain words rather than keywords.
+    fn parse_xmlparse_call(&mut self, name: ObjectName) -> Result<Function, ParserError> {
+        self.expect_token(&Token::LParen)?;
+        let is_mode = matches!(&self.peek_token_ref().token, Token::Word(word)
+            if word.quote_style.is_none()
+                && (word.value.eq_ignore_ascii_case("content")
+                    || word.value.eq_ignore_ascii_case("document")));
+        if !is_mode {
+            return self.expected_ref("CONTENT or DOCUMENT", self.peek_token_ref());
+        }
```
iffyio: impl wise these should be declared and parsed as keywords if they're missing in the list?

iffyio: oh actually, since we're hardcoding property of the xmlparse and it requires an argument, I don't think either keyword or this code is needed, the parse can just consume the next two tokens as space separated words (the info or content vs document is already being dropped before reaching the AST so the validation isn't necessary at this layer)

src/ast/mod.rs
LucaCappelletti94: Done

src/parser/mod.rs
LucaCappelletti94: Done. Also extended the associated tests, as by fuzzing on the postgres corpus I found statements that were failing like `XMLPARSE(DOCUMENT x) OVER (PARTITION BY y)`

src/parser/mod.rs
LucaCappelletti94: Done. Note the mode word is not dropped, it becomes the `name` of the `FunctionArg::Named`, which keeps the round trip exact.

iffyio (approved):

LucaCappelletti94:
I think there are some CI issues on github at this time and it bounced the merge

PR #2372, opened 2026-06-09
Parser: fix exponential parse time on table-factor paren chains

moshap-firebolt:
`parse_table_factor` speculatively parses `(...)` as a derived table; on
failure it rewinds and falls through to `parse_table_and_joins`, which
recurses back into `parse_table_factor` and re-attempts the same
speculative parse at every deeper paren. Both arms walk the remaining
chain, so on inputs like `SELECT 1 FROM ((((...` work doubles per level.
Caching the position at which the speculative arm failed short-circuits
the second descent.

Measured on `GenericDialect` with `with_recursion_limit(256)`, release
build:

| N  | Before  | After  |
|----|---------|--------|
| 10 |  20 ms  | 57 us  |
| 20 | 820 ms  | 119 us |
| 25 |  2.8 s  | 281 us |
| 30 |  7.9 s  | 345 us |

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -245,6 +245,34 @@ fn parse_prefix_case_chain(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark parsing pathological paren chains that previously caused 2^N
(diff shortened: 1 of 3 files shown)


iffyio (approved):

PR #2373, opened 2026-06-11
CI: Add cargo audit security workflow

LucaCappelletti94:
Adds a Security audit CI workflow that runs cargo-audit against the dependency tree on changes to Cargo.toml or Cargo.lock, on pull requests and the merge queue, and on a daily schedule.

Known RUSTSEC advisories fail CI via cargo audit --deny warnings, and the daily run catches newly published advisories even when dependencies have not changed.

The workflow uses no third-party actions and only contents: read permissions. Running cargo audit --deny warnings locally against the current tree passes with no advisories.


LucaCappelletti94:
@iffyio I would think this is a straightforward one? I just want to add a bit more security given the ongoing dependency attacks that are occurring.

iffyio (approved):

PR #2374, opened 2026-06-11
Parse `ALTER USER` as a synonym for `ALTER ROLE`

LucaCappelletti94:
In PostgreSQL, `ALTER USER` is an alternate spelling of `ALTER ROLE` and accepts the same option syntax (`WITH SUPERUSER`, `PASSWORD '...'`, `CONNECTION LIMIT n`, `RENAME TO ...`, `SET`/`RESET parameter`, etc.). The parser previously routed every dialect's `ALTER USER` to the Snowflake-specific `parse_alter_user`, which rejected all of these PostgreSQL forms.

This adds a `supports_alter_user_as_alter_role` dialect hook (default false, true for PostgreSQL). When set, `ALTER USER` dispatches to the existing `parse_alter_role` path and produces a `Statement::AlterRole`, since the two are semantically identical in PostgreSQL. All other dialects keep the Snowflake behavior unchanged. The shared Snowflake `ALTER USER` test is scoped to dialects that do not enable the new hook.

Reference: https://www.postgresql.org/docs/current/sql-alteruser.html


iffyio (approved):

PR #2376, opened 2026-06-11
Do not consume statement terminator in unparenthesized option lists

LucaCappelletti94:
`CREATE USER <name>` and `ALTER USER <name> SET ...` parse on their own, but `CREATE USER user1; SELECT 1` (and the two `ALTER USER ... SET` forms) fail with `Expected: end of statement, found: SELECT`. The trailing option list is read by `parse_key_value_options`, whose loop drives on `next_token()` and breaks on a `;` that it has already consumed. With the terminator gone, the top-level statement loop sees the next statement's first token directly and errors out. The bug is dialect independent and affects every unparenthesized caller of the helper: `CREATE USER` and both `ALTER USER ... SET` branches. Statements that never reach the helper (`CREATE ROLE`, `DROP USER`, `ALTER USER ... RENAME`) are unaffected, as are the parenthesized callers, which end on `)`.

This puts the semicolon back with `prev_token()` before breaking, so the caller and the top-level loop still see the separator, mirroring how the existing end-keyword arm already does it. The `EOF` arm needs no rewind. Fixing the single helper repairs all three statements at once. A regression test parses each of the three forms followed by `SELECT 1` and asserts two statements result.


tests/sqlparser_common.rs
```diff
@@ -17942,6 +17942,22 @@ fn parse_create_user() {
...
+    // An unparenthesized key-value option list must not swallow the statement
+    // terminator, otherwise any following statement fails to parse. This covers
+    // every unparenthesized caller of `parse_key_value_options`: `CREATE USER`
+    // and both `ALTER USER ... SET` forms.
+    for sql in [
+        "CREATE USER user1; SELECT 1",
+        "ALTER USER user1 SET x = 'y'; SELECT 1",
+        "ALTER USER user1 SET TAG t = 'v'; SELECT 1",
+    ] {
+        let statements = Parser::parse_sql(&GenericDialect {}, sql).unwrap();
```
iffyio: can we change this test to not hardcode the GenericDialect (can it use one of the dialect select helpers to run the test case)?

LucaCappelletti94: Done

iffyio (approved):

PR #2377, opened 2026-06-15
PostgreSQL: support right-deep join chains (deferred ON clauses)

moshap-firebolt:
PostgreSQL accepts join chains where ON clauses are deferred to the end, e.g.:

    t0 JOIN t1 JOIN t2 ON c1 ON c2

This is equivalent to `t0 JOIN (t1 JOIN t2 ON c1) ON c2` per the SQL standard's
right-associative interpretation when parentheses are absent. PostgreSQL documents
this in its JOIN syntax:
https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-JOIN

The parser already handles this correctly for dialects that return `false` from
`supports_left_associative_joins_without_parens()` (Snowflake does this). The
PostgreSQL dialect was missing the override and fell through to the default `true`,
causing a parse error on any right-deep chain of depth >= 3.

This adds the override and a test that verifies both the parse and the canonical
round-trip serialization (with explicit parens).

`cargo test --all-features` passes locally.

### DIFF
--- a/src/dialect/postgresql.rs
+++ b/src/dialect/postgresql.rs
@@ -299,6 +299,12 @@ impl Dialect for PostgreSqlDialect {
         true
     }
 
+    /// PostgreSQL supports right-deep join chains: `t0 JOIN t1 JOIN t2 ON c1 ON c2`
+    /// See: <https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-JOIN>
+    fn supports_left_associative_joins_without_parens(&self) -> bool {
+        false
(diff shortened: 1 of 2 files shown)


iffyio (approved):

iffyio:
@moshap-firebolt could you take a look at the conflict on the branch when you have the time?

moshap-firebolt:
> @moshap-firebolt could you take a look at the conflict on the branch when you have the time?

Sorry - missed this initially - done now!

PR #2381, opened 2026-06-17
Revert Teradata dialect

iffyio:
This reverts commits #2309 and #2329 that I introduced with base teradata dialect support.

I needed the dialect support for my use case and the plan was to further extend the dialect going forward. But I later opted for a less invasive approach that relies on the parser's public apis instead. As a result I no longer have the plans to extend the dialect and it feels wrong to keep it in the repo if its going to be half-implemented and no one else interested in it.

So here I figure to remove it. I can also leave as is if the preference is to keep it


moshap-firebolt:
FWIW - there is a preference to keep it in, even if not complete - it is still helpful with initial take on Teradata workloads

iffyio:
Alright, will leave it in then!

PR #2385, opened 2026-06-18
ClickHouse: Support unparenthesized IN right-hand side

### LINKED ISSUE #2384: `IN` operator rejects a ClickHouse query-parameter placeholder as its right-hand side without parenthesis
## Summary

ClickHouse supports server-side **query parameters** with the `{name:Type}` substitution syntax. When the parameter is an array, it is idiomatic to write it directly as the right-hand side of an `IN` operator:

```sql
WHERE id IN {ids: Array(UUID)}
```

`sqlparser` fails to parse this with:

```
sql parser error: Expected: (, found: { at Line: 1, Column: ...
```

The parser's `IN` handler (`Parser::parse_in`) unconditionally requires a `(` (or `UNNEST`) after the `IN` keyword, so a parameter placeholder standing in for the list is rejected. The same placeholder parses fine everywhere else (e.g. as the RHS of `=`, `BETWEEN`, a function argument), because in prefix position `{name:Type}` is parsed as a `Dictionary` expression.

However, the parenthesized form works:

```sql
WHERE span_id IN ({spanIds: Array(UUID)})   -- parses OK
```

so the two forms are inconsistent, even though ClickHouse accepts both and treats them identically.

## Reproduction (Rust)

```rust
use sqlparser::dialect::ClickHouseDialect;

fn main() {
    let d = ClickHouseDialect {};

    // FAILS: "Expected: (, found: {"
    let bare = "SELECT name FROM users WHERE id IN {ids: Array(UInt64)}";
    println!("{:?}", Parser::parse_sql(&d, bare));

    // OK
    let paren = "SELECT name FROM users WHERE id IN ({ids: Array(UInt64)})";
    println!("{:?}", Parser::parse_sql(&d, paren));

    // OK — placeholder accepted in non-IN positions
    let eq = "SELECT name FROM users WHERE id = {id: UInt64}";
    println!("{:?}", Parser::parse_sql(&d, eq));
}
```

## Reproduction (ClickHouse — confirming the syntax is valid)

```sql
-- 1. Create a dummy in-memory table.
CREATE TABLE users
(
    id   UInt64,
    name String
)
ENGINE = Memory;

-- 2. Insert dummy data.
INSERT INTO users VALUES
    (1, 'alice'),
    (2, 'bob'),
    (3, 'carol');

-- 3. Define an array query parameter for this session.
SET param_ids = '[1, 2]';

-- 4. Use the parameter as the right-hand side of IN — runs successfully.
SELECT name
FROM users
WHERE id IN {ids: Array(UInt64)};
-- → alice
--   bob
```

## Root cause

`Parser::parse_in` [(`src/parser/mod.rs`, ~line 4290 in 0.62.0):](https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/parser/mod.rs#L4382-L4414)

```rust
pub fn parse_in(&mut self, expr: Expr, negated: bool) -> Result<Expr, ParserError> {
    if self.parse_keyword(Keyword::UNNEST) { /* ... */ }
    self.expect_token(&Token::LParen)?;          // <-- rejects `{...}` placeholder
    let in_op = match self.maybe_parse(|p| p.parse_query())? {
        Some(subquery) => Expr::InSubquery { /* ... */ },
        None => Expr::InList { /* ... */ },
    };
    Ok(in_op)
}
```

`expect_token(LParen)` leaves no room for a bare placeholder. In prefix position the same token sequence is handled by  parse_lbrace_expr` → `parse_dictionary` (gated on `Dialect::supports_dictionary_syntax`, which `ClickHouseDialect` returns `true` for), producing `Expr::Dictionary(...)`. So the AST node to represent the placeholder already exists; `parse_in` simply never reaches that path.

## Suggested fix

When the token after `IN` is **not** `(` (and not `UNNEST`), and the dialect supports the placeholder/dictionary syntax, parse a single prefix expression and wrap it as the `IN` right-hand side — e.g. a new `Expr::InList`-like variant carrying a single expression, or
reuse the existing structure with a one-element list. This would make `x IN {p:Array(T)}` parse equivalently to `x IN ({p:Array(T)})`.

This should be gated behind a dialect capability (mirroring `supports_dictionary_syntax` / `supports_in_empty_list`) so other dialects are unaffected.

dinmukhamedm:
Fixes #2384

Note: The fix is a little broader than just the parameterized ClickHouse queries, e.g. `WHERE id IN {ids: Array(UUID)}`.

After opening the issue, I experimented a little more with ClickHouse, and found that it wraps in parenthesis anything on the RHS of `IN`, even for singular values, e.g.

```sql
SELECT * from users WHERE name in 'a';
-- becomes
SELECT * from users WHERE name in ('a');
```

### DIFF
--- a/src/dialect/clickhouse.rs
+++ b/src/dialect/clickhouse.rs
@@ -80,6 +80,7 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    // See <https://clickhouse.com/docs/sql-reference/operators/in>
     fn supports_in_unparenthesized_expr(&self) -> bool {
         true
     }
--- a/src/dialect/clickhouse.rs
+++ b/src/dialect/clickhouse.rs
@@ -80,6 +80,10 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    fn supports_in_unparenthesized_expr(&self) -> bool {
+        true
+    }
+
     /// See <https://clickhouse.com/docs/en/sql-reference/functions#higher-order-functions---operator-and-lambdaparams-expr-function>
     fn supports_lambda_functions(&self) -> bool {
         true
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -435,6 +435,15 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports a bare expression as the right-hand
+    /// side of `IN`, without a parenthesized list — as in `x IN 'a'` or the
+    /// ClickHouse `{name:Type}` query-parameter placeholder `x IN {ids:Array(UInt64)}`.
+    /// The expression is wrapped into a single-element list, matching ClickHouse,
+    /// which reformats `x IN 'a'` to `x IN ('a')`.
+    fn supports_in_unparenthesized_expr(&self) -> bool {
+        false
+    }
+
     /// Returns true if the dialect supports `BEGIN {DEFERRED | IMMEDIATE | EXCLUSIVE | TRY | CATCH} [TRANSACTION]` statements
     fn supports_start_transaction_modifier(&self) -> bool {
         false
@@ -2049,6 +2058,10 @@ mod tests {
 
             fn supports_in_empty_list(&self) -> bool {
                 self.0.supports_in_empty_list()
+            }
+
+            fn supports_in_unparenthesized_expr(&self) -> bool {
+                self.0.supports_in_unparenthesized_expr()
             }
 
             fn convert_type_before_value(&self) -> bool {
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -436,10 +436,7 @@ pub trait Dialect: Debug + Any {
     }
 
     /// Returns true if the dialect supports a bare expression as the right-hand
-    /// side of `IN`, without a parenthesized list — as in `x IN 'a'` or the
-    /// ClickHouse `{name:Type}` query-parameter placeholder `x IN {ids:Array(UInt64)}`.
-    /// The expression is wrapped into a single-element list, matching ClickHouse,
-    /// which reformats `x IN 'a'` to `x IN ('a')`.
+    /// side of `IN`, without a parenthesized list — as in `x IN 'a'`.
     fn supports_in_unparenthesized_expr(&self) -> bool {
         false
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4392,8 +4392,6 @@ impl<'a> Parser<'a> {
                 negated,
             });
         }
-        // ClickHouse accepts a bare expression as the IN RHS (e.g. `x IN 'a'` or
-        // a `{name:Type}` placeholder), wrapping it into a single-element list.
         if self.dialect.supports_in_unparenthesized_expr()
             && self.peek_token_ref().token != Token::LParen
         {
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -4392,6 +4392,17 @@ impl<'a> Parser<'a> {
                 negated,
             });
         }
+        // ClickHouse accepts a bare expression as the IN RHS (e.g. `x IN 'a'` or
+        // a `{name:Type}` placeholder), wrapping it into a single-element list.
+        if self.dialect.supports_in_unparenthesized_expr()
+            && self.peek_token_ref().token != Token::LParen
+        {
+            return Ok(Expr::InList {
+                expr: Box::new(expr),
+                list: vec![self.parse_expr()?],
+                negated,
+            });
+        }
         self.expect_token(&Token::LParen)?;
         let in_op = match self.maybe_parse(|p| p.parse_query())? {
             Some(subquery) => Expr::InSubquery {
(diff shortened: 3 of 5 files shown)


src/parser/mod.rs
```diff
@@ -4392,6 +4392,17 @@ impl<'a> Parser<'a> {
                 negated,
             });
         }
+        // ClickHouse accepts a bare expression as the IN RHS (e.g. `x IN 'a'` or
+        // a `{name:Type}` placeholder), wrapping it into a single-element list.
```
```suggestion
```

tests/sqlparser_clickhouse.rs
```diff
@@ -1846,6 +1846,44 @@ fn parse_inner_array_join() {
     }
 }
 
+#[test]
+fn parse_in_unparenthesized_placeholder() {
+    // ClickHouse `{name:Type}` query-parameter placeholder as the IN RHS, without parens.
+    match clickhouse().expr_parses_to("x IN {ids:Array(UInt64)}", "x IN ({ids: Array(UInt64)})") {
```
iffyio: lets rewrite these tests to use all_dialects_where and verified_stmt, we can remove the test for dialects that donts support the feature

tests/sqlparser_clickhouse.rs
```diff
@@ -1846,6 +1846,44 @@ fn parse_inner_array_join() {
...
+        Expr::InList { list, negated, .. } => {
+            assert!(!negated);
+            assert_eq!(list.len(), 1);
+            assert!(matches!(list[0], Expr::Dictionary(_)));
+        }
+        other => panic!("expected InList, got {other:?}"),
```
iffyio: we can probably drop the ast assertions, I think the funcationality is simple enough to rely only on roundtrip tests

src/dialect/mod.rs
```diff
@@ -435,6 +435,15 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Returns true if the dialect supports a bare expression as the right-hand
+    /// side of `IN`, without a parenthesized list — as in `x IN 'a'` or the
+    /// ClickHouse `{name:Type}` query-parameter placeholder `x IN {ids:Array(UInt64)}`.
+    /// The expression is wrapped into a single-element list, matching ClickHouse,
+    /// which reformats `x IN 'a'` to `x IN ('a')`.
```
```suggestion
    /// side of `IN`, without a parenthesized list — as in `x IN 'a'`.
```

src/dialect/clickhouse.rs
```diff
@@ -80,6 +80,10 @@ impl Dialect for ClickHouseDialect {
         true
     }
 
+    fn supports_in_unparenthesized_expr(&self) -> bool {
```
iffyio, line 84: lets add a link to the docs where the syntax comes from

tests/sqlparser_clickhouse.rs
dinmukhamedm: would you like me to move the tests to the common file?

Otherwise, I should have used the function `clickhouse()` at the bottom of this file that only returns the clickhouse dialect. This will be more consistent with the rest of the tests in this file.

Edit: did both for consistency with the rest of the tests

tests/sqlparser_common.rs
```diff
@@ -2375,15 +2375,45 @@ fn parse_in_unnest() {
...
+    });
+    dialects.expr_parses_to("x IN {ids:Array(UInt64)}", "x IN ({ids: Array(UInt64)})");
+    dialects.expr_parses_to(
+        "x NOT IN {ids:Array(UInt64)}",
+        "x NOT IN ({ids: Array(UInt64)})",
+    );
+    dialects.verified_expr("x IN ({ids: Array(UInt64)})");
+    // Precedence: the trailing `AND` is not swallowed.
+    dialects.verified_expr("x IN ({p: Array(UInt64)}) AND y = 1");
+}
```
iffyio: is there some special handling for dictionaries we're accounting for? if not I think this test we can probably remove

tests/sqlparser_common.rs
```diff
@@ -10834,8 +10864,11 @@ fn parse_position() {
...
 #[test]
 fn parse_position_negative() {
+    // Dialects that accept an unparenthesized IN right-hand side (e.g. ClickHouse)
+    // report a different error here, so exclude them.
     let sql = "SELECT POSITION(foo IN) from bar";
-    let res = parse_sql_statements(sql);
+    let res =
+        all_dialects_except(|d| d.supports_in_unparenthesized_expr()).parse_sql_statements(sql);
     assert_eq!(
         ParserError::ParserError("Expected: (, found: )".to_string()),
```
iffyio, line 10857: maybe we can also add a similar assertion for dialects that do support unparenthesized IN?

tests/sqlparser_common.rs
dinmukhamedm: no, good catch. This was because I started with dict, and then realized clickhouse allows for everything else unparenthesized. Removed now

tests/sqlparser_common.rs
dinmukhamedm: done

iffyio (approved):
LGTM! Thanks @dinmukhamedm!

PR #2387, opened 2026-06-23
Handle nested parentheses in INSERT subquery detection

sabir-akhadov-localstack:
```sql
INSERT INTO t ((SELECT 1));
```

Before this change the statement above failed to parse: a source query wrapped in more than one layer of parentheses was mistaken for an INSERT column list.

When parsing `INSERT INTO t <source>`, the parser must decide whether the tokens following the table name begin an explicit column list `(a, b, ...)` or the source query `(SELECT ...)`. It made this decision in `peek_subquery_start` by matching exactly `( SELECT`, so any extra leading paren tripped the column-list branch — even though the equivalent nested-parenthesized subquery parses fine in other positions.

This changes `peek_subquery_start` to skip any number of leading `(` and treat the tokens as a subquery start as long as at least one paren precedes the `SELECT`, making INSERT consistent with how nested parentheses around subqueries are already handled elsewhere via recursive descent.

A regression test is added in `tests/sqlparser_common.rs` covering `INSERT INTO t ((SELECT 1))` and `INSERT INTO t (((SELECT 1)))`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -18476,24 +18476,19 @@ impl<'a> Parser<'a> {
     }
 
     /// Returns true if the immediate tokens look like the
-    /// beginning of a subquery. `(SELECT ...`
+    /// beginning of a subquery. `(SELECT ...` or `((SELECT ...` etc.
     fn peek_subquery_start(&mut self) -> bool {
-        matches!(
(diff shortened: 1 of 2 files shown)


iffyio (approved):

PR #2388, opened 2026-06-23
Snowflake: parse CREATE WAREHOUSE

sabir-akhadov-localstack:
Snowflake supports `CREATE WAREHOUSE` to provision a virtual warehouse, and `DROP WAREHOUSE` to remove one. Neither parsed before this change.

```sql
CREATE OR REPLACE WAREHOUSE my_wh WITH WAREHOUSE_SIZE = 'XSMALL' AUTO_SUSPEND = 60;
DROP WAREHOUSE my_wh;
```

This adds `Statement::CreateWarehouse` and `ObjectType::Warehouse`. The warehouse properties and parameters are captured as `KeyValueOptions` (the same representation used for `CREATE FILE FORMAT` and friends), so the statement round-trips losslessly rather than discarding everything after the name. The optional leading `WITH` and the spaces around `=` are normalized away on display.

See the [Snowflake `CREATE WAREHOUSE` docs](https://docs.snowflake.com/en/sql-reference/sql/create-warehouse).

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4536,6 +4536,23 @@ pub enum Statement {
         comment: Option<String>,
     },
     /// ```sql
+    /// CREATE [ OR REPLACE ] WAREHOUSE [ IF NOT EXISTS ] <name>
+    ///   [ [ WITH ] <property> = <value> [ ... ] ]
+    /// ```
+    /// Snowflake-specific statement to create a virtual warehouse.
+    ///
+    /// See <https://docs.snowflake.com/en/sql-reference/sql/create-warehouse>
+    CreateWarehouse {
+        /// `OR REPLACE` flag.
+        or_replace: bool,
+        /// `IF NOT EXISTS` flag.
+        if_not_exists: bool,
+        /// Warehouse name.
+        name: ObjectName,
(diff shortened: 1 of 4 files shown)


src/ast/mod.rs
```diff
@@ -4536,6 +4536,23 @@ pub enum Statement {
...
+    CreateWarehouse {
+        /// `OR REPLACE` flag.
+        or_replace: bool,
+        /// `IF NOT EXISTS` flag.
+        if_not_exists: bool,
+        /// Warehouse name.
+        name: ObjectName,
+        /// Warehouse properties and parameters (e.g. `WAREHOUSE_SIZE = 'XSMALL'`).
+        options: KeyValueOptions,
+    },
```
iffyio: Can we change the representation to use a named struct? i.e. `CreateWarehouse(CreateWarehouse)`

sabir-akhadov-localstack: Done in 3fc58f5 — moved to a named `CreateWarehouse` struct following the `CreateUser` pattern (struct + `Display` + `From<CreateWarehouse> for Statement`).

iffyio (approved):

PR #2389, opened 2026-06-27
Allow visits from both visitor and VisitorMut at the same time without using trait disambiguation

### LINKED ISSUE #2067: Rename VisitorMut::visit to VisitorMut::visit_mut
There're cases we need to use `Visitor::visit` and `VisitorMut::visit` in the same context. Since they're of the same name, compiler would be confused.

For example: 

``` Rust
        impl Visitor For NoopVisitor { .. }
        impl VisitorMut for NoopVisitorMut { .. }

        let mut visitor = NoopVisitor {};
        let _ = stmt.visit(&mut visitor);

        let mut visitor_mut = NoopVisitorMut {};
        let _ = stmt.visit(&mut visitor_mut);
```

Would raise the error: 
``` 
the trait bound `NoopVisitorMut: sqlparser::ast::Visitor` is not satisfied
the trait `sqlparser::ast::Visitor` is implemented for `NoopVisitor`
```

docteurklein:
Closes #2067 .

Hello! I just now realize that I could have used [overlapping trait Disambiguation](https://doc.rust-lang.org/rust-by-example/trait/disambiguating.html#disambiguating-overlapping-traits):

```
let _ = Visit::visit(&statements, &mut whitelist);
let _ = VisitMut::visit(&mut statements, &mut VisitOrderBy(order));
```

But here I am with this proposition anyway.

### DIFF
--- a/src/ast/visitor.rs
+++ b/src/ast/visitor.rs
@@ -59,7 +59,7 @@ pub trait VisitMut {
     /// Implementations should call the appropriate mutable visitor hooks to
     /// traverse and allow in-place mutation of child nodes. Returning a
     /// `ControlFlow` value permits early termination of the traversal.
-    fn visit<V: VisitorMut>(&mut self, visitor: &mut V) -> ControlFlow<V::Break>;
+    fn visit_mut<V: VisitorMut>(&mut self, visitor: &mut V) -> ControlFlow<V::Break>;
 }
 
 impl<T: Visit> Visit for Option<T> {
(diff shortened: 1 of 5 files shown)


src/ast/visitor.rs
```diff
@@ -1059,7 +1059,7 @@ mod tests {
 
 #[cfg(test)]
 mod visit_mut_tests {
-    use crate::ast::{Ident, Statement, Value, ValueWithSpan, VisitMut, VisitorMut};
+    use crate::ast::{Ident, Statement, Value, ValueWithSpan, Visit, VisitMut, Visitor, VisitorMut};
```
LucaCappelletti94, line 1062: Just eyeballing, but I would wager you need to run a ` cargo fmt --all -- --check`.

src/ast/visitor.rs
```diff
@@ -1139,4 +1139,26 @@ mod visit_mut_tests {
         let mutated = do_visit_mut("SELECT a, b FROM t", &mut visitor);
         assert_eq!(mutated.to_string(), "SELECT A, B FROM T");
     }
+
+    struct DummyVisitor;
+    impl Visitor for DummyVisitor {
```
LucaCappelletti94, line 1144: I suggest you have the dummy visitor do something, like count nodes

src/ast/visitor.rs
```diff
@@ -1139,4 +1139,26 @@ mod visit_mut_tests {
...
+        type Break = ();
+    }
+
+    struct DummyVisitorMut;
+    impl VisitorMut for DummyVisitorMut {
```
LucaCappelletti94, line 1149: And the dummy visitor mut to edit the nodes it visits.

LucaCappelletti94 (changes requested):
Not a maintainer, just fellow contributor.

Consider rebasing on main, so the branch is up to speed. It is current some 20+ commits behind.

PR #2391, opened 2026-06-29
Postgres: parse LIMIT after the row-locking clause

### LINKED ISSUE #2194: Parsing Error for Supported query in Postgres
### Describe the bug

The PostgreSQL dialect parser fails to parse valid PostgreSQL syntax when `LIMIT` appears after `FOR UPDATE SKIP LOCKED`, even though PostgreSQL accepts both clause orderings.

### To Reproduce
```rust
use sqlparser::dialect::PostgreSqlDialect;

fn main() {
    let sql = r#"
        SELECT "id", "status" 
        FROM "jobs" 
        WHERE "status" = 'pending' 
        ORDER BY "createdAt" 
        FOR UPDATE SKIP LOCKED 
        LIMIT 5
    "#;
    
    let result = Parser::parse_sql(&dialect, sql);
    println!("{:?}", result);
}
```

**Error:**
```
sql parser error: Expected: end of statement, found: LIMIT at Line: X, Column: Y
```

### Expected behavior

The parser should accept both clause orderings since PostgreSQL does:
- ✅ `ORDER BY ... LIMIT 5 FOR UPDATE SKIP LOCKED` (currently works)
- ❌ `ORDER BY ... FOR UPDATE SKIP LOCKED LIMIT 5` (should also work)

Both are semantically identical in PostgreSQL.

### Additional context

- PostgreSQL documentation specifies `FOR UPDATE` should come last, but PostgreSQL actually accepts `LIMIT` after it
- This is commonly used in job queue patterns
- jOOQ had the same issue and fixed it in version 3.19.0 (see [jOOQ/jOOQ#15826](https://github.com/jOOQ/jOOQ/issues/15826))
- Real-world example: https://www.inferable.ai/blog/posts/postgres-skip-locked
- sqlparser version: 0.59

### Tested on PostgreSQL
```sql
-- Both work in PostgreSQL 16:
SELECT * FROM jobs ORDER BY id LIMIT 5 FOR UPDATE SKIP LOCKED;
SELECT * FROM jobs ORDER BY id FOR UPDATE SKIP LOCKED LIMIT 5;
```

truffle-dev:
PostgreSQL accepts `LIMIT`/`OFFSET` both before and after the row-locking clause, but the parser only consumed `LIMIT` before the lock loop, so:

```sql
SELECT * FROM t ORDER BY id FOR UPDATE SKIP LOCKED LIMIT 5;
```

failed with `Expected: end of statement, found: LIMIT`.

This adds a `supports_limit_after_locking_clause` dialect method (off by default, enabled for Postgres) and parses a trailing `LIMIT` clause after the locks when no leading one was present. The AST renders the limit in its canonical position, so the input round-trips to `... LIMIT 5 FOR UPDATE SKIP LOCKED`.

Fixes #2194.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -14480,9 +14480,12 @@ impl<'a> Parser<'a> {
                 }
             }
 
-            // PostgreSQL accepts `LIMIT`/`OFFSET` after the row-locking clause
-            // (e.g. `... FOR UPDATE SKIP LOCKED LIMIT 5`) as well as before it.
-            if limit_clause.is_none() && self.dialect.supports_limit_after_locking_clause() {
+            // Some databases (e.g. PostgreSQL) accept `LIMIT`/`OFFSET` after the
+            // row-locking clause (`... FOR UPDATE SKIP LOCKED LIMIT 5`) as well
+            // as before it. The locking clause above is parsed for every
+            // dialect, so accept a trailing limit here too rather than gating it
+            // behind a dialect flag.
+            if limit_clause.is_none() {
                 limit_clause = self.parse_optional_limit_clause()?;
             }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -14459,7 +14459,7 @@ impl<'a> Parser<'a> {
 
             let order_by = self.parse_optional_order_by()?;
 
-            let limit_clause = self.parse_optional_limit_clause()?;
+            let mut limit_clause = self.parse_optional_limit_clause()?;
 
             let settings = self.parse_settings()?;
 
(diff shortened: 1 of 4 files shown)


src/parser/mod.rs
```diff
@@ -14479,6 +14479,13 @@ impl<'a> Parser<'a> {
                     locks.push(self.parse_lock()?);
                 }
             }
+
+            // PostgreSQL accepts `LIMIT`/`OFFSET` after the row-locking clause
+            // (e.g. `... FOR UPDATE SKIP LOCKED LIMIT 5`) as well as before it.
+            if limit_clause.is_none() && self.dialect.supports_limit_after_locking_clause() {
```
```suggestion
            if limit_clause.is_none()  {
```
I think we can skip the dialect flag entirely here, since the FOR UPDATE isn't dialect guarded either?

truffle-dev: Agreed - the lock loop above isn't dialect-guarded, so gating only the trailing limit was inconsistent. Dropped `supports_limit_after_locking_clause` and its Postgres override and made the condition just `limit_clause.is_none()`. Full suite (all dialect test files + doctests) and clippy are green. c1c4b70

iffyio (approved):
Thanks @truffle-dev!

PR #2392, opened 2026-07-01
Parser: fix exponential parse time on expression-named function arguments

moshap-firebolt:
On dialects with expression-named function arguments
(`supports_named_fn_args_with_expr_name`, e.g. PostgreSQL and MSSQL),
`parse_function_args` speculatively parses the whole argument expression to
detect the `name => value` form, then on failure rewinds and re-parses the
same expression on the unnamed path. Because CAST and CASE fall back to
function-call parsing when their reserved-word form fails, both passes recurse
through the remaining chain, so on inputs like `CAST(CASE (CAST(CASE (...` work
doubles per level. Parsing the leading expression once and then checking for a
named-argument operator removes the redundant traversal; the resulting AST is
identical on valid SQL.

Measured on `PostgreSqlDialect` with `with_recursion_limit(256)`, release
build. Input: `SELECT ` + `CAST(CASE (` repeated N times + `)` repeated 3N:

| N  | Before | After  |
|----|--------|--------|
| 10 | 1.8 s  | 781 us |
| 15 | >30 s  | 967 us |
| 20 | >30 s  | 1.6 ms |
| 25 | >30 s  | 2.3 ms |

A wildcard can never be a named-argument name, so it is routed straight to the
unnamed path, preserving prior behavior. This also makes the "reserved keyword
as a bare function argument" error consistent across dialects (`SELECT
MAX(interval)` now reports the same position regardless of named-argument
support); `test_reserved_keywords_for_identifiers` is updated accordingly.

Regression test in `tests/sqlparser_common.rs` runs a 30-level chain with a 5 s
timeout (hits the timeout pre-fix, finishes in well under a millisecond
post-fix); bench under `sqlparser_bench`.

### DIFF
--- a/sqlparser_bench/benches/sqlparser_bench.rs
+++ b/sqlparser_bench/benches/sqlparser_bench.rs
@@ -273,6 +273,30 @@ fn parse_table_factor_paren_chain(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark parsing pathological `CAST(CASE (CAST(CASE (...` chains that
+/// previously caused 2^N work in `parse_function_args` on dialects with
+/// expression-named function arguments (the argument expression was parsed
+/// once to detect the named form, then re-parsed on the unnamed path).
+fn parse_function_arg_call_chain(c: &mut Criterion) {
+    let mut group = c.benchmark_group("parse_function_arg_call_chain");
+    let dialect = PostgreSqlDialect {};
+
+    for &n in &[10usize, 20, 30] {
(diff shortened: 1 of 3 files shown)


iffyio (approved):

iffyio:
@moshap-firebolt could you take a look at the conflict on the branch?

moshap-firebolt:
> @moshap-firebolt could you take a look at the conflict on the branch?

@iffyio - done!

PR #2394, opened 2026-07-02
feat(pgq): add SQL/PGQ keywords

### LINKED ISSUE #2393: Add SQL/PGQ keywords
Part of #1572.

This issue tracks the first incremental step toward SQL/PGQ support: adding the SQL/PGQ keywords to the tokenizer keyword list.

Out of scope for this issue:
  - SQL/PGQ grammar parsing
  - SQL/PGQ AST nodes
  - `GRAPH_TABLE` and graph pattern semantics

zhjwpku:
closes #2393

- Add SQL/PGQ keywords from ISO/IEC 9075-16:2023(E) to `src/keywords.rs`
- Add tokenizer coverage to ensure these words are recognized as keywords


zhjwpku:
@alamb Hi Andrew, could you please take a look at this? I plan to support the AST and query parser in separate follow-up PRs.

src/tokenizer.rs
```diff
@@ -2673,6 +2673,58 @@ mod tests {
         compare(expected, tokens);
     }
 
+    #[test]
```
LucaCappelletti94, line 2676: This test is tautological, there is no reason to add such a thing.

LucaCappelletti94 (changes requested):
I am not a maintainer, just a fellow contributor. Here are some opinions:

I believe that the syntax used in `GRAPH_TABLE` definitions is completely different from all other SQL operations.

While I agree that the standard now supports some graph queries, this is fairly new ([postgres 19 will introduce some support](https://www.postgresql.org/docs/19/queries-graph.html)), so it is not as easy to determine what is correct and what is not in the commonly used engines.

I suggest you only introduce specifically keywords to handle the SQL part of graph queries and avoid the PGQ specific ones, so if I am not mistaken the keep set would be:

```
DESTINATION
EDGE
GRAPH
GRAPH_TABLE
LABEL
NODE
PROPERTIES
PROPERTY
RELATIONSHIP
VERTEX
```

and the PGQ syntax may be kept as an unparsed string, which downhill user can choose to parse, like we do for the CSV/TSV object after a COPY statement.

PR #2395, opened 2026-07-02
Fix Hive SKEWED BY display

### LINKED ISSUE #1499: typo in Hive SKEWED BY stringification
I haven't run this code and it seems the SKEWED BY syntax can't even be parsed, but I've been looking at all the possible values of ast::Statement and this looks suspicious:

https://github.com/apache/datafusion-sqlparser-rs/blob/e2197eeca9ef2d51a26f29ac23c15515aa668a0f/src/ast/dml.rs#L271

```rust

                write!(
                    f,
                    " SKEWED BY ({})) ON ({})",
                    display_comma_separated(columns),
                    display_comma_separated(on)
                )?;
```

SKEWED BY has one open and two closing parens.

wenxiaojie1:
Fixes #1499.

This removes the extra closing parenthesis when formatting Hive SKEWED BY clauses and adds a display test for the constructed AST.

Checked locally:
- cargo fmt --check
- cargo test display_create_table_with_skewed_by
- git diff --check

### DIFF
--- a/tests/sqlparser_hive.rs
+++ b/tests/sqlparser_hive.rs
@@ -20,11 +20,12 @@
 //! Test SQL syntax specific to Hive. The parser based on the generic dialect
 //! is also tested (on the inputs it can handle).
 
+use sqlparser::ast::helpers::stmt_create_table::CreateTableBuilder;
 use sqlparser::ast::{
-    ClusteredBy, CommentDef, CreateFunction, CreateFunctionBody, CreateFunctionUsing, CreateTable,
-    Expr, Function, FunctionArgumentList, FunctionArguments, Ident, ObjectName, OrderByExpr,
-    OrderByOptions, OrderBySort, SelectItem, Set, Statement, TableFactor, UnaryOperator, Use,
-    Value,
+    ClusteredBy, ColumnDef, CommentDef, CreateFunction, CreateFunctionBody, CreateFunctionUsing,
+    CreateTable, DataType, Expr, Function, FunctionArgumentList, FunctionArguments,
+    HiveDistributionStyle, Ident, ObjectName, OrderByExpr, OrderByOptions, OrderBySort, SelectItem,
+    Set, Statement, TableFactor, UnaryOperator, Use, Value,
 };
 use sqlparser::dialect::{AnsiDialect, GenericDialect, HiveDialect};
 use sqlparser::parser::ParserError;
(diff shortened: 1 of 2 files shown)


tests/sqlparser_hive.rs
```diff
@@ -226,6 +227,28 @@ fn create_table_with_clustered_by() {
...
+            columns: vec![column("id")],
+            on: vec![column("id")],
+            stored_as_directories: true,
+        })
+        .build();
+
+    assert_eq!(
+        stmt.to_string(),
+        "CREATE TABLE test (id STRING) SKEWED BY (id STRING) ON (id STRING) STORED AS DIRECTORIES"
+    );
```
iffyio: can we simplify the test to use the format here for example?
https://github.com/wenxiaojie1/datafusion-sqlparser-rs/blob/30b2a5a686135af090cf7e757070143f36cdff84/tests/sqlparser_hive.rs#L122-L125

tests/sqlparser_hive.rs
```diff
@@ -226,6 +227,28 @@ fn create_table_with_clustered_by() {
     );
 }
 
+#[test]
+fn display_create_table_with_skewed_by() {
```
iffyio:
```suggestion
fn create_table_with_skewed_by() {
```

tests/sqlparser_hive.rs
wenxiaojie1: Thanks. I tried the `verified_stmt` form, but the current Hive parser rejects this syntax with `Expected: end of statement, found: SKEWED`. Issue #1499 explicitly notes that `SKEWED BY` cannot currently be parsed. I kept the explicit AST construction so the test remains scoped to the reported display bug rather than adding parser support in this PR. The focused test still passes.

tests/sqlparser_hive.rs
wenxiaojie1: Applied in f6122ee.

tests/sqlparser_hive.rs
iffyio: not sure I follow, if the parser doesn't accept the syntax then sounds like something more fundamental is wrong if the test was valid?

wenxiaojie1: That separation is intentional for this issue. #1499 explicitly notes that `SKEWED BY` cannot currently be parsed and reports the extra `)` in the existing `CreateTable` display path. `HiveDistributionStyle::SKEWED` is nevertheless a public AST variant that callers can construct programmatically (including through `CreateTableBuilder`), so its malformed SQL output is independently observable.

The AST-construction test exercises that exact display regression without expanding this small fix into parser support; `parse_hive_distribution` still has the separate `TODO: Support parsing for SKEWED distribution style`. If parser support is intended to be a prerequisite for fixing the formatter, I can revise the scope, but I read #1499 as deliberately limited to stringification.

PR #2396, opened 2026-07-02
Accept CREATE OR REPLACE SCHEMA

sabir-akhadov-localstack:
CREATE OR REPLACE SCHEMA did not parse. `CREATE SCHEMA` was handled after the `OR REPLACE` allowlist bail-out in `parse_create`, so any `OR REPLACE` in front of it produced an error:

```sql
CREATE OR REPLACE SCHEMA foo
CREATE OR REPLACE SCHEMA IF NOT EXISTS foo
```

This routes the `SCHEMA` keyword ahead of that bail-out and adds an `or_replace: bool` field to `Statement::CreateSchema`. Display round-trips `OR REPLACE`.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -4356,6 +4356,8 @@ pub enum Statement {
     CreateSchema {
         /// `<schema name> | AUTHORIZATION <schema authorization identifier>  | <schema name>  AUTHORIZATION <schema authorization identifier>`
         schema_name: SchemaName,
+        /// `true` when `OR REPLACE` was present.
+        or_replace: bool,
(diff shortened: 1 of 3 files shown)


c2main:
Just wondering: what dialect does support `CREATE OR REPLACE SCHEMA` ?

sabir-akhadov-localstack:
> Just wondering: what dialect does support `CREATE OR REPLACE SCHEMA` ?

[Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-schema#create-schema-clone)

iffyio (approved):

PR #2398, opened 2026-07-03
Parser: allow `TOP` as an identifier when not a row-limit clause

### LINKED ISSUE #2046: Parser fails when using alias named "TOP" in SELECT statement
Reproducer: Using datafusion-cli https://github.com/apache/datafusion/tree/main/datafusion-cli compiled from hash 5bbdb7eb1

```
> WITH t AS (SELECT 1 AS val) SELECT tt.val FROM t AS tt;
+-----+
| val |
+-----+
| 1   |
+-----+
1 row(s) fetched.
Elapsed 0.050 seconds.

> WITH t AS (SELECT 1 AS val) SELECT top.val FROM t AS top;  🤔 Invalid statement: SQL error: ParserError("Expected: literal int, found: . at Line: 1, Column: 39")
```

Found by SQLStorm https://github.com/apache/datafusion/issues/17698

truffle-dev:
Fixes #2046.

`parse_select` consumed a leading `TOP` keyword as the MSSQL/Snowflake row-limit clause unconditionally, so `TOP` could never be an ordinary column name, alias, or table name in any dialect:

```
SELECT top FROM t          -- Expected: literal int, found: FROM
SELECT top, a FROM t       -- Expected: literal int, found: ,
SELECT top.val FROM t AS top
```

A `TOP` clause is only ever `TOP (expr)` or `TOP <number>`, so `parse_top` accepts nothing else. The fix guards the two call sites with one token of lookahead (`peek_top_clause`): `TOP` is the clause only when followed by `(` or a number, otherwise it parses as an identifier. Disclosure: I'm an AI agent; this was reviewed and tested before opening.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -15115,7 +15115,8 @@ impl<'a> Parser<'a> {
 
         let mut top_before_distinct = false;
         let mut top = None;
-        if self.dialect.supports_top_before_distinct() && self.parse_keyword(Keyword::TOP) {
+        if self.dialect.supports_top_before_distinct() && self.peek_top_clause() {
+            self.expect_keyword(Keyword::TOP)?;
             top = Some(self.parse_top()?);
             top_before_distinct = true;
         }
@@ -15126,7 +15127,8 @@ impl<'a> Parser<'a> {
             self.parse_all_or_distinct()?
         };
 
-        if !self.dialect.supports_top_before_distinct() && self.parse_keyword(Keyword::TOP) {
+        if !self.dialect.supports_top_before_distinct() && self.peek_top_clause() {
+            self.expect_keyword(Keyword::TOP)?;
             top = Some(self.parse_top()?);
         }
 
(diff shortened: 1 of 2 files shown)


src/parser/mod.rs
```diff
@@ -15126,7 +15127,8 @@ impl<'a> Parser<'a> {
             self.parse_all_or_distinct()?
         };
 
-        if !self.dialect.supports_top_before_distinct() && self.parse_keyword(Keyword::TOP) {
+        if !self.dialect.supports_top_before_distinct() && self.peek_top_clause() {
+            self.expect_keyword(Keyword::TOP)?;
             top = Some(self.parse_top()?);
```
iffyio, line 15132: thinking one improvement we could do while we're making this change, can we move the expect_keyword into the parse_top function? that way the function is standalone (we can change the other callsite below to remove its expect_keyword to match)

src/parser/mod.rs
```diff
@@ -19386,6 +19388,18 @@ impl<'a> Parser<'a> {
...
+    /// `TOP` keyword followed by `(` or a number. When `TOP` is followed by
+    /// anything else it is an ordinary identifier (column, alias, or table
+    /// name), not the row-limit clause parsed by [`Self::parse_top`].
+    fn peek_top_clause(&self) -> bool {
+        self.peek_keyword(Keyword::TOP)
+            && matches!(
+                self.peek_nth_token_ref(1).token,
+                Token::LParen | Token::Number(_, _)
+            )
+    }
```
iffyio, line 19401: actually instead of having a peek and a parse for the top clause, we could change parse_top_clause to `maybe_parse_top_clause() -> Option<>`then the callers only need to check the dialect support and call if set

PR #2400, opened 2026-07-10
Apply recursive-protection to the Spanned::span walk

fmcmac:
### What

The `Spanned::span` implementations are plain unbounded recursion — one stack frame per AST node. Computing the span of a deeply nested AST (a deep expression tree, nested subqueries, or a long left-associative boolean chain) recurses as deep as the AST and can overflow the thread stack, aborting the process.

Notably, a query can **parse** successfully — the parser is already guarded by the `recursive-protection` feature — and then overflow **later**, when its span is computed by a downstream consumer (e.g. during query planning). We hit this in production via a downstream planner that calls `.span()` on user-supplied ASTs.

### Fix

The parser already opts into stack-overflow protection behind the `recursive-protection` feature (`recursive::recursive`, backed by `stacker::maybe_grow`); the span walk was simply never given the same treatment. This annotates the recursive `span()` implementations for the container nodes — `Expr`, `SetExpr`, `Query`, `Select`, `SelectItem`, `TableFactor`, `Statement` — so at least one node on every deep-recursion cycle grows the stack on demand.

- No new dependency (the `recursive` crate and feature already exist).
- A no-op when `recursive-protection` is disabled (same `#[cfg_attr(...)]` gating the parser uses).
- Annotating one node per cycle is sufficient for `stacker::maybe_grow` to fire; these seven cover the deep axes (expression nesting, set-op/subquery/join nesting, statement nesting).

### Test

Adds a regression test that builds a deeply nested AST and walks its span on a deliberately small (512 KiB) thread stack: it overflows without the annotations and completes with them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,15 +3,24 @@
 ## General Agent Workflow
 1. You will write unit tests to ensure your code change is working as expected.
 2. You will run the commands in the Pre Commit Checks section below to ensure your change is ready for a pull request.
-3. When instructed to open a PR, you will follow the instructions in the Pull Request Guidelines section below.
+3. You will not open a PR unless explicitly instructed to, following the Pull Request Guidelines section below.
 
 ## General Coding Guidelines
 1. Refrain from adding conditions on specific dialects, such as `dialect_is!(...)` or `dialect_of!(... | ...)`. Instead, define a new function in the `Dialect` trait that describes the condition, so that dialects can turn this condition on more easily.
 2. Make targeted code changes and refrain from refactoring, unless it's absolutely required.
 
+## Code Documentation Guidelines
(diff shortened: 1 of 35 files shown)


iffyio:
@fmcmac could you take a look at the ci failures when you have the time?

iffyio (commented):
Thanks @fmcmac! Could you fix up the check failures on the branch (hopefully the unrelated changes are due to the clippy and should go away after rebasing on main)

alamb:
Is there a real issue / usecase (like a sql query you parsed) that this is fixing?

alamb:
Marking as draft as I think this PR is no longer waiting on feedback and I am trying to make it easier to find PRs in need of review. Please mark it as ready for review when it is ready for another look

PR #2401, opened 2026-07-12
Dialects: implement identifier_quote_style for remaining dialect

takaebato:
`Dialect::identifier_quote_style` returned the default `None` for most dialects even when the database has a documented identifier quote character. 
For example `SnowflakeDialect {}.identifier_quote_style("id")` returned `None`, so downstream SQL generators could not tell how to quote identifiers for these dialects.

This PR implements the method for the remaining dialects: backticks for BigQuery, ClickHouse, Databricks, Hive and Spark, and double quotes for ANSI, DuckDB, Redshift and Snowflake, each with a reference to the database's documentation. `GenericDialect` intentionally keeps returning `None` since it does not assume any specific database.

The existing unit test in `src/dialect/mod.rs` is extended to cover all dialects.

### DIFF
--- a/src/dialect/ansi.rs
+++ b/src/dialect/ansi.rs
@@ -23,6 +23,11 @@ use crate::dialect::Dialect;
 pub struct AnsiDialect {}
 
 impl Dialect for AnsiDialect {
+    /// The SQL standard uses double quotes for delimited identifiers.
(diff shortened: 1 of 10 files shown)


iffyio (approved):
LGTM! Thanks @takaebato!

PR #2402, opened 2026-07-12
Fix clippy lints for Rust 1.97

takaebato:
CI lint currently fails, see https://github.com/apache/datafusion-sqlparser-rs/actions/runs/29178839222/job/86612818972 for an example. 
This PR applies the clippy suggestions as-is, no behavior change.


iffyio (approved):
LGTM! Thanks @takaebato!

PR #2404, opened 2026-07-13
Support inline aggregate WHERE filter (`AGG(expr WHERE cond)`)

moshap-firebolt:
GoogleSQL allows an aggregate call to carry an inline `WHERE` filter, e.g.
`COUNT(* WHERE cond)`, `SUM(x WHERE cond)`, `ARRAY_AGG(x WHERE cond ORDER BY y)`.
It is the in-argument spelling of the standard `AGG(expr) FILTER (WHERE cond)`.

### Changes
- Add `FunctionArgumentClause::Where(Expr)`.
- Parse it in `parse_function_argument_list` for the `BigQuery` and `Generic`
  dialects, before `ORDER BY` / `LIMIT` / `HAVING`.
- `Display` renders `WHERE <expr>`; span is computed from the predicate.
- New test `parse_aggregate_with_where_filter` in `tests/sqlparser_bigquery.rs`.

Docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -18870,6 +18870,14 @@ impl<'a> Parser<'a> {
         let duplicate_treatment = self.parse_duplicate_treatment()?;
         let args = self.parse_comma_separated(Parser::parse_function_args)?;
 
+        // GoogleSQL inline aggregate filter: `AGG(expr WHERE cond [ORDER BY ..])`.
+        // Precedes ORDER BY / LIMIT / HAVING; equivalent to `AGG(expr) FILTER (WHERE cond)`.
+        if dialect_of!(self is GenericDialect | BigQueryDialect)
+            && self.parse_keyword(Keyword::WHERE)
+        {
+            clauses.push(FunctionArgumentClause::Where(self.parse_expr()?));
+        }
+
         if self.dialect.supports_window_function_null_treatment_arg() {
             if let Some(null_treatment) = self.parse_null_treatment()? {
                 clauses.push(FunctionArgumentClause::IgnoreOrRespectNulls(null_treatment));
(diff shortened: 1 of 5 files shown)


src/parser/mod.rs
```diff
@@ -18870,6 +18870,14 @@ impl<'a> Parser<'a> {
         let duplicate_treatment = self.parse_duplicate_treatment()?;
         let args = self.parse_comma_separated(Parser::parse_function_args)?;
 
+        // GoogleSQL inline aggregate filter: `AGG(expr WHERE cond [ORDER BY ..])`.
+        // Precedes ORDER BY / LIMIT / HAVING; equivalent to `AGG(expr) FILTER (WHERE cond)`.
+        if dialect_of!(self is GenericDialect | BigQueryDialect)
+            && self.parse_keyword(Keyword::WHERE)
```
iffyio: can we always have this syntax be accepted by the parser?

moshap-firebolt: Yes - but it is not a standard syntax, it is really a BigQuery's shortcut for standard FILTER(WHERE ...) clause. So I don't think it should be enabled for all dialects. Let me know whether you still want to have it unconditionally

iffyio: Yeah, I think as long as its not ambiguous we can have it be accepted unconditionally

moshap-firebolt: Done!

src/parser/mod.rs
```diff
@@ -18870,6 +18870,14 @@ impl<'a> Parser<'a> {
         let duplicate_treatment = self.parse_duplicate_treatment()?;
         let args = self.parse_comma_separated(Parser::parse_function_args)?;
 
+        // Inline aggregate filter: `AGG(expr WHERE cond [ORDER BY ..])`, popularized by
+        // GoogleSQL. Precedes ORDER BY / LIMIT / HAVING; equivalent to the standard
+        // `AGG(expr) FILTER (WHERE cond)`. Accepted for all dialects: `WHERE` is a reserved
+        // keyword that cannot begin a function argument, so the syntax is unambiguous.
```
```suggestion
```

iffyio (approved):

PR #2406, opened 2026-07-17
Add visitors for ORDER BY and GROUP BY

rs-sac:
Although the expressions therein were visited, there was no pre- or post-handler to alert the visitor to the context of the expressions.


LucaCappelletti94 (approved):
Looks reasonable and adequately tested

iffyio (approved):
LGTM! Thanks @rs-sac!

PR #2418, opened 2026-07-30
Escape closing brackets in bracket-quoted identifiers

### LINKED ISSUE #2409: A bracket-quoted identifier containing `]]` round-trips to SQL that fails to reparse
```rust
use sqlparser::dialect::MsSqlDialect;

fn main() {
    let sql = "SELECT [a]]b]";
    let ast = Parser::parse_sql(&MsSqlDialect {}, sql).unwrap();
    let printed = ast[0].to_string();
    println!("input:    {}", sql);
    println!("printed:  {}", printed);
    let reparsed = Parser::parse_sql(&MsSqlDialect {}, &printed);
    println!("reparsed: {:?}", reparsed);
}
```

```
input:    SELECT [a]]b]
printed:  SELECT [a]b]
reparsed: Err(ParserError("Expected: end of statement, found: ] at Line: 1, Column: 12"))
```

`[a]]b]` is a bracket-quoted identifier whose value is `a]b` (the tokenizer folds the doubled `]]` into a literal `]`). Displaying the parsed AST back to SQL writes `[a]b]` instead, and that string does not parse as the same identifier: it parses as `a` followed by leftover tokens `]b]`.

Tested on sqlparser 0.62.0.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

hdimer:
Fixes #2409.

A bracket-quoted identifier whose value contains `]` (e.g. `[a]]b]`, value `a]b`) serialized back to `[a]b]`, which no longer re-parses. This doubles each `]` on display, mirroring the tokenizer which folds `]]` into `]`.

Redshift also parses nested quoted identifiers like `["a]b"]`, whose value is stored as a complete double-quoted string (`"a]b"`) with a literal inner `]`. The AST doesn't distinguish that from a plain bracket ident, so a value that is a complete `"..."` string is left unchanged; otherwise the `]` are doubled.

Added a round-trip test; the existing Redshift nested-identifier test covers the verbatim case.

Used AI assistance on this; I reviewed and tested it.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -385,7 +385,19 @@ impl fmt::Display for Ident {
                 let escaped = value::escape_quoted_string(&self.value, q);
                 write!(f, "{q}{escaped}{q}")
             }
-            Some('[') => write!(f, "[{}]", self.value),
+            Some('[') => {
+                // Redshift nested quoted identifiers (e.g. `["a]b"]`) store the
+                // value as a complete double-quoted string whose inner `]` is
+                // literal, so leave those unchanged. Otherwise double each `]`,
+                // mirroring the tokenizer folding `]]` into `]`, so the
+                // identifier round-trips (#2409).
+                let v = &self.value;
+                if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
+                    write!(f, "[{v}]")
+                } else {
+                    write!(f, "[{}]", v.replace(']', "]]"))
+                }
+            }
             None => f.write_str(&self.value),
             _ => panic!("unexpected quote style"),
         }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -388,23 +388,21 @@ impl fmt::Display for Ident {
             Some('[') => {
                 let v = &self.value;
                 if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
-                    // A nested double-quoted identifier (e.g. Redshift `["a]b"]`)
-                    // keeps its inner quotes in the value and its `]` is already
-                    // literal, so emit it unchanged.
+                    // Redshift's nested quoted identifier form (`["a b"]`) keeps
+                    // the inner quotes in the value, where `]` is literal.
                     write!(f, "[{v}]")
                 } else {
-                    // Double a lone `]` so the identifier round-trips, but leave
-                    // an already-doubled `]]` intact: in the tokenizer's no-escape
-                    // mode the value still holds the raw `]]`, and re-doubling it
-                    // would corrupt the identifier.
-                    write!(f, "[")?;
-                    let mut rest = v.as_str();
-                    while let Some(pos) = rest.find(']') {
-                        write!(f, "{}]]", &rest[..pos])?;
-                        let after = &rest[pos + 1..];
-                        rest = after.strip_prefix(']').unwrap_or(after);
+                    // Each `]` is doubled. `escape_quoted_string` can't be reused:
+                    // it skips a quote that is already doubled, which corrupts a
+                    // value containing `]]`.
+                    f.write_str("[")?;
+                    for (i, part) in v.split(']').enumerate() {
+                        if i > 0 {
+                            f.write_str("]]")?;
+                        }
+                        f.write_str(part)?;
                     }
-                    write!(f, "{rest}]")
+                    f.write_str("]")
                 }
             }
             None => f.write_str(&self.value),
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -386,16 +386,25 @@ impl fmt::Display for Ident {
                 write!(f, "{q}{escaped}{q}")
             }
             Some('[') => {
-                // Redshift nested quoted identifiers (e.g. `["a]b"]`) store the
-                // value as a complete double-quoted string whose inner `]` is
-                // literal, so leave those unchanged. Otherwise double each `]`,
-                // mirroring the tokenizer folding `]]` into `]`, so the
-                // identifier round-trips (#2409).
                 let v = &self.value;
                 if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
+                    // A nested double-quoted identifier (e.g. Redshift `["a]b"]`)
+                    // keeps its inner quotes in the value and its `]` is already
+                    // literal, so emit it unchanged.
                     write!(f, "[{v}]")
                 } else {
-                    write!(f, "[{}]", v.replace(']', "]]"))
+                    // Double a lone `]` so the identifier round-trips, but leave
+                    // an already-doubled `]]` intact: in the tokenizer's no-escape
+                    // mode the value still holds the raw `]]`, and re-doubling it
+                    // would corrupt the identifier.
+                    write!(f, "[")?;
+                    let mut rest = v.as_str();
+                    while let Some(pos) = rest.find(']') {
+                        write!(f, "{}]]", &rest[..pos])?;
+                        let after = &rest[pos + 1..];
+                        rest = after.strip_prefix(']').unwrap_or(after);
+                    }
+                    write!(f, "{rest}]")
                 }
             }
             None => f.write_str(&self.value),
--- a/tests/sqlparser_mssql.rs
+++ b/tests/sqlparser_mssql.rs
@@ -939,14 +939,28 @@ fn parse_table_name_in_square_brackets() {
 
 #[test]
 fn parse_bracket_identifier_with_escaped_closing_bracket() {
-    // A bracket-quoted identifier whose value contains `]` must serialize
-    // with the bracket doubled so it round-trips. See #2409.
+    // A `]` inside a bracket-quoted identifier is escaped by doubling it, so
+    // `[a]]b]` denotes the identifier `a]b`.
     let select = ms().verified_only_select("SELECT [a]]b]");
     assert_eq!(
         &Expr::Identifier(Ident::with_quote('[', "a]b")),
         expr_from_projection(&select.projection[0]),
     );
-    ms().verified_stmt("SELECT [a]]b] FROM [c]]d]");
+
+    // Round-trips regardless of where the escaped `]` sits.
+    for sql in [
+        "SELECT [a]]b] FROM [c]]d]",
+        "SELECT []]]",      // the identifier is a single `]`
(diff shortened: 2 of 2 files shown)


src/ast/mod.rs
```diff
@@ -385,7 +385,19 @@ impl fmt::Display for Ident {
                 let escaped = value::escape_quoted_string(&self.value, q);
                 write!(f, "{q}{escaped}{q}")
             }
-            Some('[') => write!(f, "[{}]", self.value),
+            Some('[') => {
+                // Redshift nested quoted identifiers (e.g. `["a]b"]`) store the
+                // value as a complete double-quoted string whose inner `]` is
+                // literal, so leave those unchanged. Otherwise double each `]`,
+                // mirroring the tokenizer folding `]]` into `]`, so the
+                // identifier round-trips (#2409).
```
LucaCappelletti94: No need to refer to issues in the code itself. Code comments should refer to the present code, not previous states of the code, save for areas very prone to regressions and reiterated attempts.

tests/sqlparser_mssql.rs
```diff
@@ -937,6 +937,18 @@ fn parse_table_name_in_square_brackets() {
     );
 }
 
+#[test]
+fn parse_bracket_identifier_with_escaped_closing_bracket() {
+    // A bracket-quoted identifier whose value contains `]` must serialize
+    // with the bracket doubled so it round-trips. See #2409.
```
LucaCappelletti94: Same here: comments should be about the code, not about previous states of the code. This is information appropriate for the PR post or commit, not code.

tests/sqlparser_mssql.rs
```diff
@@ -937,6 +937,18 @@ fn parse_table_name_in_square_brackets() {
     );
 }
 
+#[test]
+fn parse_bracket_identifier_with_escaped_closing_bracket() {
```
LucaCappelletti94, line 941: Testing is insufficient and the proposed changes are currently regressing working cases. For instance, in unescaped mode, `SELECT [a]]b]` in current main parses correctly, while with this PR it parses to `a]]]]b`.

src/ast/mod.rs
```diff
@@ -385,7 +385,19 @@ impl fmt::Display for Ident {
...
+                let v = &self.value;
+                if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
+                    write!(f, "[{v}]")
+                } else {
+                    write!(f, "[{}]", v.replace(']', "]]"))
```
LucaCappelletti94: Here you are adding allocation in a hot path with that replace, in a write. I believe it is unnecessary to do so, just print what you need without reallocating the string.

src/ast/mod.rs
```diff
@@ -385,7 +385,19 @@ impl fmt::Display for Ident {
...
             }
-            Some('[') => write!(f, "[{}]", self.value),
+            Some('[') => {
+                // Redshift nested quoted identifiers (e.g. `["a]b"]`) store the
+                // value as a complete double-quoted string whose inner `]` is
+                // literal, so leave those unchanged. Otherwise double each `]`,
+                // mirroring the tokenizer folding `]]` into `]`, so the
+                // identifier round-trips (#2409).
+                let v = &self.value;
+                if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
```
LucaCappelletti94, line 390: I believe there are several other cases where the current solution fails, other than the one I reported in the test comment. I suggest you fuzz with seeding/use round trip prop tests this code before pushing the next iteration of your PR, since it would have most likely immediately caught the mentioned problems, even if you vibe code this thing using AI. It is a very effective support tool when you lean on code generation, as it gives you test inputs generation and invariant testing.

LucaCappelletti94 (changes requested):

hdimer:
Thanks for the thorough review. Pushed a revision:

- **No-escape regression:** the escaping is now idempotent. It doubles a lone `]` but leaves an already-doubled `]]` alone, so `[a]]b]` round-trips under `unescape(false)` instead of becoming `[a]]]]b]`. This mirrors what `EscapeQuotedString` already does for `"`. Added a no-escape round-trip test plus a corpus covering leading/trailing/multiple `]`.
- **Allocation:** gone. The escaper writes straight into the formatter, no `replace()`.
- **Comments:** dropped the issue references.
- On testing: I brute-forced round-trips over `]`/letter combinations in both escape modes to check for other failures; the corpus captures the boundaries.

One thing worth flagging: like the `"` path, a value that already contains `]]` is treated as already-escaped, so an identifier whose logical value holds two consecutive literal `]` (e.g. default-mode `[a]]]]b]`) now serializes toward the no-escape reading rather than re-doubling. That is the same trade-off `escape_quoted_string` makes for `""`, so bracket and double-quoted identifiers now behave consistently.

hdimer:
Thanks — you're right the last iteration still had gaps. The root issue is that `Ident` doesn't record which escape mode produced its `value`, so one `Display` can't round-trip both:

- default (`unescape=true`): `[a]]]]b]` parses to value `a]]b`, which must serialize back to `[a]]]]b]`.
- no-escape (`unescape=false`): `[a]]b]` parses to value `a]]b`, which must serialize back to `[a]]b]`.

Same `value`, two required outputs — so the "leave `]]` intact" approach I pushed is wrong for the first case (confirmed: `[a]]]]b]` round-trips to `[a]]b]` today).

I'd propose making `[` consistent with `escape_quoted_string` for `"`: always double a literal `]`. That fixes default-mode round-tripping for every identifier (the actual #2409 bug), and mirrors `"`, whose no-escape mode is already display-lossy — so this makes `[` consistent rather than a special case. I'll back it with the round-trip property test you suggested.

Prefer that, or would you rather keep no-escape round-tripping for `[` (accepting that an identifier containing consecutive `]]` can't be represented in default mode)?

src/ast/mod.rs
```diff
@@ -385,7 +385,28 @@ impl fmt::Display for Ident {
...
+                    // Double a lone `]` so the identifier round-trips, but leave
+                    // an already-doubled `]]` intact: in the tokenizer's no-escape
+                    // mode the value still holds the raw `]]`, and re-doubling it
+                    // would corrupt the identifier.
+                    write!(f, "[")?;
+                    let mut rest = v.as_str();
+                    while let Some(pos) = rest.find(']') {
+                        write!(f, "{}]]", &rest[..pos])?;
+                        let after = &rest[pos + 1..];
+                        rest = after.strip_prefix(']').unwrap_or(after);
```
LucaCappelletti94: `ms().verified_stmt("SELECT [a]]]]b]")` fails because value `a]]b` formats as `[a]]b]` and reparses as `a]b`.

src/ast/mod.rs
```diff
@@ -385,7 +385,28 @@ impl fmt::Display for Ident {
                 let escaped = value::escape_quoted_string(&self.value, q);
                 write!(f, "{q}{escaped}{q}")
             }
-            Some('[') => write!(f, "[{}]", self.value),
+            Some('[') => {
+                let v = &self.value;
+                if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
```
LucaCappelletti94, line 390: `ms().verified_stmt(r#"SELECT ["a]]b"]"#)` fails because valid MSSQL and SQLite value `"a]b"` formats as malformed `SELECT ["a]b"]`.

LucaCappelletti94 (changes requested):
Please use a display-loop fuzzer and most likely criterion benchmarks. Fixing these corner cases will be hard, and I do not think that it will be easy to vibe code your way to an efficient and correct version without an extremely clear automatic feedback mechanism. I suggest you trim down the code comments as well.

hdimer:
You're right, and that's a cleaner repro than the one I had. The escaping is missing on the format side, so anything containing `]` round-trips into invalid SQL. I'll add that as a test and fix the formatter for both MSSQL and SQLite.

hdimer:
Pushed. Both repros were real; here is where they landed.

**`SELECT [a]]]]b]`**: fixed. Dropped the "leave an already-doubled `]]` alone" rule, so every `]` in the value is doubled unconditionally. That is the only escaping that round-trips a value holding consecutive `]`.

The cost is that no-escape mode no longer round-trips brackets: `with_unescape(false)` keeps the raw `]]` in the value, so `[a]]b]` now serializes to `[a]]]]b]`. The two are mutually exclusive for a `Display` that can see neither the dialect nor `ParserOptions`, and the old rule picked the no-escape side at the price of corrupting every real identifier containing `]]`. I removed the no-escape assertion I added last round, since it pinned the behaviour this change rejects. Nothing on main covered it.

**Round-trip test**: replaced the hand-listed cases with an exhaustive one. Every value of length 1..=5 over `]`, `"`, `a` and `é`, serialized and reparsed against MSSQL and SQLite, asserting both the recovered AST and byte-identical re-serialization. It fails against the previous formatter (smallest failing value: `]]`) and against main, so it is doing work.

**`SELECT ["a]]b"]`**: I do not think this one can be fixed here, and it needs your call.

In MSSQL that SQL gives the value `"a]b"`. In Redshift, `["=]"]` gives the value `"=]"` through the nested-identifier path, where the inner `]` is literal and must not be doubled. `test_parse_nested_quoted_identifier` pins exactly that. Same shape of value, opposite required output, and `Ident::fmt` has no dialect. Doubling makes that test fail with `["=]]"]`. I could not find a content-based rule that separates the two honestly, since either value can legitimately come from either dialect.

Worth knowing: that verbatim branch is already unsound on main, with no `]` involved. On current main, `redshift().verified_stmt(r#"SELECT 1 AS ["a""b"] FROM a"#)` fails. The nested path folds `""` to `"`, giving the value `"a"b"`, which serializes to `["a"b"]` and no longer parses. Same root cause, predates this PR, untested.

The structural fix, if you want it: have the Redshift nested form store `["foo"]` as `Ident { value: "foo", quote_style: '"' }` instead of stuffing the quotes into the value. `redshift.rs` already documents `foo`, `"foo"` and `["foo"]` as meaning the same thing, so that is arguably the right AST regardless. The special case in `Ident::fmt` then deletes itself and both of your repros pass. The visible change is that `["foo"]` re-serializes as `"foo"`, so `test_parse_nested_quoted_identifier` would move to `one_statement_parses_to`. Happy to do it as a separate PR, but it changes Redshift output so I did not fold it in here.

Comments trimmed.

PR #2419, opened 2026-07-31
Honour operator precedence in SQLite `REGEXP`, `MATCH` and `GLOB`

LucaCappelletti94:
`SQLiteDialect::parse_infix` parsed the right operand of `REGEXP`, `MATCH`, and `GLOB` with `parse_expr`, and previously lead to mutating the meaning of boolean operations constructed with it.

For instance, `a REGEXP 'p' AND b = 1` parsed as a `REGEXP ('p' AND b = 1)` instead of `(a REGEXP 'p') AND (b = 1)`.

The hook was ignoring the precedence its caller passed, so using it fixes all three. Nothing errored before, and Display adds no parentheses, so the wrong tree reprinted as the original text. That is why a text round trip never caught it and the new test asserts on the tree instead.

### DIFF
--- a/src/dialect/sqlite.rs
+++ b/src/dialect/sqlite.rs
@@ -79,7 +79,7 @@ impl Dialect for SQLiteDialect {
         &self,
         parser: &mut crate::parser::Parser,
         expr: &crate::ast::Expr,
-        _precedence: u8,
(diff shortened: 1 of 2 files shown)


iffyio (approved):

PR #2411, opened 2026-07-27
Add recursion checks for parse_data_type

iffyio:
Data types can be nested, e.g. `MAP<MAP<...>>`.
This adds the usual recursion guards to ensure
that we don't overflow the stack on extreme/pathological input.

PR #2199, opened 2026-02-06
Fixed overflow error, recursion counter was not included for parenthesis

LucaCappelletti94:
Closes #2198

PR #2422, opened 2026-08-04
Add recursion checks for parse_interval

shuvamk:
```sql
SELECT INTERVAL INTERVAL INTERVAL ... 1   -- 600 repetitions, ~5 KB
```

This aborts the process with `fatal runtime error: stack overflow` on a 2 MiB stack
(the default for a spawned thread) instead of returning `RecursionLimitExceeded`.

For dialects where `require_interval_qualifier()` is false, `parse_interval` calls
`parse_prefix` directly rather than `parse_expr`, so it never passes through the
recursion counter that lives in `parse_subexpr`. `parse_prefix` re-enters
`parse_interval` on the next INTERVAL keyword, leaving that cycle unguarded.
Reproduced on generic, duckdb, snowflake and postgres.

The fix applies the same two guards #2199 used for the parenthesis gap; the 600-level
input then returns `Err(RecursionLimitExceeded)`.

Test added beside the existing recursion-limit tests in `tests/sqlparser_common.rs`.
Without the parser change it fails as a stack-overflow abort. Same defect class as
#2411, on a different path.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3368,7 +3368,10 @@ impl<'a> Parser<'a> {
     /// ```
     ///
     /// Note that we do not currently attempt to parse the quoted value.
+    #[cfg_attr(feature = "recursive-protection", recursive::recursive)]
     pub fn parse_interval(&mut self) -> Result<Expr, ParserError> {
+        let _guard = self.recursion_counter.try_decrease()?;
+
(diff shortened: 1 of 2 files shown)


LucaCappelletti94 (approved):
Looks sensible.

iffyio (approved):
LGTM! Thanks @shuvamk!

PR #2423, opened 2026-08-04
MySQL: print a space before PARTITION when displaying a table factor

shuvamk:
```sql
SELECT * FROM employees PARTITION (p0, p2)
```

round-trips to `SELECT * FROM employeesPARTITION (p0, p2)`, which re-parses as a table *function* named `employeesPARTITION` taking arguments `p0` and `p2` — a different AST, not just different whitespace.

`Display for TableFactor::Table` writes `PARTITION (...)` with no leading space, unlike the sibling clauses in the same arm (` WITH ORDINALITY`, ` {alias}`, ` WITH (...)`). The fix adds the space.

Test is `parse_table_partition_selection` in `tests/sqlparser_mysql.rs`, covering `SELECT`/`UPDATE`/`DELETE`, an alias, a join and the missing-parenthesis error path; it fails without the change. Table-level `PARTITION` selection had no test coverage before.

### DIFF
--- a/src/ast/query.rs
+++ b/src/ast/query.rs
@@ -2216,7 +2216,7 @@ impl fmt::Display for TableFactor {
                     json_path.fmt(f)?;
                 }
                 if !partitions.is_empty() {
-                    write!(f, "PARTITION ({})", display_comma_separated(partitions))?;
+                    write!(f, " PARTITION ({})", display_comma_separated(partitions))?;
                 }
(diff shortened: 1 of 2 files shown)


LucaCappelletti94 (approved):

alamb (approved):
Thank you @LucaCappelletti94  and @shuvamk

PR #2424, opened 2026-08-04
Print a space between adjacent prefix unary operators

shuvamk:
`Expr::UnaryOp`'s `Display` writes `{op}{expr}` with no separator, so a nested unary operand is printed glued to the operator:

```sql
SELECT ~ ~ 1
```

prints as `SELECT ~~1`, which no longer reparses on any dialect (`Expected: an expression, found: ~~`). `SELECT - -1` prints as `SELECT --1` the same way; that one only breaks where `--` opens a line comment, so MySQL is unaffected.

On PostgreSQL it changes meaning rather than failing, because `@@` is its own operator: `SELECT @ @ 1` (abs of abs) prints as `SELECT @@1`, which reparses as `UnaryOperator::DoubleAt` applied to `1`.

The fix inserts a space only where the operator's glyph and the operand's leading glyph would tokenize as a single token. Pairs that cannot merge are left alone, so `++a`, `+@a`, `!!~a` and `-NOT a` print unchanged.

Tests are `parse_nested_unary_ops` and `parse_nested_pg_unary_ops`, which fail without the source change, plus `parse_adjacent_unary_ops_that_do_not_combine` and `parse_adjacent_pg_unary_ops_that_do_not_combine` pinning the pairs that must stay glued. The `AGENTS.md` pre-commit checks are clean locally.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1963,7 +1963,8 @@ impl fmt::Display for Expr {
                         | UnaryOperator::DoubleAt
                         | UnaryOperator::QuestionDash
                         | UnaryOperator::QuestionPipe
-                ) {
+                ) || matches!(expr.as_ref(), Expr::UnaryOp { .. })
+                {
                     write!(f, "{op} {expr}")
                 } else {
                     write!(f, "{op}{expr}")
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -196,6 +196,36 @@ where
     DisplaySeparated { slice, sep: ", " }
 }
 
+/// Returns true when the prefix operator `op` written directly before `operand`
+/// would be tokenized as a single operator rather than as two.
+fn lexes_as_one_operator(op: &UnaryOperator, operand: &Expr) -> bool {
+    let Expr::UnaryOp { op: leading, .. } = operand else {
+        return false;
+    };
+    match op {
+        UnaryOperator::Minus
+        | UnaryOperator::BitwiseNot
+        | UnaryOperator::PGSquareRoot
+        | UnaryOperator::PGCubeRoot => !matches!(
+            leading,
+            UnaryOperator::Not | UnaryOperator::PGPostfixFactorial
+        ),
+        UnaryOperator::BangNot => {
+            matches!(leading, UnaryOperator::BangNot | UnaryOperator::BitwiseNot)
+        }
+        UnaryOperator::PGAbs => matches!(
+            leading,
+            UnaryOperator::AtDashAt
+                | UnaryOperator::DoubleAt
+                | UnaryOperator::Minus
+                | UnaryOperator::PGAbs
+                | UnaryOperator::QuestionDash
+                | UnaryOperator::QuestionPipe
+        ),
+        _ => false,
+    }
+}
+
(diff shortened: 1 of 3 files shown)


LucaCappelletti94 (commented):
At this time, the current implementation fires a bit too broadly, including also for cases where it should not, such as:

```sql
SELECT ++a     -->  SELECT + +a
SELECT +@a     -->  SELECT + @a
SELECT !!~a    -->  SELECT !! ~a
SELECT -NOT a  -->  SELECT - NOT a
```

I believe it may be desirable to have a more focused alternative.

shuvamk:
Good catch — narrowed it. A space is now inserted only where the operator's glyph and the operand's leading glyph would tokenize as a single token, so `++a`, `+@a`, `!!~a` and `-NOT a` all print unchanged. (`+` and `!!` are `consume_and_return` in the tokenizer, so they never absorb what follows; a keyword operand can't merge either.)

I swept every prefix-operator pair across all dialects before and after the change and the set of round-trip breaks is identical, so the narrowing costs no coverage.

It does leave two cosmetic spaces, `SELECT - !a` and `SELECT ~ !a`. Those pairs do merge into a Postgres custom operator, and they're only unreachable because `BangNot` parses on Hive/Databricks alone — I'd rather not have `Display` encode that coincidence, but it's one token to exclude `BangNot` if you want it exactly minimal.

Your four cases are pinned in `parse_adjacent_unary_ops_that_do_not_combine` and `parse_adjacent_pg_unary_ops_that_do_not_combine`.

src/ast/mod.rs
```diff
@@ -196,6 +196,36 @@ where
     DisplaySeparated { slice, sep: ", " }
 }
 
+/// Returns true when the prefix operator `op` written directly before `operand`
+/// would be tokenized as a single operator rather than as two.
+fn lexes_as_one_operator(op: &UnaryOperator, operand: &Expr) -> bool {
+    let Expr::UnaryOp { op: leading, .. } = operand else {
```
LucaCappelletti94, line 202: Nested unary operator also reaches the output through a lower-precedence parent.

I suggest you use a display-loop fuzzer to catch these errors during your next review iteration, and consider also using a criterion benchmark to ensure the performance do not fall.

tests/sqlparser_postgres.rs
```diff
@@ -9684,6 +9684,40 @@ fn parse_right_deep_join_chain() {
...
+    let select = pg().verified_only_select("SELECT @@ 1");
+    assert_eq!(
+        SelectItem::UnnamedExpr(Expr::UnaryOp {
+            op: UnaryOperator::DoubleAt,
+            expr: Box::new(Expr::value(number("1"))),
+        }),
+        select.projection[0]
+    );
+
+    pg().verified_stmt("SELECT |/ |/1");
```
LucaCappelletti94, line 9710: Here is an example of why you should pin the lower-precedence operand case, as before the source change it fails with `left: "SELECT ~ -a * 2"`, `right: "SELECT ~-a * 2"`.

```suggestion
    pg().verified_stmt("SELECT |/ |/1");
    pg().verified_stmt("SELECT ~ -a * 2");
```

LucaCappelletti94 (changes requested):

PR #2425, opened 2026-08-04
Databricks: support CREATE TABLE USING, MAP<K, V>, LONG, DIV and the pipe operator

shuvamk:
Databricks SQL is built on Spark SQL, but `DatabricksDialect` was missing flags `SparkSqlDialect` sets. These fail on `DatabricksDialect` and parse on `SparkSqlDialect`:

```sql
CREATE TABLE t (id BIGINT) USING DELTA
CREATE TABLE t (m MAP<STRING, INT>)
SELECT 10 div 3
SELECT * FROM t |> WHERE x > 1
```

`CREATE TABLE t (id LONG)` parses, but as a `DataType::Custom`, not `DataType::BigInt(None)`.

`div` and pipeline syntax are documented for Databricks SQL:
https://docs.databricks.com/aws/en/sql/language-manual/functions/div
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-pipeline

The methods shared with Spark delegate to `SparkSqlDialect` rather than repeating the value, the same shape `RedshiftSqlDialect` uses for `PostgreSqlDialect`.

`parse_pipeline_operator_negative_tests` now uses `|> CALL 123` instead of `123invalid`: enabling the pipe operator brings Databricks into that assertion, and `supports_numeric_prefix` makes it fail at a different point. Both inputs fail identically on the dialects already covered.

Tests are in `tests/sqlparser_databricks.rs` and `tests/sqlparser_common.rs`; the five Databricks ones fail with `src/dialect/databricks.rs` reverted. The `AGENTS.md` pre-commit checks are clean locally.

### DIFF
--- a/src/dialect/databricks.rs
+++ b/src/dialect/databricks.rs
@@ -15,7 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::ast::Expr;
 use crate::dialect::Dialect;
+use crate::parser::{Parser, ParserError};
+
+use super::SparkSqlDialect;
 
 /// A [`Dialect`] for [Databricks SQL](https://www.databricks.com/)
 ///
@@ -116,18 +120,35 @@ impl Dialect for DatabricksDialect {
 
     /// See <https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using>
     fn supports_create_table_using(&self) -> bool {
-        true
+        SparkSqlDialect {}.supports_create_table_using()
     }
 
     /// `LONG` is an alias for `BIGINT` in Databricks SQL.
     ///
     /// See <https://docs.databricks.com/aws/en/sql/language-manual/data-types/bigint-type>
     fn supports_long_type_as_bigint(&self) -> bool {
-        true
+        SparkSqlDialect {}.supports_long_type_as_bigint()
     }
 
     /// See <https://docs.databricks.com/aws/en/sql/language-manual/data-types/map-type>
     fn supports_map_literal_with_angle_brackets(&self) -> bool {
-        true
+        SparkSqlDialect {}.supports_map_literal_with_angle_brackets()
+    }
+
+    /// See <https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-pipeline>
+    fn supports_pipe_operator(&self) -> bool {
+        SparkSqlDialect {}.supports_pipe_operator()
+    }
+
+    /// Parse the `DIV` keyword as integer division.
+    ///
+    /// See <https://docs.databricks.com/aws/en/sql/language-manual/functions/div>
+    fn parse_infix(
+        &self,
+        parser: &mut Parser,
+        expr: &Expr,
+        precedence: u8,
+    ) -> Option<Result<Expr, ParserError>> {
+        SparkSqlDialect {}.parse_infix(parser, expr, precedence)
     }
 }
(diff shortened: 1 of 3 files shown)


tests/sqlparser_databricks.rs
```diff
@@ -737,3 +737,55 @@ fn parse_cte_without_as() {
...
+                    format: Ident::new("DELTA")
+                })
+            );
+        }
+        s => panic!("Unexpected statement: {s:?}"),
+    }
+
+    databricks().verified_stmt("CREATE TABLE IF NOT EXISTS t (id BIGINT) USING PARQUET");
+
+    assert!(all_dialects_where(|d| !d.supports_create_table_using())
```
LucaCappelletti94: Consider adding the analogous assert also for the other flags you have set

LucaCappelletti94: Also, I believe that since this assertion regards all dialects except databricks, it should not be in the databricks tests, but in common.

src/dialect/databricks.rs
```diff
@@ -113,4 +113,21 @@ impl Dialect for DatabricksDialect {
...
+        true
+    }
+
+    /// `LONG` is an alias for `BIGINT` in Databricks SQL.
+    ///
+    /// See <https://docs.databricks.com/aws/en/sql/language-manual/data-types/bigint-type>
+    fn supports_long_type_as_bigint(&self) -> bool {
+        true
+    }
+
```
LucaCappelletti94, line 132: While you are adding support for Spark flags in Databricks, I believe you should also add `supports_pipe_operator` and `parse_infix` for `DIV`.

src/dialect/databricks.rs
LucaCappelletti94, line 31: I believe there is some argument to be made that, since several of these properties need to be kept aligned between Spark and Databricks, it may be desirable to call directly Spark methods in these methods instead of duplicating the scalar value.

LucaCappelletti94 (changes requested):

shuvamk:
Thanks — all four addressed in the follow-up commit.

`DIV` and `|>` are both documented for Databricks SQL, so I added them: `supports_pipe_operator`, and `parse_infix` delegating to Spark, which handles `DIV` and nothing else.

On delegating the flags: I converted the ones Spark and Databricks share, but left `supports_numeric_prefix`, `supports_table_versioning` and `supports_optimize_table` as they were — Spark doesn't override those, so delegating would take them from `true` to the trait default `false` and stop `SELECT 59901_user`, `VERSION AS OF` and `OPTIMIZE` from parsing on Databricks. Happy to convert the remaining shared ones too if you'd like the whole block consistent.

The cross-dialect asserts moved to `tests/sqlparser_common.rs` so Databricks is covered, with the flag-off negatives alongside them. `MAP<K, V>` needed a per-dialect loop rather than one `is_err()` — the ungated dialects split into two different error messages.

Title and description updated for the added scope.

tests/sqlparser_databricks.rs
```diff
@@ -757,3 +757,62 @@ fn parse_databricks_query_entry_points() {
...
+#[test]
+fn parse_div_operator() {
+    databricks().one_statement_parses_to("SELECT 10 div 3", "SELECT 10 DIV 3");
+    databricks().one_statement_parses_to("SELECT c1 div c2 FROM t", "SELECT c1 DIV c2 FROM t");
+}
+
+#[test]
+fn parse_pipe_operator() {
+    databricks().verified_stmt("SELECT * FROM t |> WHERE x > 1 |> SELECT x AS y |> ORDER BY y");
+}
```
LucaCappelletti94: You should drop `parse_pipe_operator`, as Databricks already enables `|>` on main (#2478).

```suggestion
```

LucaCappelletti94 (changes requested):
You should retitle to `Databricks: support CREATE TABLE USING, LONG and DIV` and drop the `MAP<K, V>` and `|>` examples, as #2478 already made both parse on main.

PR #2426, opened 2026-08-04
Reject invalid numeric underscore separators

### LINKED ISSUE #2421: postgres parser: multiple or trailing underscores should error
We can use underscores in numbers, however if we have more than one in succession `10__00` or trailing `10_00_` this should error according to postgres:

```sql
postgres=# select version();
                                                         version
--------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 18.2 (Debian 18.2-1.pgdg13+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
(1 row)

postgres=# select 10__00;
ERROR:  trailing junk after numeric literal at or near "10__00"
LINE 1: select 10__00;
               ^
postgres=# select 10_00_;
ERROR:  trailing junk after numeric literal at or near "10_00_"
LINE 1: select 10_00_;
               ^
postgres=# select 10_00;
 ?column?
----------
     1000
(1 row)
```

This seems to currently be accepted:

```rust
// tests/sqlparser_postgres.rs

#[test]
fn parse_incorrect_underscores() {
    let sql = "SELECT 10_00";
    let result = pg().parse_sql_statements(sql);
    assert!(result.is_err());

    let sql = "SELECT 10_00_";
    let result = pg().parse_sql_statements(sql);
    assert!(result.is_err());
}
```

- test fails since both of the statements parse successfully

fyrsta7:
## Summary

Fixes #2421 by rejecting invalid numeric underscore separators for dialects that support underscores in numeric literals.

## Details

Numeric separators should appear between digits. This keeps valid literals such as `10_000` and `1_000.123_456`, while rejecting trailing, consecutive, and decimal-point-adjacent underscores such as `10_00_`, `10___0`, `1_000.123_`, and `1._000`.

The change adds coverage at both the tokenizer and parser test levels.

## Validation

- `git diff --check`
- `cargo fmt --all --check`
- `cargo test tokenize_numeric_literal_underscore -- --nocapture`
- `cargo test parse_numeric_literal_underscore --test sqlparser_common -- --nocapture`
- `cargo test`

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1360,36 +1360,12 @@ impl<'a> Tokenizer<'a> {
                         );
                     }
 
-                    // Some dialects support underscore as number separator
-                    // There can only be one at a time and it must be followed by another digit
-                    let is_number_separator = |ch: char, next_char: Option<char>| {
-                        self.dialect.supports_numeric_literal_underscores()
-                            && ch == '_'
-                            && next_char.is_some_and(|next_ch| next_ch.is_ascii_hexdigit())
-                    };
-
-                    let mut s = peeking_next_take_while(chars, |ch, next_ch| {
-                        ch.is_ascii_digit() || is_number_separator(ch, next_ch)
-                    });
-
-                    if self.dialect.supports_numeric_literal_underscores()
-                        && chars.peek() == Some(&'_')
-                    {
-                        return self.tokenizer_error(chars.location(), "Unexpected character '_'");
-                    }
+                    let mut s = self.tokenize_number_part(chars, |ch| ch.is_ascii_digit())?;
 
                     // match binary literal that starts with 0x
                     if s == "0" && chars.peek() == Some(&'x') {
                         chars.next();
-                        let s2 = peeking_next_take_while(chars, |ch, next_ch| {
-                            ch.is_ascii_hexdigit() || is_number_separator(ch, next_ch)
-                        });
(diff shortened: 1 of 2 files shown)


src/tokenizer.rs
```diff
@@ -1372,12 +1372,24 @@ impl<'a> Tokenizer<'a> {
                         ch.is_ascii_digit() || is_number_separator(ch, next_ch)
                     });
 
+                    if self.dialect.supports_numeric_literal_underscores()
+                        && chars.peek() == Some(&'_')
+                    {
+                        return self.tokenizer_error(chars.location(), "Unexpected character '_'");
+                    }
```
iffyio: i couldn't see if/why this block was necessary given `is_number_separator` checks with a lookahead - is there a fallthrough code that captures the trailing underscore somewhere below?

I have the same comments about some of the other similar diffs below, is there a way to solve this in a more holistic manner vs the individual if checks - the latter is unclear/not-obvious how they interact with the rest of the code

fyrsta7: Thanks for pointing this out. You are right that the first check was redundant. I replaced the separate checks with one helper that validates separators consistently across integer, fractional, hexadecimal, and exponent parts, and added coverage for the related edge cases. The full test suite, Clippy, and formatting checks pass. Thanks again for the careful review.

iffyio (approved):
LGTM! Thanks @fyrsta7!

PR #2431, opened 2026-08-05
Parse `REVOKE GRANT OPTION FOR`

LucaCappelletti94:
`GRANT ... WITH GRANT OPTION` parses today but the `REVOKE` statement that withdraws that option did not, and this PR closes that gap.

Also, I noticed that there was a `test_revoke_with_cascade` which excluded `MySqlDialect`, which has no reason to be I could identify, so I removed the filter.

### DIFF
--- a/src/ast/dcl.rs
+++ b/src/ast/dcl.rs
@@ -486,10 +486,14 @@ impl From<Grant> for crate::ast::Statement {
 }
 
 /// REVOKE privileges ON objects FROM grantees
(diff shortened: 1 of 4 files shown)


LucaCappelletti94:
@iffyio this is the last one of the < 100 lines PRs I opened, the other ones are all larger. I tried to keep them as tiny as possible, but most of those are regarding adding statements support so they tend to be naturally larger.

iffyio (approved):

PR #2435, opened 2026-08-07
Support `NO INHERIT` modifier on `CHECK` constraints

LucaCappelletti94:
PostgreSQL marks constraints that child tables should not inherit with `NO INHERIT`, and `pg_dump` emits it for any schema using table inheritance. The parser rejected it in every position:

```sql
ALTER TABLE docs ADD CONSTRAINT c CHECK (id > 0) NO INHERIT;
-- Expected: end of statement, found: NO

CREATE TABLE t (a INT, CONSTRAINT c CHECK (a > 0) NO INHERIT);
-- Expected: ',' or ')' after column definition, found: NO
 ```
 
 Relevant doc: https://www.postgresql.org/docs/current/sql-createtable.html

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -1939,7 +1939,7 @@ pub enum ColumnOption {
     /// [<constraint_characteristics>]
     /// `).
     ForeignKey(ForeignKeyConstraint),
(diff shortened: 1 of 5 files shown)


iffyio (approved):

PR #2438, opened 2026-08-13
Snowflake: Add support for ->> (pipe) operator for chaining SQL stmts

BenSatori:
The ->> (flow/pipe) operator chains multiple SQL statements in Snowflake.

**Before:** Parser rejected pipe chains like `SELECT * FROM t1 ->> SELECT * FROM $1`

**After:** Parser correctly handles:
- Chained statements with ->> operator
- Pipe result references ($1, $2, $3, etc.)
- Round-trip SQL serialization

**Example:**
```sql
SELECT * FROM users WHERE active = true ->> SELECT COUNT(*) FROM $1
```

Changes:

- Added Statement::Pipe AST variant for chained statements
- Added TableFactor::PipeResultScan for $n result references
- Extended Snowflake dialect with pipe operator support
- Comprehensive test coverage (2 test functions, multiple edge cases)

Snowflake docs:
https://docs.snowflake.com/en/sql-reference/operators-flow#pipe

### DIFF
--- a/src/dialect/generic.rs
+++ b/src/dialect/generic.rs
@@ -320,4 +320,8 @@ impl Dialect for GenericDialect {
     fn supports_aliased_function_args(&self) -> bool {
         true
     }
+
+    fn supports_long_arrow_pipe_operator(&self) -> bool {
+        true
+    }
 }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -726,11 +726,11 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Does the dialect support the Snowflake `->>`  flow/pipe operator for chaining
-    /// SQL statements? e.g. `SELECT * FROM t ->> SELECT * FROM $1`
+    /// Does the dialect support the `->>` flow/pipe operator for chaining SQL statements?
+    /// e.g. `SELECT * FROM t ->> SELECT * FROM $1`
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/operators-flow>
-    fn supports_snowflake_pipe_operator(&self) -> bool {
+    fn supports_long_arrow_pipe_operator(&self) -> bool {
         false
     }
 
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -726,7 +726,7 @@ pub trait Dialect: Debug + Any {
         false
     }
 
-    /// Does the dialect support the Snowflake `-->`  flow/pipe operator for chaining
+    /// Does the dialect support the Snowflake `->>`  flow/pipe operator for chaining
     /// SQL statements? e.g. `SELECT * FROM t ->> SELECT * FROM $1`
     ///
     /// See <https://docs.snowflake.com/en/sql-reference/operators-flow>
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -726,6 +726,14 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Does the dialect support the Snowflake `-->`  flow/pipe operator for chaining
+    /// SQL statements? e.g. `SELECT * FROM t ->> SELECT * FROM $1`
+    ///
+    /// See <https://docs.snowflake.com/en/sql-reference/operators-flow>
+    fn supports_snowflake_pipe_operator(&self) -> bool {
+        false
+    }
+
     /// Does the dialect support MySQL-style `'user'@'host'` grantee syntax?
     fn supports_user_host_grantee(&self) -> bool {
         false
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -694,7 +694,7 @@ impl Dialect for SnowflakeDialect {
         true
     }
 
-    fn supports_snowflake_pipe_operator(&self) -> bool {
+    fn supports_long_arrow_pipe_operator(&self) -> bool {
         true
     }
 
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -445,6 +445,8 @@ impl Dialect for SnowflakeDialect {
         // Snowflake supports the `:` cast operator unlike other dialects
         match &token.token {
             Token::Colon => Some(Ok(self.prec_value(Precedence::DoubleColon))),
+            // ->> is the Snowflake pipe operator (statement-level), not a binary expression operator
+            Token::LongArrow => Some(Ok(self.prec_unknown())),
             _ => None,
         }
     }
@@ -689,6 +691,10 @@ impl Dialect for SnowflakeDialect {
 
     /// See <https://docs.snowflake.com/en/user-guide/querying-semistructured#label-higher-order-functions>
     fn supports_lambda_functions(&self) -> bool {
+        true
+    }
+
+    fn supports_snowflake_pipe_operator(&self) -> bool {
         true
     }
 
@@ -1071,6 +1077,11 @@ pub fn parse_create_table(
                 break;
             }
             Token::SemiColon => {
+                parser.prev_token();
+                break;
+            }
+            Token::LongArrow => {
+                // Snowflake pipe operator terminates the statement
                 parser.prev_token();
                 break;
             }
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -1210,6 +1210,11 @@ pub fn parse_create_database(
                 _ => return parser.expected("end of statement", next_token),
             },
             Token::SemiColon | Token::EOF => break,
+            Token::LongArrow => {
+                // Snowflake pipe operator terminates the statement
+                parser.prev_token();
+                break;
+            }
             _ => return parser.expected("end of statement", next_token),
         }
     }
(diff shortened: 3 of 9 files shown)


src/dialect/mod.rs
```diff
@@ -726,6 +726,14 @@ pub trait Dialect: Debug + Any {
         false
     }
 
+    /// Does the dialect support the Snowflake `-->`  flow/pipe operator for chaining
```
LucaCappelletti94: This looks like a typo

tests/sqlparser_snowflake.rs
```diff
@@ -4912,3 +4912,61 @@ fn test_select_dollar_column_from_stage() {
     // With table function args, without alias
     snowflake().verified_stmt("SELECT $1, $2 FROM @mystage1(file_format => 'myformat')");
 }
+
+#[test]
+fn test_snowflake_pipe_operator() {
```
LucaCappelletti94, line 4917: Several cases that I am not sure are currently corrected supported, such as the following ones, should be added to the tests. These are ones from the Snowflake docs:

```
// Exact documented SHOW shape.
verified_stmt(
    r#"SHOW WAREHOUSES
       ->> SELECT "name", "state", "type", "size" FROM $1"#,
);

// A Snowflake dialect override that currently scans until EOF/semicolon.
verified_stmt("CREATE DATABASE d ->> SELECT 1");

// Existing happy path.
verified_stmt(
    "CREATE TABLE t (id INT)
     ->> INSERT INTO t VALUES (1)
     ->> SELECT * FROM $1"
);

// Error paths.
assert_parse_error("SELECT 1 ->>");
assert_parse_error("SELECT * FROM $0");

// GenericDialect ambiguity.
verified_generic_expr("SELECT payload ->> 'name'");
verified_generic_pipe("SELECT 1 ->> SELECT 2");
```

src/parser/mod.rs
```diff
@@ -16706,6 +16735,18 @@ impl<'a> Parser<'a> {
             // Stage reference: @mystage or @namespace.stage (e.g. Snowflake)
             self.parse_snowflake_stage_table_factor()
         } else {
+            // Handle Snowflake pipe result references ($1, $2, ...) in FROM clause.
+            // See <https://docs.snowflake.com/en/sql-reference/operators-flow>
+            if self.dialect.supports_snowflake_pipe_operator() {
+                if let Token::Placeholder(ref s) = self.peek_token_ref().token.clone() {
+                    if let Some(index_str) = s.strip_prefix('$') {
+                        if let Ok(index) = index_str.parse::<u64>() {
```
LucaCappelletti94: Here you are currently accepting `$0`, maybe replace with:

```
if let Ok(index @ 1..) = index_str.parse::<u64>() {
```

tests/sqlparser_snowflake.rs
```diff
@@ -4912,3 +4912,61 @@ fn test_select_dollar_column_from_stage() {
...
+
+    // Non-SELECT statements in the chain (CREATE/INSERT)
+    snowflake()
+        .verified_stmt("CREATE TABLE t (id INT) ->> INSERT INTO t VALUES (1) ->> SELECT * FROM $1");
+
+    // Pipe operator is not parsed in non-Snowflake dialects
+    use sqlparser::dialect::GenericDialect;
+    use sqlparser::parser::Parser;
+    // In a generic dialect, ->> is a binary operator, not a pipe
+    let stmts = Parser::parse_sql(&GenericDialect {}, "SELECT 1").unwrap();
```
LucaCappelletti94: What is this even meant to test? Also, `GenericDialect` would generally be a dialect that supports all syntaxes, and should therefore likely also support this one.

LucaCappelletti94 (changes requested):

src/dialect/mod.rs
BenSatori: Right :)

tests/sqlparser_snowflake.rs
BenSatori: Added these cases

src/parser/mod.rs
BenSatori: Changed

tests/sqlparser_snowflake.rs
BenSatori: Added that to GenericDialect, and also renamed it so it doesn't look awkward (in case other DBs adds support for this as well).

src/parser/mod.rs
```diff
@@ -16733,6 +16763,18 @@ impl<'a> Parser<'a> {
             // Stage reference: @mystage or @namespace.stage (e.g. Snowflake)
             self.parse_snowflake_stage_table_factor()
         } else {
+            // Handle Snowflake pipe result references ($1, $2, ...) in FROM clause.
+            // See <https://docs.snowflake.com/en/sql-reference/operators-flow>
+            if self.dialect.supports_long_arrow_pipe_operator() {
```
LucaCappelletti94, line 16760: I believe these could be folded? But I might be mistaken.

LucaCappelletti94 (changes requested):
Just a nit pick, then it should me mergiable imho

src/parser/mod.rs
BenSatori: Slightly folded :)

LucaCappelletti94 (approved):

BenSatori:
@iffyio do you think we can merge this?

LucaCappelletti94:
@BenSatori GitHub CI / queue is currently chocking, it'll take a while for it to clear.

src/dialect/generic.rs
```diff
@@ -321,6 +321,10 @@ impl Dialect for GenericDialect {
         true
     }
 
+    fn supports_long_arrow_pipe_operator(&self) -> bool {
+        true
+    }
+
```
LucaCappelletti94: You should drop the Generic override. `->>` stays a JSON operator in `GenericDialect`, so a pipe after any statement that ends in an expression fails.

```suggestion
```

tests/sqlparser_snowflake.rs
```diff
@@ -4913,6 +4913,87 @@ fn test_select_dollar_column_from_stage() {
...
+            "Expected: identifier, found: $0".to_string()
+        ))
+    );
+
+    // GenericDialect also supports pipe syntax (it is permissive by design)
+    use sqlparser::dialect::GenericDialect;
+    use sqlparser::parser::Parser;
+    Parser::parse_sql(&GenericDialect {}, "SELECT * FROM t ->> SELECT * FROM $1").unwrap();
+    // JSON ->> binary operator still works inside expressions
+    Parser::parse_sql(&GenericDialect {}, "SELECT payload ->> 'name'").unwrap();
```
LucaCappelletti94: You should drop these Generic assertions together with the override.

```suggestion
```

src/dialect/snowflake.rs
LucaCappelletti94: You should stop the `COPY INTO` option loop at `->>`. `COPY INTO @s FROM t ->> SELECT * FROM $1` fails with `Expected: another copy option, ; or EOF', found: SELECT`.

```suggestion
                Token::SemiColon | Token::EOF => break,
                Token::LongArrow => {
                    parser.prev_token();
                    break;
                }
```

LucaCappelletti94: Analogous to the other note, you should stop the session option loop at `->>` too. `ALTER SESSION SET QUERY_TAG = 'x' ->> SELECT 1` fails with `Expected: another option or end of statement, found: ->>`.

```suggestion
            Token::SemiColon | Token::EOF | Token::LongArrow => break,
```

src/parser/mod.rs
LucaCappelletti94: Same story as the other two notes, you should stop `parse_key_value_options` at `->>` as well. Its unparenthesized callers `CREATE WAREHOUSE`, `CREATE USER`, `ALTER USER ... SET` and `PUT` fail before a pipe, e.g. `CREATE WAREHOUSE w WAREHOUSE_SIZE = 'X-SMALL' ->> SHOW WAREHOUSES` gives `Expected: another option, EOF, SemiColon, Comma or ')', found: SHOW`.

```suggestion
                }
                Token::LongArrow if self.dialect.supports_long_arrow_pipe_operator() => {
                    self.prev_token();
                    break;
                }
```

tests/sqlparser_snowflake.rs
```diff
@@ -4913,6 +4913,87 @@ fn test_select_dollar_column_from_stage() {
...
+    // Non-SELECT statements in the chain (CREATE/INSERT)
+    snowflake()
+        .verified_stmt("CREATE TABLE t (id INT) ->> INSERT INTO t VALUES (1) ->> SELECT * FROM $1");
+
+    // Exact documented SHOW shape from Snowflake docs
+    snowflake()
+        .verified_stmt(r#"SHOW WAREHOUSES ->> SELECT "name", "state", "type", "size" FROM $1"#);
+
+    // CREATE DATABASE piped into SELECT
+    snowflake().verified_stmt("CREATE DATABASE d ->> SELECT 1");
```
LucaCappelletti94: Red tests for the described failure modes.

```suggestion
    snowflake().verified_stmt("CREATE DATABASE d ->> SELECT 1");

    // Statements whose options run to the end of the statement
    snowflake().verified_stmt("ALTER SESSION SET QUERY_TAG='x' ->> SELECT 1");
    snowflake().verified_stmt("COPY INTO @s FROM t ->> SELECT * FROM $1");
    snowflake().verified_stmt("CREATE WAREHOUSE w WAREHOUSE_SIZE='X-SMALL' ->> SHOW WAREHOUSES");
```

PR #2442, opened 2026-08-17
Visit CreateView name as a relation

sabir-akhadov-localstack:
`visit_relations` / `visit_relations_mut` did not visit the view name in `CREATE VIEW` statements:

```rust
let statements = Parser::parse_sql(&GenericDialect {}, "CREATE VIEW db1.v AS SELECT * FROM t")?;
// visit_relations yielded only `t`, not `db1.v`
```

`CreateTable`, `CreateIndex`, `Msck`, and `AlterTable` already annotate their object names with `visit(with = "visit_relation")` — `CreateView.name` was missing the annotation. This PR adds it, so tools that inspect or rewrite relation references (e.g. schema rewriters) see the view name too.

### DIFF
--- a/src/ast/ddl.rs
+++ b/src/ast/ddl.rs
@@ -4386,6 +4386,7 @@ pub struct CreateView {
     /// <https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax>
     pub secure: bool,
     /// View name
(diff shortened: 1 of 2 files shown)


LucaCappelletti94 (approved):

iffyio (approved):

PR #2447, opened 2026-08-21
Keep the quoting and span of a function argument name

LucaCappelletti94:
`CREATE FUNCTION f("Role" TEXT)` parsed the argument name into an `Ident` whose value still carried the quote characters, with `quote_style: None`. `"we""ird"` came back as `"we"ird"`.

Both argument readers rebuilt the name from the token's text with `Ident::new(token.to_string())`. They now use `word.into_ident(token.span)`, like `parse_identifier` does, which keeps the quote style and the span.


iffyio (approved):

PR #2450, opened 2026-08-22
DuckDB: Add SUMMARIZE query support

t8y2:
Adds DuckDB `SUMMARIZE` support for named tables and query inputs such as `SELECT`, `FROM`, `VALUES`, parenthesized queries, and CTEs.

The syntax is represented as a query AST node so it can also be used in subqueries, for example `SELECT column_name FROM (SUMMARIZE SELECT 42 AS answer)`.

Reference: https://duckdb.org/docs/stable/guides/meta/summarize.html


tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,21 @@ fn test_duckdb_lambda_function() {
...
+fn parse_summarize() {
+    duckdb().verified_stmt("SUMMARIZE users");
+    duckdb().verified_stmt(r#"SUMMARIZE "analytics"."events""#);
+    duckdb().verified_stmt("SUMMARIZE 'users'");
+    duckdb().verified_stmt("SUMMARIZE SELECT * FROM users");
+    duckdb().verified_stmt("SUMMARIZE FROM users");
+    duckdb().verified_stmt("SUMMARIZE VALUES (1.0), (6754950520)");
+    duckdb().verified_stmt("SUMMARIZE (SELECT 42 AS answer)");
+    duckdb().verified_stmt("SUMMARIZE WITH users AS (SELECT 1 AS id) SELECT * FROM users");
+    duckdb().verified_stmt("SELECT column_name FROM (SUMMARIZE SELECT 42 AS answer)");
```
LucaCappelletti94: You should cover `SUMMARIZE` as a scalar subquery. DuckDB 1.4.4 accepts `SELECT (SUMMARIZE users)`, but this branch fails with `Expected: ), found: users`.

```suggestion
    duckdb().verified_stmt("SELECT column_name FROM (SUMMARIZE SELECT 42 AS answer)");
    duckdb().verified_stmt("SELECT (SUMMARIZE users)");
```

src/parser/mod.rs
LucaCappelletti94: You should let `peek_sub_query` recognise `SUMMARIZE`. DuckDB 1.4.4 accepts `SELECT (SUMMARIZE users)`, but this branch fails with `Expected: ), found: users` because the expression parser only treats `SELECT` and `WITH` as subquery heads.

```suggestion
    self.peek_one_of_keywords(&[Keyword::SELECT, Keyword::WITH])
            .is_some()
            || (self.dialect.supports_summarize() && self.peek_keyword(Keyword::SUMMARIZE))
```

LucaCappelletti94 (changes requested):

PR #2458, opened 2026-08-25
Separate the LAMBDA keyword syntax from the -> lambda syntax

adriangb:
## Motivation

`supports_lambda_functions()` gates two different spellings of the same feature:

- the arrow form, `x -> x + 1`
- the `LAMBDA` keyword form, `LAMBDA x : x + 1`

Because they share one flag, a dialect cannot have one without the other. That is
a problem for any dialect where `->` already means something else.

### DuckDB

DuckDB uses `->` for **both** JSON member access (`j -> '$.family'`) and the arrow
lambda form, and resolves the ambiguity from the function signature at bind time —
which a parser cannot do. DuckDB's own answer was to add the `lambda` keyword and
deprecate the arrow:

> DuckDB v1.3 deprecated the old lambda single arrow syntax (`x -> x + 1`) in
> favor of the Python-style syntax (`lambda x : x + 1`). [...] DuckDB v2.0 will
> disable the single arrow syntax by default.

DuckDB exposes the choice as `SET lambda_syntax = 'ENABLE_SINGLE_ARROW' |
'DISABLE_SINGLE_ARROW'`. Today this crate can only express the first; the two flags
here are what it takes to express the second.

The consequence is visible now. `DuckDbDialect` sets `supports_lambda_functions()
== true`, so DuckDB's own documented JSON example misparses:

| SQL (DuckDB dialect) | parses as |
| --- | --- |
| `SELECT j -> 'field' FROM t` | `Expr::Lambda { params: [j], body: 'field' }` |
| `SELECT t.j -> 'field' FROM t` | `BinaryOp { op: Arrow }` |

A bare column becomes a lambda; a qualified column stays JSON access. Both round-trip
to identical SQL, which is why no test catches it.

This PR does not change `DuckDbDialect`. Flipping it to the keyword-only spelling
fixes the table above but introduces the mirror-image misparse on `x -> x > 1`, and
that tradeoff deserves its own discussion. This change makes the correct behaviour
*expressible*, which is the prerequisite either way.

### Custom dialects

The same conflict is unavoidable for anyone building a custom dialect — derived from
PostgreSQL or not — that wants JSON accessors and lambda functions at once. Today,
enabling lambdas silently reinterprets every `->`; there is no way to have both.
Overriding only `supports_lambda_keyword_syntax` gives them lambdas with `->` left
alone. Unlike DuckDB there is no tradeoff here, since the keyword form is unambiguous.

## Change

Adds `Dialect::supports_lambda_keyword_syntax()`, which gates only the `LAMBDA`
keyword form and defaults to `supports_lambda_functions()`.

No existing dialect changes behaviour. Dialects that support the arrow form keep both
spellings; dialects that support neither still get neither. A dialect that uses `->`
for something else can now override just the new method to get lambdas without
disturbing its operator.

## Tests

`tests/sqlparser_custom_dialect.rs`:

- a dialect enabling only `supports_lambda_keyword_syntax` parses `lambda x : x + 1`
  while `->` stays a `BinaryOperator::Arrow`
- a dialect enabling only `supports_lambda_functions` still accepts both spellings,
  pinning the defaulting behaviour
- a dialect enabling the arrow form but disabling the keyword form accepts only the
  arrow spelling

`tests/sqlparser_derive_dialect.rs`:

- a derived dialect overriding only `supports_lambda_keyword_syntax` parses a lambda
  whose body is a JSON access, `lambda x : (x -> 'a')::INT + 1`, asserting the outer
  expression is a `LambdaSyntax::LambdaKeyword` lambda and the inner `->` is still
  `BinaryOperator::Arrow`

`tests/sqlparser_duckdb.rs`:

- pins `x -> x > 1` as a `LambdaSyntax::Arrow` lambda. Both readings of `->` print
  identically, so the pre-existing round-trip check passes under either one — without
  this assertion, switching `DuckDbDialect` to the keyword-only spelling leaves the
  entire suite green.

Full suite, `cargo fmt --check`, and `cargo clippy --all-targets --all-features` all
pass.


src/dialect/mod.rs
```diff
@@ -535,10 +535,30 @@ pub trait Dialect: Debug + Any {
...
+    /// ```
+    ///
+    /// This spelling does not claim the `->` token, so it can be enabled by
+    /// dialects that already give `->` a different meaning, such as PostgreSQL
+    /// and its derivatives, where `->` is JSON member access. Defaults to
+    /// [`Self::supports_lambda_functions`], so dialects supporting the `->`
+    /// spelling accept the `LAMBDA` spelling too unless they say otherwise.
+    fn supports_lambda_keyword_syntax(&self) -> bool {
+        self.supports_lambda_functions()
+    }
```
adriangb, line 562: One could argue for adding `fn supports_lambda_arrow_syntax()` as well, but I'd hold off until there is a concrete use case for enabling arrow syntax but not lambda syntax.

tests/sqlparser_derive_dialect.rs
```diff
@@ -121,3 +123,45 @@ fn test_identifier_quote_style_overrides() {
         None
     );
 }
+
+#[test]
+fn test_lambda_keyword_syntax_on_postgres_derivative() {
```
adriangb, line 130: I'm split between this (the real regression test _I_ want) and another test using a `MyDialect` in `sqlparser_custom_dialect.rs` that enables the two flags. Open to input.

tests/sqlparser_derive_dialect.rs
```diff
@@ -121,3 +123,45 @@ fn test_identifier_quote_style_overrides() {
...
+    assert!(!dialect.supports_lambda_functions());
+
+    // The `LAMBDA` spelling parses.
+    let sql = "SELECT transform(xs, lambda x : x + 1)";
+    let ast = Parser::parse_sql(&dialect, sql).unwrap();
+    assert_eq!(sql, ast[0].to_string());
+
+    // `->` and `->>` still parse as JSON member access rather than
+    // introducing a lambda parameter.
+    for (sql, expected_op) in [
```
adriangb: Can these be in a single query, ideally something like `select transform(xs, lambda x: (x->'a')::int + 1)`?

tests/sqlparser_derive_dialect.rs
```diff
@@ -121,3 +125,65 @@ fn test_identifier_quote_style_overrides() {
...
+        panic!("expected the left operand to be a cast");
+    };
+    let Expr::Nested(json_access) = expr.as_ref() else {
+        panic!("expected the cast operand to be parenthesized");
+    };
+    let Expr::BinaryOp { op, .. } = json_access.as_ref() else {
+        panic!("expected `->` to stay a binary operator");
+    };
+    assert_eq!(&BinaryOperator::Arrow, op);
+}
```
LucaCappelletti94, line 189: I believe we should add a test for the Spark/Snowflake case (they only accept `->`):

```suggestion
}

#[test]
fn custom_dialect_lambda_arrow_syntax_without_keyword() {
    // Arrow lambdas stay on while the `LAMBDA` keyword spelling is off,
    // as in engines like Spark and Snowflake.
    #[derive(Debug)]
    struct MyDialect {}

    impl Dialect for MyDialect {
        fn is_identifier_start(&self, ch: char) -> bool {
            is_identifier_start(ch)
        }

        fn is_identifier_part(&self, ch: char) -> bool {
            is_identifier_part(ch)
        }

        fn supports_lambda_functions(&self) -> bool {
            true
        }

        fn supports_lambda_keyword_syntax(&self) -> bool {
            false
        }
    }

    let dialect = MyDialect {};

    let sql = "SELECT transform(xs, x -> x + 1)";
    assert_eq!(
        sql,
        &format!("{}", Parser::parse_sql(&dialect, sql).unwrap()[0])
    );
    assert!(Parser::parse_sql(&dialect, "SELECT transform(xs, lambda x : x + 1)").is_err());
}
```

src/dialect/mod.rs
```diff
@@ -535,10 +535,30 @@ pub trait Dialect: Debug + Any {
...
+    /// that uses `->` as a binary operator should override only the latter.
     fn supports_lambda_functions(&self) -> bool {
         false
     }
 
+    /// Returns true if the dialect supports the `LAMBDA` keyword spelling of
+    /// lambda functions, for example:
+    ///
+    /// ```sql
+    /// SELECT transform(array(1, 2, 3), LAMBDA x : x + 1); -- returns [2,3,4]
```
LucaCappelletti94: I believe `transform(array(...), LAMBDA ...)` is the wrong spelling, if you meant the DuckDB syntax it would be more like:

```suggestion
    /// SELECT list_transform([1, 2, 3], lambda x : x + 1); -- returns [2, 3, 4]
```

src/dialect/mod.rs
```diff
@@ -535,10 +535,30 @@ pub trait Dialect: Debug + Any {
...
+    /// ```
+    ///
+    /// This spelling does not claim the `->` token, so it can be enabled by
+    /// dialects that already give `->` a different meaning, such as PostgreSQL
+    /// and its derivatives, where `->` is JSON member access. Defaults to
+    /// [`Self::supports_lambda_functions`], so dialects supporting the `->`
+    /// spelling accept the `LAMBDA` spelling too unless they say otherwise.
+    fn supports_lambda_keyword_syntax(&self) -> bool {
```
LucaCappelletti94, line 560:
```suggestion
    ///
    /// See <https://duckdb.org/docs/stable/sql/functions/lambda>
    fn supports_lambda_keyword_syntax(&self) -> bool {
```

src/dialect/mod.rs
LucaCappelletti94: Arrow-only is already expressible by overriding `supports_lambda_keyword_syntax` to `false`, so I believe there is no need.

tests/sqlparser_derive_dialect.rs
LucaCappelletti94: Both seem worth keeping.

LucaCappelletti94 (changes requested):
Generally ok, just a missing Snowflake syntax test, and the missing documentation to lambda, and the wrong syntax in the example.

src/dialect/mod.rs
adriangb: Thanks, applied!

src/dialect/mod.rs
adriangb: Thanks, applied!

tests/sqlparser_derive_dialect.rs
adriangb: Thanks, added but in `sqlparser_custom_dialect.rs`

LucaCappelletti94 (approved):

src/dialect/mod.rs
```diff
@@ -535,10 +535,32 @@ pub trait Dialect: Debug + Any {
...
+    /// SELECT list_transform([1, 2, 3], lambda x : x + 1); -- returns [2, 3, 4]
+    /// ```
+    ///
+    /// This spelling does not claim the `->` token, so it can be enabled by
+    /// dialects that already give `->` a different meaning, such as PostgreSQL
+    /// and its derivatives, where `->` is JSON member access. Defaults to
+    /// [`Self::supports_lambda_functions`], so dialects supporting the `->`
+    /// spelling accept the `LAMBDA` spelling too unless they say otherwise.
+    ///
+    /// See <https://duckdb.org/docs/stable/sql/functions/lambda>
```
iffyio, line 559: Im not sure I understood the problem being solved for  - the description mentions pg as an example, suggesting there is some ambiguous grammar in play but pg doesnt have lambda syntax to my knowledge? 

is there an example syntax that issupported by a dialect and the parser doesnt cover?

iffyio (commented):
left a comment, its not clear to me what the PR is looking to solve for

src/dialect/mod.rs
adriangb: This is a fair point. I replied in https://github.com/apache/datafusion-sqlparser-rs/pull/2458#issuecomment-5508739108 but reflecting a bit more I think there's an even stronger framing. Your hesitation comes from PostgreSQL being a bad example. It has no lambda syntax, so the new flag
would be unreachable there and the motivation is then speculative. I've reframed the
description to focus on DuckDB and custom dialects.

For DuckDB this enables fixing a live bug rather than a hypothetical.
`DuckDbDialect` already sets `supports_lambda_functions() == true`, so DuckDB's own
documented JSON example misparses today:

```rust
// DuckDbDialect
"SELECT j -> 'field' FROM t"    // => Expr::Lambda { params: [j], body: 'field' }  ❌
"SELECT t.j -> 'field' FROM t"  // => BinaryOp { op: Arrow }                       ✅
```

A bare column becomes a lambda, a qualified one stays JSON access. Both print back as
the same SQL, which is why no round-trip test catches it.

DuckDB itself resolves `->` from the function signature at bind time, which a parser
can't do — and that's exactly why they deprecated the arrow form in v1.3 in favor of
`lambda x : x + 1`, with v2.0 disabling it by default and `SET lambda_syntax` to
choose in between. Following that requires the two spellings to be separable, which is
all this PR does.

I've deliberately not flipped `DuckDbDialect` here: it fixes the case above but breaks
`x -> x > 1` in the other direction, so it's worth its own PR. I did add an assertion
pinning the current arrow behavior because I found the whole suite stays green when
you flip it. This way if we do flip it we can verify the change in behavior against tests.

This PR also pplies to custom dialects wanting JSON accessors and lambdas at once, which
is what the `derive_dialect!` test now covers. This could be a postgres based dialect that wants to add support for lambda functions (our case, using `derive_dialect!`) or a completely custom dialect.

adriangb: @iffyio please let me know if this explanation clarified things or if it is still unclear

adriangb:
@LucaCappelletti94 is it okay with you if I merge this?

src/dialect/mod.rs
```diff
@@ -535,10 +535,37 @@ pub trait Dialect: Debug + Any {
...
+    /// ```sql
+    /// SELECT list_transform([1, 2, 3], lambda x : x + 1); -- returns [2, 3, 4]
+    /// ```
+    ///
+    /// This spelling does not claim the `->` token, so it can be enabled by
+    /// dialects that already give `->` a different meaning — for example JSON
+    /// member access. DuckDB uses `->` for both, resolving the ambiguity from
+    /// the function signature at bind time rather than while parsing, and
+    /// deprecated the arrow lambda form in v1.3 in favour of this one; v2.0
+    /// disables the arrow form by default.
```
LucaCappelletti94, line 558: Version numbers in a doc comment go stale without anyone noticing, and the DuckDB page you link already carries them. I think we can trim this down in this fashion.

```suggestion
    /// This spelling does not claim the `->` token, so it can be enabled by
    /// dialects that already give `->` a different meaning, such as JSON
    /// member access.
```

tests/sqlparser_duckdb.rs
```diff
@@ -902,6 +902,24 @@ fn test_duckdb_lambda_function() {
     let sql_arrow = "SELECT list_filter([1, 2, 3], x -> x > 1)";
     duckdb().verified_stmt(sql_arrow);
 
+    // `->` is ambiguous in DuckDB: it is both the arrow lambda spelling and
+    // JSON member access, and DuckDB resolves it from the function signature
+    // at bind time. `DuckDbDialect` currently resolves it to a lambda. Both
+    // readings print identically, so round-tripping cannot tell them apart —
+    // assert the shape so any future change to that choice is visible here.
```
LucaCappelletti94, line 909: The bind-time detail repeats the trait doc, I think we can deduplicate it in this manner:

```suggestion
    // Both readings of `->` print identically, so round-tripping cannot tell
    // a lambda from JSON member access. Assert the shape instead.
```

LucaCappelletti94 (changes requested):
I re-read the current head and all seems good.

I left two suggestions, both cosmetic.

On merging, @iffyio's question on `src/dialect/mod.rs:564` is still unanswered after your two pings. I would let that thread close first rather than merge past an open question.

adriangb:
Thanks @LucaCappelletti94, I applied your suggestions.

Any thoughts on how we can move forward with https://github.com/apache/datafusion-sqlparser-rs/pull/2458#discussion_r3911943201? Is there something unclear in my reply that I may be missing, or are we just waiting for @iffyio to get back to us? If the latter is there any way to ping @iffyio? I don't think it's great to block for several weeks just because we don't hear back.

LucaCappelletti94:
@adriangb within a few days tops (maybe even later today) I will be merging the updated fuzzer harness. When that gets merged, if this PR clears it I will merge it directly.

In the meantime I'll ping @alamb so as to cast a tie breaking vote and decide how to move forward.

iffyio (approved):
I think we can merge this, sorry for the delay getting back @adriangb

alamb:
🎉

PR #2463, opened 2026-08-29
Preserve numeric compound field access display

LucaCappelletti94:
`CASE ... END . 2` parsed successfully, but its printed form became `CASE ... END.2`. That cannot be parsed again because `.2` is tokenized as a number.

This keeps a space around numeric dotted field access, so the printed statement reparses to the same tree.


alamb (approved):
Looks good to me

PR #2464, opened 2026-08-29
Preserve single quote literal display

LucaCappelletti94:
A string value containing adjacent quotes printed with one level of escaping missing.

`''''''` parsed as two quote characters, then printed as `''''`, which reparsed as one quote character.

This makes single-quoted string values print with their quotes escaped every time, so parse and display reach a fixed point.

### DIFF
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -1679,7 +1679,7 @@ fn check_roundtrip_of_escaped_string() {
     TestedDialects::new_with_options(vec![Box::new(MySqlDialect {})], options.clone())
         .verified_stmt(r"SELECT 'I\'m fine'");
     TestedDialects::new_with_options(vec![Box::new(MySqlDialect {})], options.clone())
-        .verified_stmt(r#"SELECT 'I''m fine'"#);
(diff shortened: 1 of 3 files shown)


tests/sqlparser_mysql.rs
```diff
@@ -1679,7 +1679,7 @@ fn check_roundtrip_of_escaped_string() {
     TestedDialects::new_with_options(vec![Box::new(MySqlDialect {})], options.clone())
         .verified_stmt(r"SELECT 'I\'m fine'");
     TestedDialects::new_with_options(vec![Box::new(MySqlDialect {})], options.clone())
-        .verified_stmt(r#"SELECT 'I''m fine'"#);
+        .one_statement_parses_to(r#"SELECT 'I''m fine'"#, "");
```
alamb, line 1682: What does this change do and what would it mean for a user ?

alamb (commented):
Thanks @LucaCappelletti94

tests/sqlparser_mysql.rs
LucaCappelletti94: I think I forgot to push the second part of this PR or something extremely dumb like that as this does not make sense, will do so when I get home.

PR #2465, opened 2026-08-30
Apply recursion limit without std

LucaCappelletti94:
`Parser::with_recursion_limit` documented a hard parser depth limit, but builds without `std` used a stub `RecursionCounter` that always succeeded. Deeply nested input could overflow the stack instead of returning `ParserError::RecursionLimitExceeded`.

This makes the counter use `alloc::rc::Rc` and `core::cell::Cell`, which are available without `std`, and adds a `--no default-features` runtime test to the no-std CI path.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -81,6 +81,7 @@ jobs:
         uses: ./.github/actions/setup-builder
         with:
           targets: 'thumbv6m-none-eabi'
+      - run: cargo test --release --no-default-features --test no_std_recursion
       - run: cargo check --no-default-features --target thumbv6m-none-eabi
       - run: cargo check --no-default-features --features visitor --target thumbv6m-none-eabi
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -72,25 +72,48 @@ macro_rules! parser_err {
 mod alter;
 mod merge;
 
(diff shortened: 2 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -81,6 +81,7 @@ jobs:
         uses: ./.github/actions/setup-builder
         with:
           targets: 'thumbv6m-none-eabi'
+      - run: cargo test --release --no-default-features --test no_std_recursion
```
alamb, line 84: 👍

src/parser/mod.rs
```diff
@@ -72,46 +72,25 @@ macro_rules! parser_err {
...
-/// Implementation [`RecursionCounter`] if std is available
 mod recursion {
-    use std::cell::Cell;
-    use std::rc::Rc;
+    use alloc::rc::Rc;
+    use core::cell::Cell;
 
     use super::ParserError;
 
-    /// Tracks remaining recursion depth. This value is decremented on
```
alamb, line 83: why remove all these comments?

LucaCappelletti94: Just seemed borderline tautological and a lot of text for a simple bit of code

alamb (approved):
Thank you @LucaCappelletti94 

I took the liberty of merging up from main and touching up some comments

iffyio (approved):

PR #2467, opened 2026-08-31
Clickhouse: Clickhouse assume table constraint

### LINKED ISSUE #2466: Support Clickhouse ASSUME table constraints (and remove required parenthetical)
Hello, currently [Clickhouse's ASSUME table constraints](https://clickhouse.com/docs/reference/statements/create/table#constraints) generate parse errors:
```bash
$ cat assume_test.sql
CREATE TABLE my_table (
    `date` Date,
    CONSTRAINT my_constraint ASSUME (date > '2025-01-01')
)
ENGINE = MergeTree
;
```
----
```bash
$ cargo run --features json_example --example cli assume_test.sql --clickhouse
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.01s
     Running `target/debug/examples/cli assume_test.sql --clickhouse`
Parsing from file 'assume_test.sql' using ClickHouseDialect
2026-08-31T00:27:53.248Z DEBUG [sqlparser::parser] Parsing sql 'CREATE TABLE my_table (
    `date` Date,
    CONSTRAINT my_constraint ASSUME (date > '2025-01-01')
)
ENGINE = MergeTree
;
'...
Error during parsing: ParserError("Expected: PRIMARY, UNIQUE, FOREIGN, CHECK, or EXCLUDE, found: ASSUME at Line: 3, Column: 30")
```

In addition, in Clickhouse table constraint expressions aren't required to have top-level parenthesis. We should support these dialect attributes for Clickhouse.

pelovett:
Resolves: #2466 

First time contributing so please let me know if I missed anything!

### DIFF
--- a/src/ast/table_constraints.rs
+++ b/src/ast/table_constraints.rs
@@ -78,15 +78,11 @@ pub enum TableConstraint {
     ///   [ON UPDATE <referential_action>] [ON DELETE <referential_action>]
     /// }`).
     ForeignKey(ForeignKeyConstraint),
-<<<<<<< Updated upstream
     /// `[ CONSTRAINT <name> ] CHECK (<expr>) [NO INHERIT] [[NOT] ENFORCED]`
-=======
-    /// `[ CONSTRAINT <name> ] CHECK (<expr>) [[NOT] ENFORCED]`
     ///
     /// The parentheses are only optional to parse when
     /// [`supports_unparenthesized_check_constraint`](crate::dialect::Dialect::supports_unparenthesized_check_constraint)
     /// is true for the dialect (e.g. ClickHouse); the constraint always displays with parentheses.
->>>>>>> Stashed changes
     Check(CheckConstraint),
     /// ClickHouse [table constraint][1]: `[ CONSTRAINT <name> ] ASSUME (<expr>)`.
     ///
--- a/src/ast/table_constraints.rs
+++ b/src/ast/table_constraints.rs
@@ -84,7 +84,8 @@ pub enum TableConstraint {
     /// [`supports_unparenthesized_check_constraint`](crate::dialect::Dialect::supports_unparenthesized_check_constraint)
     /// is true for the dialect (e.g. ClickHouse); the constraint always displays with parentheses.
     Check(CheckConstraint),
-    /// ClickHouse [table constraint][1]: `[ CONSTRAINT <name> ] ASSUME (<expr>)`.
+    /// ClickHouse [table constraint][1]: `CONSTRAINT <name> ASSUME (<expr>)`.
+    /// Unlike the other constraints here, the name is mandatory.
     ///
     /// The parentheses are optional to parse; the constraint always displays with parentheses.
     ///
@@ -247,11 +248,11 @@ impl crate::ast::Spanned for CheckConstraint {
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
-/// An `ASSUME` constraint (`[ CONSTRAINT <name> ] ASSUME <expr>`).
+/// An `ASSUME` constraint (`CONSTRAINT <name> ASSUME <expr>`).
 pub struct AssumeConstraint {
     /// Optional constraint name.
-    pub name: Option<Ident>,
-    /// The boolean expression the ASSUME constraint enforces.
+    pub name: Ident,
+    /// The boolean expression the ASSUME constraint claims is true.
     pub expr: Box<Expr>,
 }
 
@@ -261,7 +262,7 @@ impl fmt::Display for AssumeConstraint {
         write!(
             f,
             "{}ASSUME ({})",
-            display_constraint_name(&self.name),
+            display_constraint_name(&Some(self.name.clone())),
             self.expr
         )?;
         Ok(())
@@ -270,9 +271,7 @@ impl fmt::Display for AssumeConstraint {
 
 impl crate::ast::Spanned for AssumeConstraint {
     fn span(&self) -> Span {
-        self.expr
-            .span()
-            .union_opt(&self.name.as_ref().map(|i| i.span))
+        self.expr.span().union_opt(&Some(self.name.span))
     }
 }
 
--- a/src/ast/table_constraints.rs
+++ b/src/ast/table_constraints.rs
@@ -78,8 +78,22 @@ pub enum TableConstraint {
     ///   [ON UPDATE <referential_action>] [ON DELETE <referential_action>]
     /// }`).
     ForeignKey(ForeignKeyConstraint),
+<<<<<<< Updated upstream
     /// `[ CONSTRAINT <name> ] CHECK (<expr>) [NO INHERIT] [[NOT] ENFORCED]`
+=======
+    /// `[ CONSTRAINT <name> ] CHECK (<expr>) [[NOT] ENFORCED]`
+    ///
+    /// The parentheses are only optional to parse when
+    /// [`supports_unparenthesized_check_constraint`](crate::dialect::Dialect::supports_unparenthesized_check_constraint)
+    /// is true for the dialect (e.g. ClickHouse); the constraint always displays with parentheses.
+>>>>>>> Stashed changes
     Check(CheckConstraint),
+    /// ClickHouse [table constraint][1]: `[ CONSTRAINT <name> ] ASSUME (<expr>)`.
+    ///
+    /// The parentheses are optional to parse; the constraint always displays with parentheses.
+    ///
+    /// [1]: https://clickhouse.com/docs/reference/statements/create/table#constraints
+    Assume(AssumeConstraint),
     /// MySQLs [index definition][1] for index creation. Not present on ANSI so, for now, the usage
     /// is restricted to MySQL, as no other dialects that support this syntax were found.
     ///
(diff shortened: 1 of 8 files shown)


tests/sqlparser_clickhouse.rs
```diff
@@ -233,6 +233,26 @@ fn parse_create_table() {
...
+    );
+    clickhouse().one_statement_parses_to(
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME "a" > 0) ENGINE = MergeTree"#,
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" > 0)) ENGINE = MergeTree"#,
+    );
+    clickhouse().verified_stmt(
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" > 0)) ENGINE = MergeTree"#,
+    );
+}
+
```
LucaCappelletti94, line 255: These changes currently accept bare `ASSUME a > 0`, and `AssumeConstraint::name` is `Option<Ident>`, but ClickHouse does not have that form.

You may use these **RED** tests as a guide:

```suggestion

#[test]
fn parse_create_table_rejects_unnamed_assume_constraint() {
    clickhouse()
        .parse_sql_statements(
            r#"CREATE TABLE "x" ("a" "int", ASSUME "a" > 0) ENGINE = MergeTree"#,
        )
        .expect_err("ASSUME constraints require CONSTRAINT and a name");
}

#[test]
fn parse_alter_table_rejects_unnamed_assume_constraint() {
    clickhouse()
        .parse_sql_statements(r#"ALTER TABLE "x" ADD ASSUME "a" > 0"#)
        .expect_err("ASSUME constraints require CONSTRAINT and a name");
}
```

src/ast/table_constraints.rs
```diff
@@ -227,6 +244,38 @@ impl crate::ast::Spanned for CheckConstraint {
...
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+/// An `ASSUME` constraint (`[ CONSTRAINT <name> ] ASSUME <expr>`).
+pub struct AssumeConstraint {
+    /// Optional constraint name.
+    pub name: Option<Ident>,
+    /// The boolean expression the ASSUME constraint enforces.
```
LucaCappelletti94: ClickHouse's `ASSUME` is an optimizer assertion that is not enforced and may produce incorrect results when false. Not a constraint, when false it may start doing 2+2=3. It is just an optimization hint as far as I can recall, not an enforced rule.

LucaCappelletti94 (changes requested):

src/ast/table_constraints.rs
pelovett: Yes, updated with a more accurate comment

pelovett:
@LucaCappelletti94 if you have some time, could you take another look? I tried to address the issues you identified.

LucaCappelletti94:
Sure, currently I am trying to ultimate the fuzzing CI, once that is done reviews might get more swift given it should provide a stronger claim of consistency

src/parser/mod.rs
```diff
@@ -10203,6 +10210,32 @@ impl<'a> Parser<'a> {
...
+            {
+                let Some(identifier) = name else {
+                    return self.expected(
+                        "CONSTRAINT <name> before ASSUME",
+                        TokenWithSpan {
+                            token: Token::make_keyword("ASSUME"),
+                            span: next_token.span,
+                        },
+                    );
+                };
```
LucaCappelletti94: You should fall back to column parsing here instead of erroring. `ASSUME` is not reserved in ClickHouse, and this arm now rejects `CREATE TABLE x (assume Int32) ENGINE = MergeTree`, which parses on `main` and in ClickHouse 26.9. The unnamed-`ASSUME` tests still fail as they should.

```suggestion
                let Some(identifier) = name else {
                    self.prev_token();
                    return Ok(None);
                };
```

src/parser/mod.rs
```diff
@@ -10203,6 +10210,32 @@ impl<'a> Parser<'a> {
...
+                let has_paren = self.consume_token(&Token::LParen);
+                let expr = Box::new(self.parse_expr()?);
+                if has_paren {
+                    self.expect_token(&Token::RParen)?;
+                }
```
LucaCappelletti94: You should parse the whole expression and strip one `Nested` layer. A leading `(` is consumed as the wrapper today, so `CONSTRAINT c ASSUME (a + 1) > 0` and `ASSUME (a) IN (1, 2)` fail with `Expected: ',' or ')'`. ClickHouse accepts both.

```suggestion
                let expr = match self.parse_expr()? {
                    Expr::Nested(expr) => expr,
                    expr => Box::new(expr),
                };
```

src/parser/mod.rs
```diff
@@ -10180,9 +10180,16 @@ impl<'a> Parser<'a> {
                 ))
             }
             Token::Word(w) if w.keyword == Keyword::CHECK => {
-                self.expect_token(&Token::LParen)?;
+                let has_paren = if self.dialect.supports_unparenthesized_check_constraint() {
```
LucaCappelletti94: The unparenthesized `CHECK` path has the same defect. `CONSTRAINT c CHECK (a + 1) > 0` fails here but ClickHouse accepts it. Add an appropriate test for it, and correct this parser.

src/ast/table_constraints.rs
```diff
@@ -227,6 +245,36 @@ impl crate::ast::Spanned for CheckConstraint {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
+/// An `ASSUME` constraint (`CONSTRAINT <name> ASSUME <expr>`).
+pub struct AssumeConstraint {
+    /// Optional constraint name.
```
LucaCappelletti94: The name is mandatory, so this doc comment is wrong.

```suggestion
    /// Constraint name.
```

src/ast/table_constraints.rs
```diff
@@ -227,6 +245,36 @@ impl crate::ast::Spanned for CheckConstraint {
...
+        use crate::ast::ddl::display_constraint_name;
+        write!(
+            f,
+            "{}ASSUME ({})",
+            display_constraint_name(&Some(self.name.clone())),
+            self.expr
+        )?;
+        Ok(())
+    }
+}
```
LucaCappelletti94: You could write the name directly. `display_constraint_name(&Some(self.name.clone()))` allocates a clone on every render.

```suggestion
impl fmt::Display for AssumeConstraint {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "CONSTRAINT {} ASSUME ({})", self.name, self.expr)
    }
}
```

tests/sqlparser_clickhouse.rs
```diff
@@ -233,6 +233,49 @@ fn parse_create_table() {
...
+    clickhouse().verified_stmt(
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" CHECK ("a" > 0)) ENGINE = MergeTree"#,
+    );
+    clickhouse().one_statement_parses_to(
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME "a" > 0) ENGINE = MergeTree"#,
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" > 0)) ENGINE = MergeTree"#,
+    );
+    clickhouse().verified_stmt(
+        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" > 0)) ENGINE = MergeTree"#,
+    );
```
LucaCappelletti94: You should cover expressions that start with `(`. Both assertions fail on the current head.

```suggestion
    clickhouse().verified_stmt(
        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" > 0)) ENGINE = MergeTree"#,
    );
    clickhouse().verified_stmt(
        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" CHECK (("a" + 1) > 0)) ENGINE = MergeTree"#,
    );
    clickhouse().one_statement_parses_to(
        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME ("a" + 1) > 0) ENGINE = MergeTree"#,
        r#"CREATE TABLE "x" ("a" "int", CONSTRAINT "y" ASSUME (("a" + 1) > 0)) ENGINE = MergeTree"#,
    );
```

tests/sqlparser_clickhouse.rs
```diff
@@ -233,6 +233,49 @@ fn parse_create_table() {
...
+        .expect_err("ASSUME constraints require CONSTRAINT");
+    clickhouse()
+        .parse_sql_statements(
+            r#"CREATE TABLE "x" ("a" "int", CONSTRAINT ASSUME "a" > 0) ENGINE = MergeTree"#,
+        )
+        .expect_err("ASSUME constraints require name");
+    clickhouse()
+        .parse_sql_statements(r#"CREATE TABLE "x" ("a" "int", ASSUME "a" > 0) ENGINE = MergeTree"#)
+        .expect_err("ASSUME constraints require CONSTRAINT and a name");
+}
```
LucaCappelletti94: This test fails on the current head and passes on `main`.

```suggestion
}

#[test]
fn parse_create_table_column_named_assume() {
    clickhouse().verified_stmt("CREATE TABLE x (assume Int32) ENGINE = MergeTree");
}
```

LucaCappelletti94 (changes requested):

PR #2468, opened 2026-08-31
Parse PostgreSQL MERGE DO NOTHING actions

LucaCappelletti94:
`MERGE ... WHEN MATCHED THEN DO NOTHING` currently fails because merge action parsing only recognizes `UPDATE`, `DELETE`, and `INSERT`.

`DO NOTHING` has no payload, so it is represented as `MergeAction::DoNothing` while retaining both keyword spans.

### DIFF
--- a/src/parser/merge.rs
+++ b/src/parser/merge.rs
@@ -107,7 +107,16 @@ impl Parser<'_> {
                 Keyword::UPDATE,
                 Keyword::INSERT,
                 Keyword::DELETE,
+                Keyword::DO,
             ]) {
+                Some(Keyword::DO) => {
+                    let do_token = self.get_current_token().clone();
+                    let nothing_token = self.expect_keyword(Keyword::NOTHING)?;
+                    MergeAction::DoNothing {
+                        do_token: do_token.into(),
+                        nothing_token: nothing_token.into(),
+                    }
+                }
                 Some(Keyword::UPDATE) => {
                     if matches!(
                         clause_kind,
@@ -212,7 +221,7 @@ impl Parser<'_> {
                 }
                 _ => {
                     return parser_err!(
-                        "expected UPDATE, DELETE or INSERT in merge clause",
+                        "expected UPDATE, DELETE, INSERT or DO NOTHING in merge clause",
                         self.peek_token_ref().span.start
                     );
                 }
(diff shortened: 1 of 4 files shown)


tests/sqlparser_postgres.rs
```diff
@@ -9852,3 +9852,39 @@ fn parse_alter_table_constraint_check_no_inherit() {
...
+                ..
+            }
+        ]
+    ));
+}
+
+#[test]
+fn reject_merge_do_without_nothing() {
+    assert_eq!(
+        pg().parse_sql_statements(
```
sabir-akhadov-localstack: The change isn't dialect gated so probably needs to be tested with `pg_and_generic()`?

src/parser/merge.rs
```diff
@@ -107,7 +107,17 @@ impl Parser<'_> {
                 Keyword::UPDATE,
                 Keyword::INSERT,
                 Keyword::DELETE,
+                Keyword::DO,
             ]) {
+                Some(Keyword::DO) => {
+                    let do_token = self.get_current_token().clone();
+                    self.expect_keyword_is(Keyword::NOTHING)?;
```
sabir-akhadov-localstack: NIT: Could shorten to `let nothing_token = self.expect_keyword(Keyword::NOTHING)?;`

tests/sqlparser_postgres.rs
LucaCappelletti94: RIght, done

src/parser/merge.rs
LucaCappelletti94: And done

sabir-akhadov-localstack (approved):
LGTM

tests/sqlparser_postgres.rs
```diff
@@ -9852,3 +9852,39 @@ fn parse_alter_table_constraint_check_no_inherit() {
...
+                clause_kind: MergeClauseKind::NotMatched,
+                action: MergeAction::DoNothing { .. },
+                ..
+            }
+        ]
+    ));
+}
+
+#[test]
+fn reject_merge_do_without_nothing() {
```
```suggestion
```
can we merge in the negative test?

iffyio (commented):
Left a minor comment, otherwise this looks good to me!

iffyio (approved):

PR #2469, opened 2026-08-31
SQLite: parse the full PRAGMA value grammar

LucaCappelletti94:
`PRAGMA` values such as `journal_mode = WAL`, `synchronous = NORMAL`, `cache_size = -2000` and `optimize = 0x10002` were rejected. The earlier boolean handling also kept its own keyword list plus a value-type allowlist, and that duplicated allowlist is what rejected `PRAGMA case_sensitive_like = true` in the first place.

`parse_pragma_value` now follows SQLite's own definition of a pragma value, `signed-number | name | signed-literal`, and stores the result as an `Expr`. A bare name like `WAL` is kept as an identifier, so the parser no longer decides which value a given pragma accepts and leaves that to execution the way SQLite does. `parse_pragma` builds the statement in a single place.

This changes `Statement::Pragma.value` from `Option<ValueWithSpan>` to `Option<Expr>`.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -20166,44 +20166,37 @@ impl<'a> Parser<'a> {
         })
     }
 
-    fn parse_pragma_value(&mut self) -> Result<ValueWithSpan, ParserError> {
-        let v = self.parse_value()?;
-        match &v.value {
-            Value::SingleQuotedString(_) => Ok(v),
-            Value::DoubleQuotedString(_) => Ok(v),
-            Value::Number(_, _) => Ok(v),
-            Value::Placeholder(_) => Ok(v),
-            _ => {
-                self.prev_token();
-                self.expected_ref("number or string or ? placeholder", self.peek_token_ref())
-            }
+    /// Parse a SQLite `pragma-value`: `signed-number | name | signed-literal`.
+    fn parse_pragma_value(&mut self) -> Result<Expr, ParserError> {
+        if matches!(self.peek_token_ref().token, Token::Plus | Token::Minus) {
+            let op = match self.next_token().token {
+                Token::Plus => UnaryOperator::Plus,
+                _ => UnaryOperator::Minus,
+            };
+            return Ok(Expr::UnaryOp {
+                op,
+                expr: Box::new(Expr::Value(self.parse_value()?)),
+            });
+        }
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -20167,15 +20167,34 @@ impl<'a> Parser<'a> {
...
             Value::Number(_, _) => Ok(v),
+            Value::Boolean(_) => Ok(v),
             Value::Placeholder(_) => Ok(v),
             _ => {
                 self.prev_token();
-                self.expected_ref("number or string or ? placeholder", self.peek_token_ref())
+                self.expected_ref(
+                    "boolean, number, string, or ? placeholder",
+                    self.peek_token_ref(),
+                )
```
iffyio: I think we could simplify the logic to have it accept whatever parse_value returns. and keep the sqlite semantics for validation upstream? that would have avoided this bug in the initial version

LucaCappelletti94: I checked SQLite docs hoping for a limited set of allowed values, it states there is explicitly no such set, and I therefore switched to `Expr`

src/parser/mod.rs
```diff
@@ -20157,44 +20157,37 @@ impl<'a> Parser<'a> {
...
+            };
+            return Ok(Expr::UnaryOp {
+                op,
+                expr: Box::new(Expr::Value(self.parse_value()?)),
+            });
+        }
+        match self.maybe_parse(|parser| parser.parse_value())? {
+            Some(value) => Ok(Expr::Value(value)),
+            None => Ok(Expr::Identifier(self.parse_identifier()?)),
         }
```
iffyio, line 20175: hmm can this be replaced by a call to `parse_expr()`?

LucaCappelletti94: In terms of accepting things, yes it would work, but `parse_expr()` would over-extend the grammar accepting arbitrary expressions like `PRAGMA p = 1 + 2`, `foo(bar)`, `a AND b` or `(SELECT 1)`.

SQLite defines a pragma value as `signed-number | name | signed-literal`, so `parse_pragma_value` mirrors that production exactly while still leaving the per-pragma value validation to execution.

I understand I tend to lean much more strongly on the strict parser approach, but here I really believe allowing any expr would be excessively loose.

PR #2471, opened 2026-08-31
Honour operator precedence in MySQL and Spark `DIV`

### LINKED ISSUE #2460: `DIV` right operand is parsed at precedence 0 in the MySQL and Spark dialects
`DIV` is announced at `Precedence::MulDivModOp` by `get_next_precedence`, but the MySQL and Spark
dialects parse its right operand with `parse_expr()` (= `parse_subexpr(0)`), so the operand absorbs
every following operator instead of stopping at `DIV`'s own precedence.

| SQL | Parsed as | MySQL / Spark |
|---|---|---|
| `7 DIV 2 + 1` | `7 DIV (2 + 1)` = 2 | `(7 DIV 2) + 1` = 4 |
| `9 DIV 3 * 3` | `9 DIV (3 * 3)` = 1 | `(9 DIV 3) * 3` = 9 |
| `a DIV 2 = 1` | `a DIV (2 = 1)` | `(a DIV 2) = 1` |

Both systems place `DIV` with `*` and `/`: MySQL's
[operator precedence table](https://dev.mysql.com/doc/refman/8.4/en/operator-precedence.html) lists
`*, /, DIV, %, MOD` on one row, and Spark's `SqlBaseParser.g4` has
`operator=(ASTERISK | SLASH | PERCENT | DIV)` in a single left-recursive rule.

`Display` for `Expr::BinaryOp` emits no parentheses, so a mis-grouped tree round-trips back to the
original SQL — `verified_expr` / `verified_stmt` cannot catch this, which is why it has gone
unnoticed. Only a test that asserts on the tree will.

### Fix

Thread the `precedence` argument through, exactly as `SqliteDialect::parse_infix` already does for
`REGEXP` / `MATCH` / `GLOB` (#2419):

```diff
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
-        _precedence: u8,
+        precedence: u8,
-            let right = Box::new(match parser.parse_expr() {
+            let right = Box::new(match parser.parse_subexpr(precedence) {

--- a/src/dialect/spark.rs
+++ b/src/dialect/spark.rs
-        _precedence: u8,
+        precedence: u8,
-            let right = Box::new(match parser.parse_expr() {
+            let right = Box::new(match parser.parse_subexpr(precedence) {
```

### Red test

```rust
#[test]
fn parse_div_precedence() {
    // `DIV` has the same precedence as `*` and `/`, so `+` must end up at the root.
    assert_eq!(
        Expr::BinaryOp {
            left: Box::new(Expr::BinaryOp {
                left: Box::new(Expr::value(number("7"))),
                right: Box::new(Expr::value(number("2"))),
            }),
            op: BinaryOperator::Plus,
            right: Box::new(Expr::value(number("1"))),
        },
        mysql().verified_expr("7 DIV 2 + 1")
    );
}
```

Credit to @LucaCappelletti94, who spotted this and wrote both the patch and the test while reviewing
#2436. I'll open a PR once #2436 lands, to keep the two precedence changes reviewable separately —
happy for someone else to pick it up sooner.

zvonimir-dd:
Fixes #2460.

`Dialect::get_next_precedence` announces `DIV` at `Precedence::MulDivModOp`, but `MySqlDialect::parse_infix`
and `SparkSqlDialect::parse_infix` both ignored their `precedence` argument and parsed the right
operand with `parse_expr()` (= `parse_subexpr(0)`), so the operand absorbed every following operator:

| SQL | Before | After / MySQL & Spark |
|---|---|---|
| `7 DIV 2 + 1` | `7 DIV (2 + 1)` = 2 | `(7 DIV 2) + 1` = 4 |
| `9 DIV 3 * 3` | `9 DIV (3 * 3)` = 1 | `(9 DIV 3) * 3` = 9 |
| `a DIV 2 = 1` | `a DIV (2 = 1)` | `(a DIV 2) = 1` |

Both engines place `DIV` with `*` and `/`: MySQL's
[operator precedence table](https://dev.mysql.com/doc/refman/8.4/en/operator-precedence.html) lists
`*, /, DIV, %, MOD` on one row, and Spark's `SqlBaseParser.g4` has
`operator=(ASTERISK | SLASH | PERCENT | DIV)` in a single left-recursive rule.

The fix threads the caller's precedence through to `parse_subexpr`, exactly as
`SqliteDialect::parse_infix` already does for `REGEXP` / `MATCH` / `GLOB` (#2419). Since
`parse_subexpr` passes the *upcoming* operator's precedence into the dialect hook, the right operand
is parsed at `MulDivModOp`, which stops it before `+` and makes `DIV` left-associative against `*`,
`/` and itself.

`Display` for `Expr::BinaryOp` emits no parentheses, so a mis-grouped tree round-trips back to the
original SQL — `verified_stmt` / `verified_expr` alone cannot catch this. The new tests therefore
assert on the tree: `tests/sqlparser_mysql.rs::parse_div_precedence` covers `+`, `*`, `DIV` against
itself, `=`, and explicit parentheses; `tests/sqlparser_spark.rs::test_div_precedence` mirrors the
first two for Spark. I confirmed both fail without the source change.

`GenericDialect` is deliberately not included — it has no `DIV` operator support, so it never
produces `MyIntegerDivide`.

Full suite green (1605 tests, up 2 from 1603), `cargo fmt --check` and
`cargo clippy --all-targets --all-features -- -D warnings` clean.

Thanks to @LucaCappelletti94, who spotted this while reviewing #2436 and wrote both the patch and the
headline test.

### DIFF
--- a/src/dialect/mysql.rs
+++ b/src/dialect/mysql.rs
@@ -98,12 +98,12 @@ impl Dialect for MySqlDialect {
         &self,
         parser: &mut crate::parser::Parser,
         expr: &crate::ast::Expr,
-        _precedence: u8,
+        precedence: u8,
     ) -> Option<Result<crate::ast::Expr, ParserError>> {
         // Parse DIV as an operator
         if parser.parse_keyword(Keyword::DIV) {
             let left = Box::new(expr.clone());
-            let right = Box::new(match parser.parse_expr() {
+            let right = Box::new(match parser.parse_subexpr(precedence) {
                 Ok(expr) => expr,
                 Err(e) => return Some(Err(e)),
             });
--- a/src/dialect/spark.rs
+++ b/src/dialect/spark.rs
@@ -137,11 +137,11 @@ impl Dialect for SparkSqlDialect {
         &self,
         parser: &mut Parser,
         expr: &Expr,
-        _precedence: u8,
+        precedence: u8,
     ) -> Option<Result<Expr, ParserError>> {
         if parser.parse_keyword(Keyword::DIV) {
             let left = Box::new(expr.clone());
-            let right = Box::new(match parser.parse_expr() {
+            let right = Box::new(match parser.parse_subexpr(precedence) {
                 Ok(expr) => expr,
                 Err(e) => return Some(Err(e)),
             });
--- a/tests/sqlparser_mysql.rs
+++ b/tests/sqlparser_mysql.rs
@@ -3760,6 +3760,62 @@ fn parse_div_infix_propagates_parse_error() {
 }
 
 #[test]
+fn parse_div_precedence() {
+    let div = |left: Expr, right: Expr| Expr::BinaryOp {
+        left: Box::new(left),
+        op: BinaryOperator::MyIntegerDivide,
+        right: Box::new(right),
+    };
+    let num = |n: &str| Expr::value(number(n));
+
+    // `DIV` shares the precedence of `*` and `/`, so `+` must end up at the root.
+    assert_eq!(
+        Expr::BinaryOp {
+            left: Box::new(div(num("7"), num("2"))),
+            op: BinaryOperator::Plus,
+            right: Box::new(num("1")),
+        },
+        mysql().verified_expr("7 DIV 2 + 1")
+    );
+
+    // Equal precedence resolves left-associatively, both against `*` and against itself.
+    assert_eq!(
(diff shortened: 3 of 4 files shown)


LucaCappelletti94 (approved):

alamb (approved):
Thank you @zvonimir-dd and @LucaCappelletti94

PR #2476, opened 2026-09-01
Parse expression-valued LIKE ESCAPE clauses

lovasoa:
Repro:
```sql
SELECT 'a%' LIKE 'a#%' ESCAPE ('' || '#')
```

SQLite accepts this, but sqlparser currently rejects the ESCAPE operand as a non-value.

Parse and store ESCAPE as an expression so supported expression-valued clauses are representable.


iffyio (approved):
LGTM! Thanks @lovasoa!

PR #2477, opened 2026-09-01
Visit function clauses in source order

lovasoa:
Repro:
```sql
SELECT LISTAGG(value) WITHIN GROUP (ORDER BY order_key) FILTER (WHERE filter_key)
```

The visitor currently sees `filter_key` before `order_key`, despite the source and formatter ordering.

Declare `within_group` before `filter` so derived visitors traverse function clauses in source order.


alamb (approved):
Looks fine to me -- thanks @lovasoa

PR #2478, opened 2026-09-02
Databricks: enable query and nested type syntax

EthanVieira:
Enable existing parser capabilities for the Databricks dialect:
- MAP<key, value> nested types
- Backslash-escaped string literals
- SELECT * REPLACE
- FROM-first queries
- Pipe syntax (|>)

The pipeline negative test now uses CALL 123 invalid because Databricks permits numeric-leading identifiers, making the previous 123invalid input valid. The replacement remains invalid for every dialect covered by the test.

References:
- [Databricks pipeline syntax](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-pipeline)
- [SELECT syntax](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-select)
- [MAP type](https://docs.databricks.com/aws/en/sql/language-manual/data-types/map-type)

### DIFF
--- a/src/dialect/databricks.rs
+++ b/src/dialect/databricks.rs
@@ -113,4 +113,24 @@ impl Dialect for DatabricksDialect {
     fn supports_select_item_multi_column_alias(&self) -> bool {
         true
     }
+
+    fn supports_map_literal_with_angle_brackets(&self) -> bool {
+        true
+    }
+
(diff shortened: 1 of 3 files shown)


iffyio (approved):
Thanks @EthanVieira!

PR #2480, opened 2026-09-03
Redshift: Support APPROXIMATE PERCENTILE_DISC

BenSatori:
Add Redshift support for APPROXIMATE PERCENTILE_DISC with WITHIN GROUP.

Previously, queries such as:

SELECT APPROXIMATE PERCENTILE_DISC(0.5)
WITHIN GROUP (ORDER BY totalprice)

failed to parse. Added a regression test based on the documented Redshift syntax.

### DIFF
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -599,6 +599,7 @@ pub trait Dialect: Debug + Any {
     }
 
     /// Does the dialect support the `APPROXIMATE PERCENTILE_DISC` function syntax?
+    /// See <https://docs.aws.amazon.com/redshift/latest/dg/r_APPROXIMATE_PERCENTILE_DISC.html>
     fn supports_approximate_percentile_disc(&self) -> bool {
         false
     }
--- a/src/dialect/mod.rs
+++ b/src/dialect/mod.rs
@@ -598,6 +598,11 @@ pub trait Dialect: Debug + Any {
         None
     }
 
+    /// Does the dialect support the `APPROXIMATE PERCENTILE_DISC` function syntax?
+    fn supports_approximate_percentile_disc(&self) -> bool {
+        false
+    }
+
     /// Does the dialect support trailing commas around the query?
     fn supports_trailing_commas(&self) -> bool {
         false
--- a/src/dialect/redshift.rs
+++ b/src/dialect/redshift.rs
@@ -15,6 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
+#[cfg(not(feature = "std"))]
+use alloc::boxed::Box;
+
 use crate::ast::Expr;
 use crate::dialect::Dialect;
 use crate::keywords::Keyword;
--- a/src/dialect/redshift.rs
+++ b/src/dialect/redshift.rs
@@ -15,8 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::ast::Expr;
 use crate::dialect::Dialect;
 use crate::keywords::Keyword;
+use crate::parser::{Parser, ParserError};
+use crate::tokenizer::Token;
 use core::iter::Peekable;
 use core::str::Chars;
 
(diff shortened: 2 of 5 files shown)


src/dialect/redshift.rs
```diff
@@ -35,6 +41,28 @@ pub struct RedshiftSqlDialect {}
 // in the Postgres dialect, the query will be parsed as an array, while in the Redshift dialect it will
 // be a json path
 impl Dialect for RedshiftSqlDialect {
+    fn parse_prefix(&self, parser: &mut Parser) -> Option<Result<Expr, ParserError>> {
+        if matches!(&parser.peek_token_ref().token, Token::Word(word) if word.value.eq_ignore_ascii_case("approximate"))
```
iffyio: hmm this doesnt look like the correct place for this functionality, we already have function parsers in the parser and support other percentile_disc variants. so that I think we likely just need to adjust that logic to accept redshifts' syntax

iffyio (changes requested):

tests/sqlparser_redshift.rs
```diff
@@ -557,3 +557,17 @@ fn parse_unpivot_expression() {
...
+fn parse_approximate_percentile_disc() {
+    redshift().one_statement_parses_to(
+        r#"SELECT TOP 10 date.caldate,
+COUNT(totalprice), SUM(totalprice),
+APPROXIMATE PERCENTILE_DISC(0.5) WITHIN GROUP (ORDER BY totalprice)
+FROM listing
+JOIN date ON listing.dateid = date.dateid
+GROUP BY date.caldate
+ORDER BY 3 DESC"#,
+        "SELECT TOP 10 date.caldate, COUNT(totalprice), SUM(totalprice), APPROXIMATE PERCENTILE_DISC(0.5) WITHIN GROUP (ORDER BY totalprice) FROM listing JOIN date ON listing.dateid = date.dateid GROUP BY date.caldate ORDER BY 3 DESC",
```
iffyio: lets simplify this test input so that its clear what's being tested. also we should use verified_stmt

BenSatori: Sure thing. Will do. My intuition was to use the example in Redshift's documentation, but I will change.

tests/sqlparser_redshift.rs
```diff
@@ -557,3 +557,10 @@ fn parse_unpivot_expression() {
 fn test_interval_as_column_name() {
     redshift().verified_stmt("SELECT * FROM table_name WHERE interval = 78");
 }
+
+#[test]
+fn parse_approximate_percentile_disc() {
+    redshift().verified_stmt(
```
iffyio: let's use the `all_dialect_where` method. Also  let's add a test case for `SELECT APPROXIMATE PERCENTILE_DISC` which should be parsed for all dialects as a column alias I imagine?

src/parser/mod.rs
```diff
@@ -1836,6 +1836,20 @@ impl<'a> Parser<'a> {
 
         let dialect = self.dialect;
 
+        if dialect.supports_approximate_percentile_disc()
+            && matches!(&self.peek_token_ref().token, Token::Word(word) if word.value.eq_ignore_ascii_case("approximate"))
+            && matches!(&self.peek_nth_token_ref(1).token, Token::Word(word) if word.value.eq_ignore_ascii_case("percentile_disc"))
```
iffyio: impl wise these should use keywords.

src/parser/mod.rs
```diff
@@ -1836,6 +1836,20 @@ impl<'a> Parser<'a> {
 
         let dialect = self.dialect;
 
+        if dialect.supports_approximate_percentile_disc()
```
iffyio: hmm I think a better place to hook into this logic is probably [here](https://github.com/BenSatori/datafusion-sqlparser-rs/blob/8c2ef6cd95a1a594f92bf23c5fbdfb8cf2fab25d/src/parser/mod.rs#L1656)?

also maybe the check there can be to `peek (precentile_disc)` given the current work is 'approximate' and if so maybe_parse(expr)

tests/sqlparser_redshift.rs
```diff
@@ -557,3 +557,22 @@ fn parse_unpivot_expression() {
...
+#[test]
+fn parse_approximate_percentile_disc() {
+    let dialects = all_dialects_where(|d| d.supports_approximate_percentile_disc());
+    dialects.verified_stmt(
+        "SELECT APPROXIMATE PERCENTILE_DISC(0.5) WITHIN GROUP (ORDER BY totalprice)",
+    );
+}
+
+#[test]
+fn parse_approximate_percentile_disc_as_column_alias() {
```
iffyio:
```suggestion

```
we can merge the tests

src/dialect/mod.rs
```diff
@@ -598,6 +598,11 @@ pub trait Dialect: Debug + Any {
         None
     }
 
+    /// Does the dialect support the `APPROXIMATE PERCENTILE_DISC` function syntax?
+    fn supports_approximate_percentile_disc(&self) -> bool {
```
iffyio, line 603: can we add a link to the docs where the syntax is described? both here and in the redshift override? Otherwise it would be hard to know what it does without reading the code

iffyio (commented):
left a couple minor comments, otherwise this looks good overall

iffyio (approved):
LGTM! Thanks @BenSatori!

PR #2482, opened 2026-09-03
Check in Cargo.lock

alamb:
## Which issue does this PR close?

N/A

## Rationale for this change

Checking in `Cargo.lock` gives CI reproducible builds and makes dependency updates visible in review, per the [current Cargo guidance](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control) which now recommends committing the lockfile for libraries as well.

## What changes are included in this PR?

- Remove `Cargo.lock` from `.gitignore`
- Check in `Cargo.lock` generated with `cargo update` (all 46 packages locked to the latest semver-compatible versions; no `Cargo.toml` changes needed)

## Are these changes tested?

`cargo test --all-features` passes (22/22 suites, 0 failures).

## Are there any user-facing changes?

No. This only affects local development and CI of this repository; downstream users of the published crate are unaffected.


Cargo.lock
```diff
@@ -0,0 +1,412 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 4
```
LucaCappelletti94, line 3: At this time https://github.com/apache/datafusion-sqlparser-rs/blob/main/.tool-versions still specifies Rust `1.75`, and `version = 4` is only supported from `1.78`. I suspect that the current CI misses this because it tests only stable, beta, and nightly, and not specifically `1.75`.

I am not sure whether here we would want to either bump `.tool-versions` or downgrade the `version` of the `Cargo.lock`.

LucaCappelletti94 (changes requested):
To make dependency updates reliably visible in review, maybe we should use `--locked` in the CI? The `sqlparser_bench` job uses an independent manifest, so it likely needs its own committed lockfile, or possibly better a workspace integration.

LucaCappelletti94:
@alamb I pushed the follow-ups from my review to your branch. `sqlparser_bench` is now a workspace member, so one lockfile covers everything except `fuzz/`, which keeps its own workspace and ignores its lockfile as before. Every cargo command in CI now runs with `--locked`, so a dependency change that is not committed to `Cargo.lock` fails the build, and the lockfile diff is where reviewers see it.

`.tool-versions` is bumped to 1.88.0 and kept the lockfile at version 4. Making `sqlparser_derive` a workspace member also brings it under clippy, which flagged one `sort_by` that I changed to `sort_by_key`.

One consequence worth knowing is that the daily `cargo audit` now checks the committed lockfile, so a new advisory will fail it until someone runs `cargo update`.

LMK whether all looks good to you.

.tool-versions
```diff
@@ -1 +1 @@
-rust 1.75.0
\ No newline at end of file
+rust 1.88.0
```
alamb: Does this change the "msrv" of the crate (the minimum supported rust version) or just the versions that run the tests?

alamb (commented):
Thanks @LucaCappelletti94  -- other than the change to tool version this makes sense to me

.tool-versions
LucaCappelletti94: The `.tool-versions` bump leaves the crate's MSRV untouched. The crate declares no `rust-version` in `Cargo.toml` (maybe we should), so cargo enforces no MSRV for downstream users, and downstream builds resolve their own dependency graph without reading our `Cargo.lock`. CI does not read `.tool-versions` either.

The file only selects the toolchain that `asdf` and `mise` install for local development in this checkout.

alamb: > The file only selects the toolchain that asdf and mise install for local development in this checkou

I thought it changes the version of rust used for clippy, etc

So that by changing this we will implicitly start testing with a newer version of rust 🤔

LucaCappelletti94: No, those tools go by the version specified in CI, which currently is:

https://github.com/apache/datafusion-sqlparser-rs/blob/0fd4fa892c91e2f5ee6ffc0bcf614219a921b5e3/.github/workflows/rust.yml#L95

alamb: ok, I think i was confused with rust-toolchain.toml (e.g. https://github.com/apache/arrow-rs/blob/main/rust-toolchain.toml). Sorry about that

alamb:
Looks ready to me @LucaCappelletti94  -- thank you

PR #2490, opened 2026-09-10
Postgres: Fix parsing of COLLATE after :: cast

BenSatori:
Fixes parsing of queries like:
```sql
SELECT contact_name::text COLLATE "POSIX", * FROM customers;
```
which previously failed with `No infix parser for token Word(Word { value: "COLLATE", ... })`.

Per [Postgres operator precedence](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE), `::` binds tighter than `COLLATE`, so `PostgreSqlDialect::get_next_precedence` already assigns `COLLATE` a (lower) precedence relative to `::`. However, `Parser::parse_infix` had no case for the `COLLATE` keyword, so once the precedence-driven loop reached it, parsing failed. This adds the missing `Keyword::COLLATE` arm to build an `Expr::Collate` wrapping the already-parsed (e.g. cast) expression.

Added a unit test (`parse_collate_after_cast`) verifying `contact_name::TEXT COLLATE "POSIX"` parses to `Expr::Collate { expr: Expr::Cast { .. }, .. }`.

### DIFF
--- a/tests/sqlparser_postgres.rs
+++ b/tests/sqlparser_postgres.rs
@@ -519,6 +519,19 @@ fn parse_cast_in_default_expr() {
     pg().verified_stmt("CREATE TABLE t (c TEXT DEFAULT (foo())::TEXT NOT NULL)");
 }
 
+/// `::` binds tighter than `COLLATE`, so `expr::type COLLATE collation` collates the cast result.
+/// See <https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE>
+#[test]
+fn parse_collate_after_cast() {
+    match pg().verified_expr(r#"contact_name::TEXT COLLATE "POSIX""#) {
+        Expr::Collate { expr, collation } => {
+            assert!(matches!(*expr, Expr::Cast { .. }));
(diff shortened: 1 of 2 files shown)


LucaCappelletti94 (approved):
LGTM

tests/sqlparser_postgres.rs
```diff
@@ -519,6 +519,19 @@ fn parse_cast_in_default_expr() {
     pg().verified_stmt("CREATE TABLE t (c TEXT DEFAULT (foo())::TEXT NOT NULL)");
 }
 
+/// `::` binds tighter than `COLLATE`, so `expr::type COLLATE collation` collates the cast result.
+/// See <https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE>
+#[test]
+fn parse_collate_after_cast() {
+    match pg().verified_expr(r#"contact_name::TEXT COLLATE "POSIX""#) {
```
LucaCappelletti94: Only note maybe you want to write `pg_and_generic`

yoavcloud (approved):

PR #2491, opened 2026-09-10
Postgres: allow reserved keywords as bare column alias

BenSatori:
PostgreSQL allows almost any keyword (SELECT, ANALYZE, LATERAL, AND, OR, COLLATE, ...) to be used as a bare (AS-less) column alias unless it is in a small set of keywords that require a leading AS.

See https://www.postgresql.org/docs/current/sql-keywords-appendix.html

Previously:
`SELECT (SELECT c FROM tbl_name LIMIT 1) select;`

(or analyze/lateral/and/or/collate)
failed to parse.

This PR fixes that


src/dialect/postgresql.rs
```diff
@@ -54,6 +54,13 @@ const RESERVED_EXCLUSIONS_FOR_TABLE_ALIAS: &[Keyword] = &[
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct PostgreSqlDialect {}
 
+/// Keywords that PostgreSQL additionally allows (on top of
+/// [keywords::RESERVED_FOR_COLUMN_ALIAS]) to be used as a bare (`AS`-less)
+/// column alias.
+/// See <https://www.postgresql.org/docs/current/sql-keywords-appendix.html>
+const ADDITIONALLY_ALLOWED_BARE_COLUMN_ALIASES: &[Keyword] =
```
LucaCappelletti94, line 61: I believe at least these other ones are missing, I am not sure which of these others should be included: https://github.com/postgres/postgres/blob/master/src/include/parser/kwlist.h

```
&[
    Keyword::ANALYZE,
    Keyword::CLUSTER,
    Keyword::END,
    Keyword::EXCLUDE,
    Keyword::EXPLAIN,
    Keyword::LATERAL,
    Keyword::SELECT,
    Keyword::VALUES,
    Keyword::VIEW,
]
```

src/parser/mod.rs
```diff
@@ -1444,6 +1452,22 @@ impl<'a> Parser<'a> {
                 break;
             }
 
+            // `AND`/`OR`/`COLLATE` with no right-hand expression following it is not a
+            // binary operator or collation cast; it's a bare (`AS`-less) column alias,
+            // e.g. Postgres' `SELECT 1 and` or `SELECT 1 collate`.
+            if let Token::Word(w) = &self.peek_token_ref().token {
+                let kw = w.keyword;
+                if matches!(kw, Keyword::AND | Keyword::OR | Keyword::COLLATE)
```
LucaCappelletti94, line 1460: `PostgreSQL` also permits these bare aliases before the following query clauses. For example, `SELECT 1` and `FROM t` and `SELECT 1 collate FROM t` are valid, but `FROM` is not among these four lookahead tokens, so the parser still treats the alias as an operator and fails.

Please handle clause boundaries such as `FROM`, `INTO`, `WHERE`, and `ORDER BY`, and add tests with aliases followed by a clause. I may have missed some others, I may not recall all of them.

src/parser/mod.rs
```diff
@@ -1444,6 +1452,22 @@ impl<'a> Parser<'a> {
                 break;
             }
```
LucaCappelletti94, line 1453: This shared-parser change also enables these aliases for unrelated dialects. I am not sure about how to best handle excessive permissiveness.

For example, `BigQueryDialect` considers `AND` a column alias under its current predicate, so `SELECT 1 AND` now I believe would succeed instead of reporting the missing right-hand operand.

`GoogleSQL` explicitly requires reserved keywords such as `AND`, `OR`, and `COLLATE` to be quoted when used as identifiers.

Let's try to find a clean way to handle these cases and add negative non-PostgreSQL tests.

LucaCappelletti94 (changes requested):

src/dialect/postgresql.rs
```diff
@@ -54,6 +54,25 @@ const RESERVED_EXCLUSIONS_FOR_TABLE_ALIAS: &[Keyword] = &[
...
+    Keyword::LATERAL,
+    Keyword::AND,
+    Keyword::OR,
+    Keyword::COLLATE,
+    Keyword::CLUSTER,
+    Keyword::END,
+    Keyword::EXCLUDE,
+    Keyword::EXPLAIN,
+    Keyword::VALUES,
+    Keyword::VIEW,
```
LucaCappelletti94: You should also list the keywords PostgreSQL does not know at all. `next_token_starts_an_expr` reads `SORT`, `TOP`, `MINUS` and `DISTRIBUTE` as clause keywords, so `SELECT a AND sort FROM t WHERE b OR top` now fails with `Expected: end of statement, found: sort`. It parses on `main` and runs on PostgreSQL 17.

```suggestion
    Keyword::VIEW,
    // Not PostgreSQL keywords at all.
    Keyword::DISTRIBUTE,
    Keyword::MINUS,
    Keyword::SORT,
    Keyword::TOP,
```

tests/sqlparser_postgres.rs
```diff
@@ -9967,6 +9967,30 @@ fn parse_insert_by_name_keywords_as_table_and_alias() {
...
+        pg().one_statement_parses_to(
+            &format!("SELECT a {kw} FROM tbl_name"),
+            &format!("SELECT a AS {kw} FROM tbl_name"),
+        );
+    }
+
+    // `AND`/`OR`/`COLLATE` are still parsed as operators when followed by an operand.
+    pg().verified_stmt("SELECT 1 AND 2");
+    pg().verified_stmt("SELECT 1 OR 2");
+    pg().verified_stmt(r#"SELECT 1 COLLATE "de_DE" FROM tbl_name"#);
```
LucaCappelletti94: This line fails on the current head.

```suggestion
    pg().verified_stmt(r#"SELECT 1 COLLATE "de_DE" FROM tbl_name"#);
    pg().verified_stmt("SELECT a AND sort FROM tbl_name WHERE b OR top");
```

tests/sqlparser_common.rs
LucaCappelletti94: You should add the negative non-PostgreSQL test. On the current head it fails because `SnowflakeDialect` parses `SELECT 1 AND` as `SELECT 1 AS AND`.

```suggestion
}

#[test]
fn parse_and_or_not_bare_column_alias() {
    let dialects = all_dialects_but_pg();
    for sql in ["SELECT 1 AND", "SELECT 1 OR", "SELECT 1 AND, 2"] {
        assert!(dialects.parse_sql_statements(sql).is_err(), "{sql}");
    }
}
```

LucaCappelletti94 (changes requested):
Snowflake's `is_column_alias` returns `true` for any keyword not in its own list, so the new loop break accepts `SELECT 1 AND` and `SELECT 1 OR`. Snowflake reserves both ([reserved keywords](https://docs.snowflake.com/en/sql-reference/reserved-keywords)).

Do extend the keyword list, and add the necessary tests for them.

```diff
-            Keyword::FROM
+            Keyword::AND
+            | Keyword::FROM
             | Keyword::GROUP
             | Keyword::HAVING
             | Keyword::INTERSECT
             | Keyword::INTO
             | Keyword::MINUS
+            | Keyword::OR
             | Keyword::ORDER
```

PR #2493, opened 2026-09-10
Postgres: Support TRIM FROM syntax

BenSatori:
PostgreSQL documents `trim([ LEADING | TRAILING | BOTH ] [ characters text ] FROM string text)` and its non-standard comma form with optional `FROM`: https://www.postgresql.org/docs/current/functions-string.html

This change updates `parse_trim_expr` to handle `TRIM(FROM string)` for dialects that support Postgres-style comma-separated TRIM syntax, and preserves an explicit trim side for comma-form calls such as `TRIM(BOTH FROM string, characters)`.

AI assistance was used to prepare this change.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3101,6 +3101,21 @@ impl<'a> Parser<'a> {
                 trim_where = Some(self.parse_trim_where()?);
             }
         }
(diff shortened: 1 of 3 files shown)


LucaCappelletti94 (approved):

LucaCappelletti94:
LGTM, merging @alamb

PR #2498, opened 2026-09-11
DuckDB dialect support for escape string literals

corasaurus-hex:
# What is this?

DuckDB dialect support for escape string literals.

```sql
SELECT E'hello\nworld';
SELECT E'It\'s working';
SELECT E'column1\tcolumn2';
SELECT E'C:\\temp\\file';
```


tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,39 @@ fn test_duckdb_lambda_function() {
...
+    );
+    assert_eq!(
+        duckdb().verified_expr(r#"'a\nb'"#),
+        Expr::Value((Value::SingleQuotedString(r#"a\nb"#.to_string())).with_empty_span())
+    );
+    duckdb().one_statement_parses_to(
+        r#"COPY data TO E'out\n.csv'"#,
+        r#"COPY data TO 'out
+.csv'"#,
+    );
```
LucaCappelletti94: You should run the positive cases through `all_dialects_where(|d| d.supports_string_escape_constant())` instead of `duckdb()` alone. Both parser guards now key on that flag, so Redshift's `SELECT E'a\nb'` and `COPY data TO E'out\n.csv'` go from a parse error to parsing, and nothing tests it. With this change the test fails if either guard stops covering Redshift, which the current test does not catch.

```suggestion
    let escaping = all_dialects_where(|d| d.supports_string_escape_constant());
    escaping.verified_stmt(r#"SELECT E'a\nb' AS value"#);

    assert_eq!(
        escaping.verified_expr(r#"E'a\nb'"#),
        Expr::Value((Value::EscapedStringLiteral("a\nb".to_string())).with_empty_span())
    );
    assert_eq!(
        duckdb().verified_expr(r#"'a\nb'"#),
        Expr::Value((Value::SingleQuotedString(r#"a\nb"#.to_string())).with_empty_span())
    );
    escaping.one_statement_parses_to(
        r#"COPY data TO E'out\n.csv'"#,
        r#"COPY data TO 'out
.csv'"#,
    );
```

LucaCappelletti94 (changes requested):

PR #2499, opened 2026-09-12
DuckDB dialect: support % on LIMIT, like `LIMIT 5% OFFSET 20`

corasaurus-hex:
# What is this?

Adds support for the DuckDB dialect's percentage limit.

```sql
SELECT * FROM foo LIMIT 5% OFFSET 20;
```

Added very early on to DuckDB, pre-1.0 (https://github.com/duckdb/duckdb/releases/tag/v0.3.2).


src/parser/mod.rs
```diff
@@ -13636,6 +13666,57 @@ impl<'a> Parser<'a> {
...
+        }
+
+        if matches!(
+            &limit,
+            Expr::Value(value) if matches!(&value.value, Value::Number(_, true))
+        ) {
+            return self.expected_ref("an expression", self.peek_token_ref());
+        }
+
+        Ok(ParsedLimit::Percent(limit))
```
LucaCappelletti94: You should drop the `Value::Number(_, true)` check. It rejects `LIMIT 25L%` while `LIMIT 25L` and `LIMIT (25L)%` both parse, and without it `LIMIT 25L%` round-trips.

```suggestion
        Ok(ParsedLimit::Percent(limit))
```

tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,333 @@ fn test_duckdb_lambda_function() {
...
+            ParserError::ParserError("Expected: end of statement, found: %".to_string()),
+        ),
+        (
+            "SELECT 1 LIMIT 25%%",
+            ParserError::ParserError("Expected: an expression, found: %".to_string()),
+        ),
+        (
+            "SELECT 1 LIMIT 25L%",
+            ParserError::ParserError("Expected: an expression, found: EOF".to_string()),
+        ),
```
LucaCappelletti94: You should remove this case along with the check above, since it fails once the check is gone.

```suggestion
```

tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,333 @@ fn test_duckdb_lambda_function() {
...
+        (
+            "SELECT 1 LIMIT 25 PERCENT",
+            ParserError::ParserError("Expected: end of statement, found: PERCENT".to_string()),
+        ),
```
LucaCappelletti94: You may want to remove also this case. DuckDB 1.4.4 accepts `SELECT 1 LIMIT 25 PERCENT` (and `LIMIT 2.5 PERCENT`, `LIMIT 5 PERCENT OFFSET 1`), so the test asserts that valid DuckDB SQL is rejected.

```suggestion
```

LucaCappelletti94 (changes requested):

PR #2501, opened 2026-09-12
DuckDB: Support SET VARIABLE

corasaurus-hex:
# What is this?

Adds support for DuckDB's `SET VARIABLE` statement.

```sql
SET VARIABLE threshold = 42;
SET VARIABLE "my variable" TO 1 + 2 * 3;
```

Introduced in [DuckDB 1.1.0](https://duckdb.org/2024/09/09/announcing-duckdb-110#friendly-sql).


tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,314 @@ fn test_duckdb_lambda_function() {
     let sql_transform = "SELECT list_transform([1, 2, 3], lambda x : x * 2)";
     duckdb().verified_stmt(sql_transform);
 }
+
+#[test]
+fn parse_duckdb_set_variable() {
+    assert!(DuckDbDialect {}.supports_set_variable_statement());
```
LucaCappelletti94: You should drop this assertion. It restates the one-line `DuckDbDialect` override, and the parse assertions below already fail if the flag is off.

```suggestion
```

tests/sqlparser_duckdb.rs
```diff
@@ -19,12 +19,12 @@
...
 
 use helpers::attached_token::AttachedToken;
-use sqlparser::tokenizer::Span;
+use sqlparser::tokenizer::{Location, Span};
 use test_utils::*;
 
 use sqlparser::ast::*;
-use sqlparser::dialect::{DuckDbDialect, GenericDialect};
-use sqlparser::parser::ParserError;
+use sqlparser::dialect::{Dialect, DuckDbDialect, GenericDialect};
```
LucaCappelletti94: With line 931 gone, `Dialect` is unused and `clippy -D warnings` fails on it.

```suggestion
use sqlparser::dialect::{DuckDbDialect, GenericDialect};
```

tests/sqlparser_duckdb.rs
```diff
@@ -925,3 +925,314 @@ fn test_duckdb_lambda_function() {
...
+                assert_eq!(
+                    tested.parse_sql_statements(sql).unwrap_err(),
+                    ParserError::ParserError(
+                        "Expected: equals sign or TO, found: threshold".to_string()
+                    ),
+                    "{dialect_name}: {sql}"
+                );
+            }
+        }
+    }
```
LucaCappelletti94: You could leave out the `supports_set_stmt_without_operator` dialects instead of pinning MsSql's `SetSessionParamGeneric` output. 

That output is MsSql behaviour that this PR does not touch, so an unrelated MsSql `SET` change would break a DuckDB test.

The remaining assertion still fails if any other dialect starts accepting `SET VARIABLE`.

```suggestion
    let unsupported = all_dialects_where(|d| {
        !d.supports_set_variable_statement() && !d.supports_set_stmt_without_operator()
    });
    for sql in [
        "SET VARIABLE threshold = 42",
        "SET VARIABLE /* name */ threshold = 42",
    ] {
        assert_eq!(
            unsupported.parse_sql_statements(sql).unwrap_err(),
            ParserError::ParserError("Expected: equals sign or TO, found: threshold".to_string()),
            "{sql}"
        );
    }
```

LucaCappelletti94 (changes requested):

PR #2502, opened 2026-09-12
Snowflake: parse ORDER BY ALL

osipovartem:
`ORDER BY ALL` is supported by Snowflake and is already represented by the shared `OrderByKind::All` AST/parser path. This PR enables that existing capability for `SnowflakeDialect`.

The dialect-specific regression test covers a computed select expression together with `DESC NULLS FIRST`; the shared `parse_select_order_by_all` test also begins exercising Snowflake for every supported option combination.

Reference: https://docs.snowflake.com/en/sql-reference/constructs/order-by

Validation:
- `cargo +1.95.0 fmt --all -- --check`
- `cargo +1.95.0 test --test sqlparser_snowflake parse_order_by_all`
- `cargo +1.95.0 test --test sqlparser_common parse_select_order_by_all`
- `cargo +1.95.0 clippy --all-targets -- -D warnings`

### DIFF
--- a/tests/sqlparser_snowflake.rs
+++ b/tests/sqlparser_snowflake.rs
@@ -45,6 +45,19 @@ fn test_snowflake_create_table() {
 }
 
 #[test]
+fn parse_order_by_all() {
+    let query = snowflake()
+        .verified_query("SELECT value + 1 AS computed FROM source ORDER BY ALL DESC NULLS FIRST");
+    assert_eq!(
+        query.order_by.expect("ORDER BY expected").kind,
+        OrderByKind::All(OrderByOptions {
+            sort: Some(OrderBySort::Desc),
+            nulls_first: Some(true),
+        })
+    );
+}
+
+#[test]
 fn parse_sf_create_secure_view_and_materialized_view() {
     for sql in [
         "CREATE SECURE VIEW v AS SELECT 1",
--- a/src/dialect/snowflake.rs
+++ b/src/dialect/snowflake.rs
@@ -153,6 +153,11 @@ impl Dialect for SnowflakeDialect {
         true
     }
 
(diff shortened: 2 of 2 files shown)


osipovartem:
Independent read-only contribution review completed: no blocking correctness, regression, performance, API-compatibility, or test-coverage issues found. APPROVED.

tests/sqlparser_snowflake.rs
```diff
@@ -44,6 +44,19 @@ fn test_snowflake_create_table() {
...
+        .verified_query("SELECT value + 1 AS computed FROM source ORDER BY ALL DESC NULLS FIRST");
+    assert_eq!(
+        query.order_by.expect("ORDER BY expected").kind,
+        OrderByKind::All(OrderByOptions {
+            sort: Some(OrderBySort::Desc),
+            nulls_first: Some(true),
+        })
+    );
+}
+
```
LucaCappelletti94: You could drop this test. `parse_select_order_by_all` in `tests/sqlparser_common.rs` runs every dialect where `supports_order_by_all()` holds, so Snowflake now gets the same `ORDER BY ALL DESC NULLS FIRST` case plus the other eight option combinations.

```suggestion
```

LucaCappelletti94 (changes requested):

tests/sqlparser_snowflake.rs
osipovartem: Agreed. I removed the duplicate Snowflake-specific test in 712cb919. The shared `parse_select_order_by_all` test covers Snowflake and all nine option combinations; it passes, as does the full Snowflake test suite (167/167).

LucaCappelletti94 (approved):

LucaCappelletti94:
LGTM!
cc @alamb @iffyio

PR #2508, opened 2026-09-14
Postgres: Support IS DOCUMENT predicate

BenSatori:
Adds parsing and round-tripping for PostgreSQL XML `IS DOCUMENT` and `IS NOT DOCUMENT` predicates.

PostgreSQL documents these under XML predicates: https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-XML-PREDICATES

The parser handles `DOCUMENT` contextually after `IS`, gated by the existing XML expression dialect capability, so it does not become a global keyword.

AI assistance was used in preparing this change.


src/parser/mod.rs
```diff
@@ -4091,6 +4091,15 @@ impl<'a> Parser<'a> {
...
                         Ok(Expr::IsNotUnknown(Box::new(expr)))
+                    } else if let Some(negated) = if dialect.supports_xml_expressions() {
+                        self.parse_is_document_predicate()
+                    } else {
+                        None
+                    } {
+                        Ok(Expr::IsDocument {
+                            expr: Box::new(expr),
+                            negated,
+                        })
```
LucaCappelletti94: You should match `DOCUMENT` as a keyword with `parse_keyword`/`parse_keywords`, like the neighbouring `UNKNOWN` and `JSON` arms. Ofc add the `keywords.rs` entry.

```suggestion
                    } else if dialect.supports_xml_expressions()
                        && self.parse_keyword(Keyword::DOCUMENT)
                    {
                        Ok(Expr::IsDocument {
                            expr: Box::new(expr),
                            negated: false,
                        })
                    } else if dialect.supports_xml_expressions()
                        && self.parse_keywords(&[Keyword::NOT, Keyword::DOCUMENT])
                    {
                        Ok(Expr::IsDocument {
                            expr: Box::new(expr),
                            negated: true,
                        })
```

src/parser/mod.rs
```diff
@@ -4109,7 +4118,7 @@ impl<'a> Parser<'a> {
                         Ok(is_normalized)
                     } else {
                         self.expected_ref(
-                            "[NOT] NULL | TRUE | FALSE | DISTINCT | [NOT] JSON [VALUE | SCALAR | ARRAY | OBJECT] [WITH | WITHOUT UNIQUE [KEYS]] | [form] NORMALIZED FROM after IS",
+                            "[NOT] NULL | TRUE | FALSE | DISTINCT | [NOT] DOCUMENT | [NOT] JSON [VALUE | SCALAR | ARRAY | OBJECT] [WITH | WITHOUT UNIQUE [KEYS]] | [form] NORMALIZED FROM after IS",
```
LucaCappelletti94: You should keep `[NOT] DOCUMENT` out of this message, which every dialect shares.

```suggestion
                            "[NOT] NULL | TRUE | FALSE | DISTINCT | [NOT] JSON [VALUE | SCALAR | ARRAY | OBJECT] [WITH | WITHOUT UNIQUE [KEYS]] | [form] NORMALIZED FROM after IS",
```

src/parser/mod.rs
```diff
@@ -4752,6 +4761,16 @@ impl<'a> Parser<'a> {
...
+    fn parse_unquoted_word(&mut self, expected: &str) -> bool {
+        match &self.peek_token_ref().token {
+            Token::Word(w) if w.quote_style.is_none() && w.value.eq_ignore_ascii_case(expected) => {
+                self.advance_token();
+                true
+            }
+            _ => false,
+        }
+    }
+
```
LucaCappelletti94: You could drop `parse_unquoted_word`, since `parse_keyword(Keyword::DOCUMENT)` covers it. The tokenizer already gives quoted words `Keyword::NoKeyword`.

```suggestion
```

src/parser/mod.rs
```diff
@@ -12742,6 +12761,20 @@ impl<'a> Parser<'a> {
...
+        let start_index = self.index;
+        if self.parse_keyword(Keyword::NOT) && self.parse_unquoted_word("document") {
+            Some(true)
+        } else {
+            self.index = start_index;
+            None
+        }
+    }
+
```
LucaCappelletti94: You could drop `parse_is_document_predicate` along with the manual `self.index` rewind, since `parse_keywords` handles both.

```suggestion
```

LucaCappelletti94 (changes requested):
You should add `DOCUMENT` to the sorted keyword list. The PR does not touch this file, so this cannot be a suggestion.

```diff
     DO,
+    DOCUMENT,
     DOMAIN,
```

Issue #1354, opened 2024-07-27
support Unicode string literals prefixed with `U&`


lovasoa:
Postgres supports literal strings prefixed with `U&` which can include backslash-escaped Unicode code points:

> A variant of quoted identifiers allows including escaped Unicode characters identified by their code points. This variant starts with U& (upper or lower case U followed by ampersand) immediately before the opening double quote, without any spaces in between, for example U&"foo". (Note that this creates an ambiguity with the operator &. Use spaces around the operator to avoid this problem.) Inside the quotes, Unicode characters can be specified in escaped form by writing a backslash followed by the four-digit hexadecimal code point number or alternatively a backslash followed by a plus sign followed by a six-digit hexadecimal code point number. For example, the identifier "data" could be written as
> 
> U&"d\0061t\+000061"
>
> The following less trivial example writes the Russian word “slon” (elephant) in Cyrillic letters:
>
> `U&"\0441\043B\043E\043D"`
>
> If a different escape character than backslash is desired, it can be specified using the UESCAPE clause after the string, for example:
>
> `U&"d!0061t!+000061" UESCAPE '!'`
>
> The escape character can be any single character other than a hexadecimal digit, the plus sign, a single quote, a double quote, or a whitespace character. Note that the escape character is written in single quotes, not double quotes, after UESCAPE.
>
> To include the escape character in the identifier literally, write it twice.
>
> Either the 4-digit or the 6-digit escape form can be used to specify UTF-16 surrogate pairs to compose characters with code points larger than U+FFFF, although the availability of the 6-digit form technically makes this unnecessary. (Surrogate pairs are not stored directly, but are combined into a single code point.)
> 
> *https://www.postgresql.org/docs/current/sql-syntax-lexical.html*

Currently sqlparser-rs parses `U&'x'` as the binary operation `&` applied to an identifier u and a string literal. 

Initially reported in https://github.com/lovasoa/SQLpage/discussions/511

Other databases support the syntax: 
 - http://www.h2database.com/html/grammar.html#string
 - https://docs.vertica.com/23.4.x/en/sql-reference/language-elements/literals/string-literals/unicode-string-literals/
- https://docs.actian.com/vectorhadoop/5.0/index.html#page/SQLLang/String_Literals.htm

PR #2509, opened 2026-09-14
Postgres: Support U&"..." quoted identifiers and UESCAPE clause

BenSatori:
## Example

Before this PR, none of these valid Postgres queries parsed:

```sql
SELECT 1 AS U&"d\0061ta", * from customers;
SELECT * from U&"c\0075stomers" limit 10;
SELECT * from U&"c!0075stomers" UESCAPE '!' limit 10;
```

`U&"..."` Unicode-escaped quoted identifiers were not tokenized at all (only the `U&'...'` string literal form was supported), and the `UESCAPE '<char>'` clause that can follow either a `U&'...'` string or a `U&"..."` identifier to override the default `\` escape character was not implemented, even though the `UESCAPE` keyword already existed in `keywords.rs`.

## Fix

`U&"..."` identifiers now tokenize into a regular quoted identifier (decoded, e.g. `U&"d\0061ta"` -> `"data"`), and both `U&'...'` and `U&"..."` now support an optional trailing `UESCAPE '<char>'` clause.

See the [Postgres docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) (identifiers) and [UESCAPE docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE) (escape clause). Relates to #1354, which added the `U&'...'` string literal support this PR extends.

This PR was authored with the assistance of GitHub Copilot.


src/tokenizer.rs
```diff
@@ -2085,6 +2094,33 @@ impl<'a> Tokenizer<'a> {
...
+    /// Reads a Unicode string literal body introduced by the `U&` prefix, e.g. the
+    /// `\0061\0062\0063` in `U&'\0061\0062\0063'`, honoring a trailing `UESCAPE '<char>'`
+    /// clause if present.
+    /// See <https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE>
+    fn tokenize_unicode_single_quoted_string(
+        &self,
+        chars: &mut State,
+    ) -> Result<String, TokenizerError> {
+        let error_loc = chars.location();
+        let raw = self.tokenize_single_quoted_string(chars, '\'', false)?;
```
LucaCappelletti94: You should collapse the doubled quotes when `unescape` is off

```suggestion
        let mut raw = self.tokenize_single_quoted_string(chars, '\'', false)?;
        if !self.unescape {
            raw = raw.replace("''", "'");
        }
```

tests/sqlparser_postgres.rs
```diff
@@ -7205,6 +7205,29 @@ fn test_unicode_string_literal() {
     }
 }
 
+#[test]
+fn test_unicode_string_literal_uescape() {
+    // Custom escape character via UESCAPE, see the postgres docs example
+    pg_and_generic().expr_parses_to(r#"U&'d!0061t!+000061' UESCAPE '!'"#, "U&'data'");
+}
```
LucaCappelletti94: Just a red test relative to the other note.

```suggestion
}

#[test]
fn test_unicode_string_literal_no_unescape() {
    TestedDialects::new_with_options(
        vec![Box::new(PostgreSqlDialect {})],
        sqlparser::parser::ParserOptions::new().with_unescape(false),
    )
    .verified_expr("U&'a''b'");
}
```

LucaCappelletti94 (changes requested):

PR #2511, opened 2026-09-15
Update README.md

fahadullahshah261:
Added RisingWave to the users of sqlparser-rs


LucaCappelletti94:
RisingWave doesn't depend on this crate. Its parser is an in-tree fork of an old sqlparser-rs (https://github.com/risingwavelabs/risingwave/blob/main/src/sqlparser/README.md)

`sqlparser` only appears in its Cargo.lock transitively via DataFusion, which is already on the list. Closing. If RisingWave adds a direct dependency, reopen with a link to the Cargo.toml.

PR #2512, opened 2026-09-15
PostgreSQL: Add support for TABLE query command

BenSatori:
Adds support for PostgreSQL's `TABLE` query command, which is syntactic sugar for selecting all rows from a table.

Examples that now parse:

```sql
TABLE customers ORDER BY contact_name LIMIT 1;
TABLE ONLY customers;
TABLE customers *;
```

This also supports `TABLE` in query-expression positions, such as subqueries, through the existing query body parser.

PostgreSQL docs: https://www.postgresql.org/docs/current/sql-select.html

AI assistance was used in preparing this change.


tests/sqlparser_postgres.rs
```diff
@@ -9953,3 +9953,10 @@ fn parse_insert_by_name_keywords_as_table_and_alias() {
         statement => panic!("Expected INSERT statement, got: {statement:?}"),
     }
 }
+
+#[test]
+fn parse_table_command() {
+    pg().verified_stmt("TABLE customers ORDER BY contact_name LIMIT 1");
```
LucaCappelletti94: Maybe these should be `pg_and_generic`

tests/sqlparser_common.rs
```diff
@@ -4848,8 +4848,10 @@ fn parse_create_table_as_table() {
     let expected_query1 = Box::new(Query {
         with: None,
         body: Box::new(SetExpr::Table(Box::new(Table {
+            only: false,
```
LucaCappelletti94, line 4851: Maybe tests where these new attributes are not all `false` could help increase test coverage and avoid regressions.

LucaCappelletti94 (changes requested):

src/ast/query.rs
```diff
@@ -298,19 +298,30 @@ impl fmt::Display for SetQuantifier {
...
 #[cfg_attr(feature = "visitor", derive(Visit, VisitMut))]
 pub struct Table {
+    /// `ONLY` modifier before the table name.
+    pub only: bool,
     /// Optional table name (absent for e.g. `TABLE` command without argument).
     pub table_name: Option<Ident>,
     /// Optional schema/catalog name qualifying the table.
     pub schema_name: Option<Ident>,
+    /// Trailing `*` modifier after the table name.
+    pub with_asterisk: bool,
```
LucaCappelletti94: You should name this field `has_asterisk` to follow pre-existing standard.

```suggestion
    pub has_asterisk: bool,
```

src/ast/query.rs
```diff
@@ -298,19 +298,30 @@ impl fmt::Display for SetQuantifier {
...
+                write!(f, "ONLY ")?;
+            }
             if let Some(ref schema_name) = self.schema_name {
-                write!(f, "TABLE {}.{}", schema_name, table_name,)?;
+                write!(f, "{}.{}", schema_name, table_name,)?;
             } else {
-                write!(f, "TABLE {}", table_name)?;
+                write!(f, "{}", table_name)?;
+            }
+            if self.with_asterisk {
```
LucaCappelletti94: You should follow the `has_asterisk` rename here.

```suggestion
            if self.has_asterisk {
```

src/parser/mod.rs
```diff
@@ -15625,17 +15637,22 @@ impl<'a> Parser<'a> {
...
     pub fn parse_as_table(&mut self) -> Result<Table, ParserError> {
+        let only = self.parse_keyword(Keyword::ONLY);
         let first_name = self.parse_identifier()?;
         if self.consume_token(&Token::Period) {
             let second_name = self.parse_identifier()?;
             Ok(Table {
+                only,
                 table_name: Some(second_name),
                 schema_name: Some(first_name),
+                with_asterisk: self.consume_token(&Token::Mul),
```
LucaCappelletti94: You should follow the `has_asterisk` rename here.

```suggestion
                has_asterisk: self.consume_token(&Token::Mul),
```

src/parser/mod.rs
```diff
@@ -15625,17 +15637,22 @@ impl<'a> Parser<'a> {
...
             })
         } else {
             Ok(Table {
+                only,
                 table_name: Some(first_name),
                 schema_name: None,
+                with_asterisk: self.consume_token(&Token::Mul),
```
LucaCappelletti94: You should follow the `has_asterisk` rename here.

```suggestion
                has_asterisk: self.consume_token(&Token::Mul),
```

LucaCappelletti94 (changes requested):
I have not replaced all of the occurrences but basically standardize to `has_asterisk`.

PR #2515, opened 2026-09-16
PostgreSQL: Support CREATE FOREIGN TABLE

fmguerreiro:
Parses `CREATE FOREIGN TABLE [IF NOT EXISTS] name (columns) SERVER server [OPTIONS (...)]` ([docs](https://www.postgresql.org/docs/current/sql-createforeigntable.html)).

Split out of #2317 at @LucaCappelletti94's request. That PR now carries `CREATE FOREIGN DATA WRAPPER` only.

Table-level `OPTIONS` reuses the list parser pulled out of `parse_pg_create_server`, which #2317 also moves; whichever lands second needs a rebase there.

Per-column `OPTIONS (...)`, `INHERITS`, `PARTITION OF` and `ALTER FOREIGN TABLE` are out of scope; the rustdoc notes that.

Ported from the `pgmold-sqlparser` fork.


src/ast/spans.rs
```diff
@@ -3163,4 +3164,19 @@ WHERE id = 1
...
+        let mut test = SpanTest::new(dialect, sql);
+
+        // Ends at the option key, not the statement: a quoted option value is an
+        // Ident with an empty span, so it contributes nothing to the union.
+        let stmt = test.0.parse_statement().unwrap();
+        assert_eq!(
+            test.get_source(stmt.span()),
+            "ft (a INT) SERVER s OPTIONS (schema_name"
+        );
+    }
```
LucaCappelletti94: You should not pin known-wrong behaviour in tests. This span stops at `(schema_name` because `parse_identifier` gives quoted identifiers an empty span.

```suggestion
```

LucaCappelletti94 (changes requested):

src/ast/spans.rs
fmguerreiro: shoot, you're right! fixed in 4207b276

PR #2517, opened 2026-09-16
Add dialect labeler

### LINKED ISSUE #2506: Adding dialect labels
I was thinking that it could be nice to add dialect-based labels to the PRs, such as PostgreSQL, SQLite etc, so that reviewers familiar with each of these dialects can appropriately filter PRs. Maybe we can even add an auto-labeller bot.

@alamb @iffyio if you agree I will proceed to do so.

LucaCappelletti94:
This adds the first (in reality second) attempt at a dialect auto-labeller for PRs, based on the dialects identified from the PR title and body.

If no dialect tags are identifiable, it will post a comment asking the PR author to either add the dialect in the PR title/body, or reply to the bot with the dialects.

I have tested these functionalities in my fork, so I am tentatively hopeful it should work here.

Closes #2506.


.github/workflows/labeler/label_dialects.js
BenSatori, line 1: how about maria\s?db
?
and nitpicking but why mssql|ms sql and not ms\s?sql?

LucaCappelletti94: Mostly because I did not think of them, feel free to add them as suggestions and I will merge them in, otherwise I will add them later

LucaCappelletti94: Merged in

BenSatori (approved):
lgtm

adriangb (approved):
Thanks, LGTM!

Issue #2519, opened 2026-09-16
Continuous fuzzing roadmap toward OSS-Fuzz integration


LucaCappelletti94:
I believe having continuous fuzzer will help us a lot in the process more effectively the increasing number of PR and avoid merging regressions, other than of course discover current errors.

I propose the following sequence of PRs:

1. [x] Switching from honggfuzz to cargo-fuzz/libFuzzer harness. I agree that in the past honggfuzz was the goto solution, and it was also my favourite, but cargo-fuzz has improved a lot in the last few years and is required by OSS-Fuzz and ClusterFuzzLite. Done in #2520 
2. [x] ClusterFuzzLite PR gate + daily batch, same targets. Done in #2527 
3. [x] Adding a harness executing the parse -> `Display` -> re-parse loop. Done in #2528
4. [x] Actually committed seed corpus. Ready for review in #2530
5. [ ] Differential fuzzer for SQLite #2562 
6. [ ] Differential fuzzer for PostgreSQL #2563 
7. [ ] Differential fuzzer for DuckDB #2564 
8. [ ] Clean code of harnesses & rest suite by moving the list of all currently supported dialects in an utility function in the main library.
9. [ ] Bring up to speed the https://github.com/apache/datafusion-sqlparser-rs/pull/2193 Arbitrary PR
10. [ ] OSS-Fuzz application (`projects/datafusion-sqlparser-rs`) which I believe we may be approved for, given the reasonable user pool of this crate. **This PR should only be opened after we reach some fuzzer stability.**
11. [ ] Fuzz-byte-driven weighted recursive query generator (SparkFuzz/SQLsmith style)

alamb:
> Adding a harness executing the parse -> Display -> re-parse loop, plus an actually committed seed corpus.
That sounds good to me

In my mind, I would suggest startign small and targeted (rather than trying to automate everything all at once)

So for example, we could start with basic fuzzing in CI and more on demand fuzzing for longer runs.

LucaCappelletti94:
> In my mind, I would suggest startign small and targeted (rather than trying to automate everything all at once)

Yeah, maybe I should move point 3 at the end, since it is going to be the largest one by far.

> So for example, we could start with basic fuzzing in CI and more on demand fuzzing for longer runs.

Yes, this is point 5

LucaCappelletti94:
[...]

PR #2520, opened 2026-09-16
Fuzz: Replace honggfuzz harness with cargo-fuzz

LucaCappelletti94:
First step of #2519. Swaps honggfuzz for cargo-fuzz in the `fuzz` workspace, same all-dialect target, ASan on by default.


alamb (approved):
I tested this locally and it works. Thanks @LucaCappelletti94 

```shell
```

I did have to install nightly rust first

```shell
rustup default nightly
info: using existing install for nightly-aarch64-apple-darwin
info: default toolchain set to nightly-aarch64-apple-darwin

  nightly-aarch64-apple-darwin unchanged - rustc 1.99.0-nightly (969b803cb 2026-08-09)
```

LucaCappelletti94:
I am adding a note in the markdown that nightly needs to be installed so to avoid leaving it implicit to be rediscovered.

PR #2521, opened 2026-09-16
PostgreSQL: Support GROUP BY ALL and DISTINCT

BenSatori:
PostgreSQL accepts `ALL` or `DISTINCT` before a `GROUP BY` expression list, but the parser currently treats `ALL` as the complete clause. For example, `SELECT a FROM t GROUP BY ALL a` fails to parse.

This adds a dialect-gated group-by modifier AST representation and supports both documented forms while preserving the existing bare `GROUP BY ALL` syntax in other dialects.

PostgreSQL documentation: https://www.postgresql.org/docs/current/sql-select.html#SQL-GROUPBY

This contribution was created with AI assistance.


src/parser/mod.rs
```diff
@@ -13546,9 +13559,18 @@ impl<'a> Parser<'a> {
...
+                    GroupByExpr::ExpressionsWithModifier(modifier, exprs, modifiers)
+                }
+                (None, Some(exprs)) => GroupByExpr::Expressions(exprs, modifiers),
+                (Some(_), None) => {
+                    return parser_err!(
+                        "BUG: GROUP BY modifier requires expressions",
+                        self.peek_token_ref().span.start
+                    )
+                }
             };
```
LucaCappelletti94: You could drop the `(Some(_), None)` arm. It can never run, because a parsed modifier always leads to parsed expressions, and matching on `expressions` first makes the match exhaustive without it.

```suggestion
            let group_by = match (expressions, modifier) {
                (None, _) => GroupByExpr::All(modifiers),
                (Some(exprs), None) => GroupByExpr::Expressions(exprs, modifiers),
                (Some(exprs), Some(modifier)) => {
                    GroupByExpr::ExpressionsWithModifier(modifier, exprs, modifiers)
                }
            };
```

src/ast/query.rs
```diff
@@ -3819,6 +3840,10 @@ pub enum GroupByExpr {
     All(Vec<GroupByWithModifier>),
     /// `GROUP BY <expressions>` with optional modifiers.
     Expressions(Vec<Expr>, Vec<GroupByWithModifier>),
+    /// `GROUP BY ALL | DISTINCT <expressions>` with optional modifiers.
```
LucaCappelletti94: You should reference `[PostgreSQL]` in the text. A link reference definition with no reference renders nothing, so `cargo doc` output for `GroupByExpr` currently has no link to the PostgreSQL docs.

```suggestion
    /// `GROUP BY ALL | DISTINCT <expressions>` of [PostgreSQL] with optional modifiers.
```

src/ast/query.rs
LucaCappelletti94: You may want to update the enum doc, which still says `GROUP BY` has two forms now that `ExpressionsWithModifier` adds a third.

```suggestion
/// Represents the three syntactic forms that `GROUP BY` can take, `GROUP BY ALL`
/// with optional modifiers, ordinary `GROUP BY <exprs>` and `GROUP BY ALL | DISTINCT <exprs>`.
```

LucaCappelletti94 (changes requested):

PR #2522, opened 2026-09-16
Postgres: Support BETWEEN SYMMETRIC / ASYMMETRIC

BenSatori:
Before this fix, PostgreSQL's `BETWEEN SYMMETRIC` / `BETWEEN ASYMMETRIC` modifier wasn't parsed:

```sql
SELECT * FROM customers WHERE a NOT BETWEEN SYMMETRIC 10 AND 1;
```

This adds a `symmetric` field to `Expr::Between` and parses the optional `SYMMETRIC`/`ASYMMETRIC` keyword after `BETWEEN` (per the SQL standard / PostgreSQL docs: https://www.postgresql.org/docs/current/functions-comparison.html). `ASYMMETRIC` is the default behavior and is not round-tripped in the displayed SQL.

AI assistance disclosure: this change was drafted with GitHub Copilot.


src/parser/mod.rs
```diff
@@ -4457,8 +4457,14 @@ impl<'a> Parser<'a> {
...
 
-    /// Parses `BETWEEN <low> AND <high>`, assuming the `BETWEEN` keyword was already consumed.
+    /// Parses `[ASYMMETRIC | SYMMETRIC] <low> AND <high>`, assuming the `BETWEEN` keyword was already consumed.
     pub fn parse_between(&mut self, expr: Expr, negated: bool) -> Result<Expr, ParserError> {
+        let symmetric = if self.parse_keyword(Keyword::SYMMETRIC) {
+            true
+        } else {
+            let _ = self.parse_keyword(Keyword::ASYMMETRIC);
+            false
+        };
```
LucaCappelletti94: You should gate the modifier behind a dialect method. Consumed unconditionally, it breaks `a BETWEEN symmetric AND 5` in every dialect where `symmetric` is an ordinary identifier.

```suggestion
        let symmetric = self.dialect.supports_between_symmetric()
            && self.parse_one_of_keywords(&[Keyword::SYMMETRIC, Keyword::ASYMMETRIC])
                == Some(Keyword::SYMMETRIC);
```

tests/sqlparser_common.rs
```diff
@@ -20067,6 +20071,26 @@ fn parse_insert_by_name() {
...
+            assert!(!negated);
+        }
+        _ => unreachable!(),
+    }
+
+    // ASYMMETRIC is the default behavior and is not preserved as SQL text.
+    one_statement_parses_to(
+        "SELECT * FROM t WHERE 1 NOT BETWEEN ASYMMETRIC 2 AND 3",
+        "SELECT * FROM t WHERE 1 NOT BETWEEN 2 AND 3",
+    );
```
LucaCappelletti94: Some additional red tests.

```suggestion
    // See https://www.postgresql.org/docs/current/functions-comparison.html
    let supported = all_dialects_where(|d| d.supports_between_symmetric());
    match supported.verified_expr("1 BETWEEN SYMMETRIC 2 AND 3") {
        Expr::Between {
            symmetric, negated, ..
        } => {
            assert!(symmetric);
            assert!(!negated);
        }
        _ => unreachable!(),
    }

    // ASYMMETRIC is the default behavior and is not preserved as SQL text.
    supported.one_statement_parses_to(
        "SELECT * FROM t WHERE 1 NOT BETWEEN ASYMMETRIC 2 AND 3",
        "SELECT * FROM t WHERE 1 NOT BETWEEN 2 AND 3",
    );

    let unsupported = all_dialects_where(|d| !d.supports_between_symmetric());
    unsupported.verified_expr("a BETWEEN symmetric AND 5");
    unsupported.verified_expr("a BETWEEN asymmetric AND 5");
```

LucaCappelletti94 (changes requested):
You should add the `Dialect` trait method the parser suggestion calls:

```diff
     fn supports_in_empty_list(&self) -> bool {
         false
     }
+
+    /// Returns true if the dialect supports `BETWEEN [ ASYMMETRIC | SYMMETRIC ]`.
+    fn supports_between_symmetric(&self) -> bool {
+        false
+    }
```

Then, you should enable it for PostgreSQL and the generic dialect, where both keywords are reserved.

```diff
     fn supports_comment_optimizer_hint(&self) -> bool {
         true
     }
+
+    fn supports_between_symmetric(&self) -> bool {
+        true
+    }
 }
```

PR #2524, opened 2026-09-17
PostgreSQL: Parse ROWS FROM table functions

BenSatori:
Adds parsing and display support for PostgreSQL ROWS FROM table functions in FROM clauses, including WITH ORDINALITY and result aliases.

PostgreSQL docs: https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS

The docs also describe per-function column-definition lists inside ROWS FROM, which remain a separate syntax surface.

AI assistance was used to prepare this change.


src/parser/mod.rs
```diff
@@ -16813,6 +16795,67 @@ impl<'a> Parser<'a> {
...
+            args,
+            with_hints: vec![],
+            version: None,
+            partitions: vec![],
+            with_ordinality,
+            json_path: None,
+            sample: None,
+            index_hints: vec![],
+        })
+    }
```
LucaCappelletti94: You should parse each `ROWS FROM` item as a bare function call. I will provide some example red tests shortly.

```suggestion
    fn parse_rows_from_function(&mut self) -> Result<TableFactor, ParserError> {
        if self.parse_keyword(Keyword::UNNEST) {
            self.expect_token(&Token::LParen)?;
            let array_exprs = self.parse_comma_separated(Parser::parse_expr)?;
            self.expect_token(&Token::RParen)?;
            return Ok(TableFactor::UNNEST {
                alias: None,
                array_exprs,
                with_offset: false,
                with_offset_alias: None,
                with_ordinality: false,
            });
        }

        let name = self.parse_object_name(true)?;
        self.expect_token(&Token::LParen)?;
        let args = Some(self.parse_table_function_args()?);

        Ok(TableFactor::Table {
            name,
            alias: None,
            args,
            with_hints: vec![],
            version: None,
            partitions: vec![],
            with_ordinality: false,
            json_path: None,
            sample: None,
            index_hints: vec![],
        })
    }
```

tests/sqlparser_postgres.rs
```diff
@@ -10037,3 +10037,23 @@ fn parse_bitstring_literal_escaping() {
     pg_and_generic().verified_stmt("SELECT B''''");
     pg_and_generic().verified_stmt("SELECT B'it''s'");
 }
+
+#[test]
+fn parse_rows_from_with_ordinality() {
+    pg_and_generic().verified_stmt(
+        "SELECT * FROM ROWS FROM(UNNEST(ARRAY[1, 2]) WITH ORDINALITY) AS r (val, ord)",
+    );
```
LucaCappelletti94: You may want to turn this case into a negative one, since PostgreSQL rejects it with `syntax error at or near "WITH"`. The loop below fails on the current parser and passes with the change above.

```suggestion
    for sql in [
        "SELECT * FROM ROWS FROM(UNNEST(ARRAY[1, 2]) WITH ORDINALITY) AS r",
        "SELECT * FROM ROWS FROM(UNNEST(ARRAY[1]) WITH OFFSET) AS r",
        "SELECT * FROM ROWS FROM(generate_series(1, 2) WITH ORDINALITY) AS r",
        "SELECT * FROM ROWS FROM(generate_series(1, 2) AS g) AS r",
        "SELECT * FROM ROWS FROM(generate_series(1, 2) g) AS r",
    ] {
        assert!(pg_and_generic().parse_sql_statements(sql).is_err());
    }
```

LucaCappelletti94 (changes requested):

PR #2525, opened 2026-09-17
Postgres: Parse trailing asterisk after table name in FROM clause

BenSatori:
Before this PR, PostgreSQL's `FROM tbl_name*` syntax (asterisk explicitly indicating that descendant tables should be included) failed to parse:

```sql
SELECT * FROM tbl_name*
```

This was already supported for `TRUNCATE TABLE name*` and `LOCK TABLE name*`, but not for regular `FROM` items.

Docs: https://www.postgresql.org/docs/current/sql-select.html#SQL-FROM

## Fix
- Added a `has_trailing_asterisk` field to `TableFactor::Table`, parsed the same way as the existing asterisk handling in `TRUNCATE`/`LOCK TABLE`.
- Renamed the pre-existing `has_asterisk` fields on `TruncateTableTarget` and `LockTableTarget` to `has_trailing_asterisk` for naming consistency.

This PR was created with AI assistance.


src/ast/query.rs
```diff
@@ -2224,11 +2228,15 @@ impl fmt::Display for TableFactor {
...
                 index_hints,
+                has_trailing_asterisk,
             } => {
                 name.fmt(f)?;
                 if let Some(json_path) = json_path {
                     json_path.fmt(f)?;
                 }
+                if *has_trailing_asterisk {
+                    write!(f, "*")?;
+                }
```
LucaCappelletti94: You should render the `*` before `json_path`, matching the parse order. On Snowflake and Redshift (`supports_partiql`), `SELECT * FROM t*[0]` renders as `SELECT * FROM t[0]*`, which fails to parse back with `Expected: end of statement, found: *`.

```suggestion
                if *has_trailing_asterisk {
                    write!(f, "*")?;
                }
                if let Some(json_path) = json_path {
                    json_path.fmt(f)?;
                }
```

src/parser/mod.rs
```diff
@@ -16720,6 +16720,10 @@ impl<'a> Parser<'a> {
         } else {
             let name = self.parse_object_name(true)?;
 
+            // Postgres/Snowflake: `FROM tab*` explicitly includes descendant tables.
+            // https://www.postgresql.org/docs/current/sql-select.html#SQL-FROM
+            let has_trailing_asterisk = self.consume_token(&Token::Mul);
```
LucaCappelletti94: You could drop this comment. Snowflake's `FROM` grammar has no `tab*` form, and the `has_trailing_asterisk` field doc already links the Postgres reference.

```suggestion
            let has_trailing_asterisk = self.consume_token(&Token::Mul);
```

LucaCappelletti94 (changes requested):

PR #2526, opened 2026-09-18
Refactor: replace `dialect_of!` table partition check with `Dialect` trait method

jerguslejko:
## Summary

Continues the work from #2171 and #2175 by replacing one more `dialect_of!` macro usage with a `Dialect` trait method, so user-defined dialects can opt into table-factor `PARTITION` parsing.

**New trait method:**

| Method | Description | Dialects |
|--------|-------------|----------|
| `supports_table_partitions` | `PARTITION (p0, p1)` on a table factor | MySQL |

Also implemented for `GenericDialect` (returning `true`), matching the dialects the previous `dialect_of!(self is MySqlDialect | GenericDialect)` check named. The default is `false`, so behaviour is unchanged for every existing dialect.

Motivation is the one described in #1430 and #1186: a custom dialect wrapping `PostgreSqlDialect` cannot reach this branch at all today, because `dialect_of!` is a `TypeId` check that only matches built-in dialects.

## Test plan

- [x] All existing tests pass (1620 across 23 suites), including `parse_table_partition_selection`
- [x] New test in `tests/sqlparser_custom_dialect.rs`: a custom dialect opting in parses and round-trips `SELECT * FROM employees PARTITION (p0, p1)`; one that does not opt in still rejects it
- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes
- [x] `cargo fmt --all -- --check` passes


alamb (approved):
This is great -- thank you @jerguslejko

PR #2528, opened 2026-09-19
Fuzz: Display roundtrip

LucaCappelletti94:
Step 3 of #2519, stacked on #2527.

This PR adds the display roundtrip harness for all dialects.

A first 60 seconds run already finds `SELECT [a]]b]` in `MsSqlDialect` rendering as `SELECT [a]b]`, which fails to re-parse. That is the bug class of #2409, already being fixed in #2418, so this harness should help guide that PR.

`.clusterfuzzlite/build.sh` iterates over `cargo fuzz list`, so the new target is picked up by the CI builds with no workflow change.


LucaCappelletti94:
@alamb any notes on this one?

fuzz/fuzz_targets/fuzz_parse_roundtrip.rs
```diff
@@ -0,0 +1,59 @@
...
+use libfuzzer_sys::fuzz_target;
+use sqlparser::dialect::{
+    AnsiDialect, BigQueryDialect, ClickHouseDialect, DatabricksDialect, Dialect, DuckDbDialect,
+    GenericDialect, HiveDialect, MsSqlDialect, MySqlDialect, OracleDialect, PostgreSqlDialect,
+    RedshiftSqlDialect, SQLiteDialect, SnowflakeDialect,
+};
+use sqlparser::parser::Parser;
+
+fuzz_target!(|sql: &str| {
+    let dialects: [(&str, &dyn Dialect); 14] = [
```
alamb: it might be nice to use a pre-existing dialect enum for this (or add it somewhere) so that newly added dialects automatically get coverage

alamb (approved):
Thanks @LucaCappelletti94

fuzz/fuzz_targets/fuzz_parse_roundtrip.rs
LucaCappelletti94: For now, I made sure to extend the harnesses and include all of the dialects we are currently cover. I have spent a little while thinking about what could be the cleanest variant going forward but I am not too sure for now. Almost surely, it will be some code we move from the test suites into the lib under a test cfg.

alamb: Maybe just a function in sqlparser like
```rust
fn all_dialects() -> Vec<Box<dyn Dialect>>
```

or something 🤔

LucaCappelletti94: We already have such a function in the test suite, that is why I was saying it should be moved from there into the library proper.

PR #2530, opened 2026-09-19
Fuzz: Add curated seed corpus

LucaCappelletti94:
Step 4 of #2519, stacked on the roundtrip harness branch.

Adds 195 seed inputs in `fuzz/fuzz_seeds/`, with some from https://github.com/LucaCappelletti94/sql_ast_benchmark and others from the test suite.

`build.sh` now packages the directory as `<target>_seed_corpus.zip` next to each binary, which is the OSS-Fuzz and ClusterFuzzLite convention, so code-change and batch runs start from it instead of from random bytes. `fuzz/fuzz_seeds/*` is added to the RAT exclusions.

With this hot-start, the `fuzz_parse_sql` and `fuzz_parse_roundtrip` harnesses reach cov 10938 edges (previously 3238) and 12579 (previously 3360) respectively, in the same 30 seconds window.

The cluster CI job finally starts to actually mine bugs :)

PR #2531, opened 2026-09-19
PostgreSQL: Render prefix @ operator with a space

LucaCappelletti94:
`SELECT @ -2` fails to round-trip on PostgreSQL. The prefix `@` (`UnaryOperator::PGAbs`) renders with no space, so the AST displays as `SELECT @-2` and re-parsing aborts in the tokenizer with `Expected a valid binary operator after '@-'`, because `@-` is reserved as a geometric binary operator prefix. Found by the `fuzz_parse_roundtrip` target introduced in #2530.

`PGAbs` now renders with a trailing space, joining `NOT`, `#`, `@-@`, `@@`, `?-` and `?|`, which already use that arm for the same re-tokenization hazard, and matching the existing doc-comment example `@ -9`. The canonical output of `SELECT @a` becomes `SELECT @ a`.


alamb (approved):
Thank you for the small easy to review PRs

PR #2532, opened 2026-09-19
Fix unary minus rendering into -- line comments

LucaCappelletti94:
Fuzz round-trip found `SELECT - -1` rendering as `SELECT --1`, which re-parses as a line comment in every dialect except MySQL, and `SELECT - @2` rendering as `SELECT -@2` in PostgreSQL, where `-@` retokenizes as one custom operator and parsing fails. Same fusion hits `- ~1` (`-~1`), `- !!2` (`-!!2`), `- ||/2` (`-||/2`) and `- #x` (`-# x`).

`Expr::UnaryOp` display concatenates operator and operand in its final arm. The fix emits a space when the operator is `-` and the operand renders with a leading operator character. The first character is probed through a discarding write sink, so canonical output like `-1` or `-x` is unchanged and nothing allocates. Backtick-quoted operands and `-?` placeholders keep their current rendering because those dialects have no fusion rule and the characters stay out of the set.

There are some other variants of this error, but they are sufficiently distinct that I believe it is best to adress them in distinct PRs.


src/ast/mod.rs
```diff
@@ -1966,7 +1966,8 @@ impl fmt::Display for Expr {
                         | UnaryOperator::DoubleAt
                         | UnaryOperator::QuestionDash
                         | UnaryOperator::QuestionPipe
-                ) {
+                ) || (op == &UnaryOperator::Minus && starts_with_operator_char(expr))
```
alamb: Same comment as https://github.com/apache/datafusion-sqlparser-rs/pull/2533#discussion_r4061439474 -- I think if you restructured this code to match it would be easier to undertstand the behavior and coverage

alamb (approved):

alamb (approved):

alamb:
Doh! It has conflicts

PR #2533, opened 2026-09-19
Render prefix ~ operator with a space

LucaCappelletti94:
Fuzz round-trip found `SELECT ~ -1` rendering as `SELECT ~-1`, which PostgreSQL retokenizes as the single `~-` custom operator and fails to re-parse. `SELECT ~ ~1` renders `SELECT ~~1` and fails in every dialect, `~ @2` renders `~@2` and `~ #x` renders `~# x` in PostgreSQL.

`Expr::UnaryOp` display concatenates operator and operand in its final arm. `BitwiseNot` joins the spaced arm beside `Not` and `Hash`, the same remedy #2531 applies to `@`. Canonical output for `SELECT ~expr` becomes `SELECT ~ expr`, pinned with `one_statement_parses_to` in `parse_generic_unary_ops`.


src/ast/mod.rs
```diff
@@ -1961,6 +1961,7 @@ impl fmt::Display for Expr {
                 } else if matches!(
```
alamb: As a stylistic suggestion here to avoid issues in the future, I suggest changing the structure of this code to a sigle `match` so we can explicitly see which operators are handled in what form

like
```rust
match op {
  UnaryOperator::PGPostfixFactorial =>   write!(f, "{expr}{op}"),
  UnaryOperator::Not
                        | UnaryOperator::BitwiseNot
                        | UnaryOperator::Hash
                        | UnaryOperator::AtDashAt
                        | UnaryOperator::DoubleAt
                        | UnaryOperator::QuestionDash
                        | UnaryOperator::QuestionPipe =>  write!(f, "{op} {expr}"
  // TODO explcitly list out the other operators here so the list is explicitl
}
```

Also when I was typing that it seems like UnaryOperator::PGPostfixFactorial is the same as the else match 🤔

alamb (approved):
Thanks @LucaCappelletti94

src/ast/mod.rs
```diff
@@ -1955,24 +1955,23 @@ impl fmt::Display for Expr {
...
+                UnaryOperator::PGPostfixFactorial => write!(f, "{expr}{op}"),
+                UnaryOperator::Not
+                | UnaryOperator::BitwiseNot
+                | UnaryOperator::Hash
+                | UnaryOperator::AtDashAt
+                | UnaryOperator::DoubleAt
+                | UnaryOperator::PGAbs
+                | UnaryOperator::QuestionDash
+                | UnaryOperator::QuestionPipe => write!(f, "{op} {expr}"),
+                UnaryOperator::Plus
```
alamb, line 1968: ❤️

alamb (approved):

alamb:
Nice -- thank you @LucaCappelletti94

PR #2535, opened 2026-09-21
ClickHouse: Support FIRST | AFTER column position in ALTER TABLE

### LINKED ISSUE #2534: ClickHouse: `ALTER TABLE ... ADD COLUMN ... FIRST | AFTER col` fails to parse
```sql
ALTER TABLE t ADD COLUMN c UInt8 AFTER a
-- sql parser error: Expected: end of statement, found: AFTER
```

ClickHouse [supports this](https://clickhouse.com/docs/sql-reference/statements/alter/column), and the parser already handles it for MySQL. `parse_column_position` is hardcoded to `dialect_of!(MySqlDialect | GenericDialect)`, and this should become a `Dialect` trait method so ClickHouse can enable it.

s5dsn-eqee:
Closes #2534


LucaCappelletti94 (changes requested):
Given this feature is meant to be enabled also for MySQL, please add a test also for that dialect, expecially if its variant of the syntax has any quirks.

Otherwise all looks good.

LucaCappelletti94 (approved):

PR #2540, opened 2026-09-22
Fix `TABLE` command identifier quoting and token consumption

LucaCappelletti94:
This PR changes `Table` fields from `Option<String>` to `Option<Ident>` and updates [`TABLE` command](https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE) parsing to preserve identifier quoting and stop eagerly consuming trailing tokens.

Previously, `parse_as_table` discarded quote styles on table identifiers and unconditionally called `next_token()` three times, breaking roundtripping and trailing clauses:

```sql
-- Quoted table names had quotes stripped on display, breaking re-parse:
CREATE TABLE t AS TABLE `%mpty`;
-- Displayed as:  CREATE TABLE t AS TABLE %mpty
-- Re-parse error: ParserError("Expected: Table name, found: % at Line: 1, Column: 25")

-- Trailing clauses were swallowed:
CREATE TABLE t AS TABLE old_table ORDER BY x;
-- Parse error: ParserError("Expected: end of statement, found: x")

### CODE REVIEW

File: src/parser/mod.rs [hunk 1]
```diff
@@ -15625,44 +15625,16 @@ impl<'a> Parser<'a> {
...
-                Token::Word(w) => {
-                    table_name = w.value;
-                }
-                _ => {
-                    return self.expected("Table name", token3);
-                }
-            }
+        let first_name = self.parse_identifier()?;
+        if self.consume_token(&Token::Period) {
+            let second_name = self.parse_identifier()?;
```
alamb, line 15630: fixing bugs and deleting code is level 10 coding 👍

alamb (approved):
Thank you @LucaCappelletti94

PR #2544, opened 2026-09-22
Escape quotes when displaying `Word` tokens

LucaCappelletti94:
This PR ensures quotes in `Word` tokens are properly escaped on display, matching `Ident` display behavior and deduplicating identifier formatting into a shared `fmt_ident` helper.

```sql
REMOVE @````;
-- Formatted as: REMOVE @```
-- Re-parse error: Expected close delimiter '`' before EOF.
```


alamb (approved):
Thanks @LucaCappelletti94

